PR #13, opened 2021-01-06
Configured example app window size + hidden window toolbar.

matthew-carroll:
Configured example app window size + hidden window toolbar.


slightfoot (approved):
I tested on my MacMiniM1 and working great. I think we need a ticket to apply similar changes to the windows desktop so I can develop locally with similar expectations on layout.

Issue #5, opened 2021-01-05
Enter Content for new documents


bk-one:
**Context of use:** On app launch, if there is no content stored in the app yet
**Primary Actor:** User
**Triggers:** App launch

**Main Success Scenario:**
- [x] 1. The app window displays the `fallback content` "Enter your content" in Roboto at 13pt
- [x] 2. The color of the `fallback content` is `#C3C1C1`
- [x] 3. The `fallback content` is displayed directly underneath the title with a 10px vertical margin to the top
- [x] 4. The user clicks anywhere on the `fallback content`
- [ ] 5. The `fallback content` disappears, a vertical, blinking cursor is being shown (1px, black, same height as the content) at the very start of the content area. The content input area is 500px wide
- [x] 6. The user entered the text listed below 
- [ ] 7. At the end of the content input area, a new line is created. All text is left aligned.
- [x] 8. The text is displayed in Roboto at 13pt in `#312F2C` as shown below

**Alternative Success Scenario A:**
- [x] Steps 1. to 5. above apply
- [x] A-1: The user did not enter any text as the content and presses [ESC] or clicks elsewhere
- [x] A-2: The `fallback content` appears again.

**Alternative Success Scenario B:**
- [x] Steps 1. to 5. above apply
- [x] B-1: The user enters only whitespace characters* as the content and presses [ESC] or clicks elsewhere
- [x] B-2: The `fallback content` appears again.

**Notes:**
- The screenshot below expects a title to be set as defined in #4. This would also work if no title has been written yet
- The content can be edited, once the title is set of course. Feel free to implement this right away, but this can also be a separate ticket.
- The link in the user text should be displayed like any other text. There is no requirement to change the text to a link. This will be a separate ticket (it is still displayed as a link in this issue, though)
- For the time being, whitespace characters are only `U+0020` (Space), ` U+0009` (Tab), as well as `U+000A` and `U+000D` - unless there's an easy way to identify various whitespace characters easily.
- 

**User Text:**
Hanami (花見, "flower viewing") is the Japanese traditional custom of enjoying the transient beauty of flowers; flowers ("hana") are in this case almost always referring to those of the cherry ("sakura") or, less frequently, plum ("ume") trees. From the end of March to early May, cherry trees bloom all over Japan. The blossom forecast cherry blossom front is announced each year by the weather bureau - check https://www.jnto.go.jp/sakura/eng/index.php for details - and should be watched carefully by those planning hanami as the blossoms only last a week or two.

[...]

Issue #4, opened 2021-01-05
Enter Title for new documents


bk-one:
**Context of use:** On app launch, if there is no content stored in the app yet
**Primary Actor:** User
**Triggers:** App launch 

**Main Success Scenario:**
- [x] 1. The app window displays the `fallback title` "Enter your title" in Roboto at 34pt and bold
- [x] 2. The color of the `fallback title` is `#C3C1C1`
- [x] 3. The `fallback  title` is displayed 290px from the left border of the window and 70px from the top border
- [x] 4. The user clicks anywhere on the `fallback title`
- [ ] 5. The `fallback title` disappears, a vertical, blinking cursor is being shown (1px, black) at the very start of the title area. The title input area is 500px wide
- [x] 6. The user enters the text: "Visit Japan during Hanami"
- [x] 7. The text is displayed in Roboto at 34pt, bold, and in `#312F2C`

**Alternative Success Scenario A:**
- [x] Steps 1. to 5. above apply
- [x] A-1. the user did not enter any text as the title and presses [ESC] or clicks elsewhere
- [x] A-2. The `fallback title` appears again like in 1. above

**Additional Notes:**
- The content is stored locally, so when the app is closed and opened again, the title is displayed, so this use case doesn't apply anymore. This will be a separate ticket.
- The content can be edited, once the title is set of course. Feel free to implement this right away, but this can also be a separate ticket.
- All overflow is clipped (or whatever is easiest) for now on window resize

**Image Preview:**

![enter title](https://user-images.githubusercontent.com/37401/103669774-119bea80-4f79-11eb-8372-addc906d22cd.jpg)

![hanami](https://user-images.githubusercontent.com/37401/103670837-6c821180-4f7a-11eb-8a7a-6c0cdb877d90.jpg)

Issue #3, opened 2021-01-05
Mac Desktop App Window


bk-one:
**Context of use:** Start of the application on a Mac desktop computer
**Primary Actor:** User
**Triggers:** Application Start

**Main Success Scenario:**
1. On Application launch, the app displays a window with the dimensions of 1000x650 px
2. The title bar is transparent
3. The application title text is hidden
4. The background of the application window is white
5. The content view is "Full Size" (https://developer.apple.com/documentation/appkit/nswindow/stylemask/1644646-fullsizecontentview)

**Additional Notes:**
- This app will serve as the example app for the rich text editor. The app itself is not part of the rich text editor library and should either be a separate project or at least kept in a separate directory within the project.

Image Preview: 

![empty canvas](https://user-images.githubusercontent.com/37401/103666478-fc24c180-4f74-11eb-8afb-26d83bc91e26.jpg)

PR #14, opened 2021-01-06
#4 #5 example editor title and content entry

matthew-carroll:
#4 #5: Example editor title and content entry

**Merge #3 before merging this.**

Screen Shot 2021-01-05 at 6 38 35 PM
Screen Shot 2021-01-05 at 6 38 51 PM
Screen Shot 2021-01-05 at 6 39 02 PM

### DIFF
--- a/example/lib/example_editor.dart
+++ b/example/lib/example_editor.dart
@@ -1,3 +1,4 @@
+import 'package:flutter/cupertino.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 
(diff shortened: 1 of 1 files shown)


example/lib/example_editor.dart
```diff
@@ -0,0 +1,121 @@
...
+
+  Widget _buildContent() {
+    return RawKeyboardListener(
+      focusNode: _contentFocusNode,
+      onKey: (RawKeyEvent event) {
+        if (event.logicalKey == LogicalKeyboardKey.escape) {
+          _contentFocusNode.unfocus();
+        }
+      },
+      child: TextField(
```
bk-one, line 101: not sure how early we want to add abstractions, but shouldn't the sample app use widgets provided by the library package only? E.g. use a "EditContentWidget" or whatever we want to call it from the library, so we can update functionality in the package without touching the example app? At least that was my understanding - happy to discuss :)

example/lib/example_editor.dart
```diff
@@ -0,0 +1,121 @@
...
+    );
+  }
+
+  Widget _buildPage() {
+    return Column(
+      crossAxisAlignment: CrossAxisAlignment.stretch,
+      children: [
+        SizedBox(height: 70),
+        _buildTitle(),
+        _buildContent(),
```
slightfoot, line 58: If you type more than one line of content in the TextField the vertical alignment slightly changes. I wonder if wrapping this in an Expanded would help? Also perhaps we should wrap the Column in a SingleChildScrollView. Perhaps a different requirement.

example/lib/example_editor.dart
matthew-carroll: Once we introduce some kind of new functionality, we'll definitely put that in the package code. For now I put everything directly in the editor because it's plain Flutter code that's easier to work with directly in the example app.

example/lib/example_editor.dart
matthew-carroll: Committed some changes to handle these things.

slightfoot (approved):

PR #24, opened 2021-01-13
Spike: non-editable text selection.

matthew-carroll:
Spike: non-editable text selection.


bk-one:
First Functional Review:
- [ ] Tripple-click did not work for me
- [ ] double-click selects a whole word (as defined), but starting to drag with the mouse did change the selection from the whole word to the position where I started the drag. The whole word should stay selected. 
- [ ] In the same context, when initiating a double-click drag, I expect to select whole words and not character by character

slightfoot (approved):

PR #37, opened 2021-01-21
Spike custom text input

matthew-carroll:
This PR includes 2 spikes.

Spike 1: compare and contrast `TextInputClient` with `RawKeyboardListener`.

Spike 2: implement a single `RawKeyboardListener` editable text area.

For spike 2:

```
/// Spike:
/// Create a prototype of a custom text widget for reference purposes.
///
/// Features:
///  - displays I-beam cursor when hovering over text
///  - inserts caret where the user taps on text
///
///  - all latin characters, numbers, and special symbols can be entered into the field
///  - backspace key deletes the previous character when the selection is collapsed
///  - backspace key deletes all selected characters when the selection is not collapsed
///  - delete key deletes the previous character when the selection is collapsed
///  - delete key deletes all selected characters when the selection is not collapsed
///
///  - arrow keys move caret left, right, up, and down
///  - shift + arrow keys expands selection in desired direction
///  - shift + cmd + left/right selects by word
///  - shift + alt + left/right selects to beginning/end of line, respectively
///  - pressing left or right with a selection span causes the selection to collapse to a single caret
///
///  - drag selection of text
```

### DIFF
--- a/example/lib/spikes/custom_text_input/custom_text_input_spike.dart
+++ b/example/lib/spikes/custom_text_input/custom_text_input_spike.dart
@@ -1,0 +1,390 @@
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+
+/// Spike:
+/// Create a minimal widget implementation that directly handles text input.
+///
+/// This spike implements:
+///  - A widget that implements `TextInputClient` and opens a connection to
+///    the underlying platform. This approach seems to be specifically tailored
+///    to Android and iOS, where various adjustments are made to content
+///    by the platform.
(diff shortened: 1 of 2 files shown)


slightfoot:
If the compositing region is set the platform shell should handle the backspace and there should be no need to handle that ourselves.

Also, the RawKeyboard version only works with Hardware keyboards and mobile soft-keyboard inputs do not action those events.

slightfoot (approved):

matthew-carroll:
Simon and I spoke offline. We'll ignore the composing region because that's only needed for mobile.

I will make the other recommended changes in a followup spike.

PR #39, opened 2021-01-21
Add windows support to example.

slightfoot:
Add windows support to example project so I can run locally on my machine.


matthew-carroll (approved):
LGTM

Issue #36, opened 2021-01-20
Spike: limitations of spans in inline content.


slightfoot:
Spike: limitations of spans in inline content.

PR #40, opened 2021-01-21
Spike inline editing emoji

slightfoot:
`EditableText` and therefore `TextField` does not support `WidgetSpan` on any platform.

Tested WidgetSpan on Web and works.

#36

### DIFF
--- a/example/lib/spikes/inline/inline_spans.dart
+++ b/example/lib/spikes/inline/inline_spans.dart
@@ -1,0 +1,309 @@
+import 'package:flutter/material.dart';
+
+void main() {
+  runApp(
+    MaterialApp(
+      debugShowCheckedModeBanner: false,
+      home: InlineSpanSpike(),
+    ),
+  );
+}
+
+@immutable
+class InlineSpanSpike extends StatelessWidget {
+  const InlineSpanSpike({Key key}) : super(key: key);
+
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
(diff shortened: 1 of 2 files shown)


matthew-carroll:
Please rebase after your windows PR is merged.

example/lib/spikes/inline/inline_spans.dart
```diff
@@ -0,0 +1,309 @@
+import 'package:flutter/material.dart';
+
```
matthew-carroll, line 2: Please add a comment header like my other spikes that states the goal of the spike, and then provides any additional information and/or conclusions that resulted from the spike.

Future readers should be able to quickly glean why we ran a spike, and how it altered our course forward.

example/lib/spikes/inline/inline_spans.dart
```diff
@@ -0,0 +1,309 @@
...
+class InlineSpanSpike extends StatelessWidget {
+  const InlineSpanSpike({Key key}) : super(key: key);
+
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
+      appBar: AppBar(title: Text('Spike')),
+      body: Column(
+        children: [
+          ListTile(
```
matthew-carroll, line 22: Can you provide screenshots of these 2 screens, either in this PR or in Slack so that I can see what the code is doing without checking out this branch and building it, myself?

matthew-carroll:
This PR seems to be abandoned. I'm going to merge it as a reference because it's quarantined to its own directory.

PR #43, opened 2021-01-27
Spike: editor input, layout, selection, and editing

matthew-carroll:
This spike explores APIs and abstractions for displaying and editing a document.

This spike doesn't explore any event sourcing solution for document changes.


jamessuperlist (commented):
Thanks @matthew-carroll for the comprehensive documentation :)

Out of scope of this PR, but is opting into null-safety planned from the beginning of this project? Also I suspect printing to console will remain until first MVP?

example/lib/spikes/editor_input_delegation/layout/document_layout.dart
```diff
@@ -0,0 +1,318 @@
...
+    // print('Building document layout:');
+    final docComponents = _buildDocComponents();
+
+    return Column(
+      mainAxisSize: MainAxisSize.min,
+      crossAxisAlignment: CrossAxisAlignment.stretch,
+      children: [
+        for (final docComponent in docComponents) ...[
+          docComponent,
+          SizedBox(height: 16),
```
jamessuperlist, line 231: As a user, I would expect that click between paragraphs would enabled edit mode:

https://user-images.githubusercontent.com/75316882/106249363-9396d080-6212-11eb-83ed-09a3a7c6fa13.mp4

Without this spacer, editing feels more natural to me.

https://user-images.githubusercontent.com/75316882/106249637-e7a1b500-6212-11eb-9fc1-3cdf09ab3ad1.mp4

example/lib/spikes/editor_input_delegation/document/rich_text_document.dart
```diff
@@ -0,0 +1,215 @@
...
+/// A rich text document.
+///
+/// A `RichTextDocument` is comprised of a list of `DocumentNode`s,
+/// which describe the type and substance of a piece of content
+/// within the document. For example, a `ParagraphNode` holds a
+/// single paragraph of text within the document.
+///
+/// The purpose of the `RichTextDocument` is to facilitate the
+/// traversal of content within a rich text document, and edit
+/// the contents.
```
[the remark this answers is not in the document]
matthew-carroll: I've introduced a `DocumentEditor` class which will begin to take over this responsibility.

Whatever the eventual solution is, it can't be direct editing because we need to support undo/redo, and we'll need to deal with whatever complexities come up from integrating remote systems like Firebase.

example/lib/spikes/editor_input_delegation/document/rich_text_document.dart
```diff
@@ -0,0 +1,215 @@
...
+    if (nodeIndex >= 0 && nodeIndex < _nodes.length) {
+      _nodes.insert(nodeIndex + 1, newNode);
+    }
+  }
+
+  void deleteNodeAt(int index) {
+    if (index >= 0 && index < _nodes.length) {
+      _nodes.removeAt(index);
+    } else {
+      print('Could not delete node. Index out of range: $index');
```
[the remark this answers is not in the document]
matthew-carroll: This kind of logic is in heavy flux. It's probably not worth worrying about until we get to some kind of stabilization phase.

example/lib/spikes/editor_input_delegation/document/rich_text_document.dart
```diff
@@ -0,0 +1,215 @@
...
+  }
+}
+
+/// A singular position within a `RichTextDocument`.
+///
+/// The type of the `nodePosition` depends upon the type of
+/// `DocumentNode` that this position points to.
+class DocumentPosition<PositionType> {
+  const DocumentPosition({
+    this.nodeId,
```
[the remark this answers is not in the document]
matthew-carroll: We should cleanup all such details once the abstractions firm up. Until then, the constraints will just get in the way of refactoring.

example/lib/spikes/editor_input_delegation/layout/document_layout.dart
matthew-carroll: This is a matter of editor features, rather than core implementation. Right now, the desire seems to be to match something like Medium, which places space between paragraphs. If you want a newline without any space, you can use "shift+enter".

matthew-carroll:
I'm going to push a whole bunch of further changes to this PR tonight. Among them:

- All keyboard input is implemented via a priority-order list of rules to allow for extensibility.
- I've added a rule to prevent deletion of the title or 1st paragraph. This was implemented by way of the aforementioned keyboard rules. However, this approach needs to be re-evaluated because it probably belongs in some other phase of editing.
- All visual components are now built with a configurable builder.
- Hint text for an empty title and 1st paragraph. This was implemented by way of the aforementioned component builder.
- Rendering support for ordered lists, unordered lists, images, and horizontal rules.
- Initial rendering support for different types of text, e.g. paragraph vs header1.

I'm trying to continuously push the boundaries of what is here and then refactor the abstractions to a state that allows for desired extensibility.

example/lib/spikes/editor_abstractions/selectable_text/selectable_text.dart
```diff
@@ -0,0 +1,546 @@
...
+///
+/// `hasCursor` should be true when the text editing caret is sitting
+/// within this text area. This information is used to make certain
+/// widget decisions.
+///
+/// If `text` is empty, a `textSelection` with an extent >= 0 is provided, and
+/// `highlightWhenEmpty` is `true`, then `SelectableText` will paint a small
+/// highlight, despite having no content. This is useful when showing that
+/// one or more empty text areas are selected.
+class SelectableText extends StatefulWidget {
```
justinmc, line 19: FYI @chunhtai you might be interested in taking a look at this modification of SelectableText.

justinmc (commented):
I think the high level approach here is more or less how I'd imagined this being implemented.  

My biggest concern is the amount of divergence between this and the framework's codebase.  It may be hard to maintain this package as the framework's text editing code changes.

Ideally a lot of this code would be shared or baked into the framework.  Flutter's current text editing should maybe even be a subset of a rich editing approach like this.  I realize there is a lot of friction and breaking changes to start moving in that direction, though.

About control over the text selection toolbar mentioned in the [research doc](https://github.com/superlistapp/rich_text_editor/wiki/RESEARCH:-Can-we-use-existing-widgets%3F#editabletext): More control over this was recently added in https://github.com/flutter/flutter/pull/69428 and other PRs.  The interface of TextEditingControls is still overly opinionated and needs to be refactored, though.

You guys have probably seen these, but there have been other attempts to do somewhat related stuff in the community: https://github.com/Jjagg/boustro, https://github.com/memspace/zefyr

CC @longcatislooong if you want to take a look

matthew-carroll:
Hi @justinmc,

> I think the high level approach here is more or less how I'd imagined this being implemented.
Do you see any likely candidates for future issues within the current abstractions and APIs? Are there any obvious features that will likely conflict with this API?
> 
> My biggest concern is the amount of divergence between this and the framework's codebase. It may be hard to maintain this package as the framework's text editing code changes.
Doesn't the divergence work the opposite way? As you've seen in the wiki, I've outlined a number of reasons why none of the existing Flutter widgets with editing capability and selection ability are useful for our purposes. I built on top of the lowest level text-based widget that Flutter offers, which should therefore be the most stable, and most narrowly focused. While this divergence might result in more code for us to maintain in the package, and more features for us to test, to me this sounds like the easiest way to stay in sync with Flutter, not the most difficult.

On the other hand, if what you mean is that features beyond text editing will be out of sync, like a popup editing bar, etc, then I would suggest that this is a composition problem with Flutter. Widgets that pop up here and there for specialized text editing purposes should not be integrated with fundamental text editing widgets. This package should be able to refer to, and use, whatever editing widgets Flutter uses, without needing to buy into Flutter's entire `EditableText` or `SelectableText` ecosystem. I think in the area of text editing, Flutter has very much moved away from aggressive composition, towards a monolithic subsystem. I would recommend that the Flutter team work to address that.
> 
> Ideally a lot of this code would be shared or baked into the framework. Flutter's current text editing should maybe even be a subset of a rich editing approach like this. I realize there is a lot of friction and breaking changes to start moving in that direction, though.
Would you like to partner with us to move some of these low-level concepts and widgets into Flutter? Given all the factors at play with Flutter team development, I'm doubtful that any refactoring path of the existing widgets in Flutter will lead to good solutions within any reasonable period of time. Not only are various controls missing from the current API, like direct control over selection bounds/painting, but the current implementations also do too much, like implicitly incorporating scrolling. I think the odds are very low that these existing widgets ever become an ideal compositional tool for editable text. I think at best, they'll become a kind of Franken-monster living half in simple mobile world, and half in complex desktop world.
> 
> _Edit_ One other thing I thought about talking to @chunhtai. Has anyone thought of how to stringify a document for example when copy/pasting?
Did you see the current implementation of selection, copy, and paste in this package?

example/lib/spikes/editor_abstractions/core/document.dart
```diff
@@ -0,0 +1,118 @@
...
+///
+/// The `start` position must come before the `end` position in
+/// the document.
+class DocumentRange {
+  DocumentRange({
+    @required this.start,
+    @required this.end,
+  });
+
+  final DocumentPosition<dynamic> start;
```
goderbauer, line 48: In general, I'd avoid the use of "dynamic" in the API. It basically turns off type checking and that often leads to bugs,

goderbauer (commented):
This is a huge PR, so I haven't looked at all of it.

One high-level question I have: It seems (unless I missed something) that this will always layout and render the entire document (even if only a little bit is visible in the viewport). Presumably, that would be bad for performance if the document gets big. Have you thought about lazily rendering only what's visible?

example/lib/spikes/editor_abstractions/default_editor/editor.dart
```diff
@@ -0,0 +1,352 @@
...
+  paragraphBuilder,
+  unorderedListItemBuilder,
+  orderedListItemBuilder,
+  imageBuilder,
+  horizontalRuleBuilder,
+  unknownComponentBuilder,
+];
+
+/// Keyboard actions for the standard `Editor`.
+final defaultKeyboardActions = <DocumentKeyboardAction>[
```
goderbauer, line 329: Could this use Flutter's standard Shortcuts/Action mechanism?

example/lib/spikes/editor_abstractions/default_editor/document_interaction.dart
```diff
@@ -0,0 +1,759 @@
...
+  }
+
+  Widget _buildKeyboardAndMouseInput({
+    Widget child,
+  }) {
+    return Listener(
+      onPointerSignal: _onPointerSignal,
+      child: RawKeyboardListener(
+        focusNode: _rootFocusNode,
+        onKey: _onKeyPressed,
```
goderbauer, line 648: My understanding is this is also in charge of inserting characters into the document? What about virtual on-screen keyboards, e.g. on Android/iOS?

example/lib/spikes/editor_abstractions/core/document.dart
matthew-carroll: Can you reframe this as a recommended alternative? The choice to use `dynamic` for arbitrary document positions was intentional to support selection of arbitrary document content.

example/lib/spikes/editor_abstractions/default_editor/editor.dart
matthew-carroll: I'm not sure. I tried investigating the shortcut widget. It seemed overlay verbose and complicated and I couldn't figure out how to use it effectively. If you have any guides or tutorials that would show how to use it in this context, I'm open to giving it a try.

example/lib/spikes/editor_abstractions/default_editor/document_interaction.dart
matthew-carroll: For right now we are exclusively focused on desktop OS and desktop web. I decided early on that it wasn't worth the early effort to figure out how to adapt `TextClient` to the needs of a text editor. The mobile integration will eventually be implemented.

matthew-carroll:
@goderbauer 

> This is a huge PR, so I haven't looked at all of it.

I'm not sure if you received my initial email to Chris, but this is the place to start a review:
https://github.com/superlistapp/rich_text_editor/wiki/Super-Editor-Initial-Review

matthew-carroll:
> One high-level question I have: It seems (unless I missed something) that this will always layout and render the entire document (even if only a little bit is visible in the viewport). Presumably, that would be bad for performance if the document gets big. Have you thought about lazily rendering only what's visible?

I've considered that issue, but I decided that it's not an issue worth solving right now. All early use-cases that I'm aware of should work just fine with full layout on every frame. I get the impression that either we'll discover reasons why lazy layout isn't desired, or moving from full layout to lazy layout will end up being reasonably simple when such a thing becomes a priority.

matthew-carroll:
I'm going to merge this PR because it's a spike and it won't impact the package definition.

PR #44, opened 2021-02-10
Spike: copy and paste plain text.

matthew-carroll:
Spike: copy and paste plain text.


matthew-carroll:
Since reviews are not being done, and we're currently only dealing with spikes, I've temporarily turned off require approvals and I'm merging in this spike.

PR #45, opened 2021-03-08
Added SelectableText widget to the package.

matthew-carroll:
Added `SelectableText` widget to the package.

As part of this PR, the following changes were made:

- Configured the package to be non-null by default (NNBD)
- Copied and improved the `SelectableText` widget from the abstractions spike
- Refactored the demo entrypoint to support a menu of demos that we can expand moving forward
- Created a demo of `SelectableText` that demonstrates a variety of visual configurations

Screen Shot 2021-03-08 at 1 39 52 AM

### DIFF
--- a/example/lib/demo_selectable_text.dart
+++ b/example/lib/demo_selectable_text.dart
@@ -61,14 +61,14 @@ class _SelectableTextDemoState extends State<SelectableTextDemo> {
                 _buildDemo(
                   title: 'TEXT WITHOUT SELECTION OR CARET',
                   demo: SelectableText(
-                    richText: _demoText1,
+                    textSpan: _demoText1,
                   ),
                 ),
                 SizedBox(height: 24),
                 _buildDemo(
                   title: 'TEXT WITH CARET + COLLAPSED SELECTION',
                   demo: SelectableText(
-                    richText: _demoText1,
+                    textSpan: _demoText1,
                     textSelection: TextSelection.collapsed(offset: _demoText1.toPlainText().length),
                     showCaret: true,
                   ),
@@ -77,7 +77,7 @@ class _SelectableTextDemoState extends State<SelectableTextDemo> {
                 _buildDemo(
                   title: 'TEXT WITH LEFT-TO-RIGHT SELECTION + CARET',
                   demo: SelectableText(
-                    richText: _demoText1,
+                    textSpan: _demoText1,
                     textSelection: const TextSelection(baseOffset: 0, extentOffset: 12),
                     showCaret: true,
                   ),
@@ -86,7 +86,7 @@ class _SelectableTextDemoState extends State<SelectableTextDemo> {
                 _buildDemo(
                   title: 'TEXT WITH RIGHT-TO-LEFT SELECTION + CARET',
                   demo: SelectableText(
-                    richText: _demoText1,
+                    textSpan: _demoText1,
                     textSelection: TextSelection(
                         baseOffset: _demoText1.toPlainText().length,
                         extentOffset: _demoText1.toPlainText().length - 17),
@@ -102,7 +102,7 @@ class _SelectableTextDemoState extends State<SelectableTextDemo> {
                     showDebugPaint: true,
                     child: SelectableText(
                       key: _debugTextKey,
-                      richText: _demoText1,
+                      textSpan: _demoText1,
                       textSelection: TextSelection(baseOffset: 0, extentOffset: _demoText1.toPlainText().length),
                       textSelectionDecoration: TextSelectionDecoration(
                         selectionColor: Colors.yellow,
@@ -146,9 +146,8 @@ class _SelectableTextDemoState extends State<SelectableTextDemo> {
           padding: const EdgeInsets.symmetric(horizontal: 8, vertical: 4),
           decoration: BoxDecoration(
               color: Colors.red,
-              borderRadius: BorderRadius.only(
-                topLeft: Radius.circular(4),
-                topRight: Radius.circular(4),
+              borderRadius: BorderRadius.vertical(
+                top: Radius.circular(4),
               )),
           child: Text(
             title,
--- a/example/lib/demo_selectable_text.dart
+++ b/example/lib/demo_selectable_text.dart
@@ -1,0 +1,167 @@
+import 'package:flutter/material.dart' hide SelectableText;
+import 'package:flutter_richtext/flutter_richtext.dart';
+
+/// Demo of a variety of `SelectableText` configurations.
+class SelectableTextDemo extends StatefulWidget {
+  @override
+  _SelectableTextDemoState createState() => _SelectableTextDemoState();
+}
+
+class _SelectableTextDemoState extends State<SelectableTextDemo> {
+  final _demoText1 = TextSpan(
+    text: 'Super Editor',
+    style: TextStyle(
+      color: const Color(0xFF444444),
+      fontSize: 18,
+      height: 1.4,
+      fontWeight: FontWeight.bold,
+    ),
+    children: [
+      TextSpan(
+        text: ' is an open source text editor for Flutter projects.',
+        style: TextStyle(
+          color: const Color(0xFF444444),
+          fontSize: 18,
+          height: 1.4,
+          fontWeight: FontWeight.normal,
+        ),
+      ),
+    ],
+  );
+  final _debugTextKey = GlobalKey();
+
+  @override
+  Widget build(BuildContext context) {
+    return Center(
+      child: SingleChildScrollView(
+        child: SizedBox(
+          width: 600,
+          child: Padding(
+            padding: const EdgeInsets.symmetric(vertical: 48.0),
+            child: Column(
+              mainAxisSize: MainAxisSize.min,
+              crossAxisAlignment: CrossAxisAlignment.start,
+              children: [
+                _buildTitle('Selectable Text Widget'),
+                SizedBox(height: 24),
+                _buildDemo(
+                  title: 'EMPTY TEXT WITH CARET',
+                  demo: SelectableText.plain(
+                    text: '',
+                    textSelection: TextSelection.collapsed(offset: 0),
+                    showCaret: true,
+                    style: TextStyle(
+                      color: const Color(0xFF444444),
+                      fontSize: 18,
+                      height: 1.4,
+                    ),
+                  ),
+                ),
+                SizedBox(height: 24),
+                _buildDemo(
+                  title: 'TEXT WITHOUT SELECTION OR CARET',
+                  demo: SelectableText(
+                    richText: _demoText1,
+                  ),
+                ),
+                SizedBox(height: 24),
+                _buildDemo(
+                  title: 'TEXT WITH CARET + COLLAPSED SELECTION',
+                  demo: SelectableText(
+                    richText: _demoText1,
+                    textSelection: TextSelection.collapsed(offset: _demoText1.toPlainText().length),
+                    showCaret: true,
+                  ),
+                ),
+                SizedBox(height: 24),
+                _buildDemo(
+                  title: 'TEXT WITH LEFT-TO-RIGHT SELECTION + CARET',
+                  demo: SelectableText(
+                    richText: _demoText1,
+                    textSelection: const TextSelection(baseOffset: 0, extentOffset: 12),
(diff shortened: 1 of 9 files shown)


lib/src/infrastructure/selectable_text.dart
```diff
@@ -0,0 +1,903 @@
...
+  // non-trivial performance implications.
+  late int _cachedTextLength;
+  int get _textLength => _cachedTextLength;
+  void _updateTextLength() {
+    _cachedTextLength = widget.richText.toPlainText().length;
+  }
+
+  RenderParagraph? get _renderParagraph => _textKey.currentContext?.findRenderObject() as RenderParagraph;
+
+  // TODO: use TextPainter line height when Flutter makes the info available.
```
salihgueler, line 209: ❔ Do we have an open issue that we can track this?

lib/src/infrastructure/selectable_text.dart
```diff
@@ -0,0 +1,903 @@
...
+    return widget.textCaretFactory.build(
+      context: context,
+      renderParagraph: _renderParagraph!,
+      position: widget.textSelection.extent,
+      lineHeight: _lineHeight,
+      isTextEmpty: _textLength == 0,
+      showCaret: widget.showCaret,
+    );
+
+    // return CustomPaint(
```
salihgueler, line 476: ❓ Is this forgotten or to be used in next iterations? If it is forgotten, maybe we should remove it, if it is going to be used, let's add a todo so we can know the reason.

salihgueler (approved):
I am really impressed with the work you put here Matt. Great work! 🎉 

I don't see anything blocker. I just have a couple of questions to understand the code better.

example/lib/demo_selectable_text.dart
```diff
@@ -0,0 +1,167 @@
...
+      mainAxisSize: MainAxisSize.min,
+      crossAxisAlignment: CrossAxisAlignment.start,
+      children: [
+        Container(
+          padding: const EdgeInsets.symmetric(horizontal: 8, vertical: 4),
+          decoration: BoxDecoration(
+              color: Colors.red,
+              borderRadius: BorderRadius.only(
+                topLeft: Radius.circular(4),
+                topRight: Radius.circular(4),
```
miguelcmedeiros, line 151:
```suggestion
              borderRadius: BorderRadius.vertical(
                top: Radius.circular(4),
              ),
```

example/lib/main.dart
```diff
@@ -1,34 +1,244 @@
...
+                return Colors.grey.withOpacity(0.1);
+              }
+
+              return Colors.transparent;
+            }),
+            splashFactory: NoSplash.splashFactory,
+            foregroundColor:
+                MaterialStateColor.resolveWith((states) => isSelected ? Colors.white : const Color(0xFFBBBBBB)),
+            elevation: MaterialStateProperty.resolveWith((states) => 0),
+            padding: MaterialStateProperty.resolveWith(
```
miguelcmedeiros, line 227: We can probably drop the `resolveWith`'s since we are not doing anything with `states` at the moment.

lib/src/infrastructure/selectable_text.dart
```diff
@@ -0,0 +1,903 @@
...
+/// displayed. An explicit `showCaret` control is offered because multiple
+/// `SelectableText` widgets might be displayed together with a selection
+/// spanning multiple `SelectableText` widgets, but only one of the
+/// `SelectableText` widgets displays a caret.
+///
+/// If `text` is empty, and a `textSelection` with an extent >= 0 is provided, and
+/// `highlightWhenEmpty` is `true`, then `SelectableText` will paint a small
+/// highlight, despite having no content. This is useful when showing that
+/// one or more empty text areas are selected.
+// TODO: find a way to require fewer parameter for selection and caret painting
```
miguelcmedeiros, line 101: One alternative could be to remove the caret from this widget and have it as a separate widget that when used as an ancestor of `SelectableText`, would then paint it, given that `SelectableText` would have a field (callback) like `onTextPositionChanged` that gets called every time the selections changes.
It feels to some extent that they don't belong together since the caret usually indicates that the text is being edited, while `SelectableText` should not care about that aspect.
What do you think?

lib/src/infrastructure/selectable_text.dart
```diff
@@ -0,0 +1,903 @@
...
+/// for a text display.
+abstract class TextLayout {
+  /// Returns the `TextPosition` within this `TextLayout` that
+  /// is closest to the given `localOffset`, or -1 if the text
+  /// is not laid out yet.
+  TextPosition getPositionAtOffset(Offset localOffset);
+
+  /// Returns the (x, y) `Offset` closest to the given `position`, or
+  /// `null` if the text is not laid out yet.
+  Offset? getOffsetForPosition(TextPosition position);
```
miguelcmedeiros, line 16: This and the following methods are not used at the moment (the only one used is `getPositionAtOffset`).
With just using `getPositionAtOffset` seems to work fine for now and since they are not used, they could potentially be removed for now.

lib/src/infrastructure/selectable_text.dart
```diff
@@ -0,0 +1,903 @@
...
+      });
+    }
+
+    return Stack(
+      children: [
+        _buildTextSelection(),
+        _buildText(),
+        _buildTextCaret(),
+      ],
+    );
```
miguelcmedeiros, line 436: We could avoid the `if (_renderParagraph == null) return const SizedBox();` blocks in the following methods with:
```suggestion
    return Stack(
      children: [
        if (_renderParagraph == null) _buildTextSelection(),
        _buildText(),
        if (_renderParagraph == null) _buildTextCaret(),
      ],
    );
```

lib/src/infrastructure/selectable_text.dart
```diff
@@ -0,0 +1,903 @@
...
+  Widget _buildTextSelection() {
+    if (_renderParagraph == null) {
+      return SizedBox();
+    }
+
+    return widget.textSelectionDecoration.build(
+      context: context,
+      renderParagraph: _renderParagraph!,
+      selection: widget.textSelection,
+      isTextEmpty: _textLength == 0,
```
miguelcmedeiros, line 448: Since we pass `renderParagraph`, we could derive this value from `renderParagraph.text.toPlainText().isEmpty` (same for `textCaretFactory`)

miguelcmedeiros (commented):
Looking good 😎. I love the idea of the demo app 👍 
Left some minor comments and a proposal about the `SelectableText` API (selection vs caret).

example/lib/demo_selectable_text.dart
```diff
@@ -0,0 +1,167 @@
...
+                      fontSize: 18,
+                      height: 1.4,
+                    ),
+                  ),
+                ),
+                SizedBox(height: 24),
+                _buildDemo(
+                  title: 'TEXT WITHOUT SELECTION OR CARET',
+                  demo: SelectableText(
+                    richText: _demoText1,
```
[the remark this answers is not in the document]
matthew-carroll: Sure. I changed it to `textSpan`.

lib/src/infrastructure/selectable_text.dart
```diff
@@ -0,0 +1,903 @@
...
+
+  /// Returns the (x, y) `Offset` closest to the given `position`, or
+  /// `null` if the text is not laid out yet.
+  Offset? getOffsetForPosition(TextPosition position);
+
+  /// Returns the `TextPosition` at the start of the line that
+  /// contains the given `currentPosition`, or -1 if the text
+  /// is not laid out yet.
+  TextPosition getPositionAtStartOfLine({
+    required TextPosition currentPosition,
```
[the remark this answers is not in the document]
matthew-carroll: The library doesn't have any dependencies, so there shouldn't be any issue with turning on NNBD. And non-NNBD can still use an NNBD package.

lib/src/infrastructure/selectable_text.dart
```diff
@@ -0,0 +1,903 @@
...
+/// spanning multiple `SelectableText` widgets, but only one of the
+/// `SelectableText` widgets displays a caret.
+///
+/// If `text` is empty, and a `textSelection` with an extent >= 0 is provided, and
+/// `highlightWhenEmpty` is `true`, then `SelectableText` will paint a small
+/// highlight, despite having no content. This is useful when showing that
+/// one or more empty text areas are selected.
+// TODO: find a way to require fewer parameter for selection and caret painting
+class SelectableText extends StatefulWidget {
+  /// `SelectableText` that displays plain text (no text styles).
```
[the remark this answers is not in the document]
matthew-carroll: True. I adjusted the comment to say only a single text style.

lib/src/infrastructure/selectable_text.dart
matthew-carroll: We do now :)

https://github.com/superlistapp/rich_text_editor/issues/46

lib/src/infrastructure/selectable_text.dart
matthew-carroll: Good catch. Removed it.

example/lib/demo_selectable_text.dart
matthew-carroll: Sure. I made the change.

lib/src/infrastructure/selectable_text.dart
matthew-carroll: I think the concept of a caret it pretty heavily tied to the concept of selectable text, so I don't think it would buy anything to pull it out. I removed the TODO because it's not really a problem. Maybe we'll find a way to simplify the API over time, or maybe not. But for now it seems to get the job done.

lib/src/infrastructure/selectable_text.dart
matthew-carroll: This PR is the first of many that moves code from the editor abstractions spike to the package. All APIs will be needed.

lib/src/infrastructure/selectable_text.dart
matthew-carroll: Looks like a tradeoff either way, I think I'd opt to keep the primary tree simpler (without the if-statements) and then allow the other build methods to decide what to return.

lib/src/infrastructure/selectable_text.dart
matthew-carroll: `_textLength` caches the length of the text to avoid computing `toPlainText()` over and over.

example/lib/main.dart
matthew-carroll: What's the alternative to calling `resolveWith()`? I haven't really followed how these new widgets are supposed to be styled...

lib/src/infrastructure/selectable_text.dart
```diff
@@ -0,0 +1,903 @@
...
+          // Force another frame so that we can use the renderParagraph.
+        });
+      });
+    }
+
+    return Stack(
+      children: [
+        _buildTextSelection(),
+        _buildText(),
+        _buildTextCaret(),
```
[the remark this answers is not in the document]
matthew-carroll: I added a repaint boundary around the text caret builder. I also adjusted the blink controller to stop running animations when the text offset `< 0`.

matthew-carroll:
Thanks for the review. There's one open question in the comments but I don't think that needs to be a blocker. I'm going to merge this PR and hopefully have another PR up later today to continue the migration.

example/lib/main.dart
miguelcmedeiros: Same here, there have been some changes in the framework lately that I haven't catch up to :)
My previous remark was that we call `resolveWith()`, but we don't anything with `states`. We will get the same with:
```
foregroundColor: isSelected ? Colors.white : const Color(0xFFBBBBBB),
elevation: 0,
padding: const EdgeInsets.symmetric(horizontal: 16, vertical: 16),
```

lib/src/infrastructure/selectable_text.dart
miguelcmedeiros: The reasoning behind my comment is that we can have just text (can't edit and can't select as in `Text` widget), text that is selectable but not necessarily editable (no caret) and editable text (which is implicitly selectable). If anything, `SelectableText` should have selection handles.
As you mentioned, the API might need adjustments as we go (or not) and that should not be an issue :)

PR #47, opened 2021-03-12
Add AttributedSpans and AttributedText to the package

matthew-carroll:
`AttributedSpans` and `AttributedText` forms the basis for text styling within the editor. These concepts are conceivably applicable to all sorts of document rendering and editing, so they are placed within the `infrastructure` package.

A very simple logger implementation is added as well so that we can keep useful debug print statements without polluting clients' logs. I've kept this logger private to the package so that we can change it however we want moving forward.

I've included an interactive demo for `AttributedText`. Here is a screenshot:

Screen Shot 2021-03-11 at 4 55 39 PM

Eventually, I'm envisioning a kind of interactive story where users can see how an editor is built up from a single `Text` widget. I think this interactive education would help developers adopt the package with a deep understanding of how it works. For now, I'm just focused on individual demos, but they might eventually contribute to a single narrative.

### DIFF
--- a/lib/src/infrastructure/attributed_text.dart
+++ b/lib/src/infrastructure/attributed_text.dart
@@ -1,6 +1,9 @@
 import 'package:flutter/painting.dart';
 
 import 'attributed_spans.dart';
+import '_logging.dart';
+
+final _log = Logger(scope: 'AttributedText');
 
 /// Text with attributions applied to desired spans of text.
 ///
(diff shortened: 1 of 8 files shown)


lib/src/infrastructure/attributed_text.dart
```diff
@@ -0,0 +1,235 @@
...
+  ///
+  /// Any attributions that span `startOffset` are applied to all
+  /// of the inserted text. All spans that start after `startOffset`
+  /// are pushed back by the length of `textToInsert`.
+  AttributedText insertString({
+    required String textToInsert,
+    required int startOffset,
+    Set<dynamic> applyAttributions = const {},
+  }) {
+    _log.log('insertString', 'text: "$textToInsert", start: $startOffset, attributions: $applyAttributions');
```
salihgueler, line 141: I thought logger was for more using it in the example app. Do we want to keep logs in the library or are we going to remove them when we release it?

salihgueler (approved):
Looks good to me!

lib/src/infrastructure/attributed_text.dart
matthew-carroll: These logs won't be printed unless a developer explicitly activates the logger. I left them in because they're useful for debugging. We can trim them down later, if desired, but I think we want some kind of option to keep useful logs that we can turn on, when needed.

salihgueler: Ah okay, maybe I missed it, thank you for clarification 👍

PR #53, opened 2021-03-16
Added core abstractions

matthew-carroll:
Added core abstractions including `Document`, `DocumentSelection`, `DocumentEditor`, `DocumentLayout`, `DocumentComposer`, and `EditContext`.

I thought about adding tests, but most of this code is either abstract, or only makes sense with implementations like `ParagraphNode`. I think our time will be better spent writing tests for real documents.

Here's where I stand with the various abstractions:

`Document` and `DocumentSelection`: These abstractions represent purely logical document details. I'm pretty confident about these APIs.

`DocumentEditor`: Responsible for altering a logical document. This implementation (and probably API) definitely need work. The `DocumentEditor` API will be relevant for incremental document saving as well as undo/redo behavior. I put off this work because I want to see what limitations we run into with something like Firebase before developing a thorough API.

`DocumentLayout` and related APIs: I'm pretty confident in the layout and rendering behaviors. There may be a few small missteps but I don't expect much to go wrong here.

`DocumentComposer` and `EditContext`: These are somewhat fringe concepts that seemed to be necessary when I built the spike. Their purpose in the system definitely needs to exist, but time will tell if these particular APIs and responsibility breakdowns make sense.

### DIFF
--- a/lib/src/core/document.dart
+++ b/lib/src/core/document.dart
@@ -1,0 +1,165 @@
+import 'package:flutter/foundation.dart';
+import 'package:flutter/widgets.dart';
+
+/// A read-only document with styled text and multimedia elements.
+///
+/// A `Document` is comprised of a list of `DocumentNode`s,
+/// which describe the type and substance of a piece of content
+/// within the document. For example, a `ParagraphNode` holds a
+/// single paragraph of text within the document.
+///
+/// New types of content can be added by subclassing `DocumentNode`.
+///
+/// To represent a specific location within a `Document`,
+/// see `DocumentPosition`.
+///
+/// A `Document` has no opinion on the visual presentation of its
+/// content.
+///
+/// To edit the content of a document, see `DocumentEditor`.
+abstract class Document with ChangeNotifier {
+  /// Returns all of the content within the document as a list
+  /// of `DocumentNode`s.
+  List<DocumentNode> get nodes;
+
+  /// Returns the `DocumentNode` with the given `nodeId`, or `null`
+  /// if no such node exists.
+  DocumentNode? getNodeById(String nodeId);
+
+  /// Returns the `DocumentNode` at the given `index`, or `null`
+  /// if no such node exists.
+  DocumentNode? getNodeAt(int index);
+
+  /// Returns the index of the given `node`, or `-1` if the `node`
+  /// does not exist within this `Document`.
+  int getNodeIndex(DocumentNode node);
+
+  /// Returns the `DocumentNode` that appears immediately before the
+  /// given `node` in this `Document`, or null if the given `node`
+  /// is the first node, or the given `node` does not exist in this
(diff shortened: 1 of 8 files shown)


lib/src/core/document.dart
```diff
@@ -0,0 +1,165 @@
...
+  /// the nodes at `position1` and `position2`.
+  List<DocumentNode> getNodesInside(DocumentPosition position1, DocumentPosition position2);
+}
+
+/// A span within a `Document` that begins at `start` and
+/// ends at `end`.
+///
+/// The `start` position must come before the `end` position in
+/// the document.
+class DocumentRange {
```
salihgueler, line 71: Maybe it is because it is a bit late for me 😄 But I don't get the difference between DocumentPosition and TextPosition in DocumentNode?

lib/src/core/document.dart
```diff
@@ -0,0 +1,165 @@
...
+  int get hashCode => nodeId.hashCode ^ nodePosition.hashCode;
+
+  @override
+  String toString() {
+    return '[DocumentPosition] - node: "$nodeId", position: ($nodePosition)';
+  }
+}
+
+/// A single content node within a `Document`.
+abstract class DocumentNode implements ChangeNotifier {
```
salihgueler, line 135: I think the DocumentNode can use a bit more explanation because it does a lot of heavy work below.

salihgueler (approved):

lib/src/core/edit_context.dart
```diff
@@ -0,0 +1,16 @@
+import 'document_composer.dart';
+import 'document_editor.dart';
+import 'document_layout.dart';
+
+/// Collection of core artifacts used to edit a document.
+class EditContext {
```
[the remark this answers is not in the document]
matthew-carroll: This is called `Context` because it's functioning as a service locator, which is historically called a `Context`. If you have a better name, please let me know.

lib/src/core/document.dart
matthew-carroll: A `DocumentPosition` is a position within an overall document (which contains many nodes). 

A node position is a position within a single node, like a position within a single paragraph. A `TextPosition` is used as a node position for text-based nodes. But for something like an `ImageNode` or a `HorizontalRuleNode`, a `TextPosition` would not make any sense as a node position, so those nodes use a different type of node position. That's why node positions are `dynamic`.

lib/src/core/document.dart
matthew-carroll: Can you outline the details that are unclear to you? I'd rather solve specific confusion points than just throw more words at the problem :)

matthew-carroll:
The 3 comments on this PR are all related to documentation and naming, which can be improved as we go, so I'm going to merge this PR and unblock myself for the next PR.

PR #57, opened 2021-03-19
Add first round of interactable editor widgets

matthew-carroll:
This PR adds what's necessary to display a bare bones editor with `ParagraphNode`s.

The only intended content is paragraphs, but there may be some ancillary behavior that came in due to references in the copied code. Achieving basic paragraph interaction required quite a bit of other pieces because any and all document editors need gesture and keyboard interaction, as well as layout around the document, itself.

@salihgueler I think this review should only look for major issues. Otherwise, I think it's preferable that after we merge this, you begin integrating an editor on your side so that we can discover any APIs that can be improved for standard use.

You can find instructions for configuring an editor here:
https://github.com/superlistapp/rich_text_editor/wiki/Example-code-snippets

After this PR is merged and you begin integrating, please file tickets that describe any bugs or any tedious APIs here:
https://github.com/superlistapp/rich_text_editor/projects/3?add_cards_query=is%3Aopen


salihgueler (approved):
Since this is really important for the initial state of the RTE I went through it without much detail. I don't see any problem. If I come up with something to improve etc. I will share it on the link Matt mentioned.

PR #58, opened 2021-03-22
Add focusNode and downgrade uuid 

salihgueler:
In the Superlist application, we use a huge Focus tree for navigating users between components and pages. Because of that, the focus node in the library mostly loses the race condition and loses focus. This causes the following:
- User can not enter a text as the focus moves out of the Editor automatically
- The paragraphs can not be changed.
- In paragraph keyboard navigation stops working.
- When a user writes a text, it writes it to other focused areas with a cursor in the paragraph.

I tried to solve it with a Focus widget but it did not work as I expected it to be as well. 

With the provided focusNode, we can make the Editor part of our focus mechanism and, since the focusNode value is optional, we won't be breaking any current behavior.

-------------------------------------------------------------------------------------------
This fixes the following issues:

- In the Superlist app right now, we are using https://pub.dev/packages/stream_chat and it is using the uuid version 2.2.2. On the other hand, Super Editor uses 3.0.1, which is why it causes conflict.
- The Superlist app uses a lot of Focus nodes and Focusing mechanisms with keyboard shortcuts, that is why having a reference to FocusNode at the DocumentInteractor was necessary to make the RTE running for us.

### DIFF
--- a/lib/src/default_editor/document_interaction.dart
+++ b/lib/src/default_editor/document_interaction.dart
@@ -31,14 +31,16 @@ final _log = Logger(scope: 'document_interaction.dart');
 ///  - automatically scrolls up or down when the user drags near
 ///    a boundary
 class DocumentInteractor extends StatefulWidget {
-  const DocumentInteractor({
+  DocumentInteractor({
     Key? key,
     required this.editContext,
     required this.keyboardActions,
     this.scrollController,
     required this.document,
     this.showDebugPaint = false,
-  }) : super(key: key);
+    FocusNode? focusNode,
+  })  : this.focusNode = focusNode ?? FocusNode(),
+        super(key: key);
 
   /// Service locator for other editing components.
   final EditContext editContext;
@@ -59,17 +61,20 @@ class DocumentInteractor extends StatefulWidget {
   /// debugging, when true.
   final showDebugPaint;
 
+  /// Helps the keyboard actions to navigate between nodes and communicate back
+  /// with the focus tree of the main app
+  final FocusNode focusNode;
+
   @override
   _DocumentInteractorState createState() => _DocumentInteractorState();
 }
 
-class _DocumentInteractorState extends State<DocumentInteractor> with SingleTickerProviderStateMixin {
+class _DocumentInteractorState extends State<DocumentInteractor>
+    with SingleTickerProviderStateMixin {
   final _dragGutterExtent = 100;
   final _maxDragSpeed = 20;
 
   final _documentWrapperKey = GlobalKey();
-
-  late FocusNode _rootFocusNode;
 
   late ScrollController _scrollController;
 
@@ -91,10 +96,10 @@ class _DocumentInteractorState extends State<DocumentInteractor> with SingleTick
   @override
   void initState() {
     super.initState();
-    _rootFocusNode = FocusNode();
     _ticker = createTicker(_onTick);
     _scrollController =
-        _scrollController = (widget.scrollController ?? ScrollController())..addListener(_updateDragSelection);
+        _scrollController = (widget.scrollController ?? ScrollController())
+          ..addListener(_updateDragSelection);
 
     widget.editContext.composer.addListener(_onSelectionChange);
   }
@@ -111,7 +116,8 @@ class _DocumentInteractorState extends State<DocumentInteractor> with SingleTick
       if (oldWidget.scrollController == null) {
         _scrollController.dispose();
       }
-      _scrollController = (widget.scrollController ?? ScrollController())..addListener(_updateDragSelection);
+      _scrollController = (widget.scrollController ?? ScrollController())
+        ..addListener(_updateDragSelection);
     }
   }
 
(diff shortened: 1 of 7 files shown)


matthew-carroll:
Hi @salihgueler, can you describe the specific use-cases or technical interactions that require a `FocusNode` in the `DocumentInteractor`? I'd like to make sure we're solving the full problem.

For UUID, the version I had was for NNBD. I haven't pulled down your branch, but have you checked to make sure that everything still builds and runs as expected?

salihgueler:
> Hi @salihgueler, can you describe the specific use-cases or technical interactions that require a `FocusNode` in the `DocumentInteractor`? I'd like to make sure we're solving the full problem.
> 
> For UUID, the version I had was for NNBD. I haven't pulled down your branch, but have you checked to make sure that everything still builds and runs as expected?

Explanation is updated. 

Tested the example app and check the library if it breaks something, all works as expected with the downgraded uuid version

matthew-carroll:
I have some followup questions. I'll add the UUId downgrade in my next PR so that you can get that change without waiting on this PR.

> User can not enter a text as the focus moves out of the Editor automatically

This sounds like an issue with the implementation of focus within the app. Is there any particular reason to believe that this issue sits with the editor widget, itself?

> The paragraphs can not be changed.

Can you describe what this means? Are you saying that pressing `tab` doesn't move between paragraphs? If so, that is very much intentional. In a text editor, `tab` should enter a tab character. It should not move the caret to some other paragraph.

> In paragraph keyboard navigation stops working.

Can you describe the reproduction steps for this issue? Is the problem actually about keyboard navigation, or is the problem that the editor loses focus and the keyboard no longer does anything?

> When a user writes a text, it writes it to other focused areas with a cursor in the paragraph.

This may be multiple issues. On the one hand, we should probably update the editor so that when it loses focus, the caret goes away. But, as with the first issue, if the editor loses focus unintentionally, that may be an issue with the app that is displaying the editor.

lib/src/default_editor/document_interaction.dart
```diff
@@ -122,7 +128,7 @@ class _DocumentInteractorState extends State<DocumentInteractor> with SingleTick
     if (widget.scrollController == null) {
       _scrollController.dispose();
     }
-    _rootFocusNode.dispose();
+    widget.focusNode.dispose();
```
roughike, line 131: Since this change makes the focus node come from outside, shouldn't the parent widget handle disposing it?

salihgueler:
> I have some followup questions. I'll add the UUId downgrade in my next PR so that you can get that change without waiting on this PR.
> 
> > User can not enter a text as the focus moves out of the Editor automatically
> 
> This sounds like an issue with the implementation of focus within the app. Is there any particular reason to believe that this issue sits with the editor widget, itself?
> 
> > The paragraphs can not be changed.
> 
> Can you describe what this means? Are you saying that pressing `tab` doesn't move between paragraphs? If so, that is very much intentional. In a text editor, `tab` should enter a tab character. It should not move the caret to some other paragraph.
> 
> > In paragraph keyboard navigation stops working.
> 
> Can you describe the reproduction steps for this issue? Is the problem actually about keyboard navigation, or is the problem that the editor loses focus and the keyboard no longer does anything?
> 
> > When a user writes a text, it writes it to other focused areas with a cursor in the paragraph.
> 
> This may be multiple issues. On the one hand, we should probably update the editor so that when it loses focus, the caret goes away. But, as with the first issue, if the editor loses focus unintentionally, that may be an issue with the app that is displaying the editor.

I will make a video about all the broken behavior and why this fix was needed and send the link here.

lib/src/default_editor/document_interaction.dart
salihgueler: As it might have its own focus node and needs to dispose it, because of that we keep it here.

roughike: What I mean is this:

```dart
class _MyWidgetState extends State<MyWidget> {
  FocusNode _focusNode;

  @override
  void initState() {
    super.initState();
    _focusNode = FocusNode();
  }

  @override
  void dispose() {
    // Editor already disposes it ¯\_(ツ)_/¯
    // _focusNode.dispose(); 
    super.dispose();
  }

  @override
    return Editor(
      // ...
      focusNode: _focusNode,
    );
  }
}
```

The above example to me is a bit weird - if `_MyWidgetState` creates the `FocusNode`, to me, it makes sense that it should also dispose of it. [The docs for FocusNode](https://api.flutter.dev/flutter/widgets/FocusNode-class.html) also mention this:

> If, as is common, the hosting StatefulWidget is also the owner of the focus node, then it will also call dispose from its State.dispose

and [docs for FocusNode.dispose()](https://api.flutter.dev/flutter/widgets/FocusNode/dispose.html):

> This method should only be called by the object's owner.

In the above example, nothing bad would happen, but I'm wondering if there's a case where this might lead to weird things.

salihgueler:
After discussing with Matt. We decided to handle this on another approach.

PR #71, opened 2021-03-29
Add an example usage of slivers and rich text editor.

salihgueler:
Fixes: https://github.com/superlistapp/rich_text_editor/issues/70

This PR adds a Sliver example of the Editor that has been created before with the following Slivers:
- SliverAppBar
- SliverToBoxAdapter
- SliverList

@matthew-carroll In the example, you can see the [focus problem](https://github.com/superlistapp/rich_text_editor/issues/72) by doing the following:

- Open the Sliver example from the drawer, tap on the paragraph (any point), the caret is correctly set but the keyboard actions are not working (user can not input any text even though the keyboard sound is there).

### DIFF
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -1,5 +1,6 @@
 import 'package:example/demo_selectable_text.dart';
 import 'package:example/example_editor.dart';
+import 'package:example/sliver_example_editor.dart';
 import 'package:flutter/material.dart';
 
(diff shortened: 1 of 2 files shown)


matthew-carroll (approved):
LGTM

matthew-carroll:
I'm going to merge this PR so that I can work on fixing this bug.

PR #76, opened 2021-04-02
Implemented markdown serialization and deserialization

matthew-carroll:
Implemented markdown serialization and deserialization with an example demo that shows realtime serialization.

Flutter wouldn't run the markdown tests with null-unsafe dependencies, so I had to upgrade them. @salihgueler is it possible for you to upgrade the overlapping dependencies on your end?

There are a number of tests in this PR for both markdown serialization and deserialization.

An example app demo is included that shows realtime markdown serialization side-by-side with the document in the editor.

https://user-images.githubusercontent.com/7259036/113379085-6dce9a80-932d-11eb-8ece-41fdae897b03.mov


salihgueler (approved):
Approving this with the consideration of the TODOs below. None of them are blockers at the moment.

lib/src/serialization/markdown.dart
```diff
@@ -0,0 +1,440 @@
+import 'package:flutter/cupertino.dart';
+import 'package:flutter_richtext/flutter_richtext.dart';
+import 'package:markdown/markdown.dart' as md;
+
+// TODO: return a regular Document instead of a MutableDocument.
+//       For now, we return MutableDocument because DocumentEditor
+//       requires one. When the editing system matures, there should
+//       be a way to return something here that is not concrete.
+MutableDocument deserializeMarkdownToDocument(String markdown) {
+  final markdownLines = markdown.split('\n');
```
roughike, line 10: Would [LineSplitter.split](https://api.dart.dev/stable/2.12.2/dart-convert/LineSplitter-class.html) be more "foolproof"?

In the markdown package, [they do some processing](https://github.com/dart-lang/markdown/blob/607653c7392f8a25deae7179afd3d63f10946cf7/lib/src/html_renderer.dart#L34) before calling `.split('\n')`. They also use LineSplitter.split [on one occasion](https://github.com/dart-lang/markdown/blob/607653c7392f8a25deae7179afd3d63f10946cf7/lib/src/html_renderer.dart#L100), so I'm just wondering if they encountered some bugs with just `.split('\n')` ¯\\_(ツ)_/¯

matthew-carroll:
@salihgueler do you know how to get the tests to run if we downgrade the dependencies? I couldn't figure it out.

Is there other mitigation strategy that would allow this package to use null-safe dependencies while Superlist uses the null-unsafe versions?

lib/src/serialization/markdown.dart
matthew-carroll: Couldn't hurt. I'll make the replacement and push an update.

PR #77, opened 2021-04-03
Link classes in comments

Arkangel12:
I notice when I was reading the code to get related to it, that its comments are pretty useful, but you were not able to jump into a named class because it was like this 'NameClass' instead of [NameClass] that made easier to navigate to the class.

Also, I found a couple of missed `@required` in a constructor at `core/document.dart`  _(this was the only code changed)_


example/macos/Flutter/GeneratedPluginRegistrant.swift
```diff
@@ -2,8 +2,6 @@
 //  Generated file. Do not edit.
 //
 
-// clang-format off
```
salihgueler, line 5: Why this was necessary?

lib/src/core/document.dart
```diff
@@ -3,67 +3,67 @@ import 'package:flutter/widgets.dart';
...
+  /// Returns the [DocumentNode] that appears immediately after the
+  /// given `node` in this [Document], or null if the given `node`
   /// is the last node, or the given `node` does not exist in this
-  /// `Document`.
+  /// [Document].
   DocumentNode? getNodeAfter(DocumentNode node);
 
-  /// Returns the `DocumentNode` at the given `position`, or `null` if
-  /// no such node exists in this `Document`.
+  /// Returns the [DocumentNode] at the given `position`, or `null` if
```
salihgueler, line 49: we could also address the variables with the cornered brackets. When the comments get too big, we might need to go down to check it (if it is needed)

salihgueler (commented):
Great thinking in general. Are we sure that it is directing us to the correct pages?

example/macos/Flutter/GeneratedPluginRegistrant.swift
Arkangel12: let me double-check because the only thing I removed was a couple of @ in one file and add some braces, I do not remove anything, this file is auto-generated, Flutter does it itself.

lib/src/core/document.dart
Arkangel12: I thought so but I don't wanna do it without running to all of you because this was just me reading the code 😅 , I can do it later because I do think it is a good idea.

matthew-carroll:
@Arkangel12 can you resolve the conflicts that have appeared in the PR?

Arkangel12:
One of them it was remove by a codegen i don't think its a good idea add the this line. 

The second one will be done later today 👍🏽 ir is ok @salihgueler if I add the braces to the params also? Will be only for classes with large documentation

matthew-carroll:
This is a useful change, but I think it's going to keep conflicting for a little while. We're making some rapid and substantive changes to get ready for initial release. Maybe we should hold off on worrying about this until things stabilize a bit more.

Arkangel12:
Sure, i just made it while i was reading/learning the code.  

I actually get what you saying and it makes sense. Let's hold on this then 👍🏽.

salihgueler:
I am closing this as it is highly outdated and most points about this is addressed separately.

Issue #62, opened 2021-03-22
Padding should accept EdgeInsetsGeometry instead of EdgeInsets


salihgueler:
Right now the padding property within the document accepts EdgeInsets and it makes it not possible to use EdgeInsetsDirectional for the RTL languages. Using EdgeInsetsGeometry instead would make things simpler.

Issue #61, opened 2021-03-22
Each keyboard hit makes a sound


salihgueler:
Maybe it was there during the testing but while I write a text each keyboard hit with a character makes the invalid keyboard sound.

matthew-carroll:
I just sent an email to Justin on the Flutter team and CC'd you. Let's see if they have an answer.

PR #90, opened 2021-04-07
Changed Editor param from EdgeInsets to EdgeInsetsGeometry, fixed unhandled key error sound on desktop. (#61, #62)

matthew-carroll:
Changed Editor param from EdgeInsets to EdgeInsetsGeometry, fixed unhandled key error sound on desktop. (#61, #62)


Arkangel12:
Was this the sound it made when you type a new character into the editor? I notice that but i'm getting related with the code 😅 will take a look into this if is the fix for that 🙌🏽.

salihgueler (approved):

PR #93, opened 2021-04-07
Feature/Update to use pedantic 1.11

defuncart:
Updates to use pedantic version 1.11.0 over 1.8.0. This will result in more analyzer warnings that will need to be fixed, but is worth considering as the package supports dart 2.12+.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -1,4 +1,4 @@
-include: package:pedantic/analysis_options.1.8.0.yaml
+include: package:pedantic/analysis_options.1.11.0.yaml
 
 analyzer:
     exclude: [build/**]
--- a/example/lib/spikes/editor_abstractions/default_editor/paragraph.dart
+++ b/example/lib/spikes/editor_abstractions/default_editor/paragraph.dart
@@ -276,7 +276,7 @@ Future<void> _processUrlNode({
   @required String originalText,
   @required String url,
 }) async {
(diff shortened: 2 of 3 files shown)


Arkangel12:
to this been completed, it will be good to proceed and resolve all the `dart analyzer` issues at least all the ones set as `error` we can work `warnings` for later.

version update only I think it is not enough. let me know if you need help with this.

matthew-carroll:
@defuncart do you know if there is a log somewhere with the differences between those pedantic versions?

defuncart:
@matthew-carroll There is some info on the [change log](https://pub.dev/packages/pedantic/changelog), about 20 or so new rules added since 1.8.

defuncart:
@Arkangel12 Running `flutter analyze` on main reveals the following errors and warnings

Bildschirmfoto 2021-04-08 um 15 26 41

So actually this PR doesn't introduce any new errors, just new warnings (about 260 more).

Most of the errors stems from _example/integration_test_ as in _example/pubspec.yaml_ `integration_test` is commented out. I pushed a simple fix for the uri error as it is unclear if example should have integration_tests or not.

Feel free to close this PR if it is not applicable at this time.

matthew-carroll:
I'm gonna merge this in before I restructure the repo. We can deal with the warnings later.

matthew-carroll (approved):

PR #94, opened 2021-04-07
Downgrade libraries to prevent clash.

salihgueler:
This PR is not necessarily for merging. But this changes were required to make the RTE working on Superlist.

As the stream_chat library can not work with uuid right now, we need to lower down the versions of the libraries. 

After all the necessary downgrades, I was able to run the tests by running the following command:

```
flutter test --no-sound-null-safety
```

All 55 tests passed with this approach.

### DIFF
--- a/example/macos/Flutter/GeneratedPluginRegistrant.swift
+++ b/example/macos/Flutter/GeneratedPluginRegistrant.swift
@@ -2,8 +2,6 @@
 //  Generated file. Do not edit.
 //
 
-// clang-format off
-
 import FlutterMacOS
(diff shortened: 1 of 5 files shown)


matthew-carroll:
@salihgueler this is good to know for future tests. Based on our conversation it looks like we can move forward with this package being null-safe and Superlist using a slightly modified branch to downgrade dependencies. If that works for you, let's go ahead and close this PR and we can keep in mind the test command if we ever need it.

salihgueler:
Of course, but let's keep the branch for the reference :)

Issue #100, opened 2021-04-07
Styling is wrong when user types styled content


matthew-carroll:
Styling is wrong when user types styled content.

Repro steps:
- go into italics mode
- type some text
- export to markdown
- re-import

Problem:
Each character flips between bold, italics, bold, italics

Issue #99, opened 2021-04-07
Autoscroll when new content is added and also when caret moves


matthew-carroll:
Autoscroll when new content is added and also when caret moves

PR #103, opened 2021-04-08
Fix bugs: auto-scroll not working, styled text not combining spans

matthew-carroll:
Fixes a couple bugs:

Bug #99 - when adding new list items or new lines that go offscreen, auto-scroll does not move the viewport down to show the extent.

Bug #100 - when typing individual characters with a style, the style of the new character does not combine with the matching style of the character before it.


salihgueler (approved):

salihgueler:
I am merging this to pull these changes into the current Superlist

Issue #80, opened 2021-04-05
Rename source code to super_editor


matthew-carroll:
Originally this project was referred to as "rich_text_editor". We have since opted for a project name of "super_editor".

This ticket is to rename the source code project to "super_editor".

PR #104, opened 2021-04-08
Refactored to a mono-repo to support other sub-packages, and renamed module to super_editor. (#80)

matthew-carroll:
Refactored to a mono-repo to support other sub-packages, and renamed module to super_editor. (#80)

@salihgueler can you pull this branch, open it, and run it locally to make sure there are no weird structural artifacts?


matthew-carroll:
@salihgueler I'll probably need to merge this tonight even without an approval so that I can continue with issues that we're finding. No work can be done until this major refactor goes in.

PR #109, opened 2021-04-11
Clean Dart Analysis Messages

Arkangel12:
Reduce warnings to a manageable amount `10`

we can talk about the ones left to fix it later.


salihgueler:
Since this is outdated and we have not decided on which approach to follow on the code, I will also close this one.

Issue #98, opened 2021-04-07
Get rid of caret when editor loses focus


matthew-carroll:
When the `Editor` widget loses focus, it doesn't remove its caret. The caret should disappear.

Issue #97, opened 2021-04-07
Recreate internal composer when DocumentEditor changes within Editor in didUpdateWidget()


matthew-carroll:
When the `DocumentEditor` changes within an `Editor` widget, the `Composer` remains unchanged, which results in an erroneous `DocumentSelection`.

When the `DocumentEditor` changes, recreate the `Composer` or at least reset the `DocumentSelection`.

PR #112, opened 2021-04-12
Redo fix for #97 and #98 because mono-repo accidentally undid the work.

matthew-carroll:
Redo fix for #97 and #98 because mono-repo accidentally undid the work.


matthew-carroll:
I'm going to immediately merge this because it's a copy/paste of the previous solution to these issues. They were over-written by the mono-repo change.

Issue #107, opened 2021-04-10
Bold and Italics doesn't serialize correctly


matthew-carroll:
Reportedly, italics text with a space at the end is serialized back to normal text.

EDIT:
Repro steps:
 - select a word that is followed by a space
 - apply bold or italics

Buggy result:
The bold/italics includes the trailing space. The trailing space should appear after the bold/italics.

salihgueler:
After #114 still reproducable. I will share videos.

ismail-go:
In my case, still exist a problem in the serialization of the markdown styling.

If the markdown styled text has a space in the begining or end of the text it does not apply markdown styling to the text correctly.

For example;
**bold text**  -> applyed bold styling correctlty but,
**bold text ** -> it seems as it is in the markdown displaying. Cannot applying the markdown styling.

as it is in bold styling, it is also works wrong in the italic, strikethrough etc..

How can we solve the problem? @matthew-carroll @salihgueler @jezell @domesticmouse @bk-one

matthew-carroll:
@ismail-go please provide a set of reproduction steps and please provide your exact version of super_editor and super_editor_markdown

Issue #105, opened 2021-04-10
CMD+B and CMD+I style changes are not reflected in markdown


matthew-carroll:
Reported by Chris Sells:

 > ...when I use a keyboard shortcut like Cmd+b or Cmd+i the text is formatted correctly but the mark down isn't updated with * or _

I have yet to repro this issue locally.

salihgueler:
This might be related to the issue I have talked about. Bold was working for me though.

PR #114, opened 2021-04-12
AttributedText now notifies listeners when it changes. Also propogates up as a Document change.

matthew-carroll:
Adds `ChangeNotifier` to `AttributedText` so that whenever style changes occur within existing text, document listeners are updated.

I think this should address #105 and possibly #107


salihgueler (approved):
Merging this in favor of using it in project.

Issue #95, opened 2021-04-07
Allow non-ASCII characters


bk-one:
As a user I want to be able to enter non-ASCII characters that I need for my language, e.g.

German: äöüß
Spanish: ñ
Swedish: ø
etc.

So the editor should allow any UTF8 character that I need to use in my language. We can keep left-to-right and right-to-left directions of typing to a separate ticket.

**Expectations:**
- [ ] Open Editor
- [ ] I can enter at least western-language based characters (as seen above; as a first step before supporting all UTF8/16)

salihgueler:
@matthew-carroll After discussing with @bk-one, it is decided that this has the highest priority. Please take a look into it when you can :)

PR #122, opened 2021-04-14
Added support for non-English characters when entered via keyboard or alt-codes

matthew-carroll:
Added support for non-English characters when entered via keyboard or alt-codes. The OS long-press popup for 'e' and 'n' not yet supported. (#95)

On mac, apparently when you long-press "e" or "n", an OS popup appears that allows you to select alternative versions of those characters. That works (kind of) with `TextField`s on mac, but does not currently work with Super Editor. It might be the case that we need to fully support `TextInputClient` to get that effect. That essentially means we would need full mobile support to make that happen.


salihgueler (approved):
Looks good to me. This is also important, so I am merging this now.

Issue #59, opened 2021-03-22
CMD+A Behavior is missing


salihgueler:
## Use Case

As a user, I would like to be able to select all paragraphs by tapping CMD+A key combination. 

## Proposal

A user should be able to select all paragraphs with similar metadata by clicking CMD+A key combination. Assume the following `MutableDocument` nodes are given:

```
ParagraphNode(
  id: 'Title',
  text: AttributedText(
    text: 'Example Document',
  ),
  metadata: {
    'blockType': 'header1',
  },
),
ParagraphNode(
  id: 'paragraph1',
  text: AttributedText(
    text:
        'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Phasellus sed sagittis urna. Aenean mattis ante justo, quis sollicitudin metus interdum id. Aenean ornare urna ac enim consequat mollis. In aliquet convallis efficitur. Phasellus convallis purus in fringilla scelerisque. Ut ac orci a turpis egestas lobortis. Morbi aliquam dapibus sem, vitae sodales arcu ultrices eu. Duis vulputate mauris quam, eleifend pulvinar quam blandit eget.',
  ),
),
ParagraphNode(
  id: 'paragraph2',
  text: AttributedText(
      text:
          'Cras vitae sodales nisi. Vivamus dignissim vel purus vel aliquet. Sed viverra diam vel nisi rhoncus pharetra. Donec gravida ut ligula euismod pharetra. Etiam sed urna scelerisque, efficitur mauris vel, semper arcu. Nullam sed vehicula sapien. Donec id tellus volutpat, eleifend nulla eget, rutrum mauris.'),
),
``` 

As a user, I would expect to select all `ParagraphNode`s above when I do CMD+A regardless of its metadata information.

matthew-carroll:
It's not missing, it just doesn't exist yet ;)

Can you rewrite this ticket as a standard feature request, including any contextual or ancillary information about the desired behavior?

We will collect many such feature requests over time. They will need to be prioritized and then implemented in stack order.

salihgueler:
@matthew-carroll I updated the description, please let me know if it is clearer :)

PR #127, opened 2021-04-15
Add CMD+A Behavior

salihgueler:
Fixes https://github.com/superlistapp/super_editor/issues/59

Adds a new default behavior and selects all the text when the user clicks CMD+A


super_editor/lib/src/default_editor/document_keyboard_actions.dart
```diff
@@ -95,6 +95,29 @@ ExecutionInstruction pasteWhenCmdVIsPressed({
...
 }
 
+
+ExecutionInstruction selectAllWhenCmdAIsPressed({
+  required EditContext editContext,
+  required RawKeyEvent keyEvent,
+}) {
+  if (!keyEvent.isMetaPressed || keyEvent.character?.toLowerCase() != 'a') {
+    return ExecutionInstruction.continueExecution;
+  }
```
matthew-carroll, line 105: You also need a condition that checks that the document has at least 1 node.

super_editor/lib/src/default_editor/document_keyboard_actions.dart
```diff
@@ -95,6 +95,29 @@ ExecutionInstruction pasteWhenCmdVIsPressed({
   return ExecutionInstruction.haltExecution;
 }
 
+
+ExecutionInstruction selectAllWhenCmdAIsPressed({
```
matthew-carroll, line 99: Can you add tests for this? My last PR showed how you might test a keyboard action.

For this behavior, I'd expect to see tests that least cover:

- when meta is pressed, but the character isn't "a", it does nothing
- when "a" is pressed, but meta isn't pressed, it does nothing
- when the document is empty it does nothing
- it selects everything in a document with 1 node
- it selects everything in a document with 2 nodes
- it selects everything in a document with 3 nodes
- it selects everything in a multi-node document when the first node and the last node are images

matthew-carroll (commented):
I left a couple of comments

super_editor/lib/src/default_editor/document_keyboard_actions.dart
salihgueler: Check for nodes added.

super_editor/lib/src/default_editor/document_keyboard_actions.dart
salihgueler: Tests added for all scenarios.

super_editor/test/src/default_editor/shortcut_test.dart
```diff
@@ -0,0 +1,260 @@
...
+            ),
+          );
+        },
+      );
+      test(
+        'CMD + A selects three nodes (all nodes) from a variety of nodes of 3 node items',
+        () {
+          final _editContext = _createEditContext(
+            nodes: [
+              _generateParagraphNode(),
```
matthew-carroll, line 162: Based on a previous comment, I think this situation should start and end with an image. The purpose of that configuration is to ensure that a document selection works with a binary position at the beginning and a binary position at the end, while also ensuring that nothing goes wrong when selecting across more than 2 nodes.

super_editor/test/src/default_editor/shortcut_test.dart
```diff
@@ -0,0 +1,260 @@
...
+            _editContext.composer.selection!.extent,
+            DocumentPosition(
+              nodeId: _editContext.editor.document.nodes.last.id,
+              nodePosition: _editContext.editor.document.nodes.last.endPosition,
+            ),
+          );
+        },
+      );
+      test(
+        'CMD + A selects five nodes (all nodes) from a variety of nodes of 5 node items',
```
matthew-carroll, line 199: This test probably doesn't need to exist. Beyond 3 nodes, the selection process is pretty much interchangeable. The purpose of the 3-node test was to ensure that having nodes on the inside of the selection that are not at either end would not interfere with the selection process. That's already verified by your earlier test. The types of nodes in between the start/end nodes don't have any impact on selection, so you're essentially repeating something you've already verified.

super_editor/test/src/default_editor/shortcut_test.dart
```diff
@@ -0,0 +1,260 @@
+import 'package:flutter/services.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:super_editor/super_editor.dart';
+import 'package:uuid/uuid.dart';
+
+import '../_document_test_tools.dart';
+import '_text_entry_test_tools.dart';
+
+void main() {
+  group(
```
matthew-carroll, line 10: Can you wrap this group with another group called "document_keyboard_actions.dart"? When we review failing tests, it may be helpful to quickly see what file they're testing. This won't be a permanent practice, but it's better than nothing until we find a replacement policy.

super_editor/test/src/default_editor/shortcut_test.dart
```diff
@@ -0,0 +1,260 @@
+import 'package:flutter/services.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:super_editor/super_editor.dart';
```
matthew-carroll, line 3: This is a collection of unit tests. Can you import the specific file you're testing rather than the whole editor?

matthew-carroll (commented):
I added a few more comments. Most of them are about testing practices.

matthew-carroll:
I updated the PR based on my earlier comments. I'm gonna merge it in.

matthew-carroll (approved):

PR #128, opened 2021-04-17
Fix: bold/italics style toggling was adding an extra character to the styled text at the end. (#107)

matthew-carroll:
Fix: bold/italics style toggling was adding an extra character to the styled text at the end. (#107)


salihgueler (approved):
Merging this in as it is crucial for the Superlist app.

PR #129, opened 2021-04-18
Add blockquote support

matthew-carroll:
Adds support for blockquotes, including markdown serialization.

Screen Shot 2021-04-18 at 4 31 37 PM


salihgueler (approved):
I would like to add this to the app today. That is why I am going to be merging it but, I think we should write some tests for it :)

Issue #126, opened 2021-04-14
[bug] Inserting horizontal divider in the same line as a paragraph deletes the whole paragraph


osaxma:
https://user-images.githubusercontent.com/46427323/114781017-c9af0f80-9d80-11eb-8a54-a2f543304c1a.mov

### To reproduce:
- Using `example_editor.dart` 
- place cursor at the beginning of the paragraph
- type three dashes '-' then a space. 

### Expected Behavior:
- The text should not be deleted. 
- Either the paragraph has to move below the divider, or the divider should not be inserted when there's a text after it.

matthew-carroll:
Thanks for filing this. I appreciate the video reference. I'll prioritize this bug to be fixed.

@osaxma in your opinion, what should happen when those dashes are added before existing text?

@salihgueler I'd also like to get your opinion on this. What should happen when 3 dashes are added before existing text?

osaxma:
I believe the divider should be inserted in that line and the paragraph should move one line below.

Here's an example from Notion:

https://user-images.githubusercontent.com/46427323/115938628-0cb96300-a4a4-11eb-9009-c5afc6dded6a.mov

Issue #123, opened 2021-04-14
Paragraph does not take entire width for short sentences


osaxma:
Paragraph does not take entire width for short sentences which breaks the alignment of the paragraph.

### To demonstrate:
<p align="center">
<img width="522" alt="Screen Shot 2021-04-14 at 3 19 43 PM" src="https://user-images.githubusercontent.com/46427323/114709104-d9086b80-9d34-11eb-888b-fafcb823906e.png"></p>

Though, if the first sentence of the paragraph is long enough to fit the entire width, the problem disappears. That can be observed by typing on the second node until the entire line is covered, then insert new lines using `shift+enter`. 

### To reproduce: 
- Using the `example_editor` demo, replace the mutable document with: 
```dart
Document _createInitialDocument() {
  return MutableDocument(
    nodes: [
      ParagraphNode(
        id: DocumentEditor.createNodeId(),
        text: AttributedText(
          text:
              'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Phasellus sed sagittis urna. Aenean mattis ante justo, quis sollicitudin metus interdum id. Aenean ornare urna ac enim consequat mollis. In aliquet convallis efficitur. Phasellus convallis purus in fringilla scelerisque. Ut ac orci a turpis egestas lobortis. Morbi aliquam dapibus sem, vitae sodales arcu ultrices eu. Duis vulputate mauris quam, eleifend pulvinar quam blandit eget.',
        ),
        metadata: {'textAlign': 'center'},
      ),
      ParagraphNode(
          id: DocumentEditor.createNodeId(),
          text: AttributedText(
            text: ' Lorem ipsum dolor.       ',
          ),
          metadata: {'textAlign': 'center'}),
      ParagraphNode(
          id: DocumentEditor.createNodeId(),
          text: AttributedText(
            text: ' Lorem ipsum dolor sit amet..     ',
          ),
          metadata: {'textAlign': 'right'}),
       ParagraphNode(
          id: DocumentEditor.createNodeId(),
          text: AttributedText(
            text: ' Multi-line paragraph \nsecond \nthird \n4th  ',
          ),
          metadata: {'textAlign': 'center'}),
    ],
  );
}
```

### Additional Context:
In the image above:
- Light Red color is a `Container` wrapped around:
https://github.com/superlistapp/super_editor/blob/78556f359dbe2425e58ccc03676c345091cda9a3/super_editor/lib/src/default_editor/paragraph.dart#L538

[...]

Issue #118, opened 2021-04-13
Copy and Paste does not work properly in the editor


salihgueler:
Right now when you select a word and copy it, it copies the whole paragraph after the selected word.

Steps to reproduce:

- Run the Example app 
- Select and Copy the word "consectetur"
- Paste it to the end
- It pastes everything after the word that has been selected

Issue #115, opened 2021-04-13
Fill the empty spaces while dragging the selection in the same paragraph


salihgueler:
When we drag and select words, characters in the paragraph, it shows empty lines between the selected paragraph elements. But for the same paragraph, it should not show the empty space.

Current:
![Screen Shot 2021-04-13 at 12 44 09](https://user-images.githubusercontent.com/24432752/114532877-f286b600-9c55-11eb-81a5-0b89f610c855.png)

Reference (Medium):

![Screen Shot 2021-04-13 at 12 43 46](https://user-images.githubusercontent.com/24432752/114532827-e69af400-9c55-11eb-96d1-3830c9c32f27.png)

matthew-carroll:
The selection boxes are calculated by Flutter's underlying text layout system. I'm not sure what the best solution for this issue is at this time.

We could add some pixels to the height, but then what happens when the text is made larger or smaller?

Maybe the height needs to be multiplied by the `TextStyle` `height` property.

Some investigation is required.

PR #140, opened 2021-04-24
Bug fixes for #115, #118, #123, #126

matthew-carroll:
Fixes bug where selection rectangles leave gaps between lines. This problem comes from Flutter's text box calculations, so the solution for now is a hard-coded height adjustment.

Fixes bug where inserting an HR with "---" before some text would delete all following text.

Fixes bug where copying text within a single node would copy everything from the selected content to the end of the node.

Fixes bug where paragraphs that don't require full width will shrink their widget, resulting in incorrect layout for center and right aligned text that doesn't fill all available width.


salihgueler (approved):
LGTM! Merging this in to Superlist.

PR #146, opened 2021-05-01
Introduce a custom text field called SuperTextField

matthew-carroll:
This PR introduces a custom implementation for a text field that's called `SuperTextField`.

This work is initially intended to help Turtle. After it meets Turtle's needs then it will be maintained for general consumption.

This PR adds a demo screen to the `super_editor` demo app that shows a handful of use-cases for `SuperTextField`.

Screen Shot 2021-04-30 at 9 46 04 PM

### DIFF
--- a/super_editor/lib/src/infrastructure/composable_text.dart
+++ b/super_editor/lib/src/infrastructure/composable_text.dart
@@ -1,0 +1,55 @@
+import 'package:flutter/rendering.dart';
+
+/// Contract for widgets that include editable text.
+///
+/// Examples: paragraphs, list items, images with captions.
+///
+/// The text positions accepted by a [TextComposable] are [dynamic]
+/// rather than [TextPosition]s because a use-case might include
+/// complex text composition, like tables, which might choose to
+/// index positions based on cell IDs, or row and column indices.
+abstract class TextComposable {
+  /// Returns a [TextSelection] that encompasses the entire word
+  /// found at the given [textPosition].
+  ///
+  /// Throws an exception if [textPosition] is not the right type
+  /// for this [TextComposable].
+  TextSelection getWordSelectionAt(dynamic textPosition);
+
+  /// Returns all text surrounding [textPosition] that is not
+  /// broken by white space.
+  ///
+  /// Throws an exception if [textPosition] is not the right type
+  /// for this [TextComposable].
+  String getContiguousTextAt(dynamic textPosition);
+
+  /// Returns the text position that corresponds to a text location
+  /// that is one line above the given [textPosition], or [null] if
+  /// there is no position one line up.
+  ///
+  /// Throws an exception if [textPosition] is not the right type
+  /// for this [TextComposable].
+  dynamic getPositionOneLineUp(dynamic textPosition);
+
+  /// Returns the node position that corresponds to a text location
+  /// that is one line below the given [textPosition], or [null] if
+  /// there is no position one line down.
+  ///
+  /// Throws an exception if [textPosition] is not the right type
+  /// for this [TextComposable].
+  dynamic getPositionOneLineDown(dynamic textPosition);
+
+  /// Returns the node position that corresponds to the first character
+  /// in the line of text that contains the given [textPosition].
+  ///
+  /// Throws an exception if [textPosition] is not the right type
+  /// for this [TextComposable].
+  dynamic getPositionAtStartOfLine(dynamic textPosition);
+
+  /// Returns the node position that corresponds to the last character
+  /// in the line of text that contains the given [textPosition].
+  ///
+  /// Throws an exception if [textPosition] is not the right type
+  /// for this [TextComposable].
+  dynamic getPositionAtEndOfLine(dynamic textPosition);
+}
--- a/super_editor/lib/src/infrastructure/selectable_text.dart
+++ b/super_editor/lib/src/infrastructure/selectable_text.dart
@@ -196,12 +196,12 @@ class SelectableTextState extends State<SelectableText> implements TextLayout {
   }
 
   @override
-  Rect getCharacterBox(TextPosition position) {
+  TextBox getCharacterBox(TextPosition position) {
     if (_renderParagraph == null) {
-      return Rect.zero;
+      return TextBox.fromLTRBD(0, 0, 0, 0, TextDirection.ltr);
     }
 
-    return _renderParagraph!.getBoxesForSelection(TextSelection.collapsed(offset: position.offset)).first.toRect();
+    return _renderParagraph!.getBoxesForSelection(TextSelection.collapsed(offset: position.offset)).first;
   }
 
   @override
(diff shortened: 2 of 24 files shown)


matthew-carroll:
@salihgueler please have a look at this PR after you're done with #150 

I'd like to get #150 and this PR merged by sometime early in the day May 6, my time. That way I can push forward with remaining v0.1.0 work.

For this PR, I recommend running the example app and checking out the new demo that I created for `SuperTextField`. After that, if you'd like, I can walk through it with you.

super_editor/lib/src/infrastructure/selectable_text.dart
```diff
@@ -222,22 +224,42 @@ class SelectableTextState extends State<SelectableText> implements TextLayout {
...
+  @override
+  Offset getOffsetAtPosition(TextPosition position) {
     if (_renderParagraph == null) {
       throw Exception('SelectableText does not yet have a RenderParagraph. Can\'t getOffsetForPosition().');
     }
+    if (_renderParagraph!.debugNeedsLayout) {
+      // This condition was added because getOffsetForCaret() was throwing
+      // an exception when debugNeedsLayout is true. It's unclear what we're
+      // supposed to do at our level to ensure that condition doesn't happen
+      // so until we figure it out, we'll just return a zero Offset.
```
salihgueler, line 254: 👍🏼

super_editor/lib/src/infrastructure/text_layout.dart
```diff
@@ -0,0 +1,96 @@
...
+  TextPosition getPositionNearestToOffset(Offset localOffset);
+
+  /// Returns the [Offset] of the character at the given [position].
+  Offset getOffsetAtPosition(TextPosition position);
+
+  /// Returns a [List] of [TextBox]es that contain the given [selection].
+  List<TextBox> getBoxesForSelection(TextSelection selection);
+
+  /// Returns a bounding [Rect] for the character at the given [position].
+  Rect getCharacterBox(TextPosition position);
```
salihgueler, line 24: I would rename this as it gives an impression to return `TextBox` and not a `Rect`

```suggestion
  Rect getCharacterRect(TextPosition position);
```

super_editor/lib/src/infrastructure/text_layout.dart
```diff
@@ -0,0 +1,96 @@
...
+  TextPosition getPositionAtStartOfLine(TextPosition textPosition);
+
+  /// Returns the [TextPosition] that corresponds to the last character
+  /// in the line of text that contains the given [textPosition].
+  TextPosition getPositionAtEndOfLine(TextPosition textPosition);
+
+  /// Returns the `TextPosition` in the first line within this
+  /// `TextLayout` that is closest to the given `x`-value, or
+  /// -1 if the text is not laid out yet.
+  TextPosition getPositionInFirstLineAtX(double x);
```
salihgueler, line 47: What is the "X" here? Is it the position in X Cartesian axis?

super_editor/lib/src/infrastructure/text_layout.dart
```diff
@@ -0,0 +1,96 @@
...
+  TextSelection getSelectionInRect(Offset baseOffset, Offset extentOffset);
+
+  /// Returns a [TextSelection] that surrounds the given [startingPosition] and expands
+  /// outward until the given [expansion] chooses to stop expanding.
+  TextSelection expandSelection(TextPosition startingPosition, TextExpansion expansion, TextAffinity affinity);
+}
+
+typedef TextExpansion = TextSelection Function(String text, TextPosition startingPosition, TextAffinity affinity);
+
+const paragraphExpansionFilter = _paragraphExpansionFilter;
```
salihgueler, line 69: I am seeing this usage for the first time 😅  Could you please explain why don't we expose the function and do this operation instead?

super_editor/lib/src/infrastructure/composable_text.dart
```diff
@@ -0,0 +1,55 @@
...
+/// rather than [TextPosition]s because a use-case might include
+/// complex text composition, like tables, which might choose to
+/// index positions based on cell IDs, or row and column indices.
+abstract class TextComposable {
+  /// Returns a [TextSelection] that encompasses the entire word
+  /// found at the given [textPosition].
+  ///
+  /// Throws an exception if [textPosition] is not the right type
+  /// for this [TextComposable].
+  TextSelection getWordSelectionAt(dynamic textPosition);
```
salihgueler, line 17: I am not so sure if we should throw an exception without a proper definition of allowed types here. Can we put something to class documentation? I know we already have possibilities but if we let the user decide the type, then I don't think that throwing non-allowed types should be their concern more than ours.

salihgueler (commented):
Overall it looks good 👍🏼  Great work 👏🏼  I have some questions and I was wondering if we should add tests to SuperTextField and how it is created. WDYT?

super_editor/lib/src/infrastructure/selectable_text.dart
```diff
@@ -3,77 +3,79 @@ import 'package:flutter/rendering.dart';
...
-  /// rectangle formed by the span from `baseOffset` to `extentOffset`, or
-  /// a collapsed selection at -1 if the text is not laid out yet.
-  ///
-  /// The `baseOffset` determines where the selection begins. The
-  /// `extentOffset` determines where the selection ends.
-  TextSelection getSelectionInRect(Offset baseOffset, Offset extentOffset);
-}
+import 'text_layout.dart';
+
+// /// Queries and calculates text position and offset details
```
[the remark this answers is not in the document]
matthew-carroll: That was an oversight. I'll remove the code in the next commit.

super_editor/lib/src/infrastructure/text_layout.dart
matthew-carroll: Good point. I'm gonna change the return type to `TextBox` to give it symmetry with `getBoxesForSelection()`.

super_editor/lib/src/infrastructure/text_layout.dart
matthew-carroll: Yeah, the doc comment above the method refers to "x-value". That's the horizontal position. Is there another name that you would prefer?

super_editor/lib/src/infrastructure/text_layout.dart
matthew-carroll: I think I may have done this because I wasn't sure if the function could be listed as `const`, but I think it can, so I'll remove the variable and just leave the function public.

super_editor/lib/src/infrastructure/composable_text.dart
matthew-carroll: That was done so that something like a table could still be considered a `TextComposable`. It's mentioned in the class docs:

```
/// The text positions accepted by a [TextComposable] are [dynamic]
/// rather than [TextPosition]s because a use-case might include
/// complex text composition, like tables, which might choose to
/// index positions based on cell IDs, or row and column indices.
```

matthew-carroll:
@salihgueler I rebased against `main` and fixed some related issues. I made some changes based on your PR comments. And I added tests for the `SuperTextField` keyboard handlers and actions.

salihgueler (approved):
Looks good now, we can merge this. It can be used in our app as well :)

Issue #145, opened 2021-04-30
[bug] exception when inserting text at the beginning of the paragraph if the first character has attribution 


osaxma:
Steps to reproduce:

- Using `example/lib/demos/example_editor.dart`, apply style (e.g. bold or italic) to the first letter or word of a paragraph.
- place the cursor at the beginning of the word such as:
<p align="center">
<img width="463" alt="Screen Shot 2021-04-30 at 6 46 25 PM" src="https://user-images.githubusercontent.com/46427323/116720161-8beaf180-a9e4-11eb-89ce-0b9007aef820.png"></p>

- try typing, and the following exception will occur:

```
════════ Exception caught by services library ══════════════════════════════════
The following _Exception was thrown during a platform message callback:
Exception: Another AttributedSpans can only be appended after the final marker in this AttributedSpans. Final marker: [SpanMarker] - attribution: bold, offset: 0, type: SpanMarkerType.start

When the exception was thrown, this was the stack
#0      AttributedSpans.addAt
package:super_editor/…/infrastructure/attributed_spans.dart:377
#1      AttributedText.copyAndAppend
package:super_editor/…/infrastructure/attributed_text.dart:129
#2      AttributedText.insertString
package:super_editor/…/infrastructure/attributed_text.dart:168
#3      InsertTextCommand.execute
package:super_editor/…/default_editor/text.dart:543
#4      DocumentEditor.executeCommand
package:super_editor/…/core/document_editor.dart:38
...
════════════════════════════════════════════════════════════════════════════════
```

--- 
Another related exception that I came across as I'm filing the issue:
- select the first letter (i.e., 'L') of the paragraph.
- press <kbd>⌘</kbd><kbd>B</kbd> twice to bold and unbold. 
- the following exception will occur:

```
════════ Exception caught by services library ══════════════════════════════════
The following _Exception was thrown during a platform message callback:
Exception: Inconsistent attributions state. Found a `start` marker with no matching `end`.

When the exception was thrown, this was the stack
#0      AttributedSpans._isContinuousAttribution
package:super_editor/…/infrastructure/attributed_spans.dart:295
#1      AttributedSpans.toggleAttribution
package:super_editor/…/infrastructure/attributed_spans.dart:265
#2      AttributedText.toggleAttribution
package:super_editor/…/infrastructure/attributed_text.dart:88
#3      ToggleTextAttributionsCommand.execute
package:super_editor/…/default_editor/text.dart:512
[...]

Issue #148, opened 2021-05-04
README about what is supported and not supported for v0.1.0


matthew-carroll:
README about what is supported and not supported for v0.1.0

Issue #147, opened 2021-05-03
[bug] triple tap at the end of a TextNode throws `RangeError`


osaxma:
**Steps to reproduce:**
- Using `example/lib/demos/example_editor.dart`.
- Place the cursor at the end of a `TextNode` such as a paragraph  or a list item. 
- triple tap and the following exception will occur:

```
════════ Exception caught by gesture ═══════════════════════════════════════════
The following RangeError was thrown while handling a gesture:
RangeError (index): Invalid value: Not in inclusive range 0..438: 439

When the exception was thrown, this was the stack
#0      _StringBase.[] (dart:core-patch/string_patch.dart:255:55)
#1      _expandPositionToParagraph
package:super_editor/…/default_editor/text_tools.dart:83
#2      getParagraphSelection
package:super_editor/…/default_editor/text_tools.dart:60
#3      _DocumentInteractorState._selectParagraphAt
package:super_editor/…/default_editor/document_interaction.dart:371
#4      _DocumentInteractorState._onTripleTapDown
package:super_editor/…/default_editor/document_interaction.dart:285
...
Handler: "onTripleTapDown"
Recognizer: TapSequenceGestureRecognizer#1f8c4
════════════════════════════════════════════════════════════════════════════════
```

**Expected Behavior:** 
The entire text should be selected. 

The behavior is observed on the latest `main` with `Flutter (Channel stable, 2.0.6, on macOS 11.2.3 20D91 darwin-x64)`.

Issue #143, opened 2021-04-27
[BUG] Deserialization of the Blockquote creates also a paragraph node


salihgueler:
STR:
- Checkout https://github.com/superlistapp/super_editor/tree/reproduce_ondemand_serialization_deserialization branch
- Run the example app
- Go to "Ondemand Serialization Demo" tab
- Remove the text 
- Add a blockquote text
- Click on Save and afterwards click on Update 

Expected:
- Should show only blockquote text
Actual:
- Shows blockquote and regular paragraph

PR #150, opened 2021-05-05
Multiple v0.1.0 issues

matthew-carroll:
#132 - fixes web bug that prints control character names when it shouldn't
#148 - added README updates that describe platform support
#142 - created a floating text format bar in the demo app. It's still missing link control and localized hint text.

EDIT - Added a couple more small fixes:
#145 - exception thrown when inserting text at start of styled text blob
#147 - exception thrown when triple tapping after final character in text blob
#143 - deserializing a markdown blockquote produces an extra paragraph node
#42 - localization for the example editor toolbar

### DIFF
--- a/super_editor/example/lib/demos/example_editor.dart
+++ b/super_editor/example/lib/demos/example_editor.dart
@@ -1,3 +1,5 @@
+import 'dart:ui';
+
 import 'package:flutter/cupertino.dart';
 import 'package:flutter/material.dart';
 import 'package:super_editor/super_editor.dart';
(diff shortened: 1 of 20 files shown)


super_editor/example/lib/demos/example_editor.dart
```diff
@@ -12,31 +15,495 @@ class ExampleEditor extends StatefulWidget {
...
+      composer: _composer,
+      scrollController: _scrollController,
+      documentLayoutKey: _docLayoutKey,
       maxWidth: 600,
       padding: const EdgeInsets.symmetric(vertical: 56, horizontal: 24),
     );
   }
 }
 
+class TextFormatBar extends StatefulWidget {
```
salihgueler, line 132: I don't understand why this is part of the example editor and not part of the library? If a person would like to use this, should they copy and paste this code to their codebase?

salihgueler (approved):
All good, we can merge this. I addressed my comments by talking to Ben :)

PR #156, opened 2021-05-07
Recomitting a fix for #123 because it was undone in a recent commit.

matthew-carroll:
Recomitting a fix for #123 because it was undone in a recent commit.


matthew-carroll:
I'm gonna force merge this because its a minor change that refixes a bug we already fixed.

Issue #152, opened 2021-05-05
[feature] characters with multiple codeUnits are not being handled properly  (e.g. flag emojis)


osaxma:
I'm not sure if this is considered a bug or unsupported feature, but I thought I should bring this to your attention. Also, I'm not certain if there are any keyboard character that contains multiple codeUnits, so I assume this problem will only occur if an external text was pasted into the editor.  

### Current Behavior 
Taking the US flag emoji `🇺🇸` as an example which has 4 code Units (i.e., `[55356, 56826, 55356, 56824]`), you can observe in the following video that it'll take `4` arrow presses to move the caret across the character. In addition, modifying the codeUnits (inserting a character in between the codeUnits or deleting any using the backspace) will causes the entire paragraph to disappear due to an `ArgumentError` in `RichText` (provided below the video).

https://user-images.githubusercontent.com/46427323/117168272-acde8880-add0-11eb-87d5-848dd0b863c4.mov

<details> <summary>Exception when modifying codeUnits</summary>

```
════════ Exception caught by rendering library ═════════════════════════════════
The following ArgumentError was thrown during performLayout():
Invalid argument(s): string is not well-formed UTF-16

The relevant error-causing widget was
RichText-[GlobalKey#8fa21]
../…/infrastructure/selectable_text.dart:463
When the exception was thrown, this was the stack
#0      ParagraphBuilder.addText (dart:ui/text.dart:2178:7)
#1      TextSpan.build
package:flutter/…/painting/text_span.dart:204
#2      TextPainter.layout
package:flutter/…/painting/text_painter.dart:569
#3      RenderParagraph._layoutText
package:flutter/…/rendering/paragraph.dart:515
#4      RenderParagraph._layoutTextWithConstraints
package:flutter/…/rendering/paragraph.dart:538
...
The following RenderObject was being processed when the exception was fired: RenderParagraph#415d2 relayoutBoundary=up15 NEEDS-LAYOUT
RenderObject: RenderParagraph#415d2 relayoutBoundary=up15 NEEDS-LAYOUT
    parentData: <none> (can use size)
    constraints: BoxConstraints(w=548.0, 0.0<=h<=Infinity)
    size: Size(548.0, 109.0)
    textAlign: left
    textDirection: ltr
    softWrap: wrapping at box width
    overflow: clip
    locale: en_US

```
</details>

It's worth noting that an official Dart package exists for handling `user-perceived characters` that may be of use when dealing with the given issue: 

[...]

PR #158, opened 2021-05-07
Adjusted text insertion and deletion to handle multi-codepoint characters. (#152)

matthew-carroll:
Adjusted text insertion and deletion to handle multi-codepoint characters. (#152)

### DIFF
--- a/super_editor/lib/src/default_editor/text.dart
+++ b/super_editor/lib/src/default_editor/text.dart
@@ -16,6 +16,7 @@ import 'package:super_editor/src/infrastructure/_logging.dart';
 import 'package:super_editor/src/infrastructure/attributed_text.dart';
 import 'package:super_editor/src/infrastructure/composable_text.dart';
 import 'package:super_editor/src/infrastructure/selectable_text.dart';
+import 'package:super_editor/src/infrastructure/text_layout.dart';
 
 import 'multi_node_editing.dart';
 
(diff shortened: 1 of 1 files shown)


osaxma:
By the way, the function keys also come as characters:
```
'F1' .... 'F12' 
```

I assume there was a good reason not include some of the following, but just in case I'll list them:
```
'Control'
'Meta' 
'Enter'
'Backspace' 
'Arrow Up' 
'Arrow Down'
'Arrow Right'
'Arrow Left' 
```

I don't have Windows, but these may be considered as well:
```
'Num Lock'
'Scroll Lock'
'Insert'
'Paste'
'Print Screen'
'Power'
```
^ there could be more.

osaxma:
By the way, I had sometime to look more into it today. 

According to [`RawKeyEventDataWeb #L34-L39`](https://github.com/flutter/flutter/blob/1d9032c7e1/packages/flutter/lib/src/services/raw_keyboard_web.dart#L35-L39),  and to [`RawKeyEvent.fromMessage #L330-L337`](https://github.com/flutter/flutter/blob/1d9032c7e1d867f071f2277eb1673e8f9b0274e3/packages/flutter/lib/src/services/raw_keyboard.dart#L330-#L337), the value of `key` is directly assigned to `character` in web only, and those `key` values are based on this:

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values

The link above has a comprehensive list but most of them are unrelated or are for special types of keyboards. 

I believe it's safe to do the following: 
```dart
const Set<String> blackListCharacters = {
'Shift',
'Alt',
'Escape',
'CapsLock',
'PageUp',
'PageDown',
'Home',
'End',
'Control',
'Meta' ,
'Enter',
'Backspace' ,
'Delete',
'F1',
'F2',
'F3',
'F4',
'F5',
'F6',
'F7',
'F8',
'F9',
'F10',
'F11',
'F12',
'Num Lock',
'Scroll Lock',
'Insert',
'Paste',
'Print Screen',
'Power',
};

if(kIsWeb && blackListCharacters.contains(character)) {
}
```

The `blackListCharacters` can grow as more of these values are encountered.

matthew-carroll:
@osaxma I updated the PR with the blacklist that you posted.

venkatd:
@matthew-carroll sure thing. Will try it shortly.

By the way #165 is preventing me from using SuperTextField in release mode for Flutter web.

venkatd:
@matthew-carroll the branch works for me. The `Tab` character was missing from the list I think. So when I hit tab, it literally inserts "Tab". Adding that to the blacklist should resolve.

super_editor/lib/src/default_editor/text.dart
```diff
@@ -586,20 +588,33 @@ ExecutionInstruction insertCharacterInTextComposable({
...
+
   // On web, keys like shift and alt are sending their full name
   // as a character, e.g., "Shift" and "Alt". This check prevents
   // those keys from inserting their name into content.
-  if (keyEvent.character!.length > 1) {
+  //
+  // This filter is a blacklist, and therefore it will fail to
+  // catch any key that isn't explicitly listed. The eventual solution
+  // to this is for the web to honor the standard key event contract,
+  // but that's out of our control.
```
salihgueler, line 601: Maybe we should bring the link that @osaxma shared on the issue (https://github.com/flutter/flutter/issues/81940) here to keep track of the source of the problem?

salihgueler (approved):

PR #167, opened 2021-05-09
Example editor toolbar links

matthew-carroll:
This PR implements link attributions for the example editor.

To better facilitate link-style behavior, I changed the concept of text attributions from `dynamic` to a dedicated base class called `Attribution`. This required updating lots of codes and tests to use the new type.

I also added some new behavior to `AttributedSpans` based on the type of queries that we need to do in the example app when we select no link vs one link vs 2+ links.


salihgueler (approved):
I need to speed up the CI process to actually have the tests running for these kinds of PRs :)

PR #168, opened 2021-05-09
Fixed bug where SuperTextField would not expand selection beyond 1 character.

matthew-carroll:
Fixed bug where SuperTextField would not expand selection beyond 1 character.


matthew-carroll:
@venkatd this PR should fix your text field selection issue

salihgueler (approved):

Issue #164, opened 2021-05-08
[SuperTextField] [web] Exception and minLines/maxLines not functioning in release mode


venkatd:
@matthew-carroll 

This is currently a blocker to swapping TextField with SuperTextField on our prod app.

Steps to reproduce:

<details>
<summary>pubspec.yaml</summary>

```yaml
name: webtesty

publish_to: 'none'

version: 1.0.0+1

environment:
  sdk: ">=2.12.0 <3.0.0"

dependencies:
  flutter:
    sdk: flutter

  super_editor:
    git:
      url: https://github.com/superlistapp/super_editor.git
      ref: main
      path: super_editor
```
</details>

<details>
<summary>pubspec.lock</summary>

```
# Generated by pub
# See https://dart.dev/tools/pub/glossary#lockfile
packages:
  args:
    dependency: transitive
    description:
      name: args
      url: "https://pub.dartlang.org"
    source: hosted
    version: "2.1.0"
  async:
    dependency: transitive
    description:
      name: async
      url: "https://pub.dartlang.org"
    source: hosted
    version: "2.6.1"
  characters:
    dependency: transitive
    description:
      name: characters
      url: "https://pub.dartlang.org"
    source: hosted
    version: "1.1.0"
  charcode:
    dependency: transitive
    description:
      name: charcode
      url: "https://pub.dartlang.org"
    source: hosted
    version: "1.2.0"
  collection:
    dependency: transitive
    description:
      name: collection
      url: "https://pub.dartlang.org"
    source: hosted
    version: "1.15.0"
  crypto:
    dependency: transitive
    description:
      name: crypto
      url: "https://pub.dartlang.org"
    source: hosted
    version: "3.0.1"
  flutter:
    dependency: "direct main"
    description: flutter
    source: sdk
    version: "0.0.0"
  http:
    dependency: transitive
    description:
      name: http
      url: "https://pub.dartlang.org"
    source: hosted
    version: "0.13.3"
  http_parser:
    dependency: transitive
    description:
      name: http_parser
      url: "https://pub.dartlang.org"
    source: hosted
    version: "4.0.0"
  linkify:
    dependency: transitive
    description:
      name: linkify
      url: "https://pub.dartlang.org"
    source: hosted
    version: "4.0.0"
  markdown:
    dependency: transitive
    description:
      name: markdown
      url: "https://pub.dartlang.org"
    source: hosted
    version: "4.0.0"
  meta:
    dependency: transitive
    description:
      name: meta
[...]

PR #179, opened 2021-05-14
Bugfix: Don't access debugNeedsLayout in profile or release modes. (#164)

matthew-carroll:
Bugfix: Don't access debugNeedsLayout in profile or release modes. (#164)


matthew-carroll:
CC @venkatd

matthew-carroll:

Issue #170, opened 2021-05-10
[SuperTextField] [web] Bug in getPositionAtStartOfLine getPositionAtEndOfLine


venkatd:
@matthew-carroll 

Steps to reproduce:

* Run the following on the latest master with the html renderer:

<details>
<summary>main.dart</summary>

```dart
import 'package:flutter/material.dart';
import 'package:super_editor/super_editor.dart';

const text = 'a\nb';

final controller = AttributedTextEditingController(
  text: AttributedText(text: text),
  selection: TextSelection.collapsed(offset: text.length),
);

void main() async {
  runApp(
    const MaterialApp(
      home: Material(child: TextFieldTest()),
    ),
  );
}

class TextFieldTest extends StatelessWidget {
  const TextFieldTest();

  @override
  Widget build(BuildContext context) {
    return Padding(
      padding: const EdgeInsets.all(25.0),
      child: Column(
        children: [
          Expanded(flex: 1, child: Container(color: Colors.blue)),
          DecoratedBox(
            decoration: BoxDecoration(
              border: Border.all(color: Colors.black, width: 1),
            ),
            child: SuperTextField(
              padding: const EdgeInsets.symmetric(horizontal: 12.0),
              textController: controller,
              minLines: 2,
              maxLines: 2,
            ),
          ),
        ],
      ),
    );
  }
}
```

</details>

* Bring focus to the TextField
* Hit cmd+left
* The cursor will jump to the line above rather than the start of the current line

Things start to work if I use your workaround by adding `lineHeight / 2` to the position in `getPositionAtStartOfLine` and `getPositionAtEndOfLine` as you've done in other areas of the code. Meaning `Offset(_, positionOffset.dy);` becomes `Offset(_, positionOffset.dy + _lineHeight / 2);`

I'm not sure why this is only specific to web but I am guessing the web text calculations are slightly different.

Flutter doesn't expose `getLineBoundary` which is inside of TextPainter.

Here you see there is a `getWordBoundary` which is proxying `TextPainter`
https://github.com/flutter/flutter/blob/65388ee2eeaf0d2cf087eaa4a325e3689020c46a/packages/flutter/lib/src/rendering/paragraph.dart#L816-L829

There is a `getLineBoundary` available on `TextPainter` here, but it's unfortunately not exposed through `RenderParagraph`
[...]

PR #180, opened 2021-05-14
Bugfix: jumping to start/end of lines was off by a line on web

matthew-carroll:
Bugfix: jumping to start/end of lines was off by a line on web due to inaccurate position calculations in the text system. Added half a line height to the calculations to sit definitively within the current line. (#170)


matthew-carroll:
CC @venkatd

matthew-carroll:

Issue #172, opened 2021-05-10
[SuperTextField] [web] Use ctrl in place of meta on Windows platforms


venkatd:
On Windows, the meta key maps to the windows key while ctrl+v, ctrl+left, etc are the shortcuts on windows. So I would like to be able to use the ctrl modifier in place of the meta modifier in that situation.

Basically I would like to remap all the keyboard shortcuts.

If we are able to decouple keyboard combinations from the actions, that could allow us to setup good default maps based on the platform and allow us to customize them.

This platform detection may be use case specific.

For example:
- Native MacOS/Windows/Linux: can check `defaultTargetPlatform`
- Web - can check `navigator.platform`, user agent, or feature detection, but each has its own trade-offs. I would probably opt for `navigator.platform`. Depending on the browser+os combo, there may be different keyboard shortcuts.

venkatd:
@matthew-carroll I think meta is pretty specific to Mac, so we're checking for MacOS and using meta. If it's not Mac, we default to control.

It looks more reliable to detect Mac. Windows, Linux, etc, are a mess with the platform strings. But all Macs start with `Mac`. Like `Mac68K`, `MacPPC`, `MacIntel`.

Some snippets


```dart
// The dart:html side
final _platform = window.navigator.platform ?? '';
final kShortcutKey = _platform.startsWith('Mac') ? LogicalKeyboardKey.meta : LogicalKeyboardKey.control;

// The dart:io side
final kShortcutKey = Platform.isMacOS ? LogicalKeyboardKey.meta : LogicalKeyboardKey.control;
```

Note that we'd need a conditional import like so to avoid `Platform.isMacOS` from crashing on web. Like so:

```
import 'src/impl/unsupported.dart'
    if (dart.library.html) 'src/impl/html.dart'
    if (dart.library.io) 'src/impl/io.dart' show kShortcutKey;
```

Then in our codebase, we use `kShortcutKey` in place of any reference to `LogicalKeyboardKey.meta`.

PR #184, opened 2021-05-15
Use control key for shortcuts on Windows and Linux. (#172)

matthew-carroll:
Use control key for shortcuts on Windows and Linux. (#172)


matthew-carroll:
CC @venkatd

matthew-carroll:
I'm going to merge this because @salihgueler is OOO and I want to use this same approach for another PR with keyboard shortcuts.

Issue #171, opened 2021-05-10
[SuperTextField] Shift+click to extend selection


venkatd:
Currently if you shift+click, it has the same behavior as clicking for text selection. Ideally this matches platform behavior and extends the selection to the point of shift-click.

PR #186, opened 2021-05-15
Shift + click to expand selection in SuperTextField. (#171)

matthew-carroll:
Shift + click to expand selection in SuperTextField. (#171)


matthew-carroll:
CC @venkatd

matthew-carroll:

Issue #176, opened 2021-05-12
[SuperTextField] cmd+backspace to delete the current line


venkatd:
Would like to mimic the behavior on MacOS/web. Not urgent and something we could implement on our end if needed.

@matthew-carroll

venkatd:
Here's our implementation:

```dart

extension TextEditingControllerExt on AttributedTextEditingController {
  void deleteToStartOfLine({required SelectableTextState selectableTextState}) {
    final deleteStartIndex = selectableTextState
        .getPositionAtStartOfLine(TextPosition(offset: selection.extentOffset))
        .offset;
    final deleteEndIndex = selection.end;

    text = text.removeRegion(
      startOffset: deleteStartIndex,
      endOffset: deleteEndIndex,
    );
    selection = TextSelection.collapsed(offset: deleteStartIndex);
  }
}

TextFieldKeyboardHandlerResult deleteToStartOfLine({
  required AttributedTextEditingController controller,
  required SelectableTextState selectableTextState,
  required RawKeyEvent keyEvent,
}) {
  if (controller.selection.extentOffset <= 0) {
    return TextFieldKeyboardHandlerResult.notHandled;
  }

  if (controller.selection.isCollapsed) {
    controller.deleteToStartOfLine(selectableTextState: selectableTextState);
  } else {
    controller.deleteSelectedText();
  }

  return TextFieldKeyboardHandlerResult.handled;
}
```

PR #187, opened 2021-05-15
Cmd + backspace deletes the entire line before the caret. (#176)

matthew-carroll:
Cmd + backspace deletes the entire line before the caret. (#176)


matthew-carroll:
CC @venkatd

matthew-carroll:

Issue #119, opened 2021-04-13
[web]: Typing the ` or ´ characters outputs the word "Dead" and puts the caret out of place


roughike:
Can be reproduced on the example app by running it on the web and typing <kbd>´</kbd> or <kbd>`</kbd>:

![dead](https://user-images.githubusercontent.com/13744304/114545099-34b6f400-9c64-11eb-9753-009bd7ce14ed.gif)

Copy-pasting <code>Look ´ at ` me ´ using ´ some ´ ticks</code> works just fine.

When running as a macOS app, instead of _"Dead"_, there's no output on the editor.

matthew-carroll:
@roughike where are those keys on your keyboard? I have \` in the upper left on the same key as `~`. I don't think I have a tick in the other direction, I only have a non-directional `'`.

I'm able to type \` on desktop and web without issue.

roughike:
Sorry for late reply, GitHub notifications are never-ending for me.

I have the Finnish keyboard layout that looks like this:

![image](https://user-images.githubusercontent.com/13744304/117610816-0812d100-b16b-11eb-9569-ab4c16edd75c.png)

The <kbd>´</kbd> is on the upper right corner. To type <kbd>`</kbd>, I use <kbd>Shift + ´</kbd>.

PR #188, opened 2021-05-15
Ignore characters on web that are reported as 'Dead'. (#119)

matthew-carroll:
Ignore characters on web that are reported as 'Dead'. (#119)

On web, some keys report a `character` value of "Dead". This is a Flutter bug/incomplete feature issue. I don't think there is much we can do, so I've opted to ignore any character with that value.


matthew-carroll:

PR #190, opened 2021-05-17
Added section on how to run the sample editor to README

bk-one:


matthew-carroll (approved):
LGTM

creativecreatorormaybenot:
Thanks, I tested it and I think it is pretty neat :) One note: the image used in the examples does not seem to load on web for me. Might be CORS-blocked.

salihgueler:
Hi @creativecreatorormaybenot thanks for the feedback! This seems to unrelated to this PR as it is just showing a way to run the editor, also, we have tested it on the web as well, might it be related to some of your network settings.

Anyways, I would recommend creating an issue with STR for this so we can take a look into it after triaging it :)

PR #191, opened 2021-05-18
Special paste behavior for web to avoid permissions

matthew-carroll:
Special paste behavior for web to avoid permissions

### DIFF
--- a/super_editor/lib/src/default_editor/document_interaction.dart
+++ b/super_editor/lib/src/default_editor/document_interaction.dart
@@ -12,6 +12,7 @@ import 'package:super_editor/src/core/document_selection.dart';
 import 'package:super_editor/src/core/edit_context.dart';
 import 'package:super_editor/src/infrastructure/_logging.dart';
 import 'package:super_editor/src/infrastructure/multi_tap_gesture.dart';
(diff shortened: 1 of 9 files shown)


matthew-carroll:
@venkatd can you give this branch a try and let me know what you think?

venkatd:
Hi @matthew-carroll I can give this a try tomorrow. I'm starting to feel like platform-specific behavior might be better putting for later. Will need to give some thought on what our current prios are.

For example, our code currently looks like this:

```dart
return ClipboardPasteZone(
  onPaste: (content) {
    if (content.files.isNotEmpty) {
      // upload files
    } else if (content.text != null) {
      messageEditingController.replaceSelection(content.text!);
    }
  },
  child: SuperTextField(...)
);
```

If I had `deleteSelectedText` and `insertTextAtCaret` that would be preferrable for our current use case. (And if this is handled at a lower level, we may not be able to override it easily and customize how we need paste to work.)

We want control over use cases like:
- Pasting files (we initiate a file upload right now)
- Pre-processing text (we may want to trim text, normalize quotes, etc)

matthew-carroll:
We haven't moved on this in a while, so I'm going to close this. We can come back to this later, if desired.

PR #205, opened 2021-05-31
204: Fix the selection problem on null safe projects.

### LINKED ISSUE #204: [BUG] NodeSelection problem on migrated Null Safe projects
Steps to Reproduce
- Run the editor on a null safe app (you can migrate the main example app as well for trying)
- Run the application
Expected
- App shows the editor
Actual
- App gives the following error
```
type 'Null' is not a subtype of type 'TextSelection' in type cast
```

salihgueler:
Fixes #204

### DIFF
--- a/super_editor/lib/src/core/document_selection.dart
+++ b/super_editor/lib/src/core/document_selection.dart
@@ -86,7 +86,7 @@ class DocumentNodeSelection<SelectionType> {
   });
 
   final String nodeId;
-  final SelectionType nodeSelection;
+  final SelectionType? nodeSelection;
   final bool isBase;
   final bool isExtent;
   final bool highlightWhenEmpty;
--- a/super_editor/lib/src/default_editor/blockquote.dart
+++ b/super_editor/lib/src/default_editor/blockquote.dart
@@ -197,7 +197,7 @@ Widget? blockquoteBuilder(ComponentContext componentContext) {
     return null;
   }
 
-  final textSelection = componentContext.nodeSelection?.nodeSelection as TextSelection;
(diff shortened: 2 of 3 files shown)


matthew-carroll:
I don't think this should ever be `null`. Can you describe where this is an issue?

salihgueler:
> I don't think this should ever be `null`. Can you describe where this is an issue?

With any null safe app, this issue happens. For reproducing you can migrate example app with 'dart migrate --skip-import-check' and run the app.

matthew-carroll:
What's the actual error and where are you seeing it?

salihgueler:
> What's the actual error and where are you seeing it?

The error is as described in the Github issue #204. It points to the parts that I made the change. 

Basically error says, you are trying to assign null to TextSelection

matthew-carroll:
Ok, I've reproduced the issue. After some initial investigation it still looks like the property should not be nullable, so this change is not the appropriate fix. I'll put up a PR for this later tonight.

In the future, if you could post the stacktrace that would be helpful. This issue could probably have been debugged with just a stacktrace instead of a project migration and deployment.

matthew-carroll:
Taking a 2nd look at this PR, I thought the `?` was being applied to the node's selection rather than the builder's selection. I think this change might be correct after all. I'm going to take one more look and then merge this if it looks like it covers all the component builders.

matthew-carroll:
Ok, I thought maybe this was missing the `ParagraphBuilder` but I see that this condition was already taken into account in that component builder. I'm going to approve this and merge it. I'm also going to put up a PR that migrates the example app to null safety.

matthew-carroll (approved):

PR #206, opened 2021-06-01
Marketing video demo

matthew-carroll:
This PR includes an automated demo for a v0.1.0 marketing video. It also includes some bug fixes that were discovered along the way.


matthew-carroll:
I'm going to merge this without review to unblock some necessary work. Most of this PR is the addition of a demo and therefore not significant from a review perspective, however it also includes some important bug fixes that resolve issues introduced when I separated editor behaviors from keyboard handlers.

PR #207, opened 2021-06-01
Renamed Editor to SuperEditor and SelectableText to SuperSelectableText

matthew-carroll:
Renamed Editor to SuperEditor and SelectableText to SuperSelectableText to disambiguate with other widgets and classes.


matthew-carroll:
I'm going to merge this after PR validation runs so that I can fix #204 and continue with v0.1.0 preparations.

PR #208, opened 2021-06-01
Migrated example app to null safety

matthew-carroll:
Migrated example app to null safety, also fixed a bunch of analysis warnings.


matthew-carroll:
I'm going to merge this without review to continue with time-sensitive v0.1.0 updates.

PR #211, opened 2021-06-02
Created README_RELEASING.md with instructions for releasing new versions of Super Editor

matthew-carroll:
- Updated CHANGELOG.md with release notes for v0.1.0
 - Fixed a bug where pressing delete in an empty paragraph below an image/HR would move the selection up but not delete the paragraph
 - Fixed a bug where pressing left/right with an expanded selection would move the caret after collapsing, but the caret shouldn't move
 - Fixed a bug where the format bar in the example app wasn't changing position with a changing selection
 - Updated the demo of AttributedText to use the new Attribution definitions instead of Strings

### DIFF
--- a/super_editor/CHANGELOG.md
+++ b/super_editor/CHANGELOG.md
@@ -1,3 +1,21 @@
-## [0.0.1] - TODO: Add release date.
+## [0.1.0] - June 3, 2021
 
-* TODO: Describe initial release.
+The first release of Super Editor.
+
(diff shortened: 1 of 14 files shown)


salihgueler (approved):
Looks good to me 👍🏼  But let's get @bk-one's opinion on this for sure.

bk-one (approved):
Looks good to me! Thanks

PR #221, opened 2021-06-03
[docs] core/document_selection.dart

creativecreatorormaybenot:
## Summary

### Part

In this pull request in particular, I am only adding documentation to the `core/document_selection.dart` file, exclusively. This should hopefully make it easier to review the individual parts of my contribution.

## Guidelines

### Assertions

### Imports

### Meta

## Full contribution

### Reviews


super_editor/lib/src/core/document_selection.dart
```diff
@@ -16,26 +20,82 @@ import 'document.dart';
...
+  /// See also:
+  ///
+  ///  * [isCollapsed], which determines whether a selection is collapsed or
+  ///    not.
   const DocumentSelection.collapsed({
     required DocumentPosition position,
   })  : base = position,
         extent = position;
 
+  /// Creates a selection from the [base] position and [extent] position within
```
matthew-carroll, line 34: I think we should change "and" to "to" because the nature of `base` and `extent` imply a direction.

super_editor/lib/src/core/document_selection.dart
```diff
@@ -16,26 +20,82 @@ import 'document.dart';
...
+  /// the document.
   DocumentSelection({
     required this.base,
     required this.extent,
   });
 
+  /// The base position of the selection within the document.
+  ///
+  /// If the [base] position comes before the [extent] position within the
```
matthew-carroll, line 43: I think we need 3 cases here instead of 2 and blank lines between them:

If [base] equals [extent], the selection is collapsed.

If [base] comes before [extent], the selection is expanded in the downstream direction.

If [base] comes after [extent], the selection is expanded in the upstream direction.

super_editor/lib/src/core/document_selection.dart
```diff
@@ -16,26 +20,82 @@ import 'document.dart';
...
+  /// document and the specific node that the position points to, the selection
+  /// is in the downstream direction.
+  /// If the [base] position and [extent] position are equal, the selection is
+  /// collapsed (see [isCollapsed] for details).
+  /// Otherwise, the selection is in the upstream direction.
   final DocumentPosition base;
+
+  /// The extent position of the selection within the document.
+  ///
+  /// If the [extent] position comes before the [base] position within the
```
matthew-carroll, line 53: Please apply similar changes as mentioned for `base`.

super_editor/lib/src/core/document_selection.dart
```diff
@@ -16,26 +20,82 @@ import 'document.dart';
...
+  /// document and the specific node that the position points to, the selection
+  /// is in the upstream direction.
+  /// If the [extent] position and [base] position are equal, the selection is
+  /// collapsed (see [isCollapsed] for details).
+  /// Otherwise, the selection is in the downstream direction.
   final DocumentPosition extent;
 
+  /// Whether this selection is collapsed or not.
```
matthew-carroll, line 61: To follow the style of similar doc comments elsewhere in this package, the first line should read:
Returns [true] if this selection is collapsed, or [false] if this selection is expanded.

super_editor/lib/src/core/document_selection.dart
```diff
@@ -16,26 +20,82 @@ import 'document.dart';
...
+  ///
+  /// A collapsed document selection is one where the [base] position and
+  /// [extent] position are identical ([DocumentPosition.==] to be precise).
+  /// Semantically, a collapsed selection represents a single position and can
+  /// be understood as *nothing being selected*, i.e. the cursor not having
+  /// highlighted any node or character.
```
matthew-carroll, line 67: This part isn't quite accurate. Images and horizontal rules are highlighted even when the selection is collapsed because those nodes/components only have one piece of content. I think it is sufficient to say that:

A [DocumentSelection] is "collapsed" when its [base] and [extent] are equal ([DocumentPosition.==]). Otherwise, the [DocumentSelection] is "expanded".

super_editor/lib/src/core/document_selection.dart
```diff
@@ -16,26 +20,82 @@ import 'document.dart';
...
   bool get isCollapsed => base == extent;
 
   @override
   String toString() {
     return '[DocumentSelection] - \n  base: ($base),\n  extent: ($extent)';
   }
 
+  /// Returns a version of this that is collapsed at the [extent] position.
+  ///
+  /// If this selection is already collapsed, returns this instance (i.e. the
```
matthew-carroll, line 83: DocumentSelection is immutable, so I don't think we need this specification.

super_editor/lib/src/core/document_selection.dart
```diff
@@ -16,26 +20,82 @@ import 'document.dart';
...
+  /// [identical] object).
+  ///
+  /// Note that the relative position within the document does not play a role
```
matthew-carroll, line 86: We can make this more concise:
Note, the returned [DocumentSelection] is collapsed at this selection's [extent], regardless of whether this selection's [extent] comes before or after its [base].

super_editor/lib/src/core/document_selection.dart
```diff
@@ -52,6 +112,17 @@ class DocumentSelection {
   ///
   /// The source [Document] is required so that the upstream [DocumentPosition]
   /// can be selected from [base] and [extent].
+  ///
+  /// If this selection is already collapsed, returns this instance (i.e. the
```
matthew-carroll, line 116: Similar to an earlier comment, this class is immutable, so I don't think we need this note.

super_editor/lib/src/core/document_selection.dart
```diff
@@ -83,6 +154,17 @@ class DocumentSelection {
   ///
   /// The source [Document] is required so that the downstream [DocumentPosition]
   /// can be selected from [base] and [extent].
+  ///
+  /// If this selection is already collapsed, returns this instance (i.e. the
```
matthew-carroll, line 158: Same here for removing this note.

super_editor/lib/src/core/document_selection.dart
```diff
@@ -117,6 +199,8 @@ class DocumentSelection {
   @override
   int get hashCode => base.hashCode ^ extent.hashCode;
 
+  /// Creates a copy of this selection but with the given fields replaced with
```
matthew-carroll, line 202: I think this description can match that of `TextSelection.copy`:
Creates a new [DocumentSelection] based on the current selection, with the provided parameters overridden.

super_editor/lib/src/core/document_selection.dart
```diff
@@ -146,10 +246,35 @@ class DocumentNodeSelection<SelectionType> {
     this.highlightWhenEmpty = false,
   });
 
+  /// ID that points to the single node that the selection describes about.
```
matthew-carroll, line 249: I think this description can be:
The ID of the node that's selected.

super_editor/lib/src/core/document_selection.dart
```diff
@@ -146,10 +246,35 @@ class DocumentNodeSelection<SelectionType> {
...
   final String nodeId;
+
+  /// The actual node-specific selection data.
```
matthew-carroll, line 252: This description can be:
The selection within the given node.

super_editor/lib/src/core/document_selection.dart
```diff
@@ -146,10 +246,35 @@ class DocumentNodeSelection<SelectionType> {
...
   final SelectionType? nodeSelection;
+
+  /// Whether this node selection is base in the context of the surrounding
```
matthew-carroll, line 255: I think I would say:
[true] if this [DocumentNodeSelection] forms the base position of a larger document selection, false otherwise.

In the case that a document selection is limited to a single node, a [DocumentNodeSelection] will form the base position and the extent position at the same time.

super_editor/lib/src/core/document_selection.dart
```diff
@@ -146,10 +246,35 @@ class DocumentNodeSelection<SelectionType> {
...
+  /// [DocumentSelection].
+  ///
+  /// If the node that [DocumentSelection.base] points to is equal to the node
+  /// that [nodeId] points to, [isBase] is `true`. Otherwise, it is `false`.
   final bool isBase;
+
+  /// Whether this node selection is extent in the context of the surrounding
```
matthew-carroll, line 262: I would have this description match my suggestion above, but adjusted for the extent.

super_editor/lib/src/core/document_selection.dart
```diff
@@ -146,10 +246,35 @@ class DocumentNodeSelection<SelectionType> {
...
+  /// [DocumentSelection].
+  ///
+  /// If the node that [DocumentSelection.extent] points to is equal to the node
+  /// that [nodeId] points to, [isExtent] is `true`. Otherwise, it is `false`.
   final bool isExtent;
+
+  /// Whether a thin selection highlight should be shown when the text is empty,
```
matthew-carroll, line 269: I would say:
[true] if the component rendering this [DocumentNodeSelection] should paint a highlight even when the given node has no content, [false] otherwise.

This behavior is primarily intended to coordinate small highlights in empty paragraphs when a document selection spans multiple paragraphs.

matthew-carroll (changes requested):

super_editor/lib/src/core/document_selection.dart
creativecreatorormaybenot: @matthew-carroll I did actually already have all three cases, however, I used the word "Otherwise` for the last case. I will change it to use "If" every time and add the new lines.

super_editor/lib/src/core/document_selection.dart
creativecreatorormaybenot: [Effective Dart states](https://dart.dev/guides/language/effective-dart/documentation#prefer-starting-variable-getter-or-setter-comments-with-noun-phrases) otherwise:

> PREFER starting variable, getter, or setter comments with noun phrases.

Do you think I should still change it, here and elsewhere?

super_editor/lib/src/core/document_selection.dart
creativecreatorormaybenot: About this and the two below, I saw that you used this style elsewhere in the docs, though I was not sure if everywhere.

However, it felt off to me because of https://dart.dev/guides/language/effective-dart/documentation#prefer-starting-variable-getter-or-setter-comments-with-noun-phrases.  
I also checked `flutter/flutter` code to find how it is done there and I found:  

https://github.com/flutter/flutter/blob/94a41cde83695b1267d0d7fae9d075bfd92af402/packages/flutter/lib/src/material/scaffold.dart#L2114

However, I also found some code that went along the lines of "True if ...".  So I am wondering because you have mentioned this style choice here and elsewhere: what do you think we should go with? I can update all occurrences accordingly afterwards.

creativecreatorormaybenot:
Thanks @matthew-carroll - I applied most suggestions from your code review. Now, there is only the open questions about `bool` getters. Once that is resolved, I can make any last adjustments if necessary.

creativecreatorormaybenot:
@matthew-carroll Let me know when you have made a decision on https://github.com/superlistapp/super_editor/pull/221#discussion_r645969982 - I will adjust accordingly then 👍

super_editor/lib/src/core/document_selection.dart
matthew-carroll: @creativecreatorormaybenot let's do this, take a look at the section right above the one that you linked:
https://dart.dev/guides/language/effective-dart/documentation#prefer-starting-function-or-method-comments-with-third-person-verbs

The example they give is:
```
/// Returns `true` if every element satisfies the [predicate].
```

Let's take that approach. So in this case:

```
/// Returns `true` if this [DocumentNodeSelection] forms the base position of a larger document selection, `false` otherwise.
```

And one of these days I'll figure out the correct different between `[]` ` `` ` in these doc comments....

Let me know if you have any other questions about this.

creativecreatorormaybenot: @matthew-carroll Thanks, makes sense 👍 

The difference between ```` and `[]` is easy: ```` is for **formatting** as code blocks and `[]` is for **linking** to references (in the IDE for code navigation and in generated docs for hyperlinks) :v:

matthew-carroll (approved):
LGTM

PR #222, opened 2021-06-03
[docs] core/document_composer.dart

creativecreatorormaybenot:
## Summary

### Part

In this pull request in particular, I am only adding documentation to the `core/document_composer.dart` file, exclusively. This should hopefully make it easier to review the individual parts of my contribution.

## Guidelines

### Assertions

### Imports

### Meta

## Full contribution

### Reviews


matthew-carroll (approved):
LGTM

PR #223, opened 2021-06-03
[docs] core/document_editor.dart

creativecreatorormaybenot:
## Summary

### Part

In this pull request in particular, I am only adding documentation to the `core/document_editor.dart` file, exclusively. This should hopefully make it easier to review the individual parts of my contribution.

## Guidelines

### Assertions

### Imports

### Meta

## Full contribution

### Reviews


super_editor/lib/src/core/document_editor.dart
```diff
@@ -6,51 +6,58 @@ import 'package:uuid/uuid.dart';
...
-  /// Executes the given `command` to alter the `Document` that is tied
-  /// to this `DocumentEditor`.
+  /// Executes the given [command] to alter the [Document] that is tied
+  /// to this [DocumentEditor].
   void executeCommand(EditorCommand command) {
     command.execute(_document, DocumentEditorTransaction._(_document));
   }
 }
 
+/// A command that can be executed on a given document.
```
matthew-carroll, line 42: I think all we need to say here is:
A command that alters a [Document] by applying changes in a [DocumentEditorTransaction].

super_editor/lib/src/core/document_editor.dart
```diff
@@ -6,51 +6,58 @@ import 'package:uuid/uuid.dart';
...
-  /// the current content of the `document` to make appropriate changes.
+  /// The [document] is provided in case this command needs to query
+  /// the current content of the [document] to make appropriate changes.
   void execute(Document document, DocumentEditorTransaction transaction);
 }
 
-/// Functional version of an `EditorCommand` for commands that
-/// don't require variables or private functions.
+/// Functional version of an [EditorCommand] for commands that
+/// do not require variables or private functions.
```
matthew-carroll, line 57: Contractions are used on purpose, similar to Flutter website guide copy policies. That policy may not be publicly stated, but it's enforced.

super_editor/lib/src/core/document_editor.dart
```diff
@@ -69,32 +76,36 @@ class DocumentEditorTransaction {
...
     return _document.deleteNode(node);
   }
 }
 
-/// An in-memory, mutable `Document`.
+/// An in-memory, mutable [Document].
 class MutableDocument with ChangeNotifier implements Document {
+  /// Creates an in-memory, mutable version of a [Document].
+  ///
+  /// Optionally, initial [nodes] can be given. If not specified, the list of
```
matthew-carroll, line 107: I would recommend inverting this language. It reads in a passive fashion. Instead, consider:

Initializes the content of this [MutableDocument] with the given [nodes], if provided, or empty content, if [nodes] are not provided.

matthew-carroll (changes requested):

matthew-carroll (approved):
LGTM

PR #224, opened 2021-06-03
[docs] core/document_layout.dart

creativecreatorormaybenot:
## Summary

### Part

In this pull request in particular, I am only adding documentation to the `core/document_layout.dart` file, exclusively. This should hopefully make it easier to review the individual parts of my contribution.

## Guidelines

### Assertions

### Imports

### Meta

## Full contribution

### Reviews


super_editor/lib/src/core/document_layout.dart
```diff
@@ -246,12 +246,41 @@ mixin DocumentComponent<T extends StatefulWidget> on State<T> {
 /// so long as those [id]s don't conflict with existing [id]s. You're
 /// responsible for implementing whatever behavior those custom
 /// [MovementModifier]s represent.
+///
+/// There is no default value for character-by-character movement because that
+/// is the default movement that occurs when **no** movement modifiers are at
+/// play.
 class MovementModifier {
+  /// The default identifier for word-by-word selection movement.
```
matthew-carroll, line 254: I think we can remove most of this description. This value is a "default" in the same way that everything in this package is "default", so I don't think we need to say that. The top line can probably be replaced with "Move text selection word-by-word."

We can remove the 2nd line because it repeats information that's explained in the `MovementModifier` class doc and that information applies to all such built-in values.

super_editor/lib/src/core/document_layout.dart
```diff
@@ -246,12 +246,41 @@ mixin DocumentComponent<T extends StatefulWidget> on State<T> {
...
+  ///
+  /// This is understood by the default node implementations.
+  ///
+  /// See also:
+  ///
+  ///  * [line], which is the default value for line-by-line selection movement.
   static const word = MovementModifier('word');
+
+  /// The default identifier for line-by-line selection movement.
```
matthew-carroll, line 263: Similar changes can be made here as made to the `word` description.

super_editor/lib/src/core/document_layout.dart
```diff
@@ -246,12 +246,41 @@ mixin DocumentComponent<T extends StatefulWidget> on State<T> {
...
+  ///
+  /// This is understood by the default node implementations.
+  ///
+  /// See also:
+  ///
+  ///  * [word], which is the default value for word-by-word selection movement.
   static const line = MovementModifier('line');
 
+  /// Creates a movement modifier from an [id] that is either understood by the
```
matthew-carroll, line 272: Most of this description repeats information that's already described in the class doc, which is not unique to this constructor.

I would expect something more like:
"Creates a movement modifier that is globally uniquely identified by the provided [id]."

super_editor/lib/src/core/document_layout.dart
```diff
@@ -246,12 +246,41 @@ mixin DocumentComponent<T extends StatefulWidget> on State<T> {
...
+  /// default node implementation or implemented by your own nodes.
   const MovementModifier(this.id);
 
+  /// The identifier of this modifier.
```
matthew-carroll, line 276: I think that all we need to say here is:
"Globally unique ID that differentiates this `MovementModifier` from all others."

super_editor/lib/src/core/document_layout.dart
```diff
@@ -281,6 +310,13 @@ typedef ComponentBuilder = Widget? Function(ComponentContext);
 /// Information that is provided to a [ComponentBuilder] to
 /// construct an appropriate [DocumentComponent] widget.
 class ComponentContext {
+  /// Creates a context for a [ComponentBuilder] to build a [DocumentComponent]
```
matthew-carroll, line 313: I'm not sure we need a doc here. Or if we do, it's probably sufficient to state: "Creates a [ComponentContext]"

The first sentence pretty much restates the class doc. The second sentence describes the contract of the constructor, which is equally visible in generated docs.

matthew-carroll (changes requested):

matthew-carroll (approved):
LGTM

PR #225, opened 2021-06-03
[docs] core/edit_context.dart

creativecreatorormaybenot:
## Summary

### Part

In this pull request in particular, I am only adding documentation to the `core/edit_context.dart` file, exclusively. This should hopefully make it easier to review the individual parts of my contribution.

## Guidelines

### Assertions

### Imports

### Meta

## Full contribution

### Reviews


salihgueler:
Hi @creativecreatorormaybenot, thank you for your contribution. In my opinion, these changes and your proposals should be addressed in an issue first and after the decisions there, it should be added as a PR. 

I think of this way for each PR and change of yours that you opened up now. My suggestion would be opening discussion in the issue and then creating PRs.

@matthew-carroll what do you think?

matthew-carroll:
@salihgueler @creativecreatorormaybenot - For added doc comments, those are probably fine in PRs. I'll take a look at these when I get a chance.

For broader recommendations about stylistic practices or other constraints, I agree with @salihgueler that we should start with a ticket and discussion. From there, if a suggestion is accepted, it should first be reflected in the contributing guide. Then it can be applied to the source code.

creativecreatorormaybenot:
> In my opinion, these changes and your proposals should be addressed in an issue first and after the decisions there, it should be added as a PR.

Hi @salihgueler, thanks for taking a look.

The reason I wrote this docs was mainly because I wanted to gain a deep understanding of the inner workings and I figured I could contribute by filling in the gaps in the docs along the way. So in this batch, there was no pain in doing it, even if it turns out it is not useful for the repo :)  
If there are immediate issues with them, I can make the necessary changes or we do not have to go forward with them at all.

I will for sure await any feedback from yours and @matthew-carroll's side before opening new PRs 👍 We can also collect it in an issue if that helps.

---

For these first few PRs, I simply used the style that I usually use when writing code myself, for open source, or even contributing to `flutter/flutter` 🙂

super_editor/lib/src/core/edit_context.dart
```diff
@@ -1,24 +1,53 @@
...
 import 'document_composer.dart';
 import 'document_editor.dart';
 import 'document_layout.dart';
 
 /// Collection of core artifacts used to edit a document.
+///
+/// This contains the [editor], the [composer], as well as access to the
+/// current [DocumentLayout] and the common editor operations.
+///
+/// Together, they are passed along to the [DocumentInteractor] or any other
```
matthew-carroll, line 16: Let's avoid stating where this object is used because that can easily change over time and it's unlikely that this API doc will be adjusted accordingly.

super_editor/lib/src/core/edit_context.dart
```diff
@@ -1,24 +1,53 @@
...
+/// user that needs to perform editing actions.
 class EditContext {
+  /// Creates an edit context that makes up a collection of core artifacts for
+  /// editing a document.
+  ///
+  /// In particular, the context contains the editor, composer, and layout.
```
matthew-carroll, line 22: This line repeats information that's already available in the class docs. I don't think we need to repeat it.

super_editor/lib/src/core/edit_context.dart
```diff
@@ -1,24 +1,53 @@
...
+  /// While the [editor] and [composer] are passed directly, the
```
matthew-carroll, line 23: I think the explanation of `getDocumentLayout` and `commonOps` makes more sense in the class docs rather than the constructor docs because these concepts are not unique to the constructor.

super_editor/lib/src/core/edit_context.dart
```diff
@@ -1,24 +1,53 @@
...
     required this.composer,
     required this.commonOps,
   }) : _getDocumentLayout = getDocumentLayout;
 
+  /// The editor of the [Document] that allows executing commands that alter the
+  /// structure of the document.
   final DocumentEditor editor;
 
-  final DocumentLayout Function() _getDocumentLayout;
+  /// The current document layout state.
```
matthew-carroll, line 40: If you check the `DocumentLayout` definition, you'll notice that its an interface without any mixins or extensions, so the description as "state" is an over-specification here.

I think that this member description should avoid referencing "state", and it should avoid referencing a temporal relationship, e.g., "current". 

The only thing we know for sure is that this method returns the visual representation of the document. You can explain that the document layout instance may change over time and that's why it's accessed via a function, but I don't think we should imply more than that.

super_editor/lib/src/core/edit_context.dart
```diff
@@ -1,24 +1,53 @@
...
+  ///
+  /// The member will always give access to the current state of the document
+  /// layout, e.g. the state of the [DefaultDocumentLayout].
   DocumentLayout get documentLayout => _getDocumentLayout();
+  final DocumentLayout Function() _getDocumentLayout;
 
+  /// The composer of the document that maintains the [DocumentSelection].
```
matthew-carroll, line 47: It also maintains the set of active text `Attribution`s, and its not really the "composer of the document" because the document doesn't alter itself. The composer works in conjunction with the `DocumentEditor` to apply changes to the `Document`. You may want to look for a concise way to describe that.

super_editor/lib/src/core/edit_context.dart
```diff
@@ -1,24 +1,53 @@
...
   final DocumentComposer composer;
 
+  /// The common operations that can be executed on the [editor], [composer],
```
matthew-carroll, line 50: I would remove the word "The" - these aren't the only operations that exist, these are just an arbitrary set.

Also, I think this description over-specifies how the common operations work. I don't think we need to list `editor`, `composer`, and `documentLayout`. `CommonEditorOperations` applies common, complex changes to the `document`. You'll notice that none of the methods on `CommonEditorOperations` require the caller to pass in an `editor`, `composer`, or `documentLayout` because those relationships were configured before the caller obtained the `CommonEditorOperations`.

creativecreatorormaybenot:
Thanks a lot for the in-depth review @matthew-carroll. I rephrased and moved docs to address all of your concerns.

matthew-carroll (approved):
I may change these doc comments later. I'm thinking that the editor, layout, and composer should perhaps be private. I'll need to think more about it. This LGTM for now.

PR #229, opened 2021-06-04
Add a dart format check for 120 char line length

domesticmouse:
In theory, this check should fail as there are two dart files that `dart format` will want to change. I'll update them in this PR, once we have confirmed Github Action sets a failure flag.


matthew-carroll (approved):
LGTM

matthew-carroll:
@domesticmouse it looks like the new validation check found some format changes and failed.

domesticmouse:
Well, at least the line length check passes now...

matthew-carroll:
I'm going to submit this with CI failures because the failures are related to out-of-date website code.

I filed: https://github.com/superlistapp/super_editor/issues/240

Issue #202, opened 2021-05-28
[MacOS] Pressing [ESC] creates whitespace in editor


bk-one:
Steps to reproduce:
 - open the sample editor
 - be in a paragaph
 - press [ESC]
 - a blank space appears

Expected behavior:
- no blank space is created

tannermeade:
this can be closed now

PR #233, opened 2021-06-04
Bug Fix: illegal characters not allowed in Paragraph and Text

tannermeade:
Added some basic validation for the types of characters that can be inserted into a paragraph or text component.

This manifested as a bug on macOS when pressing the ESC key, it was inserting the UTF-16 character 0x27 which corresponds to the ASCII 0x27 character for ESC. https://github.com/superlistapp/super_editor/issues/202

This commit makes the control characters between 0x00 - 0x08 and 0x14 - 0x31 (inclusive) illegal for text and paragraph components. The numbers between 0x08 and 0x14 represent tabs and carriage returns, and because this commit is targeting uniformly illegal characters it is going to favor allowing potential characters that might want to be included in fridge cases such as carriage returns and vertical tabs. The main concern is not allowing characters like UTF-16 ESC to be entered.


matthew-carroll:
I'm wondering if this is solving a narrow issue within a broader set of keys/characters that should be ignored. We've already introduced quite a bit of arbitrary inspections to avoid inserting bad characters.

@venkatd and @osaxma what do you think about this particular filter?

tannermeade:
> I'm wondering if this is solving a narrow issue within a broader set of keys/characters that should be ignored. We've already introduced quite a bit of arbitrary inspections to avoid inserting bad characters.
> 
> @venkatd and @osaxma what do you think about this particular filter?



> I'm wondering if this is solving a narrow issue within a broader set of keys/characters that should be ignored. We've already introduced quite a bit of arbitrary inspections to avoid inserting bad characters.
> 
> @venkatd and @osaxma what do you think about this particular filter?

I'd propose that the core problem here is there is only one type of DocumentKeyboardAction, which forces all DocumentKeyboardAction(s) to do their own validation. This will lead to duplicate validation between common action types and events being forwarded to actions that are clearly irrelevant.

Further, I'd propose there are the following types of DocumentKeyboardActions: (as a working list expected to change)
- Component Insertion (non-control characters): a,b,c, 1, 2, 3, $, *, 😀, ...
- Component Special Insertion (computed characters): tab, shift+tab, shift+enter
- Document Control: esc, function keys, non-shift modifier + key, arrow key, shift + arrow key
- Document Insertion: enter
- Destructive: backspace, delete, shift+delete

Perhaps a more complete solution here is to set up some event forwarding based on some established primitive action types. Then these primitive action types can be defined and used to forward the keyEvent to the correct type.

matthew-carroll:
There may be an architectural discussion to be had about keyboard handlers at some point, but that really isn't the issue here. 

When I mentioned arbitrary inspections, my point wasn't that we're doing the same inspections in multiple places, my point was that we're collecting a seemingly unbounded set of conditions. We have a growing blacklist of special web keys, we have special tab handling, now we have a magic set of ASCII codes. Are we going to be back here in a couple weeks with another magic set of ASCII codes? How about weird unicode sets? Unbounded solutions are a problem and I'm wondering if this approach is an unbounded solution.

venkatd:
@matthew-carroll I agree with you that it's worth understanding the root cause, but for the time being it seems OK to add this filter.

It has already been done in SuperTextField and has been working well in our app. There's a built-in function called `LogicalKeyboardKey.isControlCharacter` which is already being used, so would prefer leaning on that vs. introducing new logic.

https://github.com/superlistapp/super_editor/blob/6a3ce1883257a37e08482b950a9b43be7e91fbe6/super_editor/lib/src/infrastructure/super_textfield.dart#L1358-L1360

Introducing a text input client long term may remove many edge cases since the OS is already properly translating keystrokes to characters. For example some other edge cases:
- The native emoji keyboard and IME input (adding accented characters) don't currently work in SuperTextField but they work on the Flutter web TextField.
- On international keyboard layouts, there's a dead key which replaces the quote key for adding accents. So you might type `'` followed by `a` and the system replaces it with `á`. (http://www.kbdedit.com/manual/ex18_latin_all_accents_chained_dead.html)

(These above are a very very low priority for us.)

But as you mentioned in the YT announcement, this may be a rabbit hole for now 😅

matthew-carroll:
@tannermeade if you replace your filter with `LogicalKeyboardKey.isControlerKeyCharacter()`, does that still solve the problem?

salihgueler:
> @tannermeade if you replace your filter with `LogicalKeyboardKey.isControlerKeyCharacter()`, does that still solve the problem?

FWIW Maybe it was misplaced but I tried to do that and it did not solve the problem. 

If it fixes the problem for @tannermeade though, that's great. Then I will readdress it in our project as well.

tannermeade:
> @tannermeade if you replace your filter with `LogicalKeyboardKey.isControlerKeyCharacter()`, does that still solve the problem?

It does solve the problem of the ESC key on my mac.



> > @tannermeade if you replace your filter with `LogicalKeyboardKey.isControlerKeyCharacter()`, does that still solve the problem?
> 
> FWIW I tried to do that and it did not solve the problem.
> 
> If it fixes the problem for @tannermeade though, that's great. Then I will readdress it in our project as well.

Can you give me some more information so I can try to reproduce what is happening for you? What device are you using? Are you running it in Chrome or Mac (or windows)? Were you on the pushed branch?

venkatd:
@salihgueler If you are able to log out what the codeUnits for the character are, that could help us isolate the issue:

FYI here's the implementation of `isControlCharacter` if we want to compare implementations and where they may differ cc @tannermeade 

```dart
  static bool isControlCharacter(String label) {
    if (label.length != 1) {
      return false;
    }
    final int codeUnit = label.codeUnitAt(0);
    return (codeUnit <= 0x1f && codeUnit >= 0x00) || (codeUnit >= 0x7f && codeUnit <= 0x9f);
  }
```

salihgueler:
> > @tannermeade if you replace your filter with `LogicalKeyboardKey.isControlerKeyCharacter()`, does that still solve the problem?
> 
> It does solve the problem of the ESC key on my mac.
> 
> > > @tannermeade if you replace your filter with `LogicalKeyboardKey.isControlerKeyCharacter()`, does that still solve the problem?
> > 
> > 
> > FWIW I tried to do that and it did not solve the problem.
> > If it fixes the problem for @tannermeade though, that's great. Then I will readdress it in our project as well.
> 
> Can you give me some more information so I can try to reproduce what is happening for you? What device are you using? Are you running it in Chrome or Mac (or windows)? Were you on the pushed branch? What exactly do you mean by "it did not solve the problem"?

What I meant was, when I tried that on main, it did not fix my problem. :) 

But I agree with @venkatd on his idea above. Maybe we should compare the implementations and add the difference. :)

tannermeade:
> > > @tannermeade if you replace your filter with `LogicalKeyboardKey.isControlerKeyCharacter()`, does that still solve the problem?
> > 
> > 
> > It does solve the problem of the ESC key on my mac.
> > > > @tannermeade if you replace your filter with `LogicalKeyboardKey.isControlerKeyCharacter()`, does that still solve the problem?
> > > 
> > > 
> > > FWIW I tried to do that and it did not solve the problem.
> > > If it fixes the problem for @tannermeade though, that's great. Then I will readdress it in our project as well.
> > 
> > 
> > Can you give me some more information so I can try to reproduce what is happening for you? What device are you using? Are you running it in Chrome or Mac (or windows)? Were you on the pushed branch? What exactly do you mean by "it did not solve the problem"?
> 
> What I meant was, when I tried that on main, it did not fix my problem. :)
> 
> But I agree with @venkatd on his idea above. Maybe we should compare the implementations and add the difference. :)

What do you mean by you "tried that on main"?

salihgueler:
> > > > @tannermeade if you replace your filter with `LogicalKeyboardKey.isControlerKeyCharacter()`, does that still solve the problem?
> > > 
> > > 
> > > It does solve the problem of the ESC key on my mac.
> > > > > @tannermeade if you replace your filter with `LogicalKeyboardKey.isControlerKeyCharacter()`, does that still solve the problem?
> > > > 
> > > > 
> > > > FWIW I tried to do that and it did not solve the problem.
> > > > If it fixes the problem for @tannermeade though, that's great. Then I will readdress it in our project as well.
> > > 
> > > 
> > > Can you give me some more information so I can try to reproduce what is happening for you? What device are you using? Are you running it in Chrome or Mac (or windows)? Were you on the pushed branch? What exactly do you mean by "it did not solve the problem"?
> > 
> > 
> > What I meant was, when I tried that on main, it did not fix my problem. :)
> > But I agree with @venkatd on his idea above. Maybe we should compare the implementations and add the difference. :)
> 
> What do you mean by you "tried that on main"?

I tried the `LogicalKeyboardKey.isControlerKeyCharacter()` on main branch for fixing the ESC problem but it did not fix the Empty space addition problem.

tannermeade:
@salihgueler ah, I think that is because somewhere above the method started to be misnamed. Can you try using the method `LogicalKeyboardKey.isControlCharacter(keyEvent.character!)` instead of `LogicalKeyboardKey.isControlerKeyCharacter()`.
I've tried that on main and it is working for me.

venkatd:
@salihgueler if you find it's still not working, could you add something like this to your `void main()` file to isolate which keys are slipping through?

```dart
RawKeyboard.instance.addListener((rawKeyEvent) {
  print(rawKeyEvent);
});
```

Then if you type an illegal character that shows up, we can see which key slipped through the filters.

matthew-carroll:
Hi all, where do we stand with opinions on this? Can each of you opine with what you think the appropriate move is at this point? @venkatd @salihgueler @tannermeade 

Should we merge the original change?

Should we use the `isControlCharacter()` instead?

Should we create a combination of both?

Should we do something else?

tannermeade:
> Hi all, where do we stand with opinions on this? Can each of you opine with what you think the appropriate move is at this point? @venkatd @salihgueler @tannermeade
> 
> Should we merge the original change?
> 
> Should we use the `isControlCharacter()` instead?
> 
> Should we create a combination of both?
> 
> Should we do something else?

The only thing holding it up was @salihgueler said it wasn't working, but I'm pretty sure he was just copying the mistyped method. I'd say it is good to merge, except now there is a test failing. I'll get that fixed, and then I'd say we can move forward.

tannermeade:
@matthew-carroll ok it's passing all tests right now. It was only failing because it was missing the resent commits that got merged into the main branch. It's good to be merged on my end.

tannermeade:
@matthew-carroll and sorry, I didn't answer your questions specifically.
Should we merge the original change? **No**
Should we use the isControlCharacter() instead? **Yes**
Should we create a combination of both? **No**
Should we do something else? **No**

The current pull request is for using isControlCharacter(). It passes all tests. I'd say it's ready to go.

matthew-carroll:
Thanks @tannermeade. I'll give @venkatd and @salihgueler a chance to respond before merging this in.

venkatd:
I can't speak about the super editor itself, but SuperTextField has been using the `isControlCharacter` filter and nothing else. We haven't had any issues with illegal characters on Mac/web.

matthew-carroll:
@domesticmouse it looks like the line length check is failing on this PR but I'm not sure why. Is it possible to add some kind of diff output to see what it thinks the problem is?

domesticmouse:
Looking at [the logged output](https://github.com/superlistapp/super_editor/pull/233/checks#step:5:7), `dart format` is complaining about `super_editor/lib/src/default_editor/text.dart`.

Running `dart format -l 120 .` locally shows the single change is a blank line deletion:

```diff
index 918eb10..5e5b305 100644
--- a/super_editor/lib/src/default_editor/text.dart
+++ b/super_editor/lib/src/default_editor/text.dart
@@ -865,7 +865,6 @@ ExecutionInstruction anyCharacterToInsertInTextContent({
     return ExecutionInstruction.continueExecution;
   }
 
-
   String character = keyEvent.character!;
 
   // On web, keys like shift and alt are sending their full name
```

salihgueler:
Hi @tannermeade sorry I was off-grid for couple of days. I would say let's merge this in. It should not be a problem. My main concern was, I tried a similar approach before and it did not work, but it. might be because of several reasons.

Let's be sure that PR checks are passing before we merge this though :)

tannermeade:
> Hi @tannermeade sorry I was off-grid for couple of days. I would say let's merge this in. It should not be a problem. My main concern was, I tried a similar approach before and it did not work, but it. might be because of several reasons.
> 
> Let's be sure that PR checks are passing before we merge this though :)

It is currently passing all tests now 👍

matthew-carroll (approved):
LGTM. All tests are passing now. Thanks everyone for the input.

PR #248, opened 2021-06-14
Fix white space, when user press ESC key

Arkangel12:
This solves issue #202, adding default behavior to Keyboard actions for `ESC` key: _do nothing_ when press.

### DIFF
--- a/super_editor/lib/src/default_editor/document_interaction.dart
+++ b/super_editor/lib/src/default_editor/document_interaction.dart
@@ -66,7 +66,8 @@ class DocumentInteractor extends StatefulWidget {
   _DocumentInteractorState createState() => _DocumentInteractorState();
 }
 
-class _DocumentInteractorState extends State<DocumentInteractor> with SingleTickerProviderStateMixin {
+class _DocumentInteractorState extends State<DocumentInteractor>
+    with SingleTickerProviderStateMixin {
   final _dragGutterExtent = 100;
   final _maxDragSpeed = 20;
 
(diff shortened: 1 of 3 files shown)


matthew-carroll:
Is this still relevant after the latest merge to ignore control keys?

Also, for future reference, please set your formatter to 120 character line length. Most of the changes in this PR are automatic reformats.

tannermeade:
> Is this still relevant after the latest merge to ignore control keys?
> 
> Also, for future reference, please set your formatter to 120 character line length. Most of the changes in this PR are automatic reformats.

This is not relevant anymore.

This pull request solves the problem by adding a new `defaultKeyboardActions` called `doNothingWhenEscIsPressed` which acts as a hard break/cancel when iterating through the list of keyboard action handlers. All keyboard actions that are after the `doNothingWhenEscIsPressed` in the list can not handle the ESC key.

While this solution does avoid adding more logic to the `anyCharacterToInsertInParagraph` and `anyCharacterToInsertInTextContent` keyboard handlers, it does moves ESC key logic into the ordering of `defaultKeyboardActions`. This makes the keyboard actions tightly coupled and does not handle any other control character situation (which the merged solution does).

We can start a conversation about how to handle this better, but it was said in the already merged pull request that handling control characters is a very very low priority by @venkatd.
 
> Introducing a text input client long term may remove many edge cases since the OS is already properly translating keystrokes to characters. For example some other edge cases:
> 
> The native emoji keyboard and IME input don't currently work in SuperTextField but they work on the Flutter web TextField.
> On international keyboard layouts, there's a dead key which replaces the quote key for adding accents. So you might type ' followed by a and the system replaces it with á. If you don't type anything, it defaults to a quote. (http://www.kbdedit.com/manual/ex18_latin_all_accents_chained_dead.html)
> (These above are a very very low priority for us.)
> 
> But as you mentioned in the YT announcement, this may be a rabbit hole for now 😅

tannermeade:
oh and @Arkangel12 thanks for supporting the repo too bro! Sorry if my last post was a bit direct. Good job getting that GDE 🙌

Arkangel12:
> Is this still relevant after the latest merge to ignore control keys?
> 
> 
> 
> Also, for future reference, please set your formatter to 120 character line length. Most of the changes in this PR are automatic reformats.

Will do!

matthew-carroll:
Based on the responses, I'm assuming that this PR isn't needed, so I'm going to close it.

PR #256, opened 2021-06-18
Added a TextInputClient demo to show the simplest possible integration

matthew-carroll:
@osaxma and @venkatd - I'd be interested to hear any thoughts you have about this demo. I created a custom textfield based on `SuperSelectableText`, with `TextInputClient`, that handles taps and drags (no scrolling/overflow). 

There were 2 or 3 methods where I wasn't quite sure what their purpose should be. Maybe one of you know.

Also, as far as you can tell, did I get anything wrong with this barebones example?

My plan is to introduce demos that build up in complexity so that everyone who runs the demo app can see the discovery process that lead to whatever implementation we end up with.

@salihgueler I'll CC you here in case you're interested, but this PR shouldn't require any serious review.


super_editor/example/lib/demos/flutter_features/textinputclient/basic_text_input_client.dart
```diff
@@ -0,0 +1,228 @@
...
+      _currentText = value.text;
+      _currentSelection = value.selection;
+    });
+  }
+
+  @override
+  void updateFloatingCursor(RawFloatingCursorPoint point) {
+    print('My TextInputClient: updateFloatingCursor(): $point');
+
+    // I'm not sure what this is supposed to do.
```
osaxma, line 192: Starting at the 22" mark (i.e. press hold space bar) -- that's the floating cursor. 

https://user-images.githubusercontent.com/46427323/116639257-c5821500-a970-11eb-8b4d-718efae5cd47.mp4

btw, I wrote more details about it in this [comment](https://github.com/superlistapp/super_editor/discussions/242#discussioncomment-883562) from yesterday in the proposal.

super_editor/example/lib/demos/flutter_features/textinputclient/basic_text_input_client.dart
```diff
@@ -0,0 +1,228 @@
...
+    // or an escape hatch for input functionality that an app needs to support,
+    // but which does not exist at the OS/platform level.
+  }
+
+  @override
+  void showAutocorrectionPromptRect(int start, int end) {
+    print('My TextInputClient: showAutocorrectionPromptRect() - start: $start, end: $end');
+
+    // I'm not sure why iOS wants to show an "autocorrection" rectangle
+    // when we already have a selection visible.
```
osaxma, line 176: It's this one:

<p align="center">Screen Shot 2021-04-12 at 7 21 27 PM</p>

`start` and `end` are the selection offsets of the remote `TextEditingValue`. The method is called by the system based on the selection (typically right after a collapsed selection lands after the word).

osaxma:
LGTM as a basic example.

venkatd:
Looks good to me! I never had a chance to play with TextInputClient. What has to be kept in sync between the two? Is it a one-way sync or two-way sync?

Can we adopt just the parts we need for TextInputClient? For example, text selection, undo, panning, scrolling, etc, seem fine to keep in the Flutter world since there doesn't appear to be a lot of complexity and we get full control over how they are handled.

Meanwhile translating keystrokes to characters, auto-correct, etc, seem best to allow the platform/TextInputClient to do the heavy lifting. Those areas seem very complex or laden with edge cases if we manage them in Flutter-land.

I can play around with TextInputClient more when I find some time.

matthew-carroll:
@venkatd it's 2-way sync. `TextInputClient` receives updates to the `TextEditingValue`, but also in this example, you can see that we change the `TextEditingValue`, too, like when the user taps or drags.

matthew-carroll:
I added rough support for a floating cursor. @osaxma do you know where we tell the platform what side the textfield is so that the floating cursor reports correct offsets, and also prevents the floating cursor from going outside the bounds of the field? I tried the method on `TextInputConnection` to set the size of the editable, but that didn't seem to have any effect.


https://user-images.githubusercontent.com/7259036/122636923-a9614100-d0a0-11eb-867f-281be77054c3.mov

osaxma:
@matthew-carroll 

I assumed that the bounds cannot be set on the system side. If you take a look at [`EditableTextState.updateFloatingCursor`](https://github.com/flutter/flutter/blob/d79295af24c3ed621c33713ecda14ad196fd9c31/packages/flutter/lib/src/widgets/editable_text.dart#L1838), you'll notice that Flutter keeps track and computes a variable called: `_lastBoundedOffset`. I didn't dive much into their implementation there but I believe they're trying to solve the same problem. 

In the experiment, I ended up setting the bounds visually only. While the actual offset can get out of bounds, the value is clamped so the floating cursor is always painted within the document. This worked fine for a full-screen SuperEditor given its size, but it is not be ideal for smaller fields where the out of bound offset can get much larger than the field itself. It's worth investigating how Flutter does it in the referenced method above.

matthew-carroll:
I'm gonna close this PR and open a new one with initial implementations for Android and iOS text fields.

PR #257, opened 2021-06-23
fix(common_editor_operatons): bug in web deleting first letter of node solved

Josehdez96:
I was checking the code for a few days to understand why in web the first letter(offset = 0) of each node/paragraph would not able to deleted. I found the error and I made this little fix, but I want to keep helping you. it is a very nice project!


Josehdez96:
I will sent the fixed one

PR #258, opened 2021-06-23
fix(common_editor_operatons): bug deleting TextPosition's offset = 0 of nodes solved

Josehdez96:
I was checking the code for a few days to understand why in web (and I think it happens on desktop too) the first letter(offset = 0) of each node/paragraph could not be forward deleted. I found the error and I made this little fix, but I want to keep helping you. it is a very nice project!


matthew-carroll (approved):
LGTM. For future work, please begin by filing an issue with a description and reproduction steps so that we can track commit history to the original issues.

PR #259, opened 2021-06-25
fix(common_editor_operations): exception when deleting without existence of nodes has been solved

Josehdez96:
An exception was getting if the user keeps trying to delete (with backspace key) when all nodes are already deleted


matthew-carroll (approved):
Thanks. LGTM.

Issue #260, opened 2021-06-26
SuperTextField example is broken 


britannio:
<img width="1112" alt="Screenshot 2021-06-26 at 23 31 26" src="https://user-images.githubusercontent.com/33752528/123527372-c4eace00-d6d6-11eb-992d-032af8d4393f.png">

The regression was introduced in https://github.com/superlistapp/super_editor/pull/227 and can either be fixed by making the following list non `const` https://github.com/superlistapp/super_editor/blob/1ca4e21171b1bbc3c2349e9d071037e06ec36bb0/super_editor/example/lib/demos/supertextfield/_interactive_demo.dart#L17
or by storing a mutable copy of the list provided to `AttributedSpans` so that it can be sorted.
https://github.com/superlistapp/super_editor/blob/65fd55fa487375effa099ba379d0055639676913/lib/src/infrastructure/attributed_spans.dart#L45

PR #227, opened 2021-06-04
Adding flutter_lints

domesticmouse:
Hey @matthew-carroll,

I tried adding [flutter_lints](https://pub.dev/packages/flutter_lints), but the tests are failing. However, the reason is unmigrated code in the spikes directory, so I'm unsure what you want to do about that.

brett

PR #261, opened 2021-06-26
Store a copy of attributions provided to AttributedSpans

britannio:
Closes https://github.com/superlistapp/super_editor/issues/260

| Before                                                                                                                                                                 | After                                                                                                                                                                  |
|------------------------------------------------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| Screenshot 2021-06-26 at 23 31 26 | Screenshot 2021-06-26 at 23 47 03 |

### DIFF
--- a/super_editor/lib/src/infrastructure/attributed_spans.dart
+++ b/super_editor/lib/src/infrastructure/attributed_spans.dart
@@ -45,7 +45,7 @@ class AttributedSpans {
   /// with no spans.
   AttributedSpans({
     List<SpanMarker>? attributions,
-  }) : _attributions = attributions ?? [] {
+  }) : _attributions = [...?attributions] {
     _sortAttributions();
   }
 


britannio:
Hey @matthew-carroll, this fix is a one liner so it should be a quick review. I also proposed an alternative solution in #260 if that's preferred.

matthew-carroll:
Thanks for filing the issue and the PR. I checked #227 that you referenced, but I don't see where the spans were changed to `const` in that PR. It also looks like the existing code that's being changed in this PR reads `attributions ?? []`, which should instantiate a non-`const` list when no attributions are provided.

Is the goal here to prevent clients from passing in a `const` list?

britannio:
The `const` keyword causing the issue is at [_interactive_demo.dart](https://github.com/superlistapp/super_editor/pull/227/files#diff-ceaa52a1e5d0b4522dfea9341c26f3cb916b59c2250b931a11e2828b8e8b43c0R17).

> Is the goal here to prevent clients from passing in a const list?

The goal is to allow clients to pass const and mutable lists without anything breaking. At the moment, `AttributedSpans` only supports mutable lists.

matthew-carroll (approved):
LGTM. Thanks.

PR #264, opened 2021-07-07
WIP - Feedback wanted: example of embedding custom content in the editor

roughike:
## Background
We want to be able to display `Task` objects, stored in a `TasksRepository`, in the rich text editor.

The `Task` objects can be created, edited, and deleted outside of the editor. Users can also create these tasks directly from the editor (but that's not included in this sample (yet?)). 

These `Task` objects can also eventually be edited on some other device and if a `Task` that's visible in the editor is modified by another user, the editor should reflect that change.

This PR includes a minimal sample of how that would look like. For simplicity, it uses Markdown with a custom syntax (`<@ task:abc123 @>`) to embed `Task` objects in the document.

I'm looking for some feedback on the general approach taken on this PR, and if we should take a different approach instead. 

## Changes required from the editor
With this approach, in order to support the custom Markdown syntax, the `deserializeMarkdownToDocument` method in  markdown.dart needs these extra parameters:

* `customBlockSyntaxes`, which is a `List<md.BlockSyntax>`, passed into `md.Document` for parsing the custom task syntax and turning it into a Markdown AST node, and
* `customMarkdownToDocumentVisitor`, which is a method that turns the Markdown AST nodes into  `DocumentNode`s.

If this seems like a sane approach, I'm willing to continue building on top of it and add the necessary tests.

## Post-merge actions
- [ ] add a comment in [this discussion](https://github.com/superlistapp/super_editor/discussions/249) about the sample


super_editor/example/lib/demos/custom_elements/task_item.dart
```diff
@@ -0,0 +1,166 @@
+import 'dart:async';
+
+import 'package:example/demos/custom_elements/task.dart';
+import 'package:flutter/material.dart';
+import 'package:super_editor/super_editor.dart';
+
+/// A [DocumentNode] representing a [Task] in a [Document].
```
salihgueler, line 7:
```suggestion
/// A [TextNode] representing a [Task] in a [Document].
```

I would suggest to state that this is a TextNode rather than DocumentNode as it is extending itself from it. People can direct themselves to DocumentNode from TextNode later on.

super_editor/example/lib/demos/custom_elements/task_item.dart
```diff
@@ -0,0 +1,166 @@
+import 'dart:async';
```
salihgueler, line 1: I think this is the key example and it is great. 🎉 One thing to clarify, for others, on this example is to explain how we did such changes. Or even better, creating a document page in repository and explaining how custom elements can be created would be great.

salihgueler (approved):
I think it looks good overall as an intermediate step. If we decide to move forward with this approach, I would REALLY love to see a nice blog/documentation over the custom element creation on the matter. 

Great work @roughike

matthew-carroll:
Thanks for the PR. I'll take a look as soon as I can.

matthew-carroll:
I started looking through the changes but I think the premise for this particular use-case has conflated two different goals. I think we need to iron this out.

The issue is: **custom widgets vs custom document data types**

When @bk-one and I spoke about this goal weeks ago, the focus was on introducing custom widgets/media elements into the document, i.e., adding things with application behavior into the middle of a document. This is tantamount to cutting a whole in a document and injecting some other interactive experience that you want to insert.

However, at first blush, what I think I'm seeing in this PR is an attempt to introduce a "task" as a new fundamental data type within a `Document`. If so, that's a fundamentally different goal. Adding a "task" type to the document model isn't a big deal, but saying that tasks can be added, edited, or removed remotely is a huge issue. What we're really talking about with those use-cases is concurrent document editing. We just happen to be focused on a little innocuous "task" node. But it wouldn't make sense to try to solve that just for tasks when it looks to clearly be a subset of a much larger behavior. A behavior that requires substantial further groundwork to support.

With that said, am I understanding this PR correctly? Are you trying to making a "task" into a part of a `Document` the way that `ParagraphNode` and `ListItemNode` are part of the document model?

If my understanding is correct, and also what matters here is the ability to display tasks that can change remotely, then I would treat a group of tasks as a single element in the `Document`. A rough overview would be:

Markdown:
```
Some paragraph of text.

[superlist:tasks:0ea421be-e2d9-11eb-ba80-0242ac130004]

Some other paragraph of text.
```

Node:
```dart
class SuperlistTaskGroupNode {
  String get groupId;
}
```

Widget:
Use a `StatefulWidget` that talks to the repository or talks directly to Firestore and then renders tasks itself.

You can also implement the necessary UI capabilities to allow drag selection across the document and capture this content, too. It's up to the node and the widget what it means to select parts of a task group.

I'd like to here everyone's thoughts. @roughike @salihgueler @bk-one

matthew-carroll:
@roughike it sounds like what you want is truly concurrent document editing. Based on your description, the task item is a legitimate piece of content in the document, much like a list item.

We don't have any support for concurrent document editing and I think that's far beyond the scope of anything that we're in a position to address right now.

@bk-one can you let us know what you'd like to happen here? I suppose I can look through the PR like any other code review, but I don't think we should add this to the example app, and I don't think we should invest time right now attempting to change the package to support this. Mobile seems to be the #1 priority, which I'm doing my best to deliver ASAP. After that, we either need to solve undo/redo before concurrent document editing, so that we know what the change log looks like, or we have to solve undo/redo and concurrent document editing at the same time. Either way, I think it needs to come after mobile, and no sooner than undo/redo. And it's going to be a significant project in itself.

If Superlist really needs this functionality, then I'd say go ahead and solve it in whatever way works for you for the time being. But the fundamental answer is concurrent document editing.

bk-one:
@matthew-carroll truly concurrent document editing is a feature that Superlist wants to include, but as we discussed earlier, not the main priority right now. I would suggest that @roughike and I are focusing on a JSON serialization for the Document Mode, as I understand that @Jethro87 also is interested in that. This will create the groundwork on our end to work with individual nodes and find ways to replace them when needed. We keep this in our own custom usage of super_editor for now and will bring learnings back that hopefully helps you and us to structure the concurrent editing eventually.

So please stick to the plan to finish mobile, focus on undo/redo next, and then we can see how to build basic use cases for collaboration and real-time updates on documents.

Does this make sense?

matthew-carroll:
> please stick to the plan to finish mobile, focus on undo/redo next, and then we can see how to build basic use cases for collaboration and real-time updates on documents

Sure, will do. What should we do with this PR?

> focusing on a JSON serialization for the Document Mode

Can someone draft a brief overview of what the goal is there? Is it just a JSON representation instead of a Markdown representation? Or is there more to that particular goal? Whenever possible, I'd like to be aware of how the editor is being used, in case that use-case requires altering plans in the future.

bk-one:
Yes, we will join the existing discussion at
https://github.com/superlistapp/super_editor/discussions/254 and propose a
path forward.

As said, this PR can either be closed without merging, or we keep it open
for research reasons and revisit it in a few months? I assume others might
find the approach helpful? But I assume it's easiest to close it but keep
the branch.

roughike:
I'll close this one until it's clearer what we want to do with it from Superlist's perspective.

Issue #270, opened 2021-07-22
`AttributedSpans` with "equivalent" `attributions` are failing equality checks 


britannio:
`AttributedSpans` have a list of `SpanMarker`s. These `SpanMarker`s can be sorted by their offset but if multiple `SpanMarker`s have the same offset then their positions depend on their insertion order. 

For this reason, the following `AttributedSpans` do not return `true` when compared via `==`:
Is this the intended behaviour? For context, I'm writing a HTML serialiser but some tests are failing when I compare documents via `Document.hasEquivalentContent()`.

```dart
AttributedSpans(
  attributions: [
    SpanMarker(
      attribution: boldAttribution,
      offset: 0,
      markerType: SpanMarkerType.start,
    ),
    SpanMarker(
      attribution: italicsAttribution,
      offset: 0,
      markerType: SpanMarkerType.start,
    ),
    SpanMarker(
      attribution: italicsAttribution,
      offset: 1,
      markerType: SpanMarkerType.end,
    ),
    SpanMarker(
      attribution: boldAttribution,
      offset: 1,
      markerType: SpanMarkerType.end,
    ),
  ],
);
```

```dart
AttributedSpans(
  attributions: [
    SpanMarker(
      attribution: italicsAttribution,
      offset: 0,
      markerType: SpanMarkerType.start,
    ),
    SpanMarker(
      attribution: boldAttribution,
      offset: 0,
      markerType: SpanMarkerType.start,
    ),
    SpanMarker(
      attribution: boldAttribution,
      offset: 1,
      markerType: SpanMarkerType.end,
    ),
    SpanMarker(
      attribution: italicsAttribution,
      offset: 1,
      markerType: SpanMarkerType.end,
    ),
  ],
);
```

PR #271, opened 2021-07-22
AttributedSpans equality check disregards order of attributions

britannio:
Closes https://github.com/superlistapp/super_editor/issues/270

### DIFF
--- a/super_editor/test/src/infrastructure/attributed_spans_test.dart
+++ b/super_editor/test/src/infrastructure/attributed_spans_test.dart
@@ -1,5 +1,6 @@
 import 'package:super_editor/src/infrastructure/attributed_spans.dart';
 import 'package:flutter_test/flutter_test.dart';
+import 'package:super_editor/super_editor.dart';
 
 // Attributions used throughout this test suite.
 const bold = NamedAttribution('bold');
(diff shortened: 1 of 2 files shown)


britannio:
Does VSCode have a setting to format lines at 120 characters?

creativecreatorormaybenot:
@britannio it has nothing to do with VSCode. But you can change the `dart format` line wrap in settings :)

britannio:
Cool, I created https://github.com/superlistapp/super_editor/pull/272 so I don't have to run the `dart format` command manually.

creativecreatorormaybenot:
@britannio Of course not what I was suggesting :) Nice PR - I meant the user settings.

matthew-carroll:
Can you add a test to the appropriate test file that verifies unordered equality?

britannio:
@matthew-carroll Just added the tests!

super_editor/test/src/infrastructure/attributed_spans_test.dart
```diff
@@ -477,6 +477,21 @@ void main() {
         expect(collapsedSpans[2].attributions.first, italics);
       });
     });
+
+    group('equality', () {
+      test('it determines that spans with equivalent content in a different order are equal', () {
```
matthew-carroll, line 482: The wording in this description is a bit passive. How about:
it is equal to another AttributedSpans with equivalent markers that are stored in a different order

super_editor/test/src/infrastructure/attributed_spans_test.dart
```diff
@@ -477,6 +477,21 @@ void main() {
         expect(collapsedSpans[2].attributions.first, italics);
       });
     });
+
+    group('equality', () {
```
matthew-carroll, line 481: If we're going to write a test for different marker order, then we should probably write a test for the standard case where the marker order is the same.

Also, how about a test for empty `AttributedSpans`s?

And, when it comes to proving things, the converse is important, too. Consider writing one or more tests that demonstrate non-equality, to make sure that equality doesn't always return `true`.

matthew-carroll (changes requested):
This is a good start on tests, but we should be more holistic with the testing range if we're going to take the time to add them.

britannio:
Thanks for the feedback! Hopefully, I've addressed it all.

super_editor/test/src/infrastructure/attributed_spans_test.dart
```diff
@@ -477,6 +478,42 @@ void main() {
...
+      const boldStart = SpanMarker(attribution: bold, offset: 0, markerType: SpanMarkerType.start);
```
matthew-carroll, line 483: I'll accept this test as-is so that we can merge it, but in general I think that test configurations should go within the tests. Right now it doesn't look too bad to have these 4 properties here, but imagine that this group of tests grows to be 50 tests. How many different configurations might we have in that case? They'd all accumulate here and become a mess. So, for tests, I recommend re-stating the conditions in each test. In rare cases, you might move some of the setup generation into a method that a test invokes, but even that should be rare.

matthew-carroll (approved):
LGTM. Thanks for the implementation and for following up with tests.

PR #272, opened 2021-07-22
Set VSCode Dart formatter line length to 120

britannio:


matthew-carroll:
Is this file at the correct level? I don't use VSCode, but for Android Studio, a developer opens the super_editor directory, rather than the root.

britannio:
@matthew-carroll Previously, I've been opening the root folder but opening the super_editor/ folder works equally well so I moved it.

matthew-carroll (approved):
LGTM

PR #273, opened 2021-07-24
MutableDocument equality check compares nodes of other document

britannio:
It was previously comparing `_nodes` with `nodes` but `nodes` is a getter that returns `_nodes` so they're identical and hence `const DeepCollectionEquality().equals(_nodes, nodes)` is always true.

This is my second PR related to misbehaving equality checks so would it be worth adding a few unit tests?

### DIFF
--- a/super_editor/test/src/core/document_editor_test.dart
+++ b/super_editor/test/src/core/document_editor_test.dart
@@ -21,5 +21,23 @@ void main() {
       // newNode exists at index 1
       expect(document.nodes.indexOf(newNode), 1);
     });
+
+    test('it is equal to a similar document with the same content', () {
+      final node = HorizontalRuleNode(id: '0');
+
+      final document1 = MutableDocument(nodes: [node]);
+      final document2 = MutableDocument(nodes: [node]);
+
+      expect(document1 == document2, isTrue);
+    });
+
+    test('it is NOT equal to a similar document with different content', () {
+      final node = HorizontalRuleNode(id: '0');
+
+      final document1 = MutableDocument(nodes: [node]);
+      final document2 = MutableDocument(nodes: [node, node]);
+
+      expect(document1 == document2, isFalse);
+    });
   });
 }
--- a/super_editor/test/src/core/document_editor_test.dart
+++ b/super_editor/test/src/core/document_editor_test.dart
@@ -22,7 +22,16 @@ void main() {
       expect(document.nodes.indexOf(newNode), 1);
     });
 
-    test('it is equal to a similar document with the same content', () {
+    test('it is equal to another document when both documents are empty', () {
+      final document1 = MutableDocument(nodes: []);
+      final document2 = MutableDocument(nodes: []);
+
+      expect(document1 == document2, isTrue);
+    });
+
+    test(
+        'it is equal to another document when each content node is equal to the corresponding node in the other document',
+        () {
       final node = HorizontalRuleNode(id: '0');
 
       final document1 = MutableDocument(nodes: [node]);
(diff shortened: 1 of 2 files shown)


matthew-carroll:
Unit tests sound like a good idea. Would you like to add some to this PR for this change?

britannio:
> Unit tests sound like a good idea. Would you like to add some to this PR for this change?

Sure

britannio:
@matthew-carroll Just added the tests!

super_editor/test/src/core/document_editor_test.dart
```diff
@@ -21,5 +21,23 @@ void main() {
       // newNode exists at index 1
       expect(document.nodes.indexOf(newNode), 1);
     });
+
+    test('it is equal to a similar document with the same content', () {
```
matthew-carroll, line 25: These test descriptions don't really describe what it means to be equal. Based on what I'm seeing, it looks like the description should read:

it is equal to another document when each content node is equal to the corresponding node in the other document

super_editor/test/src/core/document_editor_test.dart
```diff
@@ -21,5 +21,23 @@ void main() {
...
+      final document2 = MutableDocument(nodes: [node]);
+
+      expect(document1 == document2, isTrue);
+    });
+
+    test('it is NOT equal to a similar document with different content', () {
+      final node = HorizontalRuleNode(id: '0');
+
+      final document1 = MutableDocument(nodes: [node]);
+      final document2 = MutableDocument(nodes: [node, node]);
```
matthew-carroll, line 38: Tests should generally consist of situations that make sense in reality. There's no obvious situation where the same instance of a node would appear in a document, let alone back-to-back. This test should use a different node for the 2nd node in `document2`.

super_editor/test/src/core/document_editor_test.dart
```diff
@@ -21,5 +21,23 @@ void main() {
...
+    test('it is equal to a similar document with the same content', () {
+      final node = HorizontalRuleNode(id: '0');
+
+      final document1 = MutableDocument(nodes: [node]);
+      final document2 = MutableDocument(nodes: [node]);
+
+      expect(document1 == document2, isTrue);
+    });
+
+    test('it is NOT equal to a similar document with different content', () {
```
matthew-carroll, line 34: Referring to documents as "similar" is ambiguous. Similar how?

This test appears to test the following:
it is NOT equal to a document with the same starting nodes but with additional nodes at the end

super_editor/test/src/core/document_editor_test.dart
```diff
@@ -21,5 +21,23 @@ void main() {
...
+
+    test('it is equal to a similar document with the same content', () {
+      final node = HorizontalRuleNode(id: '0');
+
+      final document1 = MutableDocument(nodes: [node]);
+      final document2 = MutableDocument(nodes: [node]);
+
+      expect(document1 == document2, isTrue);
+    });
+
```
matthew-carroll, line 33: It's good to verify boundary cases and I think we're missing one here:
it is equal to another document when both documents are empty

super_editor/test/src/core/document_editor_test.dart
```diff
@@ -21,5 +21,23 @@ void main() {
...
+    test('it is NOT equal to a similar document with different content', () {
+      final node = HorizontalRuleNode(id: '0');
+
+      final document1 = MutableDocument(nodes: [node]);
+      final document2 = MutableDocument(nodes: [node, node]);
+
+      expect(document1 == document2, isFalse);
+    });
```
matthew-carroll, line 41: You've shown that a document with extra nodes will not be equal, but that means that an equality check could simply compare the `length` of the nodes and still pass these tests. It would be good to ensure that two lists with the same number of nodes are not equal when the nodes are not equal.

it is NOT equal to a document when corresponding nodes are NOT equal

matthew-carroll (changes requested):
Here are some recommendations to improve the unit tests to better communicate the purpose of the tests, and to cover more boundary conditions.

britannio:
Thanks for the feedback! Hopefully, I've addressed it all.

matthew-carroll (approved):
LGTM. Thanks.

Issue #279, opened 2021-08-11
Failed assertion: 'buttons == kPrimaryButton': is not true.


knopp:
Happens when quickly right-clicking after a left quick. `TapSequenceGestureRecognizer.isPointerAllowed` should only allow primary pointer events, but that's not always the case.

Complete StackTrace:
```
flutter: The following assertion was thrown while routing a pointer event:
flutter: 'package:super_editor/src/infrastructure/multi_tap_gesture.dart': Failed assertion: line 261 pos 12:
flutter: 'buttons == kPrimaryButton': is not true.
flutter: 
flutter: When the exception was thrown, this was the stack:
flutter: #2      TapSequenceGestureRecognizer._checkUp (package:super_editor/src/infrastructure/multi_tap_gesture.dart:261:12)
flutter: #3      TapSequenceGestureRecognizer._registerFirstTap (package:super_editor/src/infrastructure/multi_tap_gesture.dart:209:5)
flutter: #4      TapSequenceGestureRecognizer._handleEvent (package:super_editor/src/infrastructure/multi_tap_gesture.dart:123:9)
flutter: #5      PointerRouter._dispatch (package:flutter/src/gestures/pointer_router.dart:94:12)
flutter: #6      PointerRouter._dispatchEventToRoutes.<anonymous closure> (package:flutter/src/gestures/pointer_router.dart:139:9)
flutter: #7      _LinkedHashMapMixin.forEach (dart:collection-patch/compact_hash.dart:400:8)
flutter: #8      PointerRouter._dispatchEventToRoutes (package:flutter/src/gestures/pointer_router.dart:137:18)
flutter: #9      PointerRouter.route (package:flutter/src/gestures/pointer_router.dart:123:7)
flutter: #10     GestureBinding.handleEvent (package:flutter/src/gestures/binding.dart:440:19)
flutter: #11     GestureBinding.dispatchEvent (package:flutter/src/gestures/binding.dart:420:22)
flutter: #12     RendererBinding.dispatchEvent (package:flutter/src/rendering/binding.dart:278:11)
flutter: #13     GestureBinding._handlePointerEventImmediately (package:flutter/src/gestures/binding.dart:374:7)
flutter: #14     GestureBinding.handlePointerEvent (package:flutter/src/gestures/binding.dart:338:5)
flutter: #15     GestureBinding._flushPointerEventQueue (package:flutter/src/gestures/binding.dart:296:7)
```

PR #280, opened 2021-08-11
Only allow primary button pointer in TapSequenceGestureRecognizer

knopp:
Fixes https://github.com/superlistapp/super_editor/issues/279


matthew-carroll (approved):
LGTM

PR #287, opened 2021-09-09
Fix #286, Breaking upstream changes in flutter SDK

### LINKED ISSUE #286: SuperEditor doesn't build with current flutter master
Text Editing Model refactor has lended (https://github.com/flutter/flutter/commit/2382b4c046623da0cfe9cc1cafc4ad8ea1b161e7) and removed `RenderEditable.nextCharacter` (used [here](https://github.com/superlistapp/super_editor/blob/191410aba2ef54135d97616b30a17a6d2b5b0522/super_editor/lib/src/infrastructure/text_layout.dart#L120)). Same goes for previous character.

AmitPr:
As noted in #286, changes to the `RenderEditable` class removed `previousCharacter` and `nextCharacter` methods. This PR moves them into a `src/infrastructure/utilities.dart` file. I'm not too familiar with the structure of the library, but I'm fairly certain there isn't another home for these functions other than the `text_layout.dart` file itself, if that's more desirable.

Resolves #286.

### DIFF
--- a/super_editor/lib/src/infrastructure/text_layout.dart
+++ b/super_editor/lib/src/infrastructure/text_layout.dart
@@ -1,4 +1,5 @@
 import 'package:flutter/rendering.dart';
+import 'package:super_editor/src/infrastructure/utilities.dart';
 
 /// Contract to interrogate the layout of a blob of text.
 abstract class TextLayout {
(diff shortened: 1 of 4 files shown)


AmitPr:
If we’re sure these methods won’t be used elsewhere, it’s probably better to inline them within the functions in the `text_layout` file

matthew-carroll:
This issue is addressed in this PR: https://github.com/superlistapp/super_editor/pull/296

Also, for future reference, please don't name things `Utilities` or `util`. That creates "junk drawers" of functionality. They can contain anything. That makes it very difficult for developers to know where they should look for things, where new behavior should be placed, and where encapsulation boundaries should be installed.

PR #289, opened 2021-09-11
Fix #237, allowing shift clicking to expand selections in documents

### LINKED ISSUE #237: Shift+click selection for editor
When a user presses shift and taps/clicks on a text position (SuperTextField) or a content position (SuperEditor), the selection should move the extent to that position.

AmitPr:
Following in the steps of #186, this copies the implementation of Shift-Click selection expansion to work outside of just `SuperTextField`. Resolves #237.

### DIFF
--- a/super_editor/lib/src/default_editor/document_interaction.dart
+++ b/super_editor/lib/src/default_editor/document_interaction.dart
@@ -224,18 +224,30 @@ class _DocumentInteractorState extends State<DocumentInteractor> with SingleTick
 
   void _onTapDown(TapDownDetails details) {
     _log.log('_onTapDown', 'EditableDocument: onTapDown()');
-    _clearSelection();
-    _selectionType = SelectionType.position;
-
     final docOffset = _getDocOffset(details.localPosition);
     _log.log('_onTapDown', ' - document offset: $docOffset');
     final docPosition = _layout.getDocumentPositionAtOffset(docOffset);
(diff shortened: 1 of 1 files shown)


matthew-carroll:
I see that there are changes to a single tap. What happens if the user double taps or triple taps with these changes?

AmitPr:
> What happens if the user double taps or triple taps with these changes?

These changes don't modify that behavior, which is consistent with other document editors:
* Double tap selects the word the user is on, regardless of Shift being pressed
* Triple tap is the same, behavior doesn't care whether or not Shift is pressed

matthew-carroll:
@AmitPr following up on your last comment, I see that the `_selectionType` is only set for a normal tap. Can you confirm that this is the desirable behavior?

AmitPr:
Turned out to be a pretty simple fix, hope the code is self-explanatory.

matthew-carroll (approved):
This generally looks OK. There are a few details that need to be adjusted, stylistically, but I left this PR open for a long time, so I'll take care of that on my end. Also, in the future, each PR should include relevant tests. But again, since I left this PR open for a while, I'm going to ignore that in this case.

matthew-carroll:
Looks like this PR has conflicts. I tried resolving them in-place but there were too many changes to `super_editor` to make that work effectively. I've replicated these changes, with some adjustments, in #354. I'm gonna close this PR in favor of that one.

PR #291, opened 2021-09-11
Fix #290, issue with replacing selections in SuperEditor

### LINKED ISSUE #290: Replacing Selections in SuperEditor doesn't register the first keystroke
When a section of the document is selected and the user inputs any keystroke, we would expect that character to replace the entire selection. Instead the observed behavior is that the selection is deleted, but the new character is not inserted.  

Note: This is not an issue in `SuperTextField`.  

Tested Platforms: Web, macOS

AmitPr:
Fixes the issue described in #290. Resolves #290.

### DIFF
--- a/super_editor/lib/src/default_editor/document_keyboard_actions.dart
+++ b/super_editor/lib/src/default_editor/document_keyboard_actions.dart
@@ -390,7 +390,7 @@ ExecutionInstruction anyCharacterOrDestructiveKeyToDeleteSelection({
   }
 
   editContext.commonOps.deleteSelection();
-  return ExecutionInstruction.haltExecution;
+  return ExecutionInstruction.continueExecution;
 }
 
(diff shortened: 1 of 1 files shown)


matthew-carroll:
It looks like you've identified the source of the issue, but this particular change doesn't really address the broader issue.

With the introduction of this new behavior, this handler is no longer correctly named. The behavior for a destructive key, like backspace or delete, is now different from that of a content character that is pressed. For example, you mentioned `SuperTextField`, if you check the implementation, you'll see that the behavior for adding a content character includes handling an expanded selection.

The appropriate change for #290 would be to reduce this handler to just destructive keys, and then add the necessary selection inspection and content removal from the handler(s) responsible for inputing new content characters.

AmitPr:
That makes sense, unfortunately it'll probably take me too much time to get this fixed properly then, sorry for the hassle!

PR #292, opened 2021-09-12
File name compliance fix

domesticmouse:


matthew-carroll:
It looks like most of this PR isn't a file name fix. 

There's a VSCode settings file that's committed - wasn't there already one of those committed, if so, does it not apply to the example?

A bunch of lock file changes were committed. Is there a reason for that?

The rest are auto-formatting changes. They may introduce conflicts with the large PR that I have up for mobile integration for text fields, so I'd rather keep them out until after the mobile PR gets merged.

domesticmouse:
Okiday, I'll break this PR up into separate changes.

domesticmouse:
> There's a VSCode settings file that's committed - wasn't there already one of those committed, if so, does it not apply to the example?

This settings file doesn't get picked up if you open the `example` directory directly in VSCode, which is the easiest way to get VS Code happy with the example project.

PR #295, opened 2021-09-18
Add a missing timer?.cancel()

TimWhiting:


TimWhiting:
I guess technically this isn't needed because the only way for this method to be called is for the timer to complete. So you can close it if you feel it doesn't add value.

matthew-carroll:
Yeah, as you say, the purpose of that method is to run when the timer completes, so we know it's not running.

To avoid situations like this, it's helpful to file an issue with a description of something that doesn't work as intended. Then we know for sure that there's a problem to solve.

I'm going to close this PR.

PR #298, opened 2021-09-24
Rebuild runners for Flutter 2.5

domesticmouse:
Also, enforce line endings to be consistent so runners can be easily rebuilt going forward


matthew-carroll:
Can you describe what isn't working correctly for this package with the current app projects?

domesticmouse:
The reason for enforcing line endings is that the Windows runners were checked in with crlf and the rest are lf. I'm guessing the Windows runners were added on a Windows machine, but the rest on a macOS box. 

As to the reason to update runners is that the runners usually have minor fixes that are nice to have. 

If this isn't a concern, feel free to nack this PR.

matthew-carroll:
I'm fine with replacing the runners in general, but I'd like to wait until there's a material gain. Otherwise we're just introducing risk. Sometimes we get small fixes, and sometime we get small bugs. I'll close this PR for now and we can try this again when we don't have a lot of in-flight work.

PR #299, opened 2021-09-24
Drop and git ignore the `pubspec.lock` for the `super_editor` library

domesticmouse:
Split from #292


matthew-carroll:
I have the same question here. Are there things that aren't working with the current lock file? I ask because at the rate that dependencies change, we could probably churn this file every few days, but I'm not sure that would be of benefit.

domesticmouse:
An alternative would be to remove the lock file from the repo and git ignonre it. Given you aren't after reproducible builds, having the lockfile checked in just adds noise to PRs as a file that either has to be reverted or not staged for each PR created.

matthew-carroll:
Why do you say we're not interested in reproducible builds?

domesticmouse:
My understanding was that reproducible builds make sense for applications, because an application build incorporates it's dependencies at the time of build. A library which is imported as a source dependency doesn't have a freeze moment. The dependency versions a library are used with is dependent on the build of the binary it itself is included in. 

This is my understanding for why the [GitHub default Dart .gitignore](https://github.com/github/gitignore/blob/master/Dart.gitignore#L7) file includes `pubspec.lock` as an ignored file, with a comment that people building applications may want to check in their lock file.

matthew-carroll:
Ok, in that case we can get rid of the lock file for the package, but the lock file for the example app is still relevant. If people report bugs in the example app, or use the example app to reproduce bugs, then we do want a reproducible build.

domesticmouse:
Done

matthew-carroll (approved):
LGTM. Thanks.

PR #300, opened 2021-09-24
Reformat example for line length of 120

domesticmouse:


matthew-carroll:
I have the same question here as before. Isn't there already a VSCode line length format configuration file? Is it not applying to the example sub-project?

matthew-carroll:
I see you answered the question about the settings file in the other PR. Can you remove the attributions format change in this PR?

domesticmouse:
> Can you remove the attributions format change in this PR?

Done

matthew-carroll (approved):
LGTM

PR #305, opened 2021-10-04
Fixes #302: Handle Beginning/End of document arrow keys

### LINKED ISSUE #302: Unknown character inputted when trying to go beyond the end
Unknown characters are getting added to the input when we click the right arrow key when the cursor is already at the end of the text.

In the hosted web example, the text "ArrowRight" is getting entered.

https://user-images.githubusercontent.com/54908910/134962835-f6053e24-625c-4dd5-81cc-3cf616350d60.mov

In the web editor

https://user-images.githubusercontent.com/54908910/134963151-e67db8d0-41bb-4f39-bd3a-bf26e554249f.mov

brian-superlist:
When a user typed a Left Arrow at the Beginning of a Document or a Right arrow at the end of a document, SuperEditor's default keyboard action `moveUpDownLeftAndRightWithArrowKey` would `ExecutionInstruction.continueExecution`.
Then, the `anyCharacterToInsertInParagraph` would capture the Key Event and attempt to insert the Special Key into the text.

When moving the arrow key down at the same spot, the `moveUpDownLeftAndRightWithArrowKey` would actually `ExecutionInstruction.haltExecution`, which means the special key would not be inserted by `anyCharacterToInsertInParagraph`.

Therefore, the fix appeared to me to be: `haltExecution` when a left arrow key is detected at the beginning of a document or a right arrow key is detected at the end of the document. Then, the `anyCharacterToInsertInParagraph` cannot insert the special character.


brian-superlist:
Hm, looks like tests are failing due to using `RenderEditable.nextCharacter`, which is visible only for testing. There's a comment in the codebase that we may need to copy/pasta it into the project. Should I go ahead and do so?

matthew-carroll:
@brian-superlist I agree that changing the return value technically resolves this issue, but the contract for the method is no longer valid if that change is made.

It looks like the only obvious options are:

- Get rid of the `bool` response from the movement methods, or
- Change the calling method to always halt regardless of whether the arrow key causes movement
- Don't print a content character for arrow keys

Given that other areas of code may wish to utilize the same movement behavior, and those callers might care about whether a movement actually took place, I don't think it's a good idea to remove the `bool` response from the method.

I poked around briefly and couldn't find a reason to continue execution. Did you find any use-cases for arrow keys that come after that handler? We can probably get away with halting execution for now. A possible future problem is that we want to take a particular action when an arrow key is pressed at the boundary of a document. If we halt execution no matter what, then we can't install another handler to handle that situation. We'll be forced to edit an existing handler, which should be a last resort.

The final option is to adjust the handler that's printing this key. In theory, it's not supposed to matter if an arrow key makes it to the print handler because we shouldn't be trying to print an arrow character. As I mentioned in my comment on the issue ticket, web issues are predictable because Flutter doesn't appropriately report keys from web. However, I wouldn't expect any print behavior on Mac. Did you check on why that's happening?

> Hm, looks like tests are failing due to using RenderEditable.nextCharacter, which is visible only for testing. There's a comment in the codebase that we may need to copy/pasta it into the project. Should I go ahead and do so?

I have a fix for this in my custom text field PR that's open right now. I'd rather avoid conflicts with that PR.

brian-superlist:
> @brian-superlist I agree that changing the return value technically resolves this issue, but the contract for the method is no longer valid if that change is made.

Exactly, that's what lead to my question in my description above: is it safe to slightly alter the meaning of the contract to "handled arrow key" or "handled movement" instead of "did move"? Would there really be a downside in changing the nature of that  contract, since the only place `didMove` is used is from the `moveUpDownLeftAndRightWithArrowKeys` `KeyboardAction`? However, it sounds like you have plans for this particular contract and you want to keep it in place. Fair enough, definitely need to find another option :)

At the moment, I can't think of a scenario where a downstream `DocumentKeyboardAction` would want to handle arrow key presses, which is why my solution relied on halting execution. I couldn't think of scenario where you'd want to take an action, nor could I think of a good reason to continue passing this keystroke down the chain. Therefore, halting execution at this point seems to make sense to me until we have a concrete use case to allow it to continue.

  > The final option is to adjust the handler that's printing this key. In theory, it's not supposed to matter if an arrow key makes it to the print handler because we shouldn't be trying to print an arrow character. As I mentioned in my comment on the issue ticket, web issues are predictable because Flutter doesn't appropriately report keys from web. However, I wouldn't expect any print behavior on Mac. Did you check on why that's happening?

I didn't check on why it's happening, but I'm a "fail fast" kinda guy and this feels the most "defensive" approach, therefore it's my least favorite solution. I kind of thought the special character rendering was actually a helpful bug because it exposed bad logic further up the chain. Again, I have very little context on this project thus far having just dived in. If you prefer this solution, I'm happy to dive into why it's being rendered on Mac and see if we can prevent it from rendering.

matthew-carroll:
> I didn't check on why it's happening, but I'm a "fail fast" kinda guy and this feels the most "defensive" approach

I'll clarify the nature of these handlers. They sit in a Chain of Responsibility. The goal is to provide a modular, extensible structure to facilitate any number of use-case handlers, both from super_editor and from end-users. As such, the possibility of events falling out the end of the chain is a feature, not a bug. That's expected to happen for any key combination that doesn't have an associated editor behavior. Pressing an arrow key at the beginning or end of the document does not correspond to any presently desired editor behavior.

The root problem in this scenario is that a handler is responding to a key that it shouldn't respond to. The handler that's responding to this event is only supposed to turn character keys into character content. But arrow keys are control keys. Therefore, the bug is in the handler that's turning an arrow key into a String.

It's also important to recognize that handlers are about use-cases, not keyboard keys. For example, multiple handlers might care about CMD or Shift. There isn't a single handler for those keys. Similarly, the handler that moves the selection when arrow keys are pressed only exists to handle the situation where pressing an arrow key actually moves the selection. That handler is not the universal handler for all arrow key presses. Just like we don't have a universal handler for all CMD key presses.

Therefore, I think the best thing to do is determine why the arrow keys are being seen as character keys by the character input handler. Then, we should see if there's a reasonable way to eliminate that issue. In the case that there is no reasonable way to avoid turning arrow keys into content, we can accept the unfortunate change to the arrow handler and always halt execution when an arrow key is pressed. But I'd like to see that we've done the due diligence on the content handler, first.

Issue #315, opened 2021-10-13
Investigate scrolling use-cases


matthew-carroll:
Investigate needed scrolling and sliver use-cases. Determine where we have issues and what we might do about it.

matthew-carroll:
Add Task + Chat demo where chat messages are conditionally docked to bottom of viewport without `Sliver`s: https://github.com/superlistapp/super_editor/pull/316

matthew-carroll:
Proposed an approach to work with ancestor scrollables in #329 

This approach still has some issues, but it might be good enough for now. We'll see what happens in that PR.

matthew-carroll:
I merged #329 as a solution for now. We'll track and fix issues with that work as they come up. I'm going to close this investigation because I think we've handled the most critical issue.

PR #316, opened 2021-10-14
Add Task & Chat demo - investigating scrolling

matthew-carroll:
I'm investigating various scrolling use-cases. Meta ticket is #315 

In this PR I explored a situation where one displays task content, followed by messages, followed by message input.

The message input is always docked to the bottom.

The messages content is conditionally docked to the bottom - they're docked to the bottom when there isn't enough content to scroll, otherwise they're placed directly below the task content.

The purpose of this exercise was to see if such a thing could reasonably be achieved without `Sliver`s. The answer appears to be "yes". This doesn't mean we shouldn't pursue documents within `CustomScrollView`s, but this will hopefully help inform approaches and priorities.

https://user-images.githubusercontent.com/7259036/137232886-043e12e7-bfef-43aa-b32e-45de4b3e8fb7.mov

### DIFF
--- a/super_editor/example/lib/demos/scrolling/demo_task_and_chat.dart
+++ b/super_editor/example/lib/demos/scrolling/demo_task_and_chat.dart
@@ -1,0 +1,427 @@
+import 'dart:math';
+
+import 'package:flutter/cupertino.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter/rendering.dart';
+import 'package:super_editor/super_editor.dart';
+
+/// Displays a task + chat UI.
+///
+/// The content area is confined to a vertically oriented layout,
+/// similar to a "task pane" on the right side of a task app. In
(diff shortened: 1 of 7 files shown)


matthew-carroll:
The `CustomScrollView` solution is probably the better solution long term. I wanted to provide this as a stop gap. As you've found in the custom scrolling PR, there are still issues to resolve, and it's unclear how quickly we'll figure that out. So this is mostly a question for your priorities. If you need this solution quickly, then I would recommend taking this approach. If you can wait, then the `CustomScrollView` makes sense.

brian-superlist:
Excellent, thanks again for the demo and clarification on the `CustomScrollView` solution and timeframe!

matthew-carroll:
I replicated this demo code in #329. We'll go ahead and make it work with a `CustomScrollView` over there. Therefore, I'm closing this PR.

PR #320, opened 2021-10-19
Caret size doesn't match hint text

### LINKED ISSUE #314: Caret/I-Beam and Text Size is incorrect for Empty AttributedTexts
### Problem statement

We define a `ParagraphNode` with a specific `AttributedText` that includes two `SpanMarkers` at the 0 offset to set the font size to a larger / "title" / "heading" size. 

The problem: If the `ParagraphNode` is empty, SuperEditor resets the Text Style. If the user types in a character, SuperEditor calls our `AttributionStyleBuilder` with the proper inputs and we and style it correctly.

This results in a "visual" bug where the Caret / I-Beam is the wrong size at first. Then, the document shifts down and sizes the caret / i-beam correctly as you start typing.

It looks like it boils down to this section of code, which completely disables our ability to set the text style using a `AttributionStyleBuilder` since it passes in no information to the `AttributionStyleBuilder`: https://github.com/superlistapp/super_editor/blob/main/super_editor/lib/src/infrastructure/attributed_text.dart#L314

### Video

https://www.loom.com/share/74fc562f57fe4c46a5e632a8e43fcb0a

### Code Samples

<details>
  <summary>Reproduction Code</summary>

```dart

class ComputeTextSpanBug extends StatefulWidget {
  @override
  _ComputeTextSpanBugState createState() => _ComputeTextSpanBugState();
}

class _ComputeTextSpanBugState extends State<ComputeTextSpanBug> {
  final TextEditingController _controller = TextEditingController();
  late Document _doc;
  late DocumentEditor _docEditor;

  @override
  void initState() {
    super.initState();
    _doc = _createInitialDocument();
    _docEditor = DocumentEditor(document: _doc as MutableDocument);
  }

  @override
  void dispose() {
    _controller.dispose();
    super.dispose();
  }

  @override
    return SuperEditor.custom(
      editor: _docEditor,
      padding: const EdgeInsets.symmetric(vertical: 56, horizontal: 24),
      textStyleBuilder: _textStyleBuilder,
    );
  }

  TextStyle _textStyleBuilder(Set<Attribution> attributions) {
    var newStyle = defaultStyleBuilder(attributions);

    for (final attribution in attributions) {
      if (attribution == primaryHeaderAttribution) {
        newStyle = newStyle.copyWith(
          color: Colors.blue,
          fontSize: 48,
          fontWeight: FontWeight.bold,
          height: 1.0,
        );
      }
    }

    return newStyle;
  }
}

Document _createInitialDocument() {
  return MutableDocument(
    nodes: [
      ParagraphNode(
        text: AttributedText(
          text: '',
          spans: AttributedSpans(
            attributions: [
              const SpanMarker(
                attribution: primaryHeaderAttribution,
                offset: 0,
              ),
              const SpanMarker(
                attribution: primaryHeaderAttribution,
                offset: 0,
              ),
            ],
          ),
        ),
        metadata: {
          'blockType': header1Attribution,
        },
      ),
      ParagraphNode(
        text: AttributedText(
          text:
              'Nam hendrerit vitae elit ut placerat. Maecenas nec congue neque. Fusce eget tortor pulvinar, cursus neque vitae, sagittis lectus. Duis mollis libero eu scelerisque ullamcorper. Pellentesque eleifend arcu nec augue molestie, at iaculis dui rutrum. Etiam lobortis magna at magna pellentesque ornare. Sed accumsan, libero vel porta molestie, tortor lorem eleifend ante, at egestas leo felis sed nunc. Quisque mi neque, molestie vel dolor a, eleifend tempor odio.',
        ),
      ),
    ],
  );
}

class PrimaryHeaderAttribution extends NamedAttribution {
  const PrimaryHeaderAttribution() : super('NamedAttribution');
}

const primaryHeaderAttribution = NamedAttribution('header2');
````
</details>

### PR DESCRIPTION
**Pull request by matthew-carroll**
I investigated the issue where the caret doesn't match the height of the hint text. I found a few areas in the code that might contribute to the problem. To test the issue directly, I created a new component called `TextWithHintComponent` and then I painted a demo screen with multiple text hints at different sizes.

You can see in the attached video that the carets all match the hint text. Therefore, I'll need some further information to understand why this is a problem in Superlist.

https://user-images.githubusercontent.com/7259036/138003325-6ed2a6b3-2d42-47a1-a8c2-825a17a2e7aa.mov

Changes:
 - Created a demo screen for text components with hint text
 - Created ProxyDocumentComponent to easily wrap an existing DocumentComponent with decoration
 - Created ProxyTextComposable to easily wrap an existing TextComposable with decoration
 (Resolves #314)

### DIFF
--- a/super_editor/lib/src/default_editor/text.dart
+++ b/super_editor/lib/src/default_editor/text.dart
@@ -227,8 +227,8 @@ class _TextWithHintComponentState extends State<TextWithHintComponent>
       attributionsWithBlock.add(blockType);
     }
 
-    final contentStyle = widget.textStyleBuilder(attributions);
-    final hintStyle = contentStyle.merge(widget.hintStyleBuilder?.call(attributions) ?? const TextStyle());
+    final contentStyle = widget.textStyleBuilder(attributionsWithBlock);
+    final hintStyle = contentStyle.merge(widget.hintStyleBuilder?.call(attributionsWithBlock) ?? const TextStyle());
     return hintStyle;
   }
 
@@ -602,21 +602,9 @@ class _TextComponentState extends State<TextComponent> with DocumentComponent im
   Widget build(BuildContext context) {
     _log.log('build', 'Building a TextComponent with key: ${widget.key}');
 
-    Attribution? blockType = widget.metadata['blockType'];
-
-    // Surround the text with block level attributions.
-    final blockText = widget.text.copyText(0);
-    if (blockType != null) {
-      blockText.addAttribution(
-        blockType,
-        TextRange(start: 0, end: widget.text.text.length - 1),
-      );
-    }
-    final richText = blockText.computeTextSpan(widget.textStyleBuilder);
-
     return SuperSelectableText(
       key: _selectableTextKey,
-      textSpan: richText,
+      textSpan: widget.text.computeTextSpan(_textStyleWithBlockType),
       textAlign: widget.textAlign ?? TextAlign.left,
       textDirection: widget.textDirection ?? TextDirection.ltr,
       textSelection: widget.textSelection ?? const TextSelection.collapsed(offset: -1),
@@ -625,6 +613,18 @@ class _TextComponentState extends State<TextComponent> with DocumentComponent im
       textCaretFactory: TextCaretFactory(color: widget.caretColor),
       highlightWhenEmpty: widget.highlightWhenEmpty,
     );
+  }
+
+  /// Creates a `TextStyle` based on the given [attributions], plus any
+  /// "block type" that's specified in [widget.metadata].
+  TextStyle _textStyleWithBlockType(Set<Attribution> attributions) {
+    final attributionsWithBlockType = Set<Attribution>.from(attributions);
+    Attribution? blockType = widget.metadata['blockType'];
+    if (blockType != null) {
+      attributionsWithBlockType.add(blockType);
+    }
+
+    return widget.textStyleBuilder(attributionsWithBlockType);
   }
 }
 
--- a/super_editor/example/lib/demos/components/demo_text_with_hint.dart
+++ b/super_editor/example/lib/demos/components/demo_text_with_hint.dart
@@ -1,0 +1,90 @@
+import 'package:flutter/cupertino.dart';
+import 'package:flutter/material.dart';
+import 'package:super_editor/super_editor.dart';
+
+/// Example of various [TextWithHintComponent] visual configurations.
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by matthew-carroll** [comment]
@brian-superlist I've updated this PR with an approach that I think handles your situation. Can you confirm that what's in the demo can fit your requirements?

Admittedly, I think the demo shows that too much work is needed for this capability, but I want to make sure I've identified a solution before I look into refactoring for a better API.

**Comment by matthew-carroll** [comment]
@brian-superlist I rebased this PR and updated it to use the hint component that I added. The block level attribution is now always passed to the regular text and the hint text style builders, and the hint text now supports its own `AttributedText` instead of just a `String`.

Do you wanna give this a try and see how it works for your needs?

For your idea about an `InheritedWidget`, I think there's an argument for an overall typography `InheritedWidget`, but I think this issue should be solved locally. For overall document styling, we'll need to figure out how to provide useful configurations for developers that also allow extensions. If you consider the standard Material `Theme`, it's convenient when all you want to style are Material things. But there's no way to extend `Theme` to help style your own widgets. We're going to run into a similar problem with `SuperEditor`.

File: super_editor/lib/src/default_editor/text.dart [hunk 1]
```diff
@@ -626,6 +614,18 @@ class _TextComponentState extends State<TextComponent> with DocumentComponent im
       highlightWhenEmpty: widget.highlightWhenEmpty,
     );
   }
+
+  /// Creates a `TextStyle` based on the given [attributions], plus any
+  /// "block type" that's specified in [widget.metadata].
+  TextStyle _textStyleWithBlockType(Set<Attribution> attributions) {
+    final attributionsWithBlockType = Set<Attribution>.from(attributions);
+    Attribution? blockType = widget.metadata['blockType'];
```
brian-superlist, line 622: This looks like it would certainly work for our Superlist use case since this is specifically how we're storing this information! However, is this a good API for `SuperEditor` in general? Would it make more sense for `TextNodes` to have an explicit and properly typed `final Attribution baseAttribution` field instead, which this method could read? Superlist could also adjust to that typed API as well, it would make the feature more discoverable for everyone, and you wouldn't run into type issues here if someone uses `blockType` incorrectly.

Anyhow, sad to say today is my last day at Superlist! Really cool and difficult project you're working on here, been fun to work with and learn from ya :)

miguelcmedeiros (approved):
Tested it out in our app and `TextWithHintComponent` is doing what it's supposed to do 👍

PR #329, opened 2021-10-29
Making SuperEditor work in a CustomScrollView

matthew-carroll:
I'm working on figuring out what it will take to make `SuperEditor` work in a `CustomScrollView`.

I separated all keyboard handling from gesture handling.

Within the gesture interactor, I added behavior to check for an ancestor `Scrollable`. If there is an ancestor `Scrollable` then the gesture interactor bases all of its position and auto-scrolling behaviors on the ancestor.

https://user-images.githubusercontent.com/7259036/139408329-e944e718-7299-48c7-ad2f-add597414a88.mov

### DIFF
--- a/super_editor/example/lib/demos/scrolling/demo_task_and_chat.dart
+++ b/super_editor/example/lib/demos/scrolling/demo_task_and_chat.dart
@@ -1,0 +1,427 @@
+import 'dart:math';
+
+import 'package:flutter/cupertino.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter/rendering.dart';
+import 'package:super_editor/super_editor.dart';
+
+/// Displays a task + chat UI.
+///
+/// The content area is confined to a vertically oriented layout,
+/// similar to a "task pane" on the right side of a task app. In
+/// that content area is a task definition at the top, followed by
+/// some messages, and then capped off by a message input `TextField`.
+///
+/// The message input field is always docked to the bottom of the display area.
+///
+/// When the content + messages are shorter than the viewport, the messages
+/// are docked to the bottom of the viewport.
+///
+/// When the content + messages are taller than the viewport, the messages are
+/// placed directly below the content, like a `Column`.
+///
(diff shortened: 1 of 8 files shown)


matthew-carroll:
@brian-superlist I checked in the repo you linked and the sliver demo in there seems like it's still showing the image. I tried commenting out everything in the `SliverAppBar` on my end and everything still works OK.

Can you link to the specific files that reproduce the issue so that I can copy them over to my repo and quickly try them out?

matthew-carroll:
@brian-superlist nevermind. You were talking about removing the image in the document, not the app bar. Now I see the layout issue.

matthew-carroll:
@brian-superlist I think I've fixed the layout issue. I think it might be a bug in `IntrinsicHeight` so I adjusted the editor layout to avoid needing that. Remove the `IntrinsicHeight` widget and see if things work for you now.

I also moved the task and chat demo over to this PR. Can you let me know which slivers you're using to take up available space when they can, and then shrink to zero when there is no available space? I'll update the task and chat demo to use a `CustomScrollView` in a similar fashion.

brian-superlist:
> Can you let me know which slivers you're using to take up available space when they can, and then shrink to zero when there is no available space? I'll update the task and chat demo to use a CustomScrollView in a similar fashion.

Sure thing, you can see a demo & `SliverStickToBottom` class here:

https://github.com/brian-superlist/super_editor/blob/use-cases/super_editor/example/lib/demos/header_footer_use_case.dart

https://github.com/brian-superlist/super_editor/blob/use-cases/super_editor/example/lib/demos/sliver_stick_to_bottom.dart

matthew-carroll:
@miguelcmedeiros - I'm gonna merge this PR so that I can begin the mobile work.

I updated the PR to include 2 versions of the task+chat UI - one with a custom `RenderBox` and one with a bottom docked `Sliver` in a `CustomScrollView`. Both of them seem to work equally well, which provides initial validation for this approach. Please give this is a try in Superlist whenever you can, and then file any issues that you find.

PR #336, opened 2021-11-02
Add current DocumentSelection to ComponentContext (Resolves #334)

### LINKED ISSUE #334: Add document selection to ComponentContext
The `ComponentContext` is passed to each component builder so that the builder can make decisions about whether or not to return a component, and how to configure a component that is returned.

The `ComponentContext` already provides access to the `Document`, current `DocumentNode`, and current `NodeSelection`.

There are situations where the creation of an individual component requires knowledge about the document selection.

Example:
Consider the Medium editor with a title and subtitle at the top of the document. When the user selects any part of the title, both the title and subtitle show side-bar labels that read "Title" and "Subtitle". Therefore, the title component makes a decision based on selection in the subtitle and vis-a-versa.

matthew-carroll:
The ComponentContext is passed to each component builder so that the builder can make decisions about whether or not to return a component, and how to configure a component that is returned.

The ComponentContext already provides access to the Document, current DocumentNode, and current NodeSelection.

There are situations where the creation of an individual component requires knowledge about the document selection.

Example:
Consider the Medium editor with a title and subtitle at the top of the document. When the user selects any part of the title, both the title and subtitle show side-bar labels that read "Title" and "Subtitle". Therefore, the title component makes a decision based on selection in the subtitle and vis-a-versa.

### DIFF
--- a/super_editor/lib/src/core/document_layout.dart
+++ b/super_editor/lib/src/core/document_layout.dart
@@ -413,6 +413,7 @@ class ComponentContext {
   const ComponentContext({
     required this.context,
     required this.document,
+    this.documentSelection,
     required this.documentNode,
     required this.componentKey,
     required this.showCaret,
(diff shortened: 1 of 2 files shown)


matthew-carroll:
FYI @brian-superlist @roughike

PR #337, opened 2021-11-02
Feature/allow text component to take optional trailing span

jamessuperlist:
One use case is to be able to align, for instance, an icon at the end of a TextComponent, both in single and multiline scenarios.

Bildschirmfoto 2021-11-02 um 08 47 40

This cannot be achieved using Row or Wrap as alignment in a multiline scenario will not be correct.

Bildschirmfoto 2021-11-02 um 08 48 57
Bildschirmfoto 2021-11-02 um 08 49 22

This PR adds an optional `WidgetSpan` trailing to TextComponent, thus achieving the first screenshot as seen above.

For `Text`, a custom `TextEditingController` can be extended with `buildTextSpan` overriden to similarly achieve desired functionality. Thus another potential solution may be to introduce a controller for TextComponent.

Looking forward to your feedback.

### DIFF
--- a/super_editor/lib/src/default_editor/text.dart
+++ b/super_editor/lib/src/default_editor/text.dart
@@ -276,6 +276,7 @@ class TextComponent extends StatefulWidget {
     this.caretColor = Colors.black,
     this.highlightWhenEmpty = false,
     this.showDebugPaint = false,
+    this.trailing,
   }) : super(key: key);
 
   final AttributedText text;
@@ -289,6 +290,7 @@ class TextComponent extends StatefulWidget {
   final Color caretColor;
   final bool highlightWhenEmpty;
   final bool showDebugPaint;
+  final WidgetSpan? trailing;
 
   @override
   _TextComponentState createState() => _TextComponentState();
@@ -612,7 +614,12 @@ class _TextComponentState extends State<TextComponent> with DocumentComponent im
         TextRange(start: 0, end: widget.text.text.length - 1),
       );
     }
-    final richText = blockText.computeTextSpan(widget.textStyleBuilder);
+    final richText = TextSpan(
+      children: [
+        blockText.computeTextSpan(widget.textStyleBuilder),
+        if (widget.trailing != null) widget.trailing!
+      ],
+    );
 
     return SuperSelectableText(
       key: _selectableTextKey,


matthew-carroll:
Hi. There may be a reasonable set of feature requests that involve inline widgets. That set of behaviors needs to be spec'd out. We would need to collect a representative sample of diverse use-cases to fully understand the problem that we're solving.

We definitely shouldn't add a special parameter for a trailing widget. There's nothing special about the end of a line vs the beginning of a line vs the middle of a line. There's also nothing special about 1 widget vs 2+ widgets. 

For this particular case, have you tried subclassing `TextComponent` for your needs and then adding support for your own component to the list of component builders? I believe that would get you what you need without adding such a specific capability to `super_editor`.

jamessuperlist:
> For this particular case, have you tried subclassing `TextComponent` for your needs and then adding support for your own component to the list of component builders? I believe that would get you what you need without adding such a specific capability to `super_editor`.

Thanks @matthew-carroll for the feedback. Let me try this out and then I'll report back.

I agree that there is nothing special about trailing vs leading, single or multiple widgets. Nevertheless I'd imagine there is a use case that could satisfy most users, but without questionnaires and community feedback, this is presently unknown.

jamessuperlist:
> For this particular case, have you tried subclassing `TextComponent` for your needs and then adding support for your own component to the list of component builders?

@matthew-carroll I do not see any component builder for `TextComponent`. `TextComponent` is stateful, with build used to return a `SuperSelectableText` with the correct `textSpan`. One can extend `TextComponent`, however there is no access to `_TextComponentState` nor `build` as by design the state is private. Maybe I am missing something?

Thanks in advance for considering this issue and appreciate the help :)

matthew-carroll:
@jamessuperlist I didn't quite follow that description, but can you check out this guide and let me know if this clarifies how to create a custom component?

https://github.com/superlistapp/super_editor/wiki/How-to:-custom-UI-component

jamessuperlist:
> can you check out this guide and let me know if this clarifies how to create a custom component?
> 
> https://github.com/superlistapp/super_editor/wiki/How-to:-custom-UI-component

Thanks for the documentation. I'll close this PR and instead open an issue with regard to community feedback on leading/trailing etc. for TextComponent. With regard to creating custom components, I think the most feasible for now is to create a custom component which is basically a fork of TextComponent.

PR #344, opened 2021-11-09
Fix: pressing character during selection deletes selection and adds character (Resolves #290)

### LINKED ISSUE #290: Replacing Selections in SuperEditor doesn't register the first keystroke

matthew-carroll:
Fix: pressing character during selection deletes selection and adds character (Resolves #290)

The problem was that the handler responsible for deleting the selection didn't look for a pressed character key, it just deleted the selection and returned.

Now, the handler also inserts a character if one is pressed.

### DIFF
--- a/super_editor/example/pubspec.lock
+++ b/super_editor/example/pubspec.lock
@@ -354,7 +354,7 @@ packages:
       name: vector_math
       url: "https://pub.dartlang.org"
     source: hosted
-    version: "2.1.0"
+    version: "2.1.1"
   win32:
     dependency: transitive
     description:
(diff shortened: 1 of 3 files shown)


matthew-carroll:
CC @AmitPr

miguelcmedeiros (commented):
I tested it out on macOS and works as expect. 

I noticed that we are also not handling the escape key as well. When we select text and press escape, it deletes the selection and introduces a space. From looking at other apps I've seen 2 possible behaviours: do nothing (text remains selected) or collapse the selection. 
 
Not sure if you prefer this as a separate issue/PR

matthew-carroll:
Thanks @miguelcmedeiros, please file that as a new ticket. I believe this PR resolves the particular concern in its source issue.

PR #345, opened 2021-11-11
Make SuperEditor configurable for keyboard vs IME, and mouse vs touch (Resolves #341)

### LINKED ISSUE #341: Allow developers to choose keyboard vs IME and mouse vs touch
As we add support for mobile and IME interaction, `SuperEditor` needs to make it possible to choose the appropriate input modalities. The obvious combinations are:

- mouse + keyboard
- mouse + IME
- touch + IME

Currently, IME deltas are only available on mobile, but eventually all OSes should have IME delta support in Flutter. Therefore, desktop/web developers might choose direct keyboard input or IME input.

As for mouse vs touch, that's pretty much desktop/web vs mobile. In this case, the difference isn't the input device, but rather the interpretation of gestures.

For this ticket, stub the IME and touch interactors and adjust `SuperEditor` so that they're configurable.

matthew-carroll:
Make SuperEditor configurable for keyboard vs IME, and mouse vs touch (Resolves #341)

`SuperEditor` takes an `inputSource` of either `keyboard` or `ime`, and a `gestureMode` of either `mouse` or `touch`.

For now, using `ime` or `touch` shows an error widget because they're not implemented.

Most of this PR is minor adjustments in other areas. I filled in some missing dart docs, re-arranged some properties, and cleaned up the `SuperEditor` constructors because I seem to have made them factories for no reason.

The next step for mobile support will likely be touch interaction.


matthew-carroll:
@miguelcmedeiros I'm going to merge this so that I can continue with the mobile work. Please let me know if there's something we can do for a quicker turn around time on average PR reviews.

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -307,20 +318,63 @@ class _SuperEditorState extends State<SuperEditor> {
...
+        return DocumentImeInteractor(
+          focusNode: _focusNode,
+          editContext: _editContext,
+          child: child,
+        );
+    }
+  }
+
+  /// Builds the widget tree that handles user gesture interaction
+  /// with the document, e.g., mouse input on desktop, or touch input
```
miguelcmedeiros, line 351: I guess we need to prioritize one over the other, but devices like iPad's will allow for both concurrently.

I think I've seen an app that allowed user to choose for what kind of input the app should be optimized, but I can't name it. Not sure if that's the way forward, but at least with this API we will able to offer such flexibility 👍

PR #352, opened 2021-11-18
Failing test case for document equality with identical TextNodes

roughike:
It seems that document equality is broken for documents that contain `TextNode`s. This makes it a bit cumbersome for automated tests where we want to verify that a certain mapping layer produces the correct output.

I tried to investigate this myself, but wasn't able to fix it in a reasonable time, so I decided to submit a failing test case instead.

(feel free to take over and fix it)


matthew-carroll:
I'm gonna close this in favor of #355

PR #357, opened 2021-12-01
Support Flutter shortcuts from SuperEditor (Resolves #356)

### LINKED ISSUE #356: Handle keyboard shortcuts outside SuperEditor
Currently it is not possible to handle keyboard shortcuts outside `SuperEditor` once the editor gets focus. If we have a widget that handles keyboard shortcuts (e.g. a page widget) and has `SuperEditor` as a descendant, then no keyboard shortcut events bubble up to the ancestor's handler.

If we have a page with a document, then we should be able to handle shortcuts such as "⌘1", "⌘2", "⌘3", etc., independently of the document being in focus or not (unless the document itself handles those shortcuts...).

The following sample can be used to reproduce this issue.
```dart

class Sample extends StatefulWidget {
  @override
  _SampleState createState() => _SampleState();
}

class _SampleState extends State<Sample> {
  String _message = '';

  @override
    return FocusScope(
      child: FocusableActionDetector(
        autofocus: true,
        shortcuts: const {
          SingleActivator(LogicalKeyboardKey.digit1, meta: true):
              MetaAndDigitIntent(1),
          SingleActivator(LogicalKeyboardKey.digit2, meta: true):
              MetaAndDigitIntent(2),
          SingleActivator(LogicalKeyboardKey.digit3, meta: true):
              MetaAndDigitIntent(3),
        },
        actions: {
          MetaAndDigitIntent: CallbackAction<MetaAndDigitIntent>(
            onInvoke: (intent) => setState(
              () => _message = 'Meta + ${intent.value}',
            ),
          ),
        },
        child: SafeArea(
          child: Column(
            children: [
              Text(_message),
              Expanded(
                child: SuperEditor(
                  editor: DocumentEditor(
                    document: MutableDocument(
                      nodes: [
                        ParagraphNode(
                          text: AttributedText(text: 'Random paragraph....'),
                        ),
                      ],
                    ),
                  ),
                ),
              ),
            ],
          ),
        ),
      ),
    );
  }
}

class MetaAndDigitIntent extends Intent {
  const MetaAndDigitIntent(this.value);
  final int value;
}
```

matthew-carroll:
Support Flutter shortcuts from SuperEditor (Resolves #356)

This approach takes every key that `SuperEditor` receives and runs it through the a `ShortcutsManager`, if one exists in the widget tree. This should trigger any Flutter shortcut for the key combo, and if it does, the editor won't try to handle that key.

The issue with this approach is that we're not supposed to call the method on `ShortcutsManager` that runs the shortcut. But I don't see any other way to know if a shortcut is registered.

I'll clean this up before merging, if we decide that this is the right answer.


matthew-carroll:
Based on conversation with @miguelcmedeiros, this change probably only works when there's a single `Shortcuts` widget in the tree. However, this should solve Superlist's problem so I'm gonna merge this in. We can work on a more thorough solution later.

PR #358, opened 2021-12-07
Implement mobile gesture controls

matthew-carroll:
Implements drag handles, magnifiers, and floating toolbars with Android and iOS styling.

A demo for each platform style is added to the demo app.

I think that there are some opportunities to further de-duplicate various scrolling and handle behaviors, but in the interest of delivering full mobile support ASAP, I stopped trying to refactor at the code structure in this PR.

There are platform agnostic tools in `document_gestures_touch.dart`. Those tools are used by the platform-specific implementations.

The Android-style gesture control implementation is in `document_gestures_touch_android.dart`.

The iOS-style gesture control implementation is in `document_gestures_touch_ios.dart`.

I pulled out a couple auto-scrolling tools that are shared. First, there's an `AutoScroller` that can animate scrolling up/down whenever you tell it to. This isn't unique to mobile or documents, it just combines a per-frame `Ticker` with and a `ScrollPosition` so that we don't need to repeat the auto-scroll animation code everywhere.

Second, there's a `DragHandleAutoScroller`, which is built on top of `AutoScroller`. The `DragHandleAutoScroller` takes a viewport rectangle, along with a given handle offset in that rectangle, and then `DragHandleAutoScroller` makes decisions about whether to auto-scroll up, auto-scroll down, or stop auto-scrolling. Here again, we'd have to implement this for both Android and iOS if I didn't pull it out into it's own class.

I defined a `ScrollableDocument` widget that wraps the given `child` with a `SingleChildScrollView`, but only if there's no ancestor `Scrollable`. This is a widget structure that's needed by both Android and iOS.

Some of these scrollable tools may apply to desktop use-cases, too, but again, I'm trying to focus on moving us forward as quickly as possible right now, so I didn't dig up any desktop code.

Android:

https://user-images.githubusercontent.com/7259036/144971896-e91bba53-dc0b-4a85-9733-65e044e3130a.mov

iOS:

https://user-images.githubusercontent.com/7259036/144971951-e6bd8d23-23fc-4e08-8881-8c0569c856fc.mov


super_editor/lib/src/default_editor/text.dart
```diff
@@ -321,7 +321,7 @@ class _TextComponentState extends State<TextComponent> with DocumentComponent im
     }
 
     final offset = getOffsetForPosition(nodePosition);
-    final lineHeight = _selectableTextKey.currentState!.getLineHeightAtPosition(nodePosition);
+    final lineHeight = _selectableTextKey.currentState!.getHeightForCaret(nodePosition)!;
```
miguelcmedeiros, line 324: While testing this out I hit the following exception on desktop when the paragraph is empty:
```sh
flutter: ══╡ EXCEPTION CAUGHT BY SCHEDULER LIBRARY ╞═════════════════════════════════════════════════════════
flutter: The following _CastError was thrown during a scheduler callback:
flutter: Null check operator used on a null value
flutter: 
flutter: When the exception was thrown, this was the stack:
flutter: #0      _TextComponentState.getRectForPosition (package:super_editor/src/default_editor/text.dart:324:88)
flutter: #1      _DefaultDocumentLayoutState.getRectForPosition (package:super_editor/src/default_editor/layout.dart:133:37)
flutter: #2      _DocumentMouseInteractorState._ensureSelectionExtentIsVisible (package:super_editor/src/default_editor/document_gestures_mouse.dart:240:49)
flutter: #3      _DocumentMouseInteractorState._onSelectionChange.<anonymous closure> (package:super_editor/src/default_editor/document_gestures_mouse.dart:224:9)
flutter: #4      SchedulerBinding._invokeFrameCallback (package:flutter/src/scheduler/binding.dart:1145:15)
flutter: #5      SchedulerBinding.handleDrawFrame (package:flutter/src/scheduler/binding.dart:1090:9)
flutter: #6      SchedulerBinding._handleDrawFrame (package:flutter/src/scheduler/binding.dart:996:5)
flutter: #10     _invoke (dart:ui/hooks.dart:150:10)
flutter: #11     PlatformDispatcher._drawFrame (dart:ui/platform_dispatcher.dart:270:5)
flutter: #12     _drawFrame (dart:ui/hooks.dart:114:31)
flutter: (elided 3 frames from dart:async)
flutter: ════════════════════════════════════════════════════════════════════════════════════════════════════
```

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -163,7 +174,13 @@ class SuperEditor extends StatefulWidget {
   final DocumentInputSource inputSource;
 
   /// The `SuperEditor` gesture mode, e.g., mouse or touch.
-  final DocumentGestureMode gestureMode;
+  final DocumentGestureMode? gestureMode;
+
+  /// Builder that creates a floating toolbar when running on Android.
+  final WidgetBuilder? androidToolbarBuilder;
```
miguelcmedeiros, line 180: `SuperEditor` could implicitly provide these toolbars by default (with appropriate handling of actions) and expose instead an optional generic builder that the developer can specify an arbitrary set of actions (copy, paste, cut, translate, format, etc.).

miguelcmedeiros:
While testing these changes on a iOS device, I noticed that the tapping on a selection does not hide/show the toolbar (as it seems to be a common behaviour). Is this something easy to implement in this PR?

super_editor/lib/src/default_editor/text.dart
matthew-carroll: Can you provide reproduction steps for this, and any other errors you run into?

super_editor/lib/src/default_editor/super_editor.dart
matthew-carroll: That's true, but are these the right default toolbars? I didn't put much thought into this, I just threw together a couple toolbars that roughly approximate the platforms.

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -414,7 +458,8 @@ enum DocumentInputSource {
 
 enum DocumentGestureMode {
   mouse,
-  touch,
+  android,
```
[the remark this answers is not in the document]
matthew-carroll: If you check the `SuperEditor` constructor you'll see that passing these values is optional. Therefore, it's not a question of what developers have to do, but what they choose to do. We should avoid forcing our bias on developers wherever possible. As you alluded to in your comment, without this configuration we wouldn't even be able to demo this behavior without deploying to the type of device we're interested in. That's a pretty good indication that this configuration should be a choice for the end-user.

super_editor/lib/src/default_editor/text.dart
miguelcmedeiros: To reproduce we just need to place the caret in an empty paragraph. This will trigger the exception.

As for as other exceptions I didn't find anything else while testing.

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -163,7 +174,13 @@ class SuperEditor extends StatefulWidget {
...
 
   /// The `SuperEditor` gesture mode, e.g., mouse or touch.
-  final DocumentGestureMode gestureMode;
+  final DocumentGestureMode? gestureMode;
+
+  /// Builder that creates a floating toolbar when running on Android.
+  final WidgetBuilder? androidToolbarBuilder;
+
+  /// Builder that creates a floating toolbar when running on iOS.
+  final WidgetBuilder? iOSToolbarBuilder;
```
miguelcmedeiros, line 183: > but are these the right default toolbars?

I guess they are as good as it gets for now until more effort is put into it to improve fidelity.

Also, by looking around (e.g. Medium), this kind of toolbars are not specific to mobile. In Superlist we have it for macOS and it would be great if this is supported on all platforms by SuperEditor.

This should not block this PR, but should be considered in the future.

super_editor/lib/src/default_editor/super_editor.dart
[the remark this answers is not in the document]
miguelcmedeiros: I guess it just feels strange that this is so explicit in the public API of this widget.

Usually widgets infer in which platform the app is running and then pick the appropriate direction (usually using `Platform`, which it's also done here when the value of `gestureMode` is null). In case the developer should have a way to override this, there are other ways to do so (like what Material does with `TargetPlatform` as part of `ThemeData`).

Again, nothing that blocks the amazing work done in this PR, but wanted to share my 2 cents on this one :)

super_editor/lib/src/default_editor/super_editor.dart
matthew-carroll: Let's implement the toolbars at another time. The paste functionality may require some thought. Or, if you'd like to take a crack at one of the toolbars in a PR then I can take a look. But I assume that Superlist would prefer to have IME support over toolbar implementations.

super_editor/lib/src/default_editor/super_editor.dart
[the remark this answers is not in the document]
matthew-carroll: Your comment is on an `enum` so I'm not sure what you mean by "so explicit".

When we're done with mobile support, `SuperEditor` will auto-select this option based on platform, but even with that auto-selection, there's no reason to deprive a developer of an easy selection. 

The use of inherited widgets to make decisions like this is largely an anti-pattern. If a developer instantiates a `SuperEditor` and wants to configure something in that `SuperEditor`, then forcing the developer to add another widget higher in the tree is extra work and extra complexity. The primary reason to do such a thing with a `Theme` is because the same choices need to apply to dozens of different widgets throughout the tree (text fields, buttons, text, etc). In this case, we're talking about the configuration of a single widget.

Lastly, despite the names in the `enum`, these aren't really platforms. These are style choices. Do you want a "mouse style" interaction, an "Android style" interaction, or an "iOS style" interaction. An iOS developer is welcome to choose an Android style interaction. That might be a bad idea, but if you start restricting decisions to save users from poor choices, you'll pretty quickly find that you're restricting a lot of reasonable choices, too. This problem is becoming more clear in Flutter, itself, and some of those restrictions are the very reason that we have to build super_editor, so I'd like to avoid repeating them.

matthew-carroll:
@miguelcmedeiros I fixed the line height null pointer issue. I also added a toolbar toggle when the user taps on an expanded selection on iOS. To do that, I implemented a few extension methods that inspect the order/affinity of given document positions. I added tests for that new selection inspection behavior.

Please give the PR another try. If you're happy with it then please approve it and I'll merge in.

miguelcmedeiros:
> I fixed the line height null pointer issue.

I can confirm that there is no more exceptions when deleting the content of the paragraph, but...
Now the caret is not rendered after deleting the content. I made a quick recording about it:
https://www.loom.com/share/8d82809d939e482789c0eb8477f1e458

miguelcmedeiros:
> I also added a toolbar toggle when the user taps on an expanded selection on iOS. To do that, I implemented a few extension methods that inspect the order/affinity of given document positions. I added tests for that new selection inspection behavior.

🙌

matthew-carroll:
@miguelcmedeiros I pushed an update that estimates caret height when the `RenderParagraph` has a dirty layout. This seems to fix the issue on my end.

miguelcmedeiros:
@matthew-carroll 
I pushed one extra commit that removes the default value for `gestureMode`. Without this change, `gestureMode` would never be `null` and so, `SuperEditor` would not try to resolve it in case the developer does not specify it.

miguelcmedeiros (approved):
LGTM

matthew-carroll:
@miguelcmedeiros thanks for reviewing. I have a couple followups on review practices. I'd prefer that only the author of a PR push changes to the PR branch, unless the author gives express permission. Pushing to other peoples' PR branches increases the likelihood of conflicting branches that need to be nuked, and the likelihood of undesired or unexpected changes.

In this case, for example, that parameter was set to "mouse" on purpose. We don't have support for IME input and therefore no one wants a mobile UI right now. This was just step one in that direction. I carved out a place for that runtime decision to be made in the future, when we're ready.

Also, I'd prefer that only the original author merge in a PR, after appropriate approvals. In this case, I could have caught the change you made and reversed it, but because this was merged, I never had the opportunity.

miguelcmedeiros:
> I'd prefer that only the author of a PR push changes to the PR branch

👍 

> I'd prefer that only the original author merge in a PR

I actually misread your previous comment and read "please approve it and **~~I'll~~** merge in." 🤦‍♂️ 
Sorry about that.

> In this case, for example, that parameter was set to "mouse" on purpose. We don't have support for IME input and therefore no one wants a mobile UI right now. This was just step one in that direction. I carved out a place for that runtime decision to be made in the future, when we're ready.

We are planning to use this intermediate version of SuperEditor in our app, so that documents become easy to read/scroll on mobile, and for the purpose of providing early feedback. The previous version provided a complete broken experience on mobile (scrolling through the document), while this makes it usable (kind of read-only experience since IME input is not yet supported).
If you feel strongly about it, I can make a PR to revert that last commit.

PR #362, opened 2021-12-11
Implemented non-selectable components (Resolves #359)

### LINKED ISSUE #359: Add support for non-selectable document components
Let's take Medium as an example. The horizontal rule is not selectable. The user can select across the horizontal rule, but there is no visual styling when the selection spans the HR, and tapping on the HR has no effect. Thus, an HR can find itself within a larger selection, but cannot be selected, itself.

Add support for this use-case for any component/node.

matthew-carroll:
Implemented non-selectable components (Resolves #359)

Includes a test suite that contrasts the behavior between selectable and non-selectable horizontal rules.

https://user-images.githubusercontent.com/7259036/145670217-ff2ee625-43d4-4f8b-b256-cc69fdf916ef.mov

### DIFF
--- a/super_editor/example/lib/demos/components/demo_unselectable_hr.dart
+++ b/super_editor/example/lib/demos/components/demo_unselectable_hr.dart
@@ -1,0 +1,79 @@
+import 'package:flutter/material.dart';
+import 'package:super_editor/super_editor.dart';
+
+/// Example of a horizontal rule component that is not directly selectable.
+///
+/// The user can select around the horizontal rule, but cannot select it, specifically.
+class UnselectableHrDemo extends StatefulWidget {
+  @override
+  _UnselectableHrDemoState createState() => _UnselectableHrDemoState();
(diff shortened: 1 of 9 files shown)


miguelcmedeiros (commented):
Should the unelectable horizontal rules be excluded from copied text (Cmd+C)?

As it is now, the horizontal rule (`---`) will be part of the copied text, regardless of the horizontal rule being selectable or not.

matthew-carroll:
Yeah the HR should be a part of the copied text, but this change is also only targeting user interactions. The copy behavior can pick up the responsibility for deciding what to copy or not. You'll also notice that something being unselectable doesn't really mean it can't be selected, it just means it can't sit at the end of a selection. The user can't tap to select it, and if the user is expanding a selection, the user can't end the selection on the unselectable component.

miguelcmedeiros (approved):
🚀

PR #365, opened 2021-12-12
fixes shift+tab exception for non ListItemNode

### LINKED ISSUE #364: [bug] shift + tab throws an Exception when pressed outside in non-ListItemNode
To reproduce:
- put the selection on a paragraph
- press shift + tab

```
The following _CastError was thrown while processing the key message handler:
type 'ParagraphNode' is not a subtype of type 'ListItemNode' in type cast
```

Expected Behavior:
The key event should be ignored for non-ListItemNode and propagate to the rest of the keyboardActions/app.

osaxma:
fixes #364 

I added a condition for `unindentListItem` to skip `shift + tab` if the node isn't a `ListItemNode`

### DIFF
--- a/super_editor/lib/src/default_editor/common_editor_operations.dart
+++ b/super_editor/lib/src/default_editor/common_editor_operations.dart
@@ -1910,7 +1910,7 @@ class CommonEditorOperations {
 
     final baseNode = editor.document.getNodeById(composer.selection!.base.nodeId);
     final extentNode = editor.document.getNodeById(composer.selection!.extent.nodeId);
-    if (baseNode!.id != extentNode!.id) {
+    if (baseNode!.id != extentNode!.id || baseNode is! ListItemNode) {
       return false;
     }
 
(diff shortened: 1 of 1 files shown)


super_editor/lib/src/default_editor/common_editor_operations.dart
```diff
@@ -1807,7 +1807,7 @@ class CommonEditorOperations {
 
     final baseNode = editor.document.getNodeById(composer.selection!.base.nodeId);
     final extentNode = editor.document.getNodeById(composer.selection!.extent.nodeId);
-    if (baseNode!.id != extentNode!.id) {
+    if (baseNode!.id != extentNode!.id || baseNode is! ListItemNode) {
```
matthew-carroll, line 1810: Thanks for the fix. Can you do me a favor and make these 2 different if-statements? Throughout the code like this, you'll see that I generally keep independent conditions in independent if-statements. I think this makes it much easier for developers to quickly audit the various conditions that reject a behavior, rather than requiring very careful reading of unrelated conditions in a single if-statement.

osaxma: Makes sense. The statement is split now.

matthew-carroll (approved):
LGTM

osaxma:
Hi @matthew-carroll 

Just in case you're waiting for me to merge this, I can't do that btw. 

Screen Shot 2021-12-30 at 6 26 28 AM

matthew-carroll:
Sorry, I thought GitHub allowed that with an appropriate admin approval. I'll merge this in.

PR #367, opened 2021-12-14
[Bugfix] Show caret on desktop/web

### LINKED ISSUE #366: [Bug] Caret is not shown on macOS/web
Steps to reproduce:
* Checkout commit 751e2354d63fca7b287e8a555ec249e4f6ebc33c
* Run demo app on macOS/web
* Tap on some piece of text

**Expected:** caret is shown where the user tapped
**Actual:** caret is not shown, although the document has focus and can be edited

miguelcmedeiros:
Show the caret on desktop/web based on the passed or resolved value of `SuperEditor.gestureMode`.

Fixes #366

### DIFF
--- a/super_editor/lib/src/default_editor/super_editor.dart
+++ b/super_editor/lib/src/default_editor/super_editor.dart
@@ -331,6 +331,18 @@ class _SuperEditorState extends State<SuperEditor> {
     }
   }
 
+  DocumentGestureMode get _gestureMode {
+    if (widget.gestureMode != null) {
+      return widget.gestureMode!;
+    } else if (Platform.isAndroid) {
+      return DocumentGestureMode.android;
+    } else if (Platform.isIOS) {
+      return DocumentGestureMode.iOS;
+    } else {
+      return DocumentGestureMode.mouse;
+    }
+  }
+
   @override
   Widget build(BuildContext context) {
     return _buildInputSystem(
@@ -368,7 +380,7 @@ class _SuperEditorState extends State<SuperEditor> {
   Widget _buildGestureSystem({
     required Widget child,
   }) {
-    switch (_resolvedGestureMode) {
+    switch (_gestureMode) {
       case DocumentGestureMode.mouse:
         return DocumentMouseInteractor(
           focusNode: _focusNode,
@@ -423,7 +435,7 @@ class _SuperEditorState extends State<SuperEditor> {
             document: widget.editor.document,
             documentSelection: _composer.selection,
             componentBuilders: widget.componentBuilders,
-            showCaret: _focusNode.hasFocus && _resolvedGestureMode == DocumentGestureMode.mouse,
+            showCaret: _focusNode.hasFocus && _gestureMode == DocumentGestureMode.mouse,
             margin: widget.padding,
             componentVerticalSpacing: widget.componentVerticalSpacing,
             extensions: {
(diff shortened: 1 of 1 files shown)


matthew-carroll:
@miguelcmedeiros can you explain what this change is meant to do? It looks like some code was moved but I'm not seeing any impactful changes.

miguelcmedeiros:
> 

In `_buildGestureSystem` we had the logic to resolve the value of `gestureMode`, and with it use the appropriate document interactor. In `_buildDocumentLayout` we were using directly `widget.mouse` instead of also resolving the correct value.

The changes in this PR are to extract the resolution logic from and reuse in both methods.

matthew-carroll:
@miguelcmedeiros how does re-use solve the caret issue? What went wrong with the caret, and how does this fix whatever went wrong?

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -434,7 +423,7 @@ class _SuperEditorState extends State<SuperEditor> {
             document: widget.editor.document,
             documentSelection: _composer.selection,
             componentBuilders: widget.componentBuilders,
-            showCaret: _focusNode.hasFocus && widget.gestureMode == DocumentGestureMode.mouse,
+            showCaret: _focusNode.hasFocus && _resolvedGestureMode == DocumentGestureMode.mouse,
```
miguelcmedeiros, line 426: @matthew-carroll 
If I understand the intention here, showCaret should be true if the document has focus and gestureMode is mouse. Since widget.gestureMode is by default null, then showCaret was always false. 
With the fix, the comparison will be made against widget.gestureMode (if not null) or the appropriate value for the host platform.

matthew-carroll: Ok, so the root problem is that `showCaret` was getting set to `false` in some cases where we expected it to be `true`.

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -368,18 +368,7 @@ class _SuperEditorState extends State<SuperEditor> {
   Widget _buildGestureSystem({
     required Widget child,
   }) {
-    late DocumentGestureMode gestureMode;
```
matthew-carroll, line 371: Based on your latest description of the problem, that `showCaret` was sometimes `false` when it should be `true` because `widget.gestureMode` was `null`, here's what I suggest:

Bring back the if-statement. There's no need to break out a method for this. The gesture mode doesn't appear to be accessed from any place outside of `_buildGestureSystem()`.

You should be able to fix this problem by just changing `widget.gestureMode` to `gestureMode` on the `showCaret` line.

matthew-carroll (changes requested):
Please see my earlier comment in the code.

super_editor/lib/src/default_editor/super_editor.dart
miguelcmedeiros: `gestureMode` is accessed in `_buildGestureSystem` and `_buildDocumentLayout` (where `showCaret` is set), hence why the logic was extracted to a getter.

matthew-carroll: It looks like GitHub collapsed the code in just the right way to make it all look like one method.

In this case, if we're going to offer a getter method to get a non-null value, then I'd like to name the getter as just `_gestureMode`. The word `resolved` tends to introduce new questions about what's being resolved, and how, and at that point the person reading the code is spending more time than before.

So the getter would be:

```dart
DocumentGestureMode get _gestureMode => [put if-statement here]
```

And I would put that getter definition above the build() methods and beneath the State lifecycle methods.

miguelcmedeiros: Done ✅

matthew-carroll (approved):
LGTM

PR #368, opened 2021-12-16
[Bugfix] Use caret color from selections styles extensions (if defined)

miguelcmedeiros:
### Problem ###
It is possible to define a color for the caret using the selections styles extension, but in practice (for paragraphs only), this value is ignored. Instead, the text selection color (defined in the same extension) is used as the caret color.

### Solution ###
As done for other node types (e.g. list items or block quote), use the value coming from the extension (if defined).


matthew-carroll (approved):
LGTM

Issue #372, opened 2021-12-30
Chrome error: Unsupported operation: Platform._operatingSystem


Josehdez96:
When the app runs in Chrome and Edge it shows:

`
Error: Unsupported operation: Platform._operatingSystem
`

It turns out this issue is in Flutter Web, from July 2019: https://github.com/flutter/flutter/issues/36126. If you use platform.isAndroid or platform.isIOS, it'll throw an exception. It was a little bug introduced 10 days ago.

**Reproduction steps:**
Open the project on web (Chrome or Edge), it'll throw the error there.

As it was commented on the issue (link); we could use foundation here. I'll send my PR fixing this.

PR #373, opened 2021-12-30
fix(super_editor.dart): foudation to avoid web exception

Josehdez96:
Bug fixed related to https://github.com/superlistapp/super_editor/issues/372 issue


osaxma:
I would like to suggest a different approach here if I may:

I believe it's better to use `defaultTargetPlatform` so that the appropriate gesture mode is used for mobile devices even in web. `defaultTargetPlatform` is also from the foundation library and works for all platforms (i.e. it handles both `dart:io` and `dart:html`).

Here's what the code would like if it was used. 
```dart
DocumentGestureMode get _gestureMode {
  if (widget.gestureMode != null) {
    return widget.gestureMode!;
  }

  switch (defaultTargetPlatform) {
    case TargetPlatform.android:
      return DocumentGestureMode.android;
    case TargetPlatform.iOS:
      return DocumentGestureMode.iOS;
    case TargetPlatform.macOS:
    case TargetPlatform.linux:
    case TargetPlatform.windows:
    case TargetPlatform.fuchsia:
      return DocumentGestureMode.mouse;
  }
}
```

Josehdez96:
@osaxma I think it really is a better idea, so I'll make the change.

matthew-carroll:
Can you look into why the tests are failing?

osaxma:
It looks like the tests are failing due to the following reason outlined in [`foundation/platform.dart`](https://github.com/flutter/flutter/blob/c99ed373b3d254f9b10789ddf3771c3aa39e48ca/packages/flutter/lib/src/foundation/platform.dart#L20):
```dart
/// In a test environment, the platform returned is [TargetPlatform.android]
/// regardless of the host platform. (Android was chosen because the tests were
/// originally written assuming Android-like behavior, and we added platform
/// adaptations for iOS later). Tests can check iOS behavior by using the
/// platform override APIs (such as [ThemeData.platform] in the material
/// library) or by setting [debugDefaultTargetPlatformOverride].
```


The failing tests are:
- [editor_entry_smoke_test.dart](https://github.com/superlistapp/super_editor/blob/148224e5381251101def23ed7f7b00a4d51c2077/super_editor/test/editor_smoke_tests/editor_entry_smoke_test.dart#L12)
- [unselectable_component_test.dart](https://github.com/superlistapp/super_editor/blob/148224e5381251101def23ed7f7b00a4d51c2077/super_editor/test/src/default_editor/unselectable_component_test.dart#L19)


There are two options to workaround this:
1- Overriding the `debugDefaultTargetPlatformOverride` based on the testing platform.
2- Explicitly passing `gestureMode: DocumentGestureMode.mouse` to the `SuperEditor` instances created in those tests (i.e. [here](https://github.com/superlistapp/super_editor/blob/148224e5381251101def23ed7f7b00a4d51c2077/super_editor/test/editor_smoke_tests/editor_entry_smoke_test.dart#L37), [here](https://github.com/superlistapp/super_editor/blob/148224e5381251101def23ed7f7b00a4d51c2077/super_editor/test/src/default_editor/unselectable_component_test.dart#L379) and [here](https://github.com/superlistapp/super_editor/blob/148224e5381251101def23ed7f7b00a4d51c2077/super_editor/test/src/default_editor/unselectable_component_test.dart#L392)).

I believe the second option is fine for those tests because both of them are sending `keyboard` events so they're specific to desktop anyway. 

---
\*  The errors that were making the tests fail were all coming from `document_gestures_touch_android.dart`.

matthew-carroll:
I think @osaxma's recommended change makes sense. @Josehdez96 can you update with @osaxma's changes and also try updating the tests to pass? Then I can take another look.

Josehdez96:
@matthew-carroll Done.

matthew-carroll (approved):
LGTM. Thanks.

PR #374, opened 2021-12-30
Integrate IME for mobile doc editing

matthew-carroll:
Integrates IME input with document editing on Android and iOS.

The example app includes a demo for iOS and Android. Each one is styled appropriately, but regardless of which demo you run, the IME connection that it's using is based on the underlying platform. Therefore, even if the Android example is run on iOS, it's still using the standard iOS IME deltas.

Known issues:

- There are timing issues with caret placement. The caret often appears in the wrong place.
- The first time that the user taps a selection location in a document, the delete button doesn't work. This might be an Android bug, or it might be a Flutter bug. I don't think it's a super_editor bug.
- On iOS, selecting an auto-correction from the iOS keyboard inserts the desired changes, but doesn't delete the old content that it's replacing. I think this is probably a Flutter bug.
- The selection handles sit on top of the docked editing toolbar above the keyboard.
- If the user selects content that appears beneath the keyboard, there's no auto-scrolling.


matthew-carroll:
@miguelcmedeiros @jethro87 - Can each of you try out this PR for mobile document editing?

I'd like to figure out the minimal set of changes/fixes from this point to where each of you can ship your desired features.

Jethro87:
@matthew-carroll I've run `Mobile Editing - Android` and `Mobile Editing - iOS` editor configs on the iOS simulator, my physical iOS device, and a physical Android device, and I'm unable to make the keyboard appear (thus unable to type).

On the iOS simulator, selecting 'Toggle Software Keyboard' results in this output:

```
flutter: INFO: 2021-12-30 13:10:49.521762: Handling key press: RawKeyDownEvent#28a2e(logicalKey: LogicalKeyboardKey#0006a(keyId: "0x0000006a", keyLabel: "J", debugName: "Key J"), physicalKey: PhysicalKeyboardKey#7000d(usbHidUsage: "0x0007000d", debugName: "Key J"))
flutter: [document_keyboard_actions.dart] - deleteExpandedSelectionWhenCharacterOrDestructiveKeyPressed: Running...
flutter: FINER: 2021-12-30 13:10:49.616007: Received key event, but ignoring because it's not a down event: RawKeyUpEvent#6eb6d(logicalKey: LogicalKeyboardKey#0006a(keyId: "0x0000006a", keyLabel: "J", debugName: "Key J"), physicalKey: PhysicalKeyboardKey#7000d(usbHidUsage: "0x0007000d", debugName: "Key J"))
```

I get the same output when I type any letter using my physical keyboard in the iOS simulator:

```
flutter: INFO: 2021-12-30 13:16:15.770037: Handling key press: RawKeyDownEvent#e39a6(logicalKey: LogicalKeyboardKey#00064(keyId: "0x00000064", keyLabel: "D", debugName: "Key D"), physicalKey: PhysicalKeyboardKey#70007(usbHidUsage: "0x00070007", debugName: "Key D"))
flutter: [document_keyboard_actions.dart] - deleteExpandedSelectionWhenCharacterOrDestructiveKeyPressed: Running...
flutter: FINER: 2021-12-30 13:16:15.903071: Received key event, but ignoring because it's not a down event: RawKeyUpEvent#3645f(logicalKey: LogicalKeyboardKey#00064(keyId: "0x00000064", keyLabel: "D", debugName: "Key D"), physicalKey: PhysicalKeyboardKey#70007(usbHidUsage: "0x00070007", debugName: "Key D"))
```

Can you point me in the right direction here?

osaxma:
In `Mobile Editing - iOS`, when inserting a new line, I'm getting the exceptions below for the following cases:

- When the line is inserted at the end of a paragraph, the newline is not inserted and the second attempt hides the keyboard.
- When inserting the line in the middle of a paragraph, the newline is inserted yet the exception still appears.

<details><summary> click to expand</summary>
```
The following _CastError was thrown while dispatching notifications for DocumentComposer:
Null check operator used on a null value

#0      _IOSDocumentTouchInteractorState._positionCaret
package:super_editor/…/default_editor/document_gestures_touch_ios.dart:707
#1      _IOSDocumentTouchInteractorState._onSelectionChange
package:super_editor/…/default_editor/document_gestures_touch_ios.dart:249
#2      ChangeNotifier.notifyListeners
package:flutter/…/foundation/change_notifier.dart:308
#3      new DocumentComposer.<anonymous closure>
package:super_editor/…/core/document_composer.dart:20
#4      ChangeNotifier.notifyListeners
package:flutter/…/foundation/change_notifier.dart:308
#5      ComposerPreferences.clearStyles
package:super_editor/…/core/document_composer.dart:120
#6      _SuperEditorState._updateComposerPreferencesAtSelection
package:super_editor/…/default_editor/super_editor.dart:305
#7      ChangeNotifier.notifyListeners
package:flutter/…/foundation/change_notifier.dart:308
#8      DocumentComposer.selection=
package:super_editor/…/core/document_composer.dart:41
#9      CommonEditorOperations.insertBlockLevelNewline
package:super_editor/…/default_editor/common_editor_operations.dart:1592
#10     _DocumentImeInteractorState.updateEditingValueWithDeltas
package:super_editor/…/default_editor/document_input_ime.dart:215
#11     TextInput._handleTextInputInvocation
package:flutter/…/services/text_input.dart:1557
#12     MethodChannel._handleAsMethodCall
package:flutter/…/services/platform_channel.dart:404
#13     MethodChannel.setMethodCallHandler.<anonymous closure>
package:flutter/…/services/platform_channel.dart:397
#14     _DefaultBinaryMessenger.setMessageHandler.<anonymous closure>
package:flutter/…/services/binding.dart:389
#15     _DefaultBinaryMessenger.setMessageHandler.<anonymous closure>
package:flutter/…/services/binding.dart:386
#16     _invoke2.<anonymous closure> (dart:ui/hooks.dart:189:15)
#20     _invoke2 (dart:ui/hooks.dart:188:10)
#21     _ChannelCallbackRecord.invoke (dart:ui/channel_buffers.dart:42:5)
#22     _Channel.push (dart:ui/channel_buffers.dart:132:31)
#23     ChannelBuffers.push (dart:ui/channel_buffers.dart:329:17)
#24     PlatformDispatcher._dispatchPlatformMessage (dart:ui/platform_dispatcher.dart:544:22)
#25     _dispatchPlatformMessage (dart:ui/hooks.dart:83:31)
(elided 3 frames from dart:async)
The DocumentComposer sending notification was: Instance of 'DocumentComposer'

Null check operator used on a null value

```
</details>

osaxma:
> On iOS, selecting an auto-correction from the iOS keyboard inserts the desired changes, but doesn't delete the old content that it's replacing. I think this is probably a Flutter bug.

Btw, this is working fine on my end (except for the caret placement after the replacement -- I guess that's related to the first issue). I'm using `Flutter 2.8.1 stable` with `iPhone 13 (iOS 15) simulator`.  

Another note, I believe the `selection controls` should disappear after any type of insertion or deletion (auto correct or typing).

matthew-carroll:
@Jethro87 are you running the example app, or trying to use it in your own app? If you're using the example app, did you open the mobile-specific examples?

matthew-carroll:
@osaxma for the newline, are you just pressing the action button on the keyboard to get the newline, or are you doing something else?

osaxma:
I am using the action button on the virtual keyboard. It looks like the error happens because `_docLayout.getRectForPosition` method is returning `null` for some reason:

Screen Shot 2021-12-31 at 1 57 06 AM

Jethro87:
> @Jethro87 are you running the example app, or trying to use it in your own app? If you're using the example app, did you open the mobile-specific examples?

@matthew-carroll I'm running the example app. `super_editor --> example --> main.dart`, then I'm toggling `Mobile Editing - iOS` and `Mobile Editing - Android` in `Editor Configs` in the Drawer. 

I was originally running in Flutter master, then switched to stable with the same issue.

matthew-carroll:
@Jethro87 what happens when you tap around the document? Does the caret and/or handles appear? Does the keyboard come up when you place the caret?

Jethro87:
> Is it possible that you're using the latest version of super_editor on master, rather than this PR branch? Those mobile examples already exist on master but they don't have any of the IME input that's implemented in this PR.

That's it, @matthew-carroll, thanks. I initially was going to run this in my own app, as I had things set up from the last time. After switching to running the example, I neglected to check out the proper branch. Working now. Will test it thoroughly tomorrow.

Jethro87:
The following two points look to be the same as above https://github.com/superlistapp/super_editor/pull/374#issuecomment-1003192192

- Tapping return to create a new line mid-paragraph splits the paragraph, as expected, but the carat does not move. Tapping return again closes the keyboard instead of creating another new line. Here's the output when tapping to create a new line:

```
flutter: Composer preferences changed
flutter: Another exception was thrown: Instance of 'ErrorSummary'
flutter: Another exception was thrown: Instance of 'ErrorSummary'
```

- Tapping return to create a new line at the end of a paragraph creates a new line (carat does not move as per the above point), but tapping into the new, blank line is very difficult. I have to close the keyboard before I'm able to tap into the new line. When I'm finally able to get the cursor into the new line, tapping any character closes the keyboard and does not insert the character.

---

The following points are for iOS.

- Holding the spacebar on iOS should (I believe) bring up the magnifier as tapping and holding on the document does, and allow the user to move the carat over the document.

- When a piece of content is selected, the toolbar options become unselectable. This means that I am unable to, for example, bold several words within a paragraph. I understand inline formatting options are not yet available in the toolbar, so perhaps this is why this is happening right now.

- I don't seem to be able to delete the image at the top of the document, by attempting to select it or by backspacing from the title.

- Selecting content across nodes (paragraph and list item, for example) and attempting to delete sometimes does nothing. This seems to reliably occur when selecting the first paragraph, the following blockquote, and a list item. Here's the output when selecting and tapping delete (nothing ends up being deleted):

```
flutter: Composer preferences changed
flutter: INFO: 2021-12-31 11:52:01.515878: Received edit deltas from platform
flutter: INFO: 2021-12-31 11:52:01.516227: Instance of 'TextEditingDeltaDeletion'
flutter: INFO: 2021-12-31 11:52:01.516326: Applying delta: Instance of 'TextEditingDeltaDeletion'
flutter: FINE: 2021-12-31 11:52:01.516523: Deleting text: quam blandit eget.
This is a blockquote!
This is an unordered list item
flutter: FINE: 2021-12-31 11:52:01.516638: Deleted range: TextRange(start: 412, end: 483)
flutter: FINE: 2021-12-31 11:52:01.516775: Creating doc selection from IME selection: TextSelection(baseOffset: 412, extentOffset: 483, isDirectional: false)
flutter: FINE: 2021-12-31 11:52:01.516962: Returning doc selection without modification
```

- Tapping return in a list item creates a new list item with the correct content (if applicable), but the carat stays in the original list item. It should be moved to the new list item.

- When changing a paragraph node to a header node (and vice versa), the carat doesn't resize.

- When deleting an empty list item (with a list item above it), the carat does not move to the beginning of the empty line. It stays aligned with the other list items, rather than moving to the beginning of the empty line.

- With the carat in a list item, tapping H1 or H2 does nothing, but the other formatting options work as expected (ordered/unordered list item, blockquote).

miguelcmedeiros:
I also noticed that when we change node type using the keyboard toolbar, that it leaves the caret in the same coordinates with the same size. See the following video:
https://user-images.githubusercontent.com/31278849/147961565-dc486a93-ca6b-4d98-a3be-55227b2ac808.MP4

matthew-carroll:
@Jethro87 @osaxma @miguelcmedeiros - the caret timing issue should be resolved with the latest push

miguelcmedeiros:
> @Jethro87 @osaxma @miguelcmedeiros - the caret timing issue should be resolved with the latest push

I can confirm :)

Jethro87:
The following points are for Android.

- When double-tapping on a word, the formatting toolbar appears (Cut/Copy/Paste/Select All), as expected. However, when I close the keyboard, the formatting toolbar is never removed, despite the selection and handles disappearing. This toolbar should be removed when the keyboard is closed.

- When the single selection handle is visible (for a few seconds after tapping a word), if you format the paragraph to H1, for example, the selection handle doesn't move to the proper placement when the text size changes. Here's what this looks like:

Screenshot_20220104-094422

- If you press return in the middle of a paragraph (thus it is split into two paragraphs with a space in between them), then format one of the paragraphs, both will be formatted. I would assume only one should be formatted. Example:

Screenshot_20220104-095301

matthew-carroll:
@Jethro87 can you describe the difficulty with selecting an empty paragraph? I know there was a caret timing issue, and there was an issue with inserting new characters, both of which should now be fixed. But you also mentioned that you had a hard time selecting the empty paragraph. I just tried tapping anywhere horizontally within an empty paragraph and it's bringing the keyboard up for me without issue.

Jethro87:
@matthew-carroll Selecting empty paragraphs seems to work fine now. Perhaps it had something to do with the carat and inserting new characters issue that you just solved.

I did notice another empty paragraph issue while playing around this morning. If you tap into the last list item (in either the unordered or ordered list) and press return, a new list item is created. All good so far. If you press return again, the list item is removed and carat unindented, but I am unable to type a new character, add another empty line, or remove the empty line. I have to hide the keyboard and tap into the empty line in order for it to work as expected.

A couple of other issues I noticed:

- When I slightly miss the tap area (usually in an empty line), the keyboard closes. This is a bit jarring, particularly when there are several blank lines on top of each other. It often takes me several taps, with the keyboard opening and closing, to get the carat where I want it to be. What happens in the Notes app is the carat is placed in the nearest node; the keyboard is never closed. 
- I find the selection handles a bit hard to select. Often I miss the tap area and just scroll the page. Perhaps the tap area can be slightly enlarged? Don't know if this is just me, but thought I'd mention it.

miguelcmedeiros:
> I find the selection handles a bit hard to select. Often I miss the tap area and just scroll the page. Perhaps the tap area can be slightly enlarged? Don't know if this is just me, but thought I'd mention it.

I noticed the same on Android and iOS.

miguelcmedeiros:
@matthew-carroll 

With the latest changes (not sure if it was occurring before), if the selection has multiple nodes and then backspace is pressed, then not all the selected content is deleted. Here's a GIF with an example running `Mobile Editing - Android` on an Android device (same running `Mobile Editing - iOS` on an iPhone):
ezgif com-video-to-gif-6

miguelcmedeiros:
Currently when opening mobile editing demos, the keyboard is open although there is no selection at the point.
Ideally, when a user opens opens a document, it's not immediately in "edit mode".

miguelcmedeiros:
@matthew-carroll, would it be possible to configure what happens when return is pressed?

Currently it adds new a paragraph (which is fine for built-in components), but for custom components (e.g. task items), we need to be able to configure it.

matthew-carroll:
@Jethro87 @miguelcmedeiros - I've pushed a number of commits. Let's re-sync on the known issues at this point.

On my list, the following still need to be address:

- Make all areas of the document tappable to select content so that the user doesn't need to be so accurate with tapping
- Auto-scroll to extent when it's offscreen (including underneath the docked toolbar)
- Make it easier to grab selection handles
- Add iOS spacebar long-press trackpad behavior
- Make keyboard action button response configurable

@Jethro87 you mentioned shared styles when splitting a text node. That was fixed in another PR. We'll inherit that change when I merge `master` into this branch.

Also @miguelcmedeiros can you describe the various behaviors that you need to implement for action button (newline button) on the keyboard? I want to make sure that responding to the button, itself, is the right thing to do.

miguelcmedeiros:
> Also @miguelcmedeiros can you describe the various behaviors that you need to implement for action button (newline button) on the keyboard? I want to make sure that responding to the button, itself, is the right thing to do.

The use case that I have in mind for this is for custom nodes. In this case we have a task node (i.e. has at title, checkbox and other metadata). 
When the selection is in a node of this type, and the user enters a new line, we want that newly inserted node to be of the same type instead of a `ParagraphNode`.

For desktop we solved this with a keyboard action that given the context (selection is in a task node) and the "enter" key is pressed. Aside from inserting a task node, the action will also create a task in our app state.

For mobile, I'm sure how to achieve this.

miguelcmedeiros:
@matthew-carroll, it seems that on iOS the keyboard theme is always in light mode regardless of what is set in OS settings.
ezgif com-video-to-gif-7

miguelcmedeiros:
@matthew-carroll, will `SuperTextField` also IME in the PR or in a followup?

osaxma:
@matthew-carroll 

I've two things to report:

1- Exception: Expected a BinaryNodePosition for position1 but received a TextNodePosition


<details><summary> expands to see logs </summary>

```
The following _Exception was thrown while handling a gesture:
Exception: Expected a BinaryNodePosition for position1 but received a TextNodePosition

#0      HorizontalRuleNode.selectUpstreamPosition
package:super_editor/…/default_editor/horizontal_rule.dart:29
#1      InspectDocumentSelection.doesSelectionContainPosition
package:super_editor/…/core/document_selection.dart:341
#2      _IOSDocumentTouchInteractorState._onTapUp
package:super_editor/…/default_editor/document_gestures_touch_ios.dart:357
#3      TapSequenceGestureRecognizer._checkUp.<anonymous closure>
package:super_editor/…/infrastructure/multi_tap_gesture.dart:347
#4      GestureRecognizer.invokeCallback
package:flutter/…/gestures/recognizer.dart:198
#5      TapSequenceGestureRecognizer._checkUp
package:super_editor/…/infrastructure/multi_tap_gesture.dart:347
#6      TapSequenceGestureRecognizer._registerFirstTap
package:super_editor/…/infrastructure/multi_tap_gesture.dart:287
#7      TapSequenceGestureRecognizer._handleEvent
package:super_editor/…/infrastructure/multi_tap_gesture.dart:162
#8      PointerRouter._dispatch
package:flutter/…/gestures/pointer_router.dart:94
#9      PointerRouter._dispatchEventToRoutes.<anonymous closure>
package:flutter/…/gestures/pointer_router.dart:139
#10     _LinkedHashMapMixin.forEach (dart:collection-patch/compact_hash.dart:539:8)
#11     PointerRouter._dispatchEventToRoutes
package:flutter/…/gestures/pointer_router.dart:137
#12     PointerRouter.route
package:flutter/…/gestures/pointer_router.dart:123
#13     GestureBinding.handleEvent
package:flutter/…/gestures/binding.dart:439
#14     GestureBinding.dispatchEvent
package:flutter/…/gestures/binding.dart:419
#15     RendererBinding.dispatchEvent
package:flutter/…/rendering/binding.dart:322
#16     GestureBinding._handlePointerEventImmediately
package:flutter/…/gestures/binding.dart:374
#17     GestureBinding.handlePointerEvent
package:flutter/…/gestures/binding.dart:338
#18     GestureBinding._flushPointerEventQueue
package:flutter/…/gestures/binding.dart:296
#19     GestureBinding._handlePointerDataPacket
package:flutter/…/gestures/binding.dart:279
#23     _invoke1 (dart:ui/hooks.dart:169:10)
#24     PlatformDispatcher._dispatchPointerDataPacket (dart:ui/platform_dispatcher.dart:293:7)
#25     _dispatchPointerDataPacket (dart:ui/hooks.dart:88:31)
(elided 3 frames from dart:async)
Handler: "onTapUp"
Recognizer: TapSequenceGestureRecognizer#f1393
```

</details>

to reproduce:

https://user-images.githubusercontent.com/46427323/149016451-fff970fc-51a3-424f-b737-8e843f318a5e.mov






2- While I was trying to record one, I really struggle with conflict between moving the handle and scrolling.

First recording of me struggling 😅

https://user-images.githubusercontent.com/46427323/149016628-c1f0dcab-9616-4974-8230-8e8dae8de2e3.mov


Maybe gestures around the drag handles should take priority over scrolling.

osaxma:
3- it looks like there's something blocking the gestures here.

https://user-images.githubusercontent.com/46427323/149018204-0dc76680-8135-422e-88e0-e66061b76568.mov

4- i'm not sure if this will be covered by a different PR but the following aren't working: 
  - cut, copy and paste. 
  - Triple click to select entire paragraph.
  - tapping on a selection highlight should hide the selection controls & tapping again should show it.

osaxma:
> can you describe the various behaviors that you need to implement for action button (newline button) on the keyboard? I want to make sure that responding to the button, itself, is the right thing to do.

I think the button should be configurable between node changes because some nodes may require showing a specific `TextInputAction`. The button type would provide an indication to the user of what's gonna happen when it's pressed. 

For instance, I've a non-TextNode that can be edited in-place by text (e.g. inserting/editing the URL of some embedded content). Showing `TextInputAction.done` is more appropriate here since the user is editing one-line only. Clicking done, will hide the text and show the embedded content instead of inserting a newline (currently I'm using a dialog for this).

Another use-case that I can think of is if an app has a table node, then showing `TextInputAction.next` would be the appropriate to indicate that the action will move to the cursor to the next cell. Here's an example from the Notes App on iPhone:

https://user-images.githubusercontent.com/46427323/149024883-436a97c4-d500-47e7-8b40-5438f37bf738.mp4


I think the user should be able to update the `inputAction` on node changes somehow if necessary as well as have a way to listen to the action (i.e. `TextInputClient.performAction` events). 

--- 
Side Note: IIRC `TextInputAction.newline` and `TextInputAction.unspecified` (falls back to the OS default) are the only actions that insert a newline to the text and hence are the only one that causes `updateEditingValue` or `updateEditingValueWithDeltas` to be invoked with `\n`.  

FYI, `TextInputAction.none` throws an exception on iOS so that should be taken into account if the property is exposed. There could be other actions that works for one platform and not the other so that also need to be confirmed.

Jethro87:
@matthew-carroll Things are looking good. Apart from the summary of issues you mentioned, and the additional ones mentioned by osama and miguel, I've found only one additional issue from my testing today.

When you select text and the action bar appears (Cut/Copy/Paste/etc), then scroll, the action bar follows the scrolling activity even when the selection is off-screen. On iOS, the bar follows all the way when scrolling up, and only follows until the selection is off-screen when scrolling down:

https://user-images.githubusercontent.com/6467808/149035844-322cc5e9-e752-4e3b-80cc-67f8cfaa678b.mov

On Android, the action bar follows the scroll behavior fully, whether scrolling up or down:

https://user-images.githubusercontent.com/6467808/149036298-424cba30-23b6-4d80-b6c2-85f7905d70f9.mov

I think the default behavior here should be that the action bar sticks with the selected content. It should "flip" above or below the content if necessary (you've already implemented this on iOS), but it shouldn't follow the scrolling when the selected content is off-screen.

matthew-carroll:
Here's another status update.

Completed since last time:

- Made the whole document tappable to make it easier to select content
- Auto-scroll to show selection when keyboard opens
- Keyboard respects platform brightness
- Fixed document selection bug that @osaxma reported
- Fixed triple tap to select paragraph on iOS (it was already working on Android)
- Fixed toolbar tap-to-toggle on iOS - is Android supposed to have this same behavior or should a tap always set the caret?
- Updated Android and iOS popover toolbar location based on @Jethro87's comment

In progress:

- Configure keyboard action button

To do:

- Improve drag interaction with handles. I'm not sure what the exact problem is here. The touch area is already pretty large on iOS, and yet it still seems to cause vertical scrolling more than it should.
- iOS keyboard trackpad
- Prevent overlay controls from sitting on top of mounted toolbar

matthew-carroll:
Here's another status report:

 * I implemented the ability to configure and respond to the keyboard action button. However, Flutter appears to have a bug where instructing the IME to `updateConfig()` does not actually update the IME config. Therefore, at the moment, I don't think it's possible to change the keyboard action button. If I'm wrong, please let me know.
 * I've implemented iOS floating cursor support
 * I've added the ability to clip the mobile overlay controls with any desired `CustomClipper`. This can be used, for example, to prevent the drag handles from appearing on top of the toolbar that sits on the keyboard.

@Jethro87 @osaxma - I've tried reproducing the iOS handle grab issue, but I can't get any consistent reproduction. I'm lucky if I hit a problem 1 out of 10 times. Are either of you able to consistently reproduce a problem with grabbing the handles? I know something is wrong there, but if I can't reproduce it then I can't fix it.

At this point, the only item remaining on my list is the handle grabbing issues.

miguelcmedeiros:
@matthew-carroll 
Any chance for `SuperTextField` to be updated in this PR with regards to IME support?

osaxma:
@matthew-carroll 

If you mean the conflict with scrolling, dragging the handle vertically always reproduce for me (iOS 15 simulator). 


--- 

Another issue that I just notice:
- place the selection before the image on the example app (tapping on the image should do).
- press enter on the IME
- two issues:
   1.  keyboard closes
   2. a line is inserted after the image instead of before the image

I haven't confirmed if this is also an issue on the main branch with desktop/web.

miguelcmedeiros:
@matthew-carroll 

On Android ("Mobile Editing - Android" or "Mobile Editing - iOS" demo) pressing the return key does nothing.
ezgif com-gif-maker

Tested this on a real device and emulator. On an iOS device it seems to be working properly.

Jethro87:
@matthew-carroll 

For the iOS handle grab issue, it feels like the tap area is very forgiving when tapped on the inside of the selection handle. For example, grabbing the left-side of the selection handle at the end of the content nearly always works for me. Same with the opposite - grabbing the right-side of the selection handle at the beginning of the content. If I try to touch the center or outer portion of the selection handle, I tend to "miss" more often.

I also have trouble grabbing the selection handle at the beginning of the selection and dragging "up". I wonder if the formatting toolbar is blocking the "up" movement. Instead, if I grab the selection handle and first drag left or right, _then_ up, it seems to work fine.

I can reproduce @osaxma issue re: placing selection before image (on iOS physical device).

Here's a few more issues I found:

1. Selecting a word (bringing up the formatting toolbar), then single tapping elsewhere in the document results in the formatting toolbar appearing where the original selection occurred, rather than above (or below) the new selection tap area.

https://user-images.githubusercontent.com/6467808/150848611-0ed68530-2af3-435d-907f-52d9994ae34a.mov

2. There is still no auto-scrolling below the keyboard (not sure if this had been addressed yet).

https://user-images.githubusercontent.com/6467808/150848814-f9ff2031-ffdd-4a98-b646-ca7152c3d54c.mov

3. Highlighting a selection brings up the formatting toolbar (as expected), but when the selection is deleted, the toolbar remains. It should disappear.

https://user-images.githubusercontent.com/6467808/150848936-189c6cad-4dc6-4937-83bd-932d266525f8.mov

4. Splitting a paragraph node, then formatting the split portion as a block format (h1, h2, etc) formats both parts of the split content. In my mind, when content is split it should become a separate node and formatting one should not format the other.

https://user-images.githubusercontent.com/6467808/150849159-b70b2a96-0982-4d00-acbd-51de2bbe2058.mov

matthew-carroll:
> @matthew-carroll Any chance for `SuperTextField` to be updated in this PR with regards to IME support?

I'd prefer not to expand this PR beyond document editing. It's already very large. Let's handle text field stuff with another ticket and a different PR.

miguelcmedeiros:
@matthew-carroll 

It seems that the conversion from paragraph node to list item node is broken:
* Entering a "1" followed by a space no longer triggers the conversion (paragraph -> ordered list items). This is reproducible also on macOS.
* Entering a "*" followed by a space no longer triggers the conversion (paragraph -> unordered list items). This seems to work on macOS, but not on mobile.

matthew-carroll:
> @matthew-carroll
> 
> It seems that the conversion from paragraph node to list item node is broken:
> 
> * Entering a "1" followed by a space no longer triggers the conversion (paragraph -> ordered list items). This is reproducible also on macOS.
> * Entering a "*" followed by a space no longer triggers the conversion (paragraph -> unordered list items). This seems to work on macOS, but not on mobile.

If that's reproducible on `main`, please file a new ticket for that.

The "1 " situation is expected. There was an issue filed to prevent that from turning into a list item. It was recently merged in. Instead you need to use either "1. " or "1) ".

The "*" situation sounds unintentional, so that might be a bug.

matthew-carroll:
@osaxma @Jethro87 @miguelcmedeiros - Based on @osaxma's earlier comment about inserting newlines before non-text nodes, I'm wondering what the expected behavior is here.

In some editors, you can place a caret upstream or downstream near a non-text block. In my experience, these editors are typically the ones that support flowing text, e.g., Word and Docs. However, if I remember correctly, Medium does not support this. I think Medium only allows deletion of the non-text content. This leads to me wonder if editors without flowing text tend to follow that same policy.

Does anyone have any helpful references for what we should do in this situation? Adding support for upstream and downstream caret positioning for non-text nodes will require some rework of that part of the selection system for all platforms.

matthew-carroll:
@osaxma can you post a video showing repeated and consistent iOS drag handle touch area issues? I still can't seem to cause the problem on the simulator. I've tried tapping a little inside the selection and a little outside the selection. I can't seem to get the scrollview to scroll unnecessarily.

osaxma:
> can you post a video showing repeated and consistent iOS drag handle touch area issues? I still can't seem to cause the problem on the simulator. I've tried tapping a little inside the selection and a little outside the selection. I can't seem to get the scrollview to scroll unnecessarily.

- drag handle conflict with scrolling -- I expect the selection to expand here in the direction of dragging. 

https://user-images.githubusercontent.com/46427323/151340696-2aa80e94-9ede-4cd8-96cc-ef87fab5295e.mov

osaxma:
> Does anyone have any helpful references for what we should do in this situation? Adding support for upstream and downstream caret positioning for non-text nodes will require some rework of that part of the selection system for all platforms.

I personally prefer how it's done on the iOS Notes App -- I feel it's more intuitive. Here's a short recording:

https://user-images.githubusercontent.com/46427323/151342647-cdecac5e-1aa6-4641-8e9a-113403a6f72d.mp4

The behavior of selection above applies for videos too. Typing over an image/table/video selection, replaces the selection with the inserted text.

Jethro87:
@matthew-carroll I agree with @osaxma regarding upstream/downstream carat positioning in non-text nodes. I did a bunch of testing of different note-taking apps and the functionality shown in osama's Apple Notes video above felt the best to me.

matthew-carroll:
@Jethro87 @osaxma - I replaced binary selection with an upstream/downstream selection behavior.

I think all the items have been addressed. Please let me know if there are any other critical issues for mobile document editing.

osaxma:
@matthew-carroll 

Two issues that I found:

1- Typing characters before an image extends the selection over the image without inserting anything (only enter works fine)

https://user-images.githubusercontent.com/46427323/151717576-c854c6da-f816-4b01-8ddb-7ecef2329ce5.mp4


2-  It looks like the vertical conflict was solved but now there are issues when dragging the handle horizontally

https://user-images.githubusercontent.com/46427323/151717698-a721dee3-13c9-4d86-96a8-9b5b5830544e.mov

matthew-carroll:
@osaxma I added more horizontal drag area to the iOS controls. I also changed the block selection behavior to insert paragraphs before/after the block if you start typing at an edge.

matthew-carroll:
@Jethro87 @miguelcmedeiros @osaxma - are there any other features/bugs preventing you from shipping mobile document editing?

osaxma:
@matthew-carroll all looks good and works great -- thanks!

Jethro87:
@matthew-carroll I found a few more issues.

- Underline doesn't do anything (not sure if this is on purpose right now)

- Placing the carat in a word then double spacing inserts a period.

https://user-images.githubusercontent.com/6467808/152021659-d38319ed-a579-4eb5-9469-705b8c7a32f3.mov

- Typing a word incorrectly followed by pressing the spacebar inserts the autocorrected word after the incorrect word. It should replace the incorrect word.

https://user-images.githubusercontent.com/6467808/152022109-edab5867-14e0-4e2f-a2c3-6c2e2cd186d4.mov

- Highlighting a word within a line or paragraph disables the block formatting options (headers). The correct behavior should  show all formatting options (inline and block level) as tappable, but if a block format is selected, the entire block is formatted.

https://user-images.githubusercontent.com/6467808/152022578-83298ad2-c726-4c67-ba2a-582f910bc98b.mov

- When there hasn't been a selection, but the carat is simply at the beginning of a node or within a word, tapping the inline formatting options enables the functionality (as expected), but the toolbar images don't show that they are selected. Tapping on an inline formatting option should highlight the option in the toolbar and apply the formatting when typing begins.

https://user-images.githubusercontent.com/6467808/152023007-c89217a1-7978-4af4-9ae9-abff6340740c.mov

- Selecting multiple nodes and backspacing the selection results in a grey screen and the following error message for every selected node. I can replicate this by selecting a list item node and a paragraph node, as well as a header node and a paragraph node.

```
flutter: Another exception was thrown: Instance of 'ErrorSummary'
flutter: INFO: 2022-02-01 09:16:27.785405: WARNING: found component but it's not a DocumentComponent: c125e090-5670-409d-880c-0a51a61d0272, layout key: [GlobalKey#a4edf], state: null, widget: null, context: null
flutter: INFO: 2022-02-01 09:16:27.785427: Could not find any component for node position: [DocumentPosition] - node: "c125e090-5670-409d-880c-0a51a61d0272", position: (TextPosition(offset: 0, affinity: TextAffinity.downstream))
```

https://user-images.githubusercontent.com/6467808/152023120-4f89cea1-e129-4931-8725-3ee081beefc1.mov

- In a typical iOS text field, capitalization is automatically set at the beginning of a line, and >= 1 space after a period. This can be seen by the bolded shift key. Currently in super_editor, lower case is always set.

miguelcmedeiros (commented):
I managed to integrate these changes into our project and most of it works well, with one exception. 

Not sure if I'm misusing the API, but I don't seem to get the clipper passed to `createOverlayControlsClipper` to refresh, so that it can accommodate for showing/hiding the toolbar or changing its height.

miguelcmedeiros:
> Not sure if I'm misusing the API, but I don't seem to get the clipper passed to `createOverlayControlsClipper` to refresh, so that it can accommodate for showing/hiding the toolbar or changing its height.

@matthew-carroll 
I tried passing `createOverlayControlsClipper` as a value instead of a builder, and that seemed to do the trick. Is there a need to have it as a builder?

matthew-carroll:
@miguelcmedeiros the builder provides access to a `BuildContext` in case the developer needs to lookup the dimensions of some other `RenderBox` to choose clipping bounds.

Passing a `CustomClipper` instead of a builder should cause a compilation error. I'm not sure how you did that.

If the demo configuration isn't working, please provide repro steps so that I can see what's not working on my end.

matthew-carroll:
@miguelcmedeiros for the clipper, are you saying that you don't know how to re-calculate the clipper when your toolbar changes?

If so, try defining your own `CustomClipper`. In your own `CustomClipper`, take in some kind of `Listenable` that gets notified whenever your toolbar appears/disappears/changes height. Pass that `Listenable` into the `super()` constructor to cause a re-calculation whenever it notifies.

Something like:

```dart
class MyCustomClipper extends CustomClipper<Rect> {
  MyCustomClipper({
    required Listenable toolbarChangeNotifier,
  }) : super(reclip: toolbarChangeNotifier);
}
```

Then pass a builder that instantiates that clipper:

```dart
SuperEditor(
  //...
  createOverlayControlsClipper: (_) => MyCustomClipper(toolbarChangeNotifier: myToolbarWatcher),
);
```

matthew-carroll:
@Jethro87 

- Underline styling is implemented.
- Auto-insertion period problem is resolved.
- I wasn't able to repro the autocorrection issue. I think the period work may have solved that, too.
- I changed the block-level formatting options to let you change it even when multiple characters are selected.
- The mounted toolbar now highlights the active composing attributions.
- The multi-node-deletion bug is fixed
- For capitalization, I don't know what to do for the very first character, but I changed the policy to auto-capitalize for every sentence.

@miguelcmedeiros - I removed the print statements

miguelcmedeiros:
@matthew-carroll, thanks for the tip with `CustomClipper` (`reclip`) 👍 
That worked well.

miguelcmedeiros:
@matthew-carroll, while testing using the latest changes we noticed that on Android pressing and holding the backspace key when the selection is at the end of a paragraph, it "releases" the key after deleting one character.
If the selection is somewhere in the middle of the paragraph and we press and hold the backspace key, then it will continue to delete character until we lift the finger.

Jethro87:
Thanks, @matthew-carroll, the other issues look good.

Auto-capitalization of the first word in a new node, as it is now, isn't a great experience. Not only does the user have to explicitly capitalize the first word (a non-standard behavior), but adding a new node causes the keyboard to "flash", as per this video:

https://user-images.githubusercontent.com/6467808/152218105-2f8f5b23-4a5a-4807-b09d-86ce34b79ebf.mov

Can you set the keyboard to (sentence) capitalize when the carat is next to the bogus character? If so, this might be a good enough workaround, and I wonder if this would get rid of the "flash" when a user adds a new node.

matthew-carroll:
@Jethro87 - I was able to get the desired capitalization behavior by replacing the invisible bogus characters, like "~" and "^" at the front of the text, with the substring ". ". This makes the IME think that the text at the beginning of an empty paragraph is actually the start of a new sentence, so you get caps.

I recommend doing a full regression test. I can't write automated tests around this behavior because it depends upon unpredictable delta reporting from the OS.

@miguelcmedeiros - I don't think I'm gonna be able to solve that deletion problem. The issue is that whenever we send a composing `TextRange` back to the OS, on Android, if that region doesn't match what the OS expects it to be, then Flutter restarts the whole input method.

When you hold down the delete button, Android sends us a non-zero composing region. We don't care about composing regions, so we always send `TextRange.empty` back to the OS. But when we do that, Flutter panics and restarts the input method, which causes the OS to forget the deletion process it was going through.

I tried to always send the OS whatever composing region it sends us, but that resulted in occasional `TextRange` validity errors, and even when there weren't any errors, the OS would still manage to end up in an inconsistent state, triggering that input method restart.

You can see the Flutter code that forces this restart, here:
https://github.com/flutter/engine/blob/main/shell/platform/android/io/flutter/plugin/editing/TextInputPlugin.java#L448

and

https://github.com/flutter/engine/blob/main/shell/platform/android/io/flutter/plugin/editing/TextInputPlugin.java#L474

Please let me know if there are any more ship blocking issues, or if we can merge this in.

miguelcmedeiros:
> I don't think I'm gonna be able to solve that deletion problem.

@matthew-carroll, this doesn't have to block this PR, but it has to be solved before we release Android apps with SuperEditor. I'll create a ticket for this issue, so that we can tackle it separately.

miguelcmedeiros (approved):
LGTM

Didn't do an honest code review given the size/complexity of the PR, but with all the testing (demo app and superlist app), I don't see any blocker at the moment.

Jethro87:
@matthew-carroll The capitalization behavior is quite good, nice work. I did other testing and didn't find any issues, other than "Cut/Copy/Paste" still not working, but I think (?) this is intentional.

matthew-carroll:
@Jethro87 is there anything blocking you from implementing cut/copy/paste yourself?

I put that dummy popover toolbar in there just to prove that it can be placed and moved correctly. But given that Flutter still doesn't have a legitimate copy buffer, I think each app is gonna want to decide how to interpret that data. For example, when you copy the content, what format should we use? Markdown, HTML, something else?

Jethro87:
@matthew-carroll Right, that makes sense. This isn't a problem I've given thought to but I understand the issue. Will think more on it.

matthew-carroll:
Ok, I'm gonna merge this in. Fingers crossed.

Issue #376, opened 2021-12-31
requestFocus does not show caret in editor when no selection is present


jhlabs:
To reproduce:
- Create a default `SuperEditor` instance and assign an external `FocusNode`
- Trigger the `focusNode.requestFocus()`
- Observe that no caret becomes visible in the editor

When requesting the focus of the editor through e.g. keyboard shortcuts for the first time, no caret selection appears in the editor.

I will post a PR with a possible solution to set the caret position to the end of the document when no selection is present.

osaxma:
I believe there is a similar issue when requesting the focus using gestures. 

Take the following editor as an example:

```dart
import 'package:flutter/material.dart';
import 'package:super_editor/super_editor.dart';

class SimpleEditor extends StatefulWidget {
  @override
  State<SimpleEditor> createState() => _SimpleEditorState();
}

class _SimpleEditorState extends State<SimpleEditor> {
  late final document = MutableDocument(nodes: [ParagraphNode(id: '1', text: AttributedText())]);
  late final DocumentEditor editor = DocumentEditor(document: document);
  @override
  Widget build(BuildContext context) {
    return Scaffold(
      body: Padding(
        padding: const EdgeInsets.all(30.0),
        child: Container(
          decoration: BoxDecoration(border: Border.all(color: Colors.black, width: 10)),
          child: SuperEditor(
            editor: editor,
          ),
        ),
      ),
    );
  }
}
```

Which will render the following: 
<p align="center"><img width="500" alt="Screen Shot 2022-01-17 at 12 26 23 PM" src="https://user-images.githubusercontent.com/46427323/149743090-ae12f25e-f59d-4bd0-9317-6271760e09b8.png"></p>

> Note: 
> - Everything inside the black borders is SuperEditor widget.
> - The blue grey area is the `TextComponent` (color manually added for presentation purposes). 

The issue in the example above is that the gesture must be precisely placed within the blue grey area in order for the caret to show up -- which is pretty tough when the entire page is white, for instance. 

This similar to the issue because when the user touches anywhere inside SuperEditor (except the blue grey area), SuperEditor will gain focus but it'll not show where the caret is since the selection will be `null`. 


Currently I've a workaround using the following pseudo code:
```dart
  GestureDetector(
    onTap: focusNode.hasFocus 
        ? null // to avoid conflicting with SuperEditor's gestures
        : () {
            // - move the caret to the end manually using the composer
            focusNode.requestFocus();
          },
    child: SuperEditor(
[...]

PR #377, opened 2021-12-31
Moves caret to end of document on focus

jhlabs:
**Note: This is my first PR to this project, so bear with me.**

This PR solves an issue where `requestFocus` does not show a caret as detailed in #376. The `SuperTextField` solves this issue by moving the caret to the end of the text content. I used a similar solution to move the caret to the last selectable node and within the node to the last possible `NodePosition`.

If the editor composer already has a selection, this solution will maintain the current selection on focus.

### DIFF
--- a/super_editor/lib/src/default_editor/common_editor_operations.dart
+++ b/super_editor/lib/src/default_editor/common_editor_operations.dart
@@ -605,6 +605,47 @@ class CommonEditorOperations {
     return true;
   }
 
+  void _moveCaretToEndOfComponent(String nodeId, DocumentComponent component) {
+    final endOfComponent = component.getEndPosition();
+    final newPosition = DocumentPosition(
+      nodeId: nodeId,
+      nodePosition: endOfComponent,
+    );
+
+    composer.selection = DocumentSelection.collapsed(position: newPosition);
+  }
+
+  bool _selectNodeIfPossible(DocumentNode node) {
+    final component = documentLayoutResolver().getComponentByNodeId(node.id);
+
+    if (component != null && component.isVisualSelectionSupported() == true) {
+      _moveCaretToEndOfComponent(node.id, component);
+      return true;
+    } else {
+      return false;
+    }
+  }
+
+  /// Moves the caret to the last position of the last selectable [DocumentNode]
+  bool moveCaretToEnd() {
+    final lastNode = editor.document.nodes.last;
+    final isSelected = _selectNodeIfPossible(lastNode);
+
+    if (isSelected) {
+      return true;
+    } else {
+      final selectableNodeBefore = _getUpstreamSelectableNodeBefore(lastNode);
+      // There are no selectable nodes in the document if this is true
+      if (selectableNodeBefore == null) {
+        return false;
+      }
+
+      final previousIsSelected = _selectNodeIfPossible(selectableNodeBefore);
+
+      return previousIsSelected;
+    }
+  }
+
   /// Returns the first [DocumentNode] before [startingNode] whose
   /// [DocumentComponent] is visually selectable.
   DocumentNode? _getUpstreamSelectableNodeBefore(DocumentNode startingNode) {
--- a/super_editor/lib/src/default_editor/super_editor.dart
+++ b/super_editor/lib/src/default_editor/super_editor.dart
@@ -292,7 +292,9 @@ class _SuperEditorState extends State<SuperEditor> {
     // selection to the end of the document.
 
     if (_focusNode.hasFocus && !_hasFocus) {
-      _editContext.commonOps.moveCaretToEnd();
+      if (_editContext.composer.selection == null) {
+        _editContext.commonOps.moveCaretToEnd();
+      }
     }
     _hasFocus = _focusNode.hasFocus;
   }
--- a/super_editor/lib/src/default_editor/super_editor.dart
+++ b/super_editor/lib/src/default_editor/super_editor.dart
@@ -224,11 +224,14 @@ class _SuperEditorState extends State<SuperEditor> {
   late GlobalKey _docLayoutKey;
 
   late FocusNode _focusNode;
+
   late DocumentComposer _composer;
 
   DocumentPosition? _previousSelectionExtent;
 
   late EditContext _editContext;
+
+  bool _hasFocus = false;
 
   @override
   void initState() {
@@ -237,7 +240,7 @@ class _SuperEditorState extends State<SuperEditor> {
     _composer = widget.composer ?? DocumentComposer();
     _composer.addListener(_updateComposerPreferencesAtSelection);
 
-    _focusNode = widget.focusNode ?? FocusNode();
+    _focusNode = (widget.focusNode ?? FocusNode())..addListener(_onFocusChange);
 
     _docLayoutKey = widget.documentLayoutKey ?? GlobalKey();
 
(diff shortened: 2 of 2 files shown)


matthew-carroll:
I'm not sure that placing the caret at the end of the document makes sense as a solution to giving focus to a document. I can't think of any word processor, blog editor, etc. that behaves in such a manner.

I'm thinking that a more appropriate answer involves blocking focus unless a document selection is set.

Do you have any examples of other solutions to suggest that this is a behavior that the user would expect?

jhlabs:
@matthew-carroll I believe that setting focus on the document when no selection has been given yet is actually a required feature of any rich text editor when combined with other fields that the user can navigate to via keyboard shortcuts. Whether the caret should be placed at the end or beginning is debatable - it's possible to give a preference to the editor as well.

Here is one example from a popular todo app on macOS called Things:

https://user-images.githubusercontent.com/7462288/149738310-6bba3030-7b12-448f-a4a6-6c9d9b966528.mp4

You can see that I navigate with the tab key from the title of a todo to a rich text notes area. How do I achieve this without being able to focus on it with the keyboard, even if I have not a selection yet?

A similar example comes from Coda, which is arguably quite close to a word processor. It also has a title and a rich text content field that can be navigated to with tab. The cursor jumps in this case to the very beginning of the document:

https://user-images.githubusercontent.com/7462288/149738723-fa4ba41b-b718-4903-b249-b5ca32b3f683.mp4

I find that placing the cursor at the end of the rich text field is more intuitive, but as mentioned before this could also be added as an option.

I might be missing another way to do this with `SuperEditor`, but this is the only way I found to get the desired behaviour.

matthew-carroll:
@jhlabs I think you may be confusing focus with selection. Certainly, when a document has a selection, it must be focused, but what you seem to be looking for is the "selection" part rather than the "focused" part.

I haven't looked deeply into this, but my first attempt to handle the situation you brought up is that whenever the user presses `tab` on the field before the document, I would then explicitly set the selection in the editor, which should simultaneously place the caret wherever I want, and request focus.

When the user presses `tab`, I think you can execute the following to get the result you want:

```dart
composer.selection = DocumentSelection(...);
```

That `DocumentSelection` can be whatever you want. You can put it at the beginning of the document content, the end of the content, or anywhere in between. You could also choose to select the entire document. This is all based on what your particular app want to do, which is why I wouldn't want super_editor to force the same decision on everyone.

Please let me know if this approach solves your problem. If not, please let me know where it breaks down.

matthew-carroll:
@jhlabs did my last comment provide you with a path forward that solves your problem?

jhlabs:
@matthew-carroll please excuse the late reply. I understand the difference between focus and selection. 

Of course I could add a listener outside of the `super_editor` component and set a new selection. 

However, I would need to implement the logic of traversing to the last position in the last node outside of the editor functionality. Additionally, the functions needed for this depend on other private functions in the `CommonEditorOperations`. As a user of this library, I would need to manually copy those functions into my package and keep them updated when they change inside of `super_editor`. And every user of this library that wants to set a selection to the end of the document will need to do the same steps.

Wouldn't you at least agree that a better default for a selection on focusing is to select the end or start of the document instead of ...doing nothing? I have looked at plenty of rich text editors and none of them just ignore the focus event when there was no previous selection.

As a (happy) user of this library I am actually confused on why this functionality should not be implemented inside of `super_editor` - especially since I am providing a solution and the willingness to adapt it to a better API if needed.

matthew-carroll:
Hi @jhlabs, it looks like there are a few different conversation topics combined in this thread.

First, let's address the direct, technical issue. You mentioned that you need access to private behaviors to accomplish your goal.

>  I would need to implement the logic of traversing to the last position in the last node outside of the editor functionality. Additionally, the functions needed for this depend on other private functions in the CommonEditorOperations.

Perhaps, I'm missing something, but can't you accomplish your goal with the following code snippet?

```dart
composer.selection = DocumentSelection.collapsed(
  DocumentPosition(
    nodeId: document.nodes.last.id,
    nodePosition: document.nodes.last.endPosition,
  ),
);
```

I wrote that from memory, so it might not be exact, but you should be able to write something pretty close to the above code that places the caret at the end of the document. Is there more to the situation that I'm missing?

Second, about other editors...

> I have looked at plenty of rich text editors and none of them just ignore the focus event when there was no previous selection

Keep in mind, `super_editor` isn't a rich text editor. It's a toolkit for building an infinite variety of rich text editors. The `SuperEditor` widget is a convenient, drop-in configuration. Certainly, that widget contains a number of defaults. But we have to be careful with the defaults. The `SuperEditor` widget, for example, should support a blog editor without issue. If a user opens a partially written blog post, is it a sane default to scroll all the way to the end of the document just because the user pressed `tab` and gave the editor focus? I don't think that's a common behavior at all.

I'm open to default behaviors where they seem likely to satisfy the vast majority of use-cases, frustrate as few people as possible, and easily change to the behavior that you want. It's still unclear to me that this particular request meets those criteria.

This all begs the question, what's the root problem?

It should be all but trivial for each rich text editor implementation to do whatever it wants with the document selection. If it's not easy to work with selections in `super_editor`, then that indicates a deeper API problem that we should address, rather than paper over it with an accumulation of default configurations.

If this situation is a pain point, then I agree that we should look at how we might reduce the pain. It's also important to recognize that the issue that was filed, and this PR, is but one proposal for how to reduce that pain. What I'm trying to do here is better understand the root of the problem to ensure that we reduce the pain point in the most robust and scalable manner, possible.

With that said, if working with a document selection were/is easy, are there significant reasons why individual editor implementations can't, or shouldn't implement their own desired policy for initial selection?

If there are good reasons to implement a default behavior, how do we keep everyone happy? We can likely find editors that automatically insert a caret at the beginning of a document, the end of a document, or as @osaxma has demonstrated on the issue ticket, there are editors that cache the last selection position and then restore it. Do any of these represent the vast majority of use-cases? If so, how do we make it easy to also support those that don't want the default?

jhlabs:
@matthew-carroll thanks for the detailed write up.

Regarding my immediate use case: your solution solves my problem of setting the selection to the end of the document perfectly. Thanks for that. For reference for others, here is how I set the selection to the end of the document if there was none to begin with and clear the selection when the user leaves the document when focusing the editor `FocusNode`:

```dart
 widget.focusNode.addListener(() {
      if (widget.focusNode.hasFocus) {
        _composer.selection ??= DocumentSelection.collapsed(
          position: DocumentPosition(
              nodeId: _document.nodes.last.id,
              nodePosition: _document.nodes.last.endPosition),
        );
      } else {
        _composer.clearSelection();
      }
    });
```

---

@matthew-carroll I think it is actually great that `super_editor` rather wants to be a toolkit for building rich text editors. This is exactly what the Flutter ecosystem needs. In reality however the use cases for rich text editors are often related and it would be fantastic to be able to contribute solutions to this editor instead of every user inventing the wheel from scratch.

A good example in my opinion is the web editor [tiptap](https://tiptap.dev/) that is built on top of [Prosemirror](https://prosemirror.net/). Tiptap has a range of [extensions](https://tiptap.dev/extensions) that can be used to add common behaviour to the default editor experience. 

The reason tiptap exists is that it takes a big effort to understand and implement all behaviours a user wants to have in even a basic rich text editor if they were to use Prosemirror directly. Many of those behaviours are shared among apps. 

I am not sure in which direction this project wants to go, but simply wanted to point out that there might a point for making `super_editor` extensible to make it valuable to even more users and projects. I agree with you that default values for the default editor are not the solution in the long run, but a way to extend the editor with extensions or plugins that can be shared might be worth a thought.

I will close this PR since my current issue is clearly solved with a very simple solution. Thanks!

PR #379, opened 2022-01-02
SuperTextField test APIs (Resolves #375)

### LINKED ISSUE #375: Add test APIs for SuperTextField
Add APIs to enter text in a SuperTextField on any platform, and verify its content.

matthew-carroll:
This PR introduces a couple testing APIs to interact with a `SuperTextField`. They only work for desktop right now because Flutter's IME deltas don't have associated test APIs, yet.

In a test, the developer can tap on a `SuperTextField` like normal, or to control where the caret is placed in the text, the developer can call `tester.tapAtSuperTextPosition()`.

To type text in a `SuperTextField`, call `tester.enterSuperTextPlain()`.

This PR also...

* adds control over whether a caret should blink
* moves the desktop `SuperTextField` implementation into its own file so that each platform implementation has their own file
* disposes of editing controllers in the mobile text fields so that `pumpAndSettle()` will complete, as desired, when displaying a mobile `SuperTextField`.


super_editor/lib/src/infrastructure/super_textfield/android/android_textfield.dart
```diff
@@ -30,7 +30,7 @@ class SuperAndroidTextfield extends StatefulWidget {
     required this.caretColor,
     required this.selectionColor,
     required this.handlesColor,
-    this.lineHeight,
+    required this.lineHeight,
```
miguelcmedeiros, line 33: Should `SuperIOSTextField.lineHeight` also become non-null/required as well?
In the implementation of `SuperIOSTextField` it expects it to be non-null (`!` operator is used)

jamessuperlist (commented):
Works good 👍  Thanks for adding :)

jamessuperlist:
Due to https://github.com/flutter/flutter/issues/96022, '@', '#' etc. cannot be inputted. I do not think that this PR should be blocked, but it would be good to mention in README or `type_into_super_textfield_test.dart`.

super_editor/lib/src/infrastructure/super_textfield/android/android_textfield.dart
matthew-carroll: I just added `required` to the IOS version

matthew-carroll:
@miguelcmedeiros @jamessuperlist whenever you review PRs for super_editor, can you explicitly approve the PR when you're happy with it? I'm gonna go ahead and force merge this without approval so that I don't have to wait, but in the future I'd like to have approvals that are logged with GitHub.

PR #382, opened 2022-01-02
Clone metadata instead of sharing the same instance when splitting a paragraph

### LINKED ISSUE #381: Changing Text Alignment of one node is causing another independent node to change too
This video explains the issue better (notice the third node is changing whenever the alignment for the first node is modified):

https://user-images.githubusercontent.com/46427323/147884820-a7a1e885-1849-4bc9-9583-cdba13509691.mov

**Reproduction:**
I don't have a consistent way to reproduce for three nodes as in the video. But the behavior can be observed as follow:
- Run the default example on macOS.
- Split any paragraph into two by pressing enter.
- Change the alignment for the first paragraph and the second will follow.

In the video, I inserted text manually between the two paragraphs but the inserted paragraph isn't always independent as shown in the video.

osaxma:
Fixes #381:

When splitting a paragraph, create a new clone of metadata instead of sharing the same instance with the newly created paragraph nodes to prevent the behavior described in #381.

### DIFF
--- a/super_editor/lib/src/default_editor/paragraph.dart
+++ b/super_editor/lib/src/default_editor/paragraph.dart
@@ -118,11 +118,12 @@ class SplitParagraphCommand implements EditorCommand {
     node.text = startText;
 
     // Create a new node that will follow the current node. Set its text
-    // to the text that was removed from the current node.
+    // to the text that was removed from the current node. And create a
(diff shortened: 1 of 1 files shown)


matthew-carroll (approved):
LGTM

Issue #383, opened 2022-01-05
Add cursor style support for SuperTextField


salihgueler:
Right now `SuperTextField` keeps the cursor style as a private value in the `SuperTextFieldGestureInteractor` as follows

```
final _cursorStyle = ValueNotifier<MouseCursor>(SystemMouseCursors.basic);
```

This causes the `SuperTextField` to only have the basic cursor if there isn't any text. On the other hand, as a user, I would expect to have a different cursor indicator time to time. 

Therefore I would suggest to add the basic cursor style to the SuperTextField as a field with default value of `SystemMouseCursors.basic` and give users a chance to change the cursor.

matthew-carroll:
Based on the discussion in the associated PR, the root problem for this issue is unclear, and without further investigation, we shouldn't make any changes to this behavior. I'm going to move this ticket to the next milestone so that we're not holding up the mobile release.

matthew-carroll:
@miguelcmedeiros do you remember what you wanted to do with this? Did you want the text-style cursor whenever the user hovers over a `SuperTextField`, regardless of whether it's hovering over text? If so, did you think it was important to make that choice customizable, or should we make it a default and unchangeable behavior?

miguelcmedeiros:
@matthew-carroll 
I can't think of a case where the cursor should be different than the text-style when hovering the text field container. 

My proposal is to make have the cursor as text-style by default when hovering the text field container and not making it configurable.

PR #384, opened 2022-01-07
Add cursor property to the desktop super textfield.

salihgueler:
Fixes https://github.com/superlistapp/super_editor/issues/383

Right now the cursor for the default behavior of the SuperTextField is not configurable. For superlist, we would like to use custom cursors on hover to the field when there are no texts and there is no focus.

### DIFF
--- a/super_editor/lib/src/infrastructure/super_textfield/_desktop_textfield.dart
+++ b/super_editor/lib/src/infrastructure/super_textfield/_desktop_textfield.dart
@@ -56,6 +56,7 @@ class SuperTextField extends StatefulWidget {
     this.hintBehavior = HintBehavior.displayHintUntilFocus,
     this.onRightClick,
     this.keyboardHandlers = defaultTextFieldKeyboardHandlers,
+    this.cursor = SystemMouseCursors.basic,
   }) : super(key: key);
 
   final FocusNode? focusNode;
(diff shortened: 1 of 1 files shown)


matthew-carroll:
Can you accomplish the same effect by wrapping a `SuperTextField` with a `MouseRegion` and configuring your cursor style in there?

salihgueler:
> Can you accomplish the same effect by wrapping a `SuperTextField` with a `MouseRegion` and configuring your cursor style in there?

Nope, the cursor is absorbed by the SuperTextField.

matthew-carroll:
@salihgueler instead of taking in a configuration, please try setting the cursor to `MouseCursor.defer` when the user isn't over text, then try again with your own `MouseRegion` around the `SuperTextField`. Please let me know if that works.

Related docs: https://api.flutter.dev/flutter/widgets/MouseRegion/cursor.html

matthew-carroll:
@salihgueler I saw that @bk-one prioritized the ticket associated with this PR. Can you try the approach that I mentioned in my last comment and let me know if that solves the problem?

matthew-carroll:
@salihgueler can you try the "defer" approach running in a web browser and let me know if it works there? Based on some experimentation, I think this might be a Mac bug in Flutter.

matthew-carroll:
I'm going to close this PR due to inactivity. If someone wants to investigate this problem a little deeper, I welcome another attempt to solve this issue.

PR #389, opened 2022-01-12
Use LogicalKeyboardKey to detect shortcuts

jmatth:
On Windows the keyboard shortcuts don't work because `keyEvent.character` is null. This PR changes the code to match against `keyEvent.logcialKey`, which in my testing works on both Windows and MacOS. I also moved this matching logic into a public helper function to cut down on code duplication and make it easier to add custom keybindings.

### DIFF
--- a/super_editor/lib/src/infrastructure/keyboard.dart
+++ b/super_editor/lib/src/infrastructure/keyboard.dart
@@ -7,9 +7,6 @@ extension PrimaryShortcutKey on RawKeyEvent {
       (Platform.instance.isMac && isMetaPressed) || (!Platform.instance.isMac && isControlPressed);
 }
 
-bool isShortcutPressed(RawKeyEvent keyEvent, LogicalKeyboardKey key) =>
-    keyEvent.isPrimaryShortcutKeyPressed && keyEvent.logicalKey == key;
-
 /// On web, Flutter reports control character labels as
 /// the [RawKeyEvent.character], which we don't want.
 /// Until Flutter fixes the problem, this blacklist
--- a/super_editor/lib/src/infrastructure/keyboard.dart
+++ b/super_editor/lib/src/infrastructure/keyboard.dart
@@ -7,6 +7,9 @@ extension PrimaryShortcutKey on RawKeyEvent {
       (Platform.instance.isMac && isMetaPressed) || (!Platform.instance.isMac && isControlPressed);
 }
 
+bool isShortcutPressed(RawKeyEvent keyEvent, LogicalKeyboardKey key) =>
+    keyEvent.isPrimaryShortcutKeyPressed && keyEvent.logicalKey == key;
+
 /// On web, Flutter reports control character labels as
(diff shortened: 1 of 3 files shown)


super_editor/lib/src/infrastructure/keyboard.dart
```diff
@@ -7,6 +7,9 @@ extension PrimaryShortcutKey on RawKeyEvent {
       (Platform.instance.isMac && isMetaPressed) || (!Platform.instance.isMac && isControlPressed);
 }
 
+bool isShortcutPressed(RawKeyEvent keyEvent, LogicalKeyboardKey key) =>
```
matthew-carroll, line 10: I think that we'll probably regret creating this method. A "shortcut" doesn't imply that CMD or CTL is pressed. Maybe `isPrimaryShortcutKeyPressed` is poorly named, but we probably shouldn't extend the problem.

Colloquially, a "shortcut" is any key combination that maps to an application behavior. Not only is it possible that something like ALT or SHIFT is used instead of CMD/CTL, but a shortcut might involve 3+ keys. Therefore, I think this method will confuse developers in the future, and it will invite a whole bunch of slightly different methods to be defined.

I think we're better off checking the individual details that we care about in the various handlers.

matthew-carroll:
Did you also verify this behavior on web?

jmatth:
The example app is broken on web for me on `main` right now. Flutter throws an exception from inside `Platform` when it hits [this line](https://github.com/superlistapp/super_editor/blob/418d309a836232148726b06e829d75d897aa0fd2/super_editor/lib/src/default_editor/super_editor.dart#L337). I just submitted #396 to address that separately.

With that fix applied, these changes work on web with one exception: `cmd+v` is broken in Firefox. I switched back to `main` and the issue is present there too, so I filed #395.

super_editor/lib/src/infrastructure/keyboard.dart
matthew-carroll: Yeah, I'd prefer that we just remove the method. If we eventually hit a situation where this condition is particularly onerous, then we can extract a method.

matthew-carroll:
@jmatth it looks like multiple tests are failing with this PR

jmatth:
Sorry about that. I just pushed a fix.

The tests in question were setting `logicalKey` to `LogicalKeyboardKey.meta` rather than the non-modifier key that it was trying test with. This doesn't match up with what I observe in the debugger on Windows or Macos, where `logicalKey` corresponds to the non-modifier key and modifier keys are indicated via booleans like `isMetaPressed` or `isControlPressed`. I updated the broken tests to mirror this behavior. If this doesn't line up with your observations of the key event behavior please let me know.

super_editor/lib/src/infrastructure/keyboard.dart
jmatth: Cool, done in the latest commit.

matthew-carroll (approved):
LGTM. Thanks for the contribution.

PR #393, opened 2022-01-14
Added an example app demo that shows text with inline widgets in Flutter (Resolves #390)

### LINKED ISSUE #390: Spike: Investigate inline widget support
Super Editor needs to support inline widgets.

Examples:
 * Typing a pattern might convert into a small icon, e.g., `:fire:` -> :fire: - when it's an emoji, an inline widget isn't needed, but the desired image might be a small bitmap
 * A checkmark or progress indicator at the end of a line. Positioning such a visual in a `Row` won't work because if the text is multiline, we don't want the earlier rows of text to leave any space at the end of the line. 
 * Avatar images that replace a given user name, e.g., converting @SuprDeclarative to a small avatar
 * Inline code (I'm not sure if this should be solved with an inline widget or styled text with a custom painter behind it)

These use-cases involve multiple, independent concerns:
 * Rendering arbitrary content within text
 * Adjusting `AttributedText` and its associated rendering behavior to allow for non-text content
 * Selection dynamics: 
   * What happens when the user taps somewhere in the inline widget? 
   * What happens when the user drags part of the way into the inline widget? 
   * What happens when the user drags across the inline widget? 
   * What happens when the user presses `backspace` at the end of the inline widget? 
   * What happens when the user presses `delete` at the beginning of the inline widget? 
   * What happens when the user presses arrow keys into/out of the inline widget? 

This ticket is a spike to investigate the available level of control that Flutter provides for rendering inline widgets and how much access we have to the inline widgets' bounds an other properties. By the end of this ticket, we should be in a position to propose tacit answers to the questions above.

matthew-carroll:
Added an example app demo that shows text with inline widgets in Flutter (Resolves #390)

I also published a write-up for the overall spike into inline widgets: https://github.com/superlistapp/super_editor/discussions/392


super_editor/example/lib/demos/flutter_features/demo_inline_widgets.dart
```diff
@@ -0,0 +1,205 @@
...
+              child: _buildInlineCodeExample(),
+            ),
+          ],
+        ),
+      ),
+    );
+  }
+
+  Widget _buildImageExample() {
+    return RichText(
```
miguelcmedeiros, line 74: If we use `SelectableText.rich` we can even see what Flutter does with regards to selection
Screenshot 2022-01-15 at 11 33 55

matthew-carroll: I'm not sure what this means. Can you elaborate on what you're asking for?

miguelcmedeiros: Since this demo is about showing how to inline widgets with text using Flutter's APIs, I thought it would add value to make the text selectable and see how the selection looks for text and widgets.

So instead of using `RichText`, we could update the demo with `SelectableText.rich`:
``` dart
     SelectableText.rich(
       TextSpan(
         text: "Sponsored by ",
         style: const TextStyle(
           color: Colors.black,
         ),
         children: [
           WidgetSpan(
             alignment: PlaceholderAlignment.bottom,
             child: SizedBox(
               height: 22,
               child: Image.asset("assets/images/superlist_logo.png"),
             ),
           ),
         ],
       ),
       textAlign: TextAlign.start,
     );
```

miguelcmedeiros (approved):
LGTM

super_editor/example/lib/demos/flutter_features/demo_inline_widgets.dart
matthew-carroll: Sure. Updated.

PR #394, opened 2022-01-14
Pass ignoreComposerAttributions to insertCharacter

miguelcmedeiros:
`_insertCharacterInTextComposable` takes as input `ignoreComposerAttributions`, but the caller (`insertCharacter`) never passes it.

This PR updates `insertCharacter` to pass the this value.


matthew-carroll (approved):
LGTM

PR #396, opened 2022-01-17
[web] Setting gesture mode to `mouse` on web

jmatth:
The default editor throws an exception on web when it tries to access `Platform.isAndroid`, `Platform.isIOS` etc. in order to determine the gesture mode. This change makes it check for web support first to avoid using the platform getters.

Steps to reproduce:

- Navigate into `super_editor/example`
- Run `flutter run -d chrome`


osaxma:
There's already a PR open for this btw: 

- https://github.com/superlistapp/super_editor/pull/373

jmatth:
Sorry, didn't see that. I'll close this then.

PR #400, opened 2022-01-27
fix: _isContinuousAttribution now also checks for end span marker 

hongfeiyang:
fix: _isContinuousAttribution now also checks for end span marker to avoid crashing due to two spans of the same attribution within the same AttributedText

This commit is to address a problem where two intra-word styles of the same type within the same word will trigger a `Inconsistent attributions state` Exception.

Use the default editor as an example, for the first word `Lorem`, if one turns the 'bold' style on for letter `o` and then turns on the same bold style for letter `e`, this will trigger an Exception.

### DIFF
--- a/super_editor/lib/src/infrastructure/attributed_spans.dart
+++ b/super_editor/lib/src/infrastructure/attributed_spans.dart
@@ -435,17 +435,26 @@ class AttributedSpans {
     required int end,
   }) {
     _log.fine('attribution: "$attribution", range: $start -> $end');
-    SpanMarker? markerBefore = _getNearestMarkerAtOrBefore(start, attribution: attribution, type: SpanMarkerType.start);
-    _log.fine('marker before: $markerBefore');
+    SpanMarker? startMarkerBefore =
+        _getNearestMarkerAtOrBefore(start, attribution: attribution, type: SpanMarkerType.start);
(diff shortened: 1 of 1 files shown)


matthew-carroll:
@hongfeiyang thanks for your interest in the project. I think your work will be superseded by https://github.com/superlistapp/super_editor/pull/407

I'm going to close this PR based on that assumption. If you're still running into the issue after #407 merges, please file an issue ticket with exactly what a developer needs to do to reproduce the bug. In your description above, I'm not entirely sure what you mean when you say "if one turns the 'bold' style on for letter o" - do you mean the user drags a selection around the "o" character and then presses CMD+B?

hongfeiyang:
> @hongfeiyang thanks for your interest in the project. I think your work will be superseded by https://github.com/superlistapp/super_editor/pull/407
> 
> 
> 
> I'm going to close this PR based on that assumption. If you're still running into the issue after #407 merges, please file an issue ticket with exactly what a developer needs to do to reproduce the bug. In your description above, I'm not entirely sure what you mean when you say "if one turns the 'bold' style on for letter o" - do you mean the user drags a selection around the "o" character and then presses CMD+B?

The example editor has a pop up editor toolbar when a selection is made, what I mean by turn on bold style is to first select 'o' and click on 'Bold' in the tool bar to make it bold

PR #401, opened 2022-01-27
Feature/get shift characters to work in supertextfield tests

defuncart:
Presently it is not possible to enter shift characters in SuperTextField tests (as noted in https://github.com/flutter/flutter/issues/96022).

This PR facilitates entering !, @, # etc. in `tester.enterSuperTextPlain` by instead using the physical key as opposed to logical key, and overriding the character sent from key event.

Although this may seem somewhat hacky, as this is test utils which does not impact upon the widget itself, I believe it could be used as a temp fix. Moreover, it does not seem that the issue linked above is presently being worked on.

Open to feedback.


matthew-carroll (approved):
Thanks for the contribution. This looks like it's probably the least bad option for us at this point. Hopefully the Flutter team will appreciate why this stuff doesn't belong here and then build it into Flutter. Until then, we'll work with this.

PR #407, opened 2022-02-02
Optimizing AttributedSpans.collapseSpans() to run in linear time

jmatth:
The current implementation of `collapseSpans` operates by pre-generating the list of start and end points for the collapsed spans, then walking that list and repeatedly checking the list of `SpanMarker`s to determine which `Attribution`s should apply to each span. In cases with many spans this causes a large amount of duplicate work, especially in cases with overlapping spans. In my testing it takes the method almost 2 seconds to return with 1000 spans. For a less contrived example, it only needs about 50 overlapping spans to take over 16ms (time for a frame when running at 60fps).

This PR:

- Changes `collapseSpans` to run with a single linear pass of the `SpanMarker` list. It also eliminates the preprocessing step of generating the start and end locations for the collapsed spans, which wasn't significantly contributing to the runtime but fewer LoC is always a nice bonus.
- Updates the sorting logic for `SpanMarker` to always have `start` markers precede `end` markers. Without this the inner loop of `collapseSpans` needs some ugly conditionals to handle an edge case where an end marker precedes a start marker with the same offset.
- Adds a couple tests to the AttributedSpans test suite for behavior that that wasn't already covered but broke tests from other parts of the library while I was working on this.
- Adds some minor additions to `MultiAttributionSpan`: `copyWith` for convenient copying, `toString` override for better log messages.

After these changes it the method takes 3ms to return with 50 spans and 30ms with 1000. I ran the tests locally via `flutter test` and everything passed.

### DIFF
--- a/super_editor/lib/src/infrastructure/attributed_spans.dart
+++ b/super_editor/lib/src/infrastructure/attributed_spans.dart
@@ -864,7 +864,7 @@ class AttributedSpans {
       // Because we handle committing completed spans before this, we know that by this point the current marker should
       // indicate changes to the current span.
       if (marker.isStart) {
-        // Merge the new attribution into the current span.
+        // Add the new attribution to the current span.
         currentSpan.attributions.add(marker.attribution);
         _log.fine('merging ${marker.attribution}, current span is now $currentSpan.');
       } else if (marker.isEnd) {
--- a/super_editor/lib/src/infrastructure/attributed_spans.dart
+++ b/super_editor/lib/src/infrastructure/attributed_spans.dart
@@ -831,11 +831,8 @@ class AttributedSpans {
     var currentSpan = MultiAttributionSpan(attributions: {}, start: 0, end: contentLength);
 
     var i = 0;
+    var marker = _attributions[i];
     while (i < _attributions.length) {
-      var marker = _attributions[i];
-      final currentOffset = marker.offset;
-      final currentType = marker.markerType;
-
       do {
         // Consume markers as long as they are in a contiguous block with the same offset and type.
         if (marker.offset >= contentLength) {
@@ -875,8 +872,8 @@ class AttributedSpans {
         }
         marker = _attributions[i];
 
-        // If we have reached the end of this block of markers, exit the loop to commit the current span.
-      } while (marker.offset == currentOffset && marker.markerType == currentType);
+        // Keep consuming markers as long as they modify the current span.
+      } while (marker.offset <= currentSpan.start);
 
       final int currentEnd;
       if (i == _attributions.length) {
--- a/super_editor/lib/src/infrastructure/attributed_spans.dart
+++ b/super_editor/lib/src/infrastructure/attributed_spans.dart
@@ -880,8 +880,9 @@ class AttributedSpans {
     }
 
     if (collapsedSpans.last.end < contentLength - 1) {
-      // The spans committed during the loop didn't cover the entire range.  The value in currentSpan should already
-      // span from the end of the last token to the end of the requested content, so just add it to the result.
+      // The last span committed during the loop does not reach the end of the requested content range. We either ran
+      // out of markers or the remaining markers are outside the content range. In both cases the value in currentSpan
+      // should already have the correct start, end, and attributions values to cover the remaining content.
       collapsedSpans.add(currentSpan);
       _log.fine('committing last span to cover requested content length of $contentLength: ${collapsedSpans.last}');
     }
--- a/super_editor/lib/src/infrastructure/attributed_spans.dart
+++ b/super_editor/lib/src/infrastructure/attributed_spans.dart
@@ -821,6 +821,11 @@ class AttributedSpans {
       return [];
     }
 
+    if (_attributions.isEmpty || _attributions.first.offset > contentLength - 1) {
+      // There is content but no attributions that apply to it.
+      return [MultiAttributionSpan(attributions: {}, start: 0, end: contentLength - 1)];
+    }
+
     final collapsedSpans = <MultiAttributionSpan>[];
     var currentSpan = MultiAttributionSpan(attributions: {}, start: 0, end: contentLength - 1);
 
@@ -874,7 +879,7 @@ class AttributedSpans {
       }
     }
 
-    if (collapsedSpans.isEmpty || collapsedSpans.last.end < contentLength - 1) {
+    if (collapsedSpans.last.end < contentLength - 1) {
       // The spans committed during the loop didn't cover the entire range.  The value in currentSpan should already
       // span from the end of the last token to the end of the requested content, so just add it to the result.
       collapsedSpans.add(currentSpan);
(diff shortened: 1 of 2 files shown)


super_editor/lib/src/infrastructure/attributed_spans.dart
```diff
@@ -820,89 +818,60 @@ class AttributedSpans {
...
-      _log.fine('end points after change: $endPoints');
-    }
-    if (!endPoints.contains(contentLength - 1)) {
-      // This condition occurs when there are no attributions.
-      _log.fine('adding a final endpoint at end of text');
-      endPoints.add(contentLength - 1);
+      // Because we handle committing completed spans before this, we know that by this point the current marker should
+      // indicate changes to the current span.
+      if (marker.isStart) {
+        // Merge the new attribution into the current span.
```
matthew-carroll, line 858: Can you change the word "merge" in this comment and the log statement to "add" so that is clearly opposes "remove" in the else statement?

super_editor/lib/src/infrastructure/attributed_spans.dart
```diff
@@ -820,89 +818,60 @@ class AttributedSpans {
...
       }
 
-      _log.fine('start points after change: $startPoints');
-      _log.fine('end points after change: $endPoints');
-    }
-    if (!endPoints.contains(contentLength - 1)) {
-      // This condition occurs when there are no attributions.
-      _log.fine('adding a final endpoint at end of text');
-      endPoints.add(contentLength - 1);
+      // Because we handle committing completed spans before this, we know that by this point the current marker should
```
matthew-carroll, line 855: This comment is probably important, but I'm not entirely sure what it's trying to say. Can you revisit this comment and try to establish some context? The comment seems to assume that the reader already has a problem in mind, but I'm not sure what that problem is.

super_editor/lib/src/infrastructure/attributed_spans.dart
```diff
@@ -820,89 +818,60 @@ class AttributedSpans {
...
+      }
     }
 
-    if (startPoints.length != endPoints.length) {
-      _log.fine('start points: $startPoints');
-      _log.fine('end points: $endPoints');
-      throw Exception(
-          ' - mismatch between number of start points and end points. Content length: $contentLength, Start: ${startPoints.length} -> $startPoints, End: ${endPoints.length} -> $endPoints, from attributions: $_attributions');
+    if (collapsedSpans.isEmpty || collapsedSpans.last.end < contentLength - 1) {
+      // The spans committed during the loop didn't cover the entire range.  The value in currentSpan should already
```
matthew-carroll, line 869: The language in this comment mixes some times and responsibilities, making it difficult to differentiate between the description of the current situation vs the description of what needs to be done. Can you revisit this comment and try to make it clear what the situation is when coming into the if-statement, vs the work that you're about to do?

super_editor/lib/src/infrastructure/attributed_spans.dart
```diff
@@ -820,89 +818,60 @@ class AttributedSpans {
...
+        _log.fine('removing attribution ${marker.attribution}, current span is now $currentSpan.');
+      }
     }
 
-    if (startPoints.length != endPoints.length) {
-      _log.fine('start points: $startPoints');
-      _log.fine('end points: $endPoints');
-      throw Exception(
-          ' - mismatch between number of start points and end points. Content length: $contentLength, Start: ${startPoints.length} -> $startPoints, End: ${endPoints.length} -> $endPoints, from attributions: $_attributions');
+    if (collapsedSpans.isEmpty || collapsedSpans.last.end < contentLength - 1) {
```
matthew-carroll, line 868: Is it possible to check for the empty situation near the beginning of the method, similar to content length? Or is that calculation too difficult to do on its own?

matthew-carroll (changes requested):
Thanks for the improvements. I added a few requests for minor changes.

super_editor/lib/src/infrastructure/attributed_spans.dart
```diff
@@ -820,89 +818,60 @@ class AttributedSpans {
...
-          endPoints.add(marker.offset - 1);
-        }
-      }
-      if (marker.isEnd) {
-        // Add an `end` point.
-        if (!endPoints.contains(marker.offset)) {
-          _log.fine('adding an end point at: ${marker.offset}');
-          endPoints.add(marker.offset);
-        }
+        // If we encountered a start token, end the current span 1 index earlier to simulate the missing end marker.
```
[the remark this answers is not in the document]
jmatth: Yeah, this is probably a quirk of the development process working its way into the comments. In an early implementation I was preprocessing the list to insert start and end markers at the points where spans overlapped. I got rid of that because 1) it's actually unnecessary and 2) it's actually harder to read and reason about. Since those explicit start/end markers aren't there, this on the fly calculation is now "simulating" them.

I agree this wording doesn't make sense now without that particular thought process. I'll update it similar to your suggestion on the end token.

super_editor/lib/src/infrastructure/attributed_spans.dart
jmatth: The comment is there to say that because the way the code is ordered, by the time we reach this point we know that we can just apply changes to the current span based on the current marker without any other checks. A previous version of the code required a bounds check on the current marker's offset, and this comment came about when that check was made unnecessary and removed.

How about: `By the time we get here, we are guaranteed that the current marker should modify the current span. Apply changes based on the type of the marker.`?

super_editor/lib/src/infrastructure/attributed_spans.dart
```diff
@@ -820,89 +818,60 @@ class AttributedSpans {
...
-        // `end` marker.
-        if (marker.offset < contentLength - 1 && !startPoints.contains(marker.offset + 1)) {
-          _log.fine('jumping forward one to add a start point at: ${marker.offset + 1}');
-          startPoints.add(marker.offset + 1);
-        }
+        // Commit the completed span.
+        collapsedSpans.add(currentSpan.copyWith(end: currentEnd));
+        _log.fine('committed span ${collapsedSpans.last}');
+
+        // If we encountered an end token, start the next span 1 index later to simulate the missing start marker.
```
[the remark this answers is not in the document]
jmatth: Addressed in [this comment](https://github.com/superlistapp/super_editor/pull/407#discussion_r798726573).

super_editor/lib/src/infrastructure/attributed_spans.dart
matthew-carroll: Sure, that comment makes sense

super_editor/lib/src/infrastructure/attributed_spans.dart
matthew-carroll: I think I'd prefer this approach. This way readers know that if they get past the guard statements, they're definitely doing some span calculations.

jmatth: Cool, taken care of in b3c9c1ee02162dd599d82821a568c042a7490af6.

super_editor/lib/src/infrastructure/attributed_spans.dart
jmatth: Taken care of in f7a40f556656ae15bb4964cbb1fe7afb2f8cc931.

jmatth:
I think I've addressed all the outstanding PR comments.

matthew-carroll (approved):
LGTM. Thanks.

PR #409, opened 2022-02-03
Don't insert characters when CMD/CNTL pressed (Resolves #405)

### LINKED ISSUE #405: Catch all CMD/CTRL keyboard sequences, unless they have a specific behavior associated
Steps to reproduce:
- Open example app
- press CMD+F (should eventually support search, no functionality associated yet)
- the caracter "f" is written

Expected behavior:
- CMD+F should not add a character, it is an "unregistered" CMD sequence - should result in a no-operation
- All other CMD (CTRL on Win) sequences should also not result in characters but should either perform the associated action (e.g. CMD+V) or be a no-operation

matthew-carroll:
Don't insert characters when CMD/CNTL pressed (Resolves #405)


miguelcmedeiros (approved):
LGTM

bk-one (approved):

Issue #411, opened 2022-02-03
Expose `autofocus` option on SuperEditor


osaxma:
## Issue
When using `SuperEditor` with other widgets, SuperEditor is either claiming focus automatically (keyboard mode) or it does not (IME mode). 

Since SuperEditor can be with other widgets within a given view, the option for autofocus should be left for the user. 

## Additional Context:
Currently the autofocus is not consistent between `IME` and `Keyboard`:

- For Keyboard, the default is `true`:
https://github.com/superlistapp/super_editor/blob/47e05fcfbc4f9e5e7a21357ef60dcb0ce48ee814/super_editor/lib/src/default_editor/document_input_keyboard.dart#L85-L92

- For IME, the default is `false`:
https://github.com/superlistapp/super_editor/blob/47e05fcfbc4f9e5e7a21357ef60dcb0ce48ee814/super_editor/lib/src/default_editor/document_input_ime.dart#L316-L323

matthew-carroll:
I took a look at other Flutter widgets including `TextField`, `Checkbox`, `Radio`, `Slider`, and `EditableText`. They all take an `autofocus` property, so this seems like a reasonable expectation for `SuperEditor`.

Issue #410, opened 2022-02-03
`ArgumentError`  in Mobile when starting with an empty Document


osaxma:
## To reproduce:

- Take the following sample app:

```dart
class SampleApp extends StatelessWidget {
  const SampleApp({Key? key}) : super(key: key);
  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      home: Scaffold(
        body: Container(
          color: Colors.blueGrey.shade300,
          margin: const EdgeInsets.all(100.0),
          child: SuperEditor(
            editor: DocumentEditor(
              document: MutableDocument(
                nodes: [
                  // uncomment and it'll work fine
                  // ParagraphNode(id: '0', text: AttributedText(text: 'test'))
                ],
              ),
            ),
            // same issue applies to: DocumentGestureMode.android
            gestureMode: DocumentGestureMode.iOS, 
            inputSource: DocumentInputSource.ime,
          ),
        ),
      ),
    );
  }
}
```

- Tap on the grey area 
- This error should show up: `ArgumentError (Invalid argument(s): 1.0)`

The error points to the following (when I placed a debug point there, `width` was equal to `0`):
https://github.com/superlistapp/super_editor/blob/47e05fcfbc4f9e5e7a21357ef60dcb0ce48ee814/super_editor/lib/src/default_editor/layout.dart#L104

## Additional context:
- The error is reproducible using either the simulator or macOS using the latest main. 

Note:

I don't really have a use case for this. I just observed the error while working on the autofocus PR.

osaxma:
Here's a simple test to reproduce as well taken from the `autofocus` PR:

```dart
  testWidgets('Reproduce Argument error', (tester) async {
    await tester.pumpWidget(
      MaterialApp(
        home: Scaffold(
          body: SuperEditor(
            editor: DocumentEditor(document: MutableDocument(nodes: [])),
            gestureMode: DocumentGestureMode.iOS, // or .android
          ),
        ),
      ),
    );
    await tester.tap(find.byType(SuperEditor));
    await tester.pumpAndSettle();
  });
```

PR #412, opened 2022-02-03
Make `autofocus` an optional parameter

osaxma:
closes: #411

In short, this PR achieve the following:
- Exposes `autofocus` property on `SuperEditor` constructors and pass it to `Focus` widget for both Keyboard and IME. 
-  <s>The default value is set to `false` similar to the default value for `Focus` widget.</s>
- The default value is set to  `true`
  - setting the default value to `false` breaks quite a bunch of tests. 
  - note: before this PR -- `autofocus` was `true` for Keyboard input  but `false` for IME input.
- added tests

## Note (edited):
Several added tests were failing before because I had a document with empty nodes (failed due to:  https://github.com/superlistapp/super_editor/issues/410). So I added one of the sample documents instead (i.e. `singleBlocDoc`) to make the tests pass.


super_editor/test/src/default_editor/super_editor_test.dart
```diff
@@ -0,0 +1,97 @@
...
+        await tester.pumpAndSettle();
+
+        expect(focusNode.hasFocus, true);
+      }, variant: inputAndGestureVariants);
+    });
+  });
+}
+
+
+class InputGestureTuple {
```
matthew-carroll, line 76: Can you make this and `inputAndGestureVariants` private so that no one inadvertently imports it into other test files?

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -84,6 +84,7 @@ class SuperEditor extends StatefulWidget {
     this.composer,
     this.componentVerticalSpacing = 16,
     this.showDebugPaint = false,
+    this.autofocus = true,
```
matthew-carroll, line 87: We'll need each of these to be `false`. If they break other tests then we need to fix those tests.

matthew-carroll (changes requested):

super_editor/lib/src/default_editor/super_editor.dart
osaxma: done -- fixing them was easier than I thought since I only had to add `autofocus = true` for the different SuperEditor builders that are used for tests.

matthew-carroll (approved):
LGTM. Thanks.

PR #418, opened 2022-02-04
Fixing docstrings on ExecutionInstruction enum members

jmatth:
Hopefully I'm not completely misreading the code, but it looks like the docstrings for `continueExecution` and `haltExecution` were swapped.


matthew-carroll (approved):
LGTM. Thanks for the fix.

Issue #386, opened 2022-01-07
Selection highlight when the node has no content is not supported for all text components


miguelcmedeiros:
Currently it is possible to control whether a thin selection highlight when the node has not content, but only by setting it directly in `TextComponent`, `TextWithHinComponent` or `SuperSelectableText`. `OrderedListItemComponent`, `UnorderedListItemComponent` and `BlockquoteComponent` currently don't support this preference.

Aside from being able to set this preference per component, it would be great to set this preference for the whole document as well.

matthew-carroll:
@miguelcmedeiros do you know if this is fixed now, or is it still a problem?

miguelcmedeiros:
@matthew-carroll 
Just checked and some components were missing some wiring to make it work. 
Opened a PR to fix it.

PR #428, opened 2022-02-14
Full-width components, stylesheets, and layout presenter and view models (Resolves #414)

### LINKED ISSUE #414: Add support for full-width components in document layout
It should be possible to display, for example, a header image that stretches the full width of the document.

Adjust the default document layout rules to allow for this use-case.

Then, upgrade the demo editor with a full-width, and more interesting header image.

matthew-carroll:
I set out to implement full-width images and found myself refactoring a lot of the layout presentation system.

Here's a write up that explains what's happening in this PR:
https://github.com/superlistapp/super_editor/discussions/427

At the time of posting this, there are still some open questions that I need to deal with, but I'm posting in the current state to start getting eyes on what I'm changing.


matthew-carroll:
CC @osaxma - in case you have any thoughts about these changes

super_editor/example/lib/demos/components/demo_text_with_hint.dart
```diff
@@ -70,11 +70,13 @@ class _TextWithHintDemoState extends State<TextWithHintDemo> {
...
+      stylesheet: defaultDocumentStylesheet.copyWith(
+        margin: const EdgeInsets.symmetric(vertical: 56, horizontal: 24),
 
-      /// Adjust the default styles to style 3 levels of headers
-      /// with large font sizes.
-      textStyleBuilder: _textStyleBuilder,
+        /// Adjust the default styles to style 3 levels of headers
+        /// with large font sizes.
+        inlineTextStyler: (attributions, style) => _textStyleBuilder(attributions),
+      ),
```
roughike, line 79: Would it make sense to have default values from `defaultDocumentStyleheet` inlined as default arguments in the constructor?

Something like:

```dart
class SingleColumnLayoutStyleSheet {
  const SingleColumnLayoutStyleSheet({
    this.standardContentWidth = 640.0,
    this.margin = EdgeInsets.only(bottom: 96),
    this.inlineTextStyler = defaultInlineTextStyler,
    this.blockStyles = DocumentBlockStyles(),
  });
}
```

(same thing would be done for `DocumentBlockStyles`)

If you'd only want to control the margin, but preserve all the other default styles, it would look like this:

```dart
@override
  return SuperEditor(
    editor: _docEditor,
    stylesheet: SingleColumnLayoutStyleSheet(
      // We want to customize the margin but everything else stays at their
      // default values.
      margin: const EdgeInsets.symmetric(vertical: 56, horizontal: 24),
    ),
  );
}
```

This way, there wouldn't be a top-level `defaultDocumentStylesheet` variable and no need to do `copyWith()` on it to change a few things while preserving the default stylesheet for everything else.

I think this could make the API a bit nicer, since there wouldn't be a need to discover that there's a `defaultDocumentStylesheet` and that you have to do a `.copyWith()` on it.

roughike: Also, this stylesheet stuff looks very similar to `Theme` when using Flutter with the Material library. Could it be named to `SomethingSomethingTheme` instead? Or are document stylesheets conceptually different from themes?

The reason I'm asking because if it is indeed similar to a theme, a concept that most Flutter developers are familiar with, it would make sense for me to call it a theme. The "stylesheet" name is not hard to understand at all, but makes me wonder _"how is this different from a theme?"_. :-)

roughike (commented):
The `ViewModel` and `Presenter` names might be, for the lack of a better word, confusing.

It looks like `ViewModel` is not supposed to be the VM in [MVVM](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93viewmodel). Likewise, it looks like `Presenter` is not supposed to be the Presenter from [MVP](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93presenter).

I'm wondering if there are alternative names that don't clash with these ones.

super_editor/lib/src/default_editor/layout_single_column/_list_items.dart
```diff
@@ -0,0 +1,144 @@
...
+    styleBuilder: componentMetadata.textStyleBuilder,
+    indent: componentMetadata.indent,
+    textSelection: componentMetadata.selection,
+    selectionColor: componentMetadata.selectionColor,
+    showCaret: componentMetadata.caret != null,
+    caretColor: componentMetadata.caretColor,
+  );
+}
+
+Widget? newOrderedListItemBuilder(
```
miguelcmedeiros, line 29: Should this be `orderedListItemComponentBuilder`?

super_editor/lib/src/default_editor/layout_single_column/_paragraph.dart
```diff
@@ -0,0 +1,140 @@
...
+            'blockType': componentViewModel.blockType,
+          }
+        : {},
+    textAlign: componentViewModel.textAlignment,
+    textDirection: componentViewModel.textDirection,
+    textSelection: componentViewModel.selection,
+    selectionColor: componentViewModel.selectionColor,
+    showCaret: componentViewModel.caret != null,
+    caretColor: componentViewModel.caretColor,
+    highlightWhenEmpty: componentViewModel.highlightWhenEmpty,
```
miguelcmedeiros, line 44: Since we are changing quite a few things around in this PR, let me bring another one up (and feel free to ignore 😅), but it would be great if this property is also supported for other node type like list items and blockquotes.

Aside from controlling it at component level, it would be good to have an editor-default value.

I opened #386 some time ago about this.

miguelcmedeiros (commented):
Tested it out using the super_editor demos (would be nice to test it together with our app, but I can't make it at the moment) and the only thing I noticed was the color of the paragraph in the Text With Hint demo is white in a light mode. Will continue to test

super_editor/example/lib/demos/components/demo_text_with_hint.dart
matthew-carroll: > Would it make sense to have default values from defaultDocumentStyleheet inlined as default arguments in the constructor?

I'm of two minds about this. Technically, yes we can do that, and I considered it. But I'm wondering if there's any stylesheet that's likely to work well as a true default. We have the `defaultDocumentStylesheet` which is an easy option for someone who doesn't care about styles at all, but if a developer styles all the headers, is that developer likely to retain the same styles for paragraph and blockquote? If developers are likely to re-style pretty much everything, then maybe it's easier to require everything and then force developers to explicitly choose to use existing sub-structures:

```dart
myStylesheet = SingleColumnLayoutStylesheet {
  margin: // new margin
  standardContentWidth: // new width
  // Explicitly use default block styles
  blockStyles: defaultBlockStyles,
  // Explicitly use default inline text styles
  inlineTextStyler: defaultInlineTextStyler,
}
```

I could go either way with it.

> Also, this stylesheet stuff looks very similar to Theme when using Flutter with the Material library. Could it be named to SomethingSomethingTheme instead? Or are document stylesheets conceptually different from themes?

WRT themes - first, it's unclear what constitutes a theme in Flutter. Flutter offers two themes and both of them are nearly identical in applicability. So what should the concept of a "theme" include? Up to this point, a theme seems to be a set of visual details applied to an arbitrary collection of widgets.

Throughout the history of print documents, stylesheets have been used for layout and styles. When the web came along and introduced browser-based documents, we got Cascading Stylesheets (CSS). 

Is super_editor a collection of widgets like Material and Cupertino? Or is super_editor a tool for editing and styling documents? And as the APIs evolve, are we likely to end up with something more like a `ThemeData` structure, or are we more likely to end up with a loose set of rules that apply conditionally, like CSS?

matthew-carroll:
> The `ViewModel` and `Presenter` names might be, for the lack of a better word, confusing.
> 
> It looks like `ViewModel` is not supposed to be the VM in [MVVM](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93viewmodel). Likewise, it looks like `Presenter` is not supposed to be the Presenter from [MVP](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93presenter).
> 
> I'm wondering if there are alternative names that don't clash with these ones.

I've used the concept of Presenters and View Models in this way for many years and I've been very happy with the results. Frankly, I don't think anybody knows what they're talking about with those acronyms. Every time someone implements them, they look completely different. If you let go of acronyms and embrace the roles, I think it makes a lot of sense.

View Model: a model of a view. A data structure that contains all the information that a widget needs to render itself.

Presenter: connects the business domain to the UI. A Presenter interprets business information and uses that information to construct View Models, which are then rendered by the UI.

The business domain in this package are the logical document constructs: `Document`, `DocumentNode`, `DocumentSelection`, `DocumentPosition`, `ParagraphNode`, `AttributedText`, etc.

The Presenter in this PR interprets these business artifacts to generate a View Model for a single-column document layout. The View Model doesn't know how to paint anything, but it contains all the information that the layout widget needs to do the painting.

super_editor/lib/src/default_editor/layout_single_column/_paragraph.dart
matthew-carroll: Yeah, I'm not sure that a type-safe structure is the way to go with this. I'm still thinking about it, but I'm not sure how to facilitate arbitrary content with this approach. I'm also guessing that at some point we're going to want something akin to CSS selectors, which I don't think would work with this approach, either. We probably need something that is more loosely defined.

matthew-carroll:
@roughike @miguelcmedeiros - what do you guys think about using something similar to CSS? Is that overkill for super_editor? Are there any obvious alternatives that would effectively support a growing collection of node types and style properties?

osaxma:
> CC @osaxma - in case you have any thoughts about these changes

I honestly haven't had enough time to experiment with the new API but my only concern is about manipulating<sup>1</sup> and storing<sup>2</sup> the stylesheet data.

1. Manipulating the stylesheet data

    When a node is deleted, the data associated with that node needs to be deleted from stylesheet as well. Also, replacing the a node while preserving the style is another concern. For instance, if an "image" replacement causes the generation of a new `nodeID`, then the old `nodeID` needs to be replaced with the new one to preserve the style. 

2. Storing the stylesheet data

    I currently have a `toJson`/`fromJson` on `Document` as well as on every `DocumentNode` that the app uses.  When `Document.toJson` is invoked, each `DocumentNode.toJson` is also invoked recursively to create a json representation of the `Document` for storage purposes. With the new API, I'm not sure how easy it would be to store the stylesheet data in order to preserve any style preferences. 

These were the only two thoughts that came two mind so far just by looking at the new API.

matthew-carroll:
@osaxma I think those points make sense for the per-component styles. But I don't think they apply to the stylesheet, do they? The stylesheet is a data structure, so it can be serialized pretty easily. Also the stylesheet is a collection of rules, rather than styles tied to a specific node.

osaxma:
@matthew-carroll 

My bad -- I meant `componentStyles` not `styleSheet`. 

> Also the stylesheet is a collection of rules, rather than styles tied to a specific node.

Again, I mixed up because I was looking at `SingleColumnCustomComponentStyles` where widths and paddings are mapped to specific `nodeID`s and confused them with the stylesheet.

bk-one:
Just to leave my 2 cents. As I can't judge the technical implications of one way or another, from a pure UX standpoint, all I would want is to define a clickable area that will trigger the editor into input mode, even if the actual editable content is much smaller. Comparable with HTML `textarea` where there's only one line of content. I should be able to click anywhere in the textarea, even if just the first line of the textarea has any content.

matthew-carroll:
@miguelcmedeiros @roughike @osaxma - I took another stab at the styling solution. I've switched from a stylesheet data structure to a stylesheet with priority rules, similar to CSS. Can each of you take another look?

The new CSS-inspired stylesheet includes selectors, which now make it possible to select a node based on the node before it, after it, or both. As a result, you'll see that the `defaultStylesheet` applies different padding to paragraphs depending on whether they come after a header, or not.

@miguelcmedeiros - I noticed that the "highlight when empty" behavior disappeared from the editor, so I got it working again. You'll notice that I added a "task" node to the example editor to show how devs can add a new node and component. The task component also highlights when empty. Please let me know if this solves the issue that you mentioned.

@roughike - You asked in another PR about notifying change listeners when node metadata changes. I think this PR solves that problem by creating explicit "put" methods for metadata, and notifying listeners from within the nodes. Please let me know if the issue remains.

matthew-carroll:
Also @miguelcmedeiros @roughike @osaxma if any of you have a Windows or Linux machine handy, can you see if you can find out why a single test is failing on those platforms, but not on Mac?

miguelcmedeiros (approved):
LGTM

matthew-carroll:
The Windows/Linux bug was the result of `renderParagraph` returning a caret offset with a `-0.0` instead of a regular `0.0`. For some reason, the negative zero also failed the bounds check against a `Size`.

I filed a Flutter issue here: https://github.com/flutter/flutter/issues/100033

I also added a check in the code to a value near zero and then replaced it with a regular `0.0`. The problem should be solved for this PR.

PR #434, opened 2022-02-17
Bugfix: caret no longer gets stuck in middle of emojis when moving left/right (Resolves #391)

matthew-carroll:
Bugfix: caret no longer gets stuck in middle of emojis when moving left/right (Resolves #391)

#391 said errors were being thrown. I wasn't able to reproduce any errors by selecting emojis on desktop in the example app. However, the caret was getting stuck in the middle of emojis because we were moving by code point rather than by character. This PR adds machinery to all `String`s to move by character, and updates the keyboard handler to use the new behavior.

@miguelcmedeiros please let me know if you're still seeing any errors or related problems with emojis.

PR #435, opened 2022-02-17
Added list item and HR pattern matching to IME input (Resolves #422)

### LINKED ISSUE #422: Converting paragraph by pattern matching is not working on mobile
On desktop if the user type any of the following patterns followed by a space in the paragraph node, then SuperEditor will convert the node type:
* `*` (unordered list)
* `1.` (ordered list)
* `---` (horizontal rule)

On mobile, the same patterns do not trigger the node conversion.

matthew-carroll:
Added list item and HR pattern matching to IME input (Resolves #422)


matthew-carroll:
The failing test should be fixed when PR #434 is merged, so it's unrelated to this PR.

miguelcmedeiros (approved):
LGTM

PR #440, opened 2022-02-18
Converted SuperTextField into a platform adaptive widget (Resolves #413)

### LINKED ISSUE #413: Merge all super text fields into cross-platform SuperTextField widget
Clients would like to use super textfields interchangeably across platforms.

Use the `SuperTextField` name as an umbrella, which then delegates to the appropriate platform configuration, inside.

matthew-carroll:
Converted SuperTextField into a platform adaptive widget (Resolves #413)

@miguelcmedeiros can you give this PR a try in the Superlist app and let me know if the new adaptive `SuperTextField` widget satisfies your x-platform text field needs?


matthew-carroll:
@venkatd - you might be interested in this PR because it impacts the text field widget APIs

super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart
```diff
@@ -1,3 +1,244 @@
...
+          ? ImeAttributedTextEditingController(controller: widget.textController, disposeClientController: false)
+          : ImeAttributedTextEditingController();
+    }
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    switch (_configuration) {
+      case SuperTextFieldPlatformConfiguration.desktop:
+        return SuperDesktopTextField(
```
miguelcmedeiros, line 157: Currently it does not handle the desktop specific fields like `keyboardHandlers` or `hintBehavior`.

miguelcmedeiros (commented):
@matthew-carroll, we can't use it in it's current shape since we are passing `keyboardHandlers` or `hintBehavior`.

I also noticed that the desktop, iOS and Android text field widgets have new fields, but they are not used.

matthew-carroll:
@miguelcmedeiros 

> I also noticed that the desktop, iOS and Android text field widgets have new fields, but they are not used.

Can you elaborate on this? I'm not sure what you mean.

matthew-carroll:
@miguelcmedeiros 

> we can't use it in it's current shape since we are passing keyboardHandlers or hintBehavior

I unified the hint APIs across all the text fields, and exposed `hintBehavior` and `keyboardHandlers` in `SuperTextField`.

> This is a tricky one since we have a text style builder and the dev's won't really know which line height to pass here.

I changed the requirements so that you don't need to pass `lineHeight` for single-line fields, but you're going to need to provide it multi-line fields. Please see the demo app examples.

Can you take another pass and see if this now works for your app?

super_editor/lib/src/infrastructure/super_textfield/android/android_textfield.dart
```diff
@@ -44,16 +41,20 @@ class SuperAndroidTextfield extends StatefulWidget {
   /// this text field.
   final ImeAttributedTextEditingController? textController;
 
+  /// The alignment to use for text in this text field.
+  final TextAlign textAlign;
```
miguelcmedeiros, line 45: I couldn't find where this field is used in the widget.

super_editor/lib/src/infrastructure/super_textfield/ios/ios_textfield.dart
```diff
@@ -46,10 +54,21 @@ class SuperIOSTextField extends StatefulWidget {
   /// this text field.
   final ImeAttributedTextEditingController? textController;
 
+  /// The alignment to use for text in this text field.
+  final TextAlign textAlign;
```
miguelcmedeiros, line 58: I couldn't find where this field is used in the widget.

super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart
```diff
@@ -1,3 +1,259 @@
...
+    this.textController,
+    this.textAlign = TextAlign.left,
+    this.textStyleBuilder = defaultStyleBuilder,
+    this.hintBehavior = HintBehavior.displayHintUntilFocus,
+    this.hintBuilder,
+    this.controlsColor,
+    this.selectionColor,
+    this.minLines,
+    this.maxLines = 1,
+    required this.lineHeight,
```
miguelcmedeiros, line 61: From [your comment](https://github.com/superlistapp/super_editor/pull/440#issuecomment-1045293446) I thought it wouldn't required anymore.

osaxma:
I believe `autofocus` should be exposed on `SuperTextField` in a similar manner to `SuperEditor`.

miguelcmedeiros:
Tried it out in our app and noticed the following behaviour which is reproduced using the demo app:
ezgif com-gif-maker-7

In case the text of the `SuperTextField` fits in the given viewport and when moving the selection handles close to the edge of the viewport, then there is a slight motion of the content.

miguelcmedeiros:
@matthew-carroll 
It seems that the keyboard appearance (light and dark modes) is not according to the OS settings:
ezgif com-gif-maker-8

matthew-carroll:
I'll address the couple of API issues that @miguelcmedeiros mentioned.

For the rest of what you guys find, can you file individual tickets for those things? Whatever the scrolling problem or keyboard theming problem, those must be about 6 months old. I'd like to focus this review on the goal of unifying the existing text fields so that developers don't have to worry about which platform they're on, if they don't want to.

matthew-carroll:
@miguelcmedeiros `textAlign` is now used, and `lineHeight` is only required when `maxLines` is > 1

venkatd:
@matthew-carroll looks great to me!

This may allow us to remove some workarounds we've had in place to use a SuperTextField on desktop and the built-in TextField on mobile devices.

How is `defaultTextFieldKeyboardHandlers` used? I guess this is only something that is taken into consideration on platforms that have actual keyboard handling?

miguelcmedeiros (approved):
LGTM

matthew-carroll:
@venkatd 
> How is `defaultTextFieldKeyboardHandlers` used? I guess this is only something that is taken into consideration on platforms that have actual keyboard handling?

Yeah, the only way I can think to unify the API is to take mobile and desktop pieces together. Otherwise, you must know what platform you're on, in which case you wouldn't need a unified API.

I'm gonna merge this in. Please let me know ASAP if you think we should make further adjustments to the unified API.

venkatd:
@matthew-carroll all LGTM! If `defaultTextFieldKeyboardHandlers` is handled just as before, the API looks good to me 👍

PR #443, opened 2022-02-22
Migrate website to null safety and fix deployment

roughike:
In deploy previews, the `main.dart.js` can't be loaded. The base href is `/SuperEditor/`, but in the deploy preview context, it should be `/`. 

We can leave it like this for now - once merged to the `main` branch, stuff will work just fine (but not for deploy previews).


website/lib/homepage/editor_toolbar.dart
```diff
@@ -173,14 +179,22 @@ class _EditorToolbarState extends State<EditorToolbar> {
...
+      final existingNode = widget.editor.document.getNodeById(widget.composer.selection!.extent.nodeId)!;
       (existingNode as ParagraphNode).metadata['blockType'] = _getBlockTypeAttribution(newType);
+
+      // Merely changing the blockType of the ParagraphNode does not trigger any of the document listeners.
+      //
+      // As such, we have to manually tell the node that something changed - otherwise, the block type of this
+      // ParagraphNode would change only if something else in the document changed. This is a bit hacky.
+      //
+      // ignore: invalid_use_of_protected_member, invalid_use_of_visible_for_testing_member
+      existingNode.notifyListeners();
```
roughike, line 197: @matthew-carroll not related to this PR, but in SuperEditor in general, should there be a different API for modifying metadata that automatically notifies listeners? the current way feels a bit hacky.

roughike:
@matthew-carroll this is now ready for review.

One thing I wasn't able to figure out is this overflow exception when the window size is small enough:

image

website/lib/homepage/editor_toolbar.dart
matthew-carroll: Using `putMetadata()` should notify listeners. I'll still approve like this, so we can move forward, but I think this problem is solved with the right adjustment.

matthew-carroll (approved):
LGTM, thanks.

PR #445, opened 2022-02-24
Handle empty documents (Resolves #274, Resolves #410)

### LINKED ISSUE #274: Can't select empty space to start editing
In many cases, the editor will be in a larger text box, like this:
image
When the user tries to select an area in the text box, the cursor does not appear and they can't start editing. They have to actually select text currently in the box. This behaviour is also on the SuperEditor website (https://superlist.com/SuperEditor/), so this isn't just an issue with my implementation.

### LINKED ISSUE #410: `ArgumentError`  in Mobile when starting with an empty Document

matthew-carroll:
Handle empty documents (Resolves #274, Resolves #410)

Fixes a bug in empty documents where a `clamp()` method was running with a min > max, which caused an illegal argument error.

Now resolves document taps by finding the "nearest" document position, rather than the "exact" document position. This means that tapping anywhere within `SuperEditor` will place the caret, which is convenient for empty documents.

### DIFF
--- a/super_editor/lib/src/default_editor/document_gestures_mouse.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_mouse.dart
@@ -301,7 +301,7 @@ class _DocumentMouseInteractorState extends State<DocumentMouseInteractor> with 
     editorGesturesLog.info("Tap down on document");
     final docOffset = _getDocOffset(details.localPosition);
     editorGesturesLog.fine(" - document offset: $docOffset");
-    final docPosition = _docLayout.getDocumentPositionAtOffset(docOffset);
+    final docPosition = _docLayout.getDocumentPositionNearestToOffset(docOffset);
     editorGesturesLog.fine(" - tapped document position: $docPosition");
 
     _focusNode.requestFocus();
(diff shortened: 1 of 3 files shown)


miguelcmedeiros (approved):
Tested it with Superlist and works well 👍 

LGTM

PR #447, opened 2022-02-24
Bugfix: mobile handles mis-aligned on desktop (Resolves #446)

### LINKED ISSUE #446: Mobile handles are mis-aligned on desktop
Run the example editor with Android or iOS gestures, instead of mouse gestures. Notice that the handles are all mis-aligned. They're left-aligned with the screen, instead of aligned with the document.

matthew-carroll:
Bugfix: mobile handles mis-aligned on desktop (Resolves #446)

The problem was that the document layer link was applied on the outside of a `Center` widget, which caused the layer link to wrap all available space, instead of just the document. I switched the order of the two widgets so that the layer link applies to the exact bounds of the document.


miguelcmedeiros (approved):
LGMT

PR #449, opened 2022-02-25
Updated demo app to automatically adapt to the underlying platform (Resolves #436)

### LINKED ISSUE #436: Update example app to auto configure to current platform
Currently, the example app has different demos for different input methods/styles: physical keyboard, Android style IME, iOS style IME.

Update the example app to auto-configure the main demo by the host platform, and offer options in that demo to switch to other input modes.

matthew-carroll:
Updated demo app to automatically adapt to the underlying platform (Resolves #436)

Also, I moved cut, copy, and paste from the keyboard handlers and put them into the `CommonEditorOperations`, and then I used those operations in the demo apps to perform those operations.

### DIFF
--- a/super_editor/example/lib/demos/example_editor/example_editor.dart
+++ b/super_editor/example/lib/demos/example_editor/example_editor.dart
@@ -202,14 +202,7 @@ class _ExampleEditorState extends State<ExampleEditor> {
         Expanded(
           child: _buildEditor(),
         ),
-        if (_isMobile)
-          MultiListenableBuilder(
-            listenables: <Listenable>{
-              _doc,
-              _composer.selectionNotifier,
-            },
-            builder: (_) => _buildMountedToolbar(),
-          ),
+        if (_isMobile) _buildMountedToolbar(),
       ],
     );
   }
@@ -246,10 +239,16 @@ class _ExampleEditorState extends State<ExampleEditor> {
       return const SizedBox();
     }
 
-    return KeyboardEditingToolbar(
-      document: _doc,
-      composer: _composer,
-      commonOps: _docOps,
+    return MultiListenableBuilder(
+      listenables: <Listenable>{
+        _doc,
+        _composer.selectionNotifier,
+      },
+      builder: (_) => KeyboardEditingToolbar(
+        document: _doc,
+        composer: _composer,
+        commonOps: _docOps,
+      ),
     );
   }
 }
--- a/super_editor/example/lib/demos/example_editor/example_editor.dart
+++ b/super_editor/example/lib/demos/example_editor/example_editor.dart
@@ -1,3 +1,4 @@
+import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:super_editor/super_editor.dart';
 
@@ -17,12 +18,13 @@ class _ExampleEditorState extends State<ExampleEditor> {
   final GlobalKey _docLayoutKey = GlobalKey();
 
   late Document _doc;
-  DocumentEditor? _docEditor;
-  DocumentComposer? _composer;
+  late DocumentEditor _docEditor;
+  late DocumentComposer _composer;
+  late CommonEditorOperations _docOps;
 
-  FocusNode? _editorFocusNode;
+  late FocusNode _editorFocusNode;
 
-  ScrollController? _scrollController;
+  late ScrollController _scrollController;
 
   OverlayEntry? _formatBarOverlayEntry;
   final _selectionAnchor = ValueNotifier<Offset?>(null);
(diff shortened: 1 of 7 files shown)


super_editor/lib/src/default_editor/common_editor_operations.dart
```diff
@@ -2092,4 +2093,256 @@ class CommonEditorOperations {
...
+    final content = (await Clipboard.getData('text/plain'))?.text ?? '';
+
+    editor.executeCommand(
+      _PasteEditorCommand(
+        content: content,
+        pastePosition: pastePosition,
+        composer: composer,
+      ),
+    );
+  }
```
roughike, line 2232: Looks like this is old code that was moved to a better location, but I still have a question.

It seems like if the clipboard data is `null`, we still potentially delete the user's current selected content. Also, with the current behavior, if clipboard data is `null`, we default to an empty text.

Is there a reason to not restructure this into following:

```suggestion
  void paste() async {
    DocumentPosition pastePosition = composer.selection!.extent;
    
    final content = (await Clipboard.getData('text/plain'))?.text;
    if (content == null) {
      // Nothing to paste - stop execution.
      return;
    }

    // Delete all currently selected content.
    if (!composer.selection!.isCollapsed) {
      pastePosition = CommonEditorOperations.getDocumentPositionAfterExpandedDeletion(
        document: editor.document,
        selection: composer.selection!,
      );

      // Delete the selected content.
      editor.executeCommand(
        DeleteSelectionCommand(documentSelection: composer.selection!),
      );

      composer.selection = DocumentSelection.collapsed(position: pastePosition);
    }

    // TODO: figure out a general approach for asynchronous behaviors that
    //       need to be carried out in response to user input.
    editor.executeCommand(
      _PasteEditorCommand(
        content: content,
        pastePosition: pastePosition,
        composer: composer,
      ),
    );
  }
```

super_editor/lib/src/infrastructure/_listenable_builder.dart
```diff
@@ -1,5 +1,6 @@
 import 'package:flutter/widgets.dart';
 
+/// Builder that runs every time one of the given [listenables] changes.
```
roughike, line 3: question: could `MultiListenableBuilder` be simplified with `Listenable.merge`?

```dart
class MultiListenableBuilder extends StatelessWidget {
  const MultiListenableBuilder({
    Key? key,
    required this.listenables,
    required this.builder,
  }) : super(key: key);

  final List<Listenable> listenables;
  final WidgetBuilder builder;

  @override
    return ListenableBuilder(
      listenable: Listenable.merge(listenables),
      builder: builder,
    );
  }
}
```

super_editor/example/lib/demos/example_editor/example_editor.dart
```diff
@@ -145,19 +158,97 @@ class _ExampleEditorState extends State<ExampleEditor> {
...
+    return MultiListenableBuilder(
+      listenables: <Listenable>{
+        _doc,
+        _composer.selectionNotifier,
+      },
+      builder: (_) => KeyboardEditingToolbar(
+        document: _doc,
+        composer: _composer,
+        commonOps: _docOps,
+      ),
```
roughike, line 251: With the current code, `KeyboardEditingToolbar` will never be visible on mobile, because we're not listening the selection and rebuilding appropriately.

Something like this fixes it (but surfaces a bug where the `KeyboardEditingToolbar` is visible even though the keyboard isn't):

```suggestion
    return MultiListenableBuilder(
      listenables: <Listenable>{
        _doc,
        _composer.selectionNotifier,
      },
      builder: (_) { 
        final selection = _composer.selection;

        if (selection == null) {
          return const SizedBox();
        }

        return KeyboardEditingToolbar(
          document: _doc,
          composer: _composer,
          commonOps: _docOps,
        );
      },
```

super_editor/lib/src/default_editor/common_editor_operations.dart
matthew-carroll: > It seems like if the clipboard data is null, we still potentially delete the user's current selected content. Also, with the current behavior, if clipboard data is null, we default to an empty text.

Can you file a ticket about this? Given the goal of this PR, and that we're migrating code, I'd prefer to dig into details about how cut/copy/paste work in a dedicated thread, and also avoid holding up this PR for that.

super_editor/lib/src/infrastructure/_listenable_builder.dart
roughike: > Listenable doesn't seem to have a dispose() method, so how do we stop it from listening to the inner Listenables?

`AnimatedBuilder` (which extends from `AnimatedWidget`, and which is already used by `ListenableBuilder`) should do that: https://github.com/flutter/flutter/blob/097d3313d8e2c7f901932d63e537c1acefb87800/packages/flutter/lib/src/widgets/transitions.dart#L106-L125

Calling `.removeListener()` on the merged `Listenable` would stop listening to the inner listenables.

> And would it require us to instantiate and manage a new Listenable instance?

Not outside of the `MultiListenableBuilder` example I posted. Consumers of `MultiListenableBuilder` would just pass a `List<Listenable>` instead of a `Set` and everything should work like it currently does.

> The code in your example would create a new merged Listenable every time the build() method runs. I assume that's not ideal...

Yeah, maybe this would be a bit less ideal, but then again, it shouldn't be too expensive. I think we might already be doing more or less the same with the existing `MultiListenableBuilder`, as we're creating a new `Set` of stuff inside of the `build` methods everywhere where `MultiListenableBuilder` is used.

(I don't think it makes a huge difference, so I'm resolving the conversation)

super_editor/example/lib/demos/example_editor/example_editor.dart
matthew-carroll: Ok, I was able to reproduce and I made the suggested change. Thanks.

matthew-carroll:
I'm assuming that this PR is implicitly approved based on the above comments, so I'm gonna force a merge.

PR #450, opened 2022-02-27
Bugfix: desktop auto-scrolling wouldn't stop after starting (Resolves #441)

### LINKED ISSUE #441: Desktop document auto-scrolling mis-calculates scroll offset
Auto-scrolling in documents is buggy, especially when dragging from an image.

Open the example demo with the image at the top. Start dragging on the image, toward the bottom of the editor.

Auto-scrolling begins, and it keeps going to the bottom of the document, no matter where you move the mouse. Also, the selection extent no longer matches the mouse position.

matthew-carroll:
Bugfix: desktop auto-scrolling wouldn't stop after starting (Resolves #441)

There was an adjustment in the calculation that I was applying on every scroll frame. I'm not entirely sure why it was working originally, and then started failing. I might be missing some use-cases, but I removed the adjustment that I was applying to the user's drag position and it seemed to fix the problem.

### DIFF
--- a/super_editor/lib/src/default_editor/document_gestures_mouse.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_mouse.dart
@@ -105,7 +105,7 @@ class DocumentMouseInteractor extends StatefulWidget {
 }
 
 class _DocumentMouseInteractorState extends State<DocumentMouseInteractor> with SingleTickerProviderStateMixin {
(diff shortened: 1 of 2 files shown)


matthew-carroll:
@roughike I wanted to make sure you saw this PR, too

roughike (approved):

PR #454, opened 2022-03-02
Improve navigating by word

jmatth:
This PR updates the `CharacterMovement` extension on `String` to fix some bugs and more closely match the behavior of other text editors.

Bugs:

- Currently only space is considered when checking for word breaks, causing the cursor to skip over whitespace created with other Unicode characters. This is fixed by using a regex to match on any characters in the Unicode "Separator" category [^1].
- Currently moving upstream by word does not handle multiple spaces well: it will step through one space at a time until it hits a space immediately after a non-space character, then jump to the beginning of the previous word. This is fixed by changing how previous spaces are counted and when the loop is broken in `moveOffsetUpstreamByWord`.

Behavior changes:

- Most other editors jump to the end of each word when moving downstream and the start when moving upstream. The current behavior kind of does this, with upstream movement jumping by start of word but downstream movement jumping to both the start and end of words, whichever it encounters first on each move. This behavior is updated so downstream movement only considers the end of words.
- Most other editors do not consume punctuation when moving upstream and downstream. This behavior is added by updating the previously mentioned regex to also match characters in the Unicode "Punctuation" category.

Here is a recording of the current behavior:

https://user-images.githubusercontent.com/1316184/156390693-17b78612-a290-4199-9f29-d50ca5481d4d.mov

Here is the behavior with the changes from this PR:

https://user-images.githubusercontent.com/1316184/156390855-61b96a80-92c1-41b8-b9a4-51dbb0b022cc.mov

And here's a recording of the same text/moves in the macOS notes app for reference:

https://user-images.githubusercontent.com/1316184/156391037-a626a874-9a1f-408a-ab3a-6df29ad0ea5c.mov

It's worth noting that this doesn't implement perfect parity with other text editors. I observed inconsistent behavior between different editors regarding repeated punctuation and emoji, which I did not try to replicate here.

[^1]: [Unicode General_Category Values Table](http://www.unicode.org/reports/tr44/#GC_Values_Table)

### DIFF
--- a/super_editor/lib/src/infrastructure/strings.dart
+++ b/super_editor/lib/src/infrastructure/strings.dart
@@ -24,7 +24,7 @@ extension CharacterMovement on String {
       return null;
     }
 
-    return _moveOffsetByWord(textOffset, TextAffinity.upstream);
+    return _moveOffsetByWord(textOffset, true);
   }
 
   /// Returns the code point index of the character that sits
@@ -41,7 +41,7 @@ extension CharacterMovement on String {
       return null;
     }
 
-    return _moveOffsetByCharacter(textOffset, characterCount, TextAffinity.upstream);
+    return _moveOffsetByCharacter(textOffset, characterCount, true);
   }
 
   /// Returns the code point index of the character that sits
@@ -59,7 +59,7 @@ extension CharacterMovement on String {
       return null;
     }
 
-    return _moveOffsetByWord(textOffset, TextAffinity.downstream);
+    return _moveOffsetByWord(textOffset, false);
   }
 
   /// Returns the code point index of the character that sits
@@ -76,10 +76,10 @@ extension CharacterMovement on String {
       return null;
     }
 
-    return _moveOffsetByCharacter(textOffset, characterCount, TextAffinity.downstream);
+    return _moveOffsetByCharacter(textOffset, characterCount, false);
   }
 
-  int? _moveOffsetByWord(int textOffset, TextAffinity affinity) {
+  int? _moveOffsetByWord(int textOffset, bool upstream) {
     if (textOffset < 0 || textOffset > length) {
       throw Exception("Index '$textOffset' is out of string range. Length: $length");
     }
@@ -97,7 +97,7 @@ extension CharacterMovement on String {
       remainingOffset -= char.length;
       return remainingOffset >= 0;
     });
-    final moveWhile = affinity == TextAffinity.downstream ? range.expandWhile : range.dropBackWhile;
+    final moveWhile = upstream ? range.dropBackWhile : range.expandWhile;
     // Adjust the range in the requested direction as long it does not end in a word. This accounts for cases where the
     // text offset starts in between words. After this we know the range ends on a word character
     moveWhile((char) => _separatorRegex.hasMatch(char));
@@ -109,7 +109,7 @@ extension CharacterMovement on String {
     return range.current.length;
   }
 
-  int? _moveOffsetByCharacter(int textOffset, int characterCount, TextAffinity affinity) {
+  int? _moveOffsetByCharacter(int textOffset, int characterCount, bool upstream) {
     if (textOffset < 0 || textOffset > length) {
       throw Exception("Index '$textOffset' is out of string range. Length: $length");
     }
@@ -128,17 +128,14 @@ extension CharacterMovement on String {
       return remainingOffset >= 0;
     });
     // Verify that the move is possible with the requested character count
-    if (affinity == TextAffinity.downstream && range.stringAfterLength < characterCount) {
-      return null;
-    }
-    if (affinity == TextAffinity.upstream && range.current.length < characterCount) {
+    if (upstream ? range.current.length < characterCount : range.stringAfterLength < characterCount) {
       return null;
     }
     // Expand or contract the range by the requested number of characters
-    if (affinity == TextAffinity.downstream) {
-      range.expandNext(characterCount);
-    } else {
+    if (upstream) {
       range.dropLast(characterCount);
+    } else {
+      range.expandNext(characterCount);
     }
     // The range now reaches from the start of the string to our new text offset. Calculate that offset using the
     // range's string length and return it
--- a/super_editor/lib/src/infrastructure/strings.dart
+++ b/super_editor/lib/src/infrastructure/strings.dart
@@ -20,37 +20,11 @@ extension CharacterMovement on String {
   ///   word |up -> `0`
   ///   word up| -> `5`
   int? moveOffsetUpstreamByWord(int textOffset) {
-    if (textOffset < 0 || textOffset > length) {
-      throw Exception("Index '$textOffset' is out of string range. Length: $length");
-    }
-
     if (textOffset == 0) {
       return null;
     }
 
-    // Create a character range, initially with zero length
-    // Note that the getter for this object is confusingly named: it is an iterator but includes lots of functionality
-    // beyond that interface, most importantly for us a range over this string that can be manipulated in terms of
-    // characters
-    final range = characters.iterator;
-    // Expand the range so it reaches from the start of the string to the initial text offset. The text offset is passed
-    // to us in terms of bytes but the iterator deals in grapheme clusters, so we need to manually count the length of
-    // each cluster as until we reach the desired offset
-    var remainingOffset = textOffset;
-    range.expandWhile((char) {
-      remainingOffset -= char.length;
-      return remainingOffset >= 0;
-    });
-    // Shrink the range from the end as long it does not end in a word. This accounts for cases where the text offset
-    // starts in between words. After this expansion we know the range ends on a word character
-    range.dropBackWhile((char) => _separatorRegex.hasMatch(char));
-    // Shrink the range from the back until it reaches a non-word character. After this expansion we know that the range
-    // ends on the first character of a word, which is the next word upstream from the initial text offset
-    range.dropBackWhile((char) => !_separatorRegex.hasMatch(char));
-    // The range now reaches from the start of the string to our new text offset. Calculate that offset using the
-    // range's string length and return it
-    final current = range.current;
-    return current.length;
+    return _moveOffsetByWord(textOffset, TextAffinity.upstream);
   }
 
   /// Returns the code point index of the character that sits
@@ -81,36 +55,11 @@ extension CharacterMovement on String {
   ///   word |up -> `7`
   ///   word up| -> `null`
   int? moveOffsetDownstreamByWord(int textOffset) {
-    if (textOffset < 0 || textOffset > length) {
-      throw Exception("Index '$textOffset' is out of string range. Length: $length");
-    }
-
     if (textOffset == length) {
       return null;
     }
 
-    // Create a character range, initially with zero length
-    // Note that the getter for this object is confusingly named: it is an iterator but includes lots of functionality
-    // beyond that interface, most importantly for us a range over this string that can be manipulated in terms of
-    // characters
-    final range = characters.iterator;
-    // Expand the range so it reaches from the start of the string to the initial text offset. The text offset is passed
-    // to us in terms of bytes but the iterator deals in grapheme clusters, so we need to manually count the length of
-    // each cluster as until we reach the desired offset
-    var remainingOffset = textOffset;
-    range.expandWhile((char) {
-      remainingOffset -= char.length;
-      return remainingOffset >= 0;
-    });
-    // Expand the range forward as long it does not end in a word. This accounts for cases where the text offset starts
-    // in between words. After this expansion we know the range ends on a word character
-    range.expandWhile((char) => _separatorRegex.hasMatch(char));
-    // Expand the range forward until it reaches a non-word character. After this expansion we know that the range ends
-    // on the last character of a word, which is the next word downstream from the initial text offset
-    range.expandWhile((char) => !_separatorRegex.hasMatch(char));
-    // The range now reaches from the start of the string to our new text offset. Calculate that offset using the
-    // range's string length and return it
-    return range.current.length;
+    return _moveOffsetByWord(textOffset, TextAffinity.downstream);
   }
 
   /// Returns the code point index of the character that sits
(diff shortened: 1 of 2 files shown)


matthew-carroll:
Thanks for the PR. I'm working on wrapping up the v0.2.0 release. I'll come back to this PR after that's done.

matthew-carroll:
@bk-one @Jethro87 - can both of you confirm that the word movement in this PR is what you'd like to see in your products? I don't have an opinion, personally, so I wanted to run this by you.

bk-one:
This is great - especially as it also includes how Apple Notes handles it. No need to reinvent anything here, if a major not taking app behaves like that on caret movement. Yes, there will always be minor inconsistencies, but staying close to Apple Notes is a good call, imho. I'm all for it.

Jethro87:
I agree, this is great. Happy to have this behavior in super editor.

super_editor/lib/src/infrastructure/strings.dart
```diff
@@ -1,16 +1,21 @@
+import 'dart:collection';
+
 import 'package:characters/characters.dart';
 
+final _separatorRegex = RegExp(r'^[\p{Z}\p{P}]$', unicode: true);
```
matthew-carroll, line 5: Can you add a comment about what this matches and include one or two examples?

super_editor/lib/src/infrastructure/strings.dart
```diff
@@ -19,34 +24,38 @@ extension CharacterMovement on String {
       return null;
     }
 
-    bool isInSpace = false;
+    bool isInSeparator = false;
```
matthew-carroll, line 27: What does "separator" mean?

super_editor/lib/src/infrastructure/strings.dart
```diff
@@ -19,34 +24,38 @@ extension CharacterMovement on String {
...
-      }
-
-      isInSpace = character == " ";
+      isInSeparator = _separatorRegex.hasMatch(character);
       codePointIndex += character.length;
       visitedCharacterCount += 1;
 
-      if (character == " ") {
-        lastSpaceEndCodePointOffset = codePointIndex;
+      if (isInSeparator) {
```
matthew-carroll, line 38: I'm finding this code, and the comments, pretty difficult to understand. Can you take another pass at comments and naming? 

For example, as I read this, I don't know what kind of algorithm I should expect, or what "separator" means in this case, or what we're adding and removing to a queue.

super_editor/lib/src/infrastructure/strings.dart
```diff
@@ -86,16 +95,16 @@ extension CharacterMovement on String {
   }
 
   /// Returns the code point index of the character that sits
-  /// one word downstream from given [textOffset] code point
-  /// index.
+  /// after the next end of word downstream from the given
```
matthew-carroll, line 98: I think this should read "after the end of the next word"

super_editor/lib/src/infrastructure/strings.dart
```diff
@@ -104,29 +113,23 @@ extension CharacterMovement on String {
...
 
-    int lastSpaceStartCodePointOffset = 0;
-    int lastSpaceEndCodePointOffset = 0;
+    int lastSeparatorStartCodePointOffset = 0;
 
     int codePointIndex = 0;
     for (final character in characters) {
-      if (character == " " && !isInSpace) {
-        lastSpaceStartCodePointOffset = codePointIndex;
+      final characterIsSeparator = _separatorRegex.hasMatch(character);
```
matthew-carroll, line 122: In the other method there's a variable called `isInSeparator` that uses the regex, but this one is called `characterIsSeparator`, and yet there's also a variable here called `isInSeparator`. Can you check these again for consistent naming and meaning?

super_editor/test/src/infrastructure/strings_test.dart
```diff
@@ -23,14 +23,20 @@ void main() {
...
       test("a word", () {
-        expect("move a💙c words".moveOffsetUpstreamByWord(15), 10);
-        expect("move a💙c words".moveOffsetUpstreamByWord(10), 9);
-        expect("move a💙c words".moveOffsetUpstreamByWord(9), 5);
-        expect("move a💙c words".moveOffsetUpstreamByWord(8), 5);
-        expect("move a💙c words".moveOffsetUpstreamByWord(4), 0);
-        expect("move a💙c words".moveOffsetUpstreamByWord(0), null);
-        expect(() => "move a💙c words".moveOffsetUpstreamByWord(-1), throwsException);
-        expect(() => "move a💙c words".moveOffsetUpstreamByWord(16), throwsException);
+        expect("  move a💙c\u{10B3F}wo.rds".moveOffsetUpstreamByWord(19), 16);
```
matthew-carroll, line 26: Is there a reason for the changes to all these tests? The inclusion of an emoji in the original tests makes it pretty clear that we're testing a multi-codepoint character, but in the absence of comments, it's tough to know why it's important that we have empty whitespace, a period, and whatever `u10B3F` is. It's probably a good idea to remove those details from any expectation that isn't interacting with those details, and to add comments to the expectations that care.

matthew-carroll (commented):
Thanks for the contribution - I have a few comments

super_editor/lib/src/infrastructure/strings.dart
jmatth: The word "separator" is in a lot of this code because my first implementation just swapped out matching a space characters for matching any graphemes in the Unicode "Separator" category. After updating it to also consider graphemes in the "Punctuation" category there wasn't an obvious better name and it can still make sense in that we're treating these graphemes as separating the parts of the string we care about navigating to, so I just left it. I'm happy to change it to whatever if there's a more descriptive name you'd prefer. Though we probably shouldn't tie the variable names to the regex with something like `isInSeparatorOrPunctuation` just in case this ever needs to be updated, then suddenly a bunch of lines need to be changed to `isInSpaceSeparatorOrOtherPunctuation` or whatever to match

super_editor/lib/src/infrastructure/strings.dart
jmatth: I took another pass at this and realized it's pretty over-engineered. I'll fix it in the next push

super_editor/lib/src/infrastructure/strings.dart
jmatth: Done, they should be more consistently/clearly named now

super_editor/lib/src/infrastructure/strings.dart
```diff
@@ -1,52 +1,71 @@
...
     }
 
-    bool isInSpace = false;
-
-    int lastSpaceStartCodePointOffset = 0;
-    int lastSpaceEndCodePointOffset = 0;
-
-    int visitedCharacterCount = 0;
-    int codePointIndex = 0;
+    // We'll store the end of the most recently seen separator grapheme here
```
matthew-carroll, line 32: I think we can still dumb down these comments a lot. It's likely that a reader isn't going to know much about grapheme clusters, runes, codepoints, etc. Most of what you're doing here looks like it can be explained without those concepts.

I may have done a poor job with the original names and comments, so don't put too much weight in them.

Generally speaking, if you're about to jump into the implementation of an algorithm, it's a good idea to begin with a comment that explains what the algorithm is about to accomplish, and, at a high level, what it's going to do. Otherwise, the reader immediately sees comments about edge cases, but the reader is left wondering "edge cases to what?". Best to start with the big picture, and possibly include an example.

Then, wherever possible, use concepts that the user understands, such as "word" and "character". If you truly need to refer to a codepoint or grapheme, consider including a very concise description about what make a codepoint or grapheme different from a character, and why it matters to us in this case. If we were implementing the `characters` package ourselves, then we could probably assume that everyone knows these concepts. But when we're just implementing the idea of moving the caret over a word, we shouldn't expect such deep knowledge of text encoding.

super_editor/lib/src/infrastructure/strings.dart
```diff
@@ -1,52 +1,71 @@
...
+    int nextSeparatorEndCodePointOffset = 0;
+    // ...and flush it to here when we see a non-separator grapheme. This
```
matthew-carroll, line 34: As I read this comment, I find that I don't know what "flush" means. I'm also not clear what "when we see" means. I'm not sure what the "requested textOffset" is - I don't think any offset is being requested. One was provided, but that's an originating text offset, not a requested or target offset.

It's also risky to use words like "precede" and "follow" when you're moving a caret through text. Do you mean "precede" from the vantage point of caret movement, or "precede" from the vantage point of text affinity. If you mean text affinity, then a better word would be "upstream".

super_editor/lib/src/infrastructure/strings.dart
```diff
@@ -1,52 +1,71 @@
...
+    // addresses the case where multiple consecutive separator graphemes
+    // immediately precede the requested textOffset
+    int lastSeparatorEndCodePointOffset = 0;
+
+    // We want to return the code point index immediately after the last
```
matthew-carroll, line 39: I think this comment misses the goal, and also gets a bit too specific with what's about to happen.

I'm not sure, but let me know if the following would be more appropriate:

```
// The caret sits at the given textOffset. We need to find the text offset that sits 
// "one word upstream" from the given textOffset. Move character-by-character
// upstream until we find the border between two words. The border between two
// words is defined to be the offset after the last character in a preceding word.
//
// For example: "abcd efghi|" -> "abcd| efghi"
//
// The types of characters that are considered "within a word" are determined by
// _separatorRegex. 
```

Actually, I think the above suggestion might make sense as the intro for the entire algorithm. If you use that as the intro, then all you need in the comment before the "for loop" is something like:

```
// Move character by character upstream until we find a word boundary
```

matthew-carroll: Er, looking at this again, I described the algorithm incorrectly (see, that's how easy it is to get things wrong). The algorithm starts at the beginning of the text when looking for the word boundary. I guess we do that because of unknown character codepoint sizes? If so, see if you can update the recommended comments accordingly. The rationale for why we start at the beginning of the text is exactly the kind of detail that a reader is likely to want, and also a detail that the reader probably wouldn't assume without help.

super_editor/lib/src/infrastructure/strings.dart
```diff
@@ -1,52 +1,71 @@
...
+    // one ahead of the current character in each iteration of the loop
+    int nextCharacterIndex = 0;
+    int nextCodePointIndex = 0;
     for (final character in characters) {
-      if (visitedCharacterCount >= textOffset - 1) {
-        // We're at the given text offset. The upstream word offset is
-        // at lastSpaceEndCodePointOffset.
+      nextCharacterIndex += 1;
+      if (nextCharacterIndex >= textOffset) {
+        // We're at the given text offset. The upstream word offset is in
```
matthew-carroll, line 47: How about:

```
// We're looking for a text offset that's upstream from the given
// textOffset. We just passed the given textOffset, therefore we
// already found the nearest upstream word boundary, or no such
// word boundary exists. Either way, we're done.
```

super_editor/lib/src/infrastructure/strings.dart
```diff
@@ -1,52 +1,71 @@
...
-      isInSpace = character == " ";
-      codePointIndex += character.length;
-      visitedCharacterCount += 1;
-
-      if (character == " ") {
-        lastSpaceEndCodePointOffset = codePointIndex;
+      nextCodePointIndex += character.length;
+      final isInSeparator = _separatorRegex.hasMatch(character);
+
+      if (isInSeparator) {
```
matthew-carroll, line 55: Is this actually checking that the offset is in a separator, or is this checking that the offset is in a word?

super_editor/lib/src/infrastructure/strings.dart
```diff
@@ -86,48 +106,42 @@ extension CharacterMovement on String {
...
   ///
   /// Examples:
   ///   |word up -> `4`
   ///   wo|rd up -> `4`
-  ///   word| up -> `5`
+  ///   word| up -> `7`
   ///   word |up -> `7`
   ///   word up| -> `null`
-  int? moveOffsetDownstreamByWord(int textOffset, {int characterCount = 1}) {
+  int? moveOffsetDownstreamByWord(int textOffset) {
```
matthew-carroll, line 118: Please be sure to apply all appropriate changes from the other method to this one, in terms of comments, naming, structure, etc.

super_editor/test/src/infrastructure/strings_test.dart
```diff
@@ -14,23 +14,89 @@ void main() {
       });
 
       test("2 characters", () {
-        expect("a💙c".moveOffsetUpstreamByCharacter(0, characterCount: 2), null);
-        expect("a💙c".moveOffsetUpstreamByCharacter(1, characterCount: 2), null);
+        expect(
+            "a💙c".moveOffsetUpstreamByCharacter(0, characterCount: 2), null);
```
matthew-carroll, line 18: Looks like are a bunch of formatting changes without content changes? If so, please return them to the way they were. Throughout super_editor, please set your line length to 120.

super_editor/test/src/infrastructure/strings_test.dart
```diff
@@ -14,23 +14,89 @@ void main() {
...
+          expect("move. words".moveOffsetUpstreamByWord(5), 0);
+          expect("move. words".moveOffsetUpstreamByWord(4), 0);
+          expect("move. words".moveOffsetUpstreamByWord(0), null);
+          expect(() => "move. words".moveOffsetUpstreamByWord(-1),
+              throwsException);
+          expect(() => "move. words".moveOffsetUpstreamByWord(12),
+              throwsException);
+        });
+
+        test("separated by multi-byte punctuation", () {
```
matthew-carroll, line 79: The test breakdown looks good.

super_editor/test/src/infrastructure/strings_test.dart
jmatth: I only just noticed that `super_editor/.vscode` exists hasn't been applying because I had the repo root open in VS Code instead of the `super_editor` sub-directory. Totally my bad, though it might be worth moving that up to the root now that there are several Flutter packages in different sub-directories rather than just `super_editor` and the website. I'll re-format this with the correct settings in any case

super_editor/lib/src/infrastructure/strings.dart
```diff
@@ -1,52 +1,71 @@
 import 'package:characters/characters.dart';
 
+// Match any characters we want skip over while moving by word. This will match
```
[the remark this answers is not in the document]
jmatth: > "Matches any character that's considered to be part of a word, and is used to jump the caret from the beginning of a word to the end, or the end of a word to the beginning."

It's the other way around: this matches characters we don't consider part of a word (spaces and punctuation). So maybe: "Matches any character that's not considered part of a word, which is then used to find word boundaries and jump the caret to the beginning or end of a word."

super_editor/lib/src/infrastructure/strings.dart
jmatth: > The algorithm starts at the beginning of the text when looking for the word boundary. I guess we do that because of unknown character codepoint sizes?

I think so, though it was like that when I got here so its possible I'm missing something. As far as I can tell both methods take the start offset based on grapheme index but return a result based on code-point index, which seems weird but I assume is necessary because of some detail farther up in the super_editor stack. We have to walk one character at a time to account for both multi-byte code points as well as graphemes composed of multiple combining code points.

Actually, now that I'm looking closer at the [characters api](https://pub.dev/documentation/characters/latest/) I think it might be possible to simplify things even more by offloading more work to that library. I'll take another crack at it and update the docstrings again with whatever the end result is

jmatth:
@matthew-carroll After some research I was able to replace the custom character/code point counting logic with functionality from the `characters` library. I also updated the move by character methods to use the same library

matthew-carroll:
@jmatth can you confirm that all your tests pass locally? There are a bunch of test failures in this PR, but I think they're probably due to the new mono repo dependency setup...

jmatth:
@matthew-carroll Yes, the tests pass on my machine[:](https://www.leadingagile.com/wp-content/uploads/2017/03/worksonmymachine.png)

```shell
$ flutter test
...
00:10 +384: All tests passed!
```

The test failures look like they're caused by pub refusing to allow dependencies that themselves have path dependencies, which was also causing issues with updating our fork of super_editor but I haven't had time to dig too far into it

matthew-carroll (approved):
LGTM - Thanks

PR #458, opened 2022-03-09
[DO NOT MERGE] Using `customStylePhases`

miguelcmedeiros:
@matthew-carroll 
This PR intends to implement what we discussed during our call.

It currently applies the behaviour that we intended (hide the hint text whenever the document selection is not collapsed, even when it overlaps with the empty text nodes).

There is one thing that I got wrong somehow which is, the `highlightWhenEmpty` property seems to always false for the header nodes. 

Is this what you had in mind?

### DIFF
--- a/super_editor/example/lib/demos/components/demo_text_with_hint.dart
+++ b/super_editor/example/lib/demos/components/demo_text_with_hint.dart
@@ -22,6 +22,7 @@ class _TextWithHintDemoState extends State<TextWithHintDemo> {
   late MutableDocument _doc;
   late DocumentEditor _docEditor;
   late DocumentComposer _composer;
+  late List<SingleColumnLayoutStylePhase> _customStylePhases;
 
   @override
   void initState() {
@@ -29,6 +30,9 @@ class _TextWithHintDemoState extends State<TextWithHintDemo> {
     _doc = _createDocument();
     _docEditor = DocumentEditor(document: _doc);
     _composer = DocumentComposer();
+    _customStylePhases = [
+      ParagraphWithHintStyler(_composer),
+    ];
   }
 
   @override
@@ -74,9 +78,7 @@ class _TextWithHintDemoState extends State<TextWithHintDemo> {
     return SuperEditor(
       editor: _docEditor,
       composer: _composer,
-      customStylePhases: [
-        ParagraphWithHintStylesheetStyler(_composer),
-      ],
+      customStylePhases: _customStylePhases,
       stylesheet: Stylesheet(
         documentPadding: const EdgeInsets.symmetric(vertical: 56, horizontal: 24),
         rules: defaultStylesheet.rules,
@@ -170,8 +172,6 @@ class HeaderWithHintComponentBuilder implements ComponentBuilder {
       return null;
     }
 
-    final textSelection = componentViewModel.selection;
-
     return TextWithHintComponent(
       key: componentContext.componentKey,
       text: componentViewModel.text,
@@ -181,8 +181,7 @@ class HeaderWithHintComponentBuilder implements ComponentBuilder {
               'blockType': componentViewModel.blockType,
             }
           : {},
-      // This is the text displayed as a hint.
-      hintText: !componentViewModel.hideHintText && componentViewModel.caret != null
+      hintText: componentViewModel.shouldShowHintText && componentViewModel.caret != null
           ? AttributedText(
               text: 'header goes here...',
               spans: AttributedSpans(
@@ -197,7 +196,7 @@ class HeaderWithHintComponentBuilder implements ComponentBuilder {
       hintStyleBuilder: (Set<Attribution> attributions) => _textStyleBuilder(attributions).copyWith(
         color: const Color(0xFFDDDDDD),
       ),
-      textSelection: textSelection,
+      textSelection: componentViewModel.selection,
       selectionColor: componentViewModel.selectionColor,
       showCaret: componentViewModel.caret != null,
     );
@@ -219,7 +218,7 @@ class ParagraphWithHintComponentViewModel extends ParagraphComponentViewModel {
     TextPosition? caret,
     required Color caretColor,
     bool highlightWhenEmpty = false,
-    required this.hideHintText,
+    required this.shouldShowHintText,
   }) : super(
           nodeId: nodeId,
           maxWidth: maxWidth,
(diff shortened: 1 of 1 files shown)


super_editor/example/lib/demos/components/demo_text_with_hint.dart
```diff
@@ -70,6 +73,10 @@ class _TextWithHintDemoState extends State<TextWithHintDemo> {
   Widget build(BuildContext context) {
     return SuperEditor(
       editor: _docEditor,
+      composer: _composer,
+      customStylePhases: [
+        ParagraphWithHintStylesheetStyler(_composer),
```
matthew-carroll, line 78: It's probably a better idea to create this in `initState()` because one instance should outlive widget rebuilds.

super_editor/example/lib/demos/components/demo_text_with_hint.dart
```diff
@@ -154,7 +161,7 @@ class HeaderWithHintComponentBuilder implements ComponentBuilder {
   @override
   Widget? createComponent(
       SingleColumnDocumentComponentContext componentContext, SingleColumnLayoutComponentViewModel componentViewModel) {
-    if (componentViewModel is! ParagraphComponentViewModel) {
+    if (componentViewModel is! ParagraphWithHintComponentViewModel) {
       return null;
     }
```
matthew-carroll, line 167: Given that you've added a style phase, which gives you more control over producing view models, the header block attribution check below this line could probably be moved into the place where you create the view model. In other words, if the given node is not a header, you could decide not to build a `ParagraphWithHintComponentViewModel` at all, and then skip the check here.

super_editor/example/lib/demos/components/demo_text_with_hint.dart
```diff
@@ -175,15 +182,17 @@ class HeaderWithHintComponentBuilder implements ComponentBuilder {
...
-      hintText: AttributedText(
-        text: 'header goes here...',
-        spans: AttributedSpans(
-          attributions: [
-            const SpanMarker(attribution: italicsAttribution, offset: 12, markerType: SpanMarkerType.start),
-            const SpanMarker(attribution: italicsAttribution, offset: 15, markerType: SpanMarkerType.end),
-          ],
-        ),
-      ),
+      hintText: !componentViewModel.hideHintText && componentViewModel.caret != null
```
matthew-carroll, line 185: I would probably move all conditionals related to hint text into the styler. The styler has all the info needed to make this decision, right? If so, why not make the final decision available in `componentViewModel.shouldShowHintText`?

super_editor/example/lib/demos/components/demo_text_with_hint.dart
```diff
@@ -194,3 +203,150 @@ class HeaderWithHintComponentBuilder implements ComponentBuilder {
...
+    Document document,
+    SingleColumnLayoutComponentViewModel viewModel,
+  ) {
+    final node = document.getNodeById(viewModel.nodeId)!;
+
+    if (node is TextNode && viewModel is ParagraphComponentViewModel) {
+      final documentSelection = composer.selection;
+      final hideHintText = documentSelection?.isCollapsed == false;
+
+      return ParagraphWithHintComponentViewModel(
```
matthew-carroll, line 332: This is where you could decide to only build for header blocks, if that's the behavior that you want.

super_editor/example/lib/demos/components/demo_text_with_hint.dart
```diff
@@ -194,3 +203,150 @@ class HeaderWithHintComponentBuilder implements ComponentBuilder {
...
+
+  SingleColumnLayoutComponentViewModel _applySelection(
+    Document document,
+    SingleColumnLayoutComponentViewModel viewModel,
+  ) {
+    final node = document.getNodeById(viewModel.nodeId)!;
+
+    if (node is TextNode && viewModel is ParagraphComponentViewModel) {
+      final documentSelection = composer.selection;
+      final hideHintText = documentSelection?.isCollapsed == false;
```
matthew-carroll, line 330: I would make the final decision here, instead of waiting for the component builder. Both should technically work, but I think that making the hint text decision here would be more appropriate.

matthew-carroll (commented):
This looks like pretty much what I was recommending. I made a few minor suggestions.

This code does what you need it to, right?

super_editor/example/lib/demos/components/demo_text_with_hint.dart
```diff
@@ -174,23 +176,176 @@ class HeaderWithHintComponentBuilder implements ComponentBuilder {
...
+    final node = document.getNodeById(viewModel.nodeId)!;
+
+    if (node is TextNode && viewModel is ParagraphComponentViewModel) {
+      final blockAttribution = viewModel.blockType;
+      if (!(const [header1Attribution, header2Attribution, header3Attribution]).contains(blockAttribution)) {
+        return viewModel;
+      }
+
+      final documentSelection = composer.selection;
+      final shouldShowHintText = documentSelection?.isCollapsed == true;
```
miguelcmedeiros, line 329: I tried doing something like
```suggestion
      final shouldShowHintText = documentSelection?.isCollapsed == true && viewModel.caretColor != null;
```
But `viewModel.caretColor` is always null. Any idea why?

matthew-carroll: I can't quite tell where this is in the code because GitHub is just showing me a snippet, but it looks like you're implementing the custom style phase?

If so, first, you don't want to make styling decisions based on the `viewModel` that you're given, because you have no control over what has, or hasn't been done to that `viewModel` up to this point. For example, in this case, the reason `caretColor` is null is because the selection styler runs after the custom style phases. This is because user selection is the most volatile behavior, and I didn't want to re-run all the custom style phases every time the user taps somewhere, or drags across a new character.

Further, I'm guessing that the official policy for "when to show hint text" probably isn't determined by the caret color, right? I think your rule here should be "Show the hint text when the selection is collapsed, within this node". That condition would be:

```
final shouldShowHintText = documentSelection != null && documentSelection!.isCollapsed && documentSelection!.extent.nodeId == node.id;
```

I may have gotten some of those terms wrong from memory, but I think that's what you're looking for, right?

miguelcmedeiros: `viewModel.caretColor != null`, I meant `viewModel.caret != null` but copied the wrong text... 🤦‍♂️ 

Still, it would be incorrect given the presenter pipeline (selection styler runs after the custom stylers). Thanks for explaining it once again. It's getting there 🙂

miguelcmedeiros:
Closing it since the goal was achieved 👍

miguelcmedeiros:
Closing it since the goal was achieved 👍

PR #476, opened 2022-03-18
Extracted attributed text into its own package (Resolves #473, Resolves #466)

### LINKED ISSUE #466: Copy LICENSE into super_editor dir for pub.dev
Copy LICENSE into super_editor dir for pub.dev. The license type is shown on pub.dev, but pub.dev doesn't know about the root directory, so it can't find the license.

### LINKED ISSUE #473: Extract AttributedText into new project
Extract AttributedText into new Dart project in the mono-repo.

matthew-carroll:
Extracted attributed text into its own package (Resolves #473, Resolves #466)

The new package is `attributed_text` and it sits at the top of the repo's directory structure, next to `super_editor`.

There are no intentional behavior changes in this PR.

I thought it would good to keep `attributed_text` from depending on Flutter, given that it's entirely logical. This required a few changes:

 * The behavior that converts `AttributedText` to a Flutter `TextSpan` was moved into an extension method in `super_editor`
 * The concept of `TextRange` in `AttributedText` was replaced with a copy of the same thing, now called `SpanRange`
 * `AttributedText` now implements its own listener system because `ChangeNotifier` is a Flutter class
 * `VoidCallback` is declared as a typedef, because that's also a Flutter artifact

At the moment, I have `super_editor` pointing to a relative path for `attributed_text`. We'll need to change this to a legitimate pub deployment before we deploy `super_editor` to pub again.


LICENSE
```diff
@@ -0,0 +1,7 @@
+Copyright (c) 2021 Superlist, SuperDeclarative! and the contributors
```
miguelcmedeiros, line 1:
```suggestion
Copyright (c) 2022 Superlist, SuperDeclarative! and the contributors
```

attributed_text/lib/src/attributed_text.dart
```diff
@@ -348,23 +330,27 @@ class AttributedText with ChangeNotifier {
...
 
-/// Creates the desired [TextStyle] given the [attributions] associated
-/// with a span of text.
+/// A zero-parameter function that returns nothing.
 ///
-/// The [attributions] set may be empty.
-typedef AttributionStyleBuilder = TextStyle Function(Set<Attribution> attributions);
+/// This is the same as Flutter's `VoidCallback`. It's replicated in this
+/// project to avoid depending on Flutter.
+typedef VoidCallback = void Function();
```
miguelcmedeiros, line 356: It seems that Dart analyzer handles this well. I expected that if we would import `super_editor` and something like `flutter/widgets` on the same file, it would give a name clash error since `VoidCallback` is defined in both.

Somehow it handles it and the IDE points then at the `attributed_text` implementation 🤷‍♂️

miguelcmedeiros (commented):
[pr_validation.yaml](https://github.com/superlistapp/super_editor/blob/main/.github/workflows/pr_validation.yaml) could be updated to run the tests in `attributed_text`.

Overall looks good, but these changes break the website package again.

Since there are not that many lint warnings in the project, we might be reaching a point where the PR validation check also runs Dart analyzer for all packages.

LICENSE
matthew-carroll: This is a copy of the existing license. I duplicated it in the root directory so that pub.dev picks it up. I think it should stay the same because the copyright began in 2021. If a lawyer disagrees, I can change it.

matthew-carroll:
@miguelcmedeiros I updated the marketing webpage, and I added the attributed_text tests to CI.

matthew-carroll:
@miguelcmedeiros can I merge this?

miguelcmedeiros (approved):
LGTM 👍

PR #477, opened 2022-03-18
Fix typo `composable` in `pubspec.yaml`

nilsreichardt:
I think `composable` should be the correct word.


matthew-carroll (approved):
LGTM - thanks

PR #484, opened 2022-03-22
Extracted super_selectable_text package (Resolves #474)

### LINKED ISSUE #474: Extract SuperSelectableText into new project
Extract SuperSelectableText into a new Flutter project in the mono-repo.

matthew-carroll:
Extracted super_selectable_text package (Resolves #474)

The new package deserves some tests and better README documentation, but we can handle sometime later when we don't have other packages waiting to be extracted.

There shouldn't be any behavior changes, with one exception. Previously, character/codepoint movement was implemented within `super_selectable_text.dart`, but the codepoint/character mapping is fundamental to text encoding, so I moved that down into `attributed_text`.


miguelcmedeiros (approved):
LGTM

PR #485, opened 2022-03-22
Bugfix: header2 wasn't being applied on mobile

Jethro87:
H2's were not being applied properly due to an incorrect boolean check in the document IME toolbar.


matthew-carroll (approved):
LGTM - will merge when CI completes

PR #494, opened 2022-03-24
Update OrderedListItemComponent, UnorderedListItemComponent and BlockquoteComponent to show a selection highlight when empty

### LINKED ISSUE #386: Selection highlight when the node has no content is not supported for all text components

miguelcmedeiros:
Fixes #386


matthew-carroll (approved):
LGTM

Issue #491, opened 2022-03-23
Ctrl+E & Ctrl+A support on Mac


kevmoo:
Only issue I hit hacking on the doc. I'm used to using

CTRL+E for end-of-line
CTRL+A for beginning-of-line

Would be great to have this support!

matthew-carroll:
Hey look, its @kevmoo! Welcome to the content-heavy side of Flutter!

I'm sure we can add these shortcuts as long as it doesn't conflict with default expectations. @bk-one what do you think?

bk-one:
Absolutely, one of my task is actually to check the full range of expected keyboard shortcuts and send you a list. So yeah, this should be done.

Issue #479, opened 2022-03-18
Allow CMD/ALT+BACKSPACE to delete words / lines


bk-one:
**To reproduce:**
- open e.g. https://superlist.com/SuperEditor/
- try to delete a word using OPTION+BACKSPACE (on Mac)
- tro to delete the whole line using CMD+BACKSPACE (on Mac)

Both only trigger a single backspace.

**Expected behavior (like in other editors)**
- OPTION+BACKSPACE (on Mac) to delete a word
- CMD+BACKSPACE (on Mac) to delete the whole line and place caret at the beginning of the line

zurbrandon:
hello

PR #497, opened 2022-03-26
Extended Keyboard actions to support issues in #479 and #491

agenthesh:
This PR adds a few new keyboard actions, and updates a previous one, to support some of the missing keyboard shortcuts with other text editors.  (#479, #491)

Working video: 

https://user-images.githubusercontent.com/55655058/160232254-163d14c7-1abd-4e20-88cf-e614e98f043a.mp4

This is a first PR into this repository, so feedback would be appreciated!


super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -592,6 +592,9 @@ final defaultKeyboardActions = <DocumentKeyboardAction>[
   anyCharacterOrDestructiveKeyToDeleteSelection,
   anyCharacterToInsertInParagraph,
   anyCharacterToInsertInTextContent,
+  moveToLineStartOrEndWithCtrlAOrE,
```
matthew-carroll, line 595: Can you move this new items above all of the `anyCharacterXXXX` methods? The more generic a handler, the further down the list I like to keep it.

super_editor/lib/src/default_editor/document_keyboard_actions.dart
```diff
@@ -299,3 +304,85 @@ ExecutionInstruction moveUpDownLeftAndRightWithArrowKeys({
 
   return didMove ? ExecutionInstruction.haltExecution : ExecutionInstruction.continueExecution;
 }
+
+ExecutionInstruction moveToLineStartOrEndWithCtrlAOrE({
```
matthew-carroll, line 308: Can you add tests for these actions? It looks like we have some existing tests under `test/src/default_editor/document_keyboard_actions_test.dart`. It would be great if you can test the various edge cases, which typically include some tests for what *should* happen, but also some for what *shouldn't* happen. Let me know if you have any questions.

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -1276,6 +1277,40 @@ class DefaultSuperTextFieldKeyboardHandlers {
...
+  static TextFieldKeyboardHandlerResult moveCaretToStartOrEnd({
+    required AttributedTextEditingController controller,
+    SuperSelectableTextState? selectableTextState,
+    required RawKeyEvent keyEvent,
+  }) {
+    bool _moveLeft = false;
+    if (!keyEvent.isControlPressed) {
+      return TextFieldKeyboardHandlerResult.notHandled;
+    }
+    if (![LogicalKeyboardKey.keyA, LogicalKeyboardKey.keyE].contains(keyEvent.logicalKey)) {
```
matthew-carroll, line 1289: Do we use this approach of creating a set of keys elsewhere? I think I remember using traditional `&&` conditions. If so, it's good for us to maintain consistency where we can.

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -1276,6 +1277,40 @@ class DefaultSuperTextFieldKeyboardHandlers {
...
+        : keyEvent.logicalKey == LogicalKeyboardKey.keyE
+            ? _moveLeft = false
+            : null;
+
+    controller.moveCaretHorizontally(
+      selectableTextState: selectableTextState!,
+      expandSelection: false,
+      moveLeft: _moveLeft,
+      movementModifiers: {
+        'movement_unit': 'line',
```
matthew-carroll, line 1307: Just to verify, is "line" a literal everywhere else, too? Are we missing a constant for that?

super_editor/test/src/infrastructure/super_textfield/desktop/super_desktop_textfield_keyboard_test.dart
```diff
@@ -2458,6 +2458,125 @@ void main() {
           expect(controller.text.text, '-->f<--');
         });
       });
+
+      group("Move caret with control A/E", () {
+        group("MacOS", () {
```
matthew-carroll, line 2463: For clarity, and to help avoid edge cases, it would be a good idea to add tests for these movements where the caret sits at the beginning and end of the line. One might expect some weird things when you try to move by line or delete a line when you're at the beginning or end.

matthew-carroll (changes requested):
Thanks for the contribution. I left a few comments for you.

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
agenthesh: I actually referenced your function, `moveUpDownLeftAndRightWithArrowKeys` for the movementModifiers, but the constants do exist. Will apply changes to both.

agenthesh: Okay, no, now I remember why I did that, its cause the moveCaretHorizontally function takes in a Map as movementModifiers, rather than the Set of MovementModifiers. I've refactored it to use the constants now

agenthesh (commented):
Hey Matt, I've added the changes that you've requested! Let me know if they are okay.

agenthesh:
Also, related to this, I don't think it's the right behavior to expect the function to do nothing when there is a selection and you press CMD+BKSP or ALT+BKSP. You had existing tests for this, so I wrote the function to satisfy these. But normally, in a text editor when there is a selection, and you use these commands, it deletes the selection. Is this something I should refactor too?

super_editor/test/src/infrastructure/super_textfield/desktop/super_desktop_textfield_keyboard_test.dart
```diff
@@ -2458,6 +2458,474 @@ void main() {
...
+          testWidgets('Caret at 5, Control A moves caret to start', (tester) async {
```
matthew-carroll, line 2464: I'm gonna leave some comments about the various test names. It's important to express test intentions very clearly and concisely, because we accumulate a large number of tests and they provide the primary tool against regressions.

In the case of these tests that start with "Caret at 5", that statement is technically true, but it's not the intention of the test. The test aims to verify behavior with a caret in the middle of the text. Therefore, a more appropriate name would be:

"caret in middle of line, Ctl+A moves caret to start"

Also, notice a couple other details:
 * I started with a lowercase letter so that the accumulated test name might look like a sentence. I may have screwed this up in other areas, but over time I'll fix some of the caps in the test names. Ideally, only a top-level `group` begins with a capital letter.
 * Across your test names, you have different ways of describing shortcuts. Let's pick one. In this case, let's go with "Ctrl+A" and "Ctrl+E". Here again, I may have done the wrong thing elsewhere. I'll need to go back and clean it up.

Please make the appropriate changes to the various test names.

matthew-carroll:
> Also, related to this, I don't think it's the right behavior to expect the function to do nothing when there is a selection and you press CMD+BKSP or ALT+BKSP. You had existing tests for this, so I wrote the function to satisfy these. But normally, in a text editor when there is a selection, and you use these commands, it deletes the selection. Is this something I should refactor too?

@agenthesh - I think that's a reasonable change to make, if it's clear to you how to make it.

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -1300,13 +1330,11 @@ class DefaultSuperTextFieldKeyboardHandlers {
     if (keyEvent.logicalKey == LogicalKeyboardKey.arrowLeft) {
       _log.finer('moveUpDownLeftAndRightWithArrowKeys - handling left arrow key');
 
-      final movementModifiers = <String, dynamic>{
-        'movement_unit': 'character',
-      };
+      Set<MovementModifier> movementModifiers = {};
```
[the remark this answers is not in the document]
agenthesh: This may take a lot of refactoring as you've used Sets everywhere when it comes to movementModifiers. I am refactoring a lot of the code to use single movementModifier values instead. movementModifiers is now nullable (and not a Set). A value of null signifies movement character-by-character. Tests have all passed.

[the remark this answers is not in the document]
matthew-carroll: Maybe I'm misunderstanding something. The code that was there before your change was a `Map`, not a `Set`. Can you describe what you mean by "you've used Sets everywhere when it comes to movementModifiers"?

super_editor/lib/src/default_editor/document_keyboard_actions.dart
```diff
@@ -299,3 +304,85 @@ ExecutionInstruction moveUpDownLeftAndRightWithArrowKeys({
...
+  required EditContext editContext,
+  required RawKeyEvent keyEvent,
+}) {
+  if (!keyEvent.isControlPressed && Platform.instance.isMac) {
```
matthew-carroll, line 312: This if-statement took me a few times to read to grasp the goal. Can you break this into two if statements?

```
if (!keyEvent.isControlPressed) {
  return ...
}

if (!Platform.instance.isMac) {
  return ...
}
```

These conditions aren't related and when a developer is quickly skimming through these handlers, it's much easier to read independent conditions, independently.

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -1276,6 +1278,34 @@ class DefaultSuperTextFieldKeyboardHandlers {
     return TextFieldKeyboardHandlerResult.handled;
   }
 
+  static TextFieldKeyboardHandlerResult moveCaretToStartOrEnd({
```
matthew-carroll, line 1281: Can you add a Dart Doc for this method?

super_editor/test/src/infrastructure/super_textfield/desktop/super_desktop_textfield_keyboard_test.dart
```diff
@@ -2458,6 +2464,480 @@ void main() {
           expect(controller.text.text, '-->f<--');
         });
       });
+
```
matthew-carroll, line 2467: Did you end up implementing deletion for selected text with alt+backspace and cmd+backspace? If so, I want to confirm that there are appropriate tests for those. I didn't notice them, but I might be missing those tests as I scroll through them.

matthew-carroll (commented):
Just a few final things

matthew-carroll:
FYI @agenthesh it looks like tests are failing on Windows and Linux

super_editor/test/src/infrastructure/super_textfield/desktop/super_desktop_textfield_keyboard_test.dart
agenthesh: I did yeah! There were existing tests which tested the old behavior, so changed those tests to work with the new behavior instead. [This commit](https://github.com/superlistapp/super_editor/pull/497/commits/c841ba32a105b6da3a58638c073263f895f8f937) has these changes if you wanna look over them.

agenthesh:
Sorry @matthew-carroll, but do you mind helping me with why the tests are failing? 

`[document_keyboard_actions.dart] - moveToLineStartOrEndWithCtrlAOrE:  - handling Ctrl+A
[578](https://github.com/superlistapp/super_editor/runs/5779443590?check_suite_focus=true#step:5:578)
[579](https://github.com/superlistapp/super_editor/runs/5779443590?check_suite_focus=true#step:5:579)
The following TestFailure was thrown running a test:
[580](https://github.com/superlistapp/super_editor/runs/5779443590?check_suite_focus=true#step:5:580)
Expected: <2>
[581](https://github.com/superlistapp/super_editor/runs/5779443590?check_suite_focus=true#step:5:581)
  Actual: <1>
[582](https://github.com/superlistapp/super_editor/runs/5779443590?check_suite_focus=true#step:5:582)

[583](https://github.com/superlistapp/super_editor/runs/5779443590?check_suite_focus=true#step:5:583)
[584](https://github.com/superlistapp/super_editor/runs/5779443590?check_suite_focus=true#step:5:584)
#4      main.<anonymous closure>.<anonymous closure>.<anonymous closure> (file:///D:/a/super_editor/super_editor/super_editor/test/src/default_editor/upstream_downstream_selection_test.dart:529:9)
[585](https://github.com/superlistapp/super_editor/runs/5779443590?check_suite_focus=true#step:5:585)
<asynchronous suspension>
[586](https://github.com/superlistapp/super_editor/runs/5779443590?check_suite_focus=true#step:5:586)
<asynchronous suspension>
[587](https://github.com/superlistapp/super_editor/runs/5779443590?check_suite_focus=true#step:5:587)
(elided one frame from package:stack_trace)
[588](https://github.com/superlistapp/super_editor/runs/5779443590?check_suite_focus=true#step:5:588)

[589](https://github.com/superlistapp/super_editor/runs/5779443590?check_suite_focus=true#step:5:589)
This was caught by the test expectation on the following line:
[590](https://github.com/superlistapp/super_editor/runs/5779443590?check_suite_focus=true#step:5:590)
  file:///D:/a/super_editor/super_editor/super_editor/test/src/default_editor/upstream_downstream_selection_test.dart line 529
[591](https://github.com/superlistapp/super_editor/runs/5779443590?check_suite_focus=true#step:5:591)
The test description was:
[592](https://github.com/superlistapp/super_editor/runs/5779443590?check_suite_focus=true#step:5:592)
  inserts paragraph after downstream edge
[593](https://github.com/superlistapp/super_editor/runs/5779443590?check_suite_focus=true#step:5:593)
════════════════════════════════════════════════════════════════════════════════════════════════════
[594](https://github.com/superlistapp/super_editor/runs/5779443590?check_suite_focus=true#step:5:594)

[595](https://github.com/superlistapp/super_editor/runs/5779443590?check_suite_focus=true#step:5:595)
00:27 +140 -1: D:\a\super_editor\super_editor\super_editor\test\src\default_editor\upstream_downstream_selection_test.dart: Upstream-downstream block typing at boundary inserts paragraph after downstream edge [E]
[596](https://github.com/superlistapp/super_editor/runs/5779443590?check_suite_focus=true#step:5:596)
  Test failed. See exception logs above.
[597](https://github.com/superlistapp/super_editor/runs/5779443590?check_suite_focus=true#step:5:597)
  The test description was: inserts paragraph after downstream edge`

This is the only exception that is caught for both Windows and Linux, however, both of these tests are satisfied on my local machine. Not sure where to go from here since they are working for me. Could you provide some direction? 

PS: Also, did you leave a question about the movementModifier changes? I got an email about that, but there isn't anything on GitHub. Just wondering if its okay then.

agenthesh:
Sorry @matthew-carroll, but do you mind helping me with why the tests are failing?

[This](https://pastebin.com/UdQR6try) is the only exception that is caught for both Windows and Linux, however, both of these tests are satisfied on my local machine. Not sure where to go from here since they are working for me. Could you provide some direction?

PS: Also, did you leave a question about the movementModifier changes? I got an email about that, but there isn't anything on GitHub. Just wondering if its okay then.

matthew-carroll:
I can't infer much from the failures that you linked, but if the problem is only on Windows and Linux then it reminds me of this issue that I recently had:
https://github.com/flutter/flutter/issues/100033

Can you check if it's the same issue? If it is, you should be able to use the same approach that I did to hack around the issue for right now.

matthew-carroll:
Hi @agenthesh - any update on this ticket?

agenthesh:
Hey @matthew-carroll! Really sorry for the delay, but as for an update, I have none. These are the things I've tried to debug this problem below:

- Run the test on a windows machine. Tests all passed. The test which is throwing an exception was debugged, and the expected behavior is correct.
- Actually verify the behavior in SuperEditor. Created a mutable document with a HR, and then pressed the A key. There are two nodes, and the A character is after the HR at the beginning of the next line. (verified this both on windows and Mac). 
- The issue that you linked does not apply to this test. The test is failing because (for some reason) the number of nodes it finds on the document is one, when in reality it should be two (Horizontal Rule node, and the Paragraph node for the character). Related to this by the way, is the fact that the test which is failing, is not even related to the code I have changed or written. 

- I even ended up installing Linux on my Mac (Arch Linux, running on an M1 Max) to see if there was problem there (side note: Installing linux, flutter on an ARM based Mac is annoying as hell). But there isn’t. The tests all work fine, and the behavior in super_editor is perfect as well.[ I’ve attached flutter doctor and logs from the tests here.](https://drive.google.com/file/d/1mw4QRJeZtjQkyD_YDRNbp7jLd1JCDnx7/view?usp=sharing)

The only reason I can think of is something is going wrong during the CI build. Not exactly sure what could be going wrong however. What's even more confusing is this. The exact same test exists for an upstream edge rather than a downstream edge, and the CI build is accepting that test. I’m honestly really lost cause I’ve made sure that the code is not an issue here.

matthew-carroll:
@agenthesh it looks like the issue must have been a CI flake for some reason. The tests seem to pass now, so I'll go ahead and merge in. Thanks.

matthew-carroll (approved):

super_editor/lib/src/default_editor/document_keyboard_actions.dart
```diff
@@ -299,3 +304,89 @@ ExecutionInstruction moveUpDownLeftAndRightWithArrowKeys({
 
   return didMove ? ExecutionInstruction.haltExecution : ExecutionInstruction.continueExecution;
 }
+
+ExecutionInstruction moveToLineStartOrEndWithCtrlAOrE({
+  required EditContext editContext,
+  required RawKeyEvent keyEvent,
+}) {
+  if (Platform.instance.isMac) {
+    ExecutionInstruction.continueExecution;
```
nilsreichardt, line 313: Is this missing the `return`?

agenthesh: Yeah it is, that's really embarrassing! Thanks for pointing it out!

nilsreichardt: I just saw the comment by @cgestes in https://github.com/superlistapp/super_editor/commit/115b0a5d16fb8f7c48f99ff1ed8d840084cf80fa, so credits to him :D

agenthesh: @matthew-carroll Sorry about this! Do you prefer I create a new PR for this?

PR #500, opened 2022-03-27
IME for desktop (Resolves #499)

### LINKED ISSUE #499: Support IME on desktop
Hardware keyboard handlers prevent `super_editor` from supporting various input modalities, e.g., auto-correction, emoji pickers.

There are also certain characters that don't work with Flutter's hardware keyboard support, for some reason, e.g., #348, #385.

To enable these capabilities, `super_editor` should add support for IME on desktop, similar to its IME support for mobile.

matthew-carroll:
IME for desktop (Resolves #499)

IME support on desktop requires that we use the existing mobile IME integration, and also add our own physical keyboard responses for anything that isn't content input, e.g., backspace, delete, arrow keys.

As soon as possible, we should wrap these behaviors with tests and submit to the Flutter test repo so that Flutter, itself, can't change how these things behave without telling us.


matthew-carroll:
@bk-one @MarcelKaeding - can both of you try the example app in this PR and let me know if you're able to enter the special characters that you've posted about in the past?

MarcelKaeding:
https://user-images.githubusercontent.com/1116227/160363311-99cdc554-bfb8-4221-8c11-fb3763caf176.mp4

@matthew-carroll I can enter the characters now (for example: ^`´) . However, it inserts them twice into the block, one insertion at the cursor position and a second insertion at a seemingly random other place in the block. The same is happening when I hit option+n for example, which should insert ~, or option+u which inserts ¨

matthew-carroll:
I created a minimal text editing deltas demo to determine if the double-entry problem was super_editor, or Flutter. At least part of the issue appears to exist within Flutter.

Using the Flutter-only demo, when I try to enter one of the special characters, it replaces a character instead of inserting a character. Also, the `opt+u` appears in random places.

I filed https://github.com/flutter/flutter/issues/101013 with Flutter

@bk-one if this proves to an issue within Flutter, we'll be back in a Flutter rollout timeline again.

matthew-carroll:
Update: The Flutter team reported that this was fixed. I switched to the `master` branch to try it out. I'm able to insert the special characters as desired. However, if I select multiple characters and then enter a special character, the special character is inserted at the front of the selection, rather than replacing the selection. Therefore, it's still buggy.

matthew-carroll:
@MarcelKaeding can you try this PR again, using Flutter on the `master` branch? Please try typing characters, as well as replacing selections with characters, and any other common interaction you can think of.

MarcelKaeding:
@matthew-carroll the previously problematic symbols ^`´~ work fine now and are correctly set at the cursor position 👍 

One issue I'm seeing though: when you type `^` directly follow by `a` the expected result is to get `â` same for:
`~` + `n` = `ñ`
`^` + `o` = `ô`
`´` + `e` = `é`
`¨`+ `o` = `ö`
and so on.. 
`¨` can be entered by `Option + U`

matthew-carroll:
@MarcelKaeding are you saying that when you type `^` + `a` you're getting `^a` instead of `â`

Also, to make sure that I understand, you're saying that when the OS sees those two characters back-to-back, it should automatically merge them? What if you actually want `^a` instead of `â`?

MarcelKaeding:
Yes exactly, both points are correct.

If you wanted ^a you'd have to enter `^``SPACE``a`

matthew-carroll:
@MarcelKaeding is that behavior tied to a particular type of keyboard, or OS setting? If I type ^a right here in GitHub, nothing happens, nor have I ever seen such character combinations before.

MarcelKaeding:
@matthew-carroll yes, it could also be related to the keyboard layout. for me if I type `^` directly followed by `a`I get `â` in GitHub, and App Notes for example.

Here is a helpful info graphic on the key combinations that should work on a US keyboard to type the accents that are used for example in French and Spanish https://macinfo.us/how-to-type-french-accents-on-a-macbook-pro/

matthew-carroll:
@MarcelKaeding I cross-posted over on the Flutter issue. It looks like there may be some followup questions from the team. Can you answer their questions about the character combination behavior? https://github.com/flutter/flutter/issues/101013

matthew-carroll:
@MarcelKaeding @bk-one - Can you try again with this branch and Flutter on `master`? I have â working in the example editor on my end.

bk-one:
- [ ] Odd behavior on horizontal (and vertical) scrolling: https://www.loom.com/share/2a394ee80f864cfe9730bfdc3e7bc9f6
- [ ] The input works technically, but the visualization isn't right. It's supposed to show the initial modifier key underlined: https://www.loom.com/share/82fde7a85c8d44678bba6eff9632e54c

```
$ flutter doctor
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel master, 2.13.0-0.0.pre.867, on macOS 12.3.1 21E258 darwin-arm, locale en-DE)
[✓] Xcode - develop for iOS and macOS (Xcode 13.3.1)
[✗] Chrome - develop for the web (Cannot find Chrome executable at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome)
    ! Cannot find Chrome. Try setting CHROME_EXECUTABLE to a Chrome executable.
[✓] VS Code (version 1.66.2)
[✓] Connected device (1 available)
```

matthew-carroll:
@bk-one did scrolling horizontally on the trackpad always work for selection? I didn't even know that was an interaction modality. Is there any other interaction that's failing for you on this branch?

As for the underline, I think we'll need to handle that separately. We'll need to revisit text decorations to figure out the right place to impose that underline.

bk-one:
@MarcelKaeding can you please double check the behavior and answer Matts questions?

MarcelKaeding:
I've tested the branch and it works as expected for me 👍

PR #502, opened 2022-03-28
Replaced relative deps with GitHub deps

matthew-carroll:
Replaced relative deps with GitHub deps.

@miguelcmedeiros can you try this PR and see if it solves the relative dependency issue that you're having?

### DIFF
--- a/super_editor/pubspec.yaml
+++ b/super_editor/pubspec.yaml
@@ -12,7 +12,9 @@ dependencies:
     sdk: flutter
 
   attributed_text:
-    path: ../attributed_text
+    git:
+      url: https://github.com/superlistapp/super_editor.git
+      path: attributed_text
   characters: ^1.2.0
   collection: ^1.15.0
   http: ^0.13.1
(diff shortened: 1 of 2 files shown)


super_editor/pubspec.yaml
```diff
@@ -12,7 +12,9 @@ dependencies:
     sdk: flutter
 
   attributed_text:
-    path: ../attributed_text
+    git:
+      url: https://github.com/superlistapp/super_editor.git
```
miguelcmedeiros, line 16: Can't really test it since `flutter packages get` will resolve this dependency to the `main` branch, which keeps complaining about `../attributed_text` not being valid.

Probably it will work but can't say for sure until we merge it

miguelcmedeiros (approved):
LGTM (hopefully mitigates the issue with updating super_editor in projects, until the recently extracted packages get released to pub.dev)

matthew-carroll:
@miguelcmedeiros Ok, I'll merge this and then you can try again.

PR #513, opened 2022-04-01
Fixed an embedded scrolling bug (Resolves #496)

### LINKED ISSUE #496: Extent selection position is incorrect when SuperEditor is placed inside a CustomScrollView
Steps to reproduce:
* In "Sliver Demo Editor", select some text and move the selection extent

Expected: selection is painted from the base position till the extent position (close to cursor) 
Actual: extent position is not correct
ezgif com-gif-maker-10

matthew-carroll:
Fixed an embedded scrolling bug (Resolves #496)

@miguelcmedeiros please let me know if this solves your issues in Superlist, and let me know if the PR looks OK.

Most of the work in the PR is to create a scrolling minimap to help us diagnose scrolling problems in the future. It still needs a bit of work, and we probably need some tools beyond that, but I wanted to unblock you.

There are some failing golden tests due to some kind of timing issue. I'll come back tomorrow to fix those tests and merge in, if this solves your problem.


matthew-carroll:
I removed the golden tests for the scrolling minimap because I couldn't seem to get the minimap connected to a scrollable in the test. I left the minimap code in super_editor and the example app because it seems to work OK in a real app. We should come back and improve the minimap code, and add tests.

@miguelcmedeiros I'm going to merge this so that I can move forward with other work, but you should verify this fix in Superlist ASAP.

PR #518, opened 2022-04-03
Add link support to markdown serialization

### LINKED ISSUE #489: Add support for links to markdown deserialization
The ExampleEditor toolbar includes an option to add links. The links are added correctly to the document however when using serializeDocumentToMarkdown the links aren't included.

Thanks for the excellent package!

GarettCooper:
Resolves #489 

Adds basic link support to markdown serialization. As implemented, links can completely overlap with other attributes, but a partial intersection will break the link into two, such as the following:
`*test [link* text](https://link.com)` would instead be serialized as `*test *[*link*](https://link.com)[ text](https://link.com)` 
I decided to put the PR up anyway with this behaviour since it's consistent(ly broken) with the other markdown styles, which would require a much more significant change to resolve and seemed out of scope of the original issue.

### DIFF
--- a/super_editor/example/lib/demos/demo_markdown_serialization.dart
+++ b/super_editor/example/lib/demos/demo_markdown_serialization.dart
@@ -132,8 +132,19 @@ Document _createInitialDocument() {
       ListItemNode.unordered(
         id: DocumentEditor.createNodeId(),
         text: AttributedText(
-          text: 'This is a 3rd list item',
-        ),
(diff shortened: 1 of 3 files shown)


matthew-carroll (approved):
LGTM - Thanks. Can you also file an issue that describes the specific failure modes that you alluded to in this PR? It would also be great if you could add that issue number in your code/tests where you mention the problem. I'll merge this PR after that minor update.

PR #519, opened 2022-04-05
Switched super_text dependency on attributed_text from GitHub to Pub.dev

matthew-carroll:
Switched `super_text` dependency on attributed_text from GitHub to Pub.dev

Also, this PR includes release 0.1.1 for `attributed_text` because apparently it was using a version of `collection` that's incompatible with Flutter, so I downgraded that dependency within `attributed_text`.


matthew-carroll:
I think the CI errors will disappear when these version changes are merged, so I'm gonna merge and see what happens.

PR #524, opened 2022-04-06
fix toolbar location when there are multiple editors in one screen

### LINKED ISSUE #486: Default toolbar location is incorrect
By default the toolbar is shown in the wrong location.

image

We were able to resolve the issue by changing this code in example_editor.dart from:

    docBox.localToGlobal(docBoundingBox.topLeft, ancestor: context.findRenderObject()),
    docBox.localToGlobal(docBoundingBox.bottomRight, ancestor: context.findRenderObject()),

To:

    docBox.localToGlobal(docBoundingBox.topLeft),
    docBox.localToGlobal(docBoundingBox.bottomRight),

Thanks for the great package!

ThangVuNguyenViet:
Fix default toolbar location when there are multiple editors on one screen (Resolves #486 )

When there are multiple editors on one screen, the toolbar in `example_editor.dart` is not displayed correctly, as other editors except the first editor shows the toolbar based on the position of the first editor. This PR fixes the above issue, as displayed below.

This PR does not affect the `super_editor` components, but rather gives the user a better default implementation in the example file.

The PR is created thanks to hillelcoren. 

MacOS:
image

### DIFF
--- a/super_editor/example/lib/demos/example_editor/example_editor.dart
+++ b/super_editor/example/lib/demos/example_editor/example_editor.dart
@@ -151,8 +151,8 @@ class _ExampleEditorState extends State<ExampleEditor> {
           .getRectForSelection(_composer.selection!.base, _composer.selection!.extent)!;
       final docBox = _docLayoutKey.currentContext!.findRenderObject() as RenderBox;
       final overlayBoundingBox = Rect.fromPoints(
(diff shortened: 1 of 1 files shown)


matthew-carroll (approved):
LGTM - Thanks for the fix, and for the explanation in the PR

PR #533, opened 2022-04-11
Fix Auto-scroll in CustomScrollView, and Android Editor in CustomScrollView (resolve #521)

### LINKED ISSUE #521: [Mobile] Auto-scroll does not work when document is placed inside a CustomScrollView
Steps to reproduce:
* Update "Sliver Demo Editor" so that it works on iOS:
```
```
* Select a paragraph that is located at the bottom of the viewport

**Expected:** keyboard is shown and the document auto-scrolls to show the selection (caret)
**Actual:** document does not auto-scroll and selection is hidden by the keyboard

ThangVuNguyenViet:
Fix auto-scroll when Editor is in a CustomScrollView (Resolves #521 ), and fix scrolling when using Editor in a CustomScrollView with `gestureMode: DocumentGestureMode.android`

I reimplemented the scrolling trigger, which checks for content (which is a handle in this case) that is not visible within the viewport and its boundary. The scrolling could trigger either when the content is above or below the desired viewport, and it jumps to the position based on its trigger position, added with the current scroll offset.

Along the way, I found that the `_ancestorScrollPosition` is not set in `document_gesture_touch_android.dart`, compared to `document_gesture_touch_ios.dart`, which leads to errors when using in CustomScrollView. So I made the change accordingly.

iOS:

Android:

### DIFF
--- a/super_editor/test/src/default_editor/auto_scroll_test.dart
+++ b/super_editor/test/src/default_editor/auto_scroll_test.dart
@@ -17,8 +17,11 @@ void main() {
           var screenHeight = 844.0;
           const screenWidth = 390.0;
 
-          const frameCount = 60;
-          const shrinkPerFrame = 5.0;
+          // TODO: Figure out why changing the [shrinkPerFrame] and/or [frameCount]
+          // could result in the final actual scroll offset is smaller than the expected scroll offset
+          // by a fraction of [shrinkPerFrame]
+          const frameCount = 5;
+          const shrinkPerFrame = 60.0;
 
           // The position should be in the middle bottom of the screen, so that
           // if the document is layed out properly, tapping this position should place a caret to the document.
@@ -50,7 +53,28 @@ void main() {
             screenHeight,
           );
 
-          expect(scrollController.offset, greaterThanOrEqualTo(tapPosition.dy - screenHeight));
+          final handleFinder = find.byType(BlinkingCaret);
+          final handleBox = handleFinder.evaluate().last.findRenderObject() as RenderBox;
+
+          final editorFinder = find.byType(SuperEditor);
+          final editorBox = editorFinder.evaluate().first.findRenderObject() as RenderBox;
+
+          final documentOffset = handleBox.localToGlobal(editorBox.globalToLocal(Offset.zero));
+
+          // Determine the caret's height. Related to the position of the tapped caret and the document
+          const lineHeight = 18;
+
+          // Dy from the SuperEditor to its Scrollable parent
+          const editorOffsetDy = 212.0;
+
+          // DragAutoScrollBoundary.trailing of default_editor in [AndroidDocumentTouchInteractor]
+          const dragAutoScrollBoundary = 54.0;
+
+          // The math was taken from [ensureOffsetIsVisible] in [document_gestures_touch.dart]
+          expect(
+            scrollController.offset,
+            equals((documentOffset.dy + lineHeight) + dragAutoScrollBoundary - screenHeight + editorOffsetDy),
+          );
         });
       });
       group('iOS', () {
@@ -63,8 +87,11 @@ void main() {
           var screenHeight = 844.0;
           const screenWidth = 390.0;
 
-          const frameCount = 60;
-          const shrinkPerFrame = 5.0;
+          // TODO: Figure out why changing the [shrinkPerFrame] and/or [frameCount]
+          // could result in the final actual scroll offset is smaller than the expected scroll offset
+          // by a fraction of [shrinkPerFrame]
+          const frameCount = 5;
+          const shrinkPerFrame = 60.0;
 
           // The position should be in the middle bottom of the screen, so that
           // if the document is layed out properly, tapping this position should place a caret to the document.
@@ -96,7 +123,28 @@ void main() {
             screenHeight,
           );
 
-          expect(scrollController.offset, greaterThanOrEqualTo(tapPosition.dy - screenHeight));
+          final handleFinder = find.byType(BlinkingCaret);
+          final handleBox = handleFinder.evaluate().last.findRenderObject() as RenderBox;
+
+          final editorFinder = find.byType(SuperEditor);
+          final editorBox = editorFinder.evaluate().first.findRenderObject() as RenderBox;
+
+          final documentOffset = handleBox.localToGlobal(editorBox.globalToLocal(Offset.zero));
+
+          // Determine the caret's height. Related to the position of the tapped caret
+          const lineHeight = 18;
+
+          // Dy from the SuperEditor to its Scrollable parent
+          const editorOffsetDy = 212.0;
+
+          // dragAutoScrollBoundary.trailing of default_editor in [IOSDocumentTouchInteractor]
+          const dragAutoScrollBoundary = 54.0;
+
+          // The math was taken from [ensureOffsetIsVisible] in [document_gestures_touch.dart]
+          expect(
+            scrollController.offset,
+            equals((documentOffset.dy + lineHeight) + dragAutoScrollBoundary - screenHeight + editorOffsetDy),
+          );
         });
       });
     });
@@ -121,7 +169,7 @@ Future<double> _shrinkViewportAndEnsureVisibleCaret(
     await tester.pumpAndSettle();
 
     // Ensure visible caret
-    final handleBox = handleFinder.evaluate().last.renderObject as RenderBox;
+    final handleBox = handleFinder.evaluate().last.findRenderObject() as RenderBox;
     final handleOffset = handleBox.localToGlobal(Offset.zero);
 
     expect(handleOffset.dy, lessThanOrEqualTo(height));
--- a/super_editor/test/src/default_editor/auto_scroll_test.dart
+++ b/super_editor/test/src/default_editor/auto_scroll_test.dart
@@ -15,9 +15,6 @@ void main() {
           var screenHeight = 844.0;
           const screenWidth = 390.0;
 
-          // TODO: Figure out why changing the [shrinkPerFrame] and/or [frameCount]
-          // could result in the final actual scroll offset is smaller than the expected scroll offset
-          // by a fraction of [shrinkPerFrame]
           const frameCount = 5;
           const shrinkPerFrame = 60.0;
 
@@ -39,13 +36,16 @@ void main() {
           // Place a caret in the document
           await tester.tapAt(tapPosition);
 
-          screenHeight = await _shrinkViewportAndEnsureVisibleCaret(
+          final unscrolledOffset = await _shrinkViewportAndEnsureVisibleCaret(
             tester: tester,
             frameCount: frameCount,
             shrinkPerFrame: shrinkPerFrame,
             width: screenWidth,
             height: screenHeight,
           );
+
+          // After shrinking, screenHeight is reduced
+          screenHeight = screenHeight - frameCount * shrinkPerFrame;
 
           final handleFinder = find.byType(BlinkingCaret);
           final handleOffset = tester.getTopLeft(handleFinder.last);
@@ -65,9 +65,14 @@ void main() {
           const dragAutoScrollBoundary = 54.0;
 
           // The math was taken from [ensureOffsetIsVisible] in [document_gestures_touch.dart]
+          // The [unscrolledOffset] is subtracted because at the last of the shrinking frame process,
+          // the screenHeight (viewport) is reduced but it doesn't meet the condition to scroll
           expect(
             scrollController.offset,
-            equals((handleToEditorOffset.dy + lineHeight) + dragAutoScrollBoundary - screenHeight + editorOffsetDy),
+            equals((handleToEditorOffset.dy + lineHeight) +
+                dragAutoScrollBoundary -
+                (screenHeight - unscrolledOffset) +
+                editorOffsetDy),
           );
         });
       });
@@ -79,11 +84,8 @@ void main() {
           var screenHeight = 844.0;
           const screenWidth = 390.0;
 
-          // TODO: Figure out why changing the [shrinkPerFrame] and/or [frameCount]
-          // could result in the final actual scroll offset is smaller than the expected scroll offset
-          // by a fraction of [shrinkPerFrame]
-          const frameCount = 5;
-          const shrinkPerFrame = 60.0;
+          const frameCount = 60;
+          const shrinkPerFrame = 5.0;
 
           // Tap offset to select an arbitrary text position in the document.
           final tapPosition = Offset(screenWidth / 2, screenHeight - 1);
@@ -103,13 +105,16 @@ void main() {
           // Place a caret to the document
           await tester.tapAt(tapPosition);
 
-          screenHeight = await _shrinkViewportAndEnsureVisibleCaret(
+          final unscrolledOffset = await _shrinkViewportAndEnsureVisibleCaret(
             tester: tester,
             frameCount: frameCount,
             shrinkPerFrame: shrinkPerFrame,
             width: screenWidth,
             height: screenHeight,
           );
+
+          // After shrinking, screenHeight is reduced
+          screenHeight = screenHeight - frameCount * shrinkPerFrame;
 
           final handleFinder = find.byType(BlinkingCaret);
           final handleOffset = tester.getTopLeft(handleFinder.last);
@@ -129,9 +134,14 @@ void main() {
           const dragAutoScrollBoundary = 54.0;
 
           // The math was taken from [ensureOffsetIsVisible] in [document_gestures_touch.dart]
+          // The [unscrolledOffset] is subtracted because at the last of the shrinking frame process,
+          // the screenHeight (viewport) is reduced but it doesn't meet the condition to scroll
           expect(
             scrollController.offset,
-            equals((handleToEditorOffset.dy + lineHeight) + dragAutoScrollBoundary - screenHeight + editorOffsetDy),
+            equals((handleToEditorOffset.dy + lineHeight) +
+                dragAutoScrollBoundary -
+                (screenHeight - unscrolledOffset) +
+                editorOffsetDy),
           );
         });
       });
@@ -228,7 +238,8 @@ class _SliverTestEditorState extends State<_SliverTestEditor> {
 /// To mimic the keyboard showing behaviour, the window size will be shrinking
 /// for a number of [frameCount], each time it reduces an amount of [shrinkPerFrame]
 ///
-/// Returns the new height, which equals to [height] - [frameCount] * [shrinkPerFrame]
+/// During the shrinking, the editor might not scroll on every frame.
+/// Returns the total scroll offset of continuously not scrolling in those last frames.
 Future<double> _shrinkViewportAndEnsureVisibleCaret({
   required WidgetTester tester,
   required int frameCount,
@@ -237,6 +248,9 @@ Future<double> _shrinkViewportAndEnsureVisibleCaret({
   required double height,
 }) async {
   final handleFinder = find.byType(BlinkingCaret);
+
+  int framesBeforeScroll = 0;
+  double prevHandleOffsetDy = 0;
 
   for (var i = 0; i < frameCount; i++) {
     height -= shrinkPerFrame;
@@ -247,8 +261,15 @@ Future<double> _shrinkViewportAndEnsureVisibleCaret({
     final handleOffset = tester.getBottomLeft(handleFinder.last);
 
     expect(handleOffset.dy, lessThanOrEqualTo(height));
-  }
 
-  // Return the window's height after being shrunk
-  return height;
+    if (prevHandleOffsetDy != handleOffset.dy) {
+      // HandleOffset changed, means the editor scrolled
+      framesBeforeScroll = 0;
+      prevHandleOffsetDy = handleOffset.dy;
+    } else {
+      framesBeforeScroll++;
+    }
+  }
+  // Returns the total scroll offset of continuously not scrolling in the last frames
+  return framesBeforeScroll * shrinkPerFrame;
 }
--- a/super_editor/test/src/default_editor/auto_scroll_test.dart
+++ b/super_editor/test/src/default_editor/auto_scroll_test.dart
@@ -7,158 +7,100 @@ import '../_document_test_tools.dart';
 void main() {
   group('SuperEditor', () {
     group('auto-scroll', () {
-      group('Android', () {
-        testWidgets('maintain visible caret when the viewport is being minimized', (WidgetTester tester) async {
-          final scrollController = ScrollController();
-
-          // Setting initial fake screen size. The height will shrink later.
-          var screenHeight = 844.0;
-          const screenWidth = 390.0;
-
-          const frameCount = 60;
-          const shrinkPerFrame = 5.0;
-
-          // Tap offset to select an arbitrary text position in the document.
-          final tapPosition = Offset(screenWidth / 2, screenHeight - 1);
-
-          tester.binding.window
-            ..physicalSizeTestValue = Size(screenWidth, screenHeight)
-            ..textScaleFactorTestValue = 1.0
-            ..devicePixelRatioTestValue = 1.0;
-
-          final testEditor = _SliverTestEditor(
-            scrollController: scrollController,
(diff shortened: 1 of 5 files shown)


ThangVuNguyenViet:
The PR doesn't include any tests yet, as I couldn't find a way to show the keyboard in widget test. If you find a way to do that, or at least mimic the keyboard showing behavior, please tell me so I can make an update

super_editor/test/src/default_editor/auto_scroll_test.dart
```diff
@@ -0,0 +1,210 @@
...
+        gestureMode == DocumentGestureMode.iOS ? IOSCollapsedHandle : AndroidSelectionHandle,
+      );
+      expect(handleFinder.evaluate(), isNotEmpty);
+    }
+
+    expect(scrollController.offset, greaterThanOrEqualTo(tapDy - height));
+  });
+}
+
+class SliverTestEditor extends StatefulWidget {
```
matthew-carroll, line 72: Is this class intended to be used in other files? If not, it should be declared private.

super_editor/test/src/default_editor/auto_scroll_test.dart
```diff
@@ -0,0 +1,210 @@
...
+        testAutoScroll(
+          'maintain visible caret when the viewport is being minimized',
+          DocumentGestureMode.iOS,
+        );
+      });
+    });
+  });
+}
+
+void testAutoScroll(
```
matthew-carroll, line 24: Is this method supposed to be used in other files? If not, it should be declared private.

super_editor/test/src/default_editor/auto_scroll_test.dart
```diff
@@ -0,0 +1,210 @@
...
+            ),
+          ],
+        ),
+      ),
+      debugShowCheckedModeBanner: false,
+    );
+  }
+
+  Document _createInitialDocument() {
+    return MutableDocument(
```
matthew-carroll, line 168: You should probably move this method into `_document_test_tools.dart` so that we don't end up with a bunch of big document definitions in various test files. You can call it `createExampleDocument()`.

super_editor/test/src/default_editor/auto_scroll_test.dart
```diff
@@ -0,0 +1,210 @@
...
+                inputSource: DocumentInputSource.ime,
+              ),
+            ),
+            SliverList(
+              delegate: SliverChildBuilderDelegate(
+                (BuildContext context, int index) {
+                  return ListTile(
+                    title: Text('$index'),
+                    onTap: () {
+                      ScaffoldMessenger.of(context).showSnackBar(
```
matthew-carroll, line 147: I don't think we need snackbar displays in a test, do we?

super_editor/test/src/default_editor/auto_scroll_test.dart
```diff
@@ -0,0 +1,210 @@
...
+    var height = 844.0;
+    const width = 390.0;
+    const tapDy = 800.0;
+
+    tester.binding.window
+      ..physicalSizeTestValue = Size(width, height)
+      ..textScaleFactorTestValue = 1.0
+      ..devicePixelRatioTestValue = 1.0;
+
+    final testEditor = SliverTestEditor(
```
matthew-carroll, line 40: Is there a reason that this widget is assigned to variable, rather than passing this widget tree directly into `pumpWidget`?

super_editor/test/src/default_editor/auto_scroll_test.dart
```diff
@@ -0,0 +1,210 @@
...
+      scrollController: scrollController,
+      gestureMode: gestureMode,
+    );
+
+    await tester.pumpWidget(testEditor);
+    await tester.pumpAndSettle();
```
matthew-carroll, line 46: Any time that you use a `pumpAndSettle()` you should add a comment about why a single `pump` isn't enough.

super_editor/test/src/default_editor/auto_scroll_test.dart
```diff
@@ -0,0 +1,210 @@
...
+
+    expect(scrollController.offset, 0.0);
```
matthew-carroll, line 48: Please describe why you're expecting results. What does a scroll offset of zero mean? why is it important?

super_editor/test/src/default_editor/auto_scroll_test.dart
```diff
@@ -0,0 +1,210 @@
...
+
+    // Tap at the middle bottom of the screen
```
matthew-carroll, line 50: Why are we tapping at the middle bottom of the screen?

super_editor/test/src/default_editor/auto_scroll_test.dart
```diff
@@ -0,0 +1,210 @@
...
+    await tester.tapAt(const Offset(width / 2, tapDy));
+    await tester.pumpAndSettle();
+
+    // Slowly reduce screen size and pump
+    // To mimic keyboard showing behaviour
+    for (var i = 0; i < 30; i++) {
```
matthew-carroll, line 56: We should avoid magic numbers when we can. The `30` should probably go in a constant called `frameCount`, and the `10` can go into a constant called `shrinkPerFrame`. You can declare those above the for-loop.

super_editor/test/src/default_editor/auto_scroll_test.dart
```diff
@@ -0,0 +1,210 @@
...
+      height -= 10;
+      tester.binding.window.physicalSizeTestValue = Size(width, height);
+      await tester.pumpAndSettle();
+
+      // Maintain visible caret
+      final handleFinder = find.byType(
```
matthew-carroll, line 62: I think you can declare this `Finder` once, outside the loop. I think the `expect` will evaluate the `Finder` each time. Also, `isNotEmpty` isn't the ideal check, is it? Shouldn't you ensure that you find exactly 1 such handle?

BTW, have you ever seen this test fail? Are you sure the test will fail if the handle goes off-screen?

super_editor/test/src/default_editor/auto_scroll_test.dart
```diff
@@ -0,0 +1,210 @@
...
+        gestureMode == DocumentGestureMode.iOS ? IOSCollapsedHandle : AndroidSelectionHandle,
+      );
+      expect(handleFinder.evaluate(), isNotEmpty);
+    }
+
+    expect(scrollController.offset, greaterThanOrEqualTo(tapDy - height));
```
matthew-carroll, line 68: I think you should expect a specific value here. You're controlling the change in height, and the number of frames, so you should be able to calculate an exact offset, right? When you do that, please include a comment that explains the math to get that value, in case we need to change it later.

super_editor/test/src/default_editor/auto_scroll_test.dart
```diff
@@ -0,0 +1,210 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:super_editor/super_editor.dart';
+
+void main() {
+  group('SuperEditor', () {
+    group('auto-scroll', () {
+      group('Android', () {
+        testAutoScroll(
```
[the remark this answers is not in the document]
ThangVuNguyenViet: Do you mean I should remove the `testAutoScroll` function, break that function into smaller private functions and use them directly in the test?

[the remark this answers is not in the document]
matthew-carroll: Yeah, that would work. I have no idea how many smaller functions would make sense. There may only be one, but maybe more. I'd have to think about it.

But the guiding principle is that when reading a given test, you should be able to read that literal test. We don't want our developers to have to compile code in their mind to understand what the test is doing, nor do we want readers to have to jump around to gain a high level understanding of what the test is verifying, and how. We want reading tests to be as effortless as possible.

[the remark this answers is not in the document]
ThangVuNguyenViet: I just pushed an update for this

super_editor/test/src/default_editor/auto_scroll_test.dart
```diff
@@ -0,0 +1,210 @@
...
+}
+
+void testAutoScroll(
+  String description,
+  DocumentGestureMode gestureMode,
+) {
+  testWidgets(description, (WidgetTester tester) async {
+    final scrollController = ScrollController();
+
+    var height = 844.0;
```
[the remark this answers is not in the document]
ThangVuNguyenViet: Knowing the screen size for this test ensures when the document `_SliverTestEditor` is laid out, the editor is visible in the viewport and at the bottom of the screen. I put some comments on the updated commit.

super_editor/test/src/default_editor/auto_scroll_test.dart
ThangVuNguyenViet: You're right. I just pushed an update for this

super_editor/test/src/default_editor/auto_scroll_test.dart
ThangVuNguyenViet: I deleted this method to enhance test readability, based on your recommendation

super_editor/test/src/default_editor/auto_scroll_test.dart
ThangVuNguyenViet: You're right. I just pushed an update for this

super_editor/test/src/default_editor/auto_scroll_test.dart
ThangVuNguyenViet: It was meant to make sure no scrolling happens when the test taps the document but the shrinking has not occurred. It turns out not so important, so I removed it

super_editor/test/src/default_editor/auto_scroll_test.dart
ThangVuNguyenViet: We can tap on a lot of places, but if the document is laid out properly, tapping this position should easily place a caret to the document. By placing a caret at the bottom, the shrinking viewport can easily trigger the scroll.

super_editor/test/src/default_editor/auto_scroll_test.dart
ThangVuNguyenViet: I just pushed an update for this

super_editor/test/src/default_editor/auto_scroll_test.dart
ThangVuNguyenViet: You're right. It won't fail. So I've made an update for that, by determining the position of the caret and comparing it with the height. If it's less than or equal to the height, then it's in the viewport. I just pushed an update for this

super_editor/test/src/default_editor/auto_scroll_test.dart
ThangVuNguyenViet: I spent hours figuring out the math to get the value, but I couldn't. The closest I can get is described in the next update I just pushed. Its problem is when changing the value of `frameCount` and/or the `shrinkPerFrame`, the final actual scroll offset is smaller than the expected scroll offset by a fraction of `shrinkPerFrame`.

For example, when running with 
```dart
const frameCount = 10;
const shrinkPerFrame = 30.0;
```

it results in 
```
Expected: <365.5600280761719>
  Actual: <335.5600280761719>
```

Which is a fraction of 30.

I put the comment in the next update. I hope we can get some insight from that.

super_editor/test/src/default_editor/auto_scroll_test.dart
```diff
@@ -0,0 +1,263 @@
...
+            ),
+          ],
+        ),
+      ),
+      debugShowCheckedModeBanner: false,
+    );
+  }
+}
+
+/// Slowly reduce screen size and pump
```
matthew-carroll, line 238: This doc comment has a few issues. 

First, each line is far too short. You should use a similar width as compared to other doc comments. 

Second, there should be a concise sentence to start, followed by other details beneath that. Please check the Bounty Hunters video that discusses doc comments. 

Third, language tense is very important in code. Instead of "reduce screen size" you should say "reduces screen size".

That said, I have some comments on what I think this method should do, which will impact the doc comment.

super_editor/test/src/default_editor/auto_scroll_test.dart
```diff
@@ -0,0 +1,263 @@
...
+/// To mimic keyboard showing behaviour
+/// Return the reduced height
+Future<double> _shrinkViewportAndEnsureVisibleCaret(
+  WidgetTester tester,
+  int frameCount,
```
matthew-carroll, line 243: When you have more than 2 or 3 parameters, especially when they have similar types, you should consider using named parameters.

super_editor/test/src/default_editor/auto_scroll_test.dart
```diff
@@ -0,0 +1,263 @@
...
+  double shrinkPerFrame,
+  double width,
```
matthew-carroll, line 245: I don't think you need to pass `width` - this function can check the current `physicalSizeTestValue` and then re-use that value.

Same for `height`, too. I don't think you need to pass that in.

super_editor/test/src/default_editor/auto_scroll_test.dart
```diff
@@ -0,0 +1,263 @@
...
+      ),
+      debugShowCheckedModeBanner: false,
+    );
+  }
+}
+
+/// Slowly reduce screen size and pump
+/// To mimic keyboard showing behaviour
+/// Return the reduced height
+Future<double> _shrinkViewportAndEnsureVisibleCaret(
```
matthew-carroll, line 241: Given that what we're doing here is simulating keyboard expansion, why not name this method the same? `simulateKeyboardExpansion()` - then we're in a position to use this in general.

super_editor/test/src/default_editor/auto_scroll_test.dart
```diff
@@ -0,0 +1,263 @@
...
+  double height,
+) async {
+  final handleFinder = find.byType(BlinkingCaret);
+
+  for (var i = 0; i < frameCount; i++) {
+    height -= shrinkPerFrame;
+    tester.binding.window.physicalSizeTestValue = Size(width, height);
+    await tester.pumpAndSettle();
+
+    // Ensure visible caret
```
matthew-carroll, line 255: Instead of implementing a specific condition here, how about taking in a callback that you invoke every time in the loop so that clients can do whatever they want on each frame?

The callback should be async, just in case that's needed: `Future<void> Function()`

super_editor/test/src/default_editor/auto_scroll_test.dart
```diff
@@ -0,0 +1,263 @@
...
+    final handleBox = handleFinder.evaluate().last.findRenderObject() as RenderBox;
+    final handleOffset = handleBox.localToGlobal(Offset.zero);
+
+    expect(handleOffset.dy, lessThanOrEqualTo(height));
+  }
+
+  return height;
```
matthew-carroll, line 262: Do we really need to return the `height`? It's also a good idea to specify which height you're returning. Is this the content height?

super_editor/test/src/default_editor/auto_scroll_test.dart
```diff
@@ -0,0 +1,263 @@
...
+          // Dy from the SuperEditor to its Scrollable parent
+          const editorOffsetDy = 212.0;
+
+          // DragAutoScrollBoundary.trailing of default_editor in [AndroidDocumentTouchInteractor]
+          const dragAutoScrollBoundary = 54.0;
+
+          // The math was taken from [ensureOffsetIsVisible] in [document_gestures_touch.dart]
+          expect(
+            scrollController.offset,
+            equals((documentOffset.dy + lineHeight) + dragAutoScrollBoundary - screenHeight + editorOffsetDy),
```
matthew-carroll, line 76: I suppose that if the calculation is this complicated, then for now let's return to just expecting a specific number for the scroll offset.

super_editor/test/src/default_editor/auto_scroll_test.dart
```diff
@@ -0,0 +1,263 @@
...
+        testWidgets('maintain visible caret when the viewport is being minimized', (WidgetTester tester) async {
+          final scrollController = ScrollController();
+
+          // Setting initial fake screen size. The height would shrink later on.
+          // The size should be set properly so that when the _SliverTestEditor is laid out,
+          // the document is within bottom of the viewport.
+          var screenHeight = 844.0;
+          const screenWidth = 390.0;
+
+          // TODO: Figure out why changing the [shrinkPerFrame] and/or [frameCount]
```
matthew-carroll, line 20: It's definitely a good idea to figure this out. If we don't understand what's happening in our own tests then we can't be sure that we're testing what we think we are.

super_editor/test/src/default_editor/auto_scroll_test.dart
```diff
@@ -0,0 +1,263 @@
...
+          // could result in the final actual scroll offset is smaller than the expected scroll offset
+          // by a fraction of [shrinkPerFrame]
+          const frameCount = 5;
+          const shrinkPerFrame = 60.0;
+
+          // The position should be in the middle bottom of the screen, so that
```
matthew-carroll, line 26: You can probably tap almost anywhere to place a caret. This comment still doesn't say anything about why this point is tapped. Is it arbitrary? If so, all you need to say is:

`// Tap offset to select an arbitrary text position in the document.`

Remember, saying more isn't always better. The important part is to communicate your intention for others. If you can say that in a few words, then that's ideal.

Also, be careful with the terms "offset" and "position" in super_editor. We deal with "text positions", which are indices in text. We also deal with (x,y) "offsets". Whenever possible, let's try to name things accordingly.

super_editor/test/src/default_editor/auto_scroll_test.dart
```diff
@@ -0,0 +1,263 @@
...
+import '../_document_test_tools.dart';
+
+void main() {
+  group('SuperEditor', () {
+    group('auto-scroll', () {
+      group('Android', () {
+        testWidgets('maintain visible caret when the viewport is being minimized', (WidgetTester tester) async {
+          final scrollController = ScrollController();
+
+          // Setting initial fake screen size. The height would shrink later on.
```
matthew-carroll, line 14: Similar to another comment below, we're still missing the intent here. These numbers look very specific, where did they come from? Are these specific values important?

You say "The height would shrink later". I think what you want is "The height will shrink later".

I don't understand the part that begins with "The size should be..." but I think you can probably delete that part, anyway.

matthew-carroll (commented):
Thanks for the updates. I provided another round of comments for the test implementation.

Once we work through the test details, I'll take a look at the implementation. After that, we can merge.

super_editor/test/src/default_editor/auto_scroll_test.dart
ThangVuNguyenViet: The numbers don't have to be that specific, but that depends on the tapping position and how the document is laid out.

For instance, if setting the screen size is `Size(2400, 2400)`, the screen will be too wide and tapping wouldn't place a caret in the document.

I think it's a similar to the way you use `fakeScreenSize` in `toolbar_position_delegate_test.dart`

super_editor/test/src/default_editor/auto_scroll_test.dart
```diff
@@ -0,0 +1,263 @@
...
+            screenHeight,
+          );
+
+          final handleFinder = find.byType(BlinkingCaret);
+          final handleBox = handleFinder.evaluate().last.findRenderObject() as RenderBox;
+
+          final editorFinder = find.byType(SuperEditor);
+          final editorBox = editorFinder.evaluate().first.findRenderObject() as RenderBox;
+
+          final documentOffset = handleBox.localToGlobal(editorBox.globalToLocal(Offset.zero));
```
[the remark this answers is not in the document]
ThangVuNguyenViet: The `documentOffset` is the same thing as the `documentOffset` in `ensureOffsetIsVisible` in `document_gestures_touch.dart`, from which I took the name, except it's the top left instead of the bottom left (that's why I need to add the `lineHeight` after). 
I think `caretOffset` would be clearer in this case

super_editor/test/src/default_editor/auto_scroll_test.dart
```diff
@@ -0,0 +1,263 @@
...
+            ..textScaleFactorTestValue = 1.0
+            ..devicePixelRatioTestValue = 1.0;
+
+          final testEditor = _SliverTestEditor(
+            scrollController: scrollController,
+            gestureMode: DocumentGestureMode.android,
+          );
+
+          await tester.pumpWidget(testEditor);
+          await tester.pumpAndSettle();
```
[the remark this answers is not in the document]
ThangVuNguyenViet: Yea you're right. Those are redundant. I'm gonna remove it

super_editor/test/src/default_editor/auto_scroll_test.dart
ThangVuNguyenViet: The number for `scrollController.offset` is `365.5600280761719`, which I believe will confuse other developers more than ever. I suggest jumping back to the old expectation, which verifies the scrolled offset is greater or equal to the Y-axis distance of the caret offset to the current `screenHeight`

`expect(scrollController.offset, greaterThanOrEqualTo((caretOffset.dy + lineHeight) - screenHeight))`

What do you think about this though?

super_editor/test/src/default_editor/auto_scroll_test.dart
ThangVuNguyenViet: It's the content height, or screen height after being shrunk. I'll put a comment for this

super_editor/test/src/default_editor/auto_scroll_test.dart
```diff
@@ -0,0 +1,263 @@
...
+
+          // Place a caret to the document
```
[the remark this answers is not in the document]
ThangVuNguyenViet: It's embarrassing for making such a mistake. I'm gonna be using Grammarly for every comment I write later on.

super_editor/test/src/default_editor/auto_scroll_test.dart
ThangVuNguyenViet: It is, but the problem is I couldn't figure out why changing these numbers while remaining their product could change the scrolling offset. The math is complicated in this part.

As you suggested, if we were to remove the complicated calculation in the expectation part, the TODO should also be removed.

ThangVuNguyenViet:
I just refactored some of my implementations. Previously it was unnecessarily complicated.

super_editor/test/src/default_editor/auto_scroll_test.dart
ThangVuNguyenViet: I've made an update to this

super_editor/test/src/default_editor/auto_scroll_test.dart
ThangVuNguyenViet: Thanks. I've taken a more careful look at the doc and made a better update for this

matthew-carroll:
Thanks for the updates. I'll review as soon as I can. I've got to finish up a couple other things first.

ThangVuNguyenViet:
Sure. No worries

ThangVuNguyenViet:
Hi @matthew-carroll , I just found another bug when there is some content above the `Scrollable` which will scroll the document up instead of down. I've made the changes in the implementation, test and in the PR description

ThangVuNguyenViet:
Just want to inform you Matt, that the current implementation hasn't been done yet. I'll push more update

ThangVuNguyenViet:
@matthew-carroll I'm happy with my code. You can review it now.

I couldn't write comments for the implementation though, as my implementation is based on yours. The math is complicated, and I couldn't figure out how you came up with that. If you can explain your math, then I can explain my update based on your math and put them in the comments

matthew-carroll:
@ThangVuNguyenViet let's try pair programming with your implementation. I'll send you a calendar invite. Let's have you open your IDE and we'll work through the code, together.

ThangVuNguyenViet:
@matthew-carroll as you said, it shouldn't be as complicated as previous implementations. I did overly compute the numbers.

I just reimplemented the whole thing and it looks much simpler. I believe this is how things should be. Please take a look at that

ThangVuNguyenViet:
I just refactored the implementation and make the test to verify the caret position to be at the boundary, within a small margin of error

matthew-carroll (approved):
LGTM

PR #539, opened 2022-04-29
Enable markdown to have empty lines

impure:
This isn't officially supported in markdown, but it is very annoying not being able to have empty lines in a document.

This will only be done when the `allowBlankLines` flag is enabled. It does not change how parsing is done normally.


matthew-carroll:
@impure can you please file a corresponding ticket that shows a standard example of why this situation is annoying, and what you're proposing we do about it? I'd like to inspect the generic proposal before going through all the lines of code.

matthew-carroll:
I haven't heard anything back in a couple weeks, so I'm going to close this PR. If you believe that we should support non-standard Markdown behavior, I'd like to have that conversation in an issue ticket, first. We need to make sure that non-standard behavior won't hurt our future API potential.

PR #545, opened 2022-05-03
Fix removeComposingAttributions on AttributedTextEditingController (Resolves #540)

### LINKED ISSUE #540: [BUG] AttributedTextEditingController removes all composing attributions when it should only remove some
`AttributedTextEditingController` has a method called `removeComposingAttribution()`, which removes a `Set` of `Attribution`s from the controller's "composing attributions".

The implementation of this method inadvertently removes all composing attributions every time.

A line inside the method reads:

`_composingAttributions.removeWhere((attribution) => _composingAttributions.contains(attribution));`

When it should read:

`_composingAttributions.removeWhere((attribution) => attributions.contains(attribution));`

Fix the bug and write appropriate tests to ensure this method works as expected in the future.

angelosilvestre:
`AttributedTextEditingController` was always removing all composing attributions when calling `removeComposingAttributions`, regardless of the `attributions` passed as arguments

(Resolves #540)

### DIFF
--- a/super_editor/test/src/infrastructure/super_textfield/attributed_text_editing_controller_test.dart
+++ b/super_editor/test/src/infrastructure/super_textfield/attributed_text_editing_controller_test.dart
@@ -699,5 +699,62 @@ void main() {
       expect(text1.hasListeners, false);
       expect(text2.hasListeners, true);
     });
+
+    group('remove composing attributions', () {
+      test('should remove the given attributions', () {
+        final controller = AttributedTextEditingController(
+          text: AttributedText(text: 'my text'),
+        );
+        controller.addComposingAttributions(
+          {boldAttribution, italicsAttribution, underlineAttribution},
+        );
+        controller.removeComposingAttributions(
+            {boldAttribution, underlineAttribution});
+        expect(controller.composingAttributions.length, 1);
+        expect(
+            controller.composingAttributions.contains(boldAttribution), false,
+            reason: 'expected to remove boldAttribution');
+        expect(controller.composingAttributions.contains(underlineAttribution),
+            false,
+            reason: 'expected to remove underlineAttribution');
+      });
+
+      test(
+          'should not remove any attributions that are not passed as arguments',
+          () {
+        final controller = AttributedTextEditingController(
+          text: AttributedText(text: 'my text'),
+        );
+        controller.addComposingAttributions(
+          {boldAttribution, italicsAttribution, underlineAttribution},
+        );
+        controller.removeComposingAttributions({boldAttribution});
+        expect(controller.composingAttributions.length, 2);
+        expect(
+            controller.composingAttributions.contains(italicsAttribution), true,
+            reason: 'expected to NOT remove italicsAttribution');
+        expect(controller.composingAttributions.contains(underlineAttribution),
+            true,
+            reason: 'expected to NOT remove underlineAttribution');
+      });
+
+      test(
+          'should not modify any attributions if the controller does not contain any of the attributions passed as arguments',
+          () {
+        final controller = AttributedTextEditingController(
+          text: AttributedText(text: 'my text'),
+        );
+        controller.addComposingAttributions(
+          {boldAttribution, italicsAttribution},
+        );
+        controller.removeComposingAttributions({underlineAttribution});
+        expect(controller.composingAttributions.length, 2);
+        expect(
+            controller.composingAttributions.contains(italicsAttribution), true,
+            reason: 'expected to NOT remove italicsAttribution');
+        expect(controller.composingAttributions.contains(boldAttribution), true,
+            reason: 'expected to NOT remove boldAttribution');
+      });
+    });
   });
 }
--- a/super_editor/test/src/infrastructure/super_textfield/attributed_text_editing_controller_test.dart
+++ b/super_editor/test/src/infrastructure/super_textfield/attributed_text_editing_controller_test.dart
@@ -700,60 +700,41 @@ void main() {
       expect(text2.hasListeners, true);
     });
 
-    group('remove composing attributions', () {
-      test('should remove the given attributions', () {
-        final controller = AttributedTextEditingController(
-          text: AttributedText(text: 'my text'),
-        );
-        controller.addComposingAttributions(
-          {boldAttribution, italicsAttribution, underlineAttribution},
-        );
-        controller.removeComposingAttributions(
-            {boldAttribution, underlineAttribution});
-        expect(controller.composingAttributions.length, 1);
-        expect(
-            controller.composingAttributions.contains(boldAttribution), false,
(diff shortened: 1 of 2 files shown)


matthew-carroll:
FYI @angelosilvestre - please add me as a reviewer on a PR when you submit one, or at least tag me in a comment if GitHub won't let you add me as a reviewer. GitHub didn't tell me about this PR, so I just happened to find it right now as I was looking through the issues.

super_editor/test/src/infrastructure/super_textfield/attributed_text_editing_controller_test.dart
```diff
@@ -699,5 +699,62 @@ void main() {
...
+
+    group('remove composing attributions', () {
+      test('should remove the given attributions', () {
+        final controller = AttributedTextEditingController(
+          text: AttributedText(text: 'my text'),
+        );
+        controller.addComposingAttributions(
+          {boldAttribution, italicsAttribution, underlineAttribution},
+        );
+        controller.removeComposingAttributions(
```
matthew-carroll, line 711: I have a recommendation for readability of tests like this. These tests are structure like: setup > behavior under test > expectations. It helps to place a blank line between those groups of behaviors.

In this test, for example, you'd add a blank line above `removeComposingAttributions()`, and then you'd add another blank line above all the `expect` lines.

angelosilvestre:
@matthew-carroll OK, I'll do that in the next PR's

super_editor/test/src/infrastructure/super_textfield/attributed_text_editing_controller_test.dart
```diff
@@ -699,5 +699,62 @@ void main() {
...
+            {boldAttribution, underlineAttribution});
+        expect(controller.composingAttributions.length, 1);
+        expect(
+            controller.composingAttributions.contains(boldAttribution), false,
+            reason: 'expected to remove boldAttribution');
```
matthew-carroll, line 716: Sometimes providing a `reason` is very helpful. However, most of the time, it' obvious from the `expect()` call what's happening. For maintenance and quick reading, it's best to leave the `reason` out, unless there's an obscure detail that requires the extra clarification.

super_editor/test/src/infrastructure/super_textfield/attributed_text_editing_controller_test.dart
```diff
@@ -699,5 +699,62 @@ void main() {
       expect(text1.hasListeners, false);
       expect(text2.hasListeners, true);
     });
+
+    group('remove composing attributions', () {
+      test('should remove the given attributions', () {
+        final controller = AttributedTextEditingController(
+          text: AttributedText(text: 'my text'),
+        );
+        controller.addComposingAttributions(
```
matthew-carroll, line 708: It's probably a good idea to add an `expect` immediately after `addComposingAttributions()` that confirms that the attributions were actually added. If, for some reason, the attributions didn't make it into the controller, then this test would pass, even if the behavior under test was broken.

super_editor/test/src/infrastructure/super_textfield/attributed_text_editing_controller_test.dart
```diff
@@ -699,5 +699,62 @@ void main() {
       expect(text1.hasListeners, false);
       expect(text2.hasListeners, true);
     });
+
+    group('remove composing attributions', () {
```
matthew-carroll, line 703: Looking at the other groups of tests, with names like "user", "replace", and "delete", this group looks overly specific at this level. I think what you want is two groups, instead of one:

```
group('composing attributions', () {
  group('removal', () {
    //...
  });
});
```

super_editor/test/src/infrastructure/super_textfield/attributed_text_editing_controller_test.dart
```diff
@@ -699,5 +699,62 @@ void main() {
...
+        controller.addComposingAttributions(
+          {boldAttribution, italicsAttribution, underlineAttribution},
+        );
+        controller.removeComposingAttributions(
+            {boldAttribution, underlineAttribution});
+        expect(controller.composingAttributions.length, 1);
+        expect(
+            controller.composingAttributions.contains(boldAttribution), false,
+            reason: 'expected to remove boldAttribution');
+        expect(controller.composingAttributions.contains(underlineAttribution),
```
matthew-carroll, line 717: This `expect` is unnecessary because you already checked that the set has 1 element, and you confirmed that it's bold. So, unless Dart's `Set` class is broken, it's impossible that underline would still be in there.

super_editor/test/src/infrastructure/super_textfield/attributed_text_editing_controller_test.dart
```diff
@@ -699,5 +699,62 @@ void main() {
...
+            false,
+            reason: 'expected to remove underlineAttribution');
+      });
+
+      test(
+          'should not remove any attributions that are not passed as arguments',
```
matthew-carroll, line 723: What is this test verifying that isn't already verified in the first test?

super_editor/test/src/infrastructure/super_textfield/attributed_text_editing_controller_test.dart
```diff
@@ -699,5 +699,62 @@ void main() {
...
+        expect(
+            controller.composingAttributions.contains(italicsAttribution), true,
+            reason: 'expected to NOT remove italicsAttribution');
+        expect(controller.composingAttributions.contains(underlineAttribution),
+            true,
+            reason: 'expected to NOT remove underlineAttribution');
+      });
+
+      test(
+          'should not modify any attributions if the controller does not contain any of the attributions passed as arguments',
```
matthew-carroll, line 742: Let's see if we can make this name a bit more succinct and clear. How about: `does nothing when it doesn't have the given composing attributions`

matthew-carroll (commented):
The fix looks good. The tests are generally in the right direction. I left a few comments where I think we can improve them.

super_editor/test/src/infrastructure/super_textfield/attributed_text_editing_controller_test.dart
angelosilvestre: Adding an expect just for the length of the `composingAttributions` is enough or do you think that each attribution needs to be checked individually?

matthew-carroll: checking the length is probably sufficient in this case

super_editor/test/src/infrastructure/super_textfield/attributed_text_editing_controller_test.dart
angelosilvestre: Yeah, this test ended up beeing redundant

matthew-carroll (approved):
LGTM. Thanks.

PR #547, opened 2022-05-07
 Updates for Superlist tokenization work

matthew-carroll:
Moved all AttributedTextEditingController extension methods into the class, also exposed the 'innerController' from ImeTextEditingController because Superlist token behavior needs access to the inner controller.


matthew-carroll:
I'm going to merge this without review so that I can use the updates in Superlist work

PR #550, opened 2022-05-10
Add link support to markdown deserialization

GarettCooper:
Updates the markdown deserialization visitor with a new case to handle links.


matthew-carroll (approved):
LGTM. Thanks for the addition.

PR #558, opened 2022-05-13
Collapse selection when the ESC key is pressed (Resolves #350)

### LINKED ISSUE #350: Pressing escape key replaces selected text with a space character
Steps to reproduce:
* Run super_editor's demo app on macOS
* Select any piece of text (single character, multiple characters, multiple words, etc.)
* Press escape key

Expected: either collapse selection or do nothing (seen both behaviors in different tools)
Actual: selected text is replaced by a space `" "`
https://user-images.githubusercontent.com/31278849/141998731-3add3127-ebaa-4a51-b68c-1cb9643e7e05.mov

Note: on web nothing happens when pressing escape.

ThangVuNguyenViet:
Collapse selection when the ESC key is pressed (Resolves #350 )

On the desktop, when the ESC key is pressed on a hardware keyboard, a SuperEditor should respond by collapsing an expanded selection, or doing nothing if the selection is already collapsed.

Appropriate tests are included.

MacOS: 

https://user-images.githubusercontent.com/39979207/168209618-ebfccfed-d659-40cf-bd09-0e1c1ed6600a.mov


ThangVuNguyenViet:
I wonder if we should change the current test grouping

super_editor/lib/src/default_editor/document_keyboard_actions.dart
```diff
@@ -133,6 +133,11 @@ ExecutionInstruction anyCharacterOrDestructiveKeyToDeleteSelection({
     return ExecutionInstruction.continueExecution;
   }
 
+  // Do nothing if ESC is pressed because this should collapse the selection
+  if (keyEvent.logicalKey == LogicalKeyboardKey.escape) {
```
matthew-carroll, line 137: Why is this needed? I looked lower down at this handler and I don't see why ESCAPE would trigger this handler.

super_editor/lib/src/default_editor/document_keyboard_actions.dart
```diff
@@ -370,3 +375,23 @@ ExecutionInstruction deleteWordWithAltBksp({
...
+  if (keyEvent.logicalKey != LogicalKeyboardKey.escape) {
+    return ExecutionInstruction.continueExecution;
+  }
+  if (editContext.composer.selection == null || editContext.composer.selection!.isCollapsed) {
+    return ExecutionInstruction.continueExecution;
+  }
+
+  bool didCollapse = false;
+
+  didCollapse = editContext.commonOps.collapseSelection();
```
matthew-carroll, line 395: Why might the collapse not happen?

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -598,6 +598,7 @@ final defaultKeyboardActions = <DocumentKeyboardAction>[
   anyCharacterOrDestructiveKeyToDeleteSelection,
   anyCharacterToInsertInParagraph,
   anyCharacterToInsertInTextContent,
+  collapseSelectionWhenEscIsPressed,
```
matthew-carroll, line 601: The order of this list generally goes from specific to generic. `anyCharacterToInsertInTextContent` should always be the last thing in this list. I would move this handler just below `cutWhenCmdXIsPressed`.

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
```diff
@@ -512,6 +512,120 @@ void main() {
...
+                  nodePosition: TextNodePosition(offset: 19),
+                ),
+              ),
+            ),
+          );
+
+          Platform.setTestInstance(null);
+        });
+      });
+      group('key pressed without selection', () {
```
matthew-carroll, line 574: It doesn't look like this group is adding anything useful. I think you can just name the test "does nothing when escape is pressed if the selection is collapsed"

super_editor/lib/src/default_editor/document_keyboard_actions.dart
ThangVuNguyenViet: It would. It'd trigger adding a character, which is this weird looking one below
image

matthew-carroll: Ok, can you add a comment above that if-statement to explain why we're checking? My guess is that Flutter shouldn't be reporting a character, but it is.

super_editor/lib/src/default_editor/document_keyboard_actions.dart
ThangVuNguyenViet: In this case, it will always happen, since the condition for the `collapseSelection` to return false is the selection is null, which we've already checked. But I think we're still calling an operation returning a boolean value, so we still should use that value, in case `collapseSelection` changes.

If you still think we should remove the `didCollapse` variable, I'll do it.

super_editor/lib/src/default_editor/super_editor.dart
ThangVuNguyenViet: Sure. I'll change

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
ThangVuNguyenViet: Yeah, I was thinking of making a group which is in contrast to the one above.  But you're right, no group until there're more than 1 test in it

super_editor/lib/src/default_editor/document_keyboard_actions.dart
matthew-carroll: Yeah, I think you can remove the `didCollapse`. At this point you've already inspected what matters. Just run the operation and then say it's handled.

super_editor/lib/src/default_editor/document_keyboard_actions.dart
```diff
@@ -139,6 +139,13 @@ ExecutionInstruction anyCharacterOrDestructiveKeyToDeleteSelection({
     return ExecutionInstruction.continueExecution;
   }
 
+  // Specifically exclude situations where ESC key is pressed because ESC
```
matthew-carroll, line 142: The last sentence is all you need for this handler. This handler doesn't actually know what ESC is supposed to do, so stating what ESC is supposed to do is something that will eventually go wrong.

I think all you need is: "Flutter reports a character for ESC, but we don't want to add a character for ESC. Ignore this key press"

ThangVuNguyenViet: Okay. I just made an update

matthew-carroll (approved):
LGTM - thanks

matthew-carroll:
@ThangVuNguyenViet it looks like the same failing tests that I found in my SuperText PR are also happening here. So maybe its a Flutter 3 issue? I suppose you might as well see if you can solve those here and we can merge in those fixes, too.

ThangVuNguyenViet:
@matthew-carroll I just added updates for the caret to fix the auto-scroll test

matthew-carroll (approved):
LGTM - Thanks

PR #559, opened 2022-05-13
Fix auto-scroll tests after flutter 3 upgrade (Resolves #555 )

### LINKED ISSUE #555: Fix broken keyboard auto-scroll test after Flutter 3 upgrade
The upgrade to Flutter 3 caused a mysterious change to the expected scroll position within the keyboard auto-scroll tests.

I imagine that this difference could be caused by a change in the Ahem font, a pixel density change, or some other minor visual change that impacts our test.

Find what caused the breakage, fix the tests, and attempt to update the tests so that a similar breakage doesn't happen in the future.

The tests live in `auto_scroll_test.dart`.

ThangVuNguyenViet:
Fix auto-scroll tests after flutter 3 upgrade (Resolves #555  ) 

After the flutter 3 upgrade, for some reason, the caret size is changed, thus getting the top offset added with known text size is no longer the bottom offset of the caret. 

Wrapping the caret with the SizedBox should make its `RenderBox` reports the size. This also removes the existing TODO.

Migrate the `textScaleFactorTestValue` API because the old one was deprecated.

### DIFF
--- a/super_text/lib/src/caret.dart
+++ b/super_text/lib/src/caret.dart
@@ -173,7 +173,6 @@ class BlinkingCaretState extends State<BlinkingCaret> with SingleTickerProviderS
 
   @override
   Widget build(BuildContext context) {
-    // Passing the size so its [RenderBox] reports the size
     return CustomPaint(
       size: Size(widget.width, widget.caretHeight ?? 0),
       painter: _CursorPainter(
--- a/super_text/lib/src/caret.dart
+++ b/super_text/lib/src/caret.dart
@@ -173,16 +173,21 @@ class BlinkingCaretState extends State<BlinkingCaret> with SingleTickerProviderS
 
   @override
   Widget build(BuildContext context) {
-    return CustomPaint(
-      painter: _CursorPainter(
-        blinkController: _caretBlinkController,
-        caretHeight: widget.caretHeight,
-        caretOffset: widget.caretOffset,
-        width: widget.width,
-        borderRadius: widget.borderRadius,
-        caretColor: widget.color,
-        isTextEmpty: widget.isTextEmpty,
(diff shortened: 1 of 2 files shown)


super_text/lib/src/caret.dart
```diff
@@ -173,16 +173,21 @@ class BlinkingCaretState extends State<BlinkingCaret> with SingleTickerProviderS
...
-        width: widget.width,
-        borderRadius: widget.borderRadius,
-        caretColor: widget.color,
-        isTextEmpty: widget.isTextEmpty,
-        showCaret: widget.showCaret,
+    // Wrap the caret with [SizedBox] so its [RenderBox] reports the size
+    return SizedBox(
+      height: widget.caretHeight,
+      width: widget.width,
+      child: CustomPaint(
```
matthew-carroll, line 180: I think you can pass a `size` directly to the `CustomPaint`

super_text/lib/src/caret.dart
```diff
@@ -173,21 +173,18 @@ class BlinkingCaretState extends State<BlinkingCaret> with SingleTickerProviderS
...
-          blinkController: _caretBlinkController,
-          caretHeight: widget.caretHeight,
-          caretOffset: widget.caretOffset,
-          width: widget.width,
-          borderRadius: widget.borderRadius,
-          caretColor: widget.color,
-          isTextEmpty: widget.isTextEmpty,
-          showCaret: widget.showCaret,
-        ),
+    // Passing the size so its [RenderBox] reports the size
```
matthew-carroll, line 176: I don't think you need this comment. It's clear why we're passing a size.

matthew-carroll (approved):
LGTM. Thanks for tracking down that change, and updating the text scale factor deprecation.

PR #562, opened 2022-05-15
Moving flutter_lints to dev_dependencies in all relevant packages

jmatth:
A bunch of packages in this repo currently have `flutter_lints: ^1.0.0` in their `dependencies`. This breaks version solving for any downstream packages that want to update to the new version `2.0.0` of `flutter_lints`:

```
Because rich_clipboard_example depends on super_editor_markdown from git which depends on flutter_lints ^1.0.0, flutter_lints ^1.0.0 is required.
So, because rich_clipboard_example depends on flutter_lints ^2.0.0, version solving failed.
```

This PR fixes the issue by moving all the `flutter_lints` dependencies in this repo to the `dev_dependencies` section. I intentionally didn't update `flutter_lints` to `^2.0.0` while I was at it because doing so produces enough analyzer warnings that it should be handled in a dedicated PR.

### DIFF
--- a/attributed_text/pubspec.yaml
+++ b/attributed_text/pubspec.yaml
@@ -9,10 +9,9 @@ environment:
 dependencies:
   characters: ^1.2.0
   collection: ^1.15.0
-  flutter_lints: ^1.0.0
   logging: ^1.0.1
   test: ^1.19.4
 
 dev_dependencies:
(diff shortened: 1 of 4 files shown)


matthew-carroll (approved):
LGTM

PR #563, opened 2022-05-15
Auto link URLs (Resolves #406)

### LINKED ISSUE #406: Auto link URLs in Sample Editor
As a user, if I paste an URL into the editor, I expect that URL to be automatically converted (attributed) to be a link. This should be added to the sample editor.

ThangVuNguyenViet:
Auto link URLs (Resolves #406 )

Recognizing URLs in the document and converting them into links when typing and pasting

For example: 
- When the handle is at the last position in the text 'https://flutter.dev/' (could usually be from the user typing the text), pressing a space key or an enter will convert that into a link.
- When the user pastes a block of text 'some text https://flutter.dev/ some text', the part 'https://flutter.dev/' will be converted into a link.

To verify a `Uri` parsed from a text, that `Uri` needs to `hasScheme` and `hasAuthority`

MacOS: 

https://user-images.githubusercontent.com/39979207/168480121-f7122be8-a609-4ea4-aa70-1cbf0921efa1.mov

iOS:

https://user-images.githubusercontent.com/39979207/168480378-07da102c-8d2c-47e5-a89c-34ab244b7b65.mp4

### DIFF
--- a/super_editor/lib/src/default_editor/common_editor_operations.dart
+++ b/super_editor/lib/src/default_editor/common_editor_operations.dart
@@ -2256,6 +2256,12 @@ class CommonEditorOperations {
       attributedText = documentNode.text;
     }
 
+    // Specifically exclude this case because this leads to getting
+    // the previous position having an offset of -1, which is invalid
+    if (textNodeCurrPosition.offset == 0) {
+      return;
+    }
+
     final text = attributedText.text;
 
     // Position which is at the previous offset of the current selection.
--- a/super_editor/lib/src/default_editor/common_editor_operations.dart
+++ b/super_editor/lib/src/default_editor/common_editor_operations.dart
@@ -2229,47 +2229,34 @@ class CommonEditorOperations {
     );
   }
 
-  /// Get the word from the previous offset of the current selection. Convert the word to
-  /// a link if it is a url
+  /// Get the word from [position]. Convert the word to a link if it is a url
   ///
-  /// Does nothing if previous word is already a link
-  void convertPreviousUrlToLink(DocumentComposer composer, DocumentEditor editor) {
+  /// Does nothing if the word is already a link
+  void turnWordAtPositionToLink(TextNodePosition position) {
     if (composer.selection == null) {
+      return;
+    }
+
+    // It is invalid to have a position with negative offset
+    if (position.offset == -1) {
       return;
     }
 
     final documentNode = editor.document.getNodeById(composer.selection!.extent.nodeId);
 
-    if (documentNode is! TextNode && documentNode is! ParagraphNode) {
-      return;
-    }
-
-    // Position of the current selection
-    late TextNodePosition textNodeCurrPosition;
     late AttributedText attributedText;
-
     if (documentNode is ParagraphNode) {
-      textNodeCurrPosition = composer.selection!.extent.nodePosition as TextNodePosition;
       attributedText = documentNode.text;
     } else if (documentNode is TextNode) {
-      textNodeCurrPosition = composer.selection!.extent.nodePosition as TextNodePosition;
       attributedText = documentNode.text;
     }
-
-    // Specifically exclude this case because this leads to getting
-    // the previous position having an offset of -1, which is invalid
-    if (textNodeCurrPosition.offset == 0) {
-      return;
-    }
-
     final text = attributedText.text;
 
     // Position which is at the previous offset of the current selection.
     // This should also be the end position of the previous word
-    final textNodePrevPosition = textNodeCurrPosition.copyWith(offset: textNodeCurrPosition.offset - 1);
-    final attributionsAtPreviousWord = attributedText.getAllAttributionsAt(textNodePrevPosition.offset);
+    final attributions = attributedText.getAllAttributionsAt(position.offset);
 
-    final hasLinkAttribute = attributionsAtPreviousWord.firstWhereOrNull((attr) => attr is LinkAttribution) != null;
+    final hasLinkAttribute = attributions.firstWhereOrNull((attr) => attr is LinkAttribution) != null;
     if (hasLinkAttribute) {
       // Previous word has [LinkAttribute]. Do nothing
       return;
@@ -2277,7 +2264,7 @@ class CommonEditorOperations {
 
     final textSelection = expandPositionToWord(
       text: text,
-      textPosition: textNodePrevPosition,
+      textPosition: position,
     );
     final word = getTextFromTextSelection(text, textSelection);
 
@@ -2288,7 +2275,7 @@ class CommonEditorOperations {
 
       final prevWordDocumentSelection = DocumentSelection.extentFromDocumentPosition(
         documentPosition: composer.selection!.extent.copyWith(
-          nodePosition: textNodePrevPosition.copyWith(offset: textNodePrevPosition.offset - word.length),
+          nodePosition: position.copyWith(offset: position.offset - word.length),
         ),
         extentOffset: word.length,
       );
--- a/super_editor/lib/src/default_editor/common_editor_operations.dart
+++ b/super_editor/lib/src/default_editor/common_editor_operations.dart
@@ -2344,16 +2344,21 @@ class _PasteEditorCommand implements EditorCommand {
       if (!hasLinkAttribute) {
         // Attributions at paste offset doesn't have [LinkAttribute].
         // Add [LinkAttribute] to each url in the text if existed
-        splitContent.first.forEachWord(_pastePosition, ((word, documentSelection) {
+        for (final selectionWordEntry in splitContent.first.wordWithSelections(_pastePosition).entries) {
+          final documentSelection = selectionWordEntry.key;
+          final word = selectionWordEntry.value;
+
           final link = Uri.tryParse(word);
 
           if (link != null && link.hasScheme && link.hasAuthority) {
             // Valid url. Apply [LinkAttribution] to the url
             final linkAttribution = LinkAttribution(url: link);
-            AddTextAttributionsCommand(documentSelection: documentSelection, attributions: {linkAttribution})
-                .execute(document, transaction);
+            AddTextAttributionsCommand(
+              documentSelection: documentSelection,
+              attributions: {linkAttribution},
+            ).execute(document, transaction);
           }
-        }));
+        }
       }
 
       // At this point in the paste process, the document selection
--- a/super_editor/lib/src/default_editor/common_editor_operations.dart
+++ b/super_editor/lib/src/default_editor/common_editor_operations.dart
@@ -2276,7 +2276,7 @@ class CommonEditorOperations {
     final word = getTextFromTextSelection(text, textSelection);
 
     final link = Uri.tryParse(word);
-    if (link != null && link.hasScheme) {
+    if (link != null && link.hasScheme && link.hasAuthority) {
       // Valid url. Apply [LinkAttribution] to the url
       final linkAttribution = LinkAttribution(url: link);
 
@@ -2355,6 +2355,7 @@ class _PasteEditorCommand implements EditorCommand {
 
       // Check for url in the pasted text and apply [LinkAttribution] appropriately
       final hasLinkAttribute = attributionsAtPasteOffset.firstWhereOrNull((attr) => attr is LinkAttribution) != null;
+      print('hasLinkAttribute: $hasLinkAttribute');
       if (!hasLinkAttribute) {
         // Attributions at paste offset doesn't have [LinkAttribute].
         // Add [LinkAttribute] to each url in the text if existed
@@ -2363,7 +2364,7 @@ class _PasteEditorCommand implements EditorCommand {
           (word, documentSelection) {
             final link = Uri.tryParse(word);
 
-            if (link != null && link.hasScheme) {
+            if (link != null && link.hasScheme && link.hasAuthority) {
               // Valid url. Apply [LinkAttribution] to the url
               final linkAttribution = LinkAttribution(url: link);
               AddTextAttributionsCommand(documentSelection: documentSelection, attributions: {linkAttribution})
--- a/super_editor/lib/src/default_editor/common_editor_operations.dart
+++ b/super_editor/lib/src/default_editor/common_editor_operations.dart
@@ -2233,7 +2233,7 @@ class CommonEditorOperations {
   /// a link if it is a url
   ///
   /// Does nothing if previous word is already a link
-  void convertPreviousWordToLinkIfIsUrl(DocumentComposer composer, DocumentEditor editor) {
+  void convertPreviousWordToLinkIfItIsUrl(DocumentComposer composer, DocumentEditor editor) {
     if (composer.selection == null) {
       return;
     }
--- a/super_editor/lib/src/default_editor/common_editor_operations.dart
+++ b/super_editor/lib/src/default_editor/common_editor_operations.dart
@@ -2233,7 +2233,7 @@ class CommonEditorOperations {
   /// a link if it is a url
   ///
   /// Does nothing if previous word is already a link
-  void convertPreviousWordToLinkIfItIsUrl(DocumentComposer composer, DocumentEditor editor) {
+  void convertPreviousUrlToLink(DocumentComposer composer, DocumentEditor editor) {
     if (composer.selection == null) {
       return;
     }
--- a/super_editor/lib/src/default_editor/common_editor_operations.dart
+++ b/super_editor/lib/src/default_editor/common_editor_operations.dart
@@ -2286,6 +2286,26 @@ class _PasteEditorCommand implements EditorCommand {
         attributions: attributionsAtPasteOffset,
       ).execute(document, transaction);
 
+      // Check for url in the pasted text and apply [LinkAttribution] appropriately
+      final hasLinkAttribute = attributionsAtPasteOffset.firstWhereOrNull((attr) => attr is LinkAttribution) != null;
+      if (!hasLinkAttribute) {
+        // Attributions at paste offset doesn't have [LinkAttribute].
+        // Add [LinkAttribute] to each url in the text if existed
+        _forEachWordInText(
+          splitContent.first,
+          (word, documentSelection) {
+            final link = Uri.tryParse(word);
+
+            if (link != null && link.hasScheme) {
+              // Valid url. Apply [LinkAttribution] to the url's [documentSelection]
+              final linkAttribution = LinkAttribution(url: link);
+              AddTextAttributionsCommand(documentSelection: documentSelection, attributions: {linkAttribution})
+                  .execute(document, transaction);
+            }
+          },
+        );
+      }
+
       // At this point in the paste process, the document selection
       // position is at the end of the text that was just pasted.
       newSelectionPosition = DocumentPosition(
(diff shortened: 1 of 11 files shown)


ThangVuNguyenViet:
@matthew-carroll work's still in progress. There're 2 things: 

- Tests for the behavior of adding space will convert the previous word into a link are added, but only for the keyboard, not for the IME. I'm wondering whether we should make the same tests for the IME, or we should figure out how to group them together as 1 test
- Tests for copy and pasting are not added, as it seems like `Clipboard` doesn't trigger `setData`. Looks like we need to mock the Flutter platform channel, but I'm not sure if that'll do the trick.

Do you have any idea on these?

ThangVuNguyenViet:
Sorry, I found that you've already resolved the issue with Clipboard already

ThangVuNguyenViet:
@matthew-carroll I've done with all the tests. You can review this PR now.

In the `super_desktop_textfield_keyboard_test.dart` file, some tests related to Clipboards that you commented they need to be widget tests, while I found them okay to be just normal tests. I've made the change accordingly. I'll revert the commit if you find them necessary to be widget test.

matthew-carroll:
@ThangVuNguyenViet I see a few changes from yesterday. Is this PR actually ready for review?

ThangVuNguyenViet:
Yea it is. I couldn't think of any better way, at least at the moment I'm writing this comment.

super_editor/lib/src/default_editor/paragraph.dart
```diff
@@ -360,6 +360,10 @@ ExecutionInstruction anyCharacterToInsertInParagraph({
     editContext.commonOps.convertParagraphByPatternMatching(
       editContext.composer.selection!.extent.nodeId,
     );
+
+    // Check for the word before the space. If that is a url, make that a link
```
matthew-carroll, line 364: Similar to the IME issue, you're conflating very different responsibilities. This handler's job is simply to interpret incoming keyboard signals and then tell the domain behavior to do whatever it wants to do. In this case, you've added some of that domain knowledge to the handler, itself.

Looking at `convertParagraphByPatternMatching()`, that probably shouldn't be the job of this handler, either, but at least it's a bit more general. Any number of paragraph conversions might take place in there.

How about you add a new `commonOps` method called `tokenizeUpstreamWord((editContext.composer.selection!.extent.nodePosition as TextNodePosition).offset)`? Then, that method can maintain all possible tokenizations of the previous word.

super_editor/lib/src/default_editor/text.dart
```diff
@@ -244,6 +244,18 @@ class TextNodePosition extends TextPosition implements NodePosition {
 
   @override
   int get hashCode => super.hashCode ^ super.offset.hashCode;
+
+  /// Creates a new [TextNodePosition] based on the current position, with the
+  /// provided parameters overridden.
+  TextNodePosition copyWith({
```
matthew-carroll, line 250: Can you move this up above `hashCode` and `==`? I prefer to keep those at the bottom of classes.

super_editor/lib/src/default_editor/text.dart
```diff
@@ -1126,6 +1138,12 @@ ExecutionInstruction anyCharacterToInsertInTextContent({
 
   final didInsertCharacter = editContext.commonOps.insertCharacter(character);
 
+  if (didInsertCharacter && character == ' ') {
```
matthew-carroll, line 1141: Same comment here about keyboard handlers

super_editor/lib/src/default_editor/text_tools.dart
```diff
@@ -62,7 +62,7 @@ TextSelection expandPositionToWord({
 
   int start = min(textPosition.offset, text.length - 1);
   int end = min(textPosition.offset, text.length - 1);
-  while (start > 0 && text[start] != ' ') {
+  while (start > 0 && text[start - 1] != ' ') {
```
matthew-carroll, line 65: Why is this changed?

super_editor/lib/src/default_editor/text_tools.dart
```diff
@@ -155,3 +155,28 @@ TextDirection getParagraphDirection(String text) {
...
+    textSelections.add(textSelection);
+
+    offset += textSelection.end - textSelection.start + 1;
+  }
+  return textSelections;
+}
+
+/// Returns a string from a given [text] in a [textSelection] range
+String getTextFromTextSelection(String text, TextSelection textSelection) {
+  return text.substring(textSelection.start, textSelection.end);
```
matthew-carroll, line 181: I think `TextSelection` already has a method for this called `textInside`

super_editor/lib/src/default_editor/text_tools.dart
```diff
@@ -155,3 +155,28 @@ TextDirection getParagraphDirection(String text) {
...
   }
 }
+
+/// Returns list of [TextSelection] on each word in text
+List<TextSelection> getTextSelectionsForEachWord(String text) {
+  final List<TextSelection> textSelections = [];
+  var offset = 0;
+
+  while (offset < text.length) {
+    if (text[offset] == ' ') {
```
matthew-carroll, line 165: Don't we already have existing behavior that identifies word boundaries?

super_editor/test/src/default_editor/common_editor_operations_test.dart
```diff
@@ -158,5 +160,143 @@ void main() {
         expect(composer.selection!.extent.nodePosition, const TextNodePosition(offset: 0));
       });
     });
+    group("pasting", () {
+      test("it converts url in the pasted text into a link", () async {
+        // Note: We need to ensure initialized because we access the Clipboard.
+        TestWidgetsFlutterBinding.ensureInitialized();
```
matthew-carroll, line 166: The only reason that this line of code works in a regular Dart `test()` is because we're importing `flutter_test.dart`. But from an API standpoint, the use of `test()` suggests that we could run this test in a regular Dart test, but that's not true. So anything that touches Flutter should be defined within a `testWidgets()` instead.

super_editor/test/src/default_editor/common_editor_operations_test.dart
```diff
@@ -158,5 +160,143 @@ void main() {
...
+          ParagraphNode(
+            id: 'paragraph',
+            text: AttributedText(
+              text: 'This  a link',
+            ),
+          ),
+        ]);
+        final editor = DocumentEditor(document: document);
+
+        // Select block of text containing a url
```
matthew-carroll, line 178: You're not really "selecting a block of text", you're placing the caret in the middle of a block of text.

super_editor/test/src/default_editor/common_editor_operations_test.dart
```diff
@@ -158,5 +160,143 @@ void main() {
...
+        // the incoming clipboard call.
+        SystemChannels.platform.setMockMethodCallHandler((call) async {
+          if (call.method == 'Clipboard.getData') {
+            return {
+              'text': 'text: https://flutter.dev is',
+            };
+          }
+        });
+
+        commonOps.paste();
```
matthew-carroll, line 204: Notice how tiny this line looks compared to everything else in this test. It would be easy to scroll right past it. And yet this line is the entire behavior that you're testing. At a minimum, you should add a comment above this line that makes clear that this is the behavior that we're here to test.

super_editor/test/src/default_editor/common_editor_operations_test.dart
```diff
@@ -158,5 +160,143 @@ void main() {
...
+
+        WidgetsBinding.instance.addPostFrameCallback((timeStamp) {
```
matthew-carroll, line 206: Can you run `tester.pumpAndSettle()` instead of scheduling a frame callback? Related to that, I think that we should probably go ahead and render the document in a `SuperEditor`, rather than try to orchestrate a test that avoids the render tree.

super_editor/test/src/default_editor/common_editor_operations_test.dart
```diff
@@ -158,5 +160,143 @@ void main() {
...
+                  SpanMarker(attribution: linkAttribution, offset: 11, markerType: SpanMarkerType.start),
+                  SpanMarker(attribution: linkAttribution, offset: 30, markerType: SpanMarkerType.end),
+                ],
+              ),
+            ),
+          ),
+        ]);
+        final editor = DocumentEditor(document: document);
+
+        // Place the selection within the link attributed span's range
```
matthew-carroll, line 255: Consider reading this code. Wouldn't it be nice to see a direct representation of where the caret is placed?

```
Place caret at "This text: http://flutter.d|ev is..."
```

super_editor/test/src/default_editor/document_input_ime_test.dart
```diff
@@ -73,6 +73,142 @@ void main() {
       });
     });
 
+    group('inserting a space character', () {
+      test('it turns the previous url into a link', () {
```
matthew-carroll, line 77: I think what you mean is "preceding" rather than "previous", but the whole test name should probably be replaced with "automatically converts a URL to a link"

super_editor/test/src/default_editor/document_input_ime_test.dart
```diff
@@ -73,6 +73,142 @@ void main() {
...
+        final document = MutableDocument(nodes: [
+          ParagraphNode(
+            id: "1",
+            text: AttributedText(text: 'This text: https://flutter.devis a link'),
```
matthew-carroll, line 81: Is ".devis" a typo? Also, if this test passes, should it pass right now? Is "flutter.devis" a legitimate URL?

super_editor/test/src/default_editor/document_input_ime_test.dart
```diff
@@ -73,6 +73,142 @@ void main() {
...
+            selection: TextSelection.collapsed(offset: 30),
+            composing: TextRange(start: -1, end: -1),
+            oldText: 'This text: https://flutter.devis a link',
+          ),
+        ]);
+
+        final paragraphNode = document.nodes.first as ParagraphNode;
+
+        // The handler should insert a space
+        expect(paragraphNode.text.text, 'This text: https://flutter.dev is a link');
```
matthew-carroll, line 118: This should be failing right now. Notice that your earlier text had "flutter.devis" and now it says ".dev is". Why is this passing?

super_editor/test/src/default_editor/document_input_ime_test.dart
```diff
@@ -73,6 +73,142 @@ void main() {
...
+        // Adding [LinkAttribution] to a position that already has it
+        // could cause spans mismatching, which potentially leads to errors.
+        // This test prevents that regression
+        final linkAttribution = LinkAttribution(url: Uri.parse('https://flutter.dev'));
+
+        final document = MutableDocument(nodes: [
+          ParagraphNode(
+            id: "1",
+            text: AttributedText(
+              text: 'This text: https://flutter.devis a link',
```
matthew-carroll, line 146: Another typo. Please check all your strings for typos like this.

super_editor/test/src/default_editor/document_input_ime_test.dart
```diff
@@ -73,6 +73,142 @@ void main() {
...
+          {
+            AttributionSpan(
+              attribution: LinkAttribution(url: Uri.parse('https://flutter.dev')),
+              start: 11,
+              end: 29,
+            )
+          },
+        );
+      });
+      test('it does NOT turn the previous url into a link if that is already a link', () {
```
matthew-carroll, line 136: This test name can be reduced to "does nothing to an existing link"

super_editor/test/src/default_editor/text_test.dart
```diff
@@ -301,6 +301,138 @@ void main() {
         );
       });
     });
+
+    group('Inserting a space character', () {
+      test('it turns the previous url into a link', () {
```
matthew-carroll, line 306: Replace with "automatically converts a URL into a link"

super_editor/test/src/default_editor/text_test.dart
```diff
@@ -301,6 +301,138 @@ void main() {
...
+          {
+            AttributionSpan(
+              attribution: LinkAttribution(url: Uri.parse('https://flutter.dev')),
+              start: 11,
+              end: 29,
+            )
+          },
+        );
+      });
+      test('it does NOT turn the previous url into a link if that is already a link', () {
```
matthew-carroll, line 364: Please add a space above this `test()` declaration.

Also, please rename this test, similar to the other one I commented on.

super_editor/test/src/infrastructure/super_textfield/desktop/super_desktop_textfield_keyboard_test.dart
```diff
@@ -278,10 +278,9 @@ void main() {
 
       group('paste shortcut', () {
         group('Mac', () {
-          testWidgets('cmd+v pastes clipboard text', (tester) async {
+          test('cmd+v pastes clipboard text', () async {
```
matthew-carroll, line 281: You should convert these back. Any test that touches Flutter should be a `testWidgets()` call.

matthew-carroll (commented):
Here's my first round of comments

super_editor/test/src/default_editor/document_input_ime_test.dart
ThangVuNguyenViet: I did that on purpose. I'm going to insert a space later on after the "v", which should convert the "https://flutter.dev" into a link

super_editor/test/src/default_editor/document_input_ime_test.dart
ThangVuNguyenViet: It shouldn't fail, because I've just inserted a space after the "v"

super_editor/test/src/default_editor/document_input_ime_test.dart
ThangVuNguyenViet: For the same reason above. The test will insert a space later on after the "v".

super_editor/test/src/infrastructure/super_textfield/desktop/super_desktop_textfield_keyboard_test.dart
ThangVuNguyenViet: Okay, I will. But maybe we should change the comment to "// Note: this is a widget test because it touches Flutter."

super_editor/lib/src/default_editor/common_editor_operations.dart
```diff
@@ -2228,6 +2228,66 @@ class CommonEditorOperations {
...
+    }
+    final text = attributedText.text;
+
+    // Position which is at the previous offset of the current selection.
+    // This should also be the end position of the previous word
+    final attributions = attributedText.getAllAttributionsAt(position.offset);
+
+    final hasLinkAttribute = attributions.firstWhereOrNull((attr) => attr is LinkAttribution) != null;
+    if (hasLinkAttribute) {
+      // Previous word has [LinkAttribute]. Do nothing
```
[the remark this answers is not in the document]
ThangVuNguyenViet: This is my mistake of not fully refactoring. All "previous" in this context should be changed to current

super_editor/lib/src/default_editor/text_tools.dart
ThangVuNguyenViet: Previously it returned a space before the word, like " SuperEditor"

super_editor/test/src/infrastructure/super_textfield/desktop/super_desktop_textfield_keyboard_test.dart
matthew-carroll: In general, a note is fine, however, we'll end up repeating that note dozens or hundreds of times. I think the answer to this issue is education, instead.

super_editor/lib/src/default_editor/common_editor_operations.dart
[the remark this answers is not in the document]
matthew-carroll: "Current" still isn't great. Any reference to time can be very confusing. What's "current" versus "previous" versus "next" from the standpoint of code? I'd try to find a more direct way of saying what you think is important.

super_editor/lib/src/default_editor/text_tools.dart
matthew-carroll: I don't understand your explanation, please elaborate. Was the editor broken before? Why wasn't this triggering failing tests or unintended behavior? Seeing a change like this is very scary.

super_editor/test/src/default_editor/document_input_ime_test.dart
matthew-carroll: Ok. In cases like this the test name should reflect what you're doing in the test, and there should be a comment above any line that's likely to confuse the reader, like this one. As it is now, the test name doesn't really tell us anything about what's being tested, and there's no indication that the thing that looks like a typo was done on purpose.

super_editor/lib/src/default_editor/text_tools.dart
ThangVuNguyenViet: This change doesn't trigger any failing tests because this function was previously only used by `selectSurroundingWord` in `common_editor_operations.dart`, which wasn't used anywhere.

Previously, for example, when I plugged `expandPositionToWord(text: " SuperEditor is great", textPosition: TextPosition(offset: 2))`, it returns " SuperEditor", (notice the space at the start). Now it returns "SuperEditor"

super_editor/lib/src/default_editor/common_editor_operations.dart
```diff
@@ -2286,6 +2346,27 @@ class _PasteEditorCommand implements EditorCommand {
         attributions: attributionsAtPasteOffset,
       ).execute(document, transaction);
 
+      // Check for url in the pasted text and apply [LinkAttribution] appropriately
+      final hasLinkAttribute = attributionsAtPasteOffset.firstWhereOrNull((attr) => attr is LinkAttribution) != null;
+      print('hasLinkAttribute: $hasLinkAttribute');
+      if (!hasLinkAttribute) {
+        // Attributions at paste offset doesn't have [LinkAttribute].
+        // Add [LinkAttribute] to each url in the text if existed
+        _forEachWordInText(
```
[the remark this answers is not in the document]
ThangVuNguyenViet: The callback in this method has 2 parameters: the `word` and its `documentSelection`. Converting this method to return an `Iterable` will require creating a wrapper class for `String` and `DocumentSelection`, and I don't think we should do that. 

I suggest creating an extension on String, which takes in the `documentPosition` as the starting position of the `content` and return each `word` and its `documentSelection` 
For example, `content.onEachWord(documentPosition, (word, documentSelection) {});`

BTW, the `_forEachWordInText` doesn't take in `documentPosition` because it's using the `_pastePosition`

super_editor/lib/src/default_editor/text_tools.dart
matthew-carroll: Ok, we can move forward with that, then. Please be sure to add one or more tests that lock down this method's behavior, and please consider the edge-cases when you do that.

super_editor/test/src/default_editor/common_editor_operations_test.dart
ThangVuNguyenViet: Sure. Do we need to render the document in a `SuperEditor` in other tests too?

ThangVuNguyenViet:
This PR is closed and will be replaced by 2 PRs, one for encoding URLs on pasting and the other for encoding URLs on typing

PR #564, opened 2022-05-15
Fix caret misplace when resizing the app window (Resolves #538)

### LINKED ISSUE #538: Caret gets misplaced when resizing (width) the app window
Steps to reproduce:
* Place the editor with some content on the sides (e.g. a sidebar on the right side of the editor)
* Place the caret somewhere in the document
* Resize the width of the app window

Expected: caret position is updated to match the current selection
Actual: 

https://user-images.githubusercontent.com/31278849/164749483-7606cd10-282d-4454-85a5-042e5bcb88e1.mov

angelosilvestre:
Fixes caret misplaced when resizing the app window. Resolves https://github.com/superlistapp/super_editor/issues/538

The issue only mentions desktop, but I have been able to also reproduce the problem when `gestureMode` is `android` or `iOS`, rotating the phone

The problem occurs because the `caretOffset` is computed inside the `BlinkingTextCaret` build method and the `BlinkingTextCaret` is not rebuilt when screen size changes, causing the `BlinkingCaret` to maintain the old offset and draw in the wrong position. I modified the BlinkingTextCaret to respond to  `didChangeMetrics` and rebuild itselt

In Android and iOS `didChangeMetrics` was already implemented, but it was not calling `_updateHandlesAfterSelectionOrLayoutChange` to reposition the caret


angelosilvestre:
@matthew-carroll could you please review this PR?

super_editor/test/src/default_editor/caret_test.dart
```diff
@@ -0,0 +1,279 @@
...
+import 'package:super_editor/super_editor.dart';
+
+void main() { 
+  //position at doc|ument
+  const caretPosition = DocumentPosition(nodeId: '1', nodePosition: TextNodePosition(offset: 46));
+
+  group("SuperEditor", () {
+    group('window resizing', () {
+      const screenSizeBigger = Size(1000.0, 400.0);
+      const screenSizeSmaller = Size(250.0, 400.0);
```
matthew-carroll, line 12: Let's add a blank line beneath these constants to help quickly scan through the test suite.

super_editor/test/src/default_editor/caret_test.dart
```diff
@@ -0,0 +1,279 @@
...
+
+  group("SuperEditor", () {
+    group('window resizing', () {
+      const screenSizeBigger = Size(1000.0, 400.0);
+      const screenSizeSmaller = Size(250.0, 400.0);  
+      testWidgets('updates caret offset when selected position is displayed in a line below', (WidgetTester tester) async {          
+        tester.binding.window
+          ..devicePixelRatioTestValue = 1.0
+          ..platformDispatcher.textScaleFactorTestValue = 1.0
+          ..physicalSizeTestValue = screenSizeBigger;
```
matthew-carroll, line 17: Let's add a blank line between the test binding setup and the actual test code, to make it easier to read.

super_editor/test/src/default_editor/caret_test.dart
```diff
@@ -0,0 +1,279 @@
...
+        await tester.pumpWidget(
+          _createTestApp(
+            gestureMode:  DocumentGestureMode.mouse, 
+            initialPosition: caretPosition,
+          ),
+        );
+        await tester.pumpAndSettle();            
+        BlinkingCaret blinkingCaret = tester.widget<BlinkingCaret>(find.byType(BlinkingCaret));
+        final initialCaretOffset = blinkingCaret.caretOffset ?? Offset.zero;            
+        expect(initialCaretOffset.dx.floor(), 306);
```
matthew-carroll, line 27: Can you add a comment that explains what you're checking here? Also, any time that magic numbers are used, please mention where those came from. Any magic number might need to change in the future, and the next developer needs to know how to find the new number.

super_editor/test/src/default_editor/caret_test.dart
```diff
@@ -0,0 +1,279 @@
...
+          final docKey = GlobalKey();
+          await tester.pumpWidget(
+            _createTestApp(
+              gestureMode: DocumentGestureMode.android, 
+              docKey: docKey,              
+            ),
+          );           
+          await tester.pumpAndSettle();              
+
+          final tapOffset = _getOffsetForPosition(docKey, caretPosition);
```
matthew-carroll, line 93: It's helpful to add a comment before simulating an action that mentions what you're trying to do. For example, I'm not entirely sure what you're doing here. It looks like you're tapping at the caret, but if there's a caret, then the document already has a selection at that location, right?

super_editor/test/src/default_editor/caret_test.dart
```diff
@@ -0,0 +1,279 @@
...
+          await tester.tapAt(tapOffset);        
+          await tester.pumpAndSettle();  
+
+          Finder caretFinder = find.byType(BlinkingCaret);
+          final initialCaretTopLeft = tester.getTopLeft(caretFinder.last);
```
matthew-carroll, line 98: It looks like you can pass the `Finder` directly into `getTopLeft` and avoid the variable.

matthew-carroll (commented):
Thanks for working on this. I left some comments in the test suite aimed at improving readability.

super_editor/test/src/default_editor/caret_test.dart
```diff
@@ -0,0 +1,279 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:super_editor/super_editor.dart';
+
+void main() { 
+  //position at doc|ument
```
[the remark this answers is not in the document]
angelosilvestre: How about "Position at doc|ument. When the screen is resized this word will move between lines" ?

super_editor/test/src/default_editor/caret_test.dart
angelosilvestre: I will change to simulate a tap in the specified position in all tests instead of setting the initial selection. I was setting the initial selection on mouse `gestureMode` and simulating a tap in Android and iOS

super_editor/test/src/default_editor/caret_test.dart
angelosilvestre: I ran a sample app with the same document that is in the test, clicked at the word document and printed the caret offset. I added this to the test to check if that the caret always get the same offset

Do you think I shouldn't add this to the test?

super_editor/test/src/default_editor/caret_test.dart
```diff
@@ -0,0 +1,312 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:super_editor/src/default_editor/document_gestures_touch_android.dart';
+import 'package:super_editor/src/default_editor/document_gestures_touch_ios.dart';
+import 'package:super_editor/super_editor.dart';
+
+void main() { 
+  // position at doc|ument. When the screen is resized this word will move between lines
```
matthew-carroll, line 8: I recommend reading this from the perspective of someone who opens this file and doesn't know what's being tested. Would this comment make sense? Would it be useful? Can we do better? Also, if something changes in this test suite and this text position needs to be changed, will the developer know where they should move it?

How about this?

```
// We're testing the automatic movement of the caret when the available space changes. This
// text position sits at a location that should move to a different line when the available space
// is reduced.
```

Also, are these properties applicable to every test in this file? If not, should we move them into the group where they are used?

super_editor/test/src/default_editor/caret_test.dart
```diff
@@ -0,0 +1,312 @@
...
+  const textPosition = TextPosition(offset: 46);
+  final tapPosition = DocumentPosition(nodeId: '1', nodePosition: TextNodePosition(offset: textPosition.offset));  
+
+  group("SuperEditor", () {
+    group('window resizing', () {
+      const screenSizeBigger = Size(1000.0, 400.0);
+      const screenSizeSmaller = Size(250.0, 400.0);
+
+      testWidgets('updates caret offset when selected position is displayed in a line below', (WidgetTester tester) async {
```
matthew-carroll, line 17: I'm not sure what this test name means. "displayed a line below" what? Do you mean "moves caret to next line when available width contracts"?

super_editor/test/src/default_editor/caret_test.dart
```diff
@@ -0,0 +1,312 @@
...
+        final docKey = GlobalKey();
+        await tester.pumpWidget(
+          _createTestApp(
+            gestureMode:  DocumentGestureMode.mouse,
+            docKey: docKey,
+          ),
+        );
+        await tester.pumpAndSettle();
+
+        // simulate a tap at doc|ument
```
matthew-carroll, line 32: You are simulating a tap, but why? I think what you want here is:

```
// Place caret at a position that will move to the next line when the width contracts
```

super_editor/test/src/default_editor/caret_test.dart
```diff
@@ -0,0 +1,312 @@
...
+        final tapOffset = _getOffsetForPosition(docKey, tapPosition);
+        await tester.tapAt(tapOffset);        
+        await tester.pumpAndSettle();
+        
+        final initialCaretOffset = _getCurrentDesktopCaretOffset(tester);
```
matthew-carroll, line 37: Generally, it's useful to comment about each thing that you're verifying, but it doesn't look like most/any of the expectation blocks in this suite have comments.

For example, this first block of 3 lines could say:

```
// Ensure that the caret is displayed at the correct (x,y) in the document before resizing the window
```

and the block down below

```
// Ensure that after resizing the window, the caret updated its (x,y) to match the text
// position that was pushed down to the next line.
```

I recommend stepping back and asking what it might be like for some other developer to come into a file like this to update some tests, or fix some bugs. What might confuse that developer? How can you make their job easier?

super_editor/test/src/default_editor/caret_test.dart
```diff
@@ -0,0 +1,312 @@
...
+        final expectedInitialCaretOffset = _computeExpectedDesktopCaretOffset(tester, textPosition);
+        expect(initialCaretOffset, expectedInitialCaretOffset); 
+
+        await _resizeWindow(
```
matthew-carroll, line 41:
```
// Make the window more narrow, pushing the caret text position down a line.
```

super_editor/test/src/default_editor/caret_test.dart
```diff
@@ -0,0 +1,312 @@
...
+      body: SuperEditor(                
+        documentLayoutKey: docKey,
+        editor: editor,        
+        gestureMode: gestureMode,                       
+      ),
+    ),
+  );              
+}
+
+Offset _getOffsetForPosition(GlobalKey docKey, DocumentPosition position){
```
matthew-carroll, line 242: There are a lot of these similar looking methods. Please provide a brief doc comment for each one that explains what it does so that a developer can find the one they want, without needing to read the implementation of every method.

matthew-carroll:
Looks good to me. Can you resolve the merge conflict? Then we can merge this in.

angelosilvestre:
@matthew-carroll I had to add a getter to `offset` in `CaretPainter` to keep the tests working without major changes. Any problem with that?

matthew-carroll:
@angelosilvestre can you describe why that was necessary? How did changes on `main` break your original approach in the absence of that getter?

angelosilvestre:
@matthew-carroll I was getting the caret offset via `BlinkingCaret` and apparently after b317145f73b7e352dbf05ddfdc896408560f786b it was not being used in desktop anymore

matthew-carroll:
Ok. I think this is acceptable for now. Instead of exposing a getter, can you just make `_offset` public and also add a `@visibleForTesting` annotation?

angelosilvestre:
@matthew-carroll Done!

matthew-carroll:
Thanks. I'll merge when the tests pass.

Unrelated to this PR, do you mind if I have you look into a document scrolling bug on the website?

angelosilvestre:
I can look into that

matthew-carroll (approved):
LGTM

PR #571, opened 2022-05-17
Renamed super_text to super_text_layout to avoid pub.dev name conflict

matthew-carroll:
Renamed super_text to super_text_layout to avoid pub.dev name conflict

@bk-one @miguelcmedeiros @roughike - I found out that `super_text` was already taken as a name on pub.dev, so I changed our package name to `super_text_layout`, which seemed like a reasonable compromise.

After the tests run, I'm going to merge this and release the first package version.


matthew-carroll:
All the super_editor failures look like the same failures we have on `main`, so not caused by this PR. The markdown failure looks like its because its looking for `super_text_layout` in GitHub, but that won't be available until we merge this. So I'm gonna go ahead and merge.

PR #577, opened 2022-05-20
refactor: replace Logger usages with logging package

canastro:
## Related
Related with a todo in `super_editor/lib/src/infrastructure/_logging.dart`:
> TODO: get rid of this custom Logger when all references are replaced with the logging package

## Context
I'm starting to look into `super editor` and trying to get familiar with the code and I found this `TODO` which seems like a good first contribution, although I didn't find any issue associated with it. So if its something you don't want to merge right now, that's ok 👍🏽

### DIFF
--- a/super_editor/lib/src/default_editor/list_items.dart
+++ b/super_editor/lib/src/default_editor/list_items.dart
@@ -12,8 +12,6 @@ import 'layout_single_column/layout_single_column.dart';
 import 'paragraph.dart';
 import 'text.dart';
 
-final _log = Logger(scope: 'list_items.dart');
-
 class ListItemNode extends TextNode {
   ListItemNode.ordered({
     required String id,
@@ -463,7 +461,9 @@ class IndentListItemCommand implements EditorCommand {
     final node = document.getNodeById(nodeId);
     final listItem = node as ListItemNode;
     if (listItem.indent >= 6) {
-      _log.log('IndentListItemCommand', 'WARNING: Editor does not support an indent level beyond 6.');
+      editorListItemsLog.info(
+        'WARNING: Editor does not support an indent level beyond 6.',
+      );
       return;
     }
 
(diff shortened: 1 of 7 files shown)


super_editor/lib/src/default_editor/list_items.dart
```diff
@@ -463,7 +461,9 @@ class IndentListItemCommand implements EditorCommand {
     final node = document.getNodeById(nodeId);
     final listItem = node as ListItemNode;
     if (listItem.indent >= 6) {
-      _log.log('IndentListItemCommand', 'WARNING: Editor does not support an indent level beyond 6.');
+      editorListItemsLog.info(
```
canastro, line 464: at first I used `editorListItemsLog.warning` but it seems the rest of the package mostly uses `.info`

super_editor/lib/src/infrastructure/_logging.dart
```diff
@@ -28,6 +33,11 @@ final editorImeLog = logging.Logger(LogNames.editorIme);
 final editorLayoutLog = logging.Logger(LogNames.editorLayout);
 final editorDocLog = logging.Logger(LogNames.editorDocument);
 final editorOpsLog = logging.Logger(LogNames.editorCommonOps);
+final editorBlockquoteLog = logging.Logger(LogNames.editorBlockquote);
+final editorBoxComponentLog = logging.Logger(LogNames.editorBoxComponent);
+final editorListItemsLog = logging.Logger(LogNames.editorListItems);
+final editorMultiNodeEditingLog = logging.Logger(LogNames.editorMultiNodeEditing);
+final editorTextToolsLog = logging.Logger(LogNames.editorTextTools);
```
canastro, line 40: not sure if I got the convention right.

matthew-carroll:
Hi @canastro - thanks for looking into super_editor and for trying to solve the logger problem. The desired replacement of the logs is pretty much a judgement call, which means it's probably faster for me to do that directly than to review a series of PR changes. I'm gonna close this PR, but I encourage you to keep looking into super_editor. If you get to the point that you think you could make substantial contributions, consider applying to the Flutter Bounty Hunters.

PR #579, opened 2022-05-21
Upgraded all lints from 1.0.0 to 2.0.1

matthew-carroll:
Upgraded all lints from 1.0.0 to 2.0.1, except super_editor_markdown because it causes a build failure until we release super_editor v0.2.1


miguelcmedeiros (approved):
LGTM

PR #584, opened 2022-05-23
Fix overflow when having constrained height inside a scrollable widget (Resolves #462)

### LINKED ISSUE #462: Demo Website: the content in the editor exceeds the editor space
On the demo website, adding content vertically causes the content to exceed the white space for the editor. The user ends up typing down into the page.

superlist.com/SuperEditor

Instead, when the content reaches the bottom of the white space, the editor should constrain the vertical height of the content, and the document should scroll in the white space.

angelosilvestre:
Fix overflow when having constrained height inside a scrollable widget. Resolves https://github.com/superlistapp/super_editor/issues/462

Wrapping `SuperEditor`  with a scrollable widget and trying to constrain the height of `SuperEditor` will cause it to overflow when it's content exceeds the available space

The cause is that both `DocumentMouseInteractor` and `ScrollableDocument` wouldn't create it's own scrollview when an ancestor scrollable is found in the widget tree. It would also use the `RenderBox` of the ancestor scrollable to perform it's calculations instead of it's own `RenderBox`.

I changed this to check if it has `maxHeight` in it's `BoxConstraints` and, if so, behave as if it doesn't have an ancestor scrollable.


angelosilvestre:
@matthew-carroll Could you review this PR?

super_editor/lib/src/default_editor/document_gestures_mouse.dart
```diff
@@ -210,25 +211,28 @@ class _DocumentMouseInteractorState extends State<DocumentMouseInteractor> with
...
+  /// constraint, then the returned `ScrollPosition` belongs to that ancestor 
+  /// `Scrollable`, and this widget doesn't include a `ScrollView`.
   ///
-  /// If this widget doesn't have an ancestor `Scrollable`, then this
-  /// widget includes a `ScrollView` and the `ScrollView`'s position
+  /// If this widget doesn't have an ancestor `Scrollable` or has a `maxHeight`
+  /// constraint, then this widget includes a `ScrollView` and the `ScrollView`'s position
   /// is returned.
-  ScrollPosition get _scrollPosition => _ancestorScrollPosition ?? _scrollController.position;
+  ScrollPosition get _scrollPosition => _hasMaxHeightConstraint
```
matthew-carroll, line 221: I don't think we want this change. If there's an ancestor scrollable then we *must* honor that scrollable. Can you describe why you think it makes sense to ignore an ancestor scrollable based on our constraints? For example, how would auto-scrolling work?

angelosilvestre: If we don't make this change then we can't scroll SuperEditor to see the content that doesn't fit on the available space

angelosilvestre: With this change:

https://user-images.githubusercontent.com/7597082/169923794-0f70c138-ea5b-442b-a579-88d10ca2931d.mp4

Without this change:


https://user-images.githubusercontent.com/7597082/169923821-ac286157-247b-4057-86fb-97be2ce71083.mp4

matthew-carroll: I understand that this change might address some detail that you're focused on. I'm concerned about all the other use-cases. How does auto-scrolling work if we're embedded in a `Scrollable` and we don't return that `Scrollable`'s position?

angelosilvestre: Do we have an example of auto-scrolling that I can check and verify the behavior?

matthew-carroll: I don't know if we have an absolute example that represents all use-cases, but in the example app you can open the slivers demo, start selects some text in the document, and then drag down to the bottom of the viewport. You should be able to auto-scroll to the bottom of the document, and beyond, causing all the slivers below the document to scroll for as long as your mouse is near the bottom of the viewport.

angelosilvestre: Without the changes:

https://user-images.githubusercontent.com/7597082/169928583-a5652547-1139-488f-b169-d945172e4c4d.mp4

With the changes: 


https://user-images.githubusercontent.com/7597082/169928618-e9a2a9eb-0102-4251-a902-7ca61d4b0b8e.mp4

So yeah, it breaks auto-scroll when scrolling up...

I'll try to make it work

matthew-carroll: I recommend thinking in terms of meaning, rather than whether or not a particular condition solves a problem. Scrolling is a particularly nuanced, nasty, and frustrating area to work. So if you don't establish some concrete details, you'll probably find that you keep breaking one thing after another. That was my experience the first time around. As an example, when we consider that we have a property that represents the scroll position for the scrolling context, that's an invariant that we wouldn't want to redefine just to get around an unusual bug or situation. Instead, we have to treat that property as unavoidable, and then look for a more appropriate place to identify and solve layout issues.

I still haven't spent much time digging into this problem, but based on a cursory overview, I'm still not sure what the root problem is. I'm not sure why this was working before, but isn't working now. And I'm unclear as to why `SuperEditor` can't handle unconstrained height. If you dig into those questions, you might find a more appropriate place to address the bug. My guess is, there's a solution to this problem that involves a relatively small number of changes, those changes probably don't span very many places, and the solution will look surgical.

angelosilvestre: Until 52a2f273584dff5310cf5150c79974b5aae99969 constrained `SuperEditor` inside a scrollview was working, but without auto-scrolling. After this commit the overflow has started to happen.

The main problem is that `DocumentMouseInteractor` isn't creating it's own scrollview when it's height is constrained. By changing this the overflow doesn't happen anymore.

Regarding scrolling a `SuperEditor` with constrained height, we can change `_ensureSelectionExtentIsVisible`  to scroll it's own `ScrollPosition` before the ancestor's `ScrollPosition`:

https://github.com/superlistapp/super_editor/blob/1c4cdbb6ea4760568f39b57238f79cfb678510ae/super_editor/lib/src/default_editor/document_gestures_mouse.dart#L319-L328

Before assigning `newScrollPosition` I added this:

```dart
  if (hasMaxHeightConstraint && 
      hasAncestorScrollable && 
      !(_scrollController.position.atEdge && _scrollController.position.pixels > 0)
   ){        
        final newScrollPosition =
          (beyondBottomExtent + _scrollController.position.pixels).clamp(0.0, _scrollController.position.maxScrollExtent); 
        _scrollController.position.animateTo(
          newScrollPosition,
          curve: Curves.easeOut,
        );                
        return;
      }
```

By doing this, `SuperEditor` scrolls it's content and when it reaches the bottom it scrolls it's ancestor's `ScrollPosition` (scrolling up is similar)

Unconstrained height:


https://user-images.githubusercontent.com/7597082/170153526-9848f4af-54f4-4fd1-923c-e4f304165dd3.mp4

Constrained height:


https://user-images.githubusercontent.com/7597082/170153536-99595939-6afe-4af6-8349-aecef233864e.mp4

What do you think about this approach?

angelosilvestre:
@matthew-carroll Could you please take a look into the desktop implementation

Issue #351, opened 2021-11-16
Page up/down and home/end introduce box-with-question-mark-character


miguelcmedeiros:
Steps to reproduce:
* Run super_editor's demo app on macOS
* Place the caret anywhere in a paragraph node
* Press Page up/down or home/end

Expected: whatever macOS usually does when these keys are pressed in the context of text editing (probably needs to be spec'ed out)
Actual: a new character (box-with-question-mark-character) is added at the position where caret is
https://user-images.githubusercontent.com/31278849/142051000-238e3fd3-fe39-4f3f-a28a-f3775e85c96b.mov

Note: on web nothing happens when pressing these keys.

bk-one:
@miguelcmedeiros closing this - I think it's fixed, please re-open if still relevant

miguelcmedeiros:
I can still reproduce it

matthew-carroll:
@miguelcmedeiros @bk-one what should those buttons do? I don't know what the standard caret movement rules are for those keys, or scrolling rules when there is no caret position.

bk-one:
@matthew-carroll I think the names are pretty self-explanatory. However, I also don't think that the implementation is very important. So we might want to split this into various action items:
- [ ] 1. Make sure the keys do not produce question-mark-box characters
- [ ] 2. Implement page-up and page-down to scroll one full viewport length up or down. The caret stays at its current position and is not moved
- [ ]  3. Implement [home] and [end], scroll to the first line or last line of the document. The caret stays at its current position and is not moved.
- [ ] 4. double-check that on caret movement (e.g. cursor keys or entering a character) the viewport is scrolled back to have the caret visible.

matthew-carroll:
@faithoflifedev - would you like to try to implement Page Up, Page Down, Home, and End keyboard handlers for the `SuperEditor` widget?

faithoflifedev:
I'll take a look and let you know if it's something I can handle.

faithoflifedev:
@matthew-carroll I have some initial code that covers step 1.  Currently, when a user presses any "F" key, the question-mark-box will be displayed, so the code covers all "F" keys.

```dart
//document_keyboard_actions.dart
ExecutionInstruction doNothingWhenFKeyPressed({
  required EditContext editContext,
  required RawKeyEvent keyEvent,
}) {
  if (keyEvent.logicalKey.keyId < LogicalKeyboardKey.f1.keyId ||
[...]

PR #589, opened 2022-05-24
fixes for #351

faithoflifedev:


faithoflifedev:
@matthew-carroll  here is the draft PR for issue #351

matthew-carroll:
Hi @faithoflifedev - I'm about to look into the PR. Before I get to the code, I'll mention that PR titles and descriptions are good places to include useful information.

It would be helpful to see a title that describes what you're solving, rather than sending the reader to some issue ticket. You don't have to restate everything, but a concise synopsis is nice.

As for the PR description, that's your opportunity to prime the reader for what they're about to see in the code. How did you solve the problem? Did you hit any hurdles? Should the reader expect anything unusual? I recommend treating the PR write-up as your guide for the code that we're now reviewing.

With that, I'll look into the code...

super_editor/lib/src/default_editor/document_keyboard_actions.dart
```diff
@@ -12,6 +12,62 @@ import 'document_input_keyboard.dart';
 import 'paragraph.dart';
 import 'text.dart';
 
+ExecutionInstruction scrollWhenNavigationKeyPressed({
```
matthew-carroll, line 15: I see that other actions aren't commented. That's probably an oversight. These handlers should include clear explanation for why they exist, and what they do.

As for the name, I'm not sure that anyone is going to know what a "navigation key" is. It's probably clearer to simply list the keys that are relevant: `scrollOnPageUpPageDownHomeAndEnd()`

super_editor/lib/src/default_editor/document_keyboard_actions.dart
```diff
@@ -12,6 +12,62 @@ import 'document_input_keyboard.dart';
...
+  required EditContext editContext,
+  required RawKeyEvent keyEvent,
+}) {
+  //the super_editor doesn't always have a scroll controller
```
matthew-carroll, line 19: I think that it always should have one, or at least always have a `ScrollPosition`. In any event, there should always be a scrollable that surrounds a document.

super_editor/lib/src/default_editor/document_keyboard_actions.dart
```diff
@@ -12,6 +12,62 @@ import 'document_input_keyboard.dart';
...
+  if (keyEvent.logicalKey.keyId < LogicalKeyboardKey.end.keyId ||
+      keyEvent.logicalKey.keyId > LogicalKeyboardKey.pageUp.keyId) {
+    return ExecutionInstruction.continueExecution;
+  }
+
+  final scrollController = editContext.scrollController!;
+
+  double? jumpTo;
+
+  //end key or fn-arrowDn (on Mac)
```
matthew-carroll, line 33: I see a number of comments that appear to restate the key comparison in the if-statements. I don't think these comments are telling us more than the if-statements that follow. The interesting details are where you're jumping. For example, within the first if-branch, it might nice to say `// The user pressed the END key, scroll to the bottom of the document`.

super_editor/lib/src/default_editor/document_keyboard_actions.dart
```diff
@@ -12,6 +12,62 @@ import 'document_input_keyboard.dart';
...
+    jumpTo = scrollController.position.maxScrollExtent;
+    //home key or fn-arrowUp (on Mac)
+  } else if (keyEvent.logicalKey == LogicalKeyboardKey.home) {
+    jumpTo = scrollController.position.minScrollExtent;
+    //pgDn key or fn-arrowDn (on Mac)
+  } else if (keyEvent.logicalKey == LogicalKeyboardKey.pageDown) {
+    final nextJump = scrollController.offset + scrollController.position.extentInside;
+
+    jumpTo =
+        nextJump < scrollController.position.maxScrollExtent ? nextJump : scrollController.position.maxScrollExtent;
```
matthew-carroll, line 44: I think this line can be replaced with a call to `min()` when you set `nextJump`. Also, it's unclear why one thing is called `nextJump` and the other is called `jumpTo`. I'm guessing you did that because you needed a temporary variable. In this case you don't a temporary variable, but if you did, it would be preferable to choose a name that explains why you need 2 variables, e.g., `scrollOffsetOnePageDown`. That name implies that there may be a difference between "one page down" and "jump to"

super_editor/lib/src/default_editor/document_keyboard_actions.dart
```diff
@@ -12,6 +12,62 @@ import 'document_input_keyboard.dart';
...
+    //pgUp key or fn-arrowUp (on Mac)
+  } else if (keyEvent.logicalKey == LogicalKeyboardKey.pageUp) {
+    final nextJump = scrollController.offset - scrollController.position.extentInside;
+
+    jumpTo = nextJump > 0 ? nextJump : scrollController.position.minScrollExtent;
```
matthew-carroll, line 49: This line can be replaced with a call to `max()` when you set `nextJump`

super_editor/lib/src/default_editor/document_keyboard_actions.dart
```diff
@@ -12,6 +12,62 @@ import 'document_input_keyboard.dart';
...
+  }
+
+  if (jumpTo != null) {
```
matthew-carroll, line 52: How would `jumpTo` be null?

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -347,6 +347,7 @@ class _SuperEditorState extends State<SuperEditor> {
 
   void _createEditContext() {
     _editContext = EditContext(
+      scrollController: widget.scrollController,
```
matthew-carroll, line 350: Let's pass this as the last argument, for consistency with other locations.

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -576,6 +577,8 @@ final defaultComponentBuilders = <ComponentBuilder>[
 
 /// Keyboard actions for the standard [SuperEditor].
 final defaultKeyboardActions = <DocumentKeyboardAction>[
+  scrollWhenNavigationKeyPressed,
```
matthew-carroll, line 580: Do we want to scroll a document when there's no selection? If not, this handler should go beneath `doNothingWhenThereIsNoSelection`.

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -576,6 +577,8 @@ final defaultComponentBuilders = <ComponentBuilder>[
...
+  doNothingWhenFnKeyPressed,
```
matthew-carroll, line 581: Why is this handler needed? What happens if it's not here?

super_editor/test/src/default_editor/navigation_keys_test.dart
```diff
@@ -0,0 +1,162 @@
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:super_editor/super_editor.dart';
+
+enum Extent { top, bottom, pgDown, pgUp }
```
matthew-carroll, line 6: I generally recommend placing all declarations after the `main()` method. When you're reading a test file, you probably don't have any context to understand these declarations because you haven't seen any of the test definitions yet. Therefore, they're just getting in the way.

They should also include doc comments, and if they're not intended to be referenced from other test files, they should be private.

With regard to naming, unless there's an overriding factor, we shouldn't abbreviate things. `pgDown` -> `pageDown`, `pgUp` -> `pageUp`. Those couple extra characters don't cost us anything, but they avoid a moment of friction when reading the code.

super_editor/test/src/default_editor/navigation_keys_test.dart
```diff
@@ -0,0 +1,162 @@
...
+
+enum ScrollTo { bottom }
+
+void main() {
+  group('Navigation Keys', () {
+    double screenWidth = 640.0;
```
matthew-carroll, line 12: I haven't looked at where this variable is used, yet, but any value that doesn't change should either be `const` or `final`. Any value that does change should probably not be defined outside the test that changes it.

super_editor/test/src/default_editor/navigation_keys_test.dart
```diff
@@ -0,0 +1,162 @@
...
+
+    double screenHeight = 120.0;
+
+    Future<void> _testExtent(WidgetTester tester, Extent extent,
```
matthew-carroll, line 16: As I read this code from top to bottom, when I reach `_testExtent`, I have no basis to appreciate what that means. I can clean that you're probably doing something with scrolling, but I have no idea what you're testing.

As for function declarations, you should prefer each parameter on its own line, rather than a parameter list that wraps like this. Just add a `,` after the final parameter and the auto-formatter will update it.

super_editor/test/src/default_editor/navigation_keys_test.dart
```diff
@@ -0,0 +1,162 @@
...
+        {required LogicalKeyboardKey key, bool scrollToBottom = false}) async {
+      final testDocument = TestDocument(tester, screenWidth: screenWidth, screenHeight: screenHeight);
```
matthew-carroll, line 18: We have a class called `Document` in `super_editor`, so it's not a good idea to create something called `TestDocument` if it's not a `Document`. Moreover, a `Document` in `super_editor` is a purely logical construct. It has no concept of layout or painting. So this further muddies the water by adding Flutter UI concepts into this thing called a `TestDocument`.

super_editor/test/src/default_editor/navigation_keys_test.dart
```diff
@@ -0,0 +1,162 @@
...
+
+      await testDocument.buildDoc();
```
matthew-carroll, line 20: This approach to pumping a widget tree might be reasonable, but it's unexpected. I would expect to either see a call to `tester.pumpWidget()`, or invocation of a method such as `_pumpDocumentLayout()`, which would internally configure the widget tree so that we don't need to repeat it in every test. At first glance, I don't see any benefit of baking this behavior into a special object.

super_editor/test/src/default_editor/navigation_keys_test.dart
```diff
@@ -0,0 +1,162 @@
...
+
+      //caret is at begining of doc
```
matthew-carroll, line 22: The small addition of a verb can be helpful to disambiguate a comment like this: `//caret is at beginng of doc` -> `// Ensure that the caret is at the beginning of the document`.

It's not clear why the caret should be at the beginning of the doc. Did the above `buildDoc()` method also set a selection? If so, that detail should be in the name. If you move that method to a standalone function, it might look like `_pumpDocumentLayoutWithCollapsedSelection()`. Additionally, if that method is used in many tests, there's no need to verify what the method did. If that method fails to do its job, a bunch of your tests will fail and you'll figure it out. We're more concerned with verifying intermediate and final results.

Also, please add a space after the `//` for comments.

super_editor/test/src/default_editor/navigation_keys_test.dart
```diff
@@ -0,0 +1,162 @@
...
+      expect(testDocument.textNodePosition.offset == 0, true);
+
+      if (scrollToBottom) {
```
matthew-carroll, line 25: Generally, tests should not contain conditional logic. You should know exactly what behavior you're running, and what results you're verifying. Is there another way to achieve your testing goals?

super_editor/test/src/default_editor/navigation_keys_test.dart
```diff
@@ -0,0 +1,162 @@
...
+        testDocument.scrollController.jumpTo(testDocument.extentBottom);
+      }
+
+      //activate key press
```
matthew-carroll, line 29: This comment doesn't add anything. Mentioning what key we're pressing and/or why we're pressing it would be a much better use of a comment.

super_editor/test/src/default_editor/navigation_keys_test.dart
```diff
@@ -0,0 +1,162 @@
...
+      await tester.sendKeyEvent(key);
+
+      await tester.pumpAndSettle();
```
matthew-carroll, line 32: It's a good idea to explain every `pumpAndSettle()`. Ideally an action should only require a `pump()`, so if we need the "settle" then it's nice to know why.

super_editor/test/src/default_editor/navigation_keys_test.dart
```diff
@@ -0,0 +1,162 @@
...
+
+      final offset = testDocument.scrollController.offset;
+
+      switch (extent) {
```
matthew-carroll, line 36: At first glance, I don't like this switch statement. This forces the reader to become a human compiler and try to figure out what we're really verifying. I'd rather see 4 simple tests than 1 test with 4 switch branches.

super_editor/test/src/default_editor/navigation_keys_test.dart
```diff
@@ -0,0 +1,162 @@
...
+        case Extent.top:
+          //confirm no scroll beyond top
```
matthew-carroll, line 38: But why do we expect this? What are we testing here?

super_editor/test/src/default_editor/navigation_keys_test.dart
```diff
@@ -0,0 +1,162 @@
...
+          expect(offset, equals(screenHeight));
+          break;
+
+        case Extent.pgUp:
+          //confirm scroll up one page
+          expect(offset, equals(testDocument.extentBottom - screenHeight));
+          break;
+      }
+
+      //caret should not move
```
matthew-carroll, line 58: Why not? I've read from top to bottom of this method and I have no idea what we're testing. All I know is that some key was pressed, for some reason, and then we expect one of a set of things to have happened...or not happened, I guess.

super_editor/test/src/default_editor/navigation_keys_test.dart
```diff
@@ -0,0 +1,162 @@
...
+import 'package:flutter/services.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:super_editor/super_editor.dart';
+
+enum Extent { top, bottom, pgDown, pgUp }
+
+enum ScrollTo { bottom }
+
+void main() {
+  group('Navigation Keys', () {
```
matthew-carroll, line 11: The `K` can be lowercase.

super_editor/test/src/default_editor/navigation_keys_test.dart
```diff
@@ -0,0 +1,162 @@
...
+          //confirm scroll up one page
+          expect(offset, equals(testDocument.extentBottom - screenHeight));
+          break;
+      }
+
+      //caret should not move
+      expect(testDocument.textNodePosition.offset == 0, true);
+    }
+
+    group('navigation', () {
```
matthew-carroll, line 62: I noticed that the first group is capitalized, but this group is not.

super_editor/test/src/default_editor/navigation_keys_test.dart
```diff
@@ -0,0 +1,162 @@
...
+      testWidgets('pageDown key press at bottom of viewport',
+          (tester) async => _testExtent(tester, Extent.bottom, key: LogicalKeyboardKey.pageDown, scrollToBottom: true));
```
matthew-carroll, line 64: I see a test here that's completely defined in terms of a configuration to some other method. This must be a common practice in other languages because I've seen this a number of times from other people. But this should never be done.

If you think about it, immediately delegating a test to some other method doesn't help the reader at all. You haven't reduced the code in the test, you've only moved it. And worse yet, the code at the destination is sure to be generalized. By definition, that `_testExtent()` method is actually many tests in one method. So the reader has to become a compiler and execute that method in his/her head. This practice breeds all sorts of mistakes and bugs and frustrations.

Instead, define one test within one test declaration. You can centralize certain pieces of tests, like an initial tree pump, when that centralization doesn't hurt the reader's ability to appreciate what's being tested, why it's being tested, and how it's being tested.

matthew-carroll (changes requested):
I took a first pass over the source code. I think I've probably left enough comments to make some substantial updates. I can take another look when you think it's ready.

faithoflifedev:
Hi Matt,

Sorry for the delay getting to this, I had some other projects that I
needed to spend time on.  Anyway, I've been able to go through the feedback
you provided and I believe I've addressed all the items that you
identified.  The PR us available at -
https://github.com/superlistapp/super_editor/pull/608

Regards,
Chris

PR #592, opened 2022-05-27
Export again `super_textfield/style.dart`

miguelcmedeiros:
This file used to be exported and was also being used in client code. It was probably removed during recent refactoring (splitting `super_text_layout`, etc.).


matthew-carroll:
I checked the last 5 or 6 revisions of this file and I didn't see `styles.dart` exported from it anywhere. Since it seemed to have been exposed somewhere, we can merge this in, but we may need to export this from a different place in the future, depending on why it wasn't exported from this filed before.

matthew-carroll (approved):

PR #594, opened 2022-05-28
Added a hack fix to _test_tools.dart for finding the offset to tap in a text field

matthew-carroll:
Added a hack fix to _test_tools.dart for finding the offset to tap in a text field. Flutter is reporting -0.2 in the 'dy' instead of 0.0, which results in a tap outside the field.


matthew-carroll:
@miguelcmedeiros I'm gonna merge this in so that we can pull it into Superlist

Issue #554, opened 2022-05-12
Fix deprecations that showed up in Flutter 3


matthew-carroll:
Flutter 3, the latest stable version, deprecated some APIs that are used in `super_editor`. Go through and refactor those uses, as needed.

fabiancrx:
cc @matthew-carroll this can be closed now that #595 was merged

PR #595, opened 2022-05-28
Fix deprecations from flutter 3 (#554)

fabiancrx:
This PR is pretty straightforward, there wasn't too much to fix , and the fixes required just refactoring to the new recommended behavior from the deprecated API's


matthew-carroll (approved):
LGTM - Thanks.

PR #596, opened 2022-05-30
Fixed compound-character IME crash (Resolves #586)

### LINKED ISSUE #586: Typing ŭ in empty paragraph crashes the app
Typing ŭ in empty paragraph crashes the app. The crash is a complete app crash, not just a Dart crash. 

To insert the character in question, select the "German" keyboard and then press:
1. opt + u
2. u

The crash only seems to occur when inserting into an empty paragraph.

matthew-carroll:
Fixed compound-character IME crash (Resolves #586)

The root problem was that we conditionally include invisible prepending characters, e.g., ". ", when editing an empty paragraph with content above it. We do this so that the IME reports backspace button presses, and still auto-capitalizes the first character of the paragraph. I describe in #586 why this became a problem.

This PR resolves the composing region mismatch that caused the crash, and also adds a test to prevent a regression.

https://user-images.githubusercontent.com/7259036/170907332-053c0e46-547e-4376-9779-02051eaadc0d.mov

### DIFF
--- a/super_editor/lib/src/default_editor/document_input_ime.dart
+++ b/super_editor/lib/src/default_editor/document_input_ime.dart
@@ -135,10 +135,15 @@ class _DocumentImeInteractorState extends State<DocumentImeInteractor> implement
     if (selection == null) {
       _detachFromIme();
     } else {
-      if (isAttachedToIme) {
+      if (isAttachedToIme && !_isApplyingDeltas) {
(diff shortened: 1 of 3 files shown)


MarcelKaeding (approved):

miguelcmedeiros (approved):
🚀

PR #599, opened 2022-05-31
Fixed span merge bug in addAttribution() (Resolves #582)

### LINKED ISSUE #582: `[AttributedSpan addAttribution]` doesn't correctly expand existing attribution
Hi might be missing something, but the following code

```dart
  final text = AttributedText(text: '01234567');
  text.spans
      .addAttribution(newAttribution: italicsAttribution, start: 0, end: 4);
  text.spans
      .addAttribution(newAttribution: italicsAttribution, start: 4, end: 8);
  print('text $text');
```
prints
```
text [AttributedText] - "01234567"
[AttributedSpans] (2 spans):
 - [SpanMarker] - attribution: [NamedAttribution]: italics, offset: 0, type: SpanMarkerType.start
 - [SpanMarker] - attribution: [NamedAttribution]: italics, offset: 4, type: SpanMarkerType.end
 - [SpanMarker] - attribution: [NamedAttribution]: italics, offset: 8, type: SpanMarkerType.end
```
Which is quite unexpected. Shouldn't the first `SpanMarkerType.end` be removed?

matthew-carroll:
Fixed span merge bug in addAttribution() (Resolves #582)

The problem was the case when an existing span ends exactly where a new span begins. The specific behavior in the `addAttribution()` method requires this situation to be handled explicitly, by removing the existing end marker. This PR does that.

This PR also renames `_attributions` to `markers` for clarity. And `markers` is public for testing verification.

### DIFF
--- a/attributed_text/lib/src/attributed_spans.dart
+++ b/attributed_text/lib/src/attributed_spans.dart
@@ -1,6 +1,7 @@
 import 'dart:math';
 
 import 'package:collection/collection.dart';
+import 'package:meta/meta.dart';
 
 import 'attribution.dart';
 import 'logging.dart';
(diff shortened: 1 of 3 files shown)


matthew-carroll:
@knopp this should solve your problem, but I'd still like for you to post in the original issue ticket with the example of what you were actually trying to do when this problem appeared.

miguelcmedeiros (approved):
LGTM

knopp (approved):
This resolves the issue. LGTM!

PR #602, opened 2022-06-01
Using TextAffinity to determine selection

### LINKED ISSUE #549: [SuperEditor] Drag to select skips one character in the upstream direction
When creating a text selection in the upstream direction with the mouse, the selection base will start one character further upstream than expected. See the below video for a couple examples:

https://user-images.githubusercontent.com/1316184/167518847-9ce9baae-b543-4d26-bba6-16910809a8e3.mov

I tracked the cause down to the the following line in [`text.dart`](https://github.com/superlistapp/super_editor/blob/cd0efedfd835b65ce47ce668665cd6444c98c69c/super_editor/lib/src/default_editor/text.dart#L430-L444)

```dart
    // Rework the textPosition so that it reports a "downstream" affinity because
    // the editor doesn't support "upstream" positions, yet.
    //
    // Applying the "-1" to switch from upstream to downstream works everywhere, except
    // when the position is at the very end of the text. In that case, we leave the offset
    // alone.
    return TextNodePosition.fromTextPosition(textPosition.affinity == TextAffinity.downstream
        // The textPosition is already "downstream", leave it alone.
        ? textPosition
        // The textPosition if "upstream", adjust it to become "downstream", unless
        // the position sits at the very end of the text.
        : TextPosition(
            offset: textPosition.offset < widget.text.text.length ? textPosition.offset - 1 : textPosition.offset,
            affinity: TextAffinity.downstream,
          ));
```

Revising that code to always use `textPosition` without any modifications fixes the drag to select issue but presumably breaks other forms of selection elsewhere in the codebase (though no tests fail).

jmatth:
Closes #549.

Old behavior:

https://user-images.githubusercontent.com/1316184/171486869-2059dea1-5852-42ad-96cb-d82a1814bd75.mov

New behavior:

https://user-images.githubusercontent.com/1316184/171486892-88b1478c-d8a3-4b81-8e09-bf09ee22ac90.mov

I have a handful of tests failing locally but they also fail on `main` without any changes, so are presumably unrelated. Hopefully the CI run sorts it out.


jmatth:
@matthew-carroll: just want to ping you on this PR since it's been sitting around for a while and #549 is still open. It should be up to date with `main` and with the tests added in #623 have been updated to match the new behavior.

matthew-carroll:
Thanks @jmatth 

@angelosilvestre can you take a look at these changes to give me a 2nd pair of eyes on the code? Can you think of anything that's missing?

@miguelcmedeiros would you mind giving this PR branch a try in Superlist and see if anything weird happens with caret placement, or expanded selections?

super_editor/test/super_editor/supereditor_selection_test.dart
```diff
@@ -171,7 +176,12 @@ void main() {
         selection,
         DocumentSelection(
           base: DocumentPosition(nodeId: secondParagraphId, nodePosition: const TextNodePosition(offset: 0)),
-          extent: DocumentPosition(nodeId: secondParagraphId, nodePosition: const TextNodePosition(offset: 16)),
+          extent: DocumentPosition(
+              nodeId: secondParagraphId,
+              nodePosition: const TextNodePosition(
+                offset: 16,
+                affinity: TextAffinity.upstream,
+              )),
```
angelosilvestre, line 184: Should the affinity be really upstream when dragging downstream?

jmatth: I think that makes sense. Say we have some text that's soft wrapped:

```
abc
zyx
```

and we drag to select from a through c:

```
|abc|
xyz
```

The extent position should be between c and x and render at the end of the first line, i.e. `TextPosition(offset: 3, affinity: TextAffinity.upstream)`

angelosilvestre: Ok, that makes sense.

matthew-carroll: Shouldn't that example depend on the drag offset? The caret should only be upstream if the drag offset is at the y-value of that line. If the drag offset sits to the left of the "x" with a y-value on that line, then the caret should sit just before the "x", not after the "c". Is that situation handled with this PR?

jmatth: > If the drag offset sits to the left of the "x" with a y-value on that line, then the caret should sit just before the "x", not after the "c".

That's what I would expect. The example I gave was just to demonstrate why affinity isn't tied to drag direction.

> Is that situation handled with this PR?

That situation works in this PR but I didn't add any tests specifically for it. It looks like we could just add assertions for extent affinity to tests [like this one][1] and it would provide sufficient coverage.

[1]: https://github.com/superlistapp/super_editor/blob/700f0f752f9b0ac2ee05ea10e79edf97f114c26c/super_editor/test/super_editor/supereditor_selection_test.dart#L31-L37

matthew-carroll: You're suggesting that we add expectations for the affinity value for `TextNodePosition`s?

How about the actual caret painting? I think we have some tests that are related to absolute caret offset. I think it would be a good idea to either add tests for affinity, or adjust those tests to also cover affinity. Otherwise, it's possible that our logical selection is correct, but our painting behavior is incorrect.

jmatth: Good point, we should have tests for the caret rendering too. I'm guessing the tests in [`caret_test.dart`][1] are the tests for absolute caret offset you're referring to?

[1]: https://github.com/superlistapp/super_editor/blob/700f0f752f9b0ac2ee05ea10e79edf97f114c26c/super_editor/test/src/default_editor/caret_test.dart

miguelcmedeiros:
> @miguelcmedeiros would you mind giving this PR branch a try in Superlist and see if anything weird happens with caret placement, or expanded selections?

@matthew-carroll, tested it out with Superlist and didn't find any regression with regards to caret placement or expanded selections.

super_editor/test/super_editor/supereditor_selection_test.dart
matthew-carroll: Yeah, that looks like a good place.

jmatth:
Added tests for upstream/downstream caret positioning at line breaks, plus an extra test to ensure affinity doesn't change the rendering if the caret isn't at a line break.

super_editor/test/src/default_editor/caret_test.dart
```diff
@@ -13,6 +13,140 @@ void main() {
...
+          DocumentPosition(
+        nodeId: '1',
+        nodePosition: TextNodePosition.fromTextPosition(upstreamTextPosition),
+      );
+      final tapDownstreamPosition = DocumentPosition(
+        nodeId: '1',
+        nodePosition: TextNodePosition.fromTextPosition(downstreamTextPosition),
+      );
+
+      testWidgets('renders caret at end of line when affinity is upstream', (WidgetTester tester) async {
```
matthew-carroll, line 39: Don't we already have test methods that accept a size and sets up the the window, automatically?

Also, if this test should pass on all platforms, please use the test method for all platforms.

super_editor/test/src/default_editor/caret_test.dart
```diff
@@ -13,6 +13,140 @@ void main() {
     const textPosition = TextPosition(offset: 46);
     final tapPosition = DocumentPosition(nodeId: '1', nodePosition: TextNodePosition(offset: textPosition.offset));
 
+    group('affinity', () {
+      // Carefully chosen magic numbers so our taps hit where the text is soft-wrapped.
```
matthew-carroll, line 17: Can you explain your selection of these values? We don't want any magic in tests. Sooner or later, these tests will fail, and a developer will need to debug the reason why. We need to ensure that future developers know how to work through the setup and the expectations.

super_editor/test/src/default_editor/caret_test.dart
```diff
@@ -13,6 +13,140 @@ void main() {
...
+        await tester.pumpWidget(
+          _createTestApp(
+            gestureMode: DocumentGestureMode.mouse,
+            docKey: docKey,
+            composer: composer,
+          ),
+        );
+        await tester.pumpAndSettle();
+
+        final tapOffset = _getOffsetForPosition(docKey, tapUpstreamPosition);
```
matthew-carroll, line 56: If you're placing a caret in a document, please use the robot methods for that purpose, or let me know what isn't working for that.

super_editor/test/src/default_editor/caret_test.dart
```diff
@@ -13,6 +13,140 @@ void main() {
...
+
+        expect(
+            composer.selection,
+            DocumentSelection.collapsed(
+                position: DocumentPosition(
+              nodeId: '1',
+              nodePosition: TextNodePosition.fromTextPosition(upstreamTextPosition),
+            )));
+        final caretOffset = SuperEditorInspector.findCaretOffsetInDocument();
+        final expectedCaretOffset = _computeExpectedDesktopCaretOffset(tester, upstreamTextPosition);
```
matthew-carroll, line 69: This looks like a cyclical condition. It appears that what we're testing is that SuperEditor positions the caret wherever SuperEditor positions the caret. Instead, we need to ensure that SuperEditor positions the caret at the offset that we would expect.

super_editor/test/src/default_editor/caret_test.dart
jmatth: > Don't we already have test methods that accept a size and sets up the the window, automatically?

Could you point me to where that is? I based this off the setup code from the window resizing tests [here][1].

> Also, if this test should pass on all platforms, please use the test method for all platforms.

Is this not already running on all platforms? I can see it in the github actions logs for Windows, macOS, and Linux.

[1]: https://github.com/superlistapp/super_editor/blob/7b74e9ea48fbfc7638d7536111a6bc970f40861d/super_editor/test/src/default_editor/caret_test.dart#L150-L158

super_editor/test/src/default_editor/caret_test.dart
jmatth: I just set a small enough width to make the text take up multiple lines then incremented the text position offset until I found where it wrapped. I could write that offset calculation into the start of the test rather than hard-coding the final value.

I wrote it this way based off of how the window resize tests already in this file work: they expect that position 46 is on one line if the window has a width of 1000 and a different line if the window has a width of 250, and these numbers come from... ¯\_(ツ)_/¯

super_editor/test/src/default_editor/caret_test.dart
jmatth: I thought so too when I saw that `_computeExpectedDesktopCaretOffset` ultimately ends up just calling `DocumentLayout.getRectForPosition`, but I'm not sure how to better define "where we expect" other than just hardcoding an offset with magic numbers. And again, I figured this would work as a test since it's pretty much identical to expectations in other tests [already in this file][1]. These tests do fail as expected when run against the `main` branch.

[1]: https://github.com/superlistapp/super_editor/blob/7b74e9ea48fbfc7638d7536111a6bc970f40861d/super_editor/test/src/default_editor/caret_test.dart#L170-L177

super_editor/test/src/default_editor/caret_test.dart
jmatth: Will do. Interestingly, the robot method `placeCaretInParagraph` doesn't support specifying `TextAffinity`, so it's not possible to tap at the start of a line that's soft-wrapped. I can work around it in this by tapping the end of the line and then updating the selection manually.

jmatth: After looking into this I think the robot should probably be updated to take optional `TextAffinity` arguments anyway. Right now if you have text that's wrapped, say:

```
abc
efg
```

It's impossible to place the caret in the top-right corner after the `c`. That position is `TextPosition(offset: 3, affinity: TextAffinity.upstream`, and the robot methods all default to `downstream` affinities.

I can still work around it in this PR by just manipulating the selection via the composer, so maybe this should be a separate ticket to update the robot?

super_editor/test/src/default_editor/caret_test.dart
matthew-carroll: > Could you point me to where that is?

I couldn't find it, so maybe we never integrated a size parameter. We can deal with that later.

> Is this not already running on all platforms?

There are two different concepts of "running on a platform". There's the host machine, which is what you're describing. But then there's the runtime configuration, e.g., show drag handles on Android but show a vertical rectangle caret on Mac. That runtime configuration is determined by the platform variable within Flutter, not the host machine. For this reason, we have a number of test methods that configure the runtime platform as desired. Check out `test_tools.dart` at the top level of the `/test` directory.

super_editor/test/src/default_editor/caret_test.dart
matthew-carroll: We should include sufficient explanation in the test code to debug it when it fails, and adjust it when necessary.

super_editor/test/src/default_editor/caret_test.dart
matthew-carroll: I think we should add support for `TextAffinity` in the robot method(s).

super_editor/test/src/default_editor/caret_test.dart
matthew-carroll: If hardcoding a value is the only alternative, then that's what we should do. Testing that an implementation matches and implementation doesn't give us any value. If other tests do that, we'll need to improve those tests in the future.

If you hardcode a value, please be sure to explain how a developer can find that value again, if the test needs to be adjusted. For example, the default line spacing might change, which would require us to recompute that offset and update the test. We want it to be trivial for a developer to go through that process.

jmatth:
So taking a step back for a minute, I don't think the caret rendering tests belong in this PR. To fix the bug I changed how an `Offset` is translated to a `NodePosition`, such as when clicking or tapping on the document. Caret rendering is the other way around: given a `DocumentPosition`, figure out where to position the caret. I haven't touched that codepath.

It looks like the behavior I actually changed is tested in [`super_editor_robot_test.dart`][1]. I can add new tests to make sure for a given input (tap at the end of a wrapped line) we get the expected output (document selection has the correct offset and affinity). That will cover the changes introduced in this PR and prevent regressions to the old behavior.

[1]: https://github.com/superlistapp/super_editor/blob/1601fdce95ebfa34bddf80cab3e58e700b0edec3/super_editor/test/super_editor/supereditor_robot_test.dart#L11

matthew-carroll:
@jmatth I think that we should add affinity to both code paths at the same time. I don't think it makes sense to support the concept of affinity logically but not graphically, or in input but not output. Can you try to make the change across the board?

jmatth:
> I don't think it makes sense to support the concept of affinity logically but not graphically...

It's already supported graphically. It's admittedly a bit difficult to do but it's possible to get a text position with an upstream affinity on the main branch right now and it displays just fine.

I took another pass at testing and there are now tests for how `TextAffinity` is handled in three areas outlined below:

## Unit tests

The behavior that actually changed was on `TextNode` and `TextNodeSelection`, and those changes have been covered by unit tests since I opened this PR. Nothing really interesting here.

## Gesture Input

We want to verify that for a tap at a given location, the document selection is updated correctly. This shouldn't depend on the caret rendering.

I've added two tests to `supereditor_gestures_test.dart` and updated one in `supereditor_robot_test.dart` to handle this. Which file each test goes in is kind of an academic exercise though, as there's a circular dependency between those two tests: the robot tests depend on the gesture system, and the gesture tests use the robot.

## Caret rendering

We want to verify that the caret renders at the correct location given different text affinities. This shouldn't depend on the gesture system. This seems to belong in `caret_test.dart`, but the existing tests in there all use taps to change the selection. I wrote the new affinity tests to manipulate the selection directly through the `DocumentComposer` instead. This should also help alleviate the "the caret is displayed where the editor displays it" cyclical logic. There's still a bit of that in there since `SuperEditorInspector.calculateOffsetForCaret` depends on `DocumentLayout` to find where it thinks the caret should be, but I think it's fine. If that situation needs to be improved then there should just be more tests on those `DocumentLayout` methods so we can rely on them.

super_editor/lib/src/test/super_editor_test/supereditor_inspector.dart
```diff
@@ -119,6 +119,49 @@ class SuperEditorInspector {
...
+    if (composer == null) {
+      throw Exception('Could not find DocumentComposer in widget tree');
+    }
+    final previousSelection = composer.selection;
+
+    // Find the y offset of the caret when placed on the first line.
+    composer.selection = DocumentSelection.collapsed(
+      position: DocumentPosition(nodeId: nodeId, nodePosition: const TextNodePosition(offset: 0)),
+    );
+    await tester.pumpAndSettle();
```
matthew-carroll, line 134: I don't think we've relied on a `WidgetTester` anywhere else in the inspector. Can you make this work without that? Ideally, the inspector could be used within a regular app, not just tests.

super_editor/lib/src/test/super_editor_test/supereditor_inspector.dart
```diff
@@ -119,6 +119,49 @@ class SuperEditorInspector {
...
+    final firstLineCaretY = SuperEditorInspector.findCaretOffsetInDocument().dy;
+
+    // Walk the caret forward one offset at a time while keeping the affinity set to downstream. The offset where the
```
matthew-carroll, line 137: Instead of iterating character-by-character, can you query for the text position at a y-offset that's below the bottom of the character at the caret?

super_editor/test/src/default_editor/caret_test.dart
```diff
@@ -13,6 +15,126 @@ void main() {
     const textPosition = TextPosition(offset: 46);
     final tapPosition = DocumentPosition(nodeId: '1', nodePosition: TextNodePosition(offset: textPosition.offset));
 
+    group('affinity', () {
```
matthew-carroll, line 18: Let's change "affinity" to "text affinity"

super_editor/test/src/default_editor/caret_test.dart
```diff
@@ -13,6 +15,126 @@ void main() {
...
+        final docKey = GlobalKey();
+        final composer = DocumentComposer();
+        await tester.pumpWidget(
+          _createTestApp(
+            gestureMode: DocumentGestureMode.mouse,
+            docKey: docKey,
+            composer: composer,
+          ),
+        );
+        await tester.pumpAndSettle();
```
matthew-carroll, line 37: Generally we want to avoid `pumpAndSettle()`. Are you sure you need it here? If so, please add a comment about why it's required. Which piece of the layout required additional frames?

super_editor/test/src/default_editor/caret_test.dart
```diff
@@ -13,6 +15,126 @@ void main() {
...
+        final lineBreakOffset = await SuperEditorInspector.findOffsetOfLineBreak(tester);
+        composer.selection = DocumentSelection.collapsed(
```
matthew-carroll, line 39: Can you replace this direct selection manipulation with a user interaction? In general, tests should achieve their goals through user interaction, rather than manipulating the implementation details.

super_editor/test/src/default_editor/caret_test.dart
```diff
@@ -13,6 +15,126 @@ void main() {
...
+            nodeId: '1',
+            nodePosition: TextNodePosition(
+              offset: lineBreakOffset,
+              affinity: TextAffinity.upstream,
+            ),
+          ),
+        );
+        await tester.pumpAndSettle();
+
+        final caretOffset = SuperEditorInspector.findCaretOffsetInDocument();
```
matthew-carroll, line 50: It looks like you can call this method directly in the expectation, rather than use an intermediate variable.

super_editor/test/src/default_editor/caret_test.dart
```diff
@@ -13,6 +15,126 @@ void main() {
...
+        final docKey = GlobalKey();
+        final composer = DocumentComposer();
+        var app = _createTestApp(
+          gestureMode: DocumentGestureMode.mouse,
+          docKey: docKey,
+          composer: composer,
+        );
+        await tester.pumpWidget(app);
+        await tester.pumpAndSettle();
+        final unbrokenTextOffset = await SuperEditorInspector.findOffsetOfLineBreak(tester) - 1;
```
matthew-carroll, line 111: What does unbroken mean? Can you find a more descriptive name, or add a comment that clarifies what you're calculating?

super_editor/test/src/default_editor/caret_test.dart
```diff
@@ -13,6 +15,126 @@ void main() {
...
+          ..physicalSizeTestValue = screenSize;
+
+        final docKey = GlobalKey();
+        final composer = DocumentComposer();
+        var app = _createTestApp(
+          gestureMode: DocumentGestureMode.mouse,
+          docKey: docKey,
+          composer: composer,
+        );
+        await tester.pumpWidget(app);
```
matthew-carroll, line 109: Is there a reason that in this test you're assigning the app widget to an app variable?

super_editor/test/src/default_editor/caret_test.dart
```diff
@@ -13,6 +15,126 @@ void main() {
...
+            nodePosition: TextNodePosition(
+              offset: lineBreakOffset,
+              affinity: TextAffinity.upstream,
+            ),
+          ),
+        );
+        await tester.pumpAndSettle();
+
+        final caretOffset = SuperEditorInspector.findCaretOffsetInDocument();
+        final expectedCaretOffset = _computeExpectedDesktopCaretOffset(
```
matthew-carroll, line 51: We said before that `_computeExpectedDesktopCaretOffset()` is cyclical, right?

super_editor/test/src/default_editor/text_test.dart
```diff
@@ -301,6 +301,77 @@ void main() {
...
+            ),
+          );
+          expect(selectionWithDownstream.affinity, TextAffinity.downstream);
+        });
+      });
+    });
+
+    group('TextNodeSelection', () {
+      group('get base', () {
+        test('it preserves affinity', () {
```
matthew-carroll, line 356: You can remote the "it" from both of these test names, given how the overall name will be constructed.

super_editor/test/super_editor/supereditor_gestures_test.dart
```diff
@@ -344,5 +344,67 @@ spans multiple lines.''',
         ),
       );
     });
+
+    testWidgetsOnAllPlatforms("places the caret at the end of a wrapped line when tapping there", (tester) async {
+      // Configure and render a document.
+      await tester
+          .createDocument()
+          .withSingleParagraph()
+          .forDesktop()
```
matthew-carroll, line 353: If you're testing on all platforms then you shouldn't configure this editor for desktop. You should let it autoconfigure for the current platform.

super_editor/test/super_editor/supereditor_gestures_test.dart
```diff
@@ -344,5 +344,67 @@ spans multiple lines.''',
...
+          .autoFocus(true)
+          .withEditorSize(const Size(300, 700))
+          .withSelection(
+            const DocumentSelection.collapsed(
+                position: DocumentPosition(nodeId: '1', nodePosition: TextNodePosition(offset: 0))),
+          )
+          .pump();
+      await tester.pumpAndSettle();
+      final offset = await SuperEditorInspector.findOffsetOfLineBreak(tester);
```
matthew-carroll, line 362: I think we can skip this calculation entirely by pumping our own paragraph for these tests, commenting above that paragraph with the offset where you expect it to wrap, and then hardcoding this offset.

super_editor/test/super_editor/supereditor_robot_test.dart
```diff
@@ -67,22 +69,37 @@ void main() {
           .withSingleParagraph()
           .forDesktop()
           .pump();
+      await tester.pumpAndSettle();
 
       // Ensure that the document doesn't have a selection.
       expect(SuperEditorInspector.findDocumentSelection(), null);
 
+      final offsetBeforeLineBreak = await SuperEditorInspector.findOffsetOfLineBreak(tester) - 1;
```
matthew-carroll, line 77: Here again, I think we can hard code this value with a custom paragraph, including a comment that says where the line wrap is expected to appear.

super_editor/test/super_editor/supereditor_robot_test.dart
```diff
@@ -67,22 +69,37 @@ void main() {
...
+
       // Tap to place the caret in the first paragraph.
-      await tester.placeCaretInParagraph("1", 10);
+      await tester.placeCaretInParagraph("1", offsetBeforeLineBreak, affinity: TextAffinity.downstream);
 
       // Ensure that the document has the expected text caret selection.
+      final expectedSelection = DocumentSelection.collapsed(
```
matthew-carroll, line 83: Now that this test is testing and expecting two different results, please adjust the comment above this line to mention what's actually expected, rather than just "has the expected text caret selection"

super_editor/test/super_editor/supereditor_robot_test.dart
```diff
@@ -67,22 +69,37 @@ void main() {
...
-            nodePosition: TextNodePosition(offset: 10),
-          ),
-        ),
+        expectedSelection,
+      );
+
+      // Pause so we don't double-tap and select a word instead of placing the caret
+      await tester.pump(const Duration(seconds: 2));
+
+      // Tap to place the caret in the first paragraph with a different affinity.
```
matthew-carroll, line 97: Rather than say "different affinity" please say what the affinity is, and why.

super_editor/test/super_editor/supereditor_scrolling_test.dart
```diff
@@ -99,7 +99,10 @@ void main() {
         DocumentSelection(
           base: DocumentPosition(
             nodeId: lastParagraph.id,
-            nodePosition: lastParagraph.endPosition,
+            nodePosition: TextNodePosition(
+              offset: lastParagraph.endPosition.offset,
+              affinity: TextAffinity.upstream,
```
matthew-carroll, line 104: I'm noticing a lot of default value of `upstream` throughout this PR. Is there a reason that `upstream` should be the default? I would have expected `downstream` to be the default in most places.

super_editor/test/src/default_editor/caret_test.dart
jmatth: I think this is leftover from a previous revision and is no longer needed. I'll remove it and similar lines in the other tests.

super_editor/test/src/default_editor/caret_test.dart
jmatth: More leftovers from a previous attempt when I was having trouble getting the rendered editor to update. Removing.

super_editor/test/src/default_editor/text_test.dart
jmatth: That's going to be inconsistent with [the other tests] in this file but sure

[1]: https://github.com/superlistapp/super_editor/blob/48ad35a46eacd21b73e8b09432b938f7c750ed38/super_editor/test/src/default_editor/text_test.dart#L51-L52

super_editor/test/super_editor/supereditor_gestures_test.dart
jmatth: This causes the iOS and Android tests to fail. Looks like [`findCaretOffsetInDocument`] is broken when used on a document created with the utility methods from [`document_test_tools.dart`][2]. Gonna defer that for now and deal with the rest of the PR comments first.

[1]: https://github.com/superlistapp/super_editor/blob/f4d342c161432f3ccf94c72d251f72eb1ae58754/super_editor/lib/src/test/super_editor_test/supereditor_inspector.dart#L43-L49
[2]: https://github.com/superlistapp/super_editor/blob/f4d342c161432f3ccf94c72d251f72eb1ae58754/super_editor/test/super_editor/document_test_tools.dart

super_editor/test/super_editor/supereditor_scrolling_test.dart
jmatth: It looks like flutter reports either upstream or downstream depending on where exactly a gesture lands, even if it's not on a line break. You can see this in the example app if I log each time the selection changes:


https://user-images.githubusercontent.com/1316184/190299418-9cb5013a-8429-461f-abe9-a09e90e368cb.mp4

super_editor/test/src/default_editor/caret_test.dart
jmatth: An offset not at a line break. I'll change the name

super_editor/test/src/default_editor/caret_test.dart
jmatth: I don't think we should do that. Tests for how the editor is rendered shouldn't be coupled to the gesture system.

super_editor/test/super_editor/supereditor_gestures_test.dart
jmatth: Based on [this comment thread] I thought we wanted to avoid using magic values in tests as much as possible, which is why I added `findOffsetOfLineBreak`. Please confirm if that's not the case before I go and switch everything back to magic sizes and offsets.

[1]: https://github.com/superlistapp/super_editor/pull/602#discussion_r967493299

super_editor/lib/src/test/super_editor_test/supereditor_inspector.dart
jmatth: Deferring this until [this comment is addressed][1], as we can probably just remove this method entirely if we are switching back to magic values everywhere.

[1]: https://github.com/superlistapp/super_editor/pull/602#discussion_r971479102

super_editor/test/src/default_editor/caret_test.dart
jmatth: Gonna defer on this one too until I get an affirmative yes or no on whether magic values are the preferred way to write tests now. Though if the answer is yes, maybe these should just be golden tests instead. "Given these parameters, expect this arbitrary rendering output" sounds more like a golden than an widget test to me.

matthew-carroll: There's no preference either way. It's a question of which approach confirms that the code achieves the desired result. Cyclical verification is almost never useful because any changes to the implementation, by definition, will continue to pass the test, even if it breaks the feature. Golden tests have their own challenges because Mac, Windows, and Linux host machines all render different pixels. Golden tests might be a good option for a number of these tests, but I think we still lack a reliable way to write and run those tests. If we can get Golden testing infrastructure to a reliable place, then I'm open to moving a number of verifications over to Golden tests.

super_editor/test/src/default_editor/caret_test.dart
matthew-carroll: I disagree. I don't think the distinction is between "rendering" and "gestures". The critical distinction in tests is between "specified behavior" and "implementation details". From a user-specified behavior standpoint, the fact that there's an option called a `DocumentComposer`, which accepts a value called a `DocumentSelection` is an implementation detail. On the other hand, regardless of implementation, we know that `SuperEditor` must always provide a way for a user to place a caret. The way that happens is encapsulated within the test robot method for selection. Thus, no matter what happens to `SuperEditor`'s implementation, we will always have a test robot method that can place a caret at a desired offset, which means that we can safely use said method with a strong promise of stability moving forward.

This analysis is very different for unit tests and component tests, but what we're writing here is closer to a smoke test. We're testing the whole `SuperEditor` structure from a user perspective. Therefore, any exposure of implementation details is regrettable, and we should try to avoid them.

super_editor/test/super_editor/supereditor_gestures_test.dart
matthew-carroll: I mentioned this in another response, but I'll repeat it here. There aren't any rules about "always", "never", or "usually" using magic numbers vs calculations. On a case-by-case basis we need to determine whether a calculation preserves the goal of the test, or renders the test anemic.

The most important detail is cyclic verification. We never want to cyclic verification. If you find yourself writing such a calculation, then you should find some other approach. Maybe it's a magic number, maybe it's something else. But we certainly don't want a cyclic calculation.

In this case, I don't think the calculation is cyclic. However, it does look like the introduction of this calculation fundamentally changes the direction of `SuperEditorInspector`, as I've mentioned elsewhere. It just so happens that in this case it should be quite simple and easy to use a specific paragraph of text and a specific window size to give us a static offset that we can hard-code. It should also be easy to add a comment near the paragraph of text which clearly explains where the line-break is supposed to appear, and which text offset references that line-break. I've done this same thing in other tests.

So in this case, a magic number keeps us from altering the basis of `SuperEditorInspector` and it does so without creating a situation that will confuse other developers, so it's a reasonable path forward.

super_editor/test/super_editor/supereditor_scrolling_test.dart
matthew-carroll: Can you elaborate on the conclusion that we should draw from that video?

super_editor/test/super_editor/supereditor_gestures_test.dart
matthew-carroll: Do you know the nature of the failure? Is it unimplemented, or is the implementation wrong?

super_editor/test/src/default_editor/text_test.dart
matthew-carroll: We're on a long road to improve a number of testing practices. Some early missteps came from me, some from others. We're going to have inconsistencies until we reach a better alignment...

super_editor/test/super_editor/supereditor_scrolling_test.dart
jmatth: Sure, to answer the original question:

> Is there a reason that upstream should be the default?

It's not the default, these tests just happen to use gesture coordinates that cause the underlying call to [`RenderParagraph.getPositionForOffset`][1] to return an upstream affinity. With the exception of tests I added, I just went through and updated expect statements to match the new affinities.

super_editor/test/src/default_editor/caret_test.dart
jmatth: I wouldn't call DocumentComposer an implementation detail. It's the public API for manipulating text selection, and we should have tests to ensure that the UI renders the selection properly.

super_editor/test/super_editor/supereditor_gestures_test.dart
jmatth: Alternate idea: if we don't need these tests to be super precise we can avoid magic numbers by writing a single test that computes both the upstream and downstream caret position at a line break and then checks that the downstream position is below and to the left of the upstream position. Something like:

```
expect(upstreamCaretOffset.dx, greaterThan(downstreamCaretOffset.dx));
expect(upstreamCaretOffset.dy, lessThan(downstreamCaretOffset.dy));
```

super_editor/test/src/default_editor/caret_test.dart
matthew-carroll: It is an implementation detail in comparison to the user-facing specified behavior. A user doesn't know anything about a 'DocumentComposer', nor does this test need to know about a `DocumentComposer`. I've done my best to communicate that point. If you feel differently, then we'll have to agree to disagree. I'd like for Super Editor tests to leverage robot behaviors wherever possible, when testing a `SuperEditor` widget as a whole.

super_editor/test/super_editor/supereditor_gestures_test.dart
matthew-carroll: The offset comparison approach is reasonable, but I think it still needs more confidence. The expectations that you described could be triggered by something like floating point variations. We need a high confidence signal that the difference is at least close to what it should be.

super_editor/test/super_editor/supereditor_scrolling_test.dart
matthew-carroll: Ok, sounds good.

super_editor/test/super_editor/supereditor_gestures_test.dart
jmatth: The inspector can't find the caret:

```
The following StateError was thrown running a test:
Bad state: No element

#0      Iterable.single (dart:core/iterable.dart:654:25)
#1      SuperEditorInspector.findCaretOffsetInDocument (package:super_editor/src/test/super_editor_test/supereditor_inspector.dart:44:61)
#2      main.<anonymous closure>.<anonymous closure> (file:///Users/joshuamatthews/projects/forks/super_editor/super_editor/test/super_editor/supereditor_gestures_test.dart:364:45)
<asynchronous suspension>
<asynchronous suspension>
(elided one frame from package:stack_trace)

The test description was:
  places the caret at the end of a wrapped line when tapping there (on iOS)
```

Here's a link to the exact line that's throwing: [`supereditor_inspector.dart:44`][1].

Looks like the real cause of the failure is having gesture mode set to anything other than `mouse`, which causes [this switch][2] to return Android or iOS gesture systems that don't render the caret with the [magic key][3] that inspector depends on.

This isn't actually a problem in these tests now since `findOffsetOfLineBreak` doesn't rely on the finding the caret. It does cause issues using the inspector in `caret_test.dart`.

[1]: https://github.com/superlistapp/super_editor/blob/b9f868766767756ad589984d74089ef780809b95/super_editor/lib/src/test/super_editor_test/supereditor_inspector.dart#L44
[2]: https://github.com/superlistapp/super_editor/blob/b9f868766767756ad589984d74089ef780809b95/super_editor/lib/src/default_editor/super_editor.dart#L540-L574
[3]: https://github.com/superlistapp/super_editor/blob/b9f868766767756ad589984d74089ef780809b95/super_editor/lib/src/default_editor/document_caret_overlay.dart#L168

jmatth:
@matthew-carroll ready for another look. The current state is

- `findOffsetOfLineBreak` no longer relies on the widget tester, so all the tests are still using this method rather than hard-coding the offset of the line break. Let me know if we should still go with hard-coding for some reason, but the affinity tests will fail if `findOffsetOfLineBreak` develops a bug and returns an offset not at a line break so I think this is fine
- I updated `findCaretOffsetInDocument` on the inspector to support finding Android and iOS carets. It just checks for desktop, Android, and iOS carets in sequence and returns the offset for the first one it finds
- The affinity tests in `caret_test.dart` test for the upstream/downstream caret offsets relative to the offset of the caret when placed at the start of the paragraph

[1]: https://github.com/superlistapp/super_editor/blob/b9f868766767756ad589984d74089ef780809b95/super_text_layout/lib/src/text_layout.dart#L375

super_editor/test/src/default_editor/caret_test.dart
```diff
@@ -13,6 +16,62 @@ void main() {
...
+          (WidgetTester tester) async {
+        await tester //
+            .createDocument()
+            .withSingleParagraph()
+            .withEditorSize(editorSize)
+            .pump();
+
+        await tester.placeCaretInParagraph('1', 0);
+        final startOfFirstLineCaretOffset = SuperEditorInspector.findCaretOffsetInDocument();
+        final lineBreakOffset = SuperEditorInspector.findOffsetOfLineBreak('1');
```
matthew-carroll, line 39: Can you add a blank line between each assignment? It's difficult to see what's being collected here. It would also be helpful to lead each of those sections with a comment about whether you're trying to place the caret at the beginning of a line, or end of a line, or whatever offset is relevant to the verifications.

super_editor/test/src/default_editor/caret_test.dart
```diff
@@ -13,6 +16,62 @@ void main() {
...
+          (WidgetTester tester) async {
+        await tester //
+            .createDocument()
+            .withSingleParagraph()
+            .withEditorSize(editorSize)
+            .pump();
+
+        // Find an offset that is not at a line break, so that the caret should render the same with upstream or
+        // downstream affinity.
+        final textOffset = SuperEditorInspector.findOffsetOfLineBreak('1') - 1;
```
matthew-carroll, line 63: Do we need any sophisticated calculations for this test? Would placing the caret at the beginning of the text give us the same verification, or is there something more to this test?

super_editor/test/src/default_editor/caret_test.dart
```diff
@@ -13,6 +16,62 @@ void main() {
...
+
+        await tester.placeCaretInParagraph('1', textOffset, affinity: TextAffinity.downstream);
+        final downstreamCaretOffset = SuperEditorInspector.findCaretOffsetInDocument();
+        await tester.pump(const Duration(seconds: 2));
+        await tester.placeCaretInParagraph('1', textOffset, affinity: TextAffinity.upstream);
```
matthew-carroll, line 68: Here again, can you add a blank line between actions so that it's clear where one begins and ends?

super_editor/test/src/default_editor/caret_test.dart
```diff
@@ -13,6 +16,62 @@ void main() {
...
+            .withEditorSize(editorSize)
+            .pump();
+
+        // Find an offset that is not at a line break, so that the caret should render the same with upstream or
+        // downstream affinity.
+        final textOffset = SuperEditorInspector.findOffsetOfLineBreak('1') - 1;
+
+        await tester.placeCaretInParagraph('1', textOffset, affinity: TextAffinity.downstream);
+        final downstreamCaretOffset = SuperEditorInspector.findCaretOffsetInDocument();
+        await tester.pump(const Duration(seconds: 2));
```
matthew-carroll, line 67: Is there a reason that you're pumping for 2 seconds? If you're trying to avoid a double tap, I think there's a constant somewhere for the min tap time. Either way, please a comment for any pump with a specified duration so that we always know why it was selected.

super_editor/test/super_editor/supereditor_robot_test.dart
```diff
@@ -71,18 +73,39 @@ void main() {
       // Ensure that the document doesn't have a selection.
       expect(SuperEditorInspector.findDocumentSelection(), null);
 
-      // Tap to place the caret in the first paragraph.
-      await tester.placeCaretInParagraph("1", 10);
+      final offsetBeforeLineBreak = SuperEditorInspector.findOffsetOfLineBreak('1') - 1;
```
matthew-carroll, line 76: Is there a reason that this was changed? This test already existed, right? If you want to test something related to affinity, is there a reason that's not a new test?

super_editor/test/src/default_editor/caret_test.dart
jmatth: It is to avoid a double tap. I think I see the constant, so I'll change these calls to use that and add comments to each line about why the pause is there.

super_editor/test/src/default_editor/caret_test.dart
jmatth: I think that should be fine, or just use an offset of 1 to make sure it's in the middle of some text.

Bigger issue: I think this test is actually invalid. We're using the robot and according to [this other test I wrote][1], `placeCaretInParagraph` produces a selection with the same affinity when on an offset not at a line break. So without manipulating the selection directly, I don't see a way to test that the same offset with different affinities produces the same caret coordinates.

[1]: https://github.com/superlistapp/super_editor/blob/4c4d5eaa680e34ef59412276020f373489895221/super_editor/test/super_editor/supereditor_robot_test.dart#L102-L110

super_editor/test/super_editor/supereditor_robot_test.dart
jmatth: I initially wrote a new test but half of it was just copy-pasted from the existing one, and I wanted to avoid a situation where we have an extra test failing for the exact same reason and just adding noise for whoever has to debug the issue.

super_editor/test/src/default_editor/caret_test.dart
matthew-carroll: Can you add affinity as a named parameter in `placeCaretInParagraph`? I assume that would let you specify what you need...

super_editor/test/super_editor/supereditor_robot_test.dart
matthew-carroll: The old test vs new test don't seem like tests that would break for the exact same reason. This test is called "taps to place caret in middle of paragraph". Isn't it possible that caret placement in the middle of a paragraph works just fine, but placing a caret at a soft line-break doesn't work? If so, then I think those are two different tests. 

From a pragmatic point of view, if another developer ends up fixing this test for some reason, I don't know how that developer will understand the goal of the test. We're very carefully placing the caret at the end of a line, and we're heavily focusing on affinity, neither of which are communicated with the name of the test. At a minimum, I think that's an issue we need to resolve.

jmatth: I just pushed a commit moving the affinity stuff into a separate test. The part I was hung up on was checking that placing the caret produced an expected selection. In the new version I just left that `expect` call out and the test only compares the selections from upstream and downstream gestures to each other.

super_editor/test/src/default_editor/caret_test.dart
matthew-carroll: Can you summarize where we are with this thread? I'm not sure I'm following it.

You mentioned above that you think this test in `caret_test` is invalid. Then you linked to a [robot test](https://github.com/superlistapp/super_editor/blob/4c4d5eaa680e34ef59412276020f373489895221/super_editor/test/super_editor/supereditor_robot_test.dart#L102-L110). But based on your latest comment on this thread, it sounds like it's actually the robot test that's invalid. If we can't control the affinity that's returned by the engine, then we probably shouldn't have a robot test that implies we can control the affinity, right?

jmatth: For a given offset `n` that is not at a line break, `tester.placeCaretInParagraph(n, affinity: ...)` will produce the same selection regardless of the specified affinity. The current test expects this behavior. This is a bug in the robot, but not one I can see a good way to work around. Modifying the tap coordinates to shift them left or right by a single logical pixel depending on whether the specified affinity is upstream or downstream seems to work. I'll implement that as a fix for now, but it may break down the line if the robot is used in actual integration tests that use other rendering engines or real fonts that aren't just monospaced boxes.

matthew-carroll: "will produce the same selection regardless of the specified affinity" - didn't you say at the beginning of this work that affinity is only supposed to mean something at line breaks? Can you describe a specific example that demonstrates this robot failure? I'm still not clear what the problem is.

jmatth: > didn't you say at the beginning of this work that affinity is only supposed to mean something at line breaks?

And RTL/LTR boundaries, yes. The test this thread is on exists make sure that for a non-linebreak offset, the caret is rendered the same at that offset regardless of affinity. To do that I need a way to set the affinity at a given none-linebreak offset. You said [over here][1] these tests need to use the robot instead of just setting the selection directly, so here we are.

[1]: https://github.com/superlistapp/super_editor/pull/602#discussion_r973667981

matthew-carroll: Each time I read this thread, I come away with the opposite understanding of what I thought you were saying in the previous description of the problem. Let's pause on these links into the code, and links to other comments. It's only confusing me more.

In response to this comment, please describe the explicit problem as you see it. Is it a problem at line breaks, or is it a problem at non-line breaks? Also, please demonstrate an example of the problem in this thread. Don't link me to an adjusted test, just show some example text right here, along with a caret offset, and a description of "expected" vs "actual" behavior. I need a firm understanding of the issue before I can review the attempt to fix it, or work around it. For me, it looks like that is going to require a clear and explicit inline example.

jmatth: Expected: if I click on the end of a wrapped line, the caret should be placed at the end of that line

Actual: The caret is placed one character upstream from the end of the line


https://user-images.githubusercontent.com/1316184/193367812-617980c5-e011-4246-ab7e-4b13f2bbb87d.mov

Expected: If I click between two characters, the caret should be placed between those characters

Actual: if the click occurred to the left of the halfway point between the two characters, the caret is placed one character upstream from where it should be

https://user-images.githubusercontent.com/1316184/193367980-2973bcd2-7963-439c-928d-51d6d3df0c1b.mov

matthew-carroll: Thanks, those problems are now very clear.

Followup: did the two issues in those videos exist before this PR, or were they introduced by this PR?

jmatth: Those issues predate this PR. This PR fixes them. See also our previous discussion of the root cause at #549.

matthew-carroll: If this PR fixes this issue, then what are we discussing here? Sounds like we're still not talking about the right thing.

jmatth: Earlier in this PR you asked me to add affinity related tests for the caret rendering. I'm trying to add a test to check that for a text offset not at a linebreak, the caret is rendered the same for both upstream and downstream affinity. In order to write such a test, I need a way to set the text selection to the same offset but different affinities while not at a line break. Doing so through gestures via the robot involves manipulating the tap coordinates slightly to the left or right, depending on the desired affinity. I'm worried this method of setting a selection with a specific affinity could break in a variety of ways, but for the sake of getting this fix merged that's how I've updated the robot to work as of the latest commit.

matthew-carroll (approved):
Generally LGTM - I think I'm gonna need to get hands on with the affinity in real uses to see if it all makes sense, but we'll adjust later.

FYI @jmatth - we're pushing towards a stable release before introducing breaking changes. After the next release, and after merging a few of those breaking changes, I'd like to go ahead and get your Melos work merged in.

PR #608, opened 2022-06-08
Issue #351 - pageUp, pageDown, Home and End keys implemented.  Function keys no longer case '?' character to display.

faithoflifedev:
Matt's example of a helpful PR description for these changes:

This PR implements desktop keyboard behaviors for PAGE UP, PAGE DOWN, HOME, and END:
 * PAGE UP: scrolls up by the viewport height, without moving the selection
 * PAGE DOWN: scrolls down by the viewport height, without moving the selection
 * HOME: scrolls to the beginning of the document, without moving the selection
 * END: scrolls to the end of the document, without moving the selection

Each of these new behaviors require knowledge of the viewport size and control over the scroll offset. Therefore, `EditContext` is updated to include a `ScrollController`. Now, all keyboard handlers have access to `SuperEditor`'s scrolling system.

**Question** for @faithoflifedev - is it true that PAGE UP and PAGE DOWN shouldn't move the selection? When I press those on my Mac in my IDE, they move the caret up/down.


matthew-carroll:
@faithoflifedev - I see that you filed a new PR. Is this PR intended to replace the existing PR? If so, can you close the existing PR? Also, typically one would update the existing PR so that we have continuity of conversation.

At first glance, I see that this PR doesn't have a title or description that describes what problem you're solving, or how. In the interest of getting in the habit of doing that, can you adjust those appropriately?

I'll review the content of the PR as soon as I can.

matthew-carroll:
@faithoflifedev any updates?

faithoflifedev:
Sorry for the delay, I've been very busy. I'll take a look this weekend.

faithoflifedev:
Existing PR closed.  Sorry about that.

Problems solved with this PR:

 1. Make sure the function keys do not produce question-mark-box characters
 2. Implement page-up and page-down to scroll one full viewport length up or down. The caret stays at its current position and is not moved
 3. Implement [home] and [end], scroll to the first line or last line of the document. The caret stays at its current position and is not moved.
 4. double-check that on caret movement (e.g. cursor keys or entering a character) the viewport is scrolled back to have the caret visible.

matthew-carroll:
@faithoflifedev thanks for cleaning things up. Can you also add a brief description to the PR that explains how you solved the non-trivial tasks? For example, I know you had to adjust some things to get page-up and page-down to work. Can you describe what you did, and why, as well as any alternative approaches that you may have rejected, or any road blocks you discovered along the way?

Also, if you believe your PR is ready for review, please switch it out of draft mode so that CI can run your tests.

I'll take a look at the implementation after you add that description and CI shows green.

faithoflifedev:
The main change required to allow for the new key based navigation options was accomplished through the addition of a `scrollController` property to the `EditContext` class.  This change allows for the creation of an  `ExecutionInstruction` that makes changes to the scroll position of the viewport in addition to the underlying document model.

As a side note, I included an `ExecutionInstruction` that was meant to prevent a unknown '?' character from appearing when a Function key was pressed.  However, I noticed that once I merged the latest code from the main branch into mine, the issue was no longer there.  I've commented out the `doNothingWhenFnKeyPressed` from super_editor.dart, for the time being as the functionality no longer appears to be necessary.

matthew-carroll:
@faithoflifedev I'm going to review your implementation in a minute. With regard to the description you submitted as a comment - the description of the PR should go in the PR description.

Your comment description lists some technical details, which is a good start, but the description doesn't really frame the big picture of this PR. I'm going to replace your PR description with an example of what I would prefer for a PR like this, which helps me more quickly review such a PR. I'll update that description after I review your implementation.

super_editor/test/src/default_editor/navigation_keys_test.dart
```diff
@@ -0,0 +1,205 @@
...
+
+        // Let the scrolling system auto-scroll, as desired.
+        await tester.pumpAndSettle();
+
+        // Ensure that the scroll position is set to the expected position
+        // after the key press action.
+        expect(scrollState.widget.controller!.offset, equals(scrollState.position.maxScrollExtent));
+
+        // Ensure that the caret has not move during the key press action.
+        expect(textNodePosition.offset, equals(0));
```
matthew-carroll, line 29: What is this line checking? I don't see where `textNodePosition` came from, and I'm not sure what else you'd need to verify other than what you already checked above this.

super_editor/test/src/default_editor/navigation_keys_test.dart
```diff
@@ -0,0 +1,205 @@
...
+      });
+
+      testWidgets('pageDown key press at top of viewport', (tester) async {
```
matthew-carroll, line 32: I added a comment up above about grouping and naming for these tests. You should apply that across the board. But also I want to clarify something by looking at this test name. Ideally, a test description describes the behavior that the test is verifying, not an action that a test happens to take. Here's a more appropriate description:

"PAGE DOWN scrolls down by the viewport height"

super_editor/test/src/default_editor/navigation_keys_test.dart
```diff
@@ -0,0 +1,205 @@
...
+
+        // Ensure that the caret has not move during the key press action.
+        expect(textNodePosition.offset, equals(0));
+      });
+    });
+  });
+}
+
+// Initial screen dimensions for test document
+const double screenWidth = 640.0;
```
matthew-carroll, line 171: Do we need to define the screen width/height, or can we write these tests in a way that they can work with any given width/height?

super_editor/test/src/default_editor/navigation_keys_test.dart
```diff
@@ -0,0 +1,205 @@
...
+
+TextNodePosition get textNodePosition => composer.selection!.extent.nodePosition as TextNodePosition;
+
+Future<void> _pumpDocumentLayout(WidgetTester tester) async {
+  tester.binding.window
+    ..physicalSizeTestValue = const Size(screenWidth, screenHeight)
+    ..platformDispatcher.textScaleFactorTestValue = 1.0
+    ..devicePixelRatioTestValue = 1.0;
+
+  await tester.pumpWidget(
```
matthew-carroll, line 191: Can these tests be written with the new document test tools, e.g., the `DocumentTester` APIs? Those APIs are meant to provide a standard way to configure a document and an editor and pump it. We'd like to move to those APIs, where possible.

matthew-carroll (commented):
I left a number of comments. Some of the comments apply to multiple places. Please take another pass and then I'll review again. If you have questions, please let me know.

super_editor/test/src/default_editor/navigation_keys_test.dart
faithoflifedev: The `DocumentTester` API will require a `ScrollController` for the `EditContext`.  I recommend initially testing as is.  Is there any example code for the API?

super_editor/lib/src/default_editor/document_keyboard_actions.dart
```diff
@@ -12,6 +14,52 @@ import 'document_input_keyboard.dart';
...
+ExecutionInstruction scrollOnPageUpPageDownHomeAndEnd({
+  required EditContext editContext,
+  required RawKeyEvent keyEvent,
+}) {
+  if (keyEvent.logicalKey.keyId < LogicalKeyboardKey.end.keyId ||
+      keyEvent.logicalKey.keyId > LogicalKeyboardKey.pageUp.keyId) {
+    return ExecutionInstruction.continueExecution;
+  }
+
+  final scrollController = editContext.scrollController!;
```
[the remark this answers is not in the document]
faithoflifedev: https://github.com/superlistapp/super_editor/blob/a67b332367d33e1cb4fc295f540635a897853dab/super_editor/lib/src/default_editor/super_editor.dart#L159-L164

`ScrollController` is nullable, I will update the code appropriately.  

Also it's unclear to me what "`scrollController` is not used if this `SuperEditor` has an ancestor  `Scrollable`." means.

super_editor/test/src/default_editor/navigation_keys_test.dart
```diff
@@ -0,0 +1,205 @@
...
+void main() {
+  group('Keyboard based navigation', () {
+    group('Page navigation keys', () {
+      testWidgets('pageDown key press at bottom of viewport', (tester) async {
+        await _pumpDocumentLayout(tester);
+
+        final scrollState = tester.state<ScrollableState>(find.byType(Scrollable));
+
+        // Set the ScollPostion to the bottom of the scrollable area
+        scrollState.widget.controller!.jumpTo(scrollState.position.maxScrollExtent);
```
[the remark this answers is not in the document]
faithoflifedev: Done - https://github.com/superlistapp/super_editor/issues/644

super_editor/test/src/default_editor/navigation_keys_test.dart
faithoflifedev: One of the initial requirements of issue #351 was that the caret not move when any of navigation keys where activates.  This check was meant to confirm that the caret had not moved from its initial position.

However since this is the default expected behaviour in this case, I'll remove the check.

super_editor/test/src/default_editor/navigation_keys_test.dart
faithoflifedev: Overall, the test code needs to know the screen height to be able to determine if a PAGE UP or PAGE DOWN key press moves the `scrollController` the correct amount.  I'm using the code:

          tester.binding.window
            ..physicalSizeTestValue = screenSizeForTest
            ..platformDispatcher.textScaleFactorTestValue = 1.0
            ..devicePixelRatioTestValue = 1.0;

to set the parameters for the test window, similar to what's in use here - https://github.com/superlistapp/super_editor/blob/1dd77e05b4df99b250ba31cca79acb705227d2c5/super_editor/test/src/default_editor/auto_scroll_test.dart#L16-L20

faithoflifedev:
Hi @matthew-carroll, I've implemented most of the changes that you've suggested, answered questions were asked and asked a few of my own. 

The tests are now working again.  Sorry about that.

rutvik110:
hey @faithoflifedev, there are few conflicts in the PR. Is it possible for you to resolve these conflicts? 
If not, could you please tick the "Allow edits from maintainers" box so I can go ahead and make those changes myself and push the PR forward.

rutvik110:
hey @matthew-carroll , brought the PR upto the latest updates from main and also did a few updates from my side, 

* Fixed a bug which resulted in scrolling the page twice on both `key down` and `key up` event on page-down/up actions.
* Fixed broken tests and using latest test builder api's for tests.
* Moved page scrolling tests to `document_keyboard_actions_test.dart` tests suite.
* Refactored a few areas.

Document is scrolled up/down by viewport height on page up/down key press respectively. And scrolled to the top/bottom on HOME/END key press respectively. 

Latest Demo:

https://github.com/superlistapp/super_editor/assets/65209850/435596e8-d415-48fa-8846-8e3e03f2b249

super_editor/lib/src/core/edit_context.dart
```diff
@@ -23,6 +25,7 @@ class SuperEditorContext {
     required DocumentLayout Function() getDocumentLayout,
     required this.composer,
     required this.commonOps,
+    this.scrollController,
```
matthew-carroll, line 28: My initial thought is that we don't want this in the `SuperEditorContext` but I haven't come up with a better option yet. I'll think about it.

super_editor/lib/src/core/edit_context.dart
```diff
@@ -11,6 +11,8 @@ import 'editor.dart';
 /// [DocumentComposer], and [DocumentLayout].
 /// In addition, [commonOps] are available for directly applying common, complex
 /// changes to the document using the artifacts.
+/// A [scrollController] is available to allow changes to the scroll position in
```
matthew-carroll, line 14: As a general note, please add a blank line between paragraphs in comments.

super_editor/lib/src/core/edit_context.dart
```diff
@@ -44,4 +47,8 @@ class SuperEditorContext {
   /// Common operations that can be executed to apply common, complex changes to
   /// the document.
   final CommonEditorOperations commonOps;
+
+  /// The scroll position state of the viewport, used to apply position changes to
```
matthew-carroll, line 51: Notice that this Dart Doc is both technically incorrect, and also overly verbose.

Let's start with "The scroll position state of the viewport". First, a `ScrollPosition` is a real thing, and this isn't a `ScrollPosition`, so that's not quite accurate. Also, the word "state" there is confusing. It seems like an extra word that confuses the topic.

Second, "used to apply position changes to the display". In general, we don't want to be overly constraining in our documentation. While it's possible to use a `ScrollController` to change a scroll offset, that's not the only thing that one can do with it. We shouldn't declare a more narrow purpose. It's capable of doing all things that any `ScrollController` can do, and for the same reasons.

A more concise and accurate Dart Doc might be the following:

```dart
/// [ScrollController] that's attached to `SuperEditor`s viewport.
```

Moreover, I now see a missing detail that should be covered in that Dart Doc that I don't know how to answer. Does this `ScrollController` attach to ancestor scrollables, too?

When a `SuperEditor` is placed inside of another `Scrollable`, `SuperEditor` detects that, and controls the ancestor `Scrollable` instead of adding its own `Scrollable`. This is because we don't want a `Scrollable` in a `Scrollable`. How does this `ScrollController` interact with that behavior.

There's a demo of this behavior in the example app, if you want to verify behavior.

On a related note, please make sure that we have tests for the key behavior that also tries a `SuperEditor` inside of another `Scrollable`.

super_editor/lib/src/default_editor/document_hardware_keyboard/document_keyboard_actions.dart
```diff
@@ -11,6 +12,119 @@ import 'package:super_editor/src/default_editor/text.dart';
 import 'package:super_editor/src/infrastructure/_logging.dart';
 import 'package:super_editor/src/infrastructure/keyboard.dart';
 
+/// Scroll to appropriate position in viewport when the PAGE UP key is pressed.
```
matthew-carroll, line 15: Please make sure to never describe something as "appropriate" or "expected" or "desired". The reader has now way to know what that means. Dart Docs need to specify the actual behavior, rather than allude to it.

super_editor/lib/src/default_editor/document_hardware_keyboard/document_keyboard_actions.dart
```diff
@@ -11,6 +12,119 @@ import 'package:super_editor/src/default_editor/text.dart';
...
+///
+/// PAGE UP: moves the scroll position in the viewport up by an amount equal to
```
matthew-carroll, line 17: These lines down here include two pieces: the description of primary behavior, and an explanation of an exceptional scenario.

The primary behavior should be included in the original documentation line, which removes the need to also state it down here.

The exceptional case clarification can remain below the first line. I would probably state it as the following:

"If there is not a viewport worth of distance to scroll up, the viewport scrolls as far as up as possible."

The term `minScrollExtent` is a property of a `ScrollPosition`, which is a property of a `ScrollController`, which is a property of a `SuperEditorContext`. We shouldn't just drop that term in docs and expect that the reader has any idea what you're referencing. Instead, we should focus on the observable behavior, aside form technical properties and classes. 

In cases where it is important to refer to specific properties and classes, you should use a full reference, e.g., `[SuperEditContext.scrollController.position.minScrollExtent]`, so that the reader knows exactly what you're talking about. And, if a technical reference is used, it's often a good idea to provide some amount of context as to what that thing is and why the reader cares.

super_editor/lib/src/default_editor/document_hardware_keyboard/document_keyboard_actions.dart
```diff
@@ -11,6 +12,119 @@ import 'package:super_editor/src/default_editor/text.dart';
...
+  }
+
+  if (keyEvent is RawKeyUpEvent) {
+    return ExecutionInstruction.continueExecution;
+  }
+
+  final scrollController = editContext.scrollController!;
+
+  // Scroll up for one viewport length of the document.
+  scrollController.jumpTo(
```
matthew-carroll, line 36: Should this be `jumpTo`? Do other apps jump, or do they animate?

super_editor/lib/src/default_editor/document_hardware_keyboard/document_keyboard_actions.dart
```diff
@@ -11,6 +12,119 @@ import 'package:super_editor/src/default_editor/text.dart';
...
+
+  final scrollController = editContext.scrollController!;
+
+  // The user pressed the END key, scroll to the bottom of the document.
+  scrollController.jumpTo(scrollController.position.maxScrollExtent);
+
+  return ExecutionInstruction.haltExecution;
+}
+
+/// Without this action in place pressing a function key would display
```
matthew-carroll, line 114: This Dart Doc jumps into a justification before it ever mentions what it actually does.

Remember, the first line should be a concise statement of the behavior that you're documenting. Below that, you can elaborate on exceptions, explanations, justifications, etc.

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -844,7 +845,12 @@ final defaultComponentBuilders = <ComponentBuilder>[
 /// Keyboard actions for the standard [SuperEditor].
 final defaultKeyboardActions = <DocumentKeyboardAction>[
   toggleInteractionModeWhenCmdOrCtrlPressed,
+  doNothingWhenFnKeyPressed,
```
matthew-carroll, line 848: By placing this handler at the top of the priority queue, this means that nothing is allowed to ever respond to a function key, right? Is that your intention?

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
```diff
@@ -976,6 +976,217 @@ void main() {
           ),
         );
       });
+      group("page scrolling", () {
```
matthew-carroll, line 979: Please make sure to include blank lines between groups and tests.

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
```diff
@@ -976,6 +976,217 @@ void main() {
...
+        testWidgets('PAGE DOWN does not scroll past bottom of the viewport', (tester) async {
```
matthew-carroll, line 980: In general, you should never use `testWidgets()`. That method configures Flutter as if it's running on Android, and only Android. Therefore, we lose out on all platform differences, and everything behaves like a mobile device.

When writing a widget test, you first need to decide which platforms it should apply to.

Should Windows, Mac, and Linux all behave the same way with these keys? I thought I remembered some discussion in the original ticket that said Windows and Mac do different things. But maybe I'm misremembering.

Similarly, if a Bluetooth keyboard is plugged into a iPhone, iPad, or Android device, should this behavior be exactly the same?

After you investigate that question, if the answer is that all platforms should do the exact same thing, then you should the `testWidgetsOnAllPlatforms()`. That test method will run the test 5 times, once for each platform.

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
```diff
@@ -976,6 +976,217 @@ void main() {
...
+
+          await tester.tapAtDocumentPosition(
+            const DocumentPosition(
+              nodeId: '1',
+              nodePosition: TextNodePosition(offset: 0),
+            ),
+          );
+          await tester.pump();
+
+          final scrollState = tester.state<ScrollableState>(find.byType(Scrollable));
```
matthew-carroll, line 991: Do we have any existing test tools for inspecting the `SuperEditor` scroll offset?

super_editor/lib/src/core/edit_context.dart
rutvik110: oh I see. I missed on the docs but will look into it. 

I'vent checked the behaviour yet for the case you mentioned. We want the same behaviour though in that case, right?

super_editor/lib/src/default_editor/document_hardware_keyboard/document_keyboard_actions.dart
rutvik110: Checked on apple notes and google docs, they do animate the scrolling in this case. That seems like the common behaviour.

Will update it to animate the scrolling instead of direct jump.

super_editor/lib/src/default_editor/super_editor.dart
rutvik110: I assume that's the intention here. For function keys it halts the execution to avoid adding `?` characters in the doc. 

Though the presence of this method here indicates a bug somewhere. I tracked it down to this three actions in the `defaultKeyboardActions`, 
```
```

In case where none of the actions above these three handles the key event, these three are executed and then pressing page-up/page-down/home/end or any function keys is resulting in first deleting any selection in the paragraph if present and then adding an '?' character in the document at that place. 

So, instead of dealing with the bug in in those three actions, atm we're simply halting the execution early to avoid that. I'll have a better look at this and see if we can resolve the issue `doNothingWhenFnKeyPressed` is trying to solve without adding it and by just updating these last last three actions to better handle that scenario.

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
rutvik110: Will look into this.

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
rutvik110: I'vent looked. Do you want me to add those if it's not present?

super_editor/lib/src/default_editor/super_editor.dart
matthew-carroll: I was focused more on the priority of the handler, rather than the existence of the handler. Flutter inserts weird characters for various keys. This is a known issue. But if you put your handler near the top of the priority queue then you're forcing future FN key handlers to go above that.

If the issue is that the character insertion handlers do bad things with FN, then why not place your blocking handler just above those, and leave as much priority available as possible for handlers that might want to respond to FN keys?

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
matthew-carroll: We'll want them eventually, but we probably don't need to use this PR to create them. The more important question is where the scrolling access should live. I haven't come up with a better answer than `SuperEditorContext`, yet, but I'm still thinking about it.

super_editor/lib/src/default_editor/super_editor.dart
rutvik110: yes, that seems the best way to deal with it rn.

rutvik110:
So in case of `SuperEditor` is within another `Scrollable`, generally all the scroll actions work as long as the `scrollController` passed to `SuperEditor` is attached to the `Scrollable` above it. Though the scroll actions apply on the parent `Scrollable` so the actions allows scrolling past the actual `SuperEditor` content. That's expected, right?

Also there's one another thing, in case of infinite `maxScrollExtent` which is unlikely in most cases, error's thrown if we try to scroll to the bottom of the document. In this case, do we want to handle this within our handler or leave it to the user to handle it.

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
```diff
@@ -1183,8 +1171,206 @@ void main() {
...
           await tester.pumpAndSettle();
 
-          // Ensure that the scroll position is set to the expected position
-          // after the key press action.
+          // Ensure we scrolled to the bottom of the viewport.
+          expect(scrollState.widget.controller!.offset, equals(scrollState.position.maxScrollExtent));
+        });
+      });
+      
+      group("page scrolling in presense of an external scrollable", () {
```
rutvik110, line 1179: Added tests to ensure page scrolling keys behave same in presence of external scrollable. Except the widget being pumped, all other testing steps are same in this group. Let me know if you want me to refactor this to have a common function that would perform the required actions across each tests. Something like `performTestActionsForHomeKey` to reduce this duplication.

rutvik110:
@matthew-carroll Made requested updates to docs and updated tests for the scenario when ancestor scrollable is present. 

Also, there's a new handler I added `doNothingWhenPageUpOrPageDownOrHomeOrEndKeyPressed` to tackle adding `?` character on pressing page scrolling keys in absence of `scrollController`. 

Also left a few comments to get your feedback on some things.

matthew-carroll:
> So in case of SuperEditor is within another Scrollable, generally all the scroll actions work as long as the scrollController passed to SuperEditor is attached to the Scrollable above it. Though the scroll actions apply on the parent Scrollable so the actions allows scrolling past the actual SuperEditor content. That's expected, right?

When there's an ancestor scrollable, `SuperEditor` won't be given the `ScrollController`. If you look inside the part of `SuperEditor` that decides whether to add a viewport or not, you'll see that it walks up the tree to find the `ScrollPosition` for an ancestor. So we need a solution that doesn't require a `ScrollController` to be given to `SuperEditor`.

> Also there's one another thing, in case of infinite maxScrollExtent which is unlikely in most cases, error's thrown if we try to scroll to the bottom of the document. In this case, do we want to handle/ignore this within our handler or leave it to the user to handle it.

`SuperEditor` documents can't be infinite. They aren't a `ListView`. They're a finite series of nodes.

rutvik110:
> `SuperEditor` documents can't be infinite. They aren't a `ListView`. They're a finite series of nodes.

I meant when the ancestor `Scrollable` has infinite extent. Though now that I think of this, this is an unlikely case so I = think we don't need to worry abt it. I've added a check for it anyway which ignores the `Home` key press in case the document is infinite.

rutvik110:
> When there's an ancestor scrollable, `SuperEditor` won't be given the `ScrollController`. If you look inside the part of `SuperEditor` that decides whether to add a viewport or not, you'll see that it walks up the tree to find the `ScrollPosition` for an ancestor. So we need a solution that doesn't require a `ScrollController` to be given to `SuperEditor`.

Yah, that should work better instead of relying on explicitly passing `ScrollController` in this case.

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -360,9 +360,12 @@ class SuperEditorState extends State<SuperEditor> {
   void didChangeDependencies() {
     super.didChangeDependencies();
 
-    _ancestorScrollController = _findAncestorScrollController(context);
-    _createEditContext();
-    _createLayoutPresenter();
+    final ancestorScrollController = _findAncestorScrollController(context);
```
rutvik110, line 363: @matthew-carroll Updated the SuperEditor to use the ancestor `scrollController` within editContext if one is found. 

With this, we shouldn't require to pass the `scrollController` explicitly to the `SuperEditor`. 

Also in case, both the ancestor and the the SuperEditor were passed the `scrollController`, the ancestor's `scrollController` takes precedence when using within editContext.

matthew-carroll:
We started a clean PR to get the latest history on `main`: https://github.com/superlistapp/super_editor/pull/1286

Closing this one.

PR #610, opened 2022-06-08
[IME] Do not clear document selection when document loses focus

### LINKED ISSUE #609: [IME] Document selection is cleared when document loses focus
Steps to reproduce:
* Run demo app
* Select a word => toolbar is shown
* Tap on the link-button

**Expected:** toolbar remains open an input field is shown to introduce the link URL
**Actual:** toolbar is closed

https://user-images.githubusercontent.com/31278849/172638158-d5b80ebd-cd14-40e8-9552-5a8eb7aa36e4.mov

miguelcmedeiros:
Before this PR, document selection would be cleared by the IME document interactor whenever the document would lose focus. 

This behavior differs from the physical-keyboard document interactor and as shown in #609, it breaks existing functionality.

By removing the clearance of the document selection it solves the issue, but I'm not aware of potential consequences. 

Fixes #609.

### DIFF
--- a/super_editor/lib/src/default_editor/document_input_ime.dart
+++ b/super_editor/lib/src/default_editor/document_input_ime.dart
@@ -204,8 +204,6 @@ class _DocumentImeInteractorState extends State<DocumentImeInteractor> implement
 
     editorImeLog.info('Detaching TextInputClient from TextInput.');
 
-    widget.editContext.composer.selection = null;
-
     _inputConnection!.close();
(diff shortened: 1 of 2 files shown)


matthew-carroll:
You said that the behavior is different between physical keyboard and IME. What's the observable difference as a user?

You mentioned that the IME behavior breaks some existing expectations. What are those expectations? Can you add tests for those expectations?

PR #616, opened 2022-06-14
Fix SuperDesktopTextField exception when rendering text containing only emojis (Resolves #459)

### LINKED ISSUE #459: SuperTextField - Exception when rendering BlinkingTextCaret on web
```
$ flutter --version
Flutter 2.10.3 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 7e9793dee1 (7 days ago) • 2022-03-02 11:23:12 -0600
Engine • revision bd539267b4
Tools • Dart 2.16.1 • DevTools 2.9.2
```

`main.dart`

```dart

void main() {
  final controller = AttributedTextEditingController(
    text: AttributedText(text: '🐢'),
  );
  runApp(
    MaterialApp(
      home: Center(
        child: SuperDesktopTextField(textController: controller),
      ),
    ),
  );
}
```

`pubspec.yaml`

```
name: flutter_sandbox
description: A new Flutter project.

publish_to: "none"

environment:
  sdk: ">=2.16.0 <3.0.0"

dependencies:
  flutter:
    sdk: flutter
  super_editor:
    git:
      url: https://github.com/superlistapp/super_editor.git
      path: super_editor
      ref: 42aadf8175971931215023b8717231fad1831ae1
```

# Steps to reproduce
* Run `flutter run -d chrome` (Flutter stable 2.10.3)
* Hit tab to focus the TextField (or click on it to manually focus)
* You will get an exception `Bad state: No element` when trying to render `BlinkingTextCaret`

@matthew-carroll we'd be willing to sponsor a fix or workaround for the error above.

The one below isn't a priority for us since it's a specific commit Jan 30th that we just got off of.

# Related issue on later version of Flutter on MacOS or Web
* Use flutter commit `29fd0469ed38e7f81a3e73bbdf06b89907bdba95`
* Run `flutter run -d macos` (or `flutter run -d chrome`)
* Hit tab to focus the TextField (or click on it to manually focus)
* You will the same exception `Bad state: No element` when trying to render `BlinkingTextCaret`

angelosilvestre:
Fix `SuperDesktopTextField` exception when rendering text containing only emojis. Resolves https://github.com/superlistapp/super_editor/issues/459

Trying to move the selection using the keyboard when `SuperDesktopTextField` has text containing only emojis causes the exception `Bad state: No element`. The issue occurs in Windows and Web

After some investigation this is what I've found:

https://github.com/superlistapp/super_editor/blob/ca24005782b21a8640f20a0b0ed3b2a5dd206f01/super_text_layout/lib/src/caret_layer.dart#L76-L81

When the text contains only emojis the line `widget.textLayout.getHeightForCaret(widget.position!)` returns null and `widget.textLayout.getLineHeightAtPosition(widget.position!)` is called , calling `getCharacterBox` on `RenderParagraphProseTextLayout`.

At `getCharacterBox` we walk forward and backward trying to find a character with a box, but we don't handle the case in which we don't find one:

https://github.com/superlistapp/super_editor/blob/ca24005782b21a8640f20a0b0ed3b2a5dd206f01/super_text_layout/lib/src/text_layout.dart#L312-L332

The `getCharacterBox` method is only called to get the line height in this case.

For some reason, if the text contains 3 or more emojis, then `getHeightForCaret` returns a non-null value and the error doesn't happen.

I modified `getCharacterBox` to behave as if the text is empty when it doesn't find any boxes, but I'm not sure if that's the correct solution for this.

### DIFF
--- a/super_editor/test/super_textfield/super_desktop_textfield_keyboard_test.dart
+++ b/super_editor/test/super_textfield/super_desktop_textfield_keyboard_test.dart
@@ -1785,6 +1785,34 @@ void main() {
         });
       });
     });
+  
+    testWidgets("can move the selection when text contains only emojis", (tester) async {      
+      final controller = AttributedTextEditingController(
+        text: AttributedText(text: '🐢'),
+      );
+
+      await tester.pumpWidget(
+        MaterialApp(
+          home: Scaffold(            
+            body: SuperDesktopTextField(
+              textController: controller,
+              textStyleBuilder: (_) => const TextStyle(fontSize: 16),
+            ),
+          ),
+        ),
+      );
+
+      // Focus the text field
+      await tester.tap(find.byType(SuperDesktopTextField));
+      await tester.pumpAndSettle();
+
+      // Press left arrow key to move the selection to the begining of the text
+      await tester.sendKeyEvent(LogicalKeyboardKey.arrowLeft);
+      await tester.pumpAndSettle();      
+
+      // As long as this test completes without an error, it should be the case
+      // that the selection is at the begining of the text
+    });
   });
 }
 
--- a/super_editor/test/super_textfield/super_desktop_textfield_keyboard_test.dart
+++ b/super_editor/test/super_textfield/super_desktop_textfield_keyboard_test.dart
@@ -208,7 +208,7 @@ void main() {
           // We give a tiny bit of wiggle room on the value because when this test
           // is run on Windows and Linux CI, there is some kind of precision error
           // that results in a tiny positive number instead of zero.
-          expect(textLayout.getCharacterBox(const TextPosition(offset: 16)).top, lessThan(0.1));
+          expect(textLayout.getCharacterBox(const TextPosition(offset: 16))?.top, lessThan(0.1));
 
           // On Linux CI, the "top" is a very tiny negative number, so we check for that value
           // instead of the check that we actually want to do.
@@ -218,7 +218,7 @@ void main() {
           //   Actual: <-7.152557373046875e-7>
           //    Which: is not a value greater than or equal to <0>
           // expect(textLayout.getCharacterBox(const TextPosition(offset: 16)).top, greaterThanOrEqualTo(0));
-          expect(textLayout.getCharacterBox(const TextPosition(offset: 16)).top, greaterThanOrEqualTo(-0.000001));
+          expect(textLayout.getCharacterBox(const TextPosition(offset: 16))?.top, greaterThanOrEqualTo(-0.000001));
         }, skip: true);
 
         testWidgetsOnDesktop('SHIFT + LEFT ARROW expands left by character', (tester) async {
@@ -323,7 +323,7 @@ void main() {
           // We should have gone from line 1 to line 2. Make double sure by
           // checking that the bounding box for the character that's now selected
           // does not sit at the top of the text box.
-          expect(SuperTextFieldInspector.findProseTextLayout().getCharacterBox(const TextPosition(offset: 18)).top,
+          expect(SuperTextFieldInspector.findProseTextLayout().getCharacterBox(const TextPosition(offset: 18))?.top,
               isNonZero);
         });
 
--- a/super_editor/test/super_textfield/super_desktop_textfield_keyboard_test.dart
+++ b/super_editor/test/super_textfield/super_desktop_textfield_keyboard_test.dart
@@ -1785,34 +1785,6 @@ void main() {
         });
       });
     });
-  
-    testWidgets("can move the selection when text contains only emojis", (tester) async {      
-      final controller = AttributedTextEditingController(
-        text: AttributedText(text: '🐢'),
-      );
-
-      await tester.pumpWidget(
-        MaterialApp(
-          home: Scaffold(            
-            body: SuperDesktopTextField(
-              textController: controller,
-              textStyleBuilder: (_) => const TextStyle(fontSize: 16),
-            ),
-          ),
-        ),
-      );
-
-      // Focus the text field
-      await tester.tap(find.byType(SuperDesktopTextField));
-      await tester.pumpAndSettle();
-
-      // Press left arrow key to move the selection to the begining of the text
-      await tester.sendKeyEvent(LogicalKeyboardKey.arrowLeft);
-      await tester.pumpAndSettle();      
-
-      // As long as this test completes without an error, it should be the case
-      // that the selection is at the begining of the text
-    });
   });
 }
 
--- a/super_editor/test/super_textfield/super_desktop_textfield_text_layout_test.dart
+++ b/super_editor/test/super_textfield/super_desktop_textfield_text_layout_test.dart
@@ -1,0 +1,460 @@
+import 'package:flutter/material.dart' hide SelectableText;
+import 'package:flutter_test/flutter_test.dart';
+import 'package:flutter_test_robots/flutter_test_robots.dart';
+import 'package:super_editor/super_editor.dart';
+
+import 'super_textfield_inspector.dart';
+import 'super_textfield_robot.dart';
+
+void main() {
+  group('SuperDesktopTextField', () {
+    group('containing only one emoji', (){
+      testWidgets("moves left with LEFT ARROW", (tester) async {           
+        await _pumpSuperTextFieldEmojiTest(tester, 
+          text: '🐢',
+        );
+
+        // Place caret after the emoji      
+        await tester.placeCaretInSuperTextField(2);   
+
+        // Ensure we are at the end of the text
+        expect(
+          SuperTextFieldInspector.findSelection(), 
+          const TextSelection.collapsed(offset: 2),
+        );   
+
+        // Press left arrow key to move the selection to the beginning of the text
+        await tester.pressLeftArrow();      
+
+        // Ensure caret is at the beginning of the text
+        expect(
+          SuperTextFieldInspector.findSelection(), 
+          const TextSelection.collapsed(offset: 0)
+        );         
+      });
+
+      testWidgets("expands selection with SHIFT + LEFT ARROW", (tester) async {           
+        await _pumpSuperTextFieldEmojiTest(tester, 
+          text: '🐢',
+        );
+
+        // Place caret after the emoji      
+        await tester.placeCaretInSuperTextField(2); 
+
+        // Ensure we are at the end of the text
+        expect(
+          SuperTextFieldInspector.findSelection(), 
+          const TextSelection.collapsed(offset: 2),
+        );      
+
+        // Press shift + left arrow key to expand the selection to the left
+        await tester.pressShiftLeftArrow();      
+
+        // Ensure that the emoji is selected
+        expect(
+          SuperTextFieldInspector.findSelection(), 
+          const TextSelection(
+            baseOffset: 2,
+            extentOffset: 0,
+          ),
+        );         
+      });
+    
+      testWidgets("moves right with RIGHT ARROW", (tester) async {      
+        await _pumpSuperTextFieldEmojiTest(tester, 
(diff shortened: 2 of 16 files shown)


super_text_layout/lib/src/text_layout.dart
```diff
@@ -329,6 +329,13 @@ class RenderParagraphProseTextLayout implements ProseTextLayout {
       ));
     }
 
+    // If we still don't have any boxes, it means that the text contains only
+    // emojis and we can't get a measurement from Flutter
+    if (boxes.isEmpty) {
+      final lineHeightEstimate = _renderParagraph.getFullHeightForCaret(const TextPosition(offset: 0)) ?? 0.0;
+      return TextBox.fromLTRBD(0, 0, 0, lineHeightEstimate, TextDirection.ltr);
```
matthew-carroll, line 336: I think this resolution is better placed wherever this method is called. By handling the situation where we can't find a box within this method, we prevent the client from possibly taking alternative actions, because the client can't tell the difference. I'm not sure about all the places that call this method, but I think a more appropriate change would be to make the return type nullable and then return `null` when no boxes are found.

super_text_layout/lib/src/text_layout.dart
```diff
@@ -329,6 +329,13 @@ class RenderParagraphProseTextLayout implements ProseTextLayout {
       ));
     }
 
+    // If we still don't have any boxes, it means that the text contains only
+    // emojis and we can't get a measurement from Flutter
+    if (boxes.isEmpty) {
+      final lineHeightEstimate = _renderParagraph.getFullHeightForCaret(const TextPosition(offset: 0)) ?? 0.0;
```
matthew-carroll, line 335: Assuming that this method is changed to return `null`, as per my other comment, then I think the more appropriate behavior for estimating the height is to return the font size multiplied by the line height. This is a decision would that would be made by the caller of this method. A similar approach is taken elsewhere in super_editor.

Be aware that `textStyle.fontSize` and `textStyle.height` might be null. If no font size is available, then you can go ahead and treat the height as `0`, because I can't think of a better answer in general. When `textStyle.height` is `null`, treat the height multiplier as `1.0`.

super_editor/test/super_textfield/super_desktop_textfield_keyboard_test.dart
```diff
@@ -3067,6 +3067,34 @@ void main() {
         });
       });
     });
+  
+    testWidgets("can move the selection when text contains only emojis", (tester) async {
```
matthew-carroll, line 3071: With regard to categorizing this test, the behavior may be triggered with a key press, but you're not really testing keyboard actions or shortcuts. I think that I would consider this test to be more of a text layout test. Can you place this in a new test file called `super_desktop_textfield_text_layout_test.dart`?

super_editor/test/super_textfield/super_desktop_textfield_keyboard_test.dart
```diff
@@ -3067,6 +3067,34 @@ void main() {
...
+            body: SuperDesktopTextField(
+              textController: controller,
+              textStyleBuilder: (_) => const TextStyle(fontSize: 16),
+            ),
+          ),
+        ),
+      );
+
+      // Focus the text field
+      await tester.tap(find.byType(SuperDesktopTextField));
```
matthew-carroll, line 3088: Can you use the `SuperTextFieldRobot` extensions here?

super_editor/test/super_textfield/super_desktop_textfield_keyboard_test.dart
```diff
@@ -3067,6 +3067,34 @@ void main() {
...
+      await tester.pumpAndSettle();
+
+      // Press left arrow key to move the selection to the begining of the text
+      await tester.sendKeyEvent(LogicalKeyboardKey.arrowLeft);
```
matthew-carroll, line 3092: Can you use `flutter_test_robots` here?

super_editor/test/super_textfield/super_desktop_textfield_keyboard_test.dart
```diff
@@ -3067,6 +3067,34 @@ void main() {
...
+      await tester.pumpAndSettle();      
+
+      // As long as this test completes without an error, it should be the case
```
matthew-carroll, line 3095: What if the caret doesn't move? Don't we want to make sure that we can move the caret upstream to downstream, and downstream to upstream, when we only have an emoji? Also, how about SHIFT + LEFT/RIGHT ARROWs? And how about double tapping on the emoji, which should select it. You may also want to test double tapping on two back-to-back emojis, which should select both of them.

matthew-carroll (commented):
Thanks for the thorough write-up on the problem. This is a good start. I think we want to adjust the direction of the implementation and the tests a bit.

super_editor/test/super_textfield/super_desktop_textfield_text_layout_test.dart
```diff
@@ -0,0 +1,472 @@
+import 'package:flutter/material.dart' hide SelectableText;
+import 'package:flutter_test/flutter_test.dart';
+import 'package:flutter_test_robots/flutter_test_robots.dart';
+import 'package:super_editor/super_editor.dart';
+
+import 'super_textfield_inspector.dart';
+import 'super_textfield_robot.dart';
+
+void main() {
+  group('SuperDesktopTextField', () {
```
matthew-carroll, line 10: Looking at this again, let's make this test suite generic to `SuperTextField`. Let's rename the file to `super_textfield_text_layout_test.dart` and let's make the top group "SuperTextField with keyboard"

super_editor/test/super_textfield/super_desktop_textfield_text_layout_test.dart
```diff
@@ -0,0 +1,472 @@
...
+    group('containing only one emoji', (){
+      testWidgets("moves left with LEFT ARROW", (tester) async {
```
matthew-carroll, line 12: Let's run these tests on all platforms, including mobile. If we don't have a test method for all platforms yet, can you add one that works like the desktop version, but add Android and iOS?

If Android and iOS don't work because the inspector or the robot don't work with mobile yet, then ignore this and update the tests to `testWidgetsOnDesktop()`.

Also, now that we're framing these tests as layout tests, we don't need to mention the shortcuts in the test name, because that's not really what we care about. Instead this test would be "moves caret upstream around the emoji". The other names should be updated in a similar fashion.

super_editor/test/super_textfield/super_desktop_textfield_text_layout_test.dart
```diff
@@ -0,0 +1,472 @@
...
+
+      testWidgets("expands selection with SHIFT + LEFT ARROW", (tester) async {           
+        await _pumpSuperTextFieldEmojiTest(tester, 
+          text: '🐢',
+        );
+
+        // Place caret at the beginning of the text
+        await tester.placeCaretInSuperTextField(0);
+        // Move caret to the right   
+        await tester.pressRightArrow();
```
matthew-carroll, line 46: If all you want to test is selecting upstream, then you can place the caret on the right side of the emoji and avoid the right arrow.

If you're using the right arrow because there's something wrong with the robot, please add a TODO here explaining that, so we know we can get rid of the right arrow later.

super_editor/test/super_textfield/super_desktop_textfield_text_layout_test.dart
```diff
@@ -0,0 +1,472 @@
...
+        expect(
+          SuperTextFieldInspector.findSelection(), 
+          const TextSelection(
+            baseOffset: 0,
+            extentOffset: 2,
+          ),
+        );         
+      });
+
+      testWidgets("selects the content double tap", (tester) async {
```
matthew-carroll, line 118: Let's avoid ambiguous terms like "content" when we have a better option. In this case: "selects the emoji on double tap".

PS - the tapping tests might belong in a different file, but we can leave that for another time.

super_editor/test/super_textfield/super_textfield_robot.dart
```diff
@@ -39,6 +40,47 @@ extension SuperTextFieldRobot on WidgetTester {
     throw Exception("Couldn't find a SuperTextField with the given Finder: $fieldFinder");
   }
 
+  /// Double taps to select a character at the given [offset].
```
matthew-carroll, line 43: This method doesn't really try "to select a character". It "Double taps in a [SuperTextField] at the given [offset]". We don't want to make assumptions about what happens.

super_editor/test/super_textfield/super_textfield_robot.dart
```diff
@@ -39,6 +40,47 @@ extension SuperTextFieldRobot on WidgetTester {
...
+  ///
+  /// {@template supertextfield_finder}
```
matthew-carroll, line 45: Please see other uses of templates. What you want to do here is reference the existing template, not define a new one.

super_editor/test/super_textfield/super_textfield_robot.dart
```diff
@@ -39,6 +40,47 @@ extension SuperTextFieldRobot on WidgetTester {
...
+  /// By default, this method expects a single [SuperTextField] in the widget tree and
+  /// finds it `byType`. To specify one [SuperTextField] among many, pass a [superTextFieldFinder].
+  /// {@endtemplate}
+  Future<void> doubleTapAtSuperTextField(int offset,
+    [Finder? superTextFieldFinder, TextAffinity affinity = TextAffinity.downstream]) async {
+    final fieldFinder = _findInnerPlatformTextField(superTextFieldFinder ?? find.byType(SuperTextField));
```
matthew-carroll, line 51: Add a TODO here to de-duplicate this behavior with `placeCaretInSuperTextField`.

super_text_layout/lib/src/text_layout.dart
```diff
@@ -39,8 +49,12 @@ abstract class TextLayout {
   /// Returns a [List] of [TextBox]es that contain the given [selection].
   List<TextBox> getBoxesForSelection(TextSelection selection);
 
-  /// Returns a bounding [TextBox] for the character at the given [position].
-  TextBox getCharacterBox(TextPosition position);
+  /// Returns a bounding [TextBox] for the character at the given [position] or null
```
matthew-carroll, line 52: Please notate "null" as a code snippet: \`null\`

super_text_layout/lib/src/text_layout.dart
```diff
@@ -329,6 +348,12 @@ class RenderParagraphProseTextLayout implements ProseTextLayout {
       ));
     }
 
+    // If we still don't have any boxes, it means that the text contains only
```
matthew-carroll, line 351: Couldn't it also mean that there isn't any text? And how about an inline widget? Might that cause a null TextBox, too?

Generally, it's best to avoid saying what something means unless you're very confident that it's true. Otherwise, this situation will come up in the future, a developer will believe this comment, and then go down the wrong route because the cause was something else.

I think for this method you can just `return boxes.firstOrNull;`

super_editor/test/super_textfield/super_textfield_text_layout_test.dart
```diff
@@ -0,0 +1,461 @@
...
+
+void main() {
+  group('SuperTextField with keyboard', () {
+    group('containing only one emoji', (){
+      testWidgetsOnAllPlatforms("moves caret upstream around the emoji", (tester) async {           
+        await _pumpSuperTextFieldEmojiTest(tester, 
+          text: '🐢',
+        );
+
+        // TODO: placing caret on the right side of the emoji at the end of the text isn't working correctly
```
matthew-carroll, line 18: Can you include a reference to an associated issue number so that we know when we can update the test implementation?

matthew-carroll (commented):
One last tiny comment

PR #617, opened 2022-06-14
Fix BlinkController not notifying its listeners when it stops blinking (Resolves #523)

### LINKED ISSUE #523: Potential bug in the `CaretBlinkController` implementation
I am reviewing the `super_editor` repo to get familiar with it, and I have found a couple of potential issues with the `CaretBlinkController` in the following places:

- https://github.com/superlistapp/super_editor/blob/main/super_text/lib/src/caret.dart#L299-L302
  Where a `notifyListeners` call is missing after setting `_isVisible` to `true`.

  ```dart
  void stopBlinking() {
    _isVisible = true; // If we're not blinking then we need to be visible
    _stopTimer();
  }
  ```

- https://github.com/superlistapp/super_editor/blob/main/super_text/lib/src/caret.dart#L311-L326
  Where a `notifyListeners` call is also missing after setting `_isVisible` to `true`.

  ```dart
  /// Clients should call this method whenever the caret moves within
  /// the content so that the caret is made fully opaque and the blink
  /// timer restarts.
  void onCaretMoved() {
    // Immediately make the caret visible whenever the position
    // changes, e.g., when the user adds/removes a character.
    _isVisible = true;

    _timer?.cancel();

    if (!_isBlinkingEnabled) {
      return;
    }

    _startTimer();
  }
  ```

angelosilvestre:
`BlinkController` was not calling `notifyListeners` when `stopBlinking` was called. Resolves https://github.com/superlistapp/super_editor/issues/523

As `_CaretPainter` uses its blinkController as a repaint `Listenable`, it's necessary to call `notifyListeners` to cause `_CaretPainter` to repaint.

I added tests to ensure both `stopBlinking` and `startBlinking` cause the listeners to be notified

### DIFF
--- a/super_text_layout/test/blink_controller_test.dart
+++ b/super_text_layout/test/blink_controller_test.dart
@@ -1,0 +1,73 @@
+import 'package:flutter/src/scheduler/ticker.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:super_text_layout/src/infrastructure/blink_controller.dart';
+
+void main() {
+  group("BlinkController", () {
+    testWidgets("should notify its listeners when it stops blinking", (tester) async {
+      bool hasNotifiedItsListener = false;
+
+      final blinkController = BlinkController(tickerProvider: tester);
+      blinkController.addListener(() { 
+        hasNotifiedItsListener = true;
+      });      
+      
+      blinkController.stopBlinking();
+      
+      // Ensure that the callback was called
+      expect(hasNotifiedItsListener, true);
+    });
+
+    testWidgets("should notify its listeners when it starts blinking", (tester) async {
+      // duration to switch between visible and invisible
+      const flashPeriod = Duration(milliseconds: 500);
+
+      bool hasNotifiedItsListeners = false;      
+      
+      // Create a fake ticker provider to be able to manually trigger a tick
+      final tickerProvider = _TestSingleTickerProvider();          
+      final blinkController = BlinkController(
+        tickerProvider: tickerProvider, 
+        flashPeriod: flashPeriod
+      );
+      blinkController.addListener(() { 
+        hasNotifiedItsListeners = true;
+      });      
+      
+      blinkController.startBlinking();
+      // Trigger a tick with an ellapsed time greater than the flashPeriod,
+      // so the controller should change its visible state and notify its listeners
+      tickerProvider.tick(flashPeriod + const Duration(milliseconds: 1));  
+    
+      // Ensure that the callback was called
+      expect(hasNotifiedItsListeners, true);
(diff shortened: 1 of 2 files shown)


super_text_layout/test/blink_controller_test.dart
```diff
@@ -0,0 +1,73 @@
+import 'package:flutter/src/scheduler/ticker.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:super_text_layout/src/infrastructure/blink_controller.dart';
+
+void main() {
+  group("BlinkController", () {
+    testWidgets("should notify its listeners when it stops blinking", (tester) async {
```
matthew-carroll, line 7: A minor nitpick about wording: when possible, prefer statements about what something does, rather than what it should do -> `notifies listeners when it stops blinking`.

super_text_layout/test/blink_controller_test.dart
```diff
@@ -0,0 +1,73 @@
...
+    testWidgets("should notify its listeners when it starts blinking", (tester) async {
+      // duration to switch between visible and invisible
+      const flashPeriod = Duration(milliseconds: 500);
+
+      bool hasNotifiedItsListeners = false;      
+      
+      // Create a fake ticker provider to be able to manually trigger a tick
+      final tickerProvider = _TestSingleTickerProvider();          
+      final blinkController = BlinkController(
+        tickerProvider: tickerProvider,
```
matthew-carroll, line 30: Can you pass the `WidgetTester` as the `TickerProvider` and then use `pump(Duration(...))`?

super_text_layout/test/blink_controller_test.dart
```diff
@@ -0,0 +1,73 @@
...
+        hasNotifiedItsListener = true;
+      });      
+      
+      blinkController.stopBlinking();
+      
+      // Ensure that the callback was called
+      expect(hasNotifiedItsListener, true);
+    });
+
+    testWidgets("should notify its listeners when it starts blinking", (tester) async {
```
matthew-carroll, line 21: I think this test is really verifying that `notifies listeners every time it blinks`, which I think is different than `notifies listeners when it starts blinking`. Both of those deserve tests, so this test shouldn't be removed, but I think testing the `starts blinking` case requires first stopping the controller, then starting it again. I think we want/expect the controller to immediately notify listeners when it starts again, even before the first blink.

matthew-carroll (commented):
Thanks for the explanation of the problem in the PR writeup. I have a few comments for the tests, and I think we may need to add one more notification to the implementation.

matthew-carroll (approved):
LGTM, thanks.

PR #618, opened 2022-06-14
Fix incorrect Windows CTRL + Arrow Keys behavior (Resolves #275)

### LINKED ISSUE #275: Incorrect Windows CTRL + Arrow Keys behavior
On web on Windows, the <kbd>CTRL</kbd> + arrow key behavior is incorrect.

This is what tapping the left arrow key + control **should** look like:

image

And this is what it looks like in SuperEditor:

image

The behavior that SuperEditor executes on CTRL + left arrow key should actually happen for <kbd>Pos1</kbd> or <kbd>Home</kbd> + left arrow key instead.

---

* CTRL + arrow key should select the next / previous word.
* Pos1 + arrow key should select the end / beginning of the line.

angelosilvestre:
Fix incorrect Windows CTRL + Arrow Keys behavior. Resolves https://github.com/superlistapp/super_editor/issues/275

We are using `alt` to skip words on both Windows and Mac. I've updated the implementation to use `ctrl` on Windows.

I've also implemented the following shortcuts on Windows:

- `home`: move to the beginning of the line
- `end`: move to the end of the line


super_editor/lib/src/default_editor/document_keyboard_actions.dart
```diff
@@ -325,6 +329,44 @@ ExecutionInstruction moveToLineStartOrEndWithCtrlAOrE({
...
 }
 
+ExecutionInstruction moveToLineStartWithHome({
+  required EditContext editContext,
+  required RawKeyEvent keyEvent,
+}) {
+  if (defaultTargetPlatform != TargetPlatform.windows) {
+    return ExecutionInstruction.continueExecution;
+  }  
+  bool didMove = false;
```
matthew-carroll, line 339: From a readability perspective, it's better to place this immediately above the following if-statement, rather than immediately after the previous if-statement. The position of the line makes this line look like it's related to the target platform check, rather than the movement behavior.

super_editor/lib/src/default_editor/document_keyboard_actions.dart
```diff
@@ -266,8 +266,12 @@ ExecutionInstruction moveUpDownLeftAndRightWithArrowKeys({
 
   bool didMove = false;
   if (keyEvent.logicalKey == LogicalKeyboardKey.arrowLeft || keyEvent.logicalKey == LogicalKeyboardKey.arrowRight) {
-    MovementModifier? movementModifier;
-    if (keyEvent.isPrimaryShortcutKeyPressed) {
+    MovementModifier? movementModifier;    
+    if (defaultTargetPlatform == TargetPlatform.windows) {
```
matthew-carroll, line 270: What happens on Linux? Shouldn't this Windows behavior also apply to Linux?

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
```diff
@@ -106,6 +107,98 @@ void main() {
 
           Platform.setTestInstance(null);
         });
+        
+        testWidgets('beginning of line with HOME on Windows', (tester) async {
```
matthew-carroll, line 111: Can you add these same keyboard handlers to SuperTextField, along with SuperTextField tests that work like the other tests that I committed? Those tests will also show the issue that Linux is doing the wrong thing.

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
```diff
@@ -106,6 +107,98 @@ void main() {
...
+                nodeId: "1",
+                nodePosition: TextNodePosition(offset: 0),
+              ),
+            ),
+          );
+
+          debugDefaultTargetPlatformOverride = null;
+        });
+        
+        testWidgets('end of line with END on Windows', (tester) async {
```
matthew-carroll, line 134: I think that each of these `testWidgets()` calls can be replaced with `testWidgetsOnWindows()` or `testWidgetsOnWindowsAndLinux()`. Then you can get rid of the `debugDefaultTargetPlatformOverride` lines.

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
```diff
@@ -106,6 +107,98 @@ void main() {
...
+          // Start the user's selection somewhere after the beginning of the first
+          // line in the first node.
+          final editContext = await _pumpCaretMovementTestSetup(tester, textOffsetInFirstNode: 8);
+
+          await tester.pressHome();
+
+          // Ensure that the caret moved to the beginning of the line.
+          expect(
+            editContext.composer.selection,
+            const DocumentSelection.collapsed(
```
matthew-carroll, line 123: Can you use the `SuperEditorInspector` to obtain the selection in these tests rather than use the returned `editContext`?

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
```diff
@@ -106,6 +107,98 @@ void main() {
 
           Platform.setTestInstance(null);
         });
+
```
matthew-carroll, line 110: When moving to line boundaries, there are a specific selection of test conditions that we should use to avoid accidental mistakes. Notice that both of your tests would continue to pass even if the HOME and END behavior moved to the very beginning or the very end of text, regardless of line.

To more effectively test these behaviors, we should test 4 conditions:
 - pressing HOME when the caret sits on a line after the first line, in an auto-wrapping paragraph
 - pressing HOME when the caret sits on a line after the first line, in a paragraph with explicit newlines
 - pressing END when the caret sits on a line before the last line, in an auto-wrapping paragraph
 - pressing END when the caret sits on a line before the last line, in a paragraph with explicit newlines

matthew-carroll (commented):
Thanks for the PR. Here are a few comments.

super_editor/lib/src/default_editor/document_keyboard_actions.dart
angelosilvestre: After a little research it seems that linux uses the same shortcuts windows uses. I will update to also work on linux

super_editor/lib/src/default_editor/document_keyboard_actions.dart
```diff
@@ -266,8 +266,12 @@ ExecutionInstruction moveUpDownLeftAndRightWithArrowKeys({
...
   bool didMove = false;
   if (keyEvent.logicalKey == LogicalKeyboardKey.arrowLeft || keyEvent.logicalKey == LogicalKeyboardKey.arrowRight) {
-    MovementModifier? movementModifier;
-    if (keyEvent.isPrimaryShortcutKeyPressed) {
+    MovementModifier? movementModifier;    
+    if (defaultTargetPlatform == TargetPlatform.windows || defaultTargetPlatform == TargetPlatform.linux) {
+      if (keyEvent.isControlPressed) {
+        movementModifier = MovementModifier.word;
+      }
+    } else if (keyEvent.isPrimaryShortcutKeyPressed) {
```
matthew-carroll, line 274: I think we were using `isPrimaryShortcutKeyPressed` to handle `CMD` on Mac and `CTRL` on Windows. Now that you're explicitly handling Windows, and given that `CTRL` **shouldn't** move by line, I think we should change this if-statement:

`if (defaultTargetPlatform == TargetPlatform.mac && keyEvent.isMetaPressed)`

Also, for the Windows/Linux if-statement above, is there a situation that requires us to have one if-statement inside another if-statement? Can we combine them?

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
```diff
@@ -15,18 +21,18 @@ void main() {
     'Document keyboard actions',
     () {
       group('jumps to', () {
-        testWidgets('beginning of line with CMD + LEFT ARROW', (tester) async {
+        testWidgets('beginning of line with CMD + LEFT ARROW on Mac', (tester) async {
```
matthew-carroll, line 24: Can these `testWidgets()` calls become `testWidgetsOnMac()`?

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
```diff
@@ -85,17 +91,17 @@ void main() {
           Platform.setTestInstance(null);
         });
 
-        testWidgets('end of word with ALT + RIGHT ARROW', (tester) async {
+        testWidgets('end of word with ALT + RIGHT ARROW on Mac', (tester) async {
```
matthew-carroll, line 94: Can you add tests that verify that Windows shortcuts don't run on Mac?

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
```diff
@@ -106,6 +112,197 @@ void main() {
 
           Platform.setTestInstance(null);
         });
+
+        testWidgetsOnWindowsAndLinux('beginning of line with HOME in an auto-wrapping paragraph', (tester) async {
+          // Configure the screen to a size we know will cause the paragraph to auto-wrap its lines
```
matthew-carroll, line 117: Can these Windows and Linux test use the same document and binding configuration? If so, you can share the setup in a singe call, similar to `_pumpCaretMovementTestSetup` in other tests.

Also, if the physical size is being used to limit the document size, I'd rather that we surround the `SuperEditor` with a `ConstrainedBox`. I'm guessing the reason you did it this way was because the document configuration APIs didn't allow you to constrain it? If that's the case, let's add a nullable property to the appropriate configuration object for `editorSize`. By default we won't use any size, but if a size is provided, then we should enforce that when we pump the widget tree.

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
```diff
@@ -106,6 +112,197 @@ void main() {
...
+            ..platformDispatcher.textScaleFactorTestValue = 1.0
+            ..physicalSizeTestValue = const Size(400, 400);
+          
+          await tester 
+            .createDocument()
+            .withSingleParagraph()
+            .forDesktop()
+            .pump();
+
+          // Place caret at the second line at "adipiscing |elit"
```
matthew-carroll, line 129: It's worth pointing out that we avoid placing the caret in the first line to make sure that HOME doesn't move the caret all the way to beginning of the text. Similar for END.

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
```diff
@@ -106,6 +112,197 @@ void main() {
...
+            SuperEditorInspector.findDocumentSelection(),
+            const DocumentSelection.collapsed(
+              position: DocumentPosition(
+                nodeId: "1",
+                nodePosition: TextNodePosition(offset: 40),
+              ),
+            ),
+          );
+
+          tester.binding.window.clearDevicePixelRatioTestValue();
```
matthew-carroll, line 145: I think there's a method to clear all test values. Consider trying that to save a couple lines.

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
```diff
@@ -667,3 +864,14 @@ Future<EditContext> _pumpCaretMovementTestSetup(
...
+
+final _styleSheet = defaultStylesheet.copyWith(
+  addRulesAfter: [
+    StyleRule(BlockSelector.all, (doc, docNode) => {'textStyle': _textStyle}),
+  ],
+);
+
+const _textStyle = TextStyle(
+  color: Colors.black,
+  fontFamily: 'Roboto',
```
matthew-carroll, line 876: Can you figure out a text layout with Ahem and avoid the text style and stylesheet? I don't know that we'll always have Roboto available in CI or other test environments.

super_editor/test/super_textfield/super_desktop_textfield_keyboard_test.dart
```diff
@@ -1486,7 +1500,21 @@ void main() {
 
           await tester.pressShiftCtlRightArrow();
 
-          expect(SuperTextFieldInspector.findSelection(), const TextSelection(baseOffset: 6, extentOffset: 16));
+          expect(SuperTextFieldInspector.findSelection(), const TextSelection(baseOffset: 6, extentOffset: 10));
+        });        
+
+        testWidgetsOnWindowsAndLinux('END moves right to end of line', (tester) async {
```
matthew-carroll, line 1506: Let's place the HOME and END tests down here, one after the other. I think readers are more likely to expect LEFT followed by RIGHT, rather than LEFT followed by HOME followed by RIGHT

angelosilvestre: I've placed HOME after LEFT because the group description is "move caret upstream". If I move HOME together with END it will be in the group "move caret downstream", which isn't the case for HOME. Should I create a separate group?

matthew-carroll: That's a good point. You can leave them as-is.

super_editor/test/super_textfield/super_desktop_textfield_keyboard_test.dart
```diff
@@ -1198,6 +1198,136 @@ void main() {
...
+          await tester.pressCtlRightArrow();
+
+          expect(SuperTextFieldInspector.findSelection(), const TextSelection.collapsed(offset: 6));
+        });
+      });
+    });
+
+    group('on Windows', () {
+      group('move caret upstream', () {
+        testWidgetsOnWindows('LEFT ARROW does nothing when ALT is pressed', (tester) async {
```
matthew-carroll, line 1263: Are Windows and Linux supposed to do the same things with regard to these key combos? If so, can we define these tests in the same place using the `testWidgetsOnWindowsAndLinux()` method?

super_editor/test/super_textfield/super_desktop_textfield_keyboard_test.dart
```diff
@@ -1456,12 +1586,26 @@ void main() {
...
+            tester,
+            AttributedTextEditingController(
+              text: AttributedText(text: "super text field"),
+            ),
+          );
+          await tester.placeCaretInSuperTextField(10);
+
+          await tester.pressHome();
+
+          expect(SuperTextFieldInspector.findSelection(), const TextSelection.collapsed(offset: 0));
```
matthew-carroll, line 1603: Similar to other discussions, I think we want to test this behavior on a line other than the first line, so that we're sure that `HOME` doesn't move the caret to the beginning of all text.

Same issue for `END`.

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
```diff
@@ -15,96 +19,439 @@ void main() {
...
+          );
+        });
+
+        testWidgetsOnWindowsAndLinux('with ALT + DOWN ARROW', (tester) async {                              
+          await _pumpExplicitLineBreakTestSetup(tester);
+
+          // Place caret at the first line at "Lorem |ipsum"          
+          await tester.placeCaretInParagraph('1', 6);
+
+          await _pressAltDownArrow(tester);
```
matthew-carroll, line 327: I know these methods weren't available when you wrote this, but a reminder to replace with the new flutter_test_robots methods

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
```diff
@@ -667,3 +1014,56 @@ Future<EditContext> _pumpCaretMovementTestSetup(
...
+    .forDesktop()
+    .withEditorSize(const Size(400, 400))
+    .pump();
+}
+
+Future<TestDocumentContext> _pumpExplicitLineBreakTestSetup(
+  WidgetTester tester, {
+  Size? size,
+}) async {          
+  final document = MutableDocument(
```
matthew-carroll, line 1031: Can you define this inline within the widget tree? It doesn't look like the `document` is used anywhere else.

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
```diff
@@ -667,3 +1014,56 @@ Future<EditContext> _pumpCaretMovementTestSetup(
...
+    ],
+  );  
+
+  final configurator = tester 
+    .createDocument()
+    .withCustomContent(document)
+    .forDesktop();
+
+  if (size != null) {
+    configurator.withEditorSize(size);
```
matthew-carroll, line 1049: If you allow a null `size` in `withEditorSize()` then you can move this call up to the builder without an if-statement and avoid breaking up the configuration.

matthew-carroll (commented):
A few more comments. We're getting close.

super_editor/test/super_textfield/super_desktop_textfield_keyboard_test.dart
angelosilvestre: On Linux ALT does not prevent horizontal movement of the caret

matthew-carroll: Does ALT "prevent" movement on Windows? Or does ALT just not move by word on Windows?

angelosilvestre: When ALT is pressed caret doesn't move at all using the arrow keys

matthew-carroll:
@angelosilvestre can you resolve the conflict with the pubspec so that the tests can run?

matthew-carroll (approved):
LGTM, thanks.

PR #623, opened 2022-06-16
Convert SuperTextField shortcut tests to user interaction tests

matthew-carroll:
Convert `SuperTextField` shortcut tests to user interaction tests.

All of the `SuperTextField` shortcut tests were previously written as unit tests against individual handlers. There's nothing wrong with that, but it greatly limits what the tests verify. For example, I discovered a couple handlers that had implementations, but they weren't included in `SuperTextField`. We missed that because we didn't write tests that actually run shortcuts through `SuperTextField`. Now, all such tests are user interaction tests, instead of unit tests.

The Flutter Bounty Hunters have a new package called `flutter_test_robots`, which will accumulate simulations of user behavior over time. This PR uses `flutter_test_robots` to easily simulate common keyboard shortcuts, e.g., `tester.pressShiftCtlLeftArrow()`.

This PR also creates a `SuperTextFieldInspector`, which answers questions about a given `SuperTextField` during a test, such as what text is currently in the field, or what selection is currently present. Similarly, extensions were added to `WidgetTester` by way of a `SuperTextFieldRobot` extension. These extensions make it easy to interact with a `SuperTextField`. For example: `tester.placeCaretInSuperTextField(5)`.

This PR removes the `SuperTextField` test tools from the public package API. I'm guessing that no one else was using those APIs. I think we need to re-evaluate how we provide inspection APIs in the public package API.


matthew-carroll:
CC @angelosilvestre @ThangVuNguyenViet - I recommend taking a look at some of the new testing tools that are introduced in this PR. Also, take a look at the approach to testing. Let me know if you have any questions about what I chose to test, or how I chose to test those things.

@angelosilvestre - You'll probably need to update some of your PRs when this merges.

matthew-carroll:
CC @venkatd for visibility into `SuperTextField` improvements.

matthew-carroll:
@miguelcmedeiros I'm going to merge this PR to unblock @angelosilvestre but I still encourage you to take a look and let me know if anything looks wrong to you.

PR #626, opened 2022-06-18
Moved clipboard simulation to the flutter_test_robots package

matthew-carroll:
Moved clipboard simulation to the flutter_test_robots package.

I asked a member of the Flutter team if they had any plans to create a test representation of clipboard behavior and he said probably not. So I moved the temporary simulated clipboard behavior into `flutter_test_robots` to make it permanent, and publicly available.


matthew-carroll:
CC - @angelosilvestre @ThangVuNguyenViet - you may end up using this API in tests at some point, if you deal with copy/paste behavior.

PR #628, opened 2022-06-18
Auto link URLs when pasting (Resolves #406)

### LINKED ISSUE #406: Auto link URLs in Sample Editor

ThangVuNguyenViet:
Auto link URLs (Resolves https://github.com/superlistapp/super_editor/issues/406 )

Recognizing URLs in the document and converting them into links when pasting

To verify a Uri parsed from a text, that Uri needs to hasScheme and hasAuthority

Additionally, this PR also implemented link splitting, a.k.a split the link into 2 parts when pasting content in the middle of an existing link, and link expanding prevention when typing.

Note: this PR partly resolves #406  because it does not contain link encoding when typing, which will be in another separated PR

MacOS:

https://user-images.githubusercontent.com/39979207/174451934-885a24aa-33e5-4fd9-a216-4239d62ba466.mov


ThangVuNguyenViet:
@matthew-carroll the CI runs failed because I added and used a new function called `getAttributionSpansAt`, which the cache doesn't have. I ran into this on my local too and resolved it by making `super_text_layout` in the .yaml file points to a path, not a version. I think this is a dependency issue

matthew-carroll:
@ThangVuNguyenViet can you check the version of `attributed_text` that's listed in `super_text_layout`'s pubspec? If the `attributed_text` version is less than `0.1.3`, can you bump that version up and try again?

Also, it looks like you have some conflicts in this PR.

ThangVuNguyenViet:
@matthew-carroll I fixed the issue by not using the newly added method and using the old one instead. Bumping up the version doesn't resolve it. It seems like the dependency doesn't pick what's in the `dependency_override`

super_editor/test/src/default_editor/common_editor_operations_test.dart
```diff
@@ -158,5 +163,313 @@ void main() {
...
+        // Tap to place the caret in the first paragraph.
+        await tester.placeCaretInParagraph("1", 0);
+
+        // Type some text by simulating hardware keyboard key presses.
+        await tester.pressCmdV();
+
+        // Ensure that the clipboard text was pasted into the SuperTextField
+        expect(
+          SuperEditorInspector.findTextInParagraph("1").text,
+          'Link: https://flutter.dev https://google.com',
```
matthew-carroll, line 383: Should there be a space between those two URLs? If you do this same operation in Docs or Word, does it behave this way?

Edit: I see that you added a space at the end of the copied text. What happens if there isn't a space at the end of the copied text? Can you change this test to remove that space? I think if we make things work without a trailing space, then we know things will work with a trailing space, too.

super_editor/test/src/default_editor/document_input_ime_test.dart
```diff
@@ -268,6 +271,128 @@ void main() {
...
+            .withCustomContent(document)
+            .forDesktop()
+            .autoFocus(true)
+            .pump();
+
+        // Place the caret in the first paragraph at the start of the link.
+        await tester.placeCaretInParagraph('1', 0);
+
+        // Type some text by simulating hardware keyboard key presses.
+        await tester.typeKeyboardText('Go to ');
```
matthew-carroll, line 311: This test is in an IME test file, but you're typing on a hardware keyboard. I think this is in the wrong place.

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
```diff
@@ -919,6 +918,129 @@ void main() {
           Platform.setTestInstance(null);
         });
       });
+
+      group('inserting', () {
```
matthew-carroll, line 922: Let's call this "inserting near links"

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
```diff
@@ -1016,33 +1138,26 @@ Future<EditContext> _pumpCaretMovementTestSetup(
 }
 
 Future<TestDocumentContext> _pumpAutoWrappingTestSetup(WidgetTester tester) async {
-  return await tester 
-    .createDocument()
-    .withSingleParagraph()
-    .forDesktop()
-    .withEditorSize(const Size(400, 400))
-    .pump();
+  return await tester.createDocument().withSingleParagraph().forDesktop().withEditorSize(const Size(400, 400)).pump();
```
matthew-carroll, line 1141: Please return this line to the multiline version from before.

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
```diff
@@ -1016,33 +1138,26 @@ Future<EditContext> _pumpCaretMovementTestSetup(
...
 
 Future<TestDocumentContext> _pumpExplicitLineBreakTestSetup(
   WidgetTester tester, {
   Size? size,
-}) async {          
-  return await tester 
-    .createDocument()
-    .withCustomContent(
-      MutableDocument(
+}) async {
```
matthew-carroll, line 1147: Please return the formatting to what it was before

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
```diff
@@ -94,7 +94,7 @@ void main() {
             ),
           );
         });
-        
+
```
matthew-carroll, line 97: There are a bunch of formatting changes in this file. These changes will probably conflict with other autoformatters. Can you figure out how to return the formatting to the original? I'm not sure why there was whitespace before, and there isn't whitespace now, but this difference will probably keep happening if we don't resolve it.

matthew-carroll (commented):
I left a number of comments. I'm going to create a couple tools to reduce the verbosity of the document tests.

I'd like to challenge you to simplify your implementation of the link behavior. I can't point to any specific area in the link code, but overall it seems very long and very complicated for what it's trying to achieve. Now that you have tests that cover this behavior, try to make the implementation as simple, obvious, and well documented as possible.

attributed_text/lib/src/attributed_spans.dart
```diff
@@ -215,6 +215,19 @@ class AttributedSpans {
     return matchingAttributionSpans;
   }
 
+  /// Returns spans for each attribution that (at least partially) appear at
+  /// [offset], as selected by [attributionFilter].
+  Set<AttributionSpan> getAttributionSpansAt({
```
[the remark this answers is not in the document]
ThangVuNguyenViet: Okay, I'll remove the method and things related to it then

super_editor/lib/src/default_editor/common_editor_operations.dart
```diff
@@ -1377,6 +1383,36 @@ class CommonEditorOperations {
     return true;
   }
 
+  Set<Attribution> _filterAttributionsWhenInsertText({
+    required TextNode textNode,
+    required int textOffset,
+    required Set<Attribution> currentAttribution,
+  }) {
+    final linkAttributionSpan = textNode.text.spans
+        // Note: We can't use the method [AttributedSpans.getAttributionSpansAt] because it seems
```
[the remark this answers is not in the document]
ThangVuNguyenViet: It's been solved by removing the `getAttributionSpansAt` method

super_editor/lib/src/default_editor/common_editor_operations.dart
```diff
@@ -1576,7 +1612,20 @@ class CommonEditorOperations {
...
     required String url,
   }) async {
-    final response = await http.get(Uri.parse(url));
+    late http.Response response;
+
+    // This function throws [SocketException] when the [url] is not valid.
+    // For instance, when typing for https://flutter.dev, it throws
+    // Unhandled Exception: SocketException: Failed host lookup: 'f'
+    //
+    // It doesn't affect any functionality, but it throws exception and preventing
```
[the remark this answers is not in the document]
ThangVuNguyenViet: Okay I'll make the comment clearer. What I'm saying is, when the user types https://f|, it immediately calls request and throws exception

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
ThangVuNguyenViet: I should've looked at the file changed page more carefully when creating a new PR then. Will take a close look at these.

super_editor/lib/src/default_editor/common_editor_operations.dart
```diff
@@ -1377,6 +1383,36 @@ class CommonEditorOperations {
...
+    required TextNode textNode,
+    required int textOffset,
+    required Set<Attribution> currentAttribution,
+  }) {
+    final linkAttributionSpan = textNode.text.spans
+        // Note: We can't use the method [AttributedSpans.getAttributionSpansAt] because it seems
+        // like the dependency doesn't pick the [attributed_text] package in the dependency_override.
+        // Therefore, until the package [attributed_text] is updated, we'll use this method instead.
+        .getAttributionSpansInRange(
+          attributionFilter: (attr) => attr is LinkAttribution,
```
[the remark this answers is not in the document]
ThangVuNguyenViet: I wanted to make the method more general about what attribution to keep when inserting, like `tokenizedUpstreamWord`, or `convertParagraphByPatternMatching`

super_editor/lib/src/default_editor/common_editor_operations.dart
```diff
@@ -2283,9 +2340,26 @@ class _PasteEditorCommand implements EditorCommand {
       InsertTextCommand(
         documentPosition: _pastePosition,
         textToInsert: splitContent.first,
-        attributions: attributionsAtPasteOffset,
+        // For [LinkAttribution], we don't apply to the pasted text, but rather
```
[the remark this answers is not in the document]
ThangVuNguyenViet: Yea it does. It's in the test `when pasting at the end of a link, convert URLs and prevent expanding link` in file `common_editor_operations_test.dart`

super_editor/test/src/default_editor/document_input_ime_test.dart
ThangVuNguyenViet: I actually wanted that for IME. I first thought by using `forIOS` should simulate IME input when using `typeKeyboardText`, but it isn't the case. And since we're migrating to the test robot, we're no longer accessing `SoftwareKeyboardHandler`, so how do we simulate the IME input?

super_editor/lib/src/default_editor/common_editor_operations.dart
```diff
@@ -1355,12 +1355,18 @@ class CommonEditorOperations {
...
+      currentAttribution: composer.preferences.currentAttributions,
+    );
+
     editorOpsLog.fine("Executing text insertion command.");
     editor.executeCommand(
       InsertTextCommand(
         documentPosition: composer.selection!.extent,
         textToInsert: text,
-        attributions: composer.preferences.currentAttributions,
+        attributions: filteredAttributions,
```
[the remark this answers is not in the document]
ThangVuNguyenViet: I tested this by running it, it works fine. However, I might need the `flutter_test_robot` to include `pressCmdB` or anything to add attribution

https://user-images.githubusercontent.com/39979207/175276429-2233e788-f617-4306-b5b5-1b730c5d2f6f.mov

super_editor/test/src/default_editor/common_editor_operations_test.dart
```diff
@@ -158,5 +163,313 @@ void main() {
         expect(composer.selection!.extent.nodePosition, const TextNodePosition(offset: 0));
       });
     });
+
+    // This test group covers cases related to link encoding when pasting. This
+    // is not neccessary a desktop platform test, but it utilizes the simulation
+    // of pasting on desktop
+    //
+    // Note: This covers cases on mobile as well, so separated tests for mobile
```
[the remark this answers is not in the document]
ThangVuNguyenViet: We can't do that because this test depends on `pressCmdV` to simulate pasting action. As discussed, this test is not testing pasting action on platforms, but link encoding instead. Choosing Mac as a platform for this test is just an option

super_editor/lib/src/default_editor/common_editor_operations.dart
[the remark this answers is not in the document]
matthew-carroll: I just added CMD+B to `flutter_test_robots` in v0.0.16

This video doesn't appear to reproduce what I described in my original comment. You should start with blank text and then:

type "abc"
press CMD+B
paste a link
type "def"

Once you've done that, you should see "abc**https://somelink.comdef**"

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
matthew-carroll: Maybe these spaces came from someone else's editor. I guess we can leave the spaces. But let's make sure that newlines remain where they are. I think I'm just using dartfmt with 120 character lines.

super_editor/lib/src/default_editor/text_tools.dart
```diff
@@ -155,3 +157,27 @@ TextDirection getParagraphDirection(String text) {
...
+///
+/// Giving a positive [extentOffset] would result in a selection expanded
+/// in upstream direction.
+///
+/// Return a collapsed selection by
+DocumentSelection documentPositionToDocumentSelection({
+  required DocumentPosition documentPosition,
+  int extentOffset = 0,
+}) {
+  assert(documentPosition.nodePosition is TextNodePosition);
```
matthew-carroll, line 174: I think this came up in the last review, but this method isn't scoped correctly. If you take in a `DocumentPosition` then you need to be able to work with `DocumentPosition`s. If what you want is a `TextNodePosition` then that's what this method should take in. 

I'm also not sure why this method exists, which indicates that the name of the method and the dart doc for the method should be significantly different. For example `documentPositionToDocumentSelection` is kind of true, in the most technical sense, but that name doesn't indicate what we're actually doing in this method, or why. A `DocumentPosition` is a piece of a `DocumentSelection` and therefore you can't convert one into the other. So this method is doing something very different from what the name suggests.

.vscode/launch.json
```diff
@@ -0,0 +1,14 @@
+{
+    // Use IntelliSense to learn about possible attributes.
```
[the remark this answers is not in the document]
ThangVuNguyenViet: Gosh. Sorry, I wasn't intended to push this

super_editor/lib/src/default_editor/text_tools.dart
ThangVuNguyenViet: Well the reason for taking in the `DocumentPosition` and using the `TextNodePosition` is that the function is returning a `DocumentSelection` which is constructed from a given `DocumentPosition` and its `TextNodePosition.offset`.

But in case you find this doesn't make much sense, I'll delete it then.

super_editor/test/super_editor/document_test_tools.dart
```diff
@@ -67,6 +67,13 @@ class TestDocumentSelector {
     );
   }
 
+  TestDocumentConfigurator withSingleLinkParagraph() {
```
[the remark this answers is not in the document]
ThangVuNguyenViet: Okay sure. I thought this is what the "challenge" is all about

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
```diff
@@ -919,6 +919,61 @@ void main() {
...
+          // Enable bold attribution
+          await tester.pressCmdB();
+
+          // Type some text by simulating hardware keyboard key presses.
+          await tester.typeKeyboardText('Go to ');
+
+          // Ensure that the link is not being expanded, and bold attribution is preserved
+          expect(
+            SuperEditorInspector.findDocument(),
+            equalsMarkdown("**Go to **[https://google.com](https://google.com)"),
```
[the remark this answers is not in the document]
ThangVuNguyenViet: What I'm trying to test here is preventing link expansion, which is part of this PR. What I just added here is ensuring other attributes are not being affected

super_editor/test/src/default_editor/common_editor_operations_test.dart
```diff
@@ -158,5 +163,143 @@ void main() {
...
+    // is not necessary
+    group('link encoding when pasting', () {
```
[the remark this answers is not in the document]
ThangVuNguyenViet: `splits a link in two when typing in the middle of the link`

We don't split a link in two when typing Matt, we only split when we paste it. Typing in the middle of a link should expand that link

super_editor/test/src/default_editor/document_input_ime_test.dart
```diff
@@ -268,6 +269,76 @@ void main() {
         );
       });
     });
+
+    group('inserting typing characters near a link', () {
+      testWidgets('does not expand the link when typing before the link', (tester) async {
```
matthew-carroll, line 274: Shouldn't this be `testWidgetsOnMobile()`? And if this should apply to both Android and iOS, then you should remove the `forIOS()` configuration in the test.

super_editor/test/src/default_editor/document_input_ime_test.dart
```diff
@@ -268,6 +269,76 @@ void main() {
...
+        );
+
+        // Ensure that the link is not being expanded
+        expect(
+          SuperEditorInspector.findDocument(),
+          equalsMarkdown("Go to [https://google.com](https://google.com)"),
+        );
+      });
+
+      testWidgets('prevent expanding the link when inserting at the end', (tester) async {
```
matthew-carroll, line 308: Read this test name with after the group above. They don't go together.

This test is essentially a variation of the previous test, so I would expect a similar name:
"does not expand the link when typing before the link" -> "does not expand the link when typing after the link"

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
```diff
@@ -919,6 +919,54 @@ void main() {
           Platform.setTestInstance(null);
         });
       });
+
+      group('inserting near links', () {
+        // Note:This is a Mac test because it depends on `pressCmdB` to enable a bold attribution
+        testWidgetsOnMac('does not expand the link when inserting at the start', (tester) async {
```
matthew-carroll, line 925: This test name should match the other test name that does the same thing

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
```diff
@@ -919,6 +919,54 @@ void main() {
...
+          // Configure and render a document.
+          await tester //
+              .createDocument()
+              .withCustomContent(_singleParagraphWithLinkDoc())
+              .forDesktop()
```
matthew-carroll, line 930: If the test is configured for a Mac then you shouldn't need to call `forDesktop()`, right?

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
```diff
@@ -919,6 +919,54 @@ void main() {
...
+              .pump();
+
+          // Place the caret in the first paragraph at the start of the link.
+          await tester.placeCaretInParagraph('1', 0);
+
+          // Type some text by simulating hardware keyboard key presses.
+          await tester.typeKeyboardText('Go to ');
+
+          // Ensure that the link is not being expanded, and bold attribution is preserved
```
matthew-carroll, line 939: "Ensure that the link is unchanged" - Also, I don't see any bold in this expectation.

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
```diff
@@ -919,6 +919,54 @@ void main() {
           Platform.setTestInstance(null);
         });
       });
+
+      group('inserting near links', () {
+        // Note:This is a Mac test because it depends on `pressCmdB` to enable a bold attribution
```
matthew-carroll, line 924: This test is not pressing `CMD+B`. It doesn't look like it needs to be a Mac test.

super_editor/test/src/default_editor/super_editor_content_conversion_test.dart
```diff
@@ -0,0 +1,362 @@
...
+import 'package:flutter_test_robots/flutter_test_robots.dart';
+import 'package:super_editor/super_editor.dart';
+
+import '../../super_editor/document_test_tools.dart';
+import '../../super_editor/supereditor_inspector.dart';
+import '../../super_editor/supereditor_robot.dart';
+import '../../test_tools.dart';
+
+void main() {
+  // This test group covers cases related to link encoding when pasting. This
```
matthew-carroll, line 11: This comment should appear at the top of the file because this file will be used for all content conversion tests, not just links. Also, I think the first paragraph of this comment can be deleted. 

You should create a group under "SuperEditor" called "link conversion" and then adjust the other groups and tests accordingly.

The "Note:" can be moved to whichever group it actually applies to.

super_editor/test/src/default_editor/super_editor_content_conversion_test.dart
```diff
@@ -0,0 +1,362 @@
...
+        await tester //
+            .createDocument()
+            .withSingleEmptyParagraph()
+            .forDesktop()
+            .pump();
+
+        // Tap to place the caret in the first paragraph.
+        await tester.placeCaretInParagraph("1", 0);
+
+        await tester.pressCmdV();
```
matthew-carroll, line 33: Above this line, comment "// Simulate the user pasting content from clipboard"

super_editor/test/src/default_editor/super_editor_content_conversion_test.dart
```diff
@@ -0,0 +1,362 @@
...
+
+        // Ensure that the link is not being expanded
+        expect(
+          SuperEditorInspector.findDocument(),
+          equalsMarkdown("[https://google.com](https://google.com)Link: [https://flutter.dev](https://flutter.dev)"),
+        );
+      });
+    });
+
+    group("apply toggled attribution to pasted text", () {
```
matthew-carroll, line 178: I don't think "toggled attribution" has a meaning. I think what you want is "composer attributions"

super_editor/test/src/default_editor/super_editor_content_conversion_test.dart
```diff
@@ -0,0 +1,362 @@
...
+
+        await tester.pressCmdV();
+
+        await tester.typeKeyboardText('. Hello Mars');
+
+        // Ensure that URLs are converted into links
+        expect(
+          SuperEditorInspector.findDocument(),
+          equalsMarkdown(
+            '**Hello world. Hello Mars**',
```
matthew-carroll, line 203: This test doesn't look relevant to this file. Where's the link conversion happening in this test? It looks like you're testing standard `AttributdText` behavior, but that's already tested in other files.

matthew-carroll (commented):
I added comments in your tests. Most of my comments apply to multiple places. Please make all relevant changes.

super_editor/test/src/default_editor/document_input_ime_test.dart
```diff
@@ -268,6 +269,82 @@ void main() {
...
+      testWidgets('prevent expanding the link when inserting at the start', (tester) async {
+        // Configure and render a document.
+        final testerDocumentContext = await tester //
+            .createDocument()
+            .withSingleLinkParagraph()
+            .forIOS()
+            .autoFocus(true)
+            .pump();
+
+        // Place the caret in the first paragraph at the start of the link.
```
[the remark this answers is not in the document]
ThangVuNguyenViet: I found the issue though. `TesterDocumentContext.composer` returning from pumping a `SuperEditor` isn't getting an update from placing a caret, because it was not passed to the `SuperEditor`. I've made the change accordingly

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
ThangVuNguyenViet: Sorry, these were comments from previous test implementations, which I forgot to update

super_editor/test/src/default_editor/super_editor_content_conversion_test.dart
ThangVuNguyenViet: I put them in the common ops test then. They're related to pasting operation

matthew-carroll: I don't think this PR needs to test generic pasting behavior. Let's focus specifically on link encoding when pasting.

super_editor/lib/src/default_editor/common_editor_operations.dart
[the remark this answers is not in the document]
matthew-carroll: But it's not generic...your code is looking for `LinkAttribution`s exclusively...

ThangVuNguyenViet:
@matthew-carroll I think the comment you're talking about is about simulating IME input in tests. I've made a fix for that issue, and also the changes you requested.

For all the comments that I didn't respond to, it means "Okay, I'll make updates accordingly" by default. I was trying to avoid spamming you actually, but I'll start responding to every comment if that's needed.

super_editor/test/src/default_editor/super_editor_content_conversion_test.dart
ThangVuNguyenViet: Well, do you want me to separate the tests and put them in another PR?

matthew-carroll:
@ThangVuNguyenViet you don't need to respond to every comment, but you should make all the requested changes before requesting another review.

super_editor/test/src/default_editor/super_editor_content_conversion_test.dart
matthew-carroll: I'm not sure that these tests will ever be needed. You can set them aside somewhere else if you want to. But let's remove them from this PR.

ThangVuNguyenViet:
Yea I thought I checked all the boxes, but turns out I didn't. Now this should be fully ready for review

super_editor/test/src/default_editor/super_editor_content_conversion_test.dart
ThangVuNguyenViet: Well, these tests just happen to be out of the scope of this PR, but I'm sure it is necessary. It makes sure any toggled-on composer attributions will be applied to the pasted text, which is what you requested.  

> This video doesn't appear to reproduce what I described in my original comment. You should start with blank text and then:
> 
> type "abc"
> press CMD+B
> paste a link
> type "def"
> 
> Once you've done that, you should see "abchttps://somelink.comdef/"

Based on that, I think you thought this functionality already existed, but it's not.

If I were to remove the tests, should I also remove the implementation that goes with this as well?

ThangVuNguyenViet: It's in `common_editor_operations.dart`. It's in 2 closed blocks, so it's better with a pic.

image

In order to apply any toggled-on attribution to the pasted text, we have to use `currentAttributions` instead of `attributionsAtPasteOffset`. The filtering link part is for link splitting.

matthew-carroll: I'm still not following your explanation. You pasted a screenshot that's filled with references to links, but then you say that this behavior isn't about links. So I'm not sure what to tell you. As for the implementation, you should include whatever behavior is required to implement the feature, and you shouldn't include any behavior that's not required to implement the feature...

But I don't think we need those tests right now. I'm trying to reduce this PR as much as possible so that we can get through it.

ThangVuNguyenViet: Okay sure. I'll remove the tests and the implementation then.

ThangVuNguyenViet: @matthew-carroll I've removed the tests as well as the implementation relating to that

ThangVuNguyenViet:
@matthew-carroll I made the changes you requested in the comments. It's ready for another review

super_editor/test/src/default_editor/test_text_tools.dart
```diff
@@ -0,0 +1,28 @@
...
+    group('expand position to word', () {
+      test('does not expand when there is no word next to the caret', () {
+        // Notice there is a space at first
+        const text = ' SuperEditor is awesome';
+        // Place a collapsed selection at "| SuperEditor is awesome"
+        expect(
+          const TextSelection.collapsed(offset: 0),
+          expandPositionToWord(text: text, textPosition: const TextPosition(offset: 0)),
+        );
+      });
```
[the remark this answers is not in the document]
matthew-carroll: It looks like this comment still isn't addressed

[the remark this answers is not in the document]
matthew-carroll: I take that back. There are tests above that do this, but the test names didn't specify this.

super_editor/lib/src/default_editor/common_editor_operations.dart
```diff
@@ -1302,12 +1302,19 @@ class CommonEditorOperations {
     final textNode = editor.document.getNode(composer.selection!.extent) as TextNode;
     final initialTextOffset = (composer.selection!.extent.nodePosition as TextNodePosition).offset;
 
+    // Inserting text and both ends of a link should not expand the link attribution
+    Set<Attribution> filteredAttributions = _filterLinkAttributionAtBothEndsOfLink(
```
matthew-carroll, line 1306: I think the use of the term "filter" here makes it more difficult to understand than it needs to be.

I think the following would be more clear:

```dart
final attributionsForInsertedText = _removeLinksBeforeAndAfterInsertionOffset(...);
```

super_editor/lib/src/default_editor/common_editor_operations.dart
```diff
@@ -1324,6 +1331,34 @@ class CommonEditorOperations {
     return true;
   }
 
+  /// Filters [LinkAttribution] from the given [currentAttributions] if the user
```
matthew-carroll, line 1334: Let's replace this comment with:

```dart
/// Removes link attributions from [currentAttributions], if the link begins at the [textOffset], or
/// ends at the [textOffset], so that the link attribution doesn't expand with new text.
///
/// A new `Set<Attribution>` is returned. [currentAttributions] is not altered.
```

super_editor/lib/src/default_editor/common_editor_operations.dart
```diff
@@ -1324,6 +1331,34 @@ class CommonEditorOperations {
...
+  /// is inserting text at both ends of a link
+  Set<Attribution> _filterLinkAttributionAtBothEndsOfLink({
+    required TextNode textNode,
```
matthew-carroll, line 1337: This parameter can be an `AttributedText` instead of a `TextNode` and still get the job done.

super_editor/lib/src/default_editor/common_editor_operations.dart
```diff
@@ -1589,11 +1637,20 @@ class CommonEditorOperations {
     final textNode = editor.document.getNode(composer.selection!.extent) as TextNode;
     final initialTextOffset = (composer.selection!.extent.nodePosition as TextNodePosition).offset;
 
+    // Inserting text and both ends of a link should not expand the link attribution
+    final attributions = ignoreComposerAttributions
```
matthew-carroll, line 1641: This is a good example of my earlier question about where this behavior belongs. Notice that if this behavior happened inside of `InsertTextCommand` then we wouldn't need to repeat it in multiple places where we use `InsertTextCommand`. This actually seems to be a behavior related to inserting text, regardless of which operation causes the text to be inserted.

super_editor/lib/src/default_editor/common_editor_operations.dart
```diff
@@ -2226,13 +2283,33 @@ class _PasteEditorCommand implements EditorCommand {
...
 
+      // If a link spans across the past location, we split that link in two. Part
+      // of the link sits before the pasted text, and the other part of the link
+      // sits after the pasted text. Both of the link parts continue to reference
+      // the original URL, even if the text was a URL.
+      //
+      // The attribution splitting happens automatically in the text inserting execution, when the
+      // newly inserted text's attributions doesn't contain the original attribution. Hence, we only
+      // need to filter out the original link attribution.
+      final filteredLinkAttribution = attributionsAtPasteOffset //
```
matthew-carroll, line 2294: This is the 3rd place that the same behavior is replicated, followed by an `InsertTextCommand`.

super_editor/lib/src/default_editor/common_editor_operations.dart
```diff
@@ -2226,13 +2283,33 @@ class _PasteEditorCommand implements EditorCommand {
...
       // Paste the first piece of content into the selected TextNode.
       InsertTextCommand(
         documentPosition: _pastePosition,
         textToInsert: splitContent.first,
-        attributions: attributionsAtPasteOffset,
+        attributions: filteredLinkAttribution,
       ).execute(document, transaction);
 
+      // Check for url in the pasted text and apply [LinkAttribution] appropriately
+      _convertURLsInPastedTextToLink(
```
matthew-carroll, line 2306: I'm wondering about the placement of this method. I can't remember if we're doing any other content conversions when we paste code, but if we are, where are they? Shouldn't URL conversion happen wherever those other conversions happen?

If there are no other conversions, then we can leave this as-is, for now.

super_editor/lib/src/default_editor/common_editor_operations.dart
```diff
@@ -2290,4 +2367,41 @@ class _PasteEditorCommand implements EditorCommand {
...
+
+  void _convertURLsInPastedTextToLink({
+    required String text,
+    required TextNode textNode,
+    required Document document,
+    required DocumentEditorTransaction transaction,
+  }) {
+    final textPosition = _pastePosition.nodePosition as TextNodePosition;
+
+    for (var wordTextSelection in text.wordTextSelections) {
```
matthew-carroll, line 2379: When using for-each-in syntax like this, the variable should be `final`, e.g., `final wordRange in words`.

Also, we should calculate the word ranges before the for-loop to ensure that we don't call that method over and over.

matthew-carroll (commented):
Getting closer, but still some things to clean up.

super_editor/test/src/default_editor/common_editor_operations_test.dart
```diff
@@ -158,5 +164,57 @@ void main() {
...
+            .withCustomContent(_singleParagraphWithLinkDoc())
+            .forDesktop()
+            .pump();
+
+        // Tap to place the caret in the first paragraph.
+        await tester.placeCaretInParagraph("1", 11);
+        // Simulate the user pasting content from clipboard
+        await tester.pressCmdV();
+
+        // Ensure that the link is splitted
```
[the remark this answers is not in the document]
ThangVuNguyenViet: Oh now I know split is also an adjective.

super_editor/lib/src/default_editor/text.dart
```diff
@@ -187,6 +188,28 @@ extension DocumentSelectionWithText on Document {
   }
 }
 
+extension WordsDocumentPosition on String {
+  /// Returns list of [TextSelection] on each word in text
+  List<TextSelection> get wordTextSelections {
+    final List<TextSelection> textSelections = [];
+    var offset = 0;
+
+    while (offset < length) {
```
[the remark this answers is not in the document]
ThangVuNguyenViet: Wow. No, I have never thought of that. I'll make those tests right away

super_editor/lib/src/default_editor/common_editor_operations.dart
ThangVuNguyenViet: Well, as much as I know, there were `linkify` and `_processUrlNode` actions in the `convertParagraphByPatternMatching` in `common_editor_operations.dart` triggered by text insertion. To be honest, I have no idea what it does.

For conversion when pasting, it is no IMO.

super_editor/lib/src/default_editor/common_editor_operations.dart
ThangVuNguyenViet: This part is different from the other 2. Inserting text in the middle of a link would expand the link, whereas pasting text should split the link. Therefore, inserting text should only remove link attribution if the current offset is sitting at either end of the link, whereas pasting text should remove link attribution regardless of the current offset.

super_editor/lib/src/default_editor/common_editor_operations.dart
ThangVuNguyenViet: Yea you're right. I've moved that into `InsertTextCommand`.

super_editor/lib/src/default_editor/common_editor_operations.dart
ThangVuNguyenViet: Well, there are some places in the source code that are currently using the for-var. Would it make sense to have another clean-up PR?

super_editor/lib/src/default_editor/text.dart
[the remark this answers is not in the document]
ThangVuNguyenViet: I've added emoji tests for this function. Just to confirm, `🐢🐢` takes 4 code points, right? I saw other places in the code base that treats this emoji as 2 code points, and I've also try pasting link with an emoji, and that works too.

super_editor/lib/src/default_editor/common_editor_operations.dart
matthew-carroll: If those other places are in this PR, then we should fix them in this PR. If the other places aren't related to this PR's code, then you can ignore them.

super_editor/lib/src/default_editor/common_editor_operations.dart
```diff
@@ -2226,13 +2239,33 @@ class _PasteEditorCommand implements EditorCommand {
         }
       }
 
+      // If a link spans across the past location, we split that link in two. Part
```
matthew-carroll, line 2242: "past" => "paste"

super_editor/lib/src/default_editor/common_editor_operations.dart
```diff
@@ -2226,13 +2239,33 @@ class _PasteEditorCommand implements EditorCommand {
...
+      // of the link sits before the pasted text, and the other part of the link
+      // sits after the pasted text. Both of the link parts continue to reference
+      // the original URL, even if the text was a URL.
+      //
+      // The attribution splitting happens automatically in the text inserting execution, when the
```
matthew-carroll, line 2247: I think you can remove this 2nd paragraph of text. It doesn't seem to add clarity to the first paragraph.

super_editor/lib/src/default_editor/common_editor_operations.dart
```diff
@@ -2290,4 +2323,42 @@ class _PasteEditorCommand implements EditorCommand {
 
     editorOpsLog.fine('Done with paste command.');
   }
+
+  void _convertURLsInPastedTextToLink({
```
matthew-carroll, line 2327: Please add a doc comment that says what this method does.

super_editor/lib/src/default_editor/common_editor_operations.dart
```diff
@@ -2290,4 +2323,42 @@ class _PasteEditorCommand implements EditorCommand {
...
+          nodePosition: textPosition.copyWith(
+            offset: textPosition.offset + wordBoundary.end,
+          ),
+        ),
+      );
+
+      if (link != null && link.hasScheme && link.hasAuthority) {
+        // Valid url. Apply [LinkAttribution] to the url
+        final linkAttribution = LinkAttribution(url: link);
+        AddTextAttributionsCommand(
```
matthew-carroll, line 2357: Commands are intended to be run from the editor, rather than run from the middle of other code like this. What would it look like if you directly applied the attributions without this command?

super_editor/test/src/default_editor/document_input_ime_test.dart
```diff
@@ -304,3 +375,55 @@ void _expectTextEditingValue({
...
+      )
+    ],
+  );
+}
+
+extension on WidgetTester {
+  // TODO: Remove this when `SuperTestRobot` support insert IME
+  Future<void> insertTextIME({
+    required SoftwareKeyboardHandler softwareKeyboardHandler,
+    required String text,
```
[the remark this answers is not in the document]
matthew-carroll: Looks like this still needs to be addressed.

super_editor/test/src/default_editor/common_editor_operations_test.dart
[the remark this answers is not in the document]
matthew-carroll: Looks like this still needs to be addressed.

super_editor/test/src/default_editor/text_tools_test.dart
```diff
@@ -0,0 +1,112 @@
...
+            expandPositionToWord(text: text, textPosition: const TextPosition(offset: text.length));
+        expect(
+          expandedSelection,
+          const TextSelection(baseOffset: 15, extentOffset: 22),
+        );
+      });
+
+      test('expand when the caret is just before emoji', () {
+        // Notice there is a space at first
+        const text = '🐢A';
```
matthew-carroll, line 79: When the whole string is one word, we can't say with much confidence that word boundaries are being calculated appropriately. I think all of these tests should include at least 3 words, so that you can verify a word at the beginning of the string, at the end of the string, and in the middle of the string.

super_editor/test/src/default_editor/document_input_ime_test.dart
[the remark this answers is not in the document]
ThangVuNguyenViet: Oh gosh. I misunderstood what you meant. I changed the method name, not the text parameter

super_editor/lib/src/default_editor/common_editor_operations.dart
ThangVuNguyenViet: I don't get what you mean here. Can you elaborate?

matthew-carroll: You shouldn't execute a command from within a command. Instead, you should directly make the changes that you need to make.

ThangVuNguyenViet: Okay. I got it

ThangVuNguyenViet:
@matthew-carroll there's something weird. I pulled & merged the main branch to this branch, and tests related to pasting failed in my local (except tests in this branch though). However, tests pass on the CI, as you can see.

matthew-carroll:
@ThangVuNguyenViet did you resolve that issue? I'm not sure what I can do with that information...

super_editor/lib/src/default_editor/text.dart
```diff
@@ -1080,12 +1117,52 @@ class InsertTextCommand implements EditorCommand {
     }
 
     final textOffset = (documentPosition.nodePosition as TextPosition).offset;
+
+    // Inserting text and both ends of a link should not expand the link attribution
```
matthew-carroll, line 1121: "text and both ends" -> "text at either end"

super_editor/test/src/default_editor/super_editor_content_conversion_test.dart
```diff
@@ -0,0 +1,211 @@
...
+import '../../super_editor/supereditor_inspector.dart';
+import '../../super_editor/supereditor_robot.dart';
+import '../../test_tools.dart';
+
+void main() {
+  group('SuperEditor', () {
+    group('link conversion', () {
+      // This is not neccessary a desktop platform test suite, but it utilizes the
+      // simulation of pasting on desktop
+      // Note: This covers cases on mobile as well, so separated tests for mobile
```
matthew-carroll, line 15: Can you remind me why we decided to only run these tests on Mac instead of running them on all platforms?

super_editor/test/super_editor/document_test_tools.dart
```diff
@@ -444,7 +444,8 @@ class EquivalentDocumentMatcher extends Matcher {
     bool nodeTypeOrContentMismatch = false;
 
     if (_expectedDocument.nodes.length != actualDocument.nodes.length) {
-      messages.add("expected ${_expectedDocument.nodes.length} document nodes but found ${actualDocument.nodes.length}");
+      messages
```
matthew-carroll, line 447: Can you revert this formatting change so that this file doesn't appear in the change log?

super_editor/test/src/default_editor/super_editor_content_conversion_test.dart
ThangVuNguyenViet: Well because this test suite tests link conversion on pasting. We've already had a test suite for pasting on all platforms, so all we need is to pick a platform and test link conversion on pasting on it. We picked Mac.

super_editor/test/super_editor/document_test_tools.dart
ThangVuNguyenViet: That's odd. Gitlens doesn't report that was my change. But sure, I'll do it.

super_editor/lib/src/default_editor/common_editor_operations.dart
```diff
@@ -2290,4 +2319,35 @@ class _PasteEditorCommand implements EditorCommand {
...
     editorOpsLog.fine('Done with paste command.');
   }
+
+  /// Check for each word in the pasted text. If the word is a valid URL,
+  /// convert it into link by adding [LinkAttribution]
+  void _convertURLsInPastedTextToLink({
+    required String text,
+    required TextNode textNode,
+    required Document document,
+    required DocumentEditorTransaction transaction,
```
matthew-carroll, line 2329: It doesn't look like the `transaction` is needed in this method.

super_editor/lib/src/default_editor/common_editor_operations.dart
```diff
@@ -2290,4 +2319,35 @@ class _PasteEditorCommand implements EditorCommand {
 
     editorOpsLog.fine('Done with paste command.');
   }
+
+  /// Check for each word in the pasted text. If the word is a valid URL,
```
matthew-carroll, line 2323: This comment reveals more internal details than it needs to. Mentioning "words" will only confuse people, because people don't think of a URL as a "word". That's mostly an implementation detail.

What this method does is: `Finds all URLs in the text within the given [textNode] and applies a [LinkAttribution] to each one.`

super_editor/lib/src/default_editor/common_editor_operations.dart
```diff
@@ -2290,4 +2319,35 @@ class _PasteEditorCommand implements EditorCommand {
...
+  /// convert it into link by adding [LinkAttribution]
+  void _convertURLsInPastedTextToLink({
+    required String text,
+    required TextNode textNode,
```
matthew-carroll, line 2327: Why are we taking in `text` and a `textNode`? You're analyzing the `text` but then you're editing the `TextNode`. Something is wrong there.

super_editor/lib/src/default_editor/common_editor_operations.dart
ThangVuNguyenViet: Yea you're right. And the `document` too

super_editor/lib/src/default_editor/common_editor_operations.dart
ThangVuNguyenViet: Sure. I'll change that

super_editor/lib/src/default_editor/common_editor_operations.dart
ThangVuNguyenViet: Maybe it's the naming problem? `text` here is the pasted text, taken from the `_content`, which is different from the `textNode.text.text`.

matthew-carroll: In the implementation you're calculating text offsets from `text` and then you're applying those text offsets to `textNode.text`. I see that you're adding the paste offset when you do that. It would probably be safer to apply the link attributions to the pasted text, and then insert the pasted text with the link attributions into the `textNode`. I think this means finding the links before running the insertion command, instead of after.

ThangVuNguyenViet: When you say "insert the pasted text with the link attributions into the `textNode`", I'm thinking about inserting `text` and a list of `LinkAttribution`s with `SpanRange` in a single method. I can't see we can do that currently.

I'm going to change this method into `_getURLsInPastedText` which will return a `List` of `AttributionSpan` and get that list before executing the `InsertTextCommand`, and then run a for-loop to add those attribution spans. What do you think about this? Perhaps this is what you meant

ThangVuNguyenViet:
@matthew-carroll I also replace my `insertTextIME` with your newly added ime text inserting API

super_editor/lib/src/default_editor/common_editor_operations.dart
```diff
@@ -2226,13 +2239,32 @@ class _PasteEditorCommand implements EditorCommand {
...
       InsertTextCommand(
         documentPosition: _pastePosition,
         textToInsert: splitContent.first,
-        attributions: attributionsAtPasteOffset,
+        attributions: attributionsForPastedText,
       ).execute(document, transaction);
 
+      // Apply [LinkAttribution] found in the pasted text
+      for (final linkAttributionSpan in linkAttributionSpans) {
+        textNode.text.addAttribution(
```
matthew-carroll, line 2262: This code still has the same order of operations issue as the previous code. You're calculating attribution spans based on the pasted text fragment and then you're applying link attributions to the text node. This creates an opportunity for problems in between.

If you're going to calculate link attribution spans on the text fragment that you're pasting, then you should apply those link attributions to the text that you're inserting. Not the text after the insertion.

ThangVuNguyenViet: Took me days to really understand what you mean (plus days of recovering from covid though). I added `attributionSpans` as parameter to the `InsertTextCommand`, and passing down `AttributedText.insertString`. The spans in `attributionSpans` should only depend on the pasting text, not the text in the `textNode`.

A little problem occurs because we're having 2 parameters for adding `Attribution`s. The `attributionSpans` can overlap the `attributions`, but having `attributions` existing surely adds convenience. IMO, it's okay to have them coexist, but not sure if you think the same.

Please let me know if that's not a good approach.

matthew-carroll:
@ThangVuNguyenViet I pulled down your PR branch so that I could make final adjustments before merging in. That PR is at #767 - I'm going to close this PR in favor of that one. I'll follow up via email.

PR #633, opened 2022-06-21
Paint multiple user selections (Resolves #631)

### LINKED ISSUE #631: SuperEditor: Paint multiple user selections
Add support for multiple user selections in a SuperEditor. These selections need a logical representation, as well as painting support.

A design discussion is available at https://github.com/superlistapp/super_editor/discussions/604

This ticket does not include painting carets for those selections.

matthew-carroll:
Paint multiple user selections (Resolves #631)

The selections are configured as per #604

### DIFF
--- a/super_editor/example/lib/demos/components/demo_text_with_hint.dart
+++ b/super_editor/example/lib/demos/components/demo_text_with_hint.dart
@@ -163,8 +163,6 @@ class HeaderWithHintComponentBuilder implements ComponentBuilder {
       return null;
     }
 
-    final textSelection = componentViewModel.selection;
-
     return TextWithHintComponent(
       key: componentContext.componentKey,
       text: componentViewModel.text,
@@ -188,8 +186,7 @@ class HeaderWithHintComponentBuilder implements ComponentBuilder {
       hintStyleBuilder: (Set<Attribution> attributions) => _textStyleBuilder(attributions).copyWith(
         color: const Color(0xFFDDDDDD),
       ),
-      textSelection: textSelection,
-      selectionColor: componentViewModel.selectionColor,
+      styledSelections: componentViewModel.styledSelections,
       showCaret: componentViewModel.caret != null,
     );
   }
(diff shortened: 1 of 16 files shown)


matthew-carroll:
@roughike @miguelcmedeiros - Do you guys wanna give this PR a try for multi-user selection and see how it works? It's probably a good idea to make sure it works the way you want it to, before investing more effort.

roughike:
Hey @matthew-carroll, the API looks like exactly what we need - can you rebase the PR so that it would be easier to take it for a test run?

matthew-carroll:
@roughike I just rebased

roughike:
Played around with this a bit.

1. Calling `setNonPrimarySelection("john", ..)` repeatedly with an updated `DocumentSelection` should change the existing selection for "john", right? This does not seem to be the case.

I wrapped the `setNonPrimarySelection("john", ...)` part in your example with a periodic timer like this:

```dart
    var i = 0;
    Timer.periodic(const Duration(seconds: 1), (_) {
      i++;
      print(i);
      _composer.setNonPrimarySelection(
        "john",
        DocumentSelection(
          base: DocumentPosition(
            nodeId: nodeId,
            nodePosition:
                TextNodePosition.fromTextPosition(TextPosition(offset: i)),
          ),
          extent: DocumentPosition(
            nodeId: nodeId,
            nodePosition:
                TextNodePosition.fromTextPosition(TextPosition(offset: 50)),
          ),
        ),
      );
    });
```

I expected that the selection start would update every second, but that does not happen.

2. When focusing an existing non-primary selection, it paints a blinking caret. This causes the document to have two blinking carets - my caret and John's caret (but John's caret is only displayed when his paragraph is focused). As far as I'm aware, we only want to display one blinking caret. See video below of the unexpected behavior:

https://user-images.githubusercontent.com/13744304/179974352-668e2dc7-938d-4ac0-a5b5-d36e5a689dcb.mov

matthew-carroll:
@roughike I pushed a fix for the two issues you mentioned. Give it another try and let me know if it works for you.

roughike:
@matthew-carroll works!

matthew-carroll:
@roughike @miguelcmedeiros - Should we merge these changes, or do you think that we need to implement more multi-user selections in this PR before we can be confident in the changes?

PR #634, opened 2022-06-21
Add .moveNode() to MutableDocument & DocumentEditorTransaction

roughike:
We want to have a way of moving a `DocumentNode` to an arbitrary index in the list of nodes with just one `notifyListeners()` call happening. 

The end result can be achieved with `document..deleteNode()..insertNode()` calls. But we only want to get a notification of the document after a `DocumentNode` has been moved, not a "partial state" where a node has been removed, and then the "final state" where the node exists on the right place of the document. This can't be achieved currently with the existing APIs.

Since there's no `moveNode()` method available, this PR introduces just that.

### DIFF
--- a/super_editor/lib/src/core/document_editor.dart
+++ b/super_editor/lib/src/core/document_editor.dart
@@ -100,10 +100,12 @@ class DocumentEditorTransaction {
     _document.deleteNodeAt(index);
   }
 
-  /// Moves the given [node] from its current index in the [Document] to the
-  /// given [toIndex].
-  void moveNode({required DocumentNode node, required int toIndex}) {
-    _document.moveNode(node: node, toIndex: toIndex);
+  /// Moves a [DocumentNode] matching the given [nodeId] from its current index
+  /// in the [Document] to the given [targetIndex].
+  ///
+  /// If none of the nodes in this document match [nodeId], throws an error.
+  void moveNode({required String nodeId, required int targetIndex}) {
+    _document.moveNode(nodeId: nodeId, targetIndex: targetIndex);
   }
 
   /// Replaces the given [oldNode] with the given [newNode]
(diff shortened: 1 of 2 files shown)


super_editor/test/src/core/document_editor_test.dart
```diff
@@ -3,6 +3,50 @@ import 'package:super_editor/super_editor.dart';
 
 void main() {
   group('MutableDocument', () {
+    test('it moves node from its current index to the given index ', () {
```
matthew-carroll, line 6: Can you add a few tests for boundary conditions:
 * what happens when this is called on an empty document
 * what happens when this is called with an invalid node ID
 * what happens when the node is moved to a negative index
 * what happens when the node is moved to an index beyond the length of the document

super_editor/lib/src/core/document_editor.dart
```diff
@@ -273,6 +279,15 @@ class MutableDocument with ChangeNotifier implements Document {
     return isRemoved;
   }
 
+  /// Moves the given [node] from its current index in the [Document] to the
+  /// given [toIndex].
+  void moveNode({required DocumentNode node, required int toIndex}) {
+    if (nodes.remove(node)) {
+      nodes.insert(toIndex, node);
+      notifyListeners();
+    }
```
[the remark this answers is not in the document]
roughike: I went with the nodeId & targetIndex approach. I'm assuming that the `nodeId` of the node you want to move is always known to you, but figuring out the index of the node might usually be an extra step.

super_editor/test/src/core/document_editor_test.dart
roughike: I wrote the missing tests.

matthew-carroll (approved):
LGTM

PR #639, opened 2022-06-22
Add test for caret blinking while typing in SuperTextField (Resolves #460)

### LINKED ISSUE #460: SuperTextField - caret blinks while typing
Tried on the latest version of super_editor:

```
  super_editor:
    git:
      url: https://github.com/superlistapp/super_editor.git
      path: super_editor
      ref: 57a7d72166272582cbbdedc752b6f328c0493f88
```

Actual behavior:
When you start typing, the cursor flashes

Expected behavior:
The cursor stays solid while typing

Quick video explaining the issue:
https://www.loom.com/share/9352fc3b1837434696d8af4b2e2c0c39dd

For some reason, the flashing seems a bit more prominent on web.

@matthew-carroll we're happy to sponsor a fix for this

angelosilvestre:
Add test for caret blinking while typing in SuperTextField. Resolves https://github.com/superlistapp/super_editor/issues/460

This issue was already solved with the introduction of `BlinkController`, so I tried to add a test to verify this behavior.

This test alone might not be enough to ensure the caret is solid while typing, because it does not check if `CaretPainter`  is repainting itself using the opacity from `BlinkController`. However, `BlinkController` has tests to ensure it notifies its listeners when it changes its visibility, so `CaretPainter` should be repainting.

I was able to simulate a failing test commenting the body of `jumpToOpaque`:

https://github.com/superlistapp/super_editor/blob/f7b0b5997c49253d1453d487ed6c5d7cb7d21f74/super_text_layout/lib/src/infrastructure/blink_controller.dart#L69-L77

I used the default `flashPeriod` (500 ms) in the test, but we don't have a way to configure `SuperTextField` flash period, so this test could break if the default `flashPeriod` is changed.

### DIFF
--- a/super_editor/test/super_textfield/super_textfield_caret_test.dart
+++ b/super_editor/test/super_textfield/super_textfield_caret_test.dart
@@ -7,19 +7,15 @@ import 'package:flutter_test_robots/flutter_test_robots.dart';
 
 import '../test_tools.dart';
 
-void main() {
-  group("SuperTextField", () {
-    testWidgetsOnDesktop("keeps caret solid while typing", (tester) async{
+void main() { 
+  group("SuperTextField", () {    
+    // Duration to switch between visible and invisible
+    const flashPeriod = Duration(milliseconds: 500);
+
+    testWidgetsOnDesktop("caret blinks at rest", (tester) async {
       // Configure BlinkController to animate, otherwise it won't blink
       BlinkController.indeterminateAnimationsEnabled = true;
-
-      // duration to switch between visible and invisible
-      const flashPeriod = Duration(milliseconds: 500);
-      
-      // interval between each key is pressed
-      // here we add one millissecond because BlinkController checks if the 
-      // ellapsed time is bigger than the flash period
-      final typingInterval = (flashPeriod ~/ 2) + const Duration(milliseconds: 1);
+      addTearDown(() => BlinkController.indeterminateAnimationsEnabled = false);
 
       await tester.pumpWidget(
         MaterialApp(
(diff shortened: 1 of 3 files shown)


super_editor/test/super_textfield/super_textfield_caret_test.dart
```diff
@@ -0,0 +1,70 @@
...
+      expect(_isCaretVisible(tester), false);
+
+      // Trigger another frame to make caret visible again
+      await tester.pump(flashPeriod + const Duration(milliseconds: 1));
+
+      // Ensure caret is visible
+      expect(_isCaretVisible(tester), true);
+
+      // Trigger another frame after half flash period
+      await tester.pump(const Duration(milliseconds: 250));
```
matthew-carroll, line 50: Please use `flashPeriod / 2` so that we can change the flash period and still have this work.

Also, please say why you're doing this in the comment, rather than what you're doing. What's the purpose of pumping for half a period?

super_editor/test/super_textfield/super_textfield_caret_test.dart
```diff
@@ -0,0 +1,70 @@
...
+
+      await tester.typeKeyboardText("abc");
+
+      // Trigger a frame that added to the previous frame 
+      // is bigger than the flash period
+      await tester.pump(const Duration(milliseconds: 300));
```
matthew-carroll, line 56: Here again, the comment should explain why you're doing this. And you should express this time in terms of the `flashPeriod`.

super_editor/test/super_textfield/super_textfield_caret_test.dart
```diff
@@ -0,0 +1,70 @@
...
+
+      // Ensure caret is visible
+      expect(_isCaretVisible(tester), true);
```
matthew-carroll, line 59: Please adjust the comment to say why the caret should be visible at this point

super_editor/test/super_textfield/super_textfield_caret_test.dart
```diff
@@ -0,0 +1,70 @@
...
+      // Trigger another frame to make caret visible again
+      await tester.pump(flashPeriod + const Duration(milliseconds: 1));
+
+      // Ensure caret is visible
+      expect(_isCaretVisible(tester), true);
+
+      // Trigger another frame after half flash period
+      await tester.pump(const Duration(milliseconds: 250));
+
+      await tester.typeKeyboardText("abc");
```
matthew-carroll, line 52: If you're testing appearance and timing while typing, then you should probably type one character at a time, and then pump a reasonable duration in between characters. For example, if you pump `flashPeriod/2` after every character, and you type 4 characters, and the caret is visible after every character, then that indicates that you've gone through 2 periods while typing and the caret never disappeared. So whether the caret started visible or invisible, if we've gone through 2 periods while typing, and the caret was visible the entire time, then typing is correctly preventing it from disappearing.

super_editor/test/super_textfield/super_textfield_caret_test.dart
```diff
@@ -0,0 +1,91 @@
...
+void main() {
+  group("SuperTextField", () {
+    testWidgetsOnDesktop("keeps caret solid while typing", (tester) async{
+      // Configure BlinkController to animate, otherwise it won't blink
+      BlinkController.indeterminateAnimationsEnabled = true;
+
+      // duration to switch between visible and invisible
+      const flashPeriod = Duration(milliseconds: 500);
+      
+      // interval between each key is pressed
```
matthew-carroll, line 19: Let's write comments like sentences:

```
Interval between each key press. We add one millisecond so that the blink timer executes just after the end of a flash period.
```

super_editor/test/super_textfield/super_textfield_caret_test.dart
```diff
@@ -0,0 +1,91 @@
...
+      // here we add one millissecond because BlinkController checks if the 
+      // ellapsed time is bigger than the flash period
+      final typingInterval = (flashPeriod ~/ 2) + const Duration(milliseconds: 1);
```
matthew-carroll, line 22: Do you actually need the +1ms? Did you try running this without it? First, you're adding more 1ms down below. Second, when typing multiple characters, the idea behind typing 4 of them is that you span multiple flash periods, so a millisecond here or there shouldn't make a difference...

super_editor/test/super_textfield/super_textfield_caret_test.dart
```diff
@@ -0,0 +1,91 @@
...
+
+      // Trigger a frame with an ellapsed time greater than the flashPeriod,
+      // so the caret should change from visible to invisible
+      await tester.pump(flashPeriod + const Duration(milliseconds: 1));
+
+      // Ensure caret is invisible after the flash period
+      expect(_isCaretVisible(tester), false);
+
+      // Trigger another frame to make caret visible again
+      await tester.pump(flashPeriod + const Duration(milliseconds: 1));
```
matthew-carroll, line 49: BTW, if it would make things easier to test, feel free to change the blink controller to `>=` instead of `>`. I don't think anyone is going to notice a difference of 1ms, but maybe it would make the tests easier to read.

super_editor/test/super_textfield/super_textfield_caret_test.dart
```diff
@@ -0,0 +1,91 @@
...
+
+      // Ensure caret is visible
+      expect(_isCaretVisible(tester), true);
+     
+      // Type and 'wait' for ~ half a period
```
matthew-carroll, line 54: Let's split this test in two. The first part of the test seems to verify that the caret blinks at rest. Then we verify the caret doesn't blink when typing. I think that's reasonably two different tests. It's kind of surprising as a reader to get down here and find typing behavior after the earlier verifications.

super_editor/test/super_textfield/super_textfield_caret_test.dart
```diff
@@ -0,0 +1,91 @@
...
+      expect(_isCaretVisible(tester), true); 
+
+      // Type and 'wait' for ~ half a period to end the second full period
+      await tester.typeKeyboardText("d");
+      await tester.pump(typingInterval);
+
+      // Ensure typing prevented caret from disappearing
+      expect(_isCaretVisible(tester), true); 
+
+      BlinkController.indeterminateAnimationsEnabled = false;
```
matthew-carroll, line 82: Would it make sense to add this to a `addTearDown()` right below where you set it to `true`?

```dart
addTearDown(() => BlinkController.indeterminateAnimationsEnabled = false);
```

I think if this test fails, this line won't run, and a bunch of other tests will fail, too.

matthew-carroll (approved):
LGTM - Thanks

PR #640, opened 2022-06-23
Change SuperDesktopTextField to always use text cursor style (Resolves #383)

### LINKED ISSUE #383: Add cursor style support for SuperTextField

angelosilvestre:
Change SuperDesktopTextField to always use text cursor style. Resolves https://github.com/superlistapp/super_editor/issues/383

I removed both `ValueNotifier` and `ListenableBuilder` because the cursor style won't be changing anymore.

Tests were added to ensure the cursor style is `SystemMouseCursors.text` when hovering the text, empty space and padding.

### DIFF
--- a/super_editor/test/super_textfield/super_desktop_texfield_mouse_interaction_test.dart
+++ b/super_editor/test/super_textfield/super_desktop_texfield_mouse_interaction_test.dart
@@ -1,0 +1,104 @@
+import 'package:flutter/gestures.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter/rendering.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:super_editor/super_editor.dart';
+import 'package:super_text_layout/super_text_layout.dart';
+
+import '../test_tools.dart';
+
+void main() {
+  group('SuperDesktopTextField', () {
+    testWidgetsOnDesktop('has text cursor style while hovering text', (tester) async {
+      final gesture = await _pumpGestureTestApp(tester);
+
+      // Ensure the cursor type is 'basic' when not hovering SuperDesktopTextField
+      expect(RendererBinding.instance.mouseTracker.debugDeviceActiveCursor(1), SystemMouseCursors.basic);
+      
+      // Hover the text inside SuperDesktopTextField
+      await gesture.moveTo(tester.getCenter(find.byType(SuperText)));
+      await tester.pump();
+
+      // Ensure the cursor type is 'text' when hovering the text
+      expect(RendererBinding.instance.mouseTracker.debugDeviceActiveCursor(1), SystemMouseCursors.text);
+
+      // Move outside SuperDesktopTextField bounds
+      await gesture.moveTo(Offset.zero);
+
+      // Ensure the cursor type is 'basic' again
+      expect(RendererBinding.instance.mouseTracker.debugDeviceActiveCursor(1), SystemMouseCursors.basic);
+    });
+
+    testWidgetsOnDesktop('has text cursor style while hovering empty space', (tester) async {
+      final gesture = await _pumpGestureTestApp(tester);
+
+      // Ensure the cursor type is 'basic' when not hovering SuperDesktopTextField
+      expect(RendererBinding.instance.mouseTracker.debugDeviceActiveCursor(1), SystemMouseCursors.basic);
+      
+      // Hover the empty space within SuperDesktopTextField        
+      await gesture.moveTo(tester.getTopRight(find.byType(SuperText)) + const Offset(10, 0));
+      await tester.pump();
+
+      // Ensure the cursor type is 'text' when hovering the empty space
+      expect(RendererBinding.instance.mouseTracker.debugDeviceActiveCursor(1), SystemMouseCursors.text);
+
+      // Move outside SuperDesktopTextField bounds
+      await gesture.moveTo(Offset.zero);
+
+      // Ensure the cursor type is 'basic' again
+      expect(RendererBinding.instance.mouseTracker.debugDeviceActiveCursor(1), SystemMouseCursors.basic);
+    });
+
+    testWidgetsOnDesktop('has text cursor style while hovering padding region', (tester) async {
+      final gesture = await _pumpGestureTestApp(tester, padding: 20.0);
+
+      // Ensure the cursor type is 'basic' when not hovering SuperDesktopTextField
+      expect(RendererBinding.instance.mouseTracker.debugDeviceActiveCursor(1), SystemMouseCursors.basic);
+      
+      // Hover the padding within SuperDesktopTextField
+      await gesture.moveTo(tester.getTopLeft(find.byType(SuperDesktopTextField)) + const Offset(10, 10));
+      await tester.pump();
+
+      // Ensure the cursor type is 'text' when hovering the padding
+      expect(RendererBinding.instance.mouseTracker.debugDeviceActiveCursor(1), SystemMouseCursors.text);
+
+      // Move outside SuperDesktopTextField bounds
+      await gesture.moveTo(Offset.zero);
+
+      // Ensure the cursor type is 'basic' again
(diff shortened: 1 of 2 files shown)


super_editor/test/super_textfield/super_desktop_texfield_mouse_interaction_test.dart
```diff
@@ -0,0 +1,104 @@
...
+import 'package:flutter_test/flutter_test.dart';
+import 'package:super_editor/super_editor.dart';
+import 'package:super_text_layout/super_text_layout.dart';
+
+import '../test_tools.dart';
+
+void main() {
+  group('SuperDesktopTextField', () {
+    testWidgetsOnDesktop('has text cursor style while hovering text', (tester) async {
+      final gesture = await _pumpGestureTestApp(tester);
```
matthew-carroll, line 13: I don't think the gesture should be started within this method. I suggest you pump the UI and then start the gesture in the test.

In other tests, there are methods that pump a field and tap within it. That's done because the behavior is used many times, and the tap is very simple, and it's completed within the method. In this case, you're starting a custom gesture, which is complex, you're not completing it within the pump call, and there's only a few tests that needs to do this.

super_editor/test/super_textfield/super_desktop_texfield_mouse_interaction_test.dart
```diff
@@ -0,0 +1,104 @@
...
+
+      // Ensure the cursor type is 'basic' when not hovering SuperDesktopTextField
+      expect(RendererBinding.instance.mouseTracker.debugDeviceActiveCursor(1), SystemMouseCursors.basic);
+      
+      // Hover the text inside SuperDesktopTextField
```
matthew-carroll, line 18: In a test that cares a lot about the content under the cursor, please indicate in your comments whether the new position sits over padding, or empty content, or text.

super_editor/test/super_textfield/super_desktop_texfield_mouse_interaction_test.dart
```diff
@@ -0,0 +1,104 @@
...
+      await gesture.moveTo(tester.getCenter(find.byType(SuperText)));
+      await tester.pump();
+
+      // Ensure the cursor type is 'text' when hovering the text
```
matthew-carroll, line 22: The cursor isn't hovering over text, is it? The field only contains "abc", so this offset is probably an empty content area, right?

super_editor/test/super_textfield/super_desktop_texfield_mouse_interaction_test.dart
```diff
@@ -0,0 +1,104 @@
...
+      expect(RendererBinding.instance.mouseTracker.debugDeviceActiveCursor(1), SystemMouseCursors.text);
+
+      // Move outside SuperDesktopTextField bounds
+      await gesture.moveTo(Offset.zero);
+
+      // Ensure the cursor type is 'basic' again
+      expect(RendererBinding.instance.mouseTracker.debugDeviceActiveCursor(1), SystemMouseCursors.basic);
+    });
+
+    testWidgetsOnDesktop('has text cursor style while hovering empty space', (tester) async {
```
matthew-carroll, line 32: "hovering empty space" -> "hovering over empty space"

super_editor/test/super_textfield/super_desktop_texfield_mouse_interaction_test.dart
```diff
@@ -0,0 +1,104 @@
...
+      final gesture = await _pumpGestureTestApp(tester);
+
+      // Ensure the cursor type is 'basic' when not hovering SuperDesktopTextField
+      expect(RendererBinding.instance.mouseTracker.debugDeviceActiveCursor(1), SystemMouseCursors.basic);
+      
+      // Hover the empty space within SuperDesktopTextField        
+      await gesture.moveTo(tester.getTopRight(find.byType(SuperText)) + const Offset(10, 0));
```
matthew-carroll, line 39: Does `SuperTextFieldInspector` have a method to get the rect for a position? If so, can you use that instead of hardcoding the offset?

angelosilvestre: I added this offset to hover over the empty content after the text

super_editor/test/super_textfield/super_desktop_texfield_mouse_interaction_test.dart
```diff
@@ -0,0 +1,114 @@
...
+      final gesture = await tester.createGesture(kind: PointerDeviceKind.mouse);
+      await gesture.addPointer(location: Offset.zero);
+      addTearDown(gesture.removePointer);      
+      await tester.pump();
+
+      // Ensure the cursor type is 'basic' when not hovering SuperDesktopTextField
+      expect(RendererBinding.instance.mouseTracker.debugDeviceActiveCursor(1), SystemMouseCursors.basic);
+      
+      // Hover over the empty space within SuperDesktopTextField      
+      await gesture.moveTo(tester.getTopRight(find.byType(SuperText)) + const Offset(10, 0));
```
matthew-carroll, line 51: I think this offset would be more clear if it was near the right side of the field and further down on the `y`. That way, it would be obvious that you're targeting an area that doesn't have text, and isn't padding.

super_editor/test/super_textfield/super_desktop_texfield_mouse_interaction_test.dart
```diff
@@ -0,0 +1,114 @@
...
+/// Creates a test app with the given [padding] applied to [SuperDesktopTextField]
+Future<void> _pumpGestureTestApp(WidgetTester tester, {
+  double padding = 0.0
+}) async {
+  await tester.pumpWidget(
+    MaterialApp(          
+      home: Scaffold(
+        body: Container(
+          width: 300,
+          padding: const EdgeInsets.all(20.0),
```
matthew-carroll, line 101: This looks like it shouldn't be hard coded. And this also reinforces my point above about the empty space test. I think your empty space test is currently testing padding, due to this hard-coded padding.

super_editor/test/super_textfield/super_desktop_texfield_mouse_interaction_test.dart
```diff
@@ -0,0 +1,114 @@
...
+      final gesture = await tester.createGesture(kind: PointerDeviceKind.mouse);
+      await gesture.addPointer(location: Offset.zero);
+      addTearDown(gesture.removePointer);      
+      await tester.pump();
+
+      // Ensure the cursor type is 'basic' when not hovering SuperDesktopTextField
+      expect(RendererBinding.instance.mouseTracker.debugDeviceActiveCursor(1), SystemMouseCursors.basic);
+      
+      // Hover over the text inside SuperDesktopTextField
+      await gesture.moveTo(tester.getTopLeft(find.byType(SuperText)));
```
matthew-carroll, line 25: Can you add a TODO to add the ability to the inspector to lookup an offset for a content position?

super_editor/test/super_textfield/super_desktop_texfield_mouse_interaction_test.dart
angelosilvestre: This padding is to ensure we can move the mouse outside `SuperDesktopTextField` bounds. In the empty space test we hover inside `SuperDesktopTextField` bounds, so the `Container` padding shouldn't affect this test.

matthew-carroll (approved):
LGTM - Thanks

PR #643, opened 2022-06-25
Fix paragraph alignment (Resolves #637)

### LINKED ISSUE #637: Paragraph alignment demo is broken
I discovered that the paragraph alignment demo is broken. The top 3 paragraphs, which should display left, center, and right aligned, respectively, are all scrunched to the left.

My guess is that paragraph alignment broke with the introduction of the new `SuperText` layout system.

Fix `super_editor` so that we get alignment back. Then, add a golden test that recreates the paragraph alignment demo.

angelosilvestre:
Fix paragraph alignment. Resolves https://github.com/superlistapp/super_editor/issues/637

The issue was that we were not passing `TextAlign` down to `LayoutAwareRichText`.

I added a golden test with the same document found in `ParagraphsDemo`.


matthew-carroll (approved):
LGTM - Thanks

Issue #630, opened 2022-06-20
SuperEditor: Multi-user selection


matthew-carroll:
This is a meta ticket for multi-user selection:

 * #631 
 * #404 
 * #424

PR #645, opened 2022-06-27
Separate gestures from scrolling and render the caret in an overlay (Resolves #149, Resolves Resolves #404)

### LINKED ISSUE #149: Move doc gesture detector beneath scrollview
Currently the document interactor is structured as follows:

RawGestureDetector
  SingleChildScrollView
    DocumentDisplay

As a result, we can't place an interactive overlay on top of the document display itself because all gestures will be consumed by the `RawGestureDetector` before they get down to the `DocumentDisplay`.

For example, imagine that we want to display a text format bar anchored to some position within the `DocumentDisplay`. We can display the bar, but the user can't tap on the bar because all gestures are intercepted.

The workaround is to use the global overlay and convert local to global coords.

Instead, we should look at re-organizing the structure as follows:

SingleChildScrollView
  DocumentOverlay
    RawGestureDetector
      DocumentDisplay

This way we can anchor widgets in document coordinate space and still have them receive gestures.

This also comes with the added benefit that the gesture detector no longer needs to map coordinates into document space because it already exists in document space.

However, I'm not sure what will happen as the user drags. Will we be able to scroll the document during a pan gesture?

### LINKED ISSUE #404: Change desktop/mouse caret to be global
When super_editor uses a mouse for input, each component is responsible for drawing the caret. For example, a `ParagraphComponent` draws the caret within the text. This means that every component needs to know how to draw a caret, which provides granular control, but it's unclear that any component needs this control in a document layout.

When super_editor uses gesture input, the caret is drawn globally in an overlay because gesture input also needs drag handles. This gesture input UI suggests that a global caret painter should work fine.

Refactor the desktop/mouse document configuration to draw the caret globally in an overlay.

matthew-carroll:
**Separate gestures from scrolling and render the caret in an overlay (Resolves #149, Resolves Resolves #404)**

We eventually want to support easily customizable multi-user carets, toolbars that sit in document space, and other decorations that are anchored in document space. To achieve these goals, we first want to add a "document overlay", which is a space that sits on top of a document and allows any other widgets to be displayed.

The goal of a document overlay was impossible with the current `SuperEditor` widget tree structure.

Historically, the scrolling system and the gesture handlers for desktop `SuperEditor`s were implemented within the same widget. For example, the widget tree looked like this:

DesktopMouseInteractor
\> RawGestureDetector 
\> SingleChildScrollView
DocumentLayout

Using the existing widget tree structure, if we added an overlay on top of the document display, that overlay wouldn't receive any gestures, because it would sit beneath the `RawGestureDetector`. On the other hand, if we placed the overlay above the `RawGestureDetector`, then the overlay would sit outside the `Scrollable` and therefore it wouldn't be in document space, it would be in viewport space. So this structure wasn't going to work.

This PR alters the structure to the following:

DocumentScrollable
DocumentOverlays
\> DefaultCaretOverlay
DocumentLayout
DesktopMouseInteractor

Notice a few things:
 * The gesture system and the scrolling system have been separated into different widets
 * The scrolling system was moved to the top of the stack so that everything else operates in document space
 * The gesture system was moved **beneath** the document. I added `IgnorePointer`s to all the components, which lets the gestures go through the `DocumentLayout` and into the `DocumentMouseInteractor`. This is a useful arrangement because now things like `Checkbox`s can handle their own gestures. Previously the `DocumentMouseInteractor` intercepted all gestures, including interactions with widgets that were embedded in the document.

With this re-arrangement, you'll notice that we've added a list of "document overlays". By default, there is a single overlay added to that list, and that overlay paints the default caret.

Completely custom caret painting can be achieved by replacing the default caret overlay with your custom caret overlay.

**Future work:**
 * Create a multi-caret overlay that ships with super_editor for ease-of-use (#630)
 * Refactor `SuperEditor` on mobile to use the same structure as desktop (#647)


super_editor/example/lib/demos/example_editor/example_editor.dart
```diff
@@ -335,6 +336,16 @@ class _ExampleEditorState extends State<ExampleEditor> {
         focusNode: _editorFocusNode,
         scrollController: _scrollController,
         documentLayoutKey: _docLayoutKey,
+        documentOverlayBuilders: [
+          DefaultCaretOverlayBuilder(
+            CaretStyle().copyWith(color: _isLight ? Colors.black : Colors.redAccent),
```
miguelcmedeiros, line 341: While testing out the changes on our project, I noticed that `CaretStyle` is not exported by the package.

super_editor/lib/src/default_editor/document_caret_overlay.dart
```diff
@@ -0,0 +1,134 @@
...
+    _blinkController = BlinkController(tickerProvider: this)..startBlinking();
+  }
+
+  @override
+  void didUpdateWidget(CaretDocumentOverlay oldWidget) {
+    super.didUpdateWidget(oldWidget);
+
+    if (widget.composer != oldWidget.composer) {
+      oldWidget.composer.removeListener(_onSelectionChange);
+      widget.composer.addListener(_onSelectionChange);
```
miguelcmedeiros, line 47: Shouldn't we remove/re-add a listener to `selectionNotifier`?
```suggestion
      oldWidget.composer.selectionNotifier.removeListener(_onSelectionChange);
      widget.composer.selectionNotifier.addListener(_onSelectionChange);
```

miguelcmedeiros (commented):
Code looks good, but didn't have a chance to test it out with our project. 
Would like to check that the auto-scroll works as expected and that it does not clash with another implementation for auto-scrolling but in this case for drag&drop.

matthew-carroll:
@miguelcmedeiros I pushed updates for your comments. I'd like to merge this as soon as possible so that some other PRs can move forward. Please let me know as soon as you've had a chance to test against Superlist.

roughike (commented):
When testing this PR in Superlist, I could not edit any of our documents.

By default, we don't autofocus when an existing document opens, so there's no way for the user to click somewhere on the document to gain focus to begin editing.

In this PR, there are two places (`BoxComponent` and `TextComponent`) that are now wrapped with `IgnorePointer` - this could be related?

matthew-carroll:
@roughike can you provide a repro sample for that behavior in the example app? I opened a few of the demos where focus can move in and out of the document. I'm not having any issue giving focus to the document.

FYI - after the rebase, there's a problem with the selection alignment over the document. I'll fix this sometime soon. But let's try to get to the bottom of your issue first.

roughike:
@matthew-carroll I took a deeper look, and it seems that the "bug that makes the document read-only at all times" was related to some API change in how document components are made. Copy-pasting SuperEditor example code with default component builders fixes it, so it's most likely not a SuperEditor issue, but something that has changed in the past SuperEditor versions.

The changes look good.

roughike (approved):

PR #652, opened 2022-06-28
SuperTextField caret is visible when there's no selection (Resolves #649)

### LINKED ISSUE #649: [SuperTextField] - on mobile the caret is always painted
`SuperTextField` on mobile always shows a caret, even when the field doesn't have focus. For example, every `SuperTextField` in a form all show a blinking caret at the same time.

I think this bug was probably caused by recent reworking of `super_text_layout`.

matthew-carroll:
SuperTextField caret is visible when there's no selection (Resolves #649)

I noticed on my webpage that all my text fields showed a blinking cursor. I think this bug came in with the recent `super_text_layout` refactor.

I adjusted the conditions under which a caret is displayed, and I added tests to make sure this doesn't happen again.


matthew-carroll:
CC @angelosilvestre because you just worked on similar behavior.

Also, it looks like my auto-formatter removed a number of whitespaces in these files. Do you know if your editor is adding those whitespaces?

miguelcmedeiros (approved):
LGTM

PR #655, opened 2022-06-29
Change SuperTextField to place caret at the end of the text when tapping empty space (Resolves #654)

### LINKED ISSUE #654: [SuperTextField] - Tapping empty space doesn't place the caret
I'm displaying a single-line `SuperTextField` and a multi-line `SuperTextField` in a mobile browser.

When I tap part of the field without any text, the caret is initially inserted at the end of the text, as expected.

However, if I place the caret elsewhere in the text, and then tap in the empty space, the caret doesn't move to the end of the text.

I expect the caret to move to the end of the text whenever I tap in empty space in the field beyond the end of the text.

angelosilvestre:
Change SuperTextField to place caret at the end of the text when tapping empty space. Resolves https://github.com/superlistapp/super_editor/issues/654

Tapping a part of the field without any text was changing selection only when `SuperTextField` had no valid selection. 

It was working when there was no selection because we are forcing a selection when `SuperAndroidTextField` is focused:

https://github.com/superlistapp/super_editor/blob/6602da8767efbd08707f18768bccb4c5bf070280/super_editor/lib/src/infrastructure/super_textfield/android/android_textfield.dart#L264-L272

This is also true for `SuperIOSTextField`

### DIFF
--- a/super_editor/lib/src/infrastructure/super_textfield/android/_user_interaction.dart
+++ b/super_editor/lib/src/infrastructure/super_textfield/android/_user_interaction.dart
@@ -149,15 +149,14 @@ class AndroidTextFieldTouchInteractorState extends State<AndroidTextFieldTouchIn
 
     final tapTextPosition = _getTextPositionAtOffset(details.localPosition);
     if (tapTextPosition == null) {
-      _log.warning('received a tap-down event on AndroidTextFieldInteractor that is not on top of any text');
+      // This situation indicates the user tapped in empty space
       widget.textController.selection = TextSelection.collapsed(offset: widget.textController.text.text.length);
-      return;
+    } else {
+      // Update the text selection to a collapsed selection where the user tapped.
+      widget.textController.selection = tapTextPosition.offset >= 0
+          ? TextSelection.collapsed(offset: tapTextPosition.offset)
+          : const TextSelection.collapsed(offset: 0);
     }
-
-    // Update the text selection to a collapsed selection where the user tapped.
-    widget.textController.selection = tapTextPosition.offset >= 0
-        ? TextSelection.collapsed(offset: tapTextPosition.offset)
-        : const TextSelection.collapsed(offset: 0);
 
     widget.editingOverlayController.showHandles();
   }
--- a/super_editor/lib/src/infrastructure/super_textfield/android/_user_interaction.dart
+++ b/super_editor/lib/src/infrastructure/super_textfield/android/_user_interaction.dart
@@ -149,9 +149,8 @@ class AndroidTextFieldTouchInteractorState extends State<AndroidTextFieldTouchIn
 
     final tapTextPosition = _getTextPositionAtOffset(details.localPosition);
     if (tapTextPosition == null) {
-      // This shouldn't be possible, but we'll ignore the tap if we can't
-      // map it to a position within the text.
       _log.warning('received a tap-down event on AndroidTextFieldInteractor that is not on top of any text');
+      widget.textController.selection = TextSelection.collapsed(offset: widget.textController.text.text.length);
       return;
     }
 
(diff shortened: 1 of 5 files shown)


super_editor/lib/src/infrastructure/super_textfield/android/_user_interaction.dart
```diff
@@ -149,9 +149,8 @@ class AndroidTextFieldTouchInteractorState extends State<AndroidTextFieldTouchIn
 
     final tapTextPosition = _getTextPositionAtOffset(details.localPosition);
     if (tapTextPosition == null) {
-      // This shouldn't be possible, but we'll ignore the tap if we can't
-      // map it to a position within the text.
       _log.warning('received a tap-down event on AndroidTextFieldInteractor that is not on top of any text');
+      widget.textController.selection = TextSelection.collapsed(offset: widget.textController.text.text.length);
```
matthew-carroll, line 153: Now that this is a legitimate behavior, we shouldn't log a warning. Also, we should add a comment explaining that this situation indicates the user tapped in empty space in the field.

super_editor/test/super_textfield/super_textfield_touch_interaction_test.dart
```diff
@@ -0,0 +1,96 @@
+import 'package:attributed_text/attributed_text.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:super_editor/src/infrastructure/super_textfield/super_textfield.dart';
+
+import '../test_tools.dart';
+import 'super_textfield_inspector.dart';
+
+void main() {
+  group('SuperTextField touch interaction', () {
```
matthew-carroll, line 10: Let's call the file `super_textfield_gestures_test.dart` and let's make the group "SuperTextField gestures"

super_editor/test/super_textfield/super_textfield_touch_interaction_test.dart
```diff
@@ -0,0 +1,96 @@
...
+    group('tapping in an empty space selects the end of the text', () {
+      testWidgetsOnMobile("when having no selection", (tester) async {
```
matthew-carroll, line 12: group: "tapping in empty space places the caret at the end of the text"
test: "when the field does not have focus"

super_editor/test/super_textfield/super_textfield_touch_interaction_test.dart
```diff
@@ -0,0 +1,96 @@
...
+        await tester.pumpAndSettle();
+
+        // Ensure selection is at the end of the text
+        expect(
+          SuperTextFieldInspector.findSelection(),
+          const TextSelection.collapsed(offset: 3),
+        );
+      });
+
+      testWidgetsOnMobile("when having a valid selection", (tester) async {
```
matthew-carroll, line 26: test: "when the field already has focus"

super_editor/test/super_textfield/super_textfield_touch_interaction_test.dart
```diff
@@ -0,0 +1,96 @@
...
+        // Ensure selection is at the end of the text
+        expect(
+          SuperTextFieldInspector.findSelection(),
+          const TextSelection.collapsed(offset: 3),
+        );
+      });
+    });
+
+    group('tapping in an area containing text moves selection to tap position', () {
+      testWidgetsOnMobile("when having no selection", (tester) async {
```
matthew-carroll, line 45: same test name updates here, too

super_editor/test/super_textfield/super_textfield_touch_interaction_test.dart
angelosilvestre: In this test the field doesn't start focused, I just set the initial selection in `AttributedTextEditingController`. Should I change to request focus to the field?

matthew-carroll: Yeah, I think what we probably want to test here is what happens when there's already a caret in the text, and that caret isn't at the end. Rather than set a selection and request focus, I think you can use the robot to place the caret and then tap the empty space.

angelosilvestre: The robot doesn't suppor mobile right now

matthew-carroll (approved):
LGTM - thanks

PR #658, opened 2022-06-30
[SuperTextField] - changed mobile tap gesture to only give focus on UP instead of DOWN (Resolves #657)

### LINKED ISSUE #657: [SuperTextField] - Change tap selection behavior on mobile
Currently, all `SuperTextField`s receive focus, and place a caret, the moment the user presses down on them.

This behavior is reasonable on desktop because pressing down with the mouse isn't used for scrolling. Therefore, a user pressing down on a text field is a good indication that the user intends to interact with the field.

On mobile, this behavior is a problem. If a `SuperTextField` sits in a `Scrollable`, then as the user scrolls down the page, it's easy to inadvertently give focus to a field, and cause the keyboard to pop up.

On mobile, we want to require a full tap to give focus and a selection to a `SuperTextField`. However, if the field already has focus and a selection, we want "tap downs" to immediately move the selection, because I think that's the typical mobile text field behavior, in general.

matthew-carroll:
[SuperTextField] - changed mobile tap gesture to only give focus on UP instead of DOWN (Resolves #657)

### DIFF
--- a/super_editor/lib/src/infrastructure/super_textfield/android/_user_interaction.dart
+++ b/super_editor/lib/src/infrastructure/super_textfield/android/_user_interaction.dart
@@ -98,8 +98,6 @@ class AndroidTextFieldTouchInteractorState extends State<AndroidTextFieldTouchIn
     with TickerProviderStateMixin {
   final _textViewportOffsetLink = LayerLink();
 
-  TextSelection? _selectionBeforeSingleTapDown;
-
   // Whether the user is dragging a collapsed selection.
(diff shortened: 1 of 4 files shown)


matthew-carroll:
CC @venkatd based on our recent discussion of text field tap behavior

CC @angelosilvestre so that you're aware of this change in behavior

matthew-carroll:
I'm going to merge this without review because the change is relatively minor and I'd like to confirm the fix in production.

PR #664, opened 2022-06-30
Make SuperTextField handle DONE action (Resolves #661)

### LINKED ISSUE #661: [SuperTextField] - on iOS web, pressing "DONE" on the keyboard doesn't do anything
When entering text in a `SuperTextField` on iOS web, at the top of the keyboard is a button that reads "DONE".

Pressing "DONE" has no effect.

I would expect pressing "DONE" to close the keyboard and remove focus from the field in question.

angelosilvestre:
Make SuperTextField handle DONE action. Resolves https://github.com/superlistapp/super_editor/issues/661

Here I'm not sure if we should unfocus even if the widget has a `onPerformActionPressed` defined. I could change this if needed. 

```dart
  void _onPerformActionPressed(TextInputAction action) {
    if (action == TextInputAction.done) {
      _focusNode.unfocus();
    }
    widget.onPerformActionPressed?.call(action);
  }
```

Maybe we should expose `onPerformActionPressed` on `SuperTextField`'s public API to write a test to ensure it's beeing called.

### DIFF
--- a/super_editor/lib/src/infrastructure/super_textfield/input_method_engine/_ime_text_editing_controller.dart
+++ b/super_editor/lib/src/infrastructure/super_textfield/input_method_engine/_ime_text_editing_controller.dart
@@ -194,6 +194,7 @@ class ImeAttributedTextEditingController extends AttributedTextEditingController
 
   void Function(TextInputAction)? _onPerformActionPressed;
   set onPerformActionPressed(Function(TextInputAction)? callback) => _onPerformActionPressed = callback;
+  Function(TextInputAction)? get onPerformActionPressed => _onPerformActionPressed;
 
   @override
   TextEditingValue? get currentTextEditingValue => TextEditingValue(
@@ -289,7 +290,11 @@ class ImeAttributedTextEditingController extends AttributedTextEditingController
 
   @override
   void performAction(TextInputAction action) {
-    _onPerformActionPressed?.call(action);
+    if (_realController is ImeAttributedTextEditingController && (_realController as ImeAttributedTextEditingController).onPerformActionPressed != null) {
+      (_realController as ImeAttributedTextEditingController).onPerformActionPressed!(action);
+    } else {
+      _onPerformActionPressed?.call(action);
+    }
   }
 
   @override
--- a/super_editor/lib/src/infrastructure/super_textfield/input_method_engine/_ime_text_editing_controller.dart
+++ b/super_editor/lib/src/infrastructure/super_textfield/input_method_engine/_ime_text_editing_controller.dart
@@ -290,11 +290,7 @@ class ImeAttributedTextEditingController extends AttributedTextEditingController
 
   @override
   void performAction(TextInputAction action) {
-    if (_realController is ImeAttributedTextEditingController && (_realController as ImeAttributedTextEditingController).onPerformActionPressed != null) {
-      (_realController as ImeAttributedTextEditingController).onPerformActionPressed!(action);
-    } else {
-      _onPerformActionPressed?.call(action);
-    }
+    _onPerformActionPressed?.call(action);
   }
 
   @override
--- a/super_editor/lib/src/infrastructure/super_textfield/ios/ios_textfield.dart
+++ b/super_editor/lib/src/infrastructure/super_textfield/ios/ios_textfield.dart
@@ -41,7 +41,6 @@ class SuperIOSTextField extends StatefulWidget {
     this.textInputAction = TextInputAction.done,
     this.popoverToolbarBuilder = _defaultPopoverToolbarBuilder,
     this.showDebugPaint = false,
-    this.onPerformActionPressed,
   })  : assert(minLines == null || minLines == 1 || lineHeight != null, 'minLines > 1 requires a non-null lineHeight'),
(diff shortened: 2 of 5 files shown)


matthew-carroll:
I looked through some of the existing API structure related to this and I think we want to make a few changes in this PR.

1. I think we should remove the `performAction` property in the Android and iOS text fields. Instead, the client should pass a controller that's configured with a perform action callback.
2. In the mobile text fields, I think we only want to take an action if the controller doesn't have a callback set on it.
3. When we handle the action, I think we need to implement all the reasonable defaults. Take a look at Flutter's enum and implement any action where we can reasonably do the right thing: https://api.flutter.dev/flutter/services/TextInputAction.html
4. Add a test for every action that we support.
5. Add a test that ensures the default action only happens when the controller has a `null` callback, and that a non-`null` callback is called.

In the future we'll figure out what we need to expose in `SuperTextField`, but it's not clear to me what that should be. In fact, we'll probably need to take another look at the concept of `ImeAttributedTextController`. I'm not sure it's the right tool to begin with.

angelosilvestre:
Should we handle `go`, `search` and `send`  unfocusing the field like in `done` ?

I've made a change that is unrelated to the original issue:

```dat
bool get _isMultiline => widget.minLines != 1 || widget.maxLines != 1;
```
`widget.minLines` accepts null, so without setting `minLines` the field was beeing considered multiline

 `widget.maxLines` also accepts null, but it defaults to 1 in the constructor

super_editor/lib/src/infrastructure/super_textfield/input_method_engine/_ime_text_editing_controller.dart
```diff
@@ -289,7 +290,11 @@ class ImeAttributedTextEditingController extends AttributedTextEditingController
 
   @override
   void performAction(TextInputAction action) {
-    _onPerformActionPressed?.call(action);
+    if (_realController is ImeAttributedTextEditingController && (_realController as ImeAttributedTextEditingController).onPerformActionPressed != null) {
```
matthew-carroll, line 293: Do we need this if-statement? If our inner controller is also an `ImeAttributedTextEditingController` then this same situation could happen inside that controller, too. I think we're better off assuming that the client controller isn't an `ImeAttributedTextEditingController`.

super_editor/test/super_textfield/super_textfield_input_actions_test.dart
```diff
@@ -0,0 +1,250 @@
...
+
+      // Simulate a tap at the action button
+      await tester.testTextInput.receiveAction(TextInputAction.previous);
+      await tester.pump();
+
+      // Ensure focus has moved to next field
+      expect(focusNodeFirstField.hasFocus, true);
+    });
+
+    group('having a textController with onPerformActionPressed set', () {
```
matthew-carroll, line 83: "having a textController with onPerformActionPressed set" -> "with custom onPerformActionPressed callback"

super_editor/lib/src/infrastructure/super_textfield/ios/ios_textfield.dart
```diff
@@ -356,6 +356,22 @@ class SuperIOSTextFieldState extends State<SuperIOSTextField>
...
+      case TextInputAction.done:
+        _focusNode.unfocus();
+        break;
+      case TextInputAction.next:
+        _focusNode.nextFocus();
+        break;
+      case TextInputAction.previous:
+        _focusNode.previousFocus();
+        break;
+      default:
```
matthew-carroll, line 371: For your question about the remaining actions: for now, let's log a warning that the user pressed an action button that doesn't have a default behavior

super_editor/lib/src/infrastructure/super_textfield/input_method_engine/_ime_text_editing_controller.dart
angelosilvestre: `onPerformActionPressed` is defined only in `ImeAttributedTextEditingController` and `_realController` is defined as `AttributedTextEditingController`.

matthew-carroll: Right, but this class with `performAction()` **IS** an `ImeAttributedTextEditingController`, right? So this class has the callback that we want to call, just like the original implementation. Am I missing something?

angelosilvestre: When we pass an `textController` to `SuperTextField` it wraps it into another `ImeAttributedTextEditingController`: 

https://github.com/superlistapp/super_editor/blob/64e029e11660f3d5901d27344eba1105850c6bb4/super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart#L154-L161

So, to call the original callback we need this.

matthew-carroll: Ok, let's change the text fields to only wrap the given controller if the given controller is not an `ImeAttributedTextEditingController`. Let me know if that creates any problems.

matthew-carroll (approved):
LGTM - Thanks.

PR #665, opened 2022-06-30
[SuperTextField] Fix space and arrow keys on web (Resolves #659)

### LINKED ISSUE #659: [SuperTextField] - Pressing space bar on software keyboard doesn't work
I noticed a bug when typing text into single-line and multi-line `SuperTextField`s on iOS in the browser that the space bar on the software keyboard doesn't work.

This problem might be unique to iOS web, but also maybe not.

When I'm entering text, and I press the space bar, instead of a space being inserted, my page's scrollview scrolls down.

Instead, when pressing space bar, a space should be inserted, and no scrolling should happen at all.

angelosilvestre:
Fix space and arrow keys on web. Resolves https://github.com/superlistapp/super_editor/issues/659

This was also an issue on Flutter `TextField` on web: https://github.com/flutter/flutter/issues/74191

Unfortunately, to be able to test this we need to avoid using `kIsWeb` and using some getter that we could override.

If we do that, then we could write a test adding `Actions` and `Shortcuts` widgets above `SuperTextField` and inspecting if the action callback was called.

### DIFF
--- a/super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart
+++ b/super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart
@@ -1,6 +1,7 @@
 import 'package:attributed_text/attributed_text.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
 import 'package:super_editor/src/infrastructure/attributed_text_styles.dart';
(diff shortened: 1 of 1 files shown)


matthew-carroll (approved):
LGTM - Thanks

matthew-carroll:
This change seemed necessary to block `Shortcuts` from doing the wrong thing, but it's a regrettable solution because we aren't using `Shortcuts`. I filed https://github.com/flutter/flutter/issues/107014 describing this situation.

PR #675, opened 2022-07-06
Force SuperTextField to show keyboard on tap (Resolves #673)

### LINKED ISSUE #673: [SuperTextField][mobile] - Field doesn't lose focus when keyboard is closed.
The following issue was found on Android web, but it probably applies to iOS web, too. It may also apply to native mobile, but I haven't tested.

On Android web, do the following:

1. tap on a `SuperTextField` to give it focus with a caret
2. the keyboard pops up
3. press the button on the keyboard that closes the keyboard

Expected result:
The `SuperTextField` caret disappears

Actual result:
The `SuperTextField` caret continues blinking

angelosilvestre:
Force `SuperTextField` to show keyboard on tap. Resolves https://github.com/superlistapp/super_editor/issues/673

After tapping the field and closing the keyboard, tapping at the field again was not opening the keyboard.

I modified it to call `show()` on `TextInputConnection` if it's already focused and attached to ime.

### DIFF
--- a/super_editor/lib/src/infrastructure/super_textfield/android/_user_interaction.dart
+++ b/super_editor/lib/src/infrastructure/super_textfield/android/_user_interaction.dart
@@ -67,7 +67,7 @@ class AndroidTextFieldTouchInteractor extends StatefulWidget {
   /// [TextController] used to read the current selection to display
   /// editing controls, and used to update the selection based on
   /// user interactions.
-  final AttributedTextEditingController textController;
+  final ImeAttributedTextEditingController textController;
 
   final AndroidEditingOverlayController editingOverlayController;
 
@@ -148,10 +148,8 @@ class AndroidTextFieldTouchInteractorState extends State<AndroidTextFieldTouchIn
   void _onTapUp(TapUpDetails details) {
     _log.fine('User released a tap');
 
-    if (widget.focusNode.hasFocus &&
-        (widget.textController is ImeAttributedTextEditingController) &&
-        (widget.textController as ImeAttributedTextEditingController).isAttachedToIme) {
-      (widget.textController as ImeAttributedTextEditingController).showKeyboard();
+    if (widget.focusNode.hasFocus && widget.textController.isAttachedToIme) {
+      widget.textController.showKeyboard();
     } else {
       widget.focusNode.requestFocus();
     }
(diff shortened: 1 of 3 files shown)


super_editor/lib/src/infrastructure/super_textfield/android/_user_interaction.dart
```diff
@@ -148,7 +148,13 @@ class AndroidTextFieldTouchInteractorState extends State<AndroidTextFieldTouchIn
   void _onTapUp(TapUpDetails details) {
     _log.fine('User released a tap');
 
-    widget.focusNode.requestFocus();
+    if (widget.focusNode.hasFocus &&
+        (widget.textController is ImeAttributedTextEditingController) &&
```
matthew-carroll, line 152: I think we can change the type of `textController` to be an `ImeAttributedTextEditingController`. Because Android and iOS always use IME, right? And this is specifically an Android interactor. Let me know if that change works out. Then we can avoid this type checking and casting.

super_editor/test/super_textfield/super_textfield_gestures_test.dart
```diff
@@ -160,6 +159,34 @@ void main() {
         // Ensure the caret moved to the beginning of the text.
         expect(SuperTextFieldInspector.findSelection()!.extent.offset, 0);
       });
+
+      testWidgetsOnMobile("tap up shows the keyboard if the field has focus", (tester) async {
```
matthew-carroll, line 163: Let's change "the field has focus" -> "the field already has focus" to make it clear that we're interested in a pre-existing condition.

super_editor/test/super_textfield/super_textfield_gestures_test.dart
```diff
@@ -160,6 +159,34 @@ void main() {
         // Ensure the caret moved to the beginning of the text.
         expect(SuperTextFieldInspector.findSelection()!.extent.offset, 0);
       });
+
+      testWidgetsOnMobile("tap up shows the keyboard if the field has focus", (tester) async {     
+        await _pumpTestApp(tester);
+
+        bool isShowKeyboarCalled = false;
```
matthew-carroll, line 166: Type: "isShowKeyboarCalled" -> "isShowKeyboardCalled"

matthew-carroll (approved):
LGTM - Thanks.

PR #676, opened 2022-07-06
Removed the need to specify a lineHeight for multiline SuperTextField's (Resolves #622)

### LINKED ISSUE #622: Don't require a line height for SuperTextField
text_scrollview.dart line 323 blows up if text style doesn't include line height. Make it so that this doesn't happen. Write a test that proves we can pump a text field without a line height.

matthew-carroll:
Removed the need to specify a lineHeight for multiline SuperTextField's (Resolves #622)

Historically, we've required that any multi-line `SuperTextField` provide an explicit line height in pixels. This is because styled text can produce varying line heights, and text scrolling currently scrolls by line, not by pixel. So how tall should a `SuperTextField` be if, for example, we have a 5 line field with different font sizes inside of it? Our answer was to make the developer choose a line height and then `SuperTextField` would multiply that value by the number of lines.

This is an annoying requirement. Even with an explicit line height, the result can be poor. The eventual answer is to offer pixel-based scrolling, when desired. But for now, this PR removes the need to pass a line height to `SuperTextField`. Instead, if a `lineHeight` isn't provided, `SuperTextField` will calculate a reasonable estimate internally.

### DIFF
--- a/super_editor/lib/src/infrastructure/super_textfield/android/android_textfield.dart
+++ b/super_editor/lib/src/infrastructure/super_textfield/android/android_textfield.dart
@@ -37,9 +37,7 @@ class SuperAndroidTextField extends StatefulWidget {
     this.textInputAction = TextInputAction.done,
     this.popoverToolbarBuilder = _defaultAndroidToolbarBuilder,
     this.showDebugPaint = false,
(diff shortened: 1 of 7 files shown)


angelosilvestre (approved):
LGTM

PR #677, opened 2022-07-07
Fixed getRangeBetween() in MutableDocument, and added tests for getDocumentSelectionInRegion in SingleColumnDocumentLayout (Resolves #48, Resolves #50)

### LINKED ISSUE #48: Fix discrepancy with Document#getRangeBetween()
The `Document#getRangeBetween()` method does not necessarily handle a range within a node correctly. It's possible that the returned range is reversed because the `Document` doesn't know how to interpret individual node positions.

Either, the implementation of this method needs to be adjusted to ensure correct range ordering, or the method needs to bail on single nodes, or the uses of this method need to be changed to solve their problems in a different way.

### LINKED ISSUE #50: Fix DefaultDocumentLayout#getDocumentSelectionInRegion
The `DefaultDocumentLayout#getDocumentSelectionInRegion()` method does not correctly handle the case where a selection region sits on the same line within a text node. Fix this.

matthew-carroll:
Fixed getRangeBetween() in MutableDocument, and added tests for getDocumentSelectionInRegion in SingleColumnDocumentLayout (Resolves #48, Resolves #50)

#50 referenced a problem that seemed to get solved somewhere along the way. I added a couple tests to ensure that the problem doesn't return.

Fixed #48 by adjusting the implementation to use an existing calculation for document affinity, which solved that problem.

### DIFF
--- a/super_editor/lib/src/core/document_editor.dart
+++ b/super_editor/lib/src/core/document_editor.dart
@@ -2,6 +2,8 @@ import 'dart:math';
 
 import 'package:collection/collection.dart';
 import 'package:flutter/foundation.dart';
+import 'package:flutter/material.dart';
+import 'package:super_editor/src/core/document_selection.dart';
 import 'package:super_editor/src/infrastructure/_logging.dart';
 import 'package:uuid/uuid.dart';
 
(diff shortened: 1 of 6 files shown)


angelosilvestre (approved):
LGTM

PR #678, opened 2022-07-07
[SuperTextField] Fix content not scrolling when selection changes (Resolves #510)

### LINKED ISSUE #510: [SuperTextField] [iOS] [Android] Typed text not visible during typing
# Steps to reproduce
Code:
```
  const MyApp({Key? key}) : super(key: key);

  @override
    return MaterialApp(
      home: Scaffold(
        body: Center(
          child: Container(
            color: Colors.black,
            child: const SuperTextField(
              lineHeight: 16,
            ),
          ),
        ),
      ),
    );
  }
}
```
- Run the app
- Tap black rectangle (the text field)
- Type more than 1 line of text

# Actual behavior
When user types a text longer than the widget's starting size, they won't be able to see what's being typed, because the text goes outside of the widget's boundaries.

https://user-images.githubusercontent.com/3801761/160955954-aeb8d2d7-2c37-4160-bcca-d17783e05a9d.mp4

# Expected behavior
While typing, current cursor position should always be visible so the user can see what's being typed.

# Platform
```
Flutter 2.10.3 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 7e9793dee1 (4 weeks ago) • 2022-03-02 11:23:12 -0600
Engine • revision bd539267b4
Tools • Dart 2.16.1 • DevTools 2.9.2
```

CC @venkatd

angelosilvestre:
Fix content not scrolling when selection changes. Resolves https://github.com/superlistapp/super_editor/issues/510

Typing or changing the selection wasn't causing `TextScrollController` to auto scroll. Because of this, the text could become invisible to the user

Also, `TextScrollController` was only handling vertical scrolling.


super_editor/test/super_textfield/super_texfield_autoscroll_test.dart
```diff
@@ -0,0 +1,97 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:super_editor/super_editor.dart';
+import 'package:super_text_layout/super_text_layout.dart';
+
+import '../test_tools.dart';
+
+void main() {
+  group('SuperTextField', () {
+    testWidgetsOnAllPlatforms('auto scrolls horizontally when single-line', (tester) async {
```
matthew-carroll, line 10: Is it possible to write these tests with drag gestures so that we test the full widget?

super_editor/test/super_textfield/super_texfield_autoscroll_test.dart
```diff
@@ -0,0 +1,97 @@
...
+      final viewportRight = tester.getBottomRight(find.byType(SuperTextField)).dx;
+
+      // Position at the end of the text
+      final textRight = tester.getBottomRight(find.byType(SuperText)).dx;
+
+      // Ensure the text field scrolled its content horizontally
+      expect(textRight, lessThanOrEqualTo(viewportRight));
+    });
+
+    testWidgetsOnAllPlatforms('auto scrolls vertically when multi-line', (tester) async {
```
matthew-carroll, line 39: Any place where the term "auto-scroll" is used with regard to this jumping behavior should be replaced. When we talk about auto-scrolling, we mean an animated scroll that happens over time. What we're doing here might be called "jumping to a scroll position" or "revealing the caret" or something like that. But we don't want to confuse it with the auto-scroll behavior.

super_editor/test/super_textfield/super_texfield_scroll_test.dart
```diff
@@ -0,0 +1,99 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:super_editor/super_editor.dart';
+import 'package:super_text_layout/super_text_layout.dart';
+
+import '../test_tools.dart';
+
+void main() {
+  group('SuperTextField', () {
+    testWidgetsOnAllPlatforms('single-line jumps scroll position horizontally when selection changes', (tester) async {
```
matthew-carroll, line 10: Let's adjust these names from "when selection changes" to "as the user types". I know we don't have the simulation for this right now, but since that's the use-case we're interested in, it's probably best to make that explicit. When we're talking about "selection", it makes me think about users dragging things.

matthew-carroll (approved):
LGTM - Pending one last comment.

matthew-carroll (approved):
LGTM - Thanks.

PR #681, opened 2022-07-07
Adding Melos for monorepo management

### LINKED ISSUE #680: Add Melos for monorepo management
[Melos][melos] is a command line tool for managing Dart monorepos. The primary benefits it would provide for this repo are:

1. Make development across projects easier. Currently some projects within the monorepo depend on the version in pub and others have `dependency_overrides` that have to be removed during publishing. Instead we could just have pub dependencies and Melos will transparently set overrides during local development.
2. Use local dependencies during CI runs. This is really just an extension of the previous point, but currently the inconsistent dependencies on pub vs local packages means checks on pull requests can't always be trusted. Melos can be used during the CI run to make sure tests run with code from the commit and not pub.dev.
3. Make publishing updates easier. Currently the varous `dependency_overrides` sections must be removed or `pub publish` will refuse to work. At time of writing this is step would be reduced to removing a single file, `pubspec_overrides.yaml`, from any packages to be published. Melos also has some convenience commands to do things like publish any packages that are newer than the version on pub, generate updated changelogs, etc. I personally find these a little fiddly to get working but they could be useful

Using melos would mean having developers install it with `flutter pub global activate melos` and then use it to install the projects' dependencies with `melos bootstrap`. CI tasks would be updated to do the same.

As an alternative, `pubspec_overrides.yaml` files could be added to each project instead and manually maintained. This mechanism for overriding dependencies during development was added in Dart 2.17 and is what Melos uses under the hood. Currently the presence of these files blocks pub publishing the same as using `dependency_overrides`, but future versions of pub should ignore the file entirely. This would lose some of the extra functionality provided by Melos but avoid a tool dependency while still solving the intra-repo dependency issue.

[melos]: https://melos.invertase.dev/

jmatth:
See #680 for details. Let me know if the update to the root readme should be moved or copied to any of the individual project readmes. I was on the fence about doing so but it seems potentially annoying to maintain setup instructions in each individual package as the repo grows. Maybe just add a line to each one saying to refer to the root readme to set up your development environment?

Closes #680.

### DIFF
--- a/super_editor/example/pubspec.yaml
+++ b/super_editor/example/pubspec.yaml
@@ -40,16 +40,21 @@ dependencies:
   logging: ^1.0.1
   uuid: ^3.0.3
 
-  # Specify paths for direct dependencies within the mono-repo.
+  super_editor: any
+  super_editor_markdown: any
+  super_text_layout: any
+
+dependency_overrides:
+  # Override dependencies within the monorepo. These need to go here otherwise
+  # version solving can break when an internal package depends on the pub-hosted
+  # version of another internal package, such as how super_editor_markdown
+  # depends on super_editor.
   super_editor:
     path: ../
   super_editor_markdown:
     path: ../../super_editor_markdown
   super_text_layout:
     path: ../../super_text_layout
-
-dependency_overrides:
-  # Override transitive dependencies within the mono-repo.
   attributed_text:
     path: ../../attributed_text
 
--- a/super_editor/example/pubspec.yaml
+++ b/super_editor/example/pubspec.yaml
@@ -40,6 +40,7 @@ dependencies:
   logging: ^1.0.1
   uuid: ^3.0.3
 
+  # Specify paths for direct dependencies within the mono-repo.
   super_editor:
     path: ../
   super_editor_markdown:
@@ -48,8 +49,7 @@ dependencies:
     path: ../../super_text_layout
 
 dependency_overrides:
-  # Override to local mono-repo path so devs can test this repo
-  # against changes that they're making to other mono-repo packages
+  # Override transitive dependencies within the mono-repo.
   attributed_text:
     path: ../../attributed_text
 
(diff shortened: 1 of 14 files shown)


super_editor/example/pubspec.yaml
```diff
@@ -40,26 +40,18 @@ dependencies:
...
   super_editor_markdown:
     path: ../../super_editor_markdown
   super_text_layout:
     path: ../../super_text_layout
 
+dependency_overrides:
+  # Override to local mono-repo path so devs can test this repo
+  # against changes that they're making to other mono-repo packages
+  attributed_text:
+    path: ../../attributed_text
```
matthew-carroll, line 54: Is there a reason that this one dependency is left under overrides, but all the others are listed above as regular deps?

jmatth: Done, lmk if you want me to revise the new comments to be more verbose

matthew-carroll:
My plan is to get this PR and the AttributedText PR merged after the next couple releases, because we're almost done with a couple deliverables.

BTW, it looks like Windows tests are failing in CI

jmatth:
> My plan is to get this PR and the AttributedText PR merged after the next couple releases, because we're almost done with a couple deliverables.

Cool, I'll keep the branches up to date with `main` in the meantime.

> BTW, it looks like Windows tests are failing in CI

Well this is annoying. Turns out there was a bug introduced in Dart `2.17.0` that causes concurrent `pub get` processes to fail (dart-lang/pub#3420). They merged a fix back in May but it still hasn't reached the Dart SDK stable channel. I'll investigate a workaround for now.

jmatth:
@matthew-carroll This is ready for another look if you're still interested in this change. #680 describes the reasoning and proposes an alternative if you don't want to add a dependency on an external tool.

matthew-carroll:
Yeah, I'd like to get this in, but can we start with the attribution range change?

matthew-carroll:
I'm going to close this because it's so out of date. We still want this. I'll probably refactor for Melos before the next major release. That will be a different PR.

PR #682, opened 2022-07-07
A couple of misc tickets (#52, #54)

### LINKED ISSUE #52: Remove highlightWhenEmpty from DocumentNodeSelection
`DocumentNodeSelection` includes a `highlightWhenEmpty` property, which is used by the ParagraphNode to decide whether or not to paint a highlight when empty.

The problem is that `DocumentNodeSelection` is a logical construct, not a visual construct. Also, `DocumentNodeSelection` applies to all types of nodes, not just `ParagraphNode`. 

Refactor such that `highlightWhenEmpty` is used only within visual classes, and only in situations where it's relevant.

### LINKED ISSUE #54: Change implementation of _TextComponentState#getContiguousTextAt()
getContiguousTextAt() assumes only a single paragraph in the component, but this isn't true when the user presses shift+enter. Update the implementation to actually calculate contiguous text.

matthew-carroll:
Moved `DocumentNodeSelection` into the layout subsystem because that's the only place it's used (Resolves #52)

Changed `getContiguousTextAt()` to have the implementation it should have originally had, and also added `getAllText()` to `TextComposable` because that's how some clients were using `getContiguousTextAt()` (Resolves #54)


angelosilvestre (approved):
LGTM

PR #685, opened 2022-07-07
Switch web key blacklist from list to algorithm (Resolves #193)

### LINKED ISSUE #193: [SuperTextField] Generic filtering of UI keys
We kept getting reports of keys that are not in the blacklist. For example, we recently had `AudioVolumeUp` reported. Someone turned up their volume while using the app (focused on a TextField) and it inserted this string literally into the editor.

It looks many come from here: https://www.w3.org/TR/uievents-key

Because it seems tough to capture them all, it may work in a more general-purpose way if we check for a camel-case alphanumeric string with length larger than 1.

@matthew-carroll 

cc @oxama I saw you were reporting these kinds of issues in the past. Would you mind confirming if the following properly filters keys for you?

My `keyboardActions` list has a `(kIsWeb) blockWebUIEventKeys,` to ensure it doesn't get included in the native apps.

```dart

TextFieldKeyboardHandlerResult blockWebUIEventKeys({
  required AttributedTextEditingController controller,
  SelectableTextState? selectableTextState,
  required RawKeyEvent keyEvent,
}) {
  if (keyEvent.character != null && _isWebUIEventKey(keyEvent.character!))
    return TextFieldKeyboardHandlerResult.blocked;
  return TextFieldKeyboardHandlerResult.notHandled;
}

// Limitation of Flutter web which returns chars that map to UI events
// Further reading: https://www.w3.org/TR/uievents-key
// The equivalent regex would look like [A-Z][A-Za-z0-9]+
bool _isWebUIEventKey(String chars) {
  const _kA = 0x41;
  const _kZ = 0x5a;
  const _ka = 0x61;
  const _kz = 0x7a;
  const _k0 = 0x30;
  const _k9 = 0x39;

  bool isUpperAlpha(int codeUnit) => codeUnit >= _kA && codeUnit <= _kZ;

  bool isAlphaNumeric(int codeUnit) {
    return codeUnit >= _kA && codeUnit <= _kZ ||
        codeUnit >= _ka && codeUnit <= _kz ||
        codeUnit >= _k0 && codeUnit <= _k9;
  }

  if (chars.length <= 1) return false;

  final codeUnits = chars.codeUnits;
  if (!isUpperAlpha(codeUnits[0])) return false;

  for (var i = 1; i < codeUnits.length; i++) {
    if (!isAlphaNumeric(codeUnits[i])) return false;
  }

  return true;
}
```

Some rough testing:
```dart
void main() {
  testKey('a'); # false
  testKey('ChannelDown'); # true
  testKey('234'); # false
  testKey('F1'); # true
  testKey('Dead'); # true
  testKey('D##'); # false
  testKey('áé'); # false
}

void testKey(String key) {
  print("testKey('$key'); # ${_isWebUIEventKey(key)}");
}
```

matthew-carroll:
Switch web key blacklist from list to algorithm (Resolves #193)

### DIFF
--- a/super_editor/lib/src/default_editor/document_keyboard_actions.dart
+++ b/super_editor/lib/src/default_editor/document_keyboard_actions.dart
@@ -156,7 +156,7 @@ ExecutionInstruction anyCharacterOrDestructiveKeyToDeleteSelection({
   final isDestructiveKey =
       keyEvent.logicalKey == LogicalKeyboardKey.backspace || keyEvent.logicalKey == LogicalKeyboardKey.delete;
   final isCharacterKey =
-      keyEvent.character != null && keyEvent.character != '' && !webBugBlacklistCharacters.contains(keyEvent.character);
+      keyEvent.character != null && keyEvent.character != '' && !isKeyEventCharacterBlacklisted(keyEvent.character);
 
   final shouldDeleteSelection = isDestructiveKey || isCharacterKey;
   if (!shouldDeleteSelection) {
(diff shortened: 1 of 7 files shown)


matthew-carroll:
CC @venkatd - here's an update that uses your recommended rule for blacklisting characters on web.

BTW, can you explain by "234", "D##", and "áé" should be legal characters?

angelosilvestre (approved):
LGTM

venkatd:
@matthew-carroll great!

I was being defensive when coming up with those cases, so we may not need to include those exact tests.

What I wanted to test for:
- Lowercase first character fails the filter
- Numeric first character fails the filter
- If there's anything non alphanumeric such as accents or puncutation, the filter would fail

The characters themselves were random but were meant to test the above logic.

PR #687, opened 2022-07-07
Made color of mobile handles configurable in SuperEditor (Resolves #425)

### LINKED ISSUE #425: Caret and selection drag handles are not customisable on iOS/Android
It should be possible to customise the theme of the caret and selection drag handles on mobile:
* Caret: color and width (similar to what it's possible to on desktop)
* Selection drag handles: color

Probably the color for both is the same, but I'm not sure if that's the case for Android 🤷‍♂️

matthew-carroll:
Made color of mobile handles configurable in SuperEditor (Resolves #425)


angelosilvestre (approved):
LGTM

PR #691, opened 2022-07-09
[SuperTextField] Auto-scroll when field is focused (Resolves #663)

### LINKED ISSUE #663: [SuperTextField] - Giving focus to a field doesn't auto-scroll
Typically, when a user focuses a text field, that text field should cause the page to auto-scroll so that the field sits somewhere around the middle of the viewport.

On mobile web, `SuperTextField` doesn't seem to auto-scroll at all.

Investigate how text field auto-scrolling usually works in Flutter. If appropriate, add the necessary behaviors to `SuperTextField` to also achieve auto-scrolling.

angelosilvestre:
Auto-scroll when field is focused. Resolves #663

Giving focus to a `SuperTextField` inside a `Scrollable` was not causing it to auto-scroll. Because of this, the text field could be hidden by the keyboard.

I used the same approach that `SuperEditor` uses to auto-scroll its contents.

### DIFF
--- a/super_editor/lib/src/infrastructure/super_textfield/android/android_textfield.dart
+++ b/super_editor/lib/src/infrastructure/super_textfield/android/android_textfield.dart
@@ -126,6 +126,9 @@ class SuperAndroidTextField extends StatefulWidget {
 }
 
 class SuperAndroidTextFieldState extends State<SuperAndroidTextField> with TickerProviderStateMixin, WidgetsBindingObserver implements ProseTextBlock {
+  static const Duration _autoScrollAnimationDuration = Duration(milliseconds: 100);
+  static const Curve _autoScrollAnimationCurve = Curves.fastOutSlowIn;
+
   final _textFieldKey = GlobalKey();
   final _textFieldLayerLink = LayerLink();
   final _textContentLayerLink = LayerLink();
@@ -145,9 +148,6 @@ class SuperAndroidTextFieldState extends State<SuperAndroidTextField> with Ticke
   // positions the invisible touch targets for base/extent
   // dragging.
   OverlayEntry? _controlsOverlayEntry;
-
-  static const Duration _autoScrollAnimationDuration = Duration(milliseconds: 100);
-  static const Curve _autoScrollAnimationCurve = Curves.fastOutSlowIn;
 
   @override
   void initState() {
@@ -268,7 +268,7 @@ class SuperAndroidTextFieldState extends State<SuperAndroidTextField> with Ticke
     // appearance/disappearance.
     WidgetsBinding.instance.addPostFrameCallback((timeStamp) {
       if (mounted && _focusNode.hasFocus) {
-        _ensureIsVisible();
+        _autoScrollToKeepTextFieldVisible();
       }
     });
   }
@@ -291,8 +291,8 @@ class SuperAndroidTextFieldState extends State<SuperAndroidTextField> with Ticke
             textInputAction: widget.textInputAction,
             textInputType: _isMultiline ? TextInputType.multiline : TextInputType.text,
           );
-          
-          _ensureIsVisible();
+
+          _autoScrollToKeepTextFieldVisible();
           _showEditingControlsOverlay();
         });
       }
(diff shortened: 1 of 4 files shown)


matthew-carroll:
Can you check how Flutter solves this with their regular `TextField`s? I'm guessing that they use a different approach, and if so, we should probably replicate theirs.

matthew-carroll:
@angelosilvestre did you want me to re-review this? Or are you still investigating?

angelosilvestre:
@matthew-carroll I already changed the implementation to be similar to flutter textfield, could you re-review it?

super_editor/lib/src/infrastructure/super_textfield/android/android_textfield.dart
```diff
@@ -379,6 +396,42 @@ class SuperAndroidTextFieldState extends State<SuperAndroidTextField>
...
+  /// is visible on the viewport when it's focused
+  void _ensureIsVisible() {
+    // If we are not inside a [Scrollable] we don't autoscroll
+    final ancestorScrollable = Scrollable.of(context);
+    if (ancestorScrollable == null) {
+      return;
+    }
+
+    // The [RenderBox] is needed so we can calculate the final [Offset]
+    final fieldBox = context.findRenderObject() as RenderBox?;
```
matthew-carroll, line 409: We control the field box, so we know it's a `RenderBox`. You can remove the `?` and the if-statement. Also, we probably don't need the comment above this line. You can move the `fieldBox` line right above the `viewportBox` line.

super_editor/lib/src/infrastructure/super_textfield/android/android_textfield.dart
```diff
@@ -379,6 +396,42 @@ class SuperAndroidTextFieldState extends State<SuperAndroidTextField>
...
+    if (fieldBox == null) {
+      return;
+    }
+
+    final viewportBox = ancestorScrollable.context.findRenderObject() as RenderBox;
+
+    // We get the selection offset so the autoscroll also supports
+    // multi-line textfields with unbounded lines
+    final offsetInsideTextField = widget.maxLines == null && _textEditingController.selection.isValid
```
matthew-carroll, line 418: Let's call this `textFieldFocalPoint` and let's change the comment to "Compute the text field offset that should be visible to the user"

super_editor/lib/src/infrastructure/super_textfield/android/android_textfield.dart
```diff
@@ -379,6 +396,42 @@ class SuperAndroidTextFieldState extends State<SuperAndroidTextField>
...
+            TextPosition(offset: _textEditingController.selection.extentOffset),
+          )
+        : Offset.zero;
+
+    final offsetInsideViewport = viewportBox.globalToLocal(
+      fieldBox.localToGlobal(offsetInsideTextField),
+    );
+
+    fieldBox.showOnScreen(
+      rect: Rect.fromLTWH(offsetInsideViewport.dx, offsetInsideViewport.dy, fieldBox.size.width, fieldBox.size.height),
```
matthew-carroll, line 429: If the ancestor scrollable takes up the whole screen, won't `offsetInsideViewport.dy` cause the caret to move all the way up and under the status bar? Don't we want that `dy` to be roughly centered vertically in the viewport?

super_editor/lib/src/infrastructure/super_textfield/android/android_textfield.dart
```diff
@@ -379,6 +396,42 @@ class SuperAndroidTextFieldState extends State<SuperAndroidTextField>
     return KeyEventResult.handled;
   }
 
+  /// Scrolls the ancestor [Scrollable], if any, so [SuperTextField]
+  /// is visible on the viewport when it's focused
+  void _ensureIsVisible() {
```
matthew-carroll, line 401: Let's call this `_autoScrollToKeepTextFieldVisible()`

super_editor/lib/src/infrastructure/super_textfield/android/android_textfield.dart
angelosilvestre: In fact, there was a mistake in the implementation. `showOnScreen` expects the `rect` to be the visible portion within the `RenderObject`. It's fixed now

matthew-carroll: If that mistake made it into the PR, is there a test that now ensures we have the right behavior?

angelosilvestre: It was making the text field visible, but it was scrolling more than necessary. I changed the test to ensure we are not scrolling too much

super_editor/lib/src/infrastructure/super_textfield/android/android_textfield.dart
```diff
@@ -379,6 +396,43 @@ class SuperAndroidTextFieldState extends State<SuperAndroidTextField>
...
+    );
+    final fieldBox = context.findRenderObject() as RenderBox;
+
+    // The area of the text field that should be revealed.
+    // We add a small margin to leave some space between the text field and the keyboard.    
+    final textFieldFocalRect = Rect.fromLTWH(
+      textFieldFocalPoint.dx,
+      textFieldFocalPoint.dy,
+      fieldBox.size.width,
+      lineHeight + 30,
```
matthew-carroll, line 426: What's this magic number? If this number really should be here, then please add a comment on this line about why. And if this number is related to font size or line height, then it almost certainly shouldn't be a constant.

I see that later in a test there happens to be a `gap` of `30` between the caret and the keyboard. If that's what this number represents, then it should be a constant variable with an appropriate name. If you'd like to share that value with the test, then consider placing that constant in a public location that's used by the Android field, the iOS field, and the test.

super_editor/test/super_textfield/super_textfield_auto_scroll_test.dart
```diff
@@ -0,0 +1,178 @@
...
+  group('SuperTextField on mobile', () {
+    group('with an ancestor Scrollable', () {
+      const screenSizeWithoutKeyboard = Size(400, 800);
+      const screenSizeWithKeyboard = Size(400, 300);
+
+      // Space between the caret and the keyboard.
+      const gap = 30;
+
+      testWidgetsOnMobile('auto scrolls when focused in single-line', (tester) async {
+        tester.binding.window
```
matthew-carroll, line 18: It looks like all these tests repeat the `tester` config and the window resize. Can we define those things in a local test method? For example:

```dart
testWidgetsOnMobileWithKeyboard("some name", (tester, keyboardToggle) async {
  // The WidgetTester setup is already done at this point

  // TODO: implement this test's setup

  // Simulate a keyboard opening, which reduces available screen height.
  await keyboardToggle.open();

  // TODO: implement this test's expectations
});
```

super_editor/lib/src/infrastructure/super_textfield/ios/ios_textfield.dart
```diff
@@ -370,6 +387,45 @@ class SuperIOSTextFieldState extends State<SuperIOSTextField>
...
+            TextPosition(offset: _textEditingController.selection.extentOffset),
+          )
+        : Offset.zero;
+
+    final lineHeight = _textContentKey.currentState!.textLayout.getLineHeightAtPosition(
+      TextPosition(offset: _textEditingController.selection.extentOffset),
+    );
+    final fieldBox = context.findRenderObject() as RenderBox;
+
+    const gapBetweenCaretAndKeyboard = 30;
```
matthew-carroll, line 411: If this is the same everywhere, shouldn't it be defined publicly somewhere so that everyone is using the same value?

super_editor/test/super_textfield/super_textfield_auto_scroll_test.dart
```diff
@@ -0,0 +1,202 @@
...
+              ),
+            ),
+          ],
+        ),
+      ),
+    ),
+  );
+}
+
+void testWidgetsOnMobileWithKeyboard(
```
matthew-carroll, line 147: All the local classes should be made private to this file.

super_editor/lib/src/infrastructure/super_textfield/metrics.dart
```diff
@@ -0,0 +1 @@
+const gapBetweenCaretAndKeyboard = 30;
```
matthew-carroll, line 1: Please include a comment that explains what this distance means. For example, is the distance between the top of the caret and the keyboard, or the bottom of the caret? And what purpose does this number play in our system?

matthew-carroll (approved):
LGTM - Thanks

PR #694, opened 2022-07-10
Fixed mobile handle dragging with ancestor Scrollables (Resolves #525)

### LINKED ISSUE #525: [Mobile] Selection handles are not easy to "grab" when the document is inside a CustomScrollView
Steps to reproduce:
* Update "Sliver Demo Editor" so that it works on iOS:
```
```
* Double-tap on a word to select it
* Drag one of the selection handles to change the selection

**Expected:** selection changes
**Actual:** document scrolls

https://user-images.githubusercontent.com/31278849/161970539-9406b176-bbe5-4eac-b850-f14f1d6fcc18.mp4

matthew-carroll:
Fixed mobile handle dragging with ancestor Scrollables (Resolves #525)

The root problem seems to be that we were using a `PanGestureRecognizer` but `Scrollable` uses a `VerticalDragGestureRecognizer`. I thought maybe our `PanGestureRecognizer`was losing out in the gesture arena. So I changed our recognizer to also be a `VerticalDragGestureRecognizer` and it now seems to beat out the ancestor `Scrollable`.

This ancestor gesture behavior is unintuitive. I would consider this to be a bug in Flutter. Nonetheless, I'm now able to easy drag the iOS caret around in the sliver demo.

### DIFF
--- a/super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
@@ -557,7 +557,12 @@ class _IOSDocumentTouchInteractorState extends State<IOSDocumentTouchInteractor>
     widget.focusNode.requestFocus();
   }
 
+  void _onPanDown(DragDownDetails details) {
+    print("ON PAN DOWN");
+  }
+
   void _onPanStart(DragStartDetails details) {
+    print("ON PAN START");
     // TODO: to help the user drag handles instead of scrolling, try checking touch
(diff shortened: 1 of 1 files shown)


matthew-carroll:
@miguelcmedeiros can you try this PR with the sliver demo, and maybe also try it in areas of Superlist that use an ancestor `Scrollable`?

angelosilvestre (approved):
LGTM

miguelcmedeiros (approved):
LGTM

PR #696, opened 2022-07-12
Add test for tapping to select an empty document (Resolves #646)

### LINKED ISSUE #646: Add tests for tapping to select an empty document
Add tests to ensure that tapping anywhere in an empty document gives focus to the document, and places the caret at the beginning of the document.

See `supereditor_gestures_test.dart` for a place to add a test.

angelosilvestre:
Add test for tapping to select an empty document. Resolves https://github.com/superlistapp/super_editor/issues/646

Tapping anywhere in a document with a single empty paragraph should give focus to the document and place caret at the beginning of the document.

### DIFF
--- a/super_editor/test/super_editor/supereditor_gestures_test.dart
+++ b/super_editor/test/super_editor/supereditor_gestures_test.dart
@@ -9,7 +9,123 @@ import 'supereditor_inspector.dart';
 
 void main() {
   group('SuperEditor gestures', () {
-    testWidgetsOnAllPlatforms('places caret at the beginning when tapping at an empty document', (tester) async {
+    testWidgetsOnAllPlatforms('places the caret when tapping an empty document (top left)', (tester) async {
+      final testContext = await tester
+          .createDocument() //
+          .withSingleEmptyParagraph()
+          .withEditorSize(const Size(300, 300))
+          .pump();
+
+      // Ensure editor is not focused
+      expect(SuperEditorInspector.hasFocus(), isFalse);
+
+      // Tap inside SuperEditor
+      await tester.tapAt(tester.getTopLeft(find.byType(SuperEditor)) + const Offset(1, 1));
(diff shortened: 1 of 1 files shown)


super_editor/test/super_editor/supereditor_gestures_test.dart
```diff
@@ -0,0 +1,41 @@
...
+import 'package:flutter_test/flutter_test.dart';
+import 'package:super_editor/super_editor.dart';
+
+import '../test_tools.dart';
+import 'document_test_tools.dart';
+import 'supereditor_inspector.dart';
+
+void main() {
+  group('SuperEditor gestures', () {
+    testWidgetsOnAllPlatforms('places caret at the beginning when tapping at an empty document', (tester) async {
```
matthew-carroll, line 12: I think the name "places the caret when tapping an empty document" is sufficient. Also, FYI, if the document wasn't empty, the position wouldn't be at the "beginning".

super_editor/test/super_editor/supereditor_gestures_test.dart
```diff
@@ -0,0 +1,41 @@
...
+          .createDocument() //
+          .withSingleEmptyParagraph()
+          .withEditorSize(const Size(300, 300))
+          .pump();
+
+      // Ensure editor is not focused
+      expect(SuperEditorInspector.hasFocus(), isFalse);
+
+      // Tap inside SuperEditor
+      await tester.tapAt(tester.getCenter(find.byType(SuperEditor)));
```
matthew-carroll, line 23: We should be a bit more exhaustive with a set of tests. We should test taps at all four corners. You'll need to either write multiple tests, or remove focus in between taps.

matthew-carroll (commented):
LGTM - Thanks

matthew-carroll (approved):

PR #698, opened 2022-07-12
Fix serialization of spans with intersecting attributions (Resolves #526)

### LINKED ISSUE #526: Markdown serialization cannot handle attributions which share a starting index but end in different places
I spent a while trying to think of a title, this was unfortunately the best I could come up with.

Currently, the way attributed text to markdown serialization works is that at each point where the set of attributes changes, the serializer inserts the characters used to delimit that markdown formatting for *all* of the attributes on that text.

This works just fine for one off styles:
| Super Editor  | Serialized Markdown | Markdown Rendered
| ------------- | ------------- | ------------- |
| Plain **Bold** Plain| `Plain **Bold** Plain`  | Plain **Bold** Plain |
| Plain ~Strike~ Plain| `Plain ~Strike~ Plain`  | Plain  ~Strike~ Plain |

And can work in cases where a simple style is fully surrounded by a different style (despite some redundancy):
| Super Editor  | Serialized Markdown | Markdown Rendered
| ------------- | ------------- | ------------- |
| **Bold ***Bold and Italic*** Bold**| `**Bold *****Bold and Italic***** Bold**`  | **Bold *****Bold and Italic***** Bold** |
| **Bold **~Bold and Strike~** Bold**| `**Bold ****~Bold and Strike~**** Bold**`  | **Bold *****~Bold and Strike~***** Bold** |

But it breaks down when two attributes begin at the same point, but end at different ones:
| Super Editor  | Serialized Markdown | Markdown Rendered
| ------------- | ------------- | ------------- |
| ***Bold and Italic* Bold**| `***Bold and Italic***** Bold**`  | ***Bold and Italic***** Bold** |
| **~Bold and Strike~ Bold**| `**~Bold and Strike~**** Bold**`  | **~Bold and Strike~**** Bold** |

In this last example, the serializer inserts `**~` to begin the section of attributed text which is both bold and has a strikethrough. At the start of the next section, where the text no longer has a strikethrough, it closes off the previous section with the inverse set of characters (`~**`) and then immediately begins a new section with bold (`**`), which is then closed at the end of the text. The output is invalid markdown with unclosed tags, which isn't rendered correctly.

This issue is also signficant for attributed text with asymmetrical markdown, like links, which are split. Not a showstopper, but also not the behaviour a user would expect.
| Super Editor  | Serialized Markdown | Markdown Rendered
| ------------- | ------------- | ------------- |
| [**Bold and Link** Link](https://github.com)| `[**Bold and Link**](https://github.com)[ Link](https://github.com)`  | [**Bold and Link**](https://github.com)[ Link](https://github.com) |
| [Link **Bold and Link** Link](https://github.com)| `[Link ](https://github.com)[**Bold and Link**](https://github.com)[ Link](https://github.com)`  | [Link ](https://github.com)[**Bold and Link**](https://github.com)[ Link](https://github.com) |

To resolve this issue, the markdown serializer should be aware of attributions on the previous span and only insert style markers for the difference between the two attribution sets.

angelosilvestre:
Fix serialization of spans with intersecting attributions. Resolves https://github.com/superlistapp/super_editor/issues/526

Markdown serialization was not working properly for attributions which share a starting index but end in different places. 

For example:
**~Bold and Strike~ Bold** was being serialized as `**~Bold and Strike~**** Bold**`

I changed the serialization logic to don't add start/end markers for attributions that are present in the previous/next span.

### DIFF
--- a/attributed_text/lib/src/attributed_text.dart
+++ b/attributed_text/lib/src/attributed_text.dart
@@ -309,9 +309,27 @@ class AttributedText {
 
   void visitAttributions(AttributionVisitor visitor) {
     final collapsedSpans = spans.collapseSpans(contentLength: text.length);
-    for (final span in collapsedSpans) {
-      visitor(this, span.start, span.attributions, AttributionVisitEvent.start);
-      visitor(this, span.end, span.attributions, AttributionVisitEvent.end);
+    for (int i = 0; i < collapsedSpans.length; i++) {
+      final currentSpan = collapsedSpans[i];
+      final previousSpan = i > 0 ? collapsedSpans[i - 1] : null;
+      final nextSpan = i < collapsedSpans.length - 1 ? collapsedSpans[i + 1] : null;
+
+      // When the previous span ends right before the current one
+      // whe only add start markers for the attributions that weren't present
+      // in the previous span.
+      final startAtributions = previousSpan == null || previousSpan.end != currentSpan.start - 1 //
+          ? currentSpan.attributions
+          : currentSpan.attributions.where((e) => !previousSpan.attributions.contains(e)).toSet();
+
+      // When the next span starts right after the current one
+      // whe only add end markers for the attributions that won't be present
+      // in the next span.
+      final endAtributions = nextSpan == null || nextSpan.start != currentSpan.end + 1 //
+          ? currentSpan.attributions
+          : currentSpan.attributions.where((e) => !nextSpan.attributions.contains(e)).toSet();
+
+      visitor(this, currentSpan.start, startAtributions, AttributionVisitEvent.start);
+      visitor(this, currentSpan.end, endAtributions, AttributionVisitEvent.end);
     }
   }
 
--- a/attributed_text/lib/src/attributed_text.dart
+++ b/attributed_text/lib/src/attributed_text.dart
@@ -314,17 +314,15 @@ class AttributedText {
       final previousSpan = i > 0 ? collapsedSpans[i - 1] : null;
       final nextSpan = i < collapsedSpans.length - 1 ? collapsedSpans[i + 1] : null;
 
-      // When the previous span ends right before the current one
-      // whe only add start markers for the attributions that weren't present
+      // Whe only add start markers for attributions that weren't present
       // in the previous span.
-      final startAtributions = previousSpan == null || previousSpan.end != currentSpan.start - 1 //
+      final startAtributions = previousSpan == null //
           ? currentSpan.attributions
           : currentSpan.attributions.where((e) => !previousSpan.attributions.contains(e)).toSet();
 
-      // When the next span starts right after the current one
-      // whe only add end markers for the attributions that won't be present
+      // Whe only add end markers for attributions that won't be present
       // in the next span.
-      final endAtributions = nextSpan == null || nextSpan.start != currentSpan.end + 1 //
+      final endAtributions = nextSpan == null //
           ? currentSpan.attributions
           : currentSpan.attributions.where((e) => !nextSpan.attributions.contains(e)).toSet();
 
--- a/super_editor_markdown/test/super_editor_markdown_test.dart
+++ b/super_editor_markdown/test/super_editor_markdown_test.dart
@@ -153,6 +153,27 @@ This is some code
         expect(serializeDocumentToMarkdown(doc), 'This ***is a*** paragraph.');
       });
 
+      test('paragraph with non-overlapping bold and italics', () {
+        final doc = MutableDocument(nodes: [
+          ParagraphNode(
+            id: '1',
+            text: AttributedText(
+              text: 'This is a paragraph.',
+              spans: AttributedSpans(
+                attributions: [
+                  const SpanMarker(attribution: boldAttribution, offset: 0, markerType: SpanMarkerType.start),
+                  const SpanMarker(attribution: boldAttribution, offset: 6, markerType: SpanMarkerType.end),
+                  const SpanMarker(attribution: italicsAttribution, offset: 8, markerType: SpanMarkerType.start),
+                  const SpanMarker(attribution: italicsAttribution, offset: 19, markerType: SpanMarkerType.end),
+                ],
+              ),
+            ),
+          ),
+        ]);
+
+        expect(serializeDocumentToMarkdown(doc), '**This is** *a paragraph.*');
+      });
+
       test('paragraph with intersecting bold and italics', () {
         final doc = MutableDocument(nodes: [
           ParagraphNode(
@@ -172,9 +193,27 @@ This is some code
         ]);
 
         expect(serializeDocumentToMarkdown(doc), 'This ***is a** paragraph*.');
-      }, skip: '''Markdown serialization is currently broken for intersecting styles.
-                  See https://github.com/superlistapp/super_editor/issues/526''');
+      });
 
+      test('paragraph with intersecting bold and strikethrough', () {
+        final doc = MutableDocument(nodes: [
+          ParagraphNode(
+            id: '1',
+            text: AttributedText(
+              text: 'Bold and Strike Bold',
+              spans: AttributedSpans(attributions: [
+                const SpanMarker(attribution: boldAttribution, offset: 0, markerType: SpanMarkerType.start),
+                const SpanMarker(attribution: boldAttribution, offset: 19, markerType: SpanMarkerType.end),
+                const SpanMarker(attribution: strikethroughAttribution, offset: 0, markerType: SpanMarkerType.start),
+                const SpanMarker(attribution: strikethroughAttribution, offset: 14, markerType: SpanMarkerType.end),
+              ]),
+            ),
+          ),
+        ]);
+
+        expect(serializeDocumentToMarkdown(doc), '**~Bold and Strike~ Bold**');
+      });
+    
       test('paragraph with overlapping code and bold', () {
         final doc = MutableDocument(nodes: [
           ParagraphNode(
@@ -273,8 +312,28 @@ This is some code
         ]);
 
         expect(serializeDocumentToMarkdown(doc), '[This **is a** paragraph](https://example.org).');
-      }, skip: '''Markdown serialization is currently broken for intersecting styles.
-                  See https://github.com/superlistapp/super_editor/issues/526''');
+      });
+
+      test('paragraph with consecutive links', () {
+        final doc = MutableDocument(nodes: [
+          ParagraphNode(
+            id: '1',
+            text: AttributedText(
+              text: 'First LinkSecond Link',
+              spans: AttributedSpans(
+                attributions: [
+                  SpanMarker(attribution: LinkAttribution(url: Uri.https('example.org', '')), offset: 0, markerType: SpanMarkerType.start),
+                  SpanMarker(attribution: LinkAttribution(url: Uri.https('example.org', '')), offset: 9, markerType: SpanMarkerType.end),
+                  SpanMarker(attribution: LinkAttribution(url: Uri.https('github.com', '')), offset: 10, markerType: SpanMarkerType.start),
+                  SpanMarker(attribution: LinkAttribution(url: Uri.https('github.com', '')), offset: 20, markerType: SpanMarkerType.end),
+                ],
+              ),
+            ),
+          ),
+        ]);
+
+        expect(serializeDocumentToMarkdown(doc), '[First Link](https://example.org)[Second Link](https://github.com)');
+      });
 
       test('image', () {
         final doc = MutableDocument(nodes: [
--- a/super_editor_markdown/pubspec.yaml
+++ b/super_editor_markdown/pubspec.yaml
@@ -22,6 +22,8 @@ dependency_overrides:
     path: ../super_editor
   super_text_layout:
     path: ../super_text_layout
+  attributed_text:
+    path: ../attributed_text
 
 dev_dependencies:
   # TODO: upgrade lints to 2.0.1 when we release super_editor 0.2.1


attributed_text/lib/src/attributed_text.dart
```diff
@@ -309,9 +309,25 @@ class AttributedText {
 
   void visitAttributions(AttributionVisitor visitor) {
     final collapsedSpans = spans.collapseSpans(contentLength: text.length);
-    for (final span in collapsedSpans) {
-      visitor(this, span.start, span.attributions, AttributionVisitEvent.start);
-      visitor(this, span.end, span.attributions, AttributionVisitEvent.end);
+    for (int i = 0; i < collapsedSpans.length; i++) {
```
matthew-carroll, line 312: The concept of visiting attributions is that any `visitor` gets to visit every attribution. My first reaction to this change is that we shouldn't be filtering anything from the `visitor`. The `visitor` needs to decide what to do.

Is that possible in this case?

super_editor_markdown/test/super_editor_markdown_test.dart
```diff
@@ -153,6 +153,27 @@ This is some code
         expect(serializeDocumentToMarkdown(doc), 'This ***is a*** paragraph.');
       });
 
+      test('paragraph with non-overlapping bold and italics', () {
```
matthew-carroll, line 156: Somewhere around here I think we have some skipped tests related to this. At least, I think this was the problem that lead to us skipping them. Can you check for the skipped tests and see if they should now be run? Also, if those tests cover things you've written here, we can just use the skipped tests.

super_editor_markdown/lib/src/markdown.dart
```diff
@@ -464,30 +464,74 @@ extension on AttributedText {
...
+          final linkMarker = _encodeLinkMarker(startAttributions, event);
+
           spanStart = index;
           buffer
             ..write(linkMarker)
             ..write(markdownStyles);
           break;
         case AttributionVisitEvent.end:
+          // The span end is handled when we visit the next span or after we visit the last one
+          hasPendingEnd = true;
```
matthew-carroll, line 512: I'm confused by the term "hasPendingEnd". I would expect this to mean "there's a span that hasn't ended yet", but you're setting it to the `true` when a span ends...

super_editor_markdown/test/super_editor_markdown_test.dart
```diff
@@ -153,6 +153,27 @@ This is some code
...
+                  const SpanMarker(attribution: boldAttribution, offset: 6, markerType: SpanMarkerType.end),
+                  const SpanMarker(attribution: italicsAttribution, offset: 8, markerType: SpanMarkerType.start),
+                  const SpanMarker(attribution: italicsAttribution, offset: 19, markerType: SpanMarkerType.end),
+                ],
+              ),
+            ),
+          ),
+        ]);
+
+        expect(serializeDocumentToMarkdown(doc), '**This is** *a paragraph.*');
```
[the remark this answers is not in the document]
angelosilvestre: I haven't found any test including 2 different non-overlapping styles

super_editor_markdown/lib/src/markdown.dart
matthew-carroll: Neither of those really communicate any new information to me. I honestly don't know what that property is doing, so I can't suggest a better name, but that's a good indication that this name should be improved. And maybe some others, too. When I tried to read through this method, it wasn't obvious what was happening or why. I'm also wondering, does this method handle more than 2 overlapping attributions?

angelosilvestre: Yes, it should handle more than 2 overlapping attributions. I could write a test for that if needed.

When we are visiting the end of a span we still don't know what attributions will be present in the next span, so we are not able to determine the attributions we should close. 

Because of this, I changed the code to close the attributions right before we handle the start of the next span (or after the end of the last span). At that point, we have access to the attributions of both spans, so we can close/open only the attributions that aren't present in both spans.

The property indicates that there is a span end we didn't handle yet.

matthew-carroll: > When we are visiting the end of a span we still don't know what attributions will be present in the next span, so we are not able to determine the attributions we should close.

Why is this true? Shouldn't we always close a Markdown style where the END marker appears? For example, there should never be a case where "bold" ends, and then immediately begins. So if you visit the END marker for "bold", why wouldn't you serialize the markdown for the end of the bold?

angelosilvestre: This is the case when two attributions share the same start index, but end in different indexes.

Considering this example: **~Bold and Strike~ Bold**

We have a first span with both Bold and Strikethrough attributions and then a second span that has only bold attribution. Currently, we are closing both attributions and then opening bold again.

With this change, we wait until we see the start of next span, then we ignore closing/re-opening the attributions that are present in both spans.

matthew-carroll: Ok. That explanation makes sense. But I'm not seeing the connection between that explanation and the way the code reads. So... I understand that based on your explanation we don't want to end a bold span, we only want to end an italics span. But my question above doesn't preclude that.

When we run into the END marker for italics, shouldn't we serialize the end of italics, e.g., "*"? What do the other styles have to do with making that decision at the END of an italics span?

angelosilvestre: When we visit the end of the first span, the `attributions` param contains both attributions. So, without looking at the next span, we have no way to know which end markers we should add.

matthew-carroll: You're saying that we're getting an `END` `event`, and with that `END` `event` we're getting an `attributions` set that includes `bold` and `italics`, even though only `italics` ended?

angelosilvestre: Yes, after that we get an `START` `event` for `bold` again.

matthew-carroll: Ah. Ok. If that's the case then my next thought is that the root cause is a bad visitor API.

What do you think about a visitor API that passes `allActiveAttributions` and `eventAttributions`? The `eventAttributions` would only contain those attributions that are `START`ing or `END`ing.

angelosilvestre: Sounds good. I will change the visitor API.

attributed_text/lib/src/attributed_text.dart
```diff
@@ -309,9 +309,23 @@ class AttributedText {
...
```
matthew-carroll, line 312: Seeing that this operates on collapsed spans makes me think that I probably combined two different goals. I think the purpose of collapsed spans was to make it trivial to put together `TextSpan`s with a series of `TextStyle`s. But I think that in reality we want a more generic visitor.

I think we want the visitor to go marker by marker, sending each `START` and `END` marker that it comes across. The new visitor API is probably correct, but I don't think that `visitAttributions()` should use `collapsedSpans`. 

Instead, to achieve the existing style behavior, I think we could define a `CollapsedSpansAttributionVisitor`, which can then be extended to create `TextSpanBuilderVisitor`. The idea is that the `CollapsedSpansAttributionVisitor` would visit the markers and internally it would call a protected method like `onSpansChange(oldSet, newSet)`. Then the `TextSpanBuilderVisitor` would implement that method and accumulate `TextSpan`s like it does now with the collapsed spans.

matthew-carroll: Actually, in addition to what I said above, we should alter the visitor API a little more than what I said previously. I'm thinking:

`visit(AttributedText, int index, startingAttributions, endingAttributions)`

And I think we can eliminate `activeAttributions` because the visitor can track that, as needed.

attributed_text/lib/src/attributed_text.dart
```diff
@@ -308,24 +308,36 @@ class AttributedText {
...
-
-      visitor(this, currentSpan.start, currentSpan.attributions, startingAttributions, AttributionVisitEvent.start);
-      visitor(this, currentSpan.end, currentSpan.attributions, endingAttributions, AttributionVisitEvent.end);
+    final emptyAttributions = Set<Attribution>.unmodifiable([]);
+
+    // Maps the indexes to the attributions that are changing at that index.
+    Map<int, ChangingAttributions> changingIndexes = {};
+
+    // Groups start/end markers by index.
+    for (final marker in spans.markers) {
```
matthew-carroll, line 317: You're batching all change information and then running the visitor. I think you can do both at the same time.

You can track a `currentIndex`, which begins at `-1`, and maintain two sets: `startingAttributions` and `endingAttributions`. 

In your loop, when you come upon a new marker, one of two things should happen:

If the new marker is at the `currentIndex`, add the new marker to one of your two sets. Continue looping.

Else, the new marker must sit at a new offset. So you've found all the markers at `currentOffset`.
 - call the visitor for `currentOffset` and pass the two attribution sets
 - update `currentOffset` to the new marker index
 - clear all values in the attribution sets
 - add the new marker to the appropriate attribution set
 - continue

This approach eliminates the `ChangingAttributions` class, gets us down to a single loop, and might be easier to read.

super_editor_markdown/lib/src/markdown.dart
```diff
@@ -464,30 +464,48 @@ extension on AttributedText {
...
+          ..write(fullText.text.substring(spanStart, index))
+          ..write(linkMarker)
+          ..write(markdownStyles);
+
+        spanStart = index;
       }
     });
 
+    // When the last span has no attributions, we still have text that wasn't added to the buffer yet.    
+    if (spanStart <= text.length - 1) {
```
matthew-carroll, line 506: I can see that most use-cases will probably need to do something when the visitation is done.

Let's create an abstract class for visitors:

```dart
abstract class AttributionVisitor {
  void onVisitBegin();

  void visitAttributions(...);

  void onVisitEnd();
}
```

This will give visitors the opportunity to initialize at the beginning, finalize at the end, and implementers can store local variables if they use a subclass.

But we should also provide a callback version for simple use-cases:

```dart
class CallbackAttributionVisitor implements AttributionVisitor {
  CallbackAttributionVisitor({
    VoidCallback? onVisitBegin,
    required VisitAttributionsCallback visitAttributions,
    VoidCallback? onVisitEnd,
  }) //...

  //...
}
```

attributed_text/lib/src/attributed_text.dart
```diff
@@ -308,11 +308,40 @@ class AttributedText {
...
+    int currentIndex = -1;
+
+    visitor.onVisitBegin();
+
+    for (final marker in spans.markers) {
+      // If the marker offset differs from the currentIndex it means
+      // we already added all the attributions that start or end
+      // at currentIndex.
+      if (marker.offset != currentIndex) {
+        if (currentIndex != -1) {
```
matthew-carroll, line 322: It's probably safer to say `>= 0`

attributed_text/lib/src/attributed_text.dart
```diff
@@ -308,11 +308,40 @@ class AttributedText {
...
+      }
+
+      if (marker.isStart) {
+        startingAttributions.add(marker.attribution);
+      } else {
+        endingAttributions.add(marker.attribution);
+      }
     }
+
+    // After the loop, we still have the last index with markers to visit,
```
matthew-carroll, line 338: I'm not sure that this comment explains the situation. How about `Visit the final set of end markers`

attributed_text/lib/src/attributed_text.dart
```diff
@@ -308,11 +308,40 @@ class AttributedText {
...
+    // if there is at least one marker.
+    if (startingAttributions.isNotEmpty || endingAttributions.isNotEmpty) {
```
matthew-carroll, line 340: Any starting attributions at this location would be an error, right?

attributed_text/lib/src/attributed_text.dart
```diff
@@ -330,25 +359,84 @@ class AttributedText {
...
+  Set<Attribution> startingAttributions,
+  Set<Attribution> endingAttributions,
 );
 
 enum AttributionVisitEvent {
   start,
   end,
 }
 
+/// Visitor interface that allows performing actions
```
matthew-carroll, line 380: `Visitor that visits every start and end attribution marker in an [AttributedText]`

attributed_text/lib/src/attributed_text.dart
```diff
@@ -330,25 +359,84 @@ class AttributedText {
...
 
 enum AttributionVisitEvent {
   start,
   end,
 }
 
+/// Visitor interface that allows performing actions 
+/// before and after visiting the attributions.
+abstract class AttributionVisitor {
+  /// Called exactly one time before visiting the attributions.
```
matthew-carroll, line 383: `Called before visiting attributions, so that implementers can perform any desired setup.`

attributed_text/lib/src/attributed_text.dart
```diff
@@ -330,25 +359,84 @@ class AttributedText {
...
+  void onVisitBegin();
+
+  /// Called to each index that has at least one attribution that starts
```
matthew-carroll, line 386:
```
/// Visits all starting and ending attribution markers at the given [index] within [fullText].
///
/// This method isn't called for indices that don't contain any attribution markers.
```

attributed_text/lib/src/attributed_text.dart
```diff
@@ -330,25 +359,84 @@ class AttributedText {
...
+  /// or ends at that index.
+  void visitAttributions(
+    AttributedText fullText,
+    int index,
+    Set<Attribution> startingAttributions,
+    Set<Attribution> endingAttributions,
+  );
+
+  /// Called exactly one time after visiting the attributions.
```
matthew-carroll, line 395: `Called after all attribution markers have been visited by [visitAttributions].`

attributed_text/lib/src/attributed_text.dart
```diff
@@ -330,25 +359,84 @@ class AttributedText {
...
+  void onVisitEnd();
+}
+
+/// [AttributionVisitor] implementation intended for simple use-cases
```
matthew-carroll, line 399: `[AttributionVisitor] that delegates to given callbacks.`

attributed_text/lib/src/attributed_text.dart
```diff
@@ -330,25 +359,84 @@ class AttributedText {
...
+/// where callbacks are sufficient.
+class CallbackAttributionVisitor implements AttributionVisitor {
+  /// Creates a callback based [AttributionVisitor]
```
matthew-carroll, line 402: I think we can get rid of this constructor comment. It's restating the docs for the methods that already exist.

attributed_text/lib/src/attributed_text.dart
```diff
@@ -330,25 +359,84 @@ class AttributedText {
...
+  /// visiting the attributions.
+  CallbackAttributionVisitor({
+    VoidCallback? onVisitBegin,
+    required VisitAttributionsCallback visitAttributions,
+    VoidCallback? onVisitEnd,
+  })  : _onVisitBegin = onVisitBegin,
+        _onVisitEnd = onVisitEnd,
+        _onVisitAttributions = visitAttributions;
+
+  final VoidCallback? _onVisitBegin;
```
matthew-carroll, line 420: Please declare the callbacks in the same order that they're called.

super_editor_markdown/lib/src/markdown.dart
```diff
@@ -412,6 +412,74 @@ class _InlineMarkdownToDocument implements md.NodeVisitor {
 }
 
 extension on AttributedText {
+  String toMarkdown() {
+    final serializer = MarkdownSerializer(this);
+    return serializer.serialize();
+  }
+}
+
+class MarkdownSerializer implements AttributionVisitor {
```
matthew-carroll, line 421: This is an `AttributedTextMarkdownSerializer`, not to be confused with one that serializes a document or something else.

matthew-carroll: Also please add a Dart doc

super_editor_markdown/lib/src/markdown.dart
```diff
@@ -412,6 +412,74 @@ class _InlineMarkdownToDocument implements md.NodeVisitor {
...
+  MarkdownSerializer(this.attributedText);
+
+  final AttributedText attributedText;
```
matthew-carroll, line 424: Does this need to be public? In fact, should this really be a property? Or should it be passed into `serialize()`?

super_editor_markdown/lib/src/markdown.dart
```diff
@@ -412,6 +412,74 @@ class _InlineMarkdownToDocument implements md.NodeVisitor {
...
+  int _spanStart = 0;
+  final _buffer = StringBuffer();
+
+  String serialize() {
+    attributedText.visitAttributions(this);
+    return _buffer.toString();
+  }
+
+  @override
+  void onVisitBegin() {}
```
matthew-carroll, line 434: In the interface definition, let's make the begin and end calls empty stubs so that subclasses aren't forced to implement them.

matthew-carroll (approved):
LGTM - Can you move the PR out of draft mode?

PR #699, opened 2022-07-14
[Example App] Fix toolbar hiding the alignment menu (Resolves #670)

### LINKED ISSUE #670: [Example App] - Toolbar hides the alignment menu
In the example app, the popover toolbar allows the user to select text alignment options with a drop down menu. The drop down menu appears beneath the toolbar, obscuring some of the options.

You can see a screenshot of the problem here: https://github.com/superlistapp/super_editor/issues/380

angelosilvestre:
Fix toolbar hiding the alignment menu. Resolves https://github.com/superlistapp/super_editor/issues/670

Currently we use an `Overlay` to display the toolbar. This is causing the dropdown items to be displayed behind the toolbar.

After some research I found that `DropdownButton` displays its items in another route and that the overlay is always displayed above this route. Based on https://github.com/flutter/flutter/issues/22917#issuecomment-431457670 it seems that this is the expected behaviour.

I changed the example editor to display the toolbar in the stack.

Checking this ticket I've found another issue:

Clicking to change the text alignment is not applying the new alignment until an attribution is toggled. The reason is that `_RebuildOptimizedSuperTextWithSelectionState` is not invalidating the cached `SuperText` when the text alignment changes.

### DIFF
--- a/super_editor/example/lib/demos/example_editor/example_editor.dart
+++ b/super_editor/example/lib/demos/example_editor/example_editor.dart
@@ -32,11 +32,12 @@ class _ExampleEditorState extends State<ExampleEditor> {
   final _lightBackground = Colors.white;
   bool _isLight = true;
 
-  OverlayEntry? _textFormatBarOverlayEntry;
   final _textSelectionAnchor = ValueNotifier<Offset?>(null);
 
   OverlayEntry? _imageFormatBarOverlayEntry;
   final _imageSelectionAnchor = ValueNotifier<Offset?>(null);
+
+  bool _isToolbarVisible = false;
 
   @override
   void initState() {
@@ -55,10 +56,6 @@ class _ExampleEditorState extends State<ExampleEditor> {
 
   @override
   void dispose() {
-    if (_textFormatBarOverlayEntry != null) {
-      _textFormatBarOverlayEntry!.remove();
-    }
-
     _scrollController.dispose();
     _editorFocusNode.dispose();
     _composer.dispose();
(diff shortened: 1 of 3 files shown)


super_editor/example/lib/demos/example_editor/example_editor.dart
```diff
@@ -296,6 +268,7 @@ class _ExampleEditorState extends State<ExampleEditor> {
             if (_isMobile) _buildMountedToolbar(),
           ],
         ),
+        if (_isToolbarVisible) _buildEditorToolbar(),
```
matthew-carroll, line 271: Unfortunately, this isn't going to work. Editors are not always displayed fullscreen, and editors might be clipped. We need our popovers to escape the bounds of the editor, which we accomplish with the app `Overlay`.

angelosilvestre:
I changed the implementation wrapping the `Scaffold` with an `Overlay`.

That way the toolbar won't use the navigator's `Overlay` and the dropdown list is displayed correctly

super_editor/example/lib/main.dart
```diff
@@ -115,12 +115,18 @@ class _HomeScreenState extends State<HomeScreen> {
 
   @override
   Widget build(BuildContext context) {
-    return Scaffold(
-      key: _scaffoldKey,
-      appBar: _buildAppBar(context),
-      extendBodyBehindAppBar: true,
-      body: _selectedMenuItem!.pageBuilder(context),
-      drawer: _buildDrawer(),
+    return Overlay(
```
matthew-carroll, line 118: Please add a comment about we need our own `Overlay` in this demo app, especially given that this same widget is used for all the demos that we have.

matthew-carroll (approved):
LGTM with one last comment

PR #704, opened 2022-07-16
[SuperTextField] - Create a controller that supports undo/redo

matthew-carroll:
[SuperTextField] - Create a controller that supports undo/redo.

This PR creates an `EventSourcedAttributedTextEditingValue`, which holds text, a selection, and a composing region, along with a history of commands that were run to produce the current value. The value can `undo()` and `redo()` those commands.

The commands are based on an interface called `AttributedTextEditingValueCommand`.

I ported the existing `AttributedTextEditingController` interface over to a new one called `EventSourcedAttributedTextEditingController`. I implemented the interface using the new event sourced value and a group of new commands. I also wrote tests for nearly every operation in the controller, and their ability to undo an operation.

Some things that are not addressed in this PR:

- How do we handle composing attributions as things are undone and redone?
- How can a developer add new pieces of information to the history, given that all commands operate on a type-safe data structure?
- How can developers optionally batch commands in retrospect so that, e.g., similar commands in quick succession are batched and undone together.


matthew-carroll:
@venkatd - would you like to give this PR a try in your app and see if this undo/redo works for you?

venkatd:
Hi @matthew-carroll I really like the direction of this. We would want the ability to compose/group actions in order to replace our internal implementation. Having commands produce edits (or changes/events depending on ones preference) gives us undo and grouping for free in many cases.

If your time allows a quick call this month could let us compare notes on the approach and how we might be able to drop our hand-rolled undo/redo.

> How can a developer add new pieces of information to the history, given that all commands operate on a type-safe data structure?

Could you give an example of what you mean by this?

super_editor/lib/src/infrastructure/super_textfield/infrastructure/editing/commands.dart
```diff
@@ -0,0 +1,519 @@
...
+}
+
+extension on AttributedText {
+  AttributedText copy() {
+    return AttributedText(
+      text: text,
+      spans: spans.copy(),
+    );
+  }
+}
```
angelosilvestre, line 273: Wouldn't it be better if this extension was placed after all the commands instead of being placed between two commands?

angelosilvestre (approved):
LGTM with a few comments

matthew-carroll:
@venkatd it sounds like what you need is the ability for a command to look at recent history and re-group them, right?

In terms of what's introduced in this PR, are there any API decisions here that you think would prevent adoption of this approach? Should we merge in and then expand the behavior, or do you think we need to revisit what's here so far?

super_editor/lib/src/infrastructure/super_textfield/infrastructure/editing/commands.dart
```diff
@@ -0,0 +1,519 @@
...
+  TextSelection? _previousSelection;
+  TextRange? _previousComposingRange;
+
+  @override
+  AttributedTextEditingValue doExecute(AttributedTextEditingValue previousValue) {
+    _previousSelection = previousValue.selection;
+    _previousComposingRange = previousValue.composingRegion;
+
+    return AttributedTextEditingValue(
+      text: previousValue.text,
```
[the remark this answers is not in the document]
matthew-carroll: I don't think it matters because we're not changing the text, but I see that I copied the text in the "select all" command, so I'll add a `copy()` here, too.

super_editor/lib/src/infrastructure/super_textfield/infrastructure/editing/commands.dart
matthew-carroll: Yeah, I'll move it down.

matthew-carroll:
@angelosilvestre - I updated the PR for your comments. I implemented one of those empty tests and deleted the other two.

I also found a bug in the Markdown serialization. It was breaking on a single-character styles. I wrote a test suite and re-organized the code.

matthew-carroll:
FYI - I extracted the markdown serialization fixes from this PR and merged them separately. So now they've disappeared from this PR's file changes.

PR #705, opened 2022-07-16
[SuperEditor] Fix drag selection of non-selectable components (Resolves #363)

### LINKED ISSUE #363: Fix drag selection of non-selectable components
Non-selectable components mostly work as expected. However, if the user drags a selection that ends on a non-selectable component, the component is selected, when it shouldn't be.

Update the selection behavior to avoid selecting non-selectable components at the base or extent.

For the below:

- Expected behavior: pressing backspace doesn't delete the HR, because the HR is supposed to be unselectable
- Actual behavior: the non-selectable HR is selected, anyway, and then deleted

https://user-images.githubusercontent.com/7259036/145670745-a785dbd0-ba22-4e41-92d3-eb76a3c8fb84.mov

https://user-images.githubusercontent.com/7259036/145670747-a9164969-8b22-4364-8a7e-61d22f11f592.mov

You can find an example of non-selectable HRs in the example app.

angelosilvestre:
[SuperEditor]  Fix drag selection of non-selectable components. Resolves https://github.com/superlistapp/super_editor/issues/363

Dragging a selection that ends on a non-selectable component was causing the component to be selected.

I modified `getDocumentSelectionInRegion` to ignore components that are marked with `isVisuallySelectable` equal to `false`. That way, a non-selectable component can't be at the `base` or `extent` of a `DocumentSelection`.

However, we still can select the surrounding components, causing the non-selectable component to be present in the selected region.

I also added an option to configure the `componentBuilders` in the `TestDocumentConfigurator`.

### DIFF
--- a/super_editor/test/super_editor/supereditor_selection_test.dart
+++ b/super_editor/test/super_editor/supereditor_selection_test.dart
@@ -97,7 +97,7 @@ void main() {
       );
     });
 
-    testWidgetsOnArbitraryDesktop("doesn't select an unselectable component at base (upstream)", (tester) async {
+    testWidgetsOnArbitraryDesktop("doesn't select an unselectable component at base (dragging upstream)", (tester) async {
       final testContext = await _pumpUnselectableComponentTestApp(tester);
 
       final firstParagraphId = testContext.editContext.editor.document.nodes.first.id;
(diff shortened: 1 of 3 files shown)


super_editor/test/super_editor/supereditor_selection_test.dart
```diff
@@ -100,5 +96,226 @@ void main() {
         ),
       );
     });
+
+    testWidgetsOnArbitraryDesktop("doesn't select an unselectable component at base (upstream)", (tester) async {
```
matthew-carroll, line 100: The phrase "(upstream)" is a bit ambiguous. I think I'd say "(dragging upstream)".

matthew-carroll (approved):
LGTM with a test naming update

matthew-carroll (approved):
LGTM - Thanks

PR #706, opened 2022-07-17
[SuperEditor] Fix placement of caret for the second time

mHassan20896:
It was found that clearSelection() method in document_composer.dart was assigning null to selection setter to clear selection which is needed to be non-null for text to get inserted in the paragraph field after placing caret for the second.

To clear selection, selection setter is assigned with a new object with base attribute same as extent attribute of the previous selection object.

### DIFF
--- a/super_editor/lib/src/core/document_composer.dart
+++ b/super_editor/lib/src/core/document_composer.dart
@@ -48,8 +48,9 @@ class DocumentComposer with ChangeNotifier {
   final selectionNotifier = ValueNotifier<DocumentSelection?>(null);
 
   /// Clears the current [selection].
+  /// Previous extent position is set on base parameter to clear the selection
   void clearSelection() {
-    selection = null;
+    selection = _selection?.copyWith(base: _selection!.extent.copyWith());
   }
 
   final ValueNotifier<ImeConfiguration> imeConfiguration;
--- a/super_editor/lib/src/core/document_composer.dart
+++ b/super_editor/lib/src/core/document_composer.dart
@@ -48,7 +48,6 @@ class DocumentComposer with ChangeNotifier {
   final selectionNotifier = ValueNotifier<DocumentSelection?>(null);
 
   /// Clears the current [selection].
-  /// Previous extent position is set on base parameter to clear the selection
   void clearSelection() {
     selection = null;
   }
--- a/super_editor/lib/src/core/document_composer.dart
+++ b/super_editor/lib/src/core/document_composer.dart
@@ -50,7 +50,7 @@ class DocumentComposer with ChangeNotifier {
   /// Clears the current [selection].
   /// Previous extent position is set on base parameter to clear the selection
   void clearSelection() {
-    selection = _selection?.copyWith(base: _selection!.extent.copyWith());
+    selection = null;
   }
 
   final ValueNotifier<ImeConfiguration> imeConfiguration;
--- a/super_editor/test/super_editor/supereditor_robot_test.dart
+++ b/super_editor/test/super_editor/supereditor_robot_test.dart
@@ -150,8 +150,11 @@ void main() {
       // Type some text by simulating hardware keyboard key presses.
       await tester.typeKeyboardText("Hello, world!");
 
+      //Simulate the delay between first tap and second tap
+      await tester.pumpAndSettle(const Duration(milliseconds: 200));
+
       // Tap to place the caret in the last position
-      await tester.placeCaretInParagraph("1", 12);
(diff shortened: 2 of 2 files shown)


mHassan20896:
@matthew-carroll

super_editor/lib/src/core/document_composer.dart
```diff
@@ -48,8 +48,9 @@ class DocumentComposer with ChangeNotifier {
   final selectionNotifier = ValueNotifier<DocumentSelection?>(null);
 
   /// Clears the current [selection].
+  /// Previous extent position is set on base parameter to clear the selection
   void clearSelection() {
-    selection = null;
+    selection = _selection?.copyWith(base: _selection!.extent.copyWith());
```
matthew-carroll, line 53: Why is this the appropriate behavior for `clearSelection()`? The existing implementation looks correct to me.

matthew-carroll:
Thanks for posting a PR. I left a comment on the source code change.

Also, for the PR description, I'd like to see a description of the problem and solution that's easier for a reviewer to understand. This PR description mentions a number of specific technical details, but the big picture isn't clear. For example, if there's a bug in super_editor, how did we get to this point without noticing the bug? Why is the change in this PR appropriate for the problem? With the given change, how might other aspects of super_editor be effected?

mHassan20896:
Problem mentioned in the ticket is: when we place caret on paragraph and enter text by simulating keyboard input (for e.g. entered text is "Hello, World!"). Then place the caret again at the end of the previously entered text, and enter the text again (for e.g. "ABC").  After performing these 4 operations expected text is "Hello, World!ABC", but actual text is found to be "Hello, World!" which is wrong.

So far we couldn't catch this bug because our test cases did not cater the "placement of caret for the second time" scenario and I got to confirm this by executing test case given on the ticket which failed for the above explained scenario.

While debugging I found that when we enter text, insertCharacter() method of common_editor_operations.dart gets executed. There is a check in this method that sees if composer.selection == null then method returns without inserting the character if the condition is true. Then I searched for null assignment to the "selection" variable. I placed debug point on all of the searched lines to see which one and when one of these null assignment is getting executed. I found that it went to clearSelection() method of document_composer.dart file where selection variable is assigned to null. It gets into clearSelection() method when we call placeCaretInParagraph() from the test case for the second time. By doing the above brain storming I got to believe this is the line because of which the test case is failing.

To clear selection, one way is:
Assign null to selection variable. 

Another way I could thought of is:

Assign the copy of selection.extent to selection.base. I thought of it like "base" is the place where selection is starting from and "extent" is the point where selection ends. If I make both "base" and "extent" same (by assigning deep copy of extent to base), selection will be automatically cleared.

So by following the above thought process, I made the change in clearSelection() method.

To see the effects of my changes in the overall code, I ran all the test cases (test case mentioned in the ticket included) and saw that nothing failed. I took it as my changes didn't effect any of the other feature either and bug mentioned in the ticket also got resolved.

super_editor/lib/src/core/document_composer.dart
mHassan20896: I thought of it like "base" is the place where selection is starting from and "extent" is the point where selection ends. If I make both "base" and "extent" same (by assigning deep copy of extent to base), selection will be automatically cleared.

matthew-carroll:
> To see the effects of my changes in the overall code, I ran all the test cases (test case mentioned in the ticket included) and saw that nothing failed.

This is probably an indication that we didn't have meaningful tests for `clearSelection()`. The purpose of `clearSelection()` is to "remove the selection", so changing the implementation such that the selection isn't removed would seem to be an inappropriate change.

> I found that it went to clearSelection() method of document_composer.dart file where selection variable is assigned to null. It gets into clearSelection() method when we call placeCaretInParagraph() from the test case for the second time.

It's good that you traced the code. That's a good first step. However, any change that we make needs to respect the intended behavior of the API, as mentioned above. 

There are probably multiple areas of code that you could change to solve this problem. The question is, what's the most appropriate change to fix this problem. As far as I can tell, the change in this PR isn't an appropriate change to fix the problem. 

I can't tell you much more without digging into this issue, myself. However, my guess is that the "root problem" is somewhere else. Somewhere, something is happening that shouldn't happen. Or, something isn't happening, when it should. I recommend looking for that place.

mHassan20896:
Hi, I have updated the PR. Kindly check it

mHassan20896:
@matthew-carroll

matthew-carroll:
I'm going to close this PR based on some offline conversation.

Issue #46, opened 2021-03-10
Calculate true line height


matthew-carroll:
`RenderParagraph` does not provide a consistent way to query line height, forcing us to calculate our own value using font size and line spacing. This proxy calculation is not accurate.

Either Flutter needs to update the API to facilitate querying line height at all times, or we need to drop down to a `Paragraph` layout object.

matthew-carroll:
Possibly related to #115

matthew-carroll:
I filed https://github.com/flutter/flutter/issues/107196 with Flutter

PR #707, opened 2022-07-19
[SuperTextField] [Desktop] Fix text behaving scrollable when maxLines is null (Resolves #566)

### LINKED ISSUE #566: [DESKTOP] The text behaves scrollable when the maxLines is null on SuperTextField
From the example application, if you go to 

`_interactive_demo.dart` file and change the example as follows: You can see that the text is cut off in the text field. Same happens in our application as well. 

```

const brandAttribution = NamedAttribution('brand');
const flutterAttribution = NamedAttribution('flutter');

class InteractiveTextFieldDemo extends StatefulWidget {
  @override
  _InteractiveTextFieldDemoState createState() => _InteractiveTextFieldDemoState();
}

class _InteractiveTextFieldDemoState extends State<InteractiveTextFieldDemo> {
  final _textFieldController = AttributedTextEditingController(
    text: AttributedText(
        text: 'Super Editor is an open source text editor for Flutter projects.',
        spans: AttributedSpans(attributions: [
          const SpanMarker(attribution: brandAttribution, offset: 0, markerType: SpanMarkerType.start),
          const SpanMarker(attribution: brandAttribution, offset: 11, markerType: SpanMarkerType.end),
          const SpanMarker(attribution: flutterAttribution, offset: 47, markerType: SpanMarkerType.start),
          const SpanMarker(attribution: flutterAttribution, offset: 53, markerType: SpanMarkerType.end),
        ])),
  );

  OverlayEntry? _popupEntry;
  Offset _popupOffset = Offset.zero;

  FocusNode? _focusNode;

  @override
  void initState() {
    super.initState();
    _focusNode = FocusNode();
  }

  @override
  void dispose() {
    _focusNode!.dispose();
    super.dispose();
  }

  void _onRightClick(
      BuildContext textFieldContext, AttributedTextEditingController textController, Offset localOffset) {
    // Only show menu if some text is selected
    if (textController.selection.isCollapsed) {
      return;
    }

    final overlay = Overlay.of(context)!;
    final overlayBox = overlay.context.findRenderObject() as RenderBox?;
    final textFieldBox = textFieldContext.findRenderObject() as RenderBox;
    _popupOffset = textFieldBox.localToGlobal(localOffset, ancestor: overlayBox);

    if (_popupEntry == null) {
      _popupEntry = OverlayEntry(builder: (context) {
        return GestureDetector(
          behavior: HitTestBehavior.translucent,
          onTapDown: (_) {
            _closePopup();
          },
          child: Container(
            width: double.infinity,
            height: double.infinity,
            color: Colors.transparent,
            child: Stack(
              children: [
                Positioned(
                  left: _popupOffset.dx,
                  top: _popupOffset.dy,
                  child: Container(
                    decoration: BoxDecoration(
                      color: Colors.white,
                      borderRadius: BorderRadius.circular(4),
                      boxShadow: [
                        BoxShadow(
                          color: Colors.black.withOpacity(0.3),
                          blurRadius: 5,
                          offset: const Offset(3, 3),
                        ),
                      ],
                    ),
                    child: Column(
                      children: [
                        TextButton(
                          onPressed: () {
                            Clipboard.setData(ClipboardData(
                              text: textController.selection.textInside(textController.text.text),
                            ));
                            _closePopup();
                          },
                          child: const Text('Copy'),
                        ),
                      ],
                    ),
                  ),
                ),
              ],
            ),
          ),
        );
      });

      overlay.insert(_popupEntry!);
    } else {
      _popupEntry!.markNeedsBuild();
    }
  }

  void _closePopup() {
    if (_popupEntry == null) {
      return;
    }

    _popupEntry!.remove();
    _popupEntry = null;
  }

  @override
    return GestureDetector(
      behavior: HitTestBehavior.opaque,
      onTap: () {
        // Remove focus from text field when the user taps anywhere else.
        _focusNode!.unfocus();
      },
      child: Center(
        child: SizedBox(
          width: 400,
          child: GestureDetector(
            onTap: () {
              // no-op. Prevents unfocus from happening when text field is tapped.
            },
            child: SizedBox(
              width: double.infinity,
              child: SuperDesktopTextField(
                textController: _textFieldController,
                focusNode: _focusNode,
                textStyleBuilder: _textStyleBuilder,
                padding: const EdgeInsets.symmetric(horizontal: 12, vertical: 6),
                decorationBuilder: (context, child) {
                  return Container(
                    decoration: BoxDecoration(
                      borderRadius: BorderRadius.circular(4),
                      border: Border.all(
                        color: _focusNode!.hasFocus ? Colors.blue : Colors.grey.shade300,
                        width: 1,
                      ),
                    ),
                    child: child,
                  );
                },
                hintBuilder: (context) {
                  return const Text(
                    'Leave a message',
                    style: TextStyle(
                      color: Colors.grey,
                    ),
                  );
                },
                minLines: 1,
                maxLines: null,
                onRightClick: _onRightClick,
              ),
            ),
          ),
        ),
      ),
    );
  }

  TextStyle _textStyleBuilder(Set<Attribution> attributions) {
    TextStyle textStyle = const TextStyle(
      color: Colors.black,
      fontSize: 14,
    );

    if (attributions.contains(brandAttribution)) {
      textStyle = textStyle.copyWith(
        color: Colors.red,
        fontWeight: FontWeight.bold,
      );
    }
    if (attributions.contains(flutterAttribution)) {
      textStyle = textStyle.copyWith(
        color: Colors.blue,
      );
    }

    return textStyle;
  }
}

``` 

https://user-images.githubusercontent.com/24432752/168600074-ff4a2f88-0316-4b93-a678-07a91115d4b5.mov

angelosilvestre:
[SuperTextField] [Desktop] Fix text behaving scrollable when maxLines is null. Resolves #566 

We use an estimated line height to calculate the `SuperTextFieldScrollview` viewport height.  Because of this, when the true line height is bigger than the estimated line height, the text is bigger than the viewport, which causes the scrolling behavior.

The root cause of https://github.com/superlistapp/super_editor/issues/565 is the same: when the text is empty, the estimated line height differs from the true line height.

I changed to use the true line height when it's available. However, this might cause the text field to change its height when the user starts typing. If there's a better solution to this I will update the PR.

There's already an open issue to calculate true line height: https://github.com/superlistapp/super_editor/issues/46

### DIFF
--- a/super_editor/test/super_textfield/super_texfield_scroll_test.dart
+++ b/super_editor/test/super_textfield/super_texfield_scroll_test.dart
@@ -68,11 +68,12 @@ void main() {
       expect(textBottom, lessThanOrEqualTo(viewportBottom));
     });
 
-    testWidgetsOnDesktop("with unlimited lines doesn't scroll vertically", (tester) async {
+    testWidgetsOnDesktop("doesn't scroll vertically when maxLines is null", (tester) async {
       // With the Ahem font the estimated line height is equal to the true line height
       // so we need to use a custom font.
       await loadAppFonts();
 
+      // We use some padding because it affects the viewport height calculation.
       const verticalPadding = 6.0;
 
       await tester.pumpWidget(
--- a/super_editor/test/super_textfield/super_texfield_scroll_test.dart
+++ b/super_editor/test/super_textfield/super_texfield_scroll_test.dart
@@ -1,5 +1,6 @@
 import 'package:flutter/material.dart';
 import 'package:flutter_test/flutter_test.dart';
+import 'package:golden_toolkit/golden_toolkit.dart';
 import 'package:super_editor/super_editor.dart';
 import 'package:super_text_layout/super_text_layout.dart';
 
@@ -65,6 +66,49 @@ void main() {
 
       // Ensure the text field scrolled its content vertically
       expect(textBottom, lessThanOrEqualTo(viewportBottom));
+    });
+
+    testWidgetsOnDesktop("with unlimited lines doesn't scroll vertically", (tester) async {
+      // With the Ahem font the estimated line height is equal to the true line height
+      // so we need to use a custom font.
+      await loadAppFonts();
+
+      const verticalPadding = 6.0;
+
+      await tester.pumpWidget(
+        MaterialApp(
+          home: Scaffold(
+            body: ConstrainedBox(
+              constraints: const BoxConstraints(minWidth: 300),
+              child: SuperDesktopTextField(
+                padding: const EdgeInsets.symmetric(horizontal: 12, vertical: verticalPadding),
+                minLines: 1,
+                maxLines: null,
+                textController: AttributedTextEditingController(
+                  text: AttributedText(text: "SuperTextField"),
+                ),
+                textStyleBuilder: (_) => const TextStyle(
+                  fontSize: 14,
+                  height: 1,
+                  fontFamily: 'Roboto',
+                ),
+              ),
+            ),
+          ),
+        ),
+      );
+      await tester.pump();
+
+      final viewportHeight = tester.getRect(find.byType(SuperTextFieldScrollview)).height;
+
+      final layoutState = (find.byType(SuperDesktopTextField).evaluate().single as StatefulElement).state as SuperDesktopTextFieldState;
+      final contentHeight = layoutState.textLayout.getLineHeightAtPosition(const TextPosition(offset: 0));
+
+      // Vertical padding is added to both top and bottom
+      final totalHeight = contentHeight + (verticalPadding * 2);
+
+      // Ensure the viewport is big enough so the text doesn't scroll vertically
+      expect(viewportHeight, greaterThanOrEqualTo(totalHeight));
     });
   });
 }
--- a/super_editor/test/super_textfield/super_texfield_scroll_test.dart
+++ b/super_editor/test/super_textfield/super_texfield_scroll_test.dart
@@ -100,6 +100,23 @@ void main() {
       );
       await tester.pump();
 
+      // The original issue was that the textfield was calculating a viewport height 
+      // that wasn't big enough to display the text, causing it to be scrollable.
+      // The textfield sizes it's viewport based on a estimated line height, so if the
+      // true line height differs from the estimated one, the issue happens.
+      //
+      // Ideally we should test if the textfield's ScrollController has a zero maxScrollExtent,
+      // to prove that it's not scrollable.
+      // However, I couldn't reproduce a failing test with this expectation.
+      // For example: when running the app, the SuperText inside the textfield with a 
+      // fontSize of 14 is 19 pixels tall.
+      // Running the test it is 14 pixels tall, even loading the app fonts.
+      // Because of this, SuperText fits inside the textfield's Scrollview and the scrolling
+      // issue doesn't happen.
+      //
+      // This test only ensures that the viewport height isn't too small, checking if a
+      // calculated line height plus padding fits inside the viewport.
+
       final viewportHeight = tester.getRect(find.byType(SuperTextFieldScrollview)).height;
 
       final layoutState = (find.byType(SuperDesktopTextField).evaluate().single as StatefulElement).state as SuperDesktopTextFieldState;
--- a/super_editor/test/super_textfield/super_texfield_scroll_test.dart
+++ b/super_editor/test/super_textfield/super_texfield_scroll_test.dart
@@ -100,22 +100,11 @@ void main() {
       );
       await tester.pump();
 
-      // The original issue was that the textfield was calculating a viewport height 
-      // that wasn't big enough to display the text, causing it to be scrollable.
-      // The textfield sizes it's viewport based on a estimated line height, so if the
-      // true line height differs from the estimated one, the issue happens.
-      //
-      // Ideally we should test if the textfield's ScrollController has a zero maxScrollExtent,
-      // to prove that it's not scrollable.
-      // However, I couldn't reproduce a failing test with this expectation.
-      // For example: when running the app, the SuperText inside the textfield with a 
-      // fontSize of 14 is 19 pixels tall.
-      // Running the test it is 14 pixels tall, even loading the app fonts.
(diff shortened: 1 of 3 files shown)


matthew-carroll (commented):
Is it possible to test this? Can you write a test that fails without this change, and then passes with this change?

angelosilvestre:
> Is it possible to test this? Can you write a test that fails without this change, and then passes with this change?

@matthew-carroll To test this we can't use the Ahem font, because it seems that with Ahem the estimated line height is equal to the true line height. To be able to test this we need to use another font, like Roboto.

angelosilvestre:
@matthew-carroll I'm asking if I'm allowed to use the Roboto font in the tests, or if it will cause any sort of problem in the CI.

matthew-carroll:
We can try it and find out. Maybe we'll have to include the font in test assets or something. Just make sure that you return the test font configuration to what it was at the end of the test. And make sure that even when the test fails, the font config gets reset.

If you're wondering how to activate a font, you can check `golden_toolkit`. They have a method that's called something like `loadAppFonts()`. We probably don't want to call that method, because I'm not sure there's a method to undo it. But you can probably find out how to enable a font by checking their code.

angelosilvestre:
@matthew-carroll There are already other places where we call `loadAppFonts` with this comment:

```dart
// We load fonts, even for non-golden tests, so that text layout and
// line-wrapping is more predictable.
await loadAppFonts();
```
Is it OK to call it in this test?

Also, I'm not sure if this file is the right place for this test. Do you think `super_texfield_scroll_test.dart` is a better place for it?

matthew-carroll:
> Is it OK to call it in this test?

It's OK, as long as you find a way to undo it. The problem is that, I think, any tests that run after you call that method will continue to use the app fonts. If you can prove that's not the case, then it's OK to call that without undoing it. However, if it looks like the change is permanent during the test run, then we should undo the change at the end of the test.

Other places that call that method in non-golden tests should probably be undoing the change, too.

> Also, I'm not sure if this file is the right place for this test. Do you think `super_texfield_scroll_test.dart` is a better place for it?

Yeah, that seems like a reasonable place to put it.

angelosilvestre:
@matthew-carroll I'm not sure if there is a method for unloading a font, but I will see if I can find one.

angelosilvestre:
@matthew-carroll It seems it's not possible to unload a font. I tried reloading the Roboto font family using the Ahem font file, but it seems it doesn't override a font that is already loaded. 

One option is to load Roboto font with another font family name (like RobotoTest) and use it only in the tests where the custom font is needed.

Do you have another suggestion?

matthew-carroll:
"font loading" was probably the wrong thing to worry about. Rather, we want Flutter to go back to using the Ahem font. Can you find where Flutter makes the decision to use Ahem?

angelosilvestre:
@matthew-carroll At least with `SuperTextField`, if we don't specify a `fontFamily` in the `textStyle`, it will not pick the default font from the theme, so Flutter will create a Paragraph without a `fontFamily` set. In this case, Ahem will be chosen directly on the engine side.

If we call `loadAppFonts` but don't specify Roboto as the `fontFamily`, Flutter engine will still pick Ahem.

In many of the Flutter framework tests, Ahem is set manually as the `fontFamily`.

matthew-carroll:
Ok, we can move forward with `loadAppFonts()`. Can you let me know which tests are new? GitHub is showing almost everything as changed, because it looks like the grouping was altered. It'd be helpful to know what I should review, rather than try to match the old with the new.

angelosilvestre:
The new tests is in `group('with unlimited lines'`, I still need to move the test to `super_texfield_scroll_test.dart`.

angelosilvestre:
@matthew-carroll I moved the test to the other file

super_editor/test/super_textfield/super_texfield_scroll_test.dart
```diff
@@ -66,6 +67,49 @@ void main() {
       // Ensure the text field scrolled its content vertically
       expect(textBottom, lessThanOrEqualTo(viewportBottom));
     });
+
+    testWidgetsOnDesktop("with unlimited lines doesn't scroll vertically", (tester) async {
```
matthew-carroll, line 71: This test name doesn't make sense to me. Can you rephrase this?

super_editor/test/super_textfield/super_texfield_scroll_test.dart
```diff
@@ -66,6 +67,49 @@ void main() {
...
+      // With the Ahem font the estimated line height is equal to the true line height
+      // so we need to use a custom font.
+      await loadAppFonts();
+
+      const verticalPadding = 6.0;
```
matthew-carroll, line 76: Please comment why we care about this.

super_editor/test/super_textfield/super_texfield_scroll_test.dart
```diff
@@ -66,6 +67,49 @@ void main() {
...
+            ),
+          ),
+        ),
+      );
+      await tester.pump();
+
+      final viewportHeight = tester.getRect(find.byType(SuperTextFieldScrollview)).height;
+
+      final layoutState = (find.byType(SuperDesktopTextField).evaluate().single as StatefulElement).state as SuperDesktopTextFieldState;
+      final contentHeight = layoutState.textLayout.getLineHeightAtPosition(const TextPosition(offset: 0));
```
matthew-carroll, line 105: Ideally, we should avoid digging deep into widget trees like this. The problem we're solving is that the field was scrollable, when it shouldn't be, right? If so, wouldn't that situation be clear by checking the min and max scroll offsets to ensure that they're both `0.0`? If so, I think you could add methods to the inspector that reports back with the min and max scroll offsets.

super_editor/test/super_textfield/super_texfield_scroll_test.dart
angelosilvestre: I changed the test description

super_editor/test/super_textfield/super_texfield_scroll_test.dart
angelosilvestre: Comment added.

matthew-carroll (approved):
LGTM

PR #708, opened 2022-07-20
[SuperEditor] Fix scrolling issue when the editor is inside a horizontal Scrollable (Resolves #501)

### LINKED ISSUE #501: Problem using the SuperEditor with a TabBar
When adding the editor to a TabBarView once the screen is filled with text the TabBarView automatically changes to the next tab rather than scroll down in the editor. 

A related problem is when the editor is used on mobile in a TabBarView if the keyboard is shown it causes the tab bar to change to the previous tab.

Here's a simple example:

```
  @override
    return Scaffold(
      appBar: AppBar(
        bottom: TabBar(
          controller: _controller,
          tabs: [
            Tab(
              text: 'Tab 1',
            ),
            Tab(
              text: 'Tab 2',
            ),
          ],
        ),
      ),
      body: TabBarView(
        controller: _controller,
        children: [
          ExampleEditor(),
          ExampleEditor(),
        ],
      ),
    );
```

angelosilvestre:
[SuperEditor] Fix scrolling issue when the editor is inside a horizontal `Scrollable`. Resolves #501

When adding the editor to a `TabBarView`, once the screen is filled with text, the `TabBarView` automatically changes to the next tab rather than scroll down in the editor.

A similar issue happens when adding the editor inside a horizontal `ListView`, where adding text causes the `ListView` to scroll.

The problem is that when `SuperEditor` has an ancestor `Scrollable`, it doesn't create its own `SingleChildScrollView` and scrolls the ancestor `Scrollable` instead. Because of that, widgets that use a horizontal `Scrollable` are affected.

I changed it to ignore the ancestor `Scrollable` if it has a horizontal `AxisDirection`.

I also added an option in `TestDocumentConfigurator` to configure the subtree between `MaterialApp` and `SuperEditor`.

### DIFF
--- a/super_editor/lib/src/default_editor/document_scrollable.dart
+++ b/super_editor/lib/src/default_editor/document_scrollable.dart
@@ -44,6 +44,9 @@ class DocumentScrollable extends StatefulWidget {
 
   /// The [ScrollController] that governs this [DocumentScrollable]'s scroll
   /// offset.
+  ///
+  /// `scrollController` is not used if this `SuperEditor` has an ancestor
+  /// `Scrollable`.
   final ScrollController? scrollController;
 
   /// This widget's child, which should include a document.
@@ -99,7 +102,7 @@ class _DocumentScrollableState extends State<DocumentScrollable> with SingleTick
       if (oldWidget.scrollController == null) {
         _scrollController.dispose();
       }
-      _scrollController = widget.scrollController ?? ScrollController();           
+      _scrollController = widget.scrollController ?? ScrollController();
     }
 
     if (widget.autoScroller != oldWidget.autoScroller) {
@@ -120,8 +123,8 @@ class _DocumentScrollableState extends State<DocumentScrollable> with SingleTick
     // if (widget.scrollingMinimapId == null) {
     //   ScrollingMinimaps.of(context)?.put(widget.scrollingMinimapId!, null);
     // }
-    
-    if (widget.scrollController == null){
+
+    if (widget.scrollController == null) {
       _scrollController.dispose();
     }
 
@@ -137,8 +140,7 @@ class _DocumentScrollableState extends State<DocumentScrollable> with SingleTick
   /// If this widget doesn't have an ancestor `Scrollable`, then this
   /// widget includes a `ScrollView` and this `State`'s render object
   /// is the viewport `RenderBox`.
-  RenderBox get _viewport =>
-      (_findAncestorScrollable(context)?.context.findRenderObject() ?? context.findRenderObject()) as RenderBox;
+  RenderBox get _viewport => (_findAncestorScrollable(context)?.context.findRenderObject() ?? context.findRenderObject()) as RenderBox;
 
   /// Returns the `ScrollPosition` that controls the scroll offset of
   /// this widget.
@@ -159,8 +161,8 @@ class _DocumentScrollableState extends State<DocumentScrollable> with SingleTick
     }
 
     final direction = ancestorScrollable.axisDirection;
-    // If the direction is horizontal, then we are inside a widget like a TabBar 
-    // or a horizontal ListView, so we can't use the ancestor scrollable 
+    // If the direction is horizontal, then we are inside a widget like a TabBar
+    // or a horizontal ListView, so we can't use the ancestor scrollable
     if (direction == AxisDirection.left || direction == AxisDirection.right) {
       return null;
     }
--- a/super_editor/lib/src/default_editor/document_scrollable.dart
+++ b/super_editor/lib/src/default_editor/document_scrollable.dart
@@ -27,6 +27,7 @@ class DocumentScrollable extends StatefulWidget {
     required this.autoScroller,
     this.scrollingMinimapId,
     this.showDebugPaint = false,
+    this.scrollController,
     required this.child,
   }) : super(key: key);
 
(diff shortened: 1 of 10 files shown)


super_editor/test/super_editor/document_test_tools.dart
```diff
@@ -142,6 +143,14 @@ class TestDocumentConfigurator {
     return this;
   }
 
+  /// Configures the [SuperEditor] to use a custom subtree between [MaterialApp] and [SuperEditor].
+  /// 
+  /// By default, [SuperEditor] is displayed inside a [Scaffold].
+  TestDocumentConfigurator withCustomSubtree(SubtreeBuilder? builder) {
```
matthew-carroll, line 149: I'm thinking that we might as well have this builder build everything above `SuperEditor`. As such I think we could call this method `withCustomWidgetTreeBuilder()`. I think the given `builder` can be a standard `WidgetBuilder`.

super_editor/test/super_editor/supereditor_ancestor_scrollable_test.dart
```diff
@@ -0,0 +1,122 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:super_editor/super_editor.dart';
+
+import '../test_tools.dart';
+import 'document_test_tools.dart';
+import 'supereditor_robot.dart';
+
+void main() {
+  group('SuperEditor', () {
```
matthew-carroll, line 10: 'SuperEditor' -> 'SuperEditor respects horizontal scrolling'

super_editor/test/super_editor/supereditor_ancestor_scrollable_test.dart
```diff
@@ -0,0 +1,122 @@
...
+                  ),
+                ),
+              ),
+            )
+            .pump();
+
+        // Ensure SuperEditor added its own Scrollview.
+        // If the Scrollview wasn't added, the content will overflow
+        // the editor bounds.
+        expect(find.byType(SingleChildScrollView), findsOneWidget);
```
matthew-carroll, line 50: I think this is an implementation detail that we don't need to expose in the test. The question is, what user behavior was previously breaking, and can we now execute that use behavior as expected. Testing the user behavior automatically verifies our implementation details, while also focusing on what matters to the user.

super_editor/test/super_editor/supereditor_ancestor_scrollable_test.dart
```diff
@@ -0,0 +1,122 @@
...
+
+        // Select the editor.
+        await tester.placeCaretInParagraph('1', 0);
+
+        // Add new lines so the content will cause editor to scroll
+        await _addNewLines(tester, count: 20);
```
matthew-carroll, line 56: Given that Flutter can change the size of our tests without us knowing, it's probably a good idea to make sure that the scroll offset is greater than zero.

super_editor/test/super_editor/document_test_tools.dart
angelosilvestre: I'm not sure if we can get a `context` to pass as an argument to `WidgetBuilder`

super_editor/test/super_editor/supereditor_ancestor_scrollable_test.dart
angelosilvestre: I added this expectation because, without the fix, adding lines to the document causes the test to fail due to renderbox overflow.  The later expectations already ensures the correct behavior.

Should I remove it?

angelosilvestre:
@matthew-carroll I changed `SuperEditor` to pass its `scrollController` to `DocumentScrollable` because it was being ignored.

super_editor/test/super_editor/supereditor_ancestor_scrollable_test.dart
matthew-carroll: Yeah, let's remove it.

super_editor/lib/src/default_editor/document_scrollable.dart
```diff
@@ -41,6 +42,10 @@ class DocumentScrollable extends StatefulWidget {
   /// debugging, when `true`.
   final bool showDebugPaint;
 
+  /// The [ScrollController] that governs this [DocumentScrollable]'s scroll
+  /// offset.
```
matthew-carroll, line 46: This controller isn't used when there's an ancestor `Scrollable`, right? If so, you should probably mention that here, and also mention it in `SuperEditor`'s docs for its `ScrollController`.

angelosilvestre: I added the comment in `DocumentScrollable`. It was already present in `SuperEditor`

super_editor/test/super_editor/supereditor_ancestor_scrollable_test.dart
```diff
@@ -0,0 +1,124 @@
...
+      await tester
+          .createDocument()
+          .withSingleEmptyParagraph()
+          .withInputSource(DocumentInputSource.ime)
+          .withScrollController(scrollController)
+          .withCustomWidgetTreeBuilder(
+            (superEditor) => MaterialApp(
+              home: ConstrainedBox(
+                constraints: const BoxConstraints(
+                  minWidth: 300,
```
matthew-carroll, line 24: Can you add a comment about why we need such a tiny editor for the test?

matthew-carroll (approved):
LGTM - I left one last request for a clarifying comment.

super_editor/test/super_editor/supereditor_ancestor_scrollable_test.dart
angelosilvestre: Done.

PR #716, opened 2022-07-25
[SuperTextField] Fix horizontal aligment (Resolves #668)

### LINKED ISSUE #668: [SuperTextField] - Horizontal alignment is broken
As seen in https://github.com/superlistapp/super_editor/issues/509, horizontal alignment for `SuperTextField` seems to be broken. Let's fix it.

angelosilvestre:
[SuperTextField] Fix horizontal aligment. Resolves #668 

Horizontal alignment was not working, as shown in https://github.com/superlistapp/super_editor/issues/509

There are two different situations: 

**Multiline text fields**: The  `RenderParagraph` has a `maxWidth` constraint. In Flutter `TextPainter`, when `maxWidth` differs from `minWidth`, it layouts the paragraph again using `maxIntrinsicWidth` as `maxWidth`, so the `RenderParagraph` doesn't use all the available width and the alignment doesn't work as expected. As a result, the scrollbar is displayed next to the text instead of being displayed at the right edge of the text field.

The relevant lines are here:

https://github.com/flutter/flutter/blob/ca6cecf034d9c18fcb6ea70309cc855e353aa0f8/packages/flutter/lib/src/painting/text_painter.dart#L630-L652

I wrapped `SuperText` with a `SizedBox` with infinity `maxWidth`, so it forces the `RenderParagraph` to use all available width.

**Singleline text fields**: The text field has a horizontal scrollview, which gives an infinity `maxWidth` to `RenderParagraph`. In this case, wrapping with the `SizedBox` won't help.

I wrapped the `SingleChildScrollView` with an `Align` widget, with an alignment based on the text alignment.


matthew-carroll:
@angelosilvestre I didn't quite follow the explanation in the PR. Specifically, I'm wondering why alignment used to work, but it doesn't any more. My guess is that the `super_text_layout` work broke it. But that still begs the question: what was the change that broke it, and if that change was in `super_text_layout`, should we change `super_text_layout` instead of adding more widgets to `SuperTextField`?

angelosilvestre:
@matthew-carroll The main problem is that, to be able to respect the text alignment, `SuperText` needs to take all available width. Instead, it takes only its `maxIntrinsicWidth` (this is done in `TextPainter`). 

Because of this, the text is aligned only within `SuperText`'s bounds, and not within `SuperTextField`'s bounds.

The alignment stopped working in b2e0dc7cf78e4c83d740077b174d5987b8ef0309, although `center` alignment of a single line text field wasn't working before that.

The text was being displayed inside a `_FillWidthIfConstrained`. The alignment was working because, with this widget, the text was taking all the available width.

In b2e0dc7cf78e4c83d740077b174d5987b8ef0309, the text was wrapped with a `IntrinsicWidth`, which caused the text to stop taking all the available width.

I can try to avoid using the `Align` widget, but I think we still need to use a `SizedBox` or reimplement `_FillWidthIfConstrained`.

matthew-carroll:
Let's bring back `_FillWidthIfConstrained` along with a doc comment that explains why we need it. I probably deleted it because I couldn't remember why we were using it, and we didn't have any tests that failed.

super_editor/lib/src/infrastructure/super_textfield/android/android_textfield.dart
```diff
@@ -275,6 +276,8 @@ class SuperAndroidTextFieldState extends State<SuperAndroidTextField> with Ticke
 
   bool get _isMultiline => (widget.minLines ?? 1) != 1 || widget.maxLines != 1;
 
+  bool get _isBounded => widget.maxLines != null;
```
matthew-carroll, line 279: This name is too generic. What's bounded? Width? Height? Line count? Character length? What is this property really being used to specify?

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -1123,21 +1131,53 @@ class SuperTextFieldScrollviewState extends State<SuperTextFieldScrollview> with
...
+          child: Padding(
+            padding: widget.padding,
+            child: widget.child,
+          ),
         ),
       ),
     );
   }
+
+  Widget _buildWrapper(Widget child) {
```
matthew-carroll, line 1168: We should never call something a "wrapper" - this doesn't mean anything. What is the wrapper doing?

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -1123,21 +1131,53 @@ class SuperTextFieldScrollviewState extends State<SuperTextFieldScrollview> with
...
+    // Multiline textfields are already aligned correctly
+    // because they have a maxWidth constraint.
+    if (widget.isMultiline) {
+      return child;
+    }
+    // Singleline textfields have an infinity maxWidth
+    // so we need to align the whole scrollview
+    return Align(
```
matthew-carroll, line 1176: In this case of a single-line, is a `ScrollView` the only situation where this problem occurs? If so, could you change `FillWidthIfConstrained` to check for an ancestor `Scrollable` when the width is unconstrained? You can get a width from the viewport.

super_editor/lib/src/infrastructure/super_textfield/android/android_textfield.dart
angelosilvestre: After the last commits this property isn't used anymore. I will remove it.

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
angelosilvestre: I renamed to make it more clear.

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
angelosilvestre: The issue isn't caused by an ancestor `Scrollable`. It's the `SingleChildScrollView` that's added by `SuperTextFieldScrollview` itself that gives an infinity max width to the super text .

matthew-carroll: Isn't the `FillWidthIfConstrained` lower in the tree than the `SingleChildScrollView`? If so, then the `SingleChildScrollView` is an ancestor `Scrollable` from the perspective of `FillWidthIfConstrained`, right?

matthew-carroll:
@angelosilvestre it looks like this PR has conflicts

angelosilvestre:
@matthew-carroll I resolved the conflits.

super_editor/lib/src/infrastructure/super_textfield/infrastructure/fill_width_if_constrained.dart
```diff
@@ -0,0 +1,74 @@
...
+  RenderFillWidthIfConstrained({
+    double? viewportWidth,
+  }) : _viewportWidth = viewportWidth;
+
+  set viewportWidth(double? value) {
+    _viewportWidth = value;
+    markNeedsLayout();
+  }
+
+  double? _viewportWidth = 0;
```
matthew-carroll, line 46: Is there a reason that this property is given an initial value of `0`? It looks like the constructor always assigns an initial value.

super_editor/lib/src/infrastructure/super_textfield/infrastructure/fill_width_if_constrained.dart
```diff
@@ -0,0 +1,74 @@
...
+    return (scrollable.context.findRenderObject() as RenderBox?)?.size.width;
+  }
+}
+
+class RenderFillWidthIfConstrained extends RenderProxyBox {
+  RenderFillWidthIfConstrained({
+    double? viewportWidth,
+  }) : _viewportWidth = viewportWidth;
+
+  set viewportWidth(double? value) {
```
matthew-carroll, line 41: Please add a dart doc that explains what viewport this is, and why we care.

super_editor/lib/src/infrastructure/super_textfield/infrastructure/fill_width_if_constrained.dart
```diff
@@ -0,0 +1,74 @@
...
+    // If the available width is bounded, 
+    // force the child to be as wide as the available width.
+    if (constraints.hasBoundedWidth) {
+      childConstraints = BoxConstraints(
+        minWidth: constraints.maxWidth,
+        minHeight: constraints.minHeight,
+        maxWidth: constraints.maxWidth,
+        maxHeight: constraints.maxHeight,
+      );
+    } else if (_viewportWidth != null) {
```
matthew-carroll, line 61: Do we care whether the viewport scrolls vertically vs horizontally?

super_editor/lib/src/infrastructure/super_textfield/infrastructure/fill_width_if_constrained.dart
angelosilvestre: This initialization isn't needed. I removed it.

super_editor/lib/src/infrastructure/super_textfield/infrastructure/fill_width_if_constrained.dart
angelosilvestre: I also changed the property name to make it more clear.

super_editor/lib/src/infrastructure/super_textfield/infrastructure/fill_width_if_constrained.dart
angelosilvestre: We only care about horizontal scrollables.

matthew-carroll (approved):
LGTM

PR #719, opened 2022-08-02
[SuperEditor] Fix selection on focus (Resolves #376)

### LINKED ISSUE #376: requestFocus does not show caret in editor when no selection is present

angelosilvestre:
[SuperEditor] Fix selection on focus. Resolves https://github.com/superlistapp/super_editor/issues/376

Currently, giving focus to `SuperEditor` by pressing `tab` or requesting focus to the attached `FocusNode` doesn't cause the editor to be selected. Because of this, we can't press `tab` and begin typing unless we tap at the editor.

This PR changes `SuperEditor` to move the selection to the end of the last selectable node when the editor gets focused, or to restore the last valid selection, if available.


super_editor/lib/src/default_editor/document_gestures_mouse.dart
```diff
@@ -521,6 +538,60 @@ Updating drag selection:
...
+      // DocumentLayout before it is available when the editor has autofocus
+      WidgetsBinding.instance.addPostFrameCallback((timeStamp) {
+        if (mounted && _focusNode.hasFocus && widget.editContext.composer.selection == null) {
+          _moveSelectionAfterFocus();
+        }
+      });
+    }
+  }
+
+  DocumentSelection? _findNewSelectionAfterFocus() {
```
matthew-carroll, line 556: This seems like a strange thing for a gesture interactor to implement, and it's repeated. Let see what it looks like to move this into `DocumentLayout`. Let's call it `findLastSelectablePosition()` and let's have it return a `DocumentPosition`.

super_editor/lib/src/default_editor/document_gestures_mouse.dart
```diff
@@ -521,6 +538,60 @@ Updating drag selection:
     widget.editContext.composer.clearSelection();
   }
 
+  void _onFocusChange() {
+    final shouldMoveSelection = !_requestingFocusByTap;
```
matthew-carroll, line 542: I'd really like to avoid this nuanced dance where we momentarily track whether a tap occurred. It's difficult to comprehend when a developer reads the code, and it's fragile because other gestures, like a long-press, aren't automatically covered. So if new selection gestures are introduced, it's likely that we'll forget to update this.

Is there another way that we can solve this?

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
```diff
@@ -107,6 +107,12 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
   Offset? _dragEndInInteractor;
   SelectionType? _selectionType;
 
+  // Indicates if we are requesting focus by tapping the editor.
+  bool _requestingFocusByTap = false;
+
+  // Holds the last valid selection, so we can restore it when the editor is focused.
+  DocumentSelection? _lastValidSelection;
```
matthew-carroll, line 114: We're repeating most of this behavior 3 times. Can you investigate the possibility to creating a single mixin that we could add to all of these gesture interactors so we can avoid 3 implementations of the same rules?

super_editor/test/super_editor/supereditor_selection_test.dart
```diff
@@ -257,6 +257,335 @@ void main() {
         ),
       );
     });
+
+    testWidgetsOnAllPlatforms("selects last selectable node when focusing by keyboard", (tester) async {
```
matthew-carroll, line 261: I think these test names are too focused on implementation details, rather than user behaviors.

It would be more meaningful to say "places caret at end of document upon first editor focus with tab". Then there would be a test that says something about restoring the previous selection. And then, for the unusual case of unselectable components, there would be a test that mentions something about selectable components. But that case is the most unusual, so we shouldn't frame all of our tests in terms of that edge-case.

super_editor/test/super_editor/supereditor_selection_test.dart
```diff
@@ -257,6 +257,335 @@ void main() {
...
+        DocumentSelection.collapsed(
+          position: DocumentPosition(
+            nodeId: secondParagraphNodeId,
+            nodePosition: const TextNodePosition(offset: 16),
+          ),
+        ),
+      );
+    });
+
+    testWidgetsOnAllPlatforms("retains a valid selection when focusing by keyboard", (tester) async {
```
matthew-carroll, line 425: Words like "valid" should rarely be used. Those words almost always mean other things. We should instead say the other things that we mean. What is a "valid" selection? What's an "invalid" selection? Why does this test matter? Let's find a more user-centric and obvious description of what we're testing here.

super_editor/test/super_editor/supereditor_selection_test.dart
angelosilvestre: I updated the tests

super_editor/test/super_editor/supereditor_selection_test.dart
angelosilvestre: updated the tests

super_editor/lib/src/default_editor/document_gestures_mouse.dart
angelosilvestre: In fact, the only situation where we need this is when tapping at an unselectable component, as it's the only situation where we focus by tapping but without setting the selection.

There is a test that ensures the selection is null when tapping at an unselectable component: "Unselectable component rejects selection when user taps on it".

This behavior should be maintained or it should also place the caret at the end of the document?

super_editor/lib/src/default_editor/document_interactor_mixin.dart
```diff
@@ -0,0 +1,100 @@
...
+
+  // Indicates if we are requesting focus by tapping the editor.
+  bool requestingFocusByTap = false;
+
+  FocusNode? _focusNode;
+  DocumentComposer? _composer;
+  DocumentLayout Function()? _getDocumentLayout;
+
+  /// Initializes the mixin internal state. Should be called at `initState`.
+  void initDocumentInteractorMixin({
```
matthew-carroll, line 17: Rather than view this mixin from the perspective of a `State` lifecycle, think about the behavior that the mixin provides, and name the public methods based on what you're asking the mixin to do. This method appears to register a couple listeners, and those listeners cause things to happen. This method should probably say that it "starts" whatever that behavior is.

super_editor/lib/src/default_editor/document_interactor_mixin.dart
```diff
@@ -0,0 +1,100 @@
...
+  }) {
+    _focusNode = focusNode;
+    _focusNode!.addListener(_onFocusChange);
+    _composer = composer;
+    _composer!.selectionNotifier.addListener(_onSelectionChange);
+    _getDocumentLayout = getDocumentLayout;
+  }
+
+  /// Updates the mixin internal state. Should be called at `didUpdateWidget`.
+  void updateDocumentInteractorMixin({
```
matthew-carroll, line 30: Rather than strongly tie this mixin to the `State` lifecycle, consider providing individual methods to change the `FocusNode`, `DocumentComposer`, and the document layout resolver. 

For example: `onFocusNodeReplaced(newNode)`

BTW, isn't there a typedef for that layout function, instead of explicitly declaring the function signature?

super_editor/lib/src/default_editor/document_interactor_mixin.dart
```diff
@@ -0,0 +1,100 @@
...
+    if (_composer != composer){
+      _composer?.removeListener(_onSelectionChange);
+      _composer = composer;
+      _composer!.addListener(_onSelectionChange);
+    }
+    _getDocumentLayout = getDocumentLayout;
+  }
+
+  @override
+  void dispose() {
```
matthew-carroll, line 49: This method could be called "stop" something, as the inverse to the method that starts the behavior. That way, a client can make its own decision about when this behavior starts and stops.

super_editor/lib/src/default_editor/document_interactor_mixin.dart
```diff
@@ -0,0 +1,100 @@
...
+    if (position == null) {
+      return null;
+    }
+
+    return DocumentSelection.collapsed(
+      position: position,
+    );
+  }
+
+  void _moveSelectionAfterFocus() {
```
matthew-carroll, line 94: The name of this function implies an order of operations that isn't necessary. This method can simply be called `_moveCaretToEndOfDocument()`.

super_editor/lib/src/default_editor/document_interactor_mixin.dart
```diff
@@ -0,0 +1,100 @@
...
+    });
+  }
+
+  void _onSelectionChange() {
+    if (_composer?.selection != null) {
+      _previousSelection = _composer?.selection;
+    }
+  }
+
+  DocumentSelection? _findNewSelectionAfterFocus(DocumentLayout? layout) {
```
matthew-carroll, line 78: I think that this implementation can be moved into the other method

super_editor/lib/src/default_editor/document_interactor_mixin.dart
```diff
@@ -0,0 +1,100 @@
...
+    WidgetsBinding.instance.addPostFrameCallback((timeStamp) {
+      if (mounted && _focusNode!.hasFocus && _composer!.selection == null) {
+        _moveSelectionAfterFocus();
+      }
+    });
+  }
+
+  void _onSelectionChange() {
+    if (_composer?.selection != null) {
+      _previousSelection = _composer?.selection;
```
matthew-carroll, line 74: I think we should add a comment about why we're storing the current selection.

super_editor/lib/src/default_editor/document_interactor_mixin.dart
```diff
@@ -0,0 +1,100 @@
...
+    requestingFocusByTap = false;
+
+    if (!_focusNode!.hasFocus || !shouldMoveSelection) {
+      return;
+    }
+
+    // We move the selection in the next frame, so we don't try to access the
+    // DocumentLayout before it is available when the editor has autofocus
+    WidgetsBinding.instance.addPostFrameCallback((timeStamp) {
+      if (mounted && _focusNode!.hasFocus && _composer!.selection == null) {
```
matthew-carroll, line 66: It would be good to add a comment about the significance of these conditions that we're checking. What does it mean to have focus and a null selection? Why do we care?

super_editor/lib/src/default_editor/document_interactor_mixin.dart
```diff
@@ -0,0 +1,100 @@
+import 'package:flutter/material.dart';
+import 'package:super_editor/super_editor.dart';
+
+/// Mixin for common operations on document interactors
+mixin DocumentInteractorMixin<T extends StatefulWidget> on State<T> {
+  // Holds the last selection, so we can restore it when the editor is re-focused.
+  DocumentSelection? _previousSelection;
+
+  // Indicates if we are requesting focus by tapping the editor.
+  bool requestingFocusByTap = false;
```
matthew-carroll, line 10: I didn't understand your previous comment about when this flag is needed vs not needed. Can you elaborate?

super_editor/lib/src/default_editor/document_interactor_mixin.dart
angelosilvestre: Wouldn't it be confusing to call it `_moveCaretToEndOfDocument` as it can restore the last selection?

super_editor/lib/src/default_editor/document_interactor_mixin.dart
angelosilvestre: When the user taps at an unselectable component, the editor is focused, but we don't set any selection. So, when focus changes we need a way to know if we should update the selection or not.

super_editor/lib/src/default_editor/document_gestures_mouse.dart
```diff
@@ -79,22 +80,31 @@ class _DocumentMouseInteractorState extends State<DocumentMouseInteractor> with
     _focusNode = widget.focusNode ?? FocusNode();
     widget.editContext.composer.selectionNotifier.addListener(_onSelectionChange);
     widget.autoScroller.addListener(_updateDragSelection);
+
+    startUpdateSelectionOnFocus(
```
matthew-carroll, line 84: Let's call this `startSyncingSelectionWithFocus`

super_editor/lib/src/default_editor/document_selection_on_focus_mixin.dart
```diff
@@ -0,0 +1,98 @@
+import 'package:flutter/material.dart';
+import 'package:super_editor/super_editor.dart';
+
+/// Provides the functionality of updating the selection when the editor is focused.
```
matthew-carroll, line 4: This first sentence isn't true. Almost all "functionality of updating the selection" that happens while "the editor is focused" is implemented elsewhere.

super_editor/lib/src/default_editor/document_selection_on_focus_mixin.dart
```diff
@@ -0,0 +1,98 @@
...
+/// 
+/// It moves the caret to the end of the document or restores the last selection,
```
matthew-carroll, line 6: I think that this description still lacks context. First, I'd try not to describe behavior in terms of the class, itself, e.g., "It moves the caret...". Moreover, this description doesn't mention any preconditions, such as the editor previously being selected, or not, or the fact that this behavior is triggered when the editor receives focus. Nor does this description mention that there are critical lifecycle methods that must be called for this behavior to work as expected. Please try to view this mixin from the perspective of someone who has no idea what it does, and then provide a gentle, but concise explanation for why it exists and what it does.

super_editor/lib/src/default_editor/document_selection_on_focus_mixin.dart
```diff
@@ -0,0 +1,98 @@
...
+    _composer!.addListener(_onSelectionChange);
+  }
+
+  /// Should be called whenever the [DocumentLayoutResolver] is replaced.
+  void onDocumentLayoutResolverReplaced(DocumentLayoutResolver? layoutResolver) {
+    _getDocumentLayout = layoutResolver;
+  }
+
+  /// Stops the behavior of updating the selection when the editor is focused.
+  void stopUpdateSelectionOnFocus() {
```
matthew-carroll, line 52: Please move this directly below the start method so that it's easy to match setup behavior with cleanup behavior.

super_editor/lib/src/default_editor/document_selection_on_focus_mixin.dart
```diff
@@ -0,0 +1,98 @@
...
+      // We only update the selection when it's null
+      // because, when the user taps at the document the selection is
+      // already set to the correct position, so we don't override it.
+      if (mounted && _focusNode!.hasFocus && _composer!.selection == null) {
+        _updateDocumentSelection();
+      }
+    });
+  }
+
+  void _onSelectionChange() {
```
matthew-carroll, line 77: In the tap callbacks where you're currently setting `requestingFocusByTap`, isn't this callback triggered before the interactors request focus? If so, couldn't we remove `requestingFocusByTap` and still get the desired behavior, because this mixin would think that the `_perviousSelection` is the new selection? In other words, this mixin would set the selection, but the mixin would set it to the same value it already has, giving us the desired result...

super_editor/lib/src/default_editor/document_selection_on_focus_mixin.dart
```diff
@@ -0,0 +1,98 @@
...
+    // We store the last selection so the next time the editor is focused
+    // the selection is restored.
+    if (_composer?.selection != null) {
+      _previousSelection = _composer?.selection;
+    }
+  }
+
+  void _updateDocumentSelection() {
```
matthew-carroll, line 85: This method looks like it's only called from `_onFocusChange()`, so it should be declared directly beneath `_onFocusChange()` for easy reference.

matthew-carroll: In fact, I think the code would probably be clearer by simply moving this implementation into `_onFocusChange()`. The name `_updateDocumentSelection()` doesn't really describe what this method does, so readers will probably have to jump down and read the implementation, anyway.

super_editor/lib/src/default_editor/document_selection_on_focus_mixin.dart
angelosilvestre: For the majority of cases, we don't need `requestingFocusByTap`. However, when the user taps an unselectable component, we have a different situation:

- The editor has focus
- The editor doesn't have a selection
- We don't have a `_perviousSelection`
- We don't want to move the caret to the end of the document

So, in the mixin's `_onFocusChange` we need a way to know when we shouldn't change the selection.

matthew-carroll: Maybe the new mixin isn't the most appropriate place to handle that scenario?

What would it look like if the tap handlers saw that the user tapped on a non-selectable component and chose not to change the selection?

matthew-carroll: What currently happens when the user taps on an unselectable component in the document?

angelosilvestre: In the tap handler we request focus, but don't change the selection. So the document has focus, but don't have any selection.

matthew-carroll: I think the least bad answer is probably to change that behavior by moving the selection to the nearest allowable position. Can you try that?

angelosilvestre: Ok, I can try that. 
FYI there's a test that expects the selection to be null in this case, so I will probably need to update that test. The test is "rejects selection when user taps on it" on `supereditor_component_selection_test.dart`.

angelosilvestre: https://github.com/superlistapp/super_editor/blob/a4884bb6beafc47f30f7537108a6be9021debebf/super_editor/test/super_editor/supereditor_component_selection_test.dart#L346-L358

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
```diff
@@ -439,6 +465,11 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
     _clearSelection();
 
     if (docPosition != null) {
+      final tappedComponent = _docLayout.getComponentByNodeId(docPosition.nodeId)!;
```
matthew-carroll, line 468: Please add a comment that explains why we're returning early, and what happened to the selection in this situation.

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
```diff
@@ -499,6 +530,11 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
...
```
matthew-carroll, line 533: Please add a comment here, too.

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
```diff
@@ -439,6 +457,14 @@ class _IOSDocumentTouchInteractorState extends State<IOSDocumentTouchInteractor>
     if (docPosition != null) {
       final didTapOnExistingSelection = selection != null && selection.isCollapsed && selection.extent == docPosition;
 
+      final tappedComponent = _docLayout.getComponentByNodeId(docPosition.nodeId)!;
+      if (!tappedComponent.isVisualSelectionSupported()) {
+        widget.commonOps.moveSelectionToNearestSelectableNode(
```
matthew-carroll, line 462: Please add a comment

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
```diff
@@ -478,6 +504,11 @@ class _IOSDocumentTouchInteractorState extends State<IOSDocumentTouchInteractor>
     widget.composer.clearSelection();
 
     if (docPosition != null) {
+      final tappedComponent = _docLayout.getComponentByNodeId(docPosition.nodeId)!;
```
matthew-carroll, line 507: Please add a comment

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
```diff
@@ -535,6 +566,11 @@ class _IOSDocumentTouchInteractorState extends State<IOSDocumentTouchInteractor>
...
```
matthew-carroll, line 569: Please add a comment

super_editor/lib/src/default_editor/document_selection_on_focus_mixin.dart
```diff
@@ -0,0 +1,98 @@
+import 'package:flutter/material.dart';
+import 'package:super_editor/super_editor.dart';
+
+/// Provides the functionality of updating the selection when the editor receives focus.
```
matthew-carroll, line 4: Try to avoid the passive tense. It adds a bunch of words that often distract from the point.

"Synchronizes document focus with document selection."

super_editor/lib/src/default_editor/document_selection_on_focus_mixin.dart
```diff
@@ -0,0 +1,98 @@
...
+///
+/// Whenever the editor receives focus, if it was previously selected, the previous selection
+/// will be restored.
```
matthew-carroll, line 7: "will be restored" -> "is restored"

super_editor/lib/src/default_editor/document_selection_on_focus_mixin.dart
```diff
@@ -0,0 +1,98 @@
...
+/// If there is no previous selection, the caret will move to the end of the document.
```
matthew-carroll, line 8: "will move to the" -> "is moved to"

also, please add a blank line above this sentence.

super_editor/lib/src/default_editor/document_selection_on_focus_mixin.dart
```diff
@@ -0,0 +1,98 @@
...
+///
+/// `startSyncingSelectionWithFocus` must be called to start this behavior and
```
matthew-carroll, line 10: When we want developers to do something, we should phrase the doc as an instruction, not just a requirement.

"Start watching and synchronizing focus with selection by calling `startSyncingSelectionWithFocus`."

"Stop watching and synchronizing focus with selection by calling `StopSyncingSelectionWithFocus`."

"When the document's [FocusNode] changes, provide the new [FocusNode] to [onFocusNodeReplaced]."

"When the document's [DocumentComposer] changes, provide the new [DocumentComposer] to [onDocumentComposerReplaced]."

"When the document's [DocumentLayoutResolver] changes, provide the new [DocumentLayoutResolver] to [onDocumentLayoutResolverReplaced]."

super_editor/lib/src/default_editor/document_selection_on_focus_mixin.dart
```diff
@@ -0,0 +1,98 @@
...
+/// Whenever the [DocumentLayoutResolver] of the widget using this mixin is replaced, `onDocumentLayoutResolverReplaced` must be called.
+mixin DocumentSelectionOnFocusMixin<T extends StatefulWidget> on State<T> {
+  // Holds the last selection, so we can restore it when the editor is re-focused.
+  DocumentSelection? _previousSelection;
+
+  FocusNode? _focusNode;
+  DocumentComposer? _composer;
+  DocumentLayoutResolver? _getDocumentLayout;
+
+  /// Starts the behavior of updating the selection when the editor is focused.
```
matthew-carroll, line 26: Please update to active tense. Also, please include some details about what is being watched, and what actions are taken.

super_editor/lib/src/default_editor/document_selection_on_focus_mixin.dart
```diff
@@ -0,0 +1,98 @@
...
+    required DocumentLayoutResolver getDocumentLayout,
+  }) {
+    _focusNode = focusNode;
+    _focusNode!.addListener(_onFocusChange);
+    _composer = composer;
+    _composer!.selectionNotifier.addListener(_onSelectionChange);
+    _getDocumentLayout = getDocumentLayout;
+  }
+
+  /// Stops the behavior of updating the selection when the editor is focused.
```
matthew-carroll, line 39: Please update to active tense.

super_editor/test/super_editor/supereditor_component_selection_test.dart
```diff
@@ -343,7 +343,7 @@ void main() {
       );
     });
 
-    testWidgetsOnAllPlatforms("rejects selection when user taps on it", (tester) async {
+    testWidgetsOnAllPlatforms("selects nearest selectable node when user taps on it", (tester) async {
```
matthew-carroll, line 346: I want to confirm, given the change to selection behavior, is there a test with a document that only contains a single, non-selectable node?

angelosilvestre: There wasn't one. I added a test.

angelosilvestre:
@matthew-carroll I added more tests.

matthew-carroll (approved):
LGTM

matthew-carroll:
@angelosilvestre can you resolve the conflicts in this PR?

angelosilvestre:
@matthew-carroll Conflicts resolved.

matthew-carroll:
@miguelcmedeiros you may want to see how this PR with selection behavior changes effects Superlist

PR #723, opened 2022-08-04
[SuperTextField] [iOS] Remove focus when detached from IME (Resolves #662)

### LINKED ISSUE #662: [SuperTextField] - Tapping outside a field doesn't remove focus
On iOS web, when I give a field focus, and then tap outside the field, the keyboard closes, but the field remains focused. Moreover, if I try tapping on the field again, the keyboard doesn't re-appear (because the field thinks its focused).

When the keyboard disappears, the field should lose focus.

It's also probably a good idea that when the user taps a field, even if it already has focus, we should connect to the IME if we're not already connected.

~This problem might exist for Android, too, but I'm not sure.~
^ This issue doesn't seem to happen on Android. Instead, on Android, tapping outside the field doesn't collapse it.

angelosilvestre:
[SuperTextField] [iOS] Remove focus when detached from IME. Resolves https://github.com/superlistapp/super_editor/issues/662

On iOS web, tapping outside the field or clicking on the OK button of the software keyboard was closing the IME connection but keeping the field focused. That way, clicking on the field a second time wasn't causing it to attach to IME again, so the keyboard wasn't re-appearing.

This PR adds a callback to `ImeAttributedTextEditingController` to notify when the IME connection changes. It also changes `SuperIOSTextFieldState` to unfocus when the IME connection is closed on web.

Another possible solution it to change `IOSTextFieldTouchInteractorState` to reattach to IME in `_onTapUp` if it has focus but it was not attached to IME.

### DIFF
--- a/super_editor/lib/src/infrastructure/super_textfield/ios/_user_interaction.dart
+++ b/super_editor/lib/src/infrastructure/super_textfield/ios/_user_interaction.dart
@@ -146,6 +146,10 @@ class IOSTextFieldTouchInteractorState extends State<IOSTextFieldTouchInteractor
     if (widget.focusNode.hasFocus && widget.textController.isAttachedToIme) {
       widget.textController.showKeyboard();
     } else if (widget.focusNode.hasFocus) {
+      // This situation can happen on iOS web when the user taps outside the field
+      // or clicks on the OK button of the software keyboard.
+      // In this situation, the IME connection is closed but the field remains focused.
+      // We need to attach to IME so the keyboard is displayed again.
       widget.textController.attachToIme();
     } else {
       widget.focusNode.requestFocus();
--- a/super_editor/lib/src/infrastructure/super_textfield/ios/_user_interaction.dart
+++ b/super_editor/lib/src/infrastructure/super_textfield/ios/_user_interaction.dart
@@ -145,6 +145,8 @@ class IOSTextFieldTouchInteractorState extends State<IOSTextFieldTouchInteractor
 
     if (widget.focusNode.hasFocus && widget.textController.isAttachedToIme) {
       widget.textController.showKeyboard();
+    } else if (widget.focusNode.hasFocus) {
+      widget.textController.attachToIme();
     } else {
       widget.focusNode.requestFocus();
     }
--- a/super_editor/test/super_textfield/super_textfield_gestures_test.dart
+++ b/super_editor/test/super_textfield/super_textfield_gestures_test.dart
@@ -188,18 +188,52 @@ void main() {
         // Ensure we requested the keyboard to the platform
         expect(isShowKeyboardCalled, true);
       });
+
+      testWidgetsOnIos("tap up attaches to IME if the field already has focus", (tester) async {
+        final controller = ImeAttributedTextEditingController();
+
+        await _pumpTestApp(tester, controller: controller);
+
+        // Tap down and up so the field is focused.
+        await tester.tap(find.byType(SuperTextField));
+        await tester.pumpAndSettle();
+
+        // Ensure we are connected.
+        expect(controller.isAttachedToIme, true);
+
+        // Disconnect from IME.
+        controller.detachFromIme();
+        await tester.pumpAndSettle();
+
+        // Ensure we are not connected.
+        expect(controller.isAttachedToIme, false);
+
+        // Avoid a double tap.
+        await tester.pump(kDoubleTapTimeout + const Duration(milliseconds: 1));
+
+        // Tap down and up again.
+        await tester.tap(find.byType(SuperTextField));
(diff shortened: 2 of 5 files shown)


matthew-carroll:
@angelosilvestre let's try the approach where we reconnect on tap if we're no longer connected. That sounds like less plumbing for the same result.

angelosilvestre:
@matthew-carroll I changed the implementation

super_editor/lib/src/infrastructure/super_textfield/ios/_user_interaction.dart
```diff
@@ -145,6 +145,8 @@ class IOSTextFieldTouchInteractorState extends State<IOSTextFieldTouchInteractor
 
     if (widget.focusNode.hasFocus && widget.textController.isAttachedToIme) {
       widget.textController.showKeyboard();
+    } else if (widget.focusNode.hasFocus) {
+      widget.textController.attachToIme();
```
matthew-carroll, line 149: It would be good to add a comment that describes a situation where this might happen. It's very common to come back to an if-statement like this in a few months and wonder why we thought this condition was important. In fact, I've done that, and it's possible that I had this if-statement before, and then deleted it because I couldn't remember why it was there.

super_editor/test/super_textfield/super_textfield_gestures_test.dart
```diff
@@ -188,18 +188,52 @@ void main() {
...
+        await _pumpTestApp(tester, controller: controller);
+
+        // Tap down and up so the field is focused.
+        await tester.tap(find.byType(SuperTextField));
+        await tester.pumpAndSettle();
+
+        // Ensure we are connected.
+        expect(controller.isAttachedToIme, true);
+
+        // Disconnect from IME.
```
matthew-carroll, line 204: Let's include a mention about how/why an IME might disconnect in a real app.

super_editor/test/super_textfield/super_textfield_gestures_test.dart
```diff
@@ -188,18 +188,52 @@ void main() {
...
+        controller.detachFromIme();
+        await tester.pumpAndSettle();
+
+        // Ensure we are not connected.
+        expect(controller.isAttachedToIme, false);
+
+        // Avoid a double tap.
```
matthew-carroll, line 211: Can we move this up to where the tap happens so that the connection is more obvious?

super_editor/lib/src/infrastructure/super_textfield/ios/_user_interaction.dart
```diff
@@ -145,6 +145,12 @@ class IOSTextFieldTouchInteractorState extends State<IOSTextFieldTouchInteractor
 
     if (widget.focusNode.hasFocus && widget.textController.isAttachedToIme) {
       widget.textController.showKeyboard();
+    } else if (widget.focusNode.hasFocus) {
+      // This situation can happen on iOS web when the user taps outside the field
```
matthew-carroll, line 149: I think this comment describes the wrong tap, doesn't it? Aren't there two things that need to happen?

When the field is focused with an open IME connection, the user taps outside the field or closes the keyboard, which closes the IME, but doesn't remove focus. Then, the user taps on the field again, which still has focus, but the IME connection is closed. So we re-open the IME connection.

Also, I want to double check, this only happens on iOS web? Not native iOS? Not Android web? etc? Do we know why it only happens on web?

super_editor/test/super_textfield/super_textfield_gestures_test.dart
```diff
@@ -188,18 +188,53 @@ void main() {
...
+        // In a real app this could happen when the user taps outside the field 
+        // or clicks on the OK button of the software keyboard.
+        controller.detachFromIme();
+        await tester.pumpAndSettle();
+
+        // Ensure we are not connected.
+        expect(controller.isAttachedToIme, false);
+
+        // Avoid a double tap.
+        await tester.pump(kDoubleTapTimeout + const Duration(milliseconds: 1));
```
matthew-carroll, line 214: In my previous request, I was asking if you could move this line up to where you performed the tap. That's lines `198` and `199`. By the time we get down here, it's immediately clear what we're avoiding, or where the previous tap took place. Better to do this near the place where you executed the tap.

angelosilvestre: Updated.

matthew-carroll: Looks like this pump is still here. Now you have one up above and here.

angelosilvestre: Updated

matthew-carroll (approved):
LGTM

PR #725, opened 2022-08-07
[AttributedText] Add method to get the attributed range for a set of attributions. (Resolves #590)

### LINKED ISSUE #590: Add AttributedSpans method to get attribution about index
It's useful to get the span range that includes a given offset. For example:

"Hello, [world]!"

Imagine that you have the offset for the "o" character, and now you want the span range for the whole word that's attributed around the "o".

Add an API that provides this capability.

angelosilvestre:
[AttributedText] Add method to get the attributed range for a set of attributions. Resolves https://github.com/superlistapp/super_editor/issues/590

This PR adds a method to return the span range that's attributed around an offset.

For example:

`He**ll~o Wor**ld~` 

Calling this method with the offset of "W" and the attributions `bold` and `strikethrough` will return the range of "o Wor".

I made `getStartingMarkerAtOrBefore` and `getEndingMarkerAtOrAfter` public to avoid duplicating the code.

### DIFF
--- a/attributed_text/lib/src/attributed_spans.dart
+++ b/attributed_text/lib/src/attributed_spans.dart
@@ -5,6 +5,7 @@ import 'package:meta/meta.dart';
 
 import 'attribution.dart';
 import 'logging.dart';
+import 'span_range.dart';
 
 final _log = attributionsLog;
 
@@ -109,11 +110,11 @@ class AttributedSpans {
     int offset, {
     Attribution? attribution,
   }) {
-    SpanMarker? markerBefore = getStartingMarkerAtOrBefore(offset, attribution: attribution);
+    SpanMarker? markerBefore = _getStartingMarkerAtOrBefore(offset, attribution: attribution);
     if (markerBefore == null) {
       return false;
     }
-    SpanMarker? markerAfter = getEndingMarkerAtOrAfter(markerBefore.offset, attribution: attribution);
+    SpanMarker? markerAfter = _getEndingMarkerAtOrAfter(markerBefore.offset, attribution: attribution);
     if (markerAfter == null) {
       throw Exception('Found an open-ended attribution. It starts with: $markerBefore');
     }
@@ -140,8 +141,8 @@ class AttributedSpans {
     // The following methods should be guaranteed to produce non-null
     // values because we already verified that the given attribution
     // exists at the given offset.
-    SpanMarker markerBefore = getStartingMarkerAtOrBefore(offset, attribution: attribution)!;
-    SpanMarker markerAfter = getEndingMarkerAtOrAfter(markerBefore.offset, attribution: attribution)!;
+    SpanMarker markerBefore = _getStartingMarkerAtOrBefore(offset, attribution: attribution)!;
+    SpanMarker markerAfter = _getEndingMarkerAtOrAfter(markerBefore.offset, attribution: attribution)!;
 
     return AttributionSpan(
       attribution: attribution,
@@ -215,10 +216,43 @@ class AttributedSpans {
     return matchingAttributionSpans;
   }
 
+  /// Returns the range that's attributed around [offset]. [attributions] must not be empty.
+  ///
+  /// It will return only the portion in which all attributions are present.
+  SpanRange getAttributedRange(Set<Attribution> attributions, int offset) {
+    if (attributions.isEmpty) {
+      throw Exception('getAttributedRange requires a non empty set of attributions');
+    }
+
+    int? maxStartMarkerOffset;
+    int? minEndMarkerOffset;
+
+    for (final attribution in attributions) {
+      if (!hasAttributionAt(offset, attribution: attribution)) {
+        throw Exception('Tried to get the attributed range of ($attribution) at offset "$offset" but the given attribution does not exist at that offset.');
+      }
+      int startMarkerOffset = _getStartingMarkerAtOrBefore(offset, attribution: attribution)!.offset;
+      int endMarkerOffset = _getEndingMarkerAtOrAfter(offset, attribution: attribution)!.offset;
+
+      if (maxStartMarkerOffset == null || startMarkerOffset > maxStartMarkerOffset) {
+        maxStartMarkerOffset = startMarkerOffset;
+      }
+
+      if (minEndMarkerOffset == null || endMarkerOffset < minEndMarkerOffset) {
+        minEndMarkerOffset = endMarkerOffset;
+      }
+    }
+
+    return SpanRange(
+      start: maxStartMarkerOffset!,
+      end: minEndMarkerOffset!,
+    );
+  }
+
   /// Finds and returns the nearest [start] marker that appears at or before the
   /// given [offset], optionally looking specifically for a marker with
   /// the given [attribution].
-  SpanMarker? getStartingMarkerAtOrBefore(int offset, {Attribution? attribution}) {
+  SpanMarker? _getStartingMarkerAtOrBefore(int offset, {Attribution? attribution}) {
     return markers //
         .reversed // search from the end so its the nearest start marker
         .where((marker) {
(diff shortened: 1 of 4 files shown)


attributed_text/lib/src/attributed_text.dart
```diff
@@ -142,6 +142,39 @@ class AttributedText {
     );
   }
 
+  /// Returns the range that's attributed around [offset]. [attributions] must not be empty.
+  ///
+  /// It will return only the portion in which all attributions are present.
+  SpanRange getAttributedRange(Set<Attribution> attributions, int offset) {
```
matthew-carroll, line 148: If we updated the existing `AttributedSpans` method to accept a `Set`, couldn't we get this implementation for free? You could get the result from `AttributedSpans` and then convert it to a `SpanRange`. Am I missing anythign?

angelosilvestre: `expandAttributionToSpan` returns an `AttributionSpan`, which has only one attribution.

Wouldn't it be weird to pass a set of attributions and get back an object containing only one?

matthew-carroll: That's a good point. Ok, then let's essentially move your new implementation in `AttributedSpans`. Then have `AttributedText` define a similarly named method and forward the call to `AttributedSpans`.

The idea here is that we want `AttributedSpans` to be responsible for inspecting itself. We don't want other objects doing surgery on it. For example, you had to make private things public to implement this behavior in `AttributedText`. That's usually a warning sign.

angelosilvestre: Should I also change the tests to test `AttributedSpans` directly?

matthew-carroll: we should test both, but the tests directly on `AttributedSpans` should probably be more exhaustive than those on `AttributedText`

angelosilvestre: Could you take a look to see what tests should I add to `AttributedSpans` ?

attributed_text/lib/src/attributed_spans.dart
```diff
@@ -215,6 +216,39 @@ class AttributedSpans {
     return matchingAttributionSpans;
   }
 
+  /// Returns the range that's attributed around [offset]. [attributions] must not be empty.
```
matthew-carroll, line 219: The first line of a Dart doc should always be a single sentence.

Also, I think a more accurate description would be "Returns the range about [offset], which is attributed with all given [attributions]."

attributed_text/lib/src/attributed_spans.dart
```diff
@@ -215,6 +216,39 @@ class AttributedSpans {
...
+  ///
+  /// It will return only the portion in which all attributions are present.
```
matthew-carroll, line 221: This line is superfluous, if you change the first sentence to what I recommended above.

attributed_text/lib/src/attributed_spans.dart
```diff
@@ -215,6 +216,39 @@ class AttributedSpans {
...
+    if (attributions.isEmpty) {
+      throw Exception('getAttributedRange requires a non empty set of attributions');
+    }
+
+    int? maxStartMarkerOffset;
+    int? minEndMarkerOffset;
+
+    for (final attribution in attributions) {
+      if (!hasAttributionAt(offset, attribution: attribution)) {
+        throw Exception('Tried to get the attributed range of ($attribution) at offset "$offset" but the given attribution does not exist at that offset.');
```
matthew-carroll, line 232: Please include the missing attribution in the exception message so that the developer might debug the problem.

attributed_text/lib/src/attributed_spans.dart
```diff
@@ -215,6 +216,39 @@ class AttributedSpans {
...
+        maxStartMarkerOffset = startMarkerOffset;
+      }
+
+      if (minEndMarkerOffset == null || endMarkerOffset < minEndMarkerOffset) {
+        minEndMarkerOffset = endMarkerOffset;
+      }
+    }
+
+    return SpanRange(
+      start: maxStartMarkerOffset!,
```
matthew-carroll, line 247: Is it possible that either of these offsets is still null?

attributed_text/lib/src/attributed_text.dart
```diff
@@ -142,6 +142,13 @@ class AttributedText {
     );
   }
 
+  /// Returns the range that's attributed around [offset]. [attributions] must not be empty.
```
matthew-carroll, line 145: Same doc comment here as for the attributed spans.

attributed_text/lib/src/attributed_spans.dart
angelosilvestre: It's already included in the message.

attributed_text/lib/src/attributed_spans.dart
angelosilvestre: As we check that all attributions are present at the given offset, both offsets should always be non-null. I added a comment explaining this.

attributed_text/test/attributed_text_test.dart
```diff
@@ -178,5 +178,98 @@ void main() {
         );
       });
     });
+
+    group('getAttributedRange', () {
```
matthew-carroll, line 182: Let's try to represent user behavior in our tests where we can, rather than tests of specific methods, or nuanced technical details.

We already have specific tests in `AttributedRange` that focus on tech details. What we probably want to focus on here is real-world uses of that ability.

Therefore, you might define a test that finds all bold text around a character. You might also write a test thad finds all bold and italic text around a character. You might write a test that finds bold, italic, and strikethrough text that appears within a word that also includes some bold and italics without strikethrough.

See if you can name this group and these tests more towards the feature we're providing, and less about the methods and data structures.

matthew-carroll:
@angelosilvestre is this PR ready for re-review?

angelosilvestre:
@matthew-carroll I still need to address the requested changes of the last review. I'll do that today.

attributed_text/test/attributed_text_test.dart
angelosilvestre: Please take a look if the test descriptions aren't too long.

attributed_text/lib/src/attributed_spans.dart
```diff
@@ -215,6 +216,46 @@ class AttributedSpans {
...
+
+    int? maxStartMarkerOffset;
+    int? minEndMarkerOffset;
+
+    for (final attribution in attributions) {
+      if (!hasAttributionAt(offset, attribution: attribution)) {
+        throw Exception(
+            'Tried to get the attributed range of ($attribution) at offset "$offset" but the given attribution does not exist at that offset.');
+      }
+      // The following methods should be guaranteed to produce non-null
```
matthew-carroll, line 235: I know I had a question below about nullability, but you don't need this comment for these variables because the type is `int` and not `int?`.

attributed_text/lib/src/attributed_spans.dart
```diff
@@ -215,6 +216,46 @@ class AttributedSpans {
...
+      if (maxStartMarkerOffset == null || startMarkerOffset > maxStartMarkerOffset) {
+        maxStartMarkerOffset = startMarkerOffset;
+      }
+
+      if (minEndMarkerOffset == null || endMarkerOffset < minEndMarkerOffset) {
+        minEndMarkerOffset = endMarkerOffset;
+      }
+    }
+
+    // Both offsets should be guaranteed to be non-null
```
matthew-carroll, line 250: Let's avoid the word "should" unless there's some unknown.

```
// Note: maxStartMarkerOffset and minEndMarkerOffset are non-null because we verified
// that all desired attributions are present at the given offset.
```

matthew-carroll (approved):
LGTM, with a couple comment changes

angelosilvestre:
@matthew-carroll I updated the comments

PR #730, opened 2022-08-12
Fix hint component and example app toolbar focus (Resolves #726)

### LINKED ISSUE #726: [super_editor] - Can't place caret in hint component, and other bugs
A few bugs have appeared with the new gesture widget organization:

- The hint text in the hint text component is absorbing the pointer before the gesture gets to the document
- The cursor should be a text cursor throughout the document, not just over nodes
- The toolbar focus in the example app is broken again - tapping anything on the toolbar makes the toolbar disappear

matthew-carroll:
Fix hint component and example app toolbar focus (Resolves #726)

The hint component's hint text was absorbing gesture interactions, preventing the placement of the caret. Now it ignores the pointer. The tests for this behavior required a modification to our test tools so that we can place the caret in a component that wraps a text component, instead of just a text component.

I also changed the cursor within the editor to always look like a text cursor, unless the cursor is over a selectable box, at which point it looks like a regular cursor again.

Lastly, I did my best to fix the popover toolbar focus issue. By adding a `FocusScope` above our custom `Overlay` in the example app, the user can tap toolbar buttons like "bold" and "italics" without issue. However, the dropdowns are still broken because they steal focus, and there doesn't seem to be a way to share that focus with the editor.

### DIFF
--- a/super_editor/example/lib/demos/example_editor/_toolbar.dart
+++ b/super_editor/example/lib/demos/example_editor/_toolbar.dart
@@ -2,8 +2,8 @@ import 'dart:math';
 
 import 'package:example/logging.dart';
 import 'package:flutter/material.dart';
-import 'package:super_editor/super_editor.dart';
 import 'package:flutter_gen/gen_l10n/app_localizations.dart';
+import 'package:super_editor/super_editor.dart';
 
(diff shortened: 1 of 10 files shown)


angelosilvestre (approved):
LGTM

PR #752, opened 2022-08-20
[super_editor_markdown] Fix serialization of text alignment, strikethrough and underline (Resolves #717)

### LINKED ISSUE #717: [super_editor_markdown] - Text alignment (textAlign) is ignored during serialization/deserialization
Thanks for the plugin. 
Everything is very super. 
But I noticed a little problem: Metadata 'textAlign' is ignored during serialization/deserialization.

angelosilvestre:
[super_editor_markdown] Fix serialization of text alignment, strikethrough, and underline. Resolves #717

Strikethrough 
---
Strikethrough is being ignored during deserialization. The markdown package used for deserialization already supports the strikethrough syntax, however, it isn't enabled by default.

This PR enables the strikethrough syntax for deserialization.

Underline
---
Underline is being ignored in both serialization and deserialization. The markdown spec doesn't have a syntax for underline text. Some parsers allow the usage of a `ins` tag to specify the underline, however, we'd like to avoid using HTML, so a custom notation is needed.

This PR introduces a custom notation for underline:
```
This is ¬underline¬.
```

The `¬` character could be changed if a more appropriate one is suggested.

Text Alignment
---
Text alignment is being ignored in both serialization and deserialization. The markdown spec doesn't have a syntax for text alignment. Some parsers allow the usage of a `p` tag to specify the alignment, however, we'd like to avoid using HTML, so a custom notation is needed.

This PR introduces a custom notation for text alignment, similar to what is used to table alignment. An alignment indicator should precede the paragraph. For example:

```
:---
This is left-aligned. (Can be omitted, as it's the default)

:---:
This is center-aligned.

---:
This is right-aligned.
```

The usage of these custom notations is configurable.

### DIFF
--- a/super_editor_markdown/lib/src/markdown.dart
+++ b/super_editor_markdown/lib/src/markdown.dart
@@ -9,10 +9,8 @@ import 'package:super_editor/super_editor.dart';
 //       requires one. When the editing system matures, there should
 //       be a way to return something here that is not concrete.
 
-/// Deserializes markdown to a document.
-///
-/// If [syntax] is `MarkdownSyntax.superEditor`, a custom notation is used to allow
-/// text alignment and underline.
+/// The given [syntax] controls how the [markdown] is parsed, e.g., [MarkdownSyntax.normal] for strict
+/// Markdown parsing, or [MarkdownSyntax.superEditor] to use Super Editor's extended syntax.
 MutableDocument deserializeMarkdownToDocument(
   String markdown, {
   MarkdownSyntax syntax = MarkdownSyntax.superEditor,
@@ -38,21 +36,8 @@ MutableDocument deserializeMarkdownToDocument(
   return MutableDocument(nodes: nodeVisitor.content);
 }
 
-/// Serializes a document to markdown format.
-///
-/// If [syntax] is `MarkdownSyntax.superEditor`, a custom notation is used to allow
-/// text alignment and underline.
-///
-/// Underline text is serialized between a pair of `¬`.
-///
-/// Text alignment is serialized using an alignment notation at the
-/// line preceding the paragraph:
-///
-/// `:---` represents left alignment. (The default)
-///
-/// `:---:` represents center alignment.
-///
-/// `---:` represents right alignment.
+/// The given [syntax] controls how the [doc] is serialized, e.g., [MarkdownSyntax.normal] for standard
+/// Markdown syntax, or [MarkdownSyntax.superEditor] to use Super Editor's extended syntax.
 String serializeDocumentToMarkdown(
   Document doc, {
   MarkdownSyntax syntax = MarkdownSyntax.superEditor,
@@ -647,22 +632,22 @@ class _ParagraphWithAlignmentSyntax extends md.ParagraphSyntax {
 
     final nextLine = parser.peek(1);
 
-    // The alignment token is at the last line of the document.
-    // As there are no lines after the token, it is considered a left-aligned paragraph,
-    // containing the alignment token as text.
+    // We found a match for a paragraph alignment token. However, the alignment token is the last
+    // line of content in the document. Therefore, it's not really a paragraph alignment token, and we
+    // should treat it as regular content.
     if (nextLine == null) {
       return false;
     }
 
-    // There is at least one more line after the current parser offset. The next line represents
-    // a different block element. The parser is not currently looking at an aligned paragraph.
+    /// We found a paragraph alignment token, but the block after the alignment token isn't a paragraph.
+    /// Therefore, the paragraph alignment token is actually regular content. This parser doesn't need to
+    /// take any action.
     if (_standardNonParagraphBlockSyntaxes.any((syntax) => syntax.pattern.hasMatch(nextLine))) {
       return false;
     }
 
-    // As we've found an alignment token followed by at least one more line that
-    // isn't a different block element, then it should be enough to consider
-    // we are looking at an aligned paragraph.
+    // We found a paragraph alignment token, followed by a paragraph. Therefore, this parser should
+    // parse the given content.
     return true;
   }
 
@@ -677,11 +662,7 @@ class _ParagraphWithAlignmentSyntax extends md.ParagraphSyntax {
     // Parse the paragraph using the standard Markdown paragraph parser.
     final paragraph = super.parse(parser);
 
-    // The standard paragraph parser may return an empty text node when the paragraph 
-    // contains only reference link definitions. If it's the case, we return the parsed
-    // node as is.
     if (paragraph is md.Element) {
-      // If we get to the parse method then it should be guaranteed we have an alignment token.      
       paragraph.attributes.addAll({'textAlign': _convertMarkdownAlignmentTokenToSuperEditorAlignment(match!.input)});
     }
 
@@ -725,5 +706,16 @@ enum MarkdownSyntax {
   normal,
 
   /// Extended syntax which supports serialization of text alignment, strikethrough and underline.
+  ///
+  /// Underline text is serialized between a pair of `¬`.
+  ///
+  /// Text alignment is serialized using an alignment notation at the
+  /// line preceding the paragraph:
+  ///
+  /// `:---` represents left alignment. (The default)
+  ///
+  /// `:---:` represents center alignment.
+  ///
+  /// `---:` represents right alignment.
   superEditor,
 }
--- a/super_editor_markdown/lib/src/markdown.dart
+++ b/super_editor_markdown/lib/src/markdown.dart
@@ -11,8 +11,8 @@ import 'package:super_editor/super_editor.dart';
 
 /// Deserializes markdown to a document.
 ///
-/// If [extendedSyntax] is `true` text alignment is parsed
-/// using a custom notation.
+/// If [extendedSyntax] is `true`, a custom notation is used to allow
+/// text alignment and underline.
 MutableDocument deserializeMarkdownToDocument(
   String markdown, {
   bool extendedSyntax = true,
@@ -40,8 +40,19 @@ MutableDocument deserializeMarkdownToDocument(
 
 /// Serializes a document to markdown format.
 ///
-/// If [extendedSyntax] is `true` text alignment is serialized
-/// using a custom notation.
+/// If [extendedSyntax] is `true`, a custom notation is used to allow
+/// text alignment and underline.
+///
+/// Underline text is serialized between a pair of `¬`.
+///
+/// Text alignment is serialized using an alignment notation at the
+/// line preceding the paragraph:
+///
+/// `:---` represents left alignment. (The default)
+///
+/// `:---:` represents center alignment.
+///
+/// `---:` represents right alignment.
 String serializeDocumentToMarkdown(
   Document doc, {
   bool extendedSyntax = true,
@@ -359,7 +370,10 @@ class _MarkdownToDocument implements md.NodeVisitor {
     final inlineParser = md.InlineParser(
       element.textContent,
       md.Document(
-        inlineSyntaxes: [md.StrikethroughSyntax()],
+        inlineSyntaxes: [
+          md.StrikethroughSyntax(),
+          UnderlineSyntax(),
+        ],
       ),
     );
     final inlineVisitor = _InlineMarkdownToDocument();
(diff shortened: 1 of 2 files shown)


super_editor_markdown/test/super_editor_markdown_test.dart
```diff
@@ -295,6 +295,24 @@ This is some code
         expect(serializeDocumentToMarkdown(doc), '[This **is a** paragraph](https://example.org).');
       });
 
+      test('paragraph with underline', () {
```
matthew-carroll, line 298: Do we already have tests for strikethrough?

super_editor_markdown/lib/src/markdown.dart
```diff
@@ -80,6 +110,14 @@ String serializeDocumentToMarkdown(Document doc) {
           ..writeln(node.text.toMarkdown()) //
           ..write('```');
       } else {
+        final String? textAlign = node.getMetadataValue('textAlign');
+        // Left alignment is the default, so there is no need to add the indicator.
+        if (extendedSyntax && textAlign != null && textAlign != 'left') {
```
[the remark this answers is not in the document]
angelosilvestre: That is correct. When we're left aligned the paragraph is serialized without any additional syntax.

super_editor_markdown/test/super_editor_markdown_test.dart
```diff
@@ -763,6 +838,81 @@ This is some code
...
+
+      test('paragraph with right alignment', () {
+        final doc = deserializeMarkdownToDocument('---:\nParagraph1');
+
+        final paragraph = doc.nodes.first as ParagraphNode;
+        expect(paragraph.getMetadataValue('textAlign'), 'right');
+        expect(paragraph.text.text, 'Paragraph1');
+      });
+
+      test('treats alignment notation as content at the end of the document', () {
```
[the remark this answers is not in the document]
angelosilvestre: It means the deserialization will produce a paragraph containing the alignment token as text instead of using it to determine the text alignment.

Would 'treats alignment token as text at the end of the document' be a better name?

super_editor_markdown/test/super_editor_markdown_test.dart
angelosilvestre: I added a serialization test for strikethrough.

super_editor_markdown/lib/src/markdown.dart
```diff
@@ -532,3 +621,97 @@ class AttributedTextMarkdownSerializer extends AttributionVisitor {
...
+
+  @override
+  bool canParse(md.BlockParser parser) {
+    if (!_alignmentNotationPattern.hasMatch(parser.current)) {
+      return false;
+    }
+
+    final nextLine = parser.peek(1);
+
+    // The indicator is at the end of the document,
```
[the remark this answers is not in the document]
angelosilvestre: I updated the comment.

super_editor_markdown/lib/src/markdown.dart
```diff
@@ -532,3 +621,97 @@ class AttributedTextMarkdownSerializer extends AttributionVisitor {
...
+    if (nextLine == null) {
+      return false;
+    }
+
+    // The line following the indicator is another block syntax.
+    if (_standardNonParagraphBlockSyntaxes.any((syntax) => syntax.pattern.hasMatch(nextLine))) {
+      return false;
+    }
+
+    return true;
```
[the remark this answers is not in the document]
angelosilvestre: I added a comment.

super_editor_markdown/lib/src/markdown.dart
```diff
@@ -532,3 +621,97 @@ class AttributedTextMarkdownSerializer extends AttributionVisitor {
...
+  }
+
+  @override
+  md.Node parse(md.BlockParser parser) {
+    final match = _alignmentNotationPattern.firstMatch(parser.current);
+
+    // If we get to the parse method than we should have an alignment indicator.
+    // Using the default paragraph parser just to avoid throwing an exception.
```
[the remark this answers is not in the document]
angelosilvestre: I was referring to throwing an exception if `parse` is called and the text doesn't contain an alignment token.
But, as we already check we have a match in `canParse`, it should be safe to remove this check and assume we will have a match.

super_editor_markdown/lib/src/markdown.dart
```diff
@@ -532,3 +621,97 @@ class AttributedTextMarkdownSerializer extends AttributionVisitor {
...
+
+    // As we already checked that there is a line following the indicator
+    // in canParse, it should be safe to assume the next line will be present.
+    parser.advance();
+
+    // We let the default paragraph parser handle the parsing of the paragraph
+    // and just add the alignment attribute.
+    final paragraph = super.parse(parser);
+
+    // The parser may return an empty text node.
```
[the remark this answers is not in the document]
angelosilvestre: The standard paragraph parser returns an empty `Text` node if the paragraph contains only reference links. I updated the comment.

super_editor_markdown/lib/src/markdown.dart
```diff
@@ -532,3 +621,97 @@ class AttributedTextMarkdownSerializer extends AttributionVisitor {
...
+    if (paragraph is md.Element) {
+      paragraph.attributes.addAll({'textAlign': _convertAlignmentIndicator(match.input)});
+    }
+
+    return paragraph;
+  }
+
+  String _convertAlignmentIndicator(String indicator) {
```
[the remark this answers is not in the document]
angelosilvestre: Dart doc added.

super_editor_markdown/test/super_editor_markdown_test.dart
```diff
@@ -316,6 +334,63 @@ This is some code
...
+          ParagraphNode(
+            id: '1',
+            text: AttributedText(text: 'Paragraph1'),
+            metadata: {
+              'textAlign': 'left',
+            },
+          ),
+        ]);
+
+        // Left alignment is the default, so it shouldn't be added.
```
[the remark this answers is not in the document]
angelosilvestre: I updated the comment.

super_editor_markdown/lib/src/markdown.dart
```diff
@@ -8,10 +8,22 @@ import 'package:super_editor/super_editor.dart';
 //       For now, we return MutableDocument because DocumentEditor
 //       requires one. When the editing system matures, there should
 //       be a way to return something here that is not concrete.
-MutableDocument deserializeMarkdownToDocument(String markdown) {
+
+/// Deserializes markdown to a document.
+///
+/// If [syntax] is `MarkdownSyntax.superEditor`, a custom notation is used to allow
```
matthew-carroll, line 14: This comment is going to become out-of-date as soon as we add one more Super Editor capability. Let's describe the parameter without locking into a specific feature set:

```
/// The given [syntax] controls how the [markdown] is parsed, e.g., [MarkdownSyntax.normal] for strict
/// Markdown parsing, or [MarkdownSyntax.superEditor] to use Super Editor's extended syntax.
```

super_editor_markdown/lib/src/markdown.dart
```diff
@@ -26,7 +38,25 @@ MutableDocument deserializeMarkdownToDocument(String markdown) {
   return MutableDocument(nodes: nodeVisitor.content);
 }
 
-String serializeDocumentToMarkdown(Document doc) {
+/// Serializes a document to markdown format.
+///
+/// If [syntax] is `MarkdownSyntax.superEditor`, a custom notation is used to allow
```
matthew-carroll, line 43: Similar issue here with the details about individual syntax features.

super_editor_markdown/lib/src/markdown.dart
```diff
@@ -26,7 +38,25 @@ MutableDocument deserializeMarkdownToDocument(String markdown) {
...
+/// text alignment and underline.
+///
+/// Underline text is serialized between a pair of `¬`.
```
matthew-carroll, line 46: These extension syntax definitions should probably go in a Dart Doc for the `MarkdownSyntax.superEditor` enum. Consider what would happen to this Dart Doc if we added a 3rd syntax.

super_editor_markdown/lib/src/markdown.dart
```diff
@@ -532,3 +620,110 @@ class AttributedTextMarkdownSerializer extends AttributionVisitor {
...
+
+  @override
+  bool canParse(md.BlockParser parser) {
+    if (!_alignmentNotationPattern.hasMatch(parser.current)) {
+      return false;
+    }
+
+    final nextLine = parser.peek(1);
+
+    // The alignment token is at the last line of the document.
```
matthew-carroll, line 650: This comment still misses the context that a reader would want. It jumps right in about an "alignment token" without mentioning why we're talking about an alignment token.

```
// We found a match for a paragraph alignment token. However, the alignment token is the last
// line of content in the document. Therefore, it's not really a paragraph alignment token, and we
// should treat it as regular content.
```

super_editor_markdown/lib/src/markdown.dart
```diff
@@ -532,3 +620,110 @@ class AttributedTextMarkdownSerializer extends AttributionVisitor {
...
+    // As there are no lines after the token, it is considered a left-aligned paragraph,
+    // containing the alignment token as text.
+    if (nextLine == null) {
+      return false;
+    }
+
+    // There is at least one more line after the current parser offset. The next line represents
```
matthew-carroll, line 657: This comment doesn't make much sense to me. It reads like 3 disconnected statements. I'm not sure how they're intended to tie together. You said "the next line represents a different block element" - a different block element from what?

If I'm understanding this comment correctly, and if I understand the code, then it sounds like what you want to say is:

```
/// We found a paragraph alignment token, but the block after the alignment token isn't a paragraph.
/// Therefore, the paragraph alignment token is actually regular content. This parser doesn't need to
/// take any action.
```

super_editor_markdown/lib/src/markdown.dart
```diff
@@ -532,3 +620,110 @@ class AttributedTextMarkdownSerializer extends AttributionVisitor {
...
+    // a different block element. The parser is not currently looking at an aligned paragraph.
+    if (_standardNonParagraphBlockSyntaxes.any((syntax) => syntax.pattern.hasMatch(nextLine))) {
+      return false;
+    }
+
+    // As we've found an alignment token followed by at least one more line that
```
matthew-carroll, line 663: This comment can be more concise:

```
// We found a paragraph alignment token, followed by a paragraph. Therefore, this parser should
// parse the given content.
```

super_editor_markdown/lib/src/markdown.dart
```diff
@@ -532,3 +620,110 @@ class AttributedTextMarkdownSerializer extends AttributionVisitor {
...
+    final match = _alignmentNotationPattern.firstMatch(parser.current);
+
+    // We've parsed the alignment token on the current line. We know a paragraph starts on the
+    // next line. Move the parser to the next line so that we can parse the paragraph.
+    parser.advance();
+
+    // Parse the paragraph using the standard Markdown paragraph parser.
+    final paragraph = super.parse(parser);
+
+    // The standard paragraph parser may return an empty text node when the paragraph
```
matthew-carroll, line 680: As a reader, I'm not sure what to do with this comment. I can't tell whether this comment is describing what's in the if-statement block, or describing a situation that bypasses the if-statement block.

super_editor_markdown/lib/src/markdown.dart
```diff
@@ -532,3 +620,110 @@ class AttributedTextMarkdownSerializer extends AttributionVisitor {
...
+    parser.advance();
+
+    // Parse the paragraph using the standard Markdown paragraph parser.
+    final paragraph = super.parse(parser);
+
+    // The standard paragraph parser may return an empty text node when the paragraph 
+    // contains only reference link definitions. If it's the case, we return the parsed
+    // node as is.
+    if (paragraph is md.Element) {
+      // If we get to the parse method then it should be guaranteed we have an alignment token.
```
matthew-carroll, line 684: "If we" and "it should" both look like words that shouldn't appear in this comment. If this comment is needed, please reword it as a statement about what we're doing and/or why we're doing it.

matthew-carroll (approved):
LGTM, with a handful of Dart comment changes

super_editor_markdown/lib/src/markdown.dart
angelosilvestre: I added this comment just because I'm using the null check operator on `match!.input`. I think I could just remove it.

super_editor_markdown/lib/src/markdown.dart
angelosilvestre: This comment isn't really adding any useful information. I think I could just remove it.

angelosilvestre:
@matthew-carroll I updated the comments.

PR #754, opened 2022-08-21
Fixing super_editor/example compilation by adding missing import

jmatth:
A missing import in `super_editor/example/lib/main.dart` is causing the example app to fail to compile. This PR just adds the import back in. It looks like the import was removed as part of [this commit][1].

[1]: https://github.com/superlistapp/super_editor/commit/574bc21770a50d41fb48788aa7ea2df995723d26#diff-87261c6ba97adf363861c4f7733e9ff6edbd7cf8732724357ec0e8a8dc22b1bfL21


matthew-carroll (approved):
LGTM

PR #755, opened 2022-08-22
Fix default gesture mode of supereditor widget (Resolves #741)

### LINKED ISSUE #741: [super_editor] - The default gesture mode should be based on the platform
Currently, the `SuperEditor` widget always configures itself for mouse input, regardless of platform.

`SuperEditor` should default to mouse input for Mac, Windows, Linux, and Web. But it should default to Android gestures on Android, and iOS gestures on iOS.

debasmitasarkar:
[SuperEditor] Fix default gesture mode of super editor widget . Resolves https://github.com/superlistapp/super_editor/issues/741

This PR changes the default initialization of optional parameter `gestureMode` in `SuperEditor` widget.
`SuperEditor` now will default to mouse input for Mac, Windows, Linux, and Web and it will to Android gestures on Android, and iOS gestures on iOS.


matthew-carroll:
@debasmitasarkar it looks like this PR is missing tests

debasmitasarkar:
> 

Sure adding test cases

debasmitasarkar:
> @debasmitasarkar it looks like this PR is missing tests

@matthew-carroll added test cases

super_editor/test/src/default_editor/super_editor_test.dart
```diff
@@ -73,6 +74,26 @@ void main() {
         expect(SuperEditorInspector.findParagraphStyle("1")!.color, Colors.white);
       });
     });
+
+    group("gesture mode", () {
+      testWidgets(
```
matthew-carroll, line 79: Please check how we group and name other tests and try to do something similar here.

super_editor/test/src/default_editor/super_editor_test.dart
```diff
@@ -73,6 +74,26 @@ void main() {
...
+        // Configure and render a document.
+        debugDefaultTargetPlatformOverride =
+            _targetPlatformAndGestureVariants.currentValue!.platform;
+        await tester //
+            .createDocument()
+            .withSingleParagraph()
+            .pump();
+
+        // Ensure that the gesture mode is correct.
+        expect(SuperEditorInspector.findGestureMode(),
```
matthew-carroll, line 92: Is there a demonstrable user interaction behavior that would ensure we're using the correct gesture mode? In theory, the reason that we switch gesture modes is so that we get different behavior. Can you look into that behavior and see if there are any side effects that we can look for, instead of querying properties?

super_editor/test/src/default_editor/super_editor_test.dart
```diff
@@ -73,6 +74,26 @@ void main() {
...
+            _targetPlatformAndGestureVariants.currentValue!.gestureMode);
+        debugDefaultTargetPlatformOverride = null;
+      }, variant: _targetPlatformAndGestureVariants);
```
matthew-carroll, line 95: I don't think this variant behavior is necessary. We already have test methods that will run tests on all desktop platforms and all mobile platforms.

matthew-carroll:
@debasmitasarkar can you update this PR with the requested changes and resolve the conflicts? Please request another review, when it's ready.

super_editor/test/src/default_editor/super_editor_test.dart
debasmitasarkar: > @debasmitasarkar can you update this PR with the requested changes and resolve the conflicts? Please request another review, when it's ready.

@matthew-carroll I am trying to figure out how to check the user interaction with different gesture mode. I am making this PR in draft for now or closing this. Will another PR when the comments resolved.

debasmitasarkar: @matthew-carroll I could not find user interaction that we can test for different gestures. If you have something in mind, let's discuss that.

super_editor/test/src/default_editor/super_editor_test.dart
debasmitasarkar: @matthew-carroll I could not find user interaction that we can test for different gestures. If you have something in mind, let's discuss that.

matthew-carroll: I looked into this. I started by asking the question "If there's no difference in behavior, then why do have different gesture modes?". I checked the docs for `AndroidDocumentTouchInteractor` and it says:

```dart
/// Document gesture interactor that's designed for Android touch input, e.g.,
/// drag to scroll, and handles to control selection.
```

So a document scrolls up/down when the user taps and drags up/down. But on all desktop platforms, tapping and dragging should move the user's selection. That's a difference in behavior based on gesture mode, right?

Another difference is that drag handles appear on mobile, but they don't appear on desktop.

debasmitasarkar: @matthew-carroll based on this I am updating the test cases

matthew-carroll:
Closing this PR because I've re-assigned the ticket, and this PR has gone stale.

PR #757, opened 2022-08-23
Ignore pointers in document overlay so that doc selection works (Resolves #756)

### LINKED ISSUE #756: [super_editor] - Ignore pointers in document overlay
The document overlay that we added to the desktop configuration seems to be interfering with document selection in some apps.

Add an `IgnorePointer` to the appropriate place in the document overlay so that selection works as expected.

matthew-carroll:
Ignore pointers in document overlay so that doc selection works (Resolves #756)

@miguelcmedeiros let me know if this PR works for you. I can't verify on my end, because the selection problem doesn't happen in the example app.

### DIFF
--- a/super_editor/lib/src/default_editor/super_editor.dart
+++ b/super_editor/lib/src/default_editor/super_editor.dart
@@ -598,20 +598,24 @@ class SuperEditorState extends State<SuperEditor> {
                 ),
               ),
               // The document that the user is editing.
-              Align(
-                alignment: Alignment.topCenter,
-                child: Stack(
-                  children: [
-                    documentLayout,
-                    // We display overlay builders in this inner-Stack so that they
-                    // match the document size, rather than the viewport size.
-                    for (final overlayBuilder in widget.documentOverlayBuilders)
+              IgnorePointer(
+                ignoring: true,
+                child: Align(
+                  alignment: Alignment.topCenter,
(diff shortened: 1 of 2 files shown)


super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -606,8 +606,10 @@ class SuperEditorState extends State<SuperEditor> {
                     // We display overlay builders in this inner-Stack so that they
                     // match the document size, rather than the viewport size.
                     for (final overlayBuilder in widget.documentOverlayBuilders)
-                      Positioned.fill(
-                        child: overlayBuilder.build(context, editContext),
+                      IgnorePointer(
```
miguelcmedeiros, line 609: This throws an exception since `Positioned ` has to be a child of `Stack`. I tried making `IgnorePointer` the child of `Positioned`, but there was no change with compared to yesterday.

matthew-carroll:
@miguelcmedeiros now the PR ignores the pointer in the `Stack`, but doesn't ignore the `document`. Let me know if this works for you.

matthew-carroll:
We found that the root problem in Superlist was a number of `MouseRegion`s set to `opaque` and a `CustomPainter` that was absorbing hits. No change to `super_editor` is required. Closing this PR.

PR #765, opened 2022-08-31
[SuperEditor] Fix caret display on focus (Resolves #722)

### LINKED ISSUE #722: [SuperEditor] Caret not visible when document has selection and no interaction has taken place
**To Reproduce**
Build a `SuperEditor` with autofocus set to `true`, a non-empty list of nodes and an initial collapsed selection.
The expected behaviour is for the caret to be visible however this isn't the case.

I attempted to figure out why:

[`CaretDocumentOverlay`](https://github.com/superlistapp/super_editor/blob/main/super_editor/lib/src/default_editor/document_caret_overlay.dart#L30) is initialised with a null caret rect. It rebuilds if the composer's [`selectionNotifier`]()https://github.com/superlistapp/super_editor/blob/main/super_editor/lib/src/default_editor/document_caret_overlay.dart#L37 notifies listeners of an update but this doesn't happen with the initial selection so the caret rect is never updated. Adding [`_onSelectionChange();`](https://github.com/superlistapp/super_editor/blob/main/super_editor/lib/src/default_editor/document_caret_overlay.dart#L58) to the end of `initState()` seems to fix the issue but I wasn't able to write a passing test to confirm it.

Failing test:
```dart
testWidgets('displays cursor when autofocus is true', (tester) async {
        // Configure and render a document.
        await tester //
            .createDocument()
            .withSingleParagraph()
            .withInputSource(_inputAndGestureVariants.currentValue!.inputSource)
            .withGestureMode(_inputAndGestureVariants.currentValue!.gestureMode)
            .autoFocus(true)
            .pump();

        // Wait for any post frame callbacks to complete
        await tester.pumpAndSettle();

        expect(find.byKey(primaryCaretKey), findsOneWidget); // "none were found but one was expected"
      }, variant: _inputAndGestureVariants);
```

angelosilvestre:
[SuperEditor] Fix caret display on focus. Resolves #722 

Creating a `SuperEditor` with autofocus set to `true` and with an initial selection wasn't displaying the caret, even though the document has a selection.

This PR changes the caret overlay and touch interactors to call `_onSelectionChange`, so the caret position is updated. It also updates the selection tests to ensure the caret is being displayed.

During manual tests I've found a situation happening only on desktop: when the editor has autofocus, by the time we get to `initState`, the focus node already has focus, so the `DocumentSelectionOnFocusMixin` isn't placing the selection at the end of the document. This PR adds a fix to this case by calling `_onFocusChange` in the mixin if we were given a focus node which already has focus. I couldn't reproduce this behavior during tests.

### DIFF
--- a/super_editor/lib/src/default_editor/document_caret_overlay.dart
+++ b/super_editor/lib/src/default_editor/document_caret_overlay.dart
@@ -36,6 +36,11 @@ class _CaretDocumentOverlayState extends State<CaretDocumentOverlay> with Single
     super.initState();
     widget.composer.selectionNotifier.addListener(_onSelectionChange);
     _blinkController = BlinkController(tickerProvider: this)..startBlinking();
+
+    // If we already have a selection, we need to display the caret.
+    if (widget.composer.selection != null) {
+      _onSelectionChange();
+    }
   }
 
   @override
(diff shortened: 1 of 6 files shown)


matthew-carroll (approved):
LGTM

PR #770, opened 2022-09-01
Allow blockquote mouse gestures by ignoring pointer events on container

### LINKED ISSUE #769: Blockquote does not allow selection interactions on Linux/Web
When hovering over blockquotes the cursor goes from text back to basic, and does not allow for any type of selection interaction, directly on the region.

Super Editor Version: 16fa1e9
Flutter Version: 3.3.0
Platforms: Linux / Web

jpopesculian:
Fixes #769

I just started to dabble into Flutter and the gesture stuff is a little over my head, but I think if I understand correctly, to fix this issue we just need to ignore the pointer events here


matthew-carroll (approved):
LGTM - Thanks

Issue #771, opened 2022-09-01
Example app is broken in Flutter 3.3.0


jmatth:
- The mouse cursor does not stay as a text cursor but switches back to the default cursor. I haven't nailed down the exact cause but it seems to happen after clicking on the document for the first time
- Scrolling on a trackpad in macOS changes the selection instead of scrolling the document. I'm guessing the gesture system needs to be updated to support the new [trackpad gestures][1]

https://user-images.githubusercontent.com/1316184/187991830-590ed1fa-56b4-4f89-af20-cd83e0f60ee9.mov

[1]: https://flutter.dev/go/trackpad-gestures

algusarov:
Not sure if it's an issue due to Flutter 3.3.0 upgrade, but I just built an example app on macOS and opening any menu in a hovering toolbar doesn't work as well currently.

https://user-images.githubusercontent.com/46132428/188264850-5aaaf31c-2652-4f89-9fff-475336da3f81.mov

jmatth:
I think the toolbar issue is unrelated. It's also an issue for me on `3.0.5` and has its own issue at #731.

matthew-carroll:
@angelosilvestre can you prioritize this?

angelosilvestre:
The mouse cursor issue seems to be a flutter bug. 

I filled https://github.com/flutter/flutter/issues/111425

PR #772, opened 2022-09-02
Add trackpad scrolling support for desktop

jpopesculian:
This seems to be a part of #771 

Because trackpad events are registered as Pan events on desktop (it seems on the web these get simulated as scroll events), they end up changing the selection and not actually scrolling. Again, I'm a noob, so let me know if you have any suggestions here!


matthew-carroll:
@angelosilvestre took over this work and added tests in #780, which was merged. So I'm going to close this PR. Thanks.

PR #773, opened 2022-09-02
Allow image mouse gestures by ignoring pointer events

jpopesculian:
Similar to #769, image components seem to block gestures (most noticeably here scrolling). The solution is also the same as #770 to just ignore the pointer events

### DIFF
--- a/super_editor/lib/src/default_editor/image.dart
+++ b/super_editor/lib/src/default_editor/image.dart
@@ -159,15 +159,17 @@ class ImageComponent extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    return Center(
-      child: SelectableBox(
-        selection: selection,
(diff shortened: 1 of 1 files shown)


matthew-carroll:
@jpopesculian we should add tests for this behavior. Do you feel comfortable figuring that out, or should @angelosilvestre take over the PR and add the tests?

jpopesculian:
hey @matthew-carroll i unfortunately don't think i'll have the time this week and would be happy to let @angelosilvestre take over

matthew-carroll:
@angelosilvestre can you take over this PR and add tests?

angelosilvestre:
@matthew-carroll Sure!

matthew-carroll:
I'm closing this PR in favor of #783

Issue #731, opened 2022-08-12
[super_editor] - Example app toolbar dropdowns make the toolbar disappear


matthew-carroll:
In the `SuperEditor` example app, select some text in the document. A popover toolbar appears.

Tap on the node type dropdown, or the alignment dropdown. Notice that the toolbar immediately disappears.

The problem is that the dropdown menus appear in their own route, which steals focus. Upon loss of focus, the IME interactor clears the document's selection. Without a selection, the toolbar has nothing to edit, so the toolbar disappears.

I think this issue is blocked on Flutter. Either, Flutter needs to change these dropdowns to play nicely with focus, or Flutter needs to provide focus tooling that lets us share focus between routes.

I filed https://github.com/flutter/flutter/issues/106923

matthew-carroll:
@angelosilvestre please keep an eye on this issue. After we merge the custom dropdown, let's see if this is resolved.

PR #776, opened 2022-09-08
[SuperEditor] [DemoApp] Fix incorrect request focus (Resolves #775)

### LINKED ISSUE #775: SuperEditor does not release focus when tapping elsewhere
Steps to reproduce:
* Place a `SuperTextField` above `SuperEditor`
* Tap on `SuperEditor` to give focus to it
* Tap on `SuperTextField` to give focus to it

Expected: focus moves from `SuperEditor` to `SuperTextField`
Actual: focus stays in `SuperEditor`

It is possible to reproduce this in `super_editor`'s demo app, by placing a `SuperTextField` above `SuperEditor` in `ExampleEditor`.

angelosilvestre:
[SuperEditor] [DemoApp] Fix incorrect request focus. Resolves #775 

The demo app is always requesting focus to the editor when `_hideEditorToolbar` is called. This is done probably due to reasons explained in https://github.com/superlistapp/super_editor/issues/731

If another focusable widget is placed in the same widget tree, then the editor never releases focus. When taping another focusable widget the following happens:

1. Focus is removed from the editor, which closes the IME connection.
2. Closing the IME connection clears the selection.
3. Changing selection causes  `_hideEditorToolbar` to be called, which requests focus to the editor again.

c866122e52b22c995630d673c2af8258722e04ec changed `SuperEditor` to restore the selection when the editor receives focus, so we are in a loop of losing focus -> clearing selection -> requesting focus -> restoring selection.

Users following the demo app might run into this issue.

This PR changes the demo app to only request focus if a toolbar is present.

### DIFF
--- a/super_editor/example/lib/demos/example_editor/example_editor.dart
+++ b/super_editor/example/lib/demos/example_editor/example_editor.dart
@@ -177,14 +177,14 @@ class _ExampleEditorState extends State<ExampleEditor> {
       // and non-null implies the entry is in the overlay.
       _textFormatBarOverlayEntry!.remove();
       _textFormatBarOverlayEntry = null;
-    }
 
-    // Ensure that focus returns to the editor.
-    //
-    // I tried explicitly unfocus()'ing the URL textfield
-    // in the toolbar but it didn't return focus to the
-    // editor. I'm not sure why.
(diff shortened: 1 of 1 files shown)


matthew-carroll (approved):
LGTM

PR #778, opened 2022-09-10
[SuperEditor] Fix caret misplaced when indenting a list item (Resolves #777)

### LINKED ISSUE #777: Caret misplaced when the indentation of a list item (ordered or unordered) changes
Steps to reproduce:
* Create a list item node (unordered or ordered)
* Press TAB

Expected: List item node gets indented and caret moves to the new start position of the node
Actual: List item node gets indented but the caret remains in the old position

https://user-images.githubusercontent.com/31278849/189396028-a9233274-6be3-47a5-ad53-4bd587d3768e.mov

angelosilvestre:
[SuperEditor] Fix caret misplaced when indenting a list item. Resolves #777

When indenting or unindenting a list item via keyboard (pressing tab/backspace) the node gets indented but the caret remains in the old position, as seen in the video: 

https://user-images.githubusercontent.com/31278849/189396028-a9233274-6be3-47a5-ad53-4bd587d3768e.mov

The issue happens because `CaretDocumentOverlay` isn't being notified that something changed, as it only listens for selection changes.

This PR changes `CaretDocumentOverlay` to update the caret on all document changes.

### DIFF
--- a/super_editor/lib/src/default_editor/document_caret_overlay.dart
+++ b/super_editor/lib/src/default_editor/document_caret_overlay.dart
@@ -23,6 +23,12 @@ class CaretDocumentOverlay extends StatefulWidget {
   final DocumentLayout Function() documentLayoutResolver;
 
   /// The editor's [Document].
+  /// 
+  /// Some operations that affect caret position don't trigger a selection change, e.g.,
+  /// indenting a list item.
+  /// 
+  /// We need to listen to all document changes to update the caret position when these
+  /// operations happen.
   final Document document;
 
   /// The visual style of the caret that this overlay paints.
@@ -40,13 +46,13 @@ class _CaretDocumentOverlayState extends State<CaretDocumentOverlay> with Single
   @override
   void initState() {
     super.initState();
-    widget.composer.selectionNotifier.addListener(_onSelectionChange);
-    widget.document.addListener(_onSelectionChange);
+    widget.composer.selectionNotifier.addListener(_scheduleCaretUpdate);
+    widget.document.addListener(_scheduleCaretUpdate);
     _blinkController = BlinkController(tickerProvider: this)..startBlinking();
 
     // If we already have a selection, we need to display the caret.
     if (widget.composer.selection != null) {
-      _onSelectionChange();
+      _scheduleCaretUpdate();
     }
   }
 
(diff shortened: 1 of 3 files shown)


super_editor/test/src/default_editor/list_items_test.dart
```diff
@@ -233,6 +237,142 @@ void main() {
...
+        );
+
+        // Ensure the list indentation was actually performed.
+        expect(computedOffsetAfterIndent.dx, greaterThan(caretOffsetBeforeIndent.dx));
+
+        // Find the offset at which the caret is currently being displayed.
+        final caretOffsetAfterIndent = SuperEditorInspector.findCaretOffsetInDocument();
+
+        // Ensure the caret is being displayed at the correct position.
+        expect(caretOffsetAfterIndent, offsetMoreOrLessEquals(computedOffsetAfterIndent));
```
matthew-carroll, line 271: Doesn't this 2nd expectation make the first expectation redundant? Also, it doesn't look like you need to place the caret offset in a variable. You can just call `SuperEditorInspector.findCaretOffsetInDocument()` where you need it.

super_editor/test/src/default_editor/list_items_test.dart
```diff
@@ -233,6 +237,142 @@ void main() {
...
+        );
+
+        // Ensure the list indentation was actually performed.
+        expect(computedOffsetAfterUnindent.dx, lessThan(caretOffsetBeforeUnindent.dx));
+
+        // Find the offset at which the caret is currently being displayed.
+        final caretOffsetAfterUnindent = SuperEditorInspector.findCaretOffsetInDocument();
+
+        // Ensure the caret is being displayed at the correct position.
+        expect(caretOffsetAfterUnindent, offsetMoreOrLessEquals(computedOffsetAfterUnindent));
```
matthew-carroll, line 306: Same comments here, and possibly elsewhere, as mentioned in the first test.

super_editor/lib/src/default_editor/document_caret_overlay.dart
```diff
@@ -35,6 +41,7 @@ class _CaretDocumentOverlayState extends State<CaretDocumentOverlay> with Single
   void initState() {
     super.initState();
     widget.composer.selectionNotifier.addListener(_onSelectionChange);
+    widget.document.addListener(_onSelectionChange);
```
matthew-carroll, line 44: We shouldn't call a method called `_onSelectionChange` when the document changes. A document change is not a selection change. Please rename the method accordingly.

super_editor/lib/src/default_editor/document_caret_overlay.dart
```diff
@@ -19,6 +22,9 @@ class CaretDocumentOverlay extends StatefulWidget {
   /// that the current selection can be mapped to an (x,y) offset and a height.
   final DocumentLayout Function() documentLayoutResolver;
 
+  /// The editor's [Document].
```
matthew-carroll, line 25: Please add an explanation about why we need to listen to document changes. It's not obvious why a `Document` is needed.

super_editor/test/src/default_editor/list_items_test.dart
angelosilvestre: The first expectation is to ensure that pressing tab actually indented the list. Without this expectation, if for some reason pressing tab didn't indent the list, the test will still pass.

super_editor/lib/src/default_editor/document_caret_overlay.dart
angelosilvestre: Method renamed.

super_editor/lib/src/default_editor/document_caret_overlay.dart
angelosilvestre: Added a dart doc.

matthew-carroll (approved):
LGTM

PR #783, opened 2022-09-15
[SuperEditor] Fix image scrolling with mouse wheel

angelosilvestre:
[SuperEditor] Fix image scrolling with mouse wheel

Currently, when hovering an image it isn't possible to scroll the document using the mouse wheel.

This PR wraps the image component tree using an `IgnorePointer`. It also wraps the `IgnorePointer` with a `MouseRegion`, so the mouse cursor remains as `basic`. Otherwise, the mouse cursor will be `text`

This PR also adds the ability to customize the inner image of the `ImageComponent`. That way, we can easily "mock" an image during tests.

Initially, I tried implementing a fake `HttpClient` to always return a fixed image, but a relatively large amount of code is required for this to work. Also, even when the fake `HttpClient` is returning the data, the test completes without the image being updated with the content returned by the `HttpClient` (I didn't dig deeper in the `Image` code to investigate why this is happening).

Based on https://github.com/superlistapp/super_editor/pull/773


super_editor/lib/src/default_editor/image.dart
```diff
@@ -150,24 +150,35 @@ class ImageComponent extends StatelessWidget {
...
     this.selectionColor = Colors.blue,
     this.selection,
+    this.imageBuilder,
   }) : super(key: key);
 
   final GlobalKey componentKey;
   final String imageUrl;
   final Color selectionColor;
   final UpstreamDownstreamNodeSelection? selection;
+  final Widget Function(BuildContext context, String imageUrl)? imageBuilder;
```
matthew-carroll, line 160: Please add a comment for why this was added, and what a `null` value means.

super_editor/lib/src/default_editor/image.dart
```diff
@@ -150,24 +150,35 @@ class ImageComponent extends StatelessWidget {
...
   @override
   Widget build(BuildContext context) {
-    return Center(
-      child: SelectableBox(
-        selection: selection,
-        selectionColor: selectionColor,
-        child: BoxComponent(
-          key: componentKey,
-          child: Image.network(
+    final image = imageBuilder != null
```
matthew-carroll, line 164: I think this can be inlined without creating a readability issue

super_editor/test/src/default_editor/image_test.dart
```diff
@@ -0,0 +1,110 @@
+import 'package:flutter/gestures.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter/rendering.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:super_editor/super_editor.dart';
+
+import '../../super_editor/document_test_tools.dart';
+import '../../super_editor/test_documents.dart';
+import '../../test_tools.dart';
+
```
matthew-carroll, line 10: I'm trying to get test files out of `src/default_editor`. Is there a reasonable place under the `super_editor` test directory where this would make sense?

super_editor/test/src/default_editor/image_test.dart
```diff
@@ -0,0 +1,110 @@
...
+      // Ensure the cursor type is 'basic' when hovering the image.
+      expect(RendererBinding.instance.mouseTracker.debugDeviceActiveCursor(1), SystemMouseCursors.basic);
+    });
+  });
+}
+
+/// Pump a SuperEditor containing an image which will render as an 100x100 box
+/// and content big enough to cause the document to be scrollable.
+Future<void> _pumpScaffold(
```
matthew-carroll, line 59: I believe that we use the term "pump scaffold" represent pumping the widgets that surrounds content. This method pumps the whole widget tree. Can you compare this to other recent tests and see if there's a more appropriate name?

super_editor/test/src/default_editor/image_test.dart
```diff
@@ -0,0 +1,110 @@
...
+
+      await _pumpScaffold(tester, scrollController: controller);
+
+      // Ensure the document started without any scrolling.
+      expect(controller.offset, 0.0);
+
+      final pointer = TestPointer(1, PointerDeviceKind.mouse);
+
+      // Hover the image.
+      pointer.hover(tester.getCenter(find.byType(ImageComponent)));
```
matthew-carroll, line 24: Would it make sense to add an inspector method that looks something like this?

```dart
pointer.hover(SuperEditorInspector.getComponentOffset("1", Alignment.center));
```

super_editor/test/src/default_editor/image_test.dart
angelosilvestre: Renamed.

super_editor/test/src/default_editor/image_test.dart
angelosilvestre: Method added.

super_editor/test/src/default_editor/image_test.dart
angelosilvestre: Moved into `supereditor_components_test.dart`

super_editor/lib/src/default_editor/image.dart
angelosilvestre: Dart doc added.

matthew-carroll (approved):
LGTM

PR #784, opened 2022-09-15
[super_editor_markdown] Fix deserialization of empty markdown (Resolves #779)

### LINKED ISSUE #779: deserializeMarkdownToDocument doesn't create an empty node when the markdown is empty
Issue Summary:  
- Super Editor seems to require that there be at least 1 node in the Document, or nothing is selectable
- If you delete all text from a Document with Super Editor, it leaves you with a single Paragraph Node in the Document with contents "", which still works fine.
- deserializeMarkdownToDocument creates a MutableDocument with 0 nodes if it consumes markdown text that is empty, creating a document that doesn't allow a user to click inside.

The work around for this is simple enough - on consumption of MD, override the 0 Node document in any number of ways - but it seems like the SuperEditor MD handling should produce what SuperEditors expect in this scenario.

angelosilvestre:
[super_editor_markdown] Fix deserialization of empty markdown. Resolves #779

Parsing empty markdown results in a document with an empty list of nodes. For the user to be able to interact with the editor, at least one node is required. So, after parsing the empty markdown, the user can't tap the editor and start typing.

This PR changes the markdown parsing to return a document containing an empty paragraph if no nodes were parsed.

### DIFF
--- a/super_editor_markdown/lib/src/markdown.dart
+++ b/super_editor_markdown/lib/src/markdown.dart
@@ -35,7 +35,18 @@ MutableDocument deserializeMarkdownToDocument(
     node.accept(nodeVisitor);
   }
 
-  return MutableDocument(nodes: nodeVisitor.content);
+  final documentNodes = nodeVisitor.content;
+
+  if (documentNodes.isEmpty) {
+    // An empty markdown was parsed.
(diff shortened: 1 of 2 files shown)


matthew-carroll (approved):
LGTM

PR #785, opened 2022-09-18
[SuperEditor] [iOS] Fix list indentation using TAB key (Resolves #745)

### LINKED ISSUE #745: [super_editor] - Tab button on software keyboard should indent list items
On tablets, like the iPad, the software keyboard includes a tab button. This tab button should behave the same way as a physical keyboard, e.g., indent list items. However, the tab key only inserts spaces.

Update the IME handler to correctly apply tabs.

angelosilvestre:
[SuperEditor] [iOS] Fix list indentation using TAB key. Resolves #745 

Pressing the TAB button on the iOS software keyboard isn't indenting list items. The tabs are reported as an insertion delta of `\t` and being inserted as text.

This PR changes the IME handler to trigger the list indentation command when a `\t` is received.

I tested some android tablet simulators and none of them have the tab key, so this command is only triggered on iOS.

### DIFF
--- a/super_editor/test/super_editor/supereditor_keyboard_test.dart
+++ b/super_editor/test/super_editor/supereditor_keyboard_test.dart
@@ -4,6 +4,7 @@ import 'package:flutter_test_robots/flutter_test_robots.dart';
 import 'package:super_editor/super_editor.dart';
 import 'package:super_editor/super_editor_test.dart';
 
+import '../test_tools.dart';
 import 'document_test_tools.dart';
 import 'supereditor_test_tools.dart';
 
@@ -281,6 +282,34 @@ void main() {
       });
     });
   });
+
+  group('SuperEditor software keyboard', () {
+    testWidgetsOnIos('pressing tab indent list', (tester) async {
+      final testContext = await _pumpUnorderedList(tester);
+      final doc = testContext.editContext.editor.document;
+      final nodeId = doc.nodes.first.id;
+
+      // Ensure we started with indentation level 0.
+      expect((doc.nodes.first as ListItemNode).indent, 0);
+
+      await tester.placeCaretInParagraph(nodeId, 0);
+
+      final selectionBeforeTab = SuperEditorInspector.findDocumentSelection();
+      final textBeforeTab = SuperEditorInspector.findTextInParagraph(nodeId);
+
+      // Simulate the user pressing TAB on the software keyboard.
+      await tester.typeImeText("\t");
+
+      // Ensure we indented the list item.
+      expect((doc.nodes.first as ListItemNode).indent, 1);
+
+      // Ensure the selection didn't change.
+      expect(SuperEditorInspector.findDocumentSelection(), selectionBeforeTab);
+
+      // Ensure the content of the list item didn't change.
+      expect(SuperEditorInspector.findTextInParagraph(nodeId), textBeforeTab);
+    });
+  });
 }
 
 Future<String> _pumpSingleLineWithCaret(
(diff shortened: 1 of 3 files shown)


super_editor/test/super_editor/supereditor_keyboard_test.dart
```diff
@@ -281,6 +282,34 @@ void main() {
...
+      final textBeforeTab = SuperEditorInspector.findTextInParagraph(nodeId);
+
+      // Simulate the user pressing TAB on the software keyboard.
+      await tester.typeImeText("\t");
+
+      // Ensure we indented the list item.
+      expect((doc.nodes.first as ListItemNode).indent, 1);
+
+      // Ensure the selection didn't change.
+      expect(SuperEditorInspector.findDocumentSelection(), selectionBeforeTab);
```
matthew-carroll, line 307: Is there a need to do a before/after comparison in this test? It looks like you should be able to expect an exact selection, and exact text, without missing anything..

super_editor/test/super_editor/supereditor_keyboard_test.dart
```diff
@@ -281,6 +282,34 @@ void main() {
...
+      // Ensure we started with indentation level 0.
+      expect((doc.nodes.first as ListItemNode).indent, 0);
+
+      await tester.placeCaretInParagraph(nodeId, 0);
+
+      final selectionBeforeTab = SuperEditorInspector.findDocumentSelection();
+      final textBeforeTab = SuperEditorInspector.findTextInParagraph(nodeId);
+
+      // Simulate the user pressing TAB on the software keyboard.
+      await tester.typeImeText("\t");
```
matthew-carroll, line 301: Can you file an issue with `flutter_test_robots` to add a tab press? Please mention in that ticket that it's for IME, and what we're looking for is `\t`.

super_editor/test/super_editor/supereditor_keyboard_test.dart
```diff
@@ -281,6 +282,34 @@ void main() {
...
   });
+
+  group('SuperEditor software keyboard', () {
+    testWidgetsOnIos('pressing tab indent list', (tester) async {
+      final testContext = await _pumpUnorderedList(tester);
+      final doc = testContext.editContext.editor.document;
+      final nodeId = doc.nodes.first.id;
+
+      // Ensure we started with indentation level 0.
+      expect((doc.nodes.first as ListItemNode).indent, 0);
```
matthew-carroll, line 293: How about we add a method to the inspector called `DocumentNode getNodeAt<T>(int index)`, which returns the node at the given index, cast to the given type. It should throw an exception for a bad index, and it should throw an exception for a bad type.

Since you access the node twice, and you use the node ID twice, you should probably store the node in a local variable and use the node across those sitautions.

super_editor/test/super_editor/supereditor_keyboard_test.dart
angelosilvestre: Updated.

super_editor/test/super_editor/supereditor_keyboard_test.dart
angelosilvestre: Method added.

super_editor/test/super_editor/supereditor_keyboard_test.dart
angelosilvestre: Ok, will do that.

angelosilvestre: Filed https://github.com/Flutter-Bounty-Hunters/flutter_test_robots/issues/9

matthew-carroll (approved):
LGTM

PR #786, opened 2022-09-18
[SuperEditor] Fix selection when tapping beyond end of content (Resolves #744)

### LINKED ISSUE #744: [super_editor] - Tapping beyond end of content should place caret at end of document
**Expected behavior:**
When there's empty space after (or below) the content in a `SuperEditor`, if the user taps the empty space, the caret should be placed at the end of the document's content. In other words, when the user taps at a location beyond the end of the document, the caret should always be placed at the end of the document, regardless of the x-position of the mouse cursor.

**Actual behavior:**
When the user taps in the empty space, the caret is placed in the last line of content, but it's placed at the same horizontal position as the user's tap.

angelosilvestre:
[SuperEditor] Fix selection when tapping beyond end of content. Resolves #744

Tapping at the empty space after the content in `SuperEditor` is placing the caret at the horizontal position where the tap happened.

This PR changes `SuperEditor` to place the caret at the end of the document when a tap happens beyond the end of the content.

### DIFF
--- a/super_editor/lib/src/default_editor/layout_single_column/_layout.dart
+++ b/super_editor/lib/src/default_editor/layout_single_column/_layout.dart
@@ -136,6 +136,15 @@ class _SingleColumnDocumentLayoutState extends State<SingleColumnDocumentLayout>
     );
     editorLayoutLog.info('Getting document position near offset: $documentOffset');
 
+    if (_isBeyondDocumentEnd(documentOffset)) {
+      // The given offset is beyond the end of the content.
+      // Return the position at the end of the last node.
+      final lastPosition = _findLastPosition();
+      if (lastPosition != null) {
+        return lastPosition;
+      }
+    }
+
     final componentKey = _findComponentClosestToOffset(documentOffset);
     if (componentKey == null || componentKey.currentContext == null) {
       return null;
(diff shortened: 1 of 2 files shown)


super_editor/lib/src/default_editor/layout_single_column/_layout.dart
```diff
@@ -136,6 +136,15 @@ class _SingleColumnDocumentLayoutState extends State<SingleColumnDocumentLayout>
     );
     editorLayoutLog.info('Getting document position near offset: $documentOffset');
 
+    if (_isBeyondDocumentEnd(documentOffset)) {
+      // The given offset is beyond the end of the content.
+      // Return the position at the end of the last node.
+      final lastPosition = _findLastPosition();
+      if (lastPosition != null) {
```
matthew-carroll, line 143: What happens if this condition isn't triggered?

super_editor/lib/src/default_editor/layout_single_column/_layout.dart
```diff
@@ -136,6 +136,15 @@ class _SingleColumnDocumentLayoutState extends State<SingleColumnDocumentLayout>
     );
     editorLayoutLog.info('Getting document position near offset: $documentOffset');
 
+    if (_isBeyondDocumentEnd(documentOffset)) {
```
matthew-carroll, line 139: If our problem is in the layout offset lookup, should we also handle the case when we're above the document start?

super_editor/lib/src/default_editor/layout_single_column/_layout.dart
```diff
@@ -456,6 +479,21 @@ class _SingleColumnDocumentLayoutState extends State<SingleColumnDocumentLayout>
     return nearestComponentKey;
   }
 
+  /// Returns the [DocumentPosition] at the end of the last node or `null` if the document is empty.
+  DocumentPosition? _findLastPosition() {
```
matthew-carroll, line 483: Don't we already have a method somewhere that finds the last position in the document?

angelosilvestre: We have a method to find the last selectable position. This method always returns the last position, even if the component is not selectable.

super_editor/lib/src/default_editor/layout_single_column/_layout.dart
angelosilvestre: `_findLastPosition` only returns null if the document doesn't have any node. So in this case there will be no selection.

super_editor/lib/src/default_editor/layout_single_column/_layout.dart
angelosilvestre: Tap above start implemented.

matthew-carroll (approved):
LGTM

PR #789, opened 2022-09-21
Create read-only SuperDocument (Resolves #788)

### LINKED ISSUE #788: [SuperDocument] - Create a read-only document
Create a widget called `SuperDocument`, which is like `SuperEditor`, except that it's non-editable.

Details:

- User should be able to drag a selection and copy content, but the user shouldn't be able to place the caret
- Double/triple tap should select words and paragraphs as with `SuperEditor`.
- Up/down arrows should scroll the document

matthew-carroll:
Creates a new widget called `SuperReader`, which is similar to `SuperEditor`, except that it's a read-only document display.

With `SuperReader`, the user can still select content, copy with the keyboard, use `shift` to expand the selection, and use up/down arrows for scrolling. Collapsed selections aren't supported.

Where possible, I pulled out common behavior from `SuperEditor` and moved it into the `infrastructure` package.

In certain classes I replaced the `EditContext` with individual properties, such as `document`, `documentLayout`, and `selection`. I did this because `EditContext` is a service locator that's designed specifically for the needs of the `SuperEditor` widget. Developers might like to use some of these interactors to build their own editors, and those editors might have their own bundle of dependencies.

The package references are getting a little bit messy. I plan to do a larger refactor after we merge the upcoming breaking changes.


matthew-carroll:
@Jethro87 - Can you give this PR a try in your app? I poked around on Mac, Android, and iOS and at first glance, `SuperDocument` seems to work as I expect it to.

We still need to fill out the test suite, and we should probably refactor a lot of shared behavior with `SuperEditor`. But before we start locking things down, I'd like to know if this widget's behavior does what you need it to.

DmitrySboychakov:
SuperDocument is a bit confusing name, wouldn't it better to call it SuperReader or something like that?

Jethro87:
@matthew-carroll The general functionality has worked well in my testing and is what we require.

I found that double tapping at the beginning of a blank node (I was experimenting with `TextWithHintComponent`), or at the end of a node, will bring up the blinking carat. It's not possible to type, but the carat remains there and blinking. This should not happen.

Also, there was some talk in #303 and in our emails about easily switching between a read-only and editable document. It wouldn't make sense to do this on tap if we're allowing users to select content, but I'm curious if you're considering this functionality in this PR or in a future PR.

Jethro87:
Also, the copy/paste dialog doesn't appear when selecting text on mobile.

matthew-carroll:
@Jethro87 those issues should now be fixed. Let me know if you find any others.

> Also, there was some talk in https://github.com/superlistapp/super_editor/discussions/303 and in our emails about easily switching between a read-only and editable document. It wouldn't make sense to do this on tap if we're allowing users to select content, but I'm curious if you're considering this functionality in this PR or in a future PR.

I'm not sure that a toggle-able read-only mode is a good idea in the long term. I expect that over time, the behavior expected for read-only and editing will diverge further and further. From an implementation perspective, it isn't really a mode or a toggle. Given that developers can include whatever keyboard actions they want in `SuperEditor`, and given that `SuperEditor` is based heavily on a `DocumentEditor` object and a `DocumentComposer`, neither of which make sense for read-only documents, I'm not sure it's a good idea to try to force `SuperEditor` into a read-only mode. If you see a path, I'm happy to consider it. But it's not obvious to me how we'd accomplish a read-only `SuperEditor` mode without creating many other problems that we'd rather not have.

matthew-carroll:
@DmitrySboychakov - the name `SuperReader` seems reasonable. I'll refactor the name after everything is working as desired.

Jethro87:
@matthew-carroll Sounds good re: toggle-able read-only mode.

Just noticed that the blinking carat still appears when triple-tapping on an empty node.

matthew-carroll:
@Jethro87 the triple tap should now be fixed. Let me know if there's anything else.

matthew-carroll:
@Jethro87 are there any other changes that you need for read-only documents?

Jethro87:
@matthew-carroll Just tested again. All looks good 👍

super_editor/lib/src/super_reader/read_only_document_keyboard_interactor.dart
```diff
@@ -0,0 +1,595 @@
...
+    }
+
+    if (defaultTargetPlatform == TargetPlatform.windows && keyEvent.isAltPressed) {
+      return ExecutionInstruction.continueExecution;
+    }
+
+    if (defaultTargetPlatform == TargetPlatform.linux &&
+        keyEvent.isAltPressed &&
+        (keyEvent.logicalKey == LogicalKeyboardKey.arrowUp || keyEvent.logicalKey == LogicalKeyboardKey.arrowDown)) {
+      return ExecutionInstruction.continueExecution;
```
angelosilvestre, line 300: Same here.

super_editor/lib/src/super_reader/super_reader.dart
```diff
@@ -0,0 +1,604 @@
...
+  /// All actions that this editor takes in response to key
+  /// events, e.g., text entry, newlines, character deletion,
+  /// copy, paste, etc.
+  ///
+  /// These actions are only used when in [DocumentInputSource.keyboard]
+  /// mode.
+  final List<ReadOnlyDocumentKeyboardAction> keyboardActions;
+
+  /// The [SuperEditor] input source, e.g., keyboard or Input Method Engine.
+  final DocumentInputSource? inputSource;
```
angelosilvestre, line 125: I think we are not checking for the `inputSource` when creating the input interactor.

super_editor/lib/src/super_reader/super_reader.dart
```diff
@@ -0,0 +1,604 @@
...
+
+  /// The [SuperEditor] gesture mode, e.g., mouse or touch.
```
angelosilvestre, line 127: [SuperEditor] -> [SuperReader]

super_editor/lib/src/super_reader/super_reader.dart
```diff
@@ -0,0 +1,604 @@
...
+          componentBuilders: widget.componentBuilders,
+          showDebugPaint: widget.debugPaint.layout,
+        ),
+      ),
+    );
+  }
+
+  /// Builds the widget tree that applies user input, e.g., key
+  /// presses from a keyboard, or text deltas from the IME.
+  Widget _buildInputSystem({
```
angelosilvestre, line 291: I think we don't handle text deltas in `SuperReader`.

super_editor/lib/src/test/super_reader_test/super_reader_inspector.dart
```diff
@@ -0,0 +1,166 @@
...
+  ///
+  /// The given [index] must be a valid node index inside the [Document].The node at [index]
+  /// must be of type [NodeType].
+  ///
+  /// {@macro super_document_finder}
+  static NodeType getNodeAt<NodeType extends DocumentNode>(int index, [Finder? superDocumentFinder]) {
+    final doc = findDocument(superDocumentFinder);
+
+    if (doc == null) {
+      throw Exception('SuperEditor not found');
```
angelosilvestre, line 136: SuperEditor -> SuperReader

super_editor/lib/src/test/super_reader_test/super_reader_robot.dart
```diff
@@ -0,0 +1,202 @@
...
+    await gesture.up();
+    await gesture.removePointer();
+    await pumpAndSettle();
+  }
+
+  /// Types the given [text] into a [SuperReader] by simulating IME text deltas from
+  /// the platform.
+  Future<void> typeImeText(String text) async {
+    await ime.typeText(text, getter: imeClientGetter);
+  }
```
angelosilvestre, line 190: Isn't `typeImeText` unnecessary for a read-only document?

super_editor/test/super_reader/reader_test_tools.dart
```diff
@@ -0,0 +1,482 @@
...
+import 'package:super_editor/src/default_editor/document_scrollable.dart';
+import 'package:super_editor/super_editor.dart';
+import 'package:super_editor/super_editor_test.dart';
+import 'package:super_editor_markdown/super_editor_markdown.dart';
+import 'package:text_table/text_table.dart';
+
+import 'test_documents.dart';
+
+/// Extensions on [WidgetTester] that configure and pump [SuperEditor]
+/// document editors.
```
angelosilvestre, line 16: [SuperEditor] -> [SuperReader]

super_editor/test/super_reader/reader_test_tools.dart
```diff
@@ -0,0 +1,482 @@
...
+      theme: _appTheme,
+      home: Scaffold(
+        body: superEditor,
+      ),
+    );
+  }
+}
+
+/// Must return a widget tree containing the given [superEditor]
+typedef WidgetTreeBuilder = Widget Function(Widget superEditor);
```
angelosilvestre, line 301: superEditor -> superReader

super_editor/test/super_reader/reader_test_tools.dart
```diff
@@ -0,0 +1,482 @@
...
+    if (target is Document) {
+      actualDocument = target;
+    } else {
+      // If we weren't given a Document, then we expect to receive a Finder
+      // that locates a SuperEditor, which contains a Document.
+      if (target is! Finder) {
+        return "the given target isn't a Document or a Finder: $target";
+      }
+
+      final document = SuperEditorInspector.findDocument(target);
```
angelosilvestre, line 364: I think it should be a `SuperReaderInspector` here.

super_editor/test/super_reader/reader_test_tools.dart
```diff
@@ -0,0 +1,482 @@
...
```
angelosilvestre, line 429: I think it should be a `SuperReaderInspector` here.

super_editor/test/super_reader/reader_test_tools_test.dart
```diff
@@ -0,0 +1,58 @@
+import 'package:flutter_test/flutter_test.dart';
+import 'package:super_editor/super_reader_test.dart';
+
+import '../test_tools.dart';
+import 'reader_test_tools.dart';
+
+void main() {
+  group("SuperEditor test tools", () {
```
angelosilvestre, line 8: SuperEditor -> SuperReader

super_editor/test/super_reader/super_reader_keyboard_test.dart
```diff
@@ -0,0 +1,321 @@
...
+import 'package:flutter_test/flutter_test.dart';
+import 'package:flutter_test_robots/flutter_test_robots.dart';
+import 'package:super_editor/super_editor.dart';
+import 'package:super_editor/super_reader_test.dart';
+
+import '../test_tools.dart';
+import 'reader_test_tools.dart';
+
+void main() {
+  group('SuperDocument keyboard', () {
```
angelosilvestre, line 11: SuperDocument -> SuperReader

super_editor/test/super_reader/super_reader_scrolling_test.dart
```diff
@@ -0,0 +1,222 @@
...
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:super_editor/super_reader_test.dart';
+import 'package:super_editor/super_editor.dart';
+
+import '../test_tools.dart';
+import 'reader_test_tools.dart';
+
+void main() {
+  group("SuperDocument scrolling", () {
```
angelosilvestre, line 11: SuperDocument -> SuperReader

super_editor/test/super_reader/super_reader_selection_test.dart
```diff
@@ -0,0 +1,514 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:super_editor/super_reader_test.dart';
+import 'package:super_editor/super_editor.dart';
+
+import '../test_tools.dart';
+import 'reader_test_tools.dart';
+
+void main() {
+  group("SuperDocument selection", () {
```
angelosilvestre, line 10: SuperDocument -> SuperReader

super_editor/example/lib/main.dart
```diff
@@ -285,6 +286,13 @@ final _menu = <_MenuGroup>[
   _MenuGroup(
     title: 'READ-ONLY DOCS',
     items: [
+      _MenuItem(
+        icon: Icons.text_snippet,
+        title: 'SuperDocument',
```
angelosilvestre, line 291: SuperDocument -> SuperReader

angelosilvestre (commented):
Added some comments.

super_editor/lib/src/document_operations/selection_operations.dart
```diff
@@ -0,0 +1,679 @@
...
+      // We're at the beginning/end of the document and can't go
+      // anywhere.
+      return false;
+    }
+
+    newExtentNodeId = nextNode.id;
+    final nextComponent = documentLayout.getComponentByNodeId(nextNode.id);
+    if (nextComponent == null) {
+      throw Exception('Could not find next component to move the selection horizontally. Next node ID: ${nextNode.id}');
+    }
```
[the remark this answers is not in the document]
matthew-carroll: I added an exception to `moveCaretUpstream`. This condition indicates that there's a problem with the document layout. Every node in the document should have a component in the layout.

super_editor/lib/src/super_reader/read_only_document_ios_touch_interactor.dart
```diff
@@ -0,0 +1,1010 @@
...
+  }
+
+  void _ensureSelectionExtentIsVisible() {
+    editorGesturesLog.fine("Ensuring selection extent is visible");
+    final collapsedHandleOffset = _editingController.collapsedHandleOffset;
+    final extentHandleOffset = _editingController.downstreamHandleOffset;
+    if (collapsedHandleOffset == null && extentHandleOffset == null) {
+      // There's no selection. We don't need to take any action.
+      return;
+    }
```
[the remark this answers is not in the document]
matthew-carroll: Looks like that's handled in an if-statement further down.

super_editor/lib/src/super_reader/super_reader.dart
matthew-carroll: I got rid of it

PR #802, opened 2022-10-03
[SuperEditor] Fix default gesture mode (Resolves #741)

### LINKED ISSUE #741: [super_editor] - The default gesture mode should be based on the platform

angelosilvestre:
[SuperEditor] Fix default gesture mode. Resolves #741

`SuperEditor` is using `mouse` as a default `gestureMode`.

There is already a getter named `_gestureMode`, which chooses the gesture mode based on the target platform if there isn't a `gestureMode` configured.

This PR just removes the default `gestureMode`. Also, removes the `_defaultGestureMode` from the test tools, as the editor will be configured automatically.


super_editor/test/super_editor/super_editor_configuration_test.dart
```diff
@@ -0,0 +1,61 @@
...
+import 'package:flutter_test/flutter_test.dart';
+import 'package:super_editor/src/test/super_editor_test/supereditor_inspector.dart';
+import 'package:super_editor/src/test/super_editor_test/supereditor_robot.dart';
+
+import '../test_tools.dart';
+import 'document_test_tools.dart';
+
+void main() {
+  group('SuperEditor', () {
+    testWidgetsOnAndroid('configures default gesture mode (on Android)', (tester) async {
```
matthew-carroll, line 11: Can you describe why you believe these tests belong in a new file? What about the existing gestures test suite?

Creating a test file for all "configurations" seems like it will bring together a bunch of otherwise unrelated behaviors, which could make it more difficult to locate desired tests.

super_editor/test/super_editor/super_editor_configuration_test.dart
```diff
@@ -0,0 +1,61 @@
...
+          .createDocument()
+          .withSingleParagraph()
+          .withAndroidToolbarBuilder((_) => SizedBox(key: toolbarKey))
+          .pump();
+
+      // Double tap to show the toolbar.
+      await tester.doubleTapInParagraph(SuperEditorInspector.findDocument()!.nodes.first.id, 0);
+
+      // Ensure the toolbar is displayed.
+      expect(find.byKey(toolbarKey), findsOneWidget);
```
matthew-carroll, line 24: Wouldn't it make more sense to look for drag handles than a toolbar? Toolbars are optional, but drag handles always exist.

super_editor/test/super_editor/super_editor_configuration_test.dart
angelosilvestre: I was not sure if I should add these tests to the gestures test suite. I moved the tests now.

super_editor/test/super_editor/super_editor_configuration_test.dart
angelosilvestre: Updated.

super_editor/test/super_editor/supereditor_gestures_test.dart
```diff
@@ -451,5 +451,45 @@ spans multiple lines.''',
...
+      await tester //
+          .createDocument()
+          .withSingleParagraph()
+          .pump();
+
+      // Double tap to show the drag handle.
+      await tester.doubleTapInParagraph(SuperEditorInspector.findDocument()!.nodes.first.id, 0);
+
+      // Ensure the drag handle is displayed.
+      // There are two AndroidSelectionHandle's. I'm not sure why.
```
matthew-carroll, line 465: There are two because you double tapped. You selected a range of text. Is there any reason that you didn't use a single tap to place a caret with one handle?

angelosilvestre: Makes sense. I update to use `placeCaretInParagraph` .

matthew-carroll (approved):
LGTM

PR #803, opened 2022-10-04
[SuperTextField] Expose padding property (Resolves #588)

### LINKED ISSUE #588: Expose padding property in SuperTextField
We need to add/expose `padding` property to `SuperTextField`. 

`SuperDesktopTextField` already has [padding](https://github.com/superlistapp/super_editor/blob/d65376dce5cbfd0236f8406db3a1c922527012db/super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart#L87) but it's not available through `SuperTextField` public API.

Padding should be tappable for caret placement.

**With padding**

https://user-images.githubusercontent.com/4456735/170072322-d1633053-4f2d-43e8-a840-913aa1b974b3.mp4

**Without padding**

https://user-images.githubusercontent.com/4456735/170071896-8e029ae9-e5de-4640-a1c0-2d24ab7b999c.mp4

angelosilvestre:
[SuperTextField] Expose padding property. Resolves #588 

Currently `SuperDesktopTextField` exposes a `padding` property, but `SuperTextField` don't.

This PR exposes the  `padding` property at `SuperTextField` public API and implements padding support in `SuperAndroidTextField` and `SuperIOSTextField`.

### DIFF
--- a/super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
+++ b/super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
@@ -354,19 +354,19 @@ class _TextScrollViewState extends State<TextScrollView>
     final estimatedContentHeight = (linesOfText * estimatedLineHeight) + totalVerticalPadding;
     _log.finer(' - estimated content height: $estimatedContentHeight');
 
-    final minContentHeight = (widget.minLines != null //
+    final minContentHeight = widget.minLines != null //
         ? widget.minLines! * estimatedLineHeight
-        : estimatedLineHeight); // Can't be shorter than 1 line.
+        : estimatedLineHeight; // Can't be shorter than 1 line.
 
     final minHeight = minContentHeight + totalVerticalPadding;
 
-    final maxContentHeight = (widget.maxLines != null //
+    final maxContentHeight = widget.maxLines != null //
         ? (widget.maxLines! * estimatedLineHeight) //
-        : null);
+        : null;
 
-    final maxHeight = (maxContentHeight != null //
+    final maxHeight = maxContentHeight != null //
         ? maxContentHeight + totalVerticalPadding
-        : null);
+        : null;
 
     _log.finer(' - minHeight: $minHeight, maxHeight: $maxHeight');
 
--- a/super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
+++ b/super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
@@ -33,6 +33,7 @@ class TextScrollView extends StatefulWidget {
     this.perLineAutoScrollDuration = Duration.zero,
     this.showDebugPaint = false,
     this.textAlign = TextAlign.left,
+    this.padding,
     required this.child,
   }) : super(key: key);
 
@@ -92,6 +93,9 @@ class TextScrollView extends StatefulWidget {
 
   /// The text alignment within the scrollview.
   final TextAlign textAlign;
+
+  /// Padding placed around the text content of this text field.
+  final EdgeInsets? padding;
 
   /// The child widget.
   final Widget child;
@@ -344,15 +348,18 @@ class _TextScrollViewState extends State<TextScrollView>
       }
     }
 
-    final estimatedContentHeight = linesOfText * estimatedLineHeight;
+    final verticalPadding = widget.padding?.vertical ?? 0.0;
+
+    final estimatedContentHeight = (linesOfText * estimatedLineHeight) + verticalPadding;
     _log.finer(' - estimated content height: $estimatedContentHeight');
 
-    final minHeight = widget.minLines != null
-        ? widget.minLines! * estimatedLineHeight
-        : estimatedLineHeight; // Can't be shorter than 1 line
-    final maxHeight = widget.maxLines != null //
-        ? widget.maxLines! * estimatedLineHeight //
-        : null;
+    final minHeight = (widget.minLines != null //
+            ? widget.minLines! * estimatedLineHeight
+            : estimatedLineHeight) + // Can't be shorter than 1 line
+        verticalPadding;
+    final maxHeight = (widget.maxLines != null //
+        ? (widget.maxLines! * estimatedLineHeight) + verticalPadding //
+        : null);
     _log.finer(' - minHeight: $minHeight, maxHeight: $maxHeight');
 
     double? viewportHeight;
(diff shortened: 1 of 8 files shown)


super_editor/test/super_textfield/super_textfield_gestures_test.dart
```diff
@@ -44,6 +44,80 @@ void main() {
...
+    group('tapping at padding', () {
+      testWidgetsOnAllPlatforms('at left places the caret', (tester) async {
+        await _pumpTestApp(
+          tester,
+          padding: const EdgeInsets.only(left: 20),
+          text: '',
+        );
+
+        // Tap in a place inside the padding.
+        await tester.tapAt(tester.getTopLeft(find.byType(SuperTextField)) + const Offset(18, 1));
```
matthew-carroll, line 56: The offset that you apply seems strange. You're moving the touch point more to the right than you are moving it down. Isn't this offset closer to a test of the top padding than the left padding? I would think that you'd want to tap in the vertical center, and maybe one pixel in from the left edge, rather than 18 pixels...

super_editor/test/super_textfield/super_textfield_gestures_test.dart
```diff
@@ -226,14 +300,17 @@ void main() {
 Future<void> _pumpTestApp(
   WidgetTester tester, {
   AttributedTextEditingController? controller,
+  EdgeInsets? padding,
+  String text = 'abc',
```
matthew-carroll, line 304: I don't think a reader is going to expect a text value when none is provided. Let's either default to empty, or make this a required property.

super_editor/test/super_textfield/super_textfield_test.dart
```diff
@@ -201,6 +201,31 @@ void main() {
         expect(_isCaretPresent(tester), isTrue);
       });
     });
+
+    group('padding', () {
+      testWidgetsOnAllPlatforms('is applied when configured', (tester) async {
```
matthew-carroll, line 206: I'm not sure I see why this test is useful. Can you describe a scenario where this test would fail, but all others would pass?

super_editor/test/super_textfield/super_textfield_gestures_test.dart
```diff
@@ -44,6 +44,80 @@ void main() {
       });
     });
 
+    group('tapping at padding', () {
+      testWidgetsOnAllPlatforms('at left places the caret', (tester) async {
+        await _pumpTestApp(
+          tester,
+          padding: const EdgeInsets.only(left: 20),
+          text: '',
```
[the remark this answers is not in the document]
angelosilvestre: Updated.

super_editor/test/super_textfield/super_textfield_gestures_test.dart
angelosilvestre: Updated.

super_editor/test/super_textfield/super_textfield_test.dart
angelosilvestre: The other tests ensure that tapping where it's supposed to have padding will place the caret. This test ensures that the padding was actually applied.

super_editor/test/super_textfield/super_textfield_gestures_test.dart
angelosilvestre: I removed the parameter, as now the tests don't use an empty field.

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -344,15 +349,25 @@ class _TextScrollViewState extends State<TextScrollView>
...
     }
 
-    final estimatedContentHeight = linesOfText * estimatedLineHeight;
+    final totalVerticalPadding = widget.padding?.vertical ?? 0.0;
+
+    final estimatedContentHeight = (linesOfText * estimatedLineHeight) + totalVerticalPadding;
     _log.finer(' - estimated content height: $estimatedContentHeight');
 
-    final minHeight = widget.minLines != null
+    final minContentHeight = (widget.minLines != null //
```
matthew-carroll, line 357: In all 3 assignments in this method, there doesn't appear to be any need for the outermost parentheses.

super_editor/test/super_textfield/super_textfield_gestures_test.dart
```diff
@@ -44,6 +44,80 @@ void main() {
...
+
+        final finder = find.byType(SuperTextField);
+        // Tap in a place inside the padding.
+        await tester.tapAt(tester.getBottomRight(finder) - Offset(tester.getSize(finder).width / 2, 1));
+        await tester.pumpAndSettle();
+
+        // Ensure caret was placed.
+        expect(
+          SuperTextFieldInspector.findSelection(),
+          const TextSelection.collapsed(offset: 3),
```
matthew-carroll, line 98: I'm noticing that tapping in the horizontal center is placing the caret at the end. Do you know why that is? I would assume that the horizontal center is either just before or after the 2nd character, so I would expect an offset of 1 or 2, but not 3...

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
angelosilvestre: Updated

super_editor/test/super_textfield/super_textfield_gestures_test.dart
angelosilvestre: I think the text field is wider than its content, so its center is after the end of the text.

matthew-carroll: it's probably a good idea to create situations that are different between middle selection and end selection. Imagine if our code interpreted a tap below the bottom of the text as an intention to place the caret at the end of the text (like a document). We wouldn't catch that because these tests are looking for the same result for horizontal middle taps, and end taps.

angelosilvestre: I changed to use a center-aligned text field. This way we can simulate a tap at the middle of the text.

super_editor/test/super_textfield/super_textfield_gestures_test.dart
```diff
@@ -44,6 +44,82 @@ void main() {
...
+
+        final finder = find.byType(SuperTextField);
+        // Tap in a place inside the padding.
+        await tester.tapAt(tester.getTopLeft(finder) + Offset(tester.getSize(finder).width / 2, 1));
+        await tester.pumpAndSettle();
+
+        // Ensure caret was placed.
+        expect(
+          SuperTextFieldInspector.findSelection(),
+          const TextSelection.collapsed(offset: 3),
```
matthew-carroll, line 80: This test is still expecting the end of the text, instead of the middle.

super_editor/test/super_textfield/super_textfield_gestures_test.dart
```diff
@@ -44,6 +44,82 @@ void main() {
...
+        );
+      });
+
+      testWidgetsOnAllPlatforms('on the bottom', (tester) async {
+        await _pumpTestApp(
+          tester,
+          padding: const EdgeInsets.only(bottom: 20),
+          textAlign: TextAlign.center,
```
matthew-carroll, line 88: Please add a comment about why you're centering the text.

super_editor/test/super_textfield/super_textfield_gestures_test.dart
```diff
@@ -44,6 +44,82 @@ void main() {
...
+      });
+
+      testWidgetsOnAllPlatforms('on the right side', (tester) async {
+        await _pumpTestApp(
+          tester,
+          padding: const EdgeInsets.only(right: 20),
+        );
+
+        final finder = find.byType(SuperTextField);
+        // Tap in a place inside the padding.
```
matthew-carroll, line 111: Each of these tests says "Tap in a place inside the padding" and yet they're all tapping at different locations. Wouldn't it be more useful to say where you're tapping?

PR #805, opened 2022-10-05
[SuperEditor] Fix list item style when converting to paragraph and back again (Resolves #795)

### LINKED ISSUE #795: Converting from a ListItemNode to a normal paragraph (or something else) and back doesn't respect Stylesheet styles
**Reproduction:**
Run the sample editor Flutter app with the example document. Upon opening the example document, converting the unordered list items using the Convert To Paragraph button, and then back to a ListItemNode using the unordered list button causes the padding to be different from what it originally was. 

I've tested this using different configurations and styles, and this problem persists in other similar scenarios as well (such as converting the ListItemNode directly to a Header, and then back), which leads me to believe it is a bug much lower level in the editor code.

Just to clarify further, the expected behavior is that converting from a ListNode with an arbitrary style configuration to another node type or block type, and then back should produce the same ListNode and style configuration, but the actual behaviour is that the style configuration after the reverse operation is not the same. This is most noticable if the text color is changed, then it is clear that that first conversion into a ListNode does not respect the style configuration.

An additional note, its just that initial 'undo-ing', or the first conversion from another Node to a ListNode, that does not respect style configurations. Pressing enter while on a ListNode (which takes you to a new line and produces a new ListNode) respects the styling properly.

angelosilvestre:
[SuperEditor] Fix list item style when converting to paragraph and back again. Resolves #795 

Converting a list item to a paragraph and back again wasn't respecting the styles for a list item.

The cause is that only the `ListItemNode` named constructors are setting the `blockType` metadata to `listItem`, and `ConvertParagraphToListItemCommand` uses the unnamed constructor. 

As the `StyleRule` for the list items uses a `listItem` block selector, this rule isn't picked.

Deserializing from markdown might also be affected, as it also uses the unnamed constructor.

This PR changes `ListItemNode` unnamed constructor to set the `blockType` if it isn't set.

### DIFF
--- a/super_editor/test/src/default_editor/list_items_test.dart
+++ b/super_editor/test/src/default_editor/list_items_test.dart
@@ -239,8 +239,7 @@ void main() {
         Platform.setTestInstance(null);
       });
 
-      testWidgetsOnArbitraryDesktop(
-          "respects stylesheet when converting unordered list item to paragraph and back again",
+      testWidgetsOnArbitraryDesktop("applies styles when unordered list item is converted to and from a paragraph",
           (WidgetTester tester) async {
         final testContext = await _pumpUnorderedList(
           tester,
@@ -284,7 +283,7 @@ void main() {
         expect(richText.text.style!.color, Colors.blue);
       });
 
-      testWidgetsOnArbitraryDesktop("respects stylesheet when converting ordered list item to paragraph and back again",
+      testWidgetsOnArbitraryDesktop("applies styles when ordered list item is converted to and from a paragraph",
           (WidgetTester tester) async {
         final testContext = await _pumpOrderedList(
           tester,
--- a/super_editor/test/src/default_editor/list_items_test.dart
+++ b/super_editor/test/src/default_editor/list_items_test.dart
@@ -1,9 +1,11 @@
+import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:flutter_test_robots/flutter_test_robots.dart';
 import 'package:super_editor/src/infrastructure/platform_detector.dart';
 import 'package:super_editor/super_editor.dart';
 import 'package:super_editor/super_editor_test.dart';
+import 'package:super_text_layout/super_text_layout.dart';
 
 import '../../super_editor/document_test_tools.dart';
 import '../../test_tools.dart';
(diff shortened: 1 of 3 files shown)


super_editor/test/src/default_editor/list_items_test.dart
```diff
@@ -236,6 +238,95 @@ void main() {
 
         Platform.setTestInstance(null);
       });
+
+      testWidgetsOnArbitraryDesktop(
+          "respects stylesheet when converting unordered list item to paragraph and back again",
```
matthew-carroll, line 243: Future developers won't know what "respects stylesheet" implies. Can this be expressed as a concrete detail? If I convert from list item to paragraph back to a list item, without the fix, what do I see?

super_editor/test/super_editor/document_test_tools.dart
```diff
@@ -217,7 +217,7 @@ class TestDocumentConfigurator {
   }
 
   /// Configures the [SuperEditor] to use the given [stylesheet].
-  TestDocumentConfigurator useStylesheet(Stylesheet stylesheet) {
+  TestDocumentConfigurator useStylesheet(Stylesheet? stylesheet) {
```
matthew-carroll, line 220: Why is this nullable?

angelosilvestre: I changed it to be nullable so the test pump methods with optional `Stylesheet` parameters can call `useStylesheet` without checking for null first. For example:

```dart
Future<TestDocumentContext> _pumpOrderedList(
  WidgetTester tester, {
  Stylesheet? styleSheet,
}) async {
  return await tester 
      .createDocument()
      .fromMarkdown('')
      .useStylesheet(styleSheet) // stylesheet is nullable
      .pump();
}
```

There are various `with*`  methods in the `TestDocumentConfigurator` which accept null, so I thought it was ok to change the `useStylesheet` method.

super_editor/test/src/default_editor/list_items_test.dart
angelosilvestre: Do you think "applies listItem stylesheet rule when converting unordered list item to paragraph and back again" is a good enough description?

super_editor/test/super_editor/document_test_tools.dart
matthew-carroll: Ok, makes sense

super_editor/test/src/default_editor/list_items_test.dart
matthew-carroll: How about "applies styles when unordered list item is converted to and from a paragraph"?

angelosilvestre: Updated

matthew-carroll (approved):
LGTM

tanzilzubair:
First off, thank you for such a quick fix! Secondly, I’d just like to mention, deserialising from markdown is indeed affected, as I was testing that package a few days back, and upon deserialisation of the document, the list items had no style.

matthew-carroll:
@tanzilzubair are you saying that there's still a markdown bug? Or there was a markdown bug, and now it's fixed?

matthew-carroll:
Please re-verify those behaviors with the current version on `main`. We merged in a fix, so if there's still a problem then we need to look at it again.

tanzilzubair:
Yep, I plan to do that soon. Just to check though, there are no breaking changes or major differences between the release and latest version of the package on `main` right? Just fixes and smaller changes?

matthew-carroll:
There are lots of changes. I imagine some of them are breaking changes, but I'm not sure.

tanzilzubair:
I tested it a bit and the parsing seems to be fixed. Also, I don't know if this is the right place to say this but there's definitely some changed behaviour with the `DocumentComposer` class and how it works, but the commits from the release version till now don't really have that as a changed file I think.

For example, I have an app that listens directly to a `DocumentComposer` in a `ListenableBuilder`, which immediately stopped working after updating. I had to use `DocumentComposer.selectionNotifier` instead to fix that, but even then, it was no longer properly informing of changes to name attributions anymore, when I hovered the cursor over text on iOS using the Spacebar-Trackpad functionality. (I am using this to be able to highlight the different styling icons on a modified version of the `KeyboardEditingToolbar`. So for example, in the previous working version, if I hovered over text that was bold, the bold icon would change color). 

I'm happy to open this as a separate issue if needed, or describe it as much as I can (even after hours of testing I'm still not sure where exactly the problem is) in the appropriate channels, so do let me know. As for now, there's too many breaking changes that I believe shouldn't be there, and I am not able to find solutions for most of them, so I'm downgrading again.

matthew-carroll:
Yes, please file a new issue and include repro steps and a screenshot or video.

Breaking changes are something that happens. But a breaking change is different from a bug. So if you're finding bugs, you should file those issues, otherwise they probably won't be fixed.

PR #807, opened 2022-10-05
Fix issues found when auditing example app with Flutter 3.3.x

matthew-carroll:
I audited the example app looking for issues after upgrading to Flutter 3.3.x.

I found some scrolling issues that seem to be caused by Flutter 3.3.x and I filed an issue or two with Flutter. I don't think there's anything we can do until Flutter addresses those issues.

I found that https://github.com/flutter/flutter/issues/111425 seems to be impacting our I-beam text cursor throughout SuperEditor. That's also outside our control.

There didn't seem to be any Flutter changes in 3.3.x that caused problems in SuperEditor that we can fix, but I did find and fix some pre-existing issues. Those fixes are in this PR.


angelosilvestre (approved):
LGTM

PR #810, opened 2022-10-05
[SuperEditor] Fix default input source (Resolves #740)

### LINKED ISSUE #740: [super_editor] - The default input source should be based on the platform
Current, `SuperEditor` always configures itself for keyboard input, regardless of platform.

`SuperEditor` should default to keyboard input for Mac, Windows, Linux, and Web. But it should default to IME for Android and iOS.

Eventually, the default will always be IME, once we've proven stability.

angelosilvestre:
[SuperEditor] Fix default input source. Resolves #740

SuperEditor is using `keyboard` as its default `inputSource`.

This PR removes the default `inputSource` and changes the editor to choose the input source based on the platform if the editor's `inputSource` isn't configured.

### DIFF
--- a/super_editor/lib/src/default_editor/super_editor.dart
+++ b/super_editor/lib/src/default_editor/super_editor.dart
@@ -1,5 +1,5 @@
 import 'package:attributed_text/attributed_text.dart';
-import 'package:flutter/foundation.dart' show defaultTargetPlatform;
+import 'package:flutter/foundation.dart' show defaultTargetPlatform, kIsWeb;
 import 'package:flutter/material.dart' hide SelectableText;
 import 'package:super_editor/src/core/document.dart';
 import 'package:super_editor/src/core/document_composer.dart';
@@ -145,7 +145,7 @@ class SuperEditor extends StatefulWidget {
     this.customStylePhases = const [],
     List<ComponentBuilder>? componentBuilders,
     SelectionStyles? selectionStyle,
-    this.inputSource = DocumentInputSource.keyboard,
+    this.inputSource,
     this.gestureMode,
     List<DocumentKeyboardAction>? keyboardActions,
     this.softwareKeyboardHandler,
@@ -210,7 +210,7 @@ class SuperEditor extends StatefulWidget {
   final List<SingleColumnLayoutStylePhase> customStylePhases;
 
   /// The `SuperEditor` input source, e.g., keyboard or Input Method Engine.
-  final DocumentInputSource inputSource;
+  final DocumentInputSource? inputSource;
 
   /// The `SuperEditor` gesture mode, e.g., mouse or touch.
   final DocumentGestureMode? gestureMode;
@@ -493,6 +493,26 @@ class SuperEditorState extends State<SuperEditor> {
     }
   }
 
+  /// Returns the [DocumentInputSource] which should be used.
+  ///
+  /// If the `inputSource` is configured, it is used. Otherwise,
+  /// the [DocumentInputSource] is chosen based on the platform.
+  DocumentInputSource get _inputSource {
+    if (widget.inputSource != null) {
+      return widget.inputSource!;
+    }
+    if (kIsWeb) {
+      return DocumentInputSource.keyboard;
+    }
+    switch (defaultTargetPlatform) {
+      case TargetPlatform.android:
+      case TargetPlatform.iOS:
+        return DocumentInputSource.ime;
+      default:
+        return DocumentInputSource.keyboard;
+    }
+  }
+
   @override
   Widget build(BuildContext context) {
     return _buildInputSystem(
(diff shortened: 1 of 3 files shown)


super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -493,6 +493,26 @@ class SuperEditorState extends State<SuperEditor> {
...
 
+  /// Returns the [DocumentInputSource] which should be used.
+  ///
+  /// If the `inputSource` is configured, it is used. Otherwise,
+  /// the [DocumentInputSource] is chosen based on the platform.
+  DocumentInputSource get _inputSource {
+    if (widget.inputSource != null) {
+      return widget.inputSource!;
+    }
+    if (kIsWeb) {
```
matthew-carroll, line 504: Is this the correct condition for mobile web?

super_editor/test/super_editor/supereditor_keyboard_test.dart
```diff
@@ -313,6 +313,62 @@ void main() {
       expect(node.text.text, 'list item 1');
     });
   });
+
+  group('SuperEditor inputSource', () {
+    testWidgetsOnMobile('configures for IME input', (tester) async {
```
matthew-carroll, line 318: These test names should mention something about defaults because we're not testing the ability for the developer to configure the input mode.

super_editor/test/super_editor/supereditor_keyboard_test.dart
```diff
@@ -313,6 +313,62 @@ void main() {
...
+
+      final document = SuperEditorInspector.findDocument()!;
+
+      // Ensure the document was created with one node.
+      expect(document.nodes.length, 1);
+
+      // Tap to give focus to the editor.
+      await tester.placeCaretInParagraph(document.nodes.first.id, 0);
+
+      // Simulate a tap in the newline button, which is handled
```
matthew-carroll, line 355: I'm confused about what you're doing here. How is this confirming that we're in keyboard mode?

super_editor/test/super_editor/supereditor_keyboard_test.dart
angelosilvestre: Updated.

super_editor/test/super_editor/supereditor_keyboard_test.dart
angelosilvestre: First I check that the IME action wasn't handled and then I check if typing with keyboard (using key events) changes the text.

super_editor/test/super_editor/supereditor_keyboard_test.dart
```diff
@@ -313,6 +313,62 @@ void main() {
       expect(node.text.text, 'list item 1');
     });
   });
+
+  group('SuperEditor inputSource', () {
+    testWidgetsOnMobile('configures for IME input by default', (tester) async {
```
matthew-carroll, line 318: I think you still need the platform types in these test names. It doesn't "configure for IME by default", it "configures for IME input by default on mobile"

super_editor/test/super_editor/supereditor_keyboard_test.dart
```diff
@@ -313,6 +313,62 @@ void main() {
...
+
+      final document = SuperEditorInspector.findDocument()!;
+
+      // Ensure the document was created with one node.
+      expect(document.nodes.length, 1);
+
+      // Tap to give focus to the editor.
+      await tester.placeCaretInParagraph(document.nodes.first.id, 0);
+
+      // Simulate a tap in the newline button, which will be handled
```
matthew-carroll, line 332: Rather than just describe what you're doing, this is a good example of a place to explain why you're doing it. It looks like pressing "newline" is a proxy condition for something that can only happen through the IME, but not through a physical keyboard. This isn't something that's likely to be obvious for developers who aren't working on these behaviors on a regular basis. Can you add a concise explanation about why this particular action is relevant to this test?

super_editor/test/super_editor/supereditor_keyboard_test.dart
matthew-carroll: Similar to my comment above, I think there's probably a more direct explanation for this behavior that captures what you're doing, and why you're doing it.

super_editor/test/super_editor/supereditor_keyboard_test.dart
angelosilvestre: I updated the comment, but I'm not sure it's good enough.

matthew-carroll: I think we can say this more concisely:

```
Ensure that IME input is enabled. To check IME input, we arbitrarily simulate a newline action from
the IME. If the editor responds to the newline, it means IME input is enabled. We expect the newline
to insert a new paragraph node.
```

matthew-carroll (approved):
LGTM

PR #820, opened 2022-10-11
Fixed desktop scroll momentum that was broken by Flutter 3.3.3 (Resolves #806)

### LINKED ISSUE #806: [SuperEditor] - Scroll momentum is broken
Flutter 3.3.x broke scrolling momentum in `SuperEditor`, where we use a `Listener` widget.

I filed https://github.com/flutter/flutter/issues/112880

matthew-carroll:
Flutter 3.3.3 broke desktop scrolling. The listener widget no longer includes scrolling momentum in the pointer signal. As a result, we now need to implement momentum, and cancel the momentum at the appropriate time.

This PR was guided by comments in https://github.com/flutter/flutter/issues/112880

### DIFF
--- a/super_editor/lib/src/default_editor/document_gestures_mouse.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_mouse.dart
@@ -408,6 +408,7 @@ class _DocumentMouseInteractorState extends State<DocumentMouseInteractor>
     if (_panGestureDevice == PointerDeviceKind.trackpad) {
       // The user ended a pan gesture with two fingers on a trackpad.
       // We already scrolled the document.
+      widget.autoScroller.goBallistic(-details.velocity.pixelsPerSecond.dy);
       return;
     }
     _onDragEnd();
@@ -442,6 +443,13 @@ class _DocumentMouseInteractorState extends State<DocumentMouseInteractor>
     if (event is PointerScrollEvent) {
       _scrollVertically(event.scrollDelta.dy);
     }
+  }
+
+  /// Beginning with Flutter 3.3.3, we are responsible for starting and
+  /// stopping scroll momentum. This method cancels any scroll momentum
+  /// in our scroll controller.
+  void _cancelScrollMomentum() {
+    widget.autoScroller.goIdle();
   }
 
   void _updateDragSelection() {
@@ -577,6 +585,8 @@ Updating drag selection:
   Widget build(BuildContext context) {
     return Listener(
       onPointerSignal: _scrollOnMouseWheel,
+      onPointerHover: (event) => _cancelScrollMomentum(),
+      onPointerDown: (event) => _cancelScrollMomentum(),
       child: _buildCursorStyle(
         child: _buildGestureInput(
           child: _buildDocumentContainer(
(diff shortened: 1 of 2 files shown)


super_editor/lib/src/default_editor/document_gestures_mouse.dart
```diff
@@ -577,6 +585,8 @@ Updating drag selection:
   Widget build(BuildContext context) {
     return Listener(
       onPointerSignal: _scrollOnMouseWheel,
+      onPointerHover: (event) => _cancelScrollMomentum(),
+      onPointerDown: (event) => _cancelScrollMomentum(),
```
tannermeade, line 589: On a macOS touchpad, this does not stop scrolling if the touchpad is touched. `onPointerDown()` will allow the scrolling to stop when clicking.
Adding `onPointerPanZoomStart: (event) => _cancelScrollMomentum(),` to the Listener seems to work well for me.

tannermeade: `onPointerHover` causes an m1 MacBook air (using a touchpad) to stop scrolling whenever the mouse is moved.

(Speaking only about the touchpad experience on macOS) replacing `onPointerHover` with `onPointerPanZoomStart` makes the functionality relatively accurate from my testing so far.

tannermeade: What is not covered by the previous suggestion is when I put 1 finger down, the expected scrolling behavior is it should stop scrolling which isn't covered by any solution here yet.

angelosilvestre (approved):
LGTM

super_editor/lib/src/default_editor/document_gestures_mouse.dart
matthew-carroll: @tannermeade I just pushed the zoom callback.

Do you know of any place that we can get the signal for the one finger down event? None of the `Listener` methods seem to be relevant. I suppose a `GestureDetector` might work, but a `GestureDetector` would absorb the gestures and break the rest of the document editing experience...

tannermeade: @matthew-carroll from my initial testing I haven't found one so far that triggers an event when I touch it. I'll try to take a closer look later.

matthew-carroll: Update: `onTapDown` doesn't trigger in a `GestureDetector`, so I'm not sure where we're supposed to get a signal for a single pointer down on trackpad...

tannermeade: I haven't been able to get anything triggering from extending directly from GestureDetector and touching the touchpad with a single finger. So I'm inclined to believe the problem might be pretty deep.

tannermeade: So far I don't have a verified demonstration that Flutter has access to that event. The regular scroll views don't respond to this event either.

matthew-carroll: Callum commented in the Flutter thread that we won't get access to this signal until Flutter 3.4. So we'll have to move forward with what we have here, until then.

PR #822, opened 2022-10-12
Export infrastructure/platform_detector.dart

miguelcmedeiros:
`Platform` class from `src/infrastructure/platform_detector.dart` needs to be exported so that the test client code can set the platform in widget tests.

This came up when client code relies on `PrimaryShortcutKey.isPrimaryShortcutKeyPressed` extension method.

### DIFF
--- a/super_editor/lib/super_editor.dart
+++ b/super_editor/lib/super_editor.dart
@@ -51,6 +51,7 @@ export 'src/infrastructure/focus.dart';
 export 'src/infrastructure/ime_input_owner.dart';
 export 'src/infrastructure/keyboard.dart';
 export 'src/infrastructure/multi_tap_gesture.dart';
+export 'src/infrastructure/platform_detector.dart';
 export 'src/infrastructure/platforms/android/android_document_controls.dart';
 export 'src/infrastructure/platforms/ios/ios_document_controls.dart';
 export 'src/infrastructure/platforms/mobile_documents.dart';


matthew-carroll:
Notes: the exposure of `Platform` is needed for Superlist tests because the Superlist code relies on this existing extension:

```dart
extension PrimaryShortcutKey on RawKeyEvent {
      (Platform.instance.isMac && isMetaPressed) || (!Platform.instance.isMac && isControlPressed);
}
```

At a minimum, we should probably rename `Platform` before making it public, so that it doesn't conflict with `dart:io`.

We should also see if `Platform` is still needed. Maybe `defaultTargetPlatform` is sufficient.

If we continue with this PR, we should move the export of `Platform` to `super_editor_test.dart`.

matthew-carroll:
@angelosilvestre can you investigate our use of our definition of `Platform` and see if we still need it? I'm thinking that maybe `defaultTargetPlatform` is sufficient. Maybe we never really needed `Platform`, or maybe Flutter has changed since then. Can you let me know how it looks to you?

angelosilvestre:
@matthew-carroll We only check our `Platform` instance on the keyboard actions code. `defaultTargetPlatform` should be sufficient.

matthew-carroll:
@angelosilvestre can you try removing `Platform`, running the tests, and also try out any relevant behavior in the example app to make sure? If we really don't need it, then please put up a PR that removes it, and we'll take that one instead of this one.

angelosilvestre:
@matthew-carroll Sure. Will do that.

miguelcmedeiros:
Closing in favour of #834

PR #826, opened 2022-10-12
Added ability to align text with stylesheet rules (Resolves #823)

### LINKED ISSUE #823: [SuperEditor] - Add stylesheet support for text alignment
Text alignment in `SuperEditor` can be controlled with `metadata` on a `TextNode`, but there's no way to control text alignment from a `Stylesheet`.

As a developer, I'd like to adjust text alignment based on the type of block. Example:

```dart
StyleRule(
        const BlockSelector("header1"),
        (doc, docNode) {
          return {
            "padding": const CascadingPadding.only(top: 40),
            "textAlign": "center",
            "textStyle": const TextStyle(
              color: Colors.white,
              fontSize: 48,
              fontWeight: FontWeight.bold,
              height: 1.1,
            ),
          };
        },
      ),
```

This then begs the question, what happens when `metadata` conflicts with a `Stylesheet`? My initial thought is that a `Stylesheet` is applied on top of nodes, and therefore the `Stylesheet` should take priority, similar to the way that later styles in a stylesheet take priority over earlier styles.

matthew-carroll:
Added ability to align text with stylesheet rules (Resolves #823)

### DIFF
--- a/super_editor/lib/src/default_editor/text.dart
+++ b/super_editor/lib/src/default_editor/text.dart
@@ -313,6 +313,8 @@ mixin TextComponentViewModel on SingleColumnLayoutComponentViewModel {
   void applyStyles(Map<String, dynamic> styles) {
     super.applyStyles(styles);
 
+    textAlignment = styles["textAlign"] ?? textAlignment;
+
     textStyleBuilder = (attributions) {
       final baseStyle = styles["textStyle"] ?? noStyleBuilder({});
       final inlineTextStyler = styles["inlineTextStyler"] as AttributionStyleAdjuster;
--- a/super_editor/lib/src/default_editor/layout_single_column/_presenter.dart
+++ b/super_editor/lib/src/default_editor/layout_single_column/_presenter.dart
@@ -163,24 +163,24 @@ class SingleColumnLayoutPresenter {
     if (newViewModel == null) {
       // The document changed. All view models were invalidated. Create a
       // new base document view model.
-      final components = <SingleColumnLayoutComponentViewModel>[];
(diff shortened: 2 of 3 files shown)


super_editor/lib/src/default_editor/text.dart
```diff
@@ -313,6 +313,8 @@ mixin TextComponentViewModel on SingleColumnLayoutComponentViewModel {
   void applyStyles(Map<String, dynamic> styles) {
     super.applyStyles(styles);
 
+    textAlignment = styles["textAlign"] ?? textAlignment;
```
angelosilvestre, line 316: Just one question: Are the viewmodels always recreated? Because, if we are overriding the `textAlignment` value and, at some point, another rule without `textAlign` is applied, we would lose the original alignment, right?

For example, if we have a rule with the selector `const BlockSelector("paragraph").after("header1")`, which has `textAlign` set and another one with the selector `const BlockSelector("paragraph")`, which doesn't have `textAlign`, if we delete the header, the paragraph will use the second rule, but will maintain the `textAlign` of the first rule.

Or am I missing something?

matthew-carroll: Yeah, the view models are replicated at every stage of the presentation pipeline. This code should be operating on a copy of the upstream view model, which itself is a copy of further upstream view model. That's what allows us to only re-run part of the presentation pipeline - we can invalidate a style phase in the middle of the pipeline.

angelosilvestre: Ok. Thanks for clarifying.

angelosilvestre (approved):
LGTM

PR #830, opened 2022-10-15
Added support for custom Markdown block syntaxes (Resolves #829)

### LINKED ISSUE #829: [super_editor_markdown] - Allow custom serialization and deserialization
`super_editor_markdown` supports parsing and serializing standard Markdown syntax. This is sufficient for typical prose document content. However, the standard Markdown syntax is not sufficient for all nodes in a document.

Consider a blog post. We'd like for `SuperReader` to render the entire post, from top to bottom, so that users can select all content continuously. That blog post might display author information, advertisements, videos, and other content, which is not part of the primary document content. We'd like a way for developers to be able to serialize and deserialize such non-standard content.

Add the ability for developers to serialize and deserialize custom Markdown block syntax in `super_editor_markdown`.

matthew-carroll:
Added support for custom Markdown block syntaxes (Resolves #829)

I also refactored the source code to accomplish a couple things:

- Separated serialization and deserialization because they're mostly independent, and it's easier to read them when separated
- Converted serialization to use a list of serializer objects, instead of hard-coding all the supported nodes in the global serializer

### DIFF
--- a/super_editor_markdown/lib/src/document_to_markdown_serializer.dart
+++ b/super_editor_markdown/lib/src/document_to_markdown_serializer.dart
@@ -2,11 +2,18 @@ import 'package:super_editor/super_editor.dart';
 
 import 'super_editor_syntax.dart';
 
-/// The given [syntax] controls how the [doc] is serialized, e.g., [MarkdownSyntax.normal] for standard
-/// Markdown syntax, or [MarkdownSyntax.superEditor] to use Super Editor's extended syntax.
+/// Serializes the given [doc] to Markdown text.
+///
+/// The given [syntax] controls how the [doc] is serialized, e.g., [MarkdownSyntax.normal]
+/// for standard Markdown syntax, or [MarkdownSyntax.superEditor] to use Super Editor's
+/// extended syntax.
+///
+/// To serialize [DocumentNode]s that aren't part of Super Editor's standard serialization,
+/// provide [customNodeSerializers] to serialize those custom nodes.
 String serializeDocumentToMarkdown(
   Document doc, {
   MarkdownSyntax syntax = MarkdownSyntax.superEditor,
+  List<DocumentNodeMarkdownSerializer> customNodeSerializers = const [],
 }) {
   StringBuffer buffer = StringBuffer();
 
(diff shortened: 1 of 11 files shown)


super_editor_markdown/lib/src/document_to_markdown_serializer.dart
```diff
@@ -0,0 +1,350 @@
...
+      final markdownStyles = _sortAndSerializeAttributions(endingAttributions, AttributionVisitEvent.end);
+      // Links are different from the plain styles since they are both not NamedAttributions (and therefore
+      // can't be checked using equality comparison) and asymmetrical in markdown.
+      final linkMarker = _encodeLinkMarker(endingAttributions, AttributionVisitEvent.end);
+
+      // +1 on end index because this visitor has inclusive indices
+      // whereas substring() expects an exclusive ending index.
+      _buffer
+        ..write(markdownStyles)
+        ..write(linkMarker);
```
angelosilvestre, line 261: I think this comment is out of place.

super_editor_markdown/lib/src/markdown_to_document_parsing.dart
```diff
@@ -497,154 +418,27 @@ class _InlineMarkdownToDocument implements md.NodeVisitor {
...
-  }
+/// Converts a deserialized Markdown element into a [DocumentNode].
+///
+/// For example, the Markdown parser might identify an element called
+/// "blockquote". A corresponding [ElementToNodeConverter] would receive
+/// the "blockquote" element and create an appropriate [ParagraphNode] to
+/// represent that blockquote in the deserialized [Document].
+abstract class ElementToNodeConverter {
+  DocumentNode? handleElement(md.Element element);
 }
```
angelosilvestre, line 429: Wouldn't it be good to have a base converter that accepts only elements with a configured tag name? Like NodeTypeDocumentNodeMarkdownSerializer does for serialization.

angelosilvestre (approved):
LGTM with a few comments.

super_editor_markdown/lib/src/markdown_to_document_parsing.dart
matthew-carroll: What's the situation that you're worried about with the current API? Where do you think it falls short, or creates additional work?

angelosilvestre: The current API is fine. We could have another class that implements `ElementToNodeConverter` with a tagname property. That way, it could automatically return null if the tagName is different from the tagName the converter cares about.

matthew-carroll: It might be worth doing that when/if we convert all the existing tag behaviors to implement this converter. For now, there are only two of them, and they only exist for testing. Probably better to wait and see what kind of requirements appear for the majority of typical conversions.

Issue #748, opened 2022-08-17
[super_editor] - Make commands atomic


matthew-carroll:
While working on #736 it became apparent that notifying listeners separately for document changes and selection changes create problems when taking post processing actions.

For example, we're trying to tokenize content in paragraphs. This tokenization process depends on both the content in the paragraph, and the position of the caret. Currently, two different change notifications are fired for a single edit. The document change is fired, and the selection change is fired.

We need change notifications to fire one time per logical edit. Also, we generally need to clean up the command system because we have edits that are firing multiple commands for a single operation, and some of those handlers are inspecting information in between the commands. This isn't how commands are supposed to work.

A possible approach:

- Calculate all necessary information in a handler before running any commands
- Batch all relevant commands into a `BatchCommand`, and then execute it
- Create a unified listener system and invoke the unified notification after the command or batch command is done

abidex4yemi:
@matthew-carroll will this be merged anytime soon? I'm still blocked on this #808

matthew-carroll:
We don't have a timeline, but it's in active development

raulmateolabhouse:
Any news on this issue? Not to put pressure, just informative intentions

matthew-carroll:
@raulmateolabhouse this is under active development. We have a client that's adapting their codebase to verify our work, and I'm adjusting the API.

matthew-carroll:
We've merged an initial API for atomic commands via the `Editor` class. We'll continue to work through items in this meta ticket.

Issue #51, opened 2021-03-16
Try DefaultDocumentLayout without caching GlobalKeys


matthew-carroll:
`DefaultDocumentLayout` cached the `GlobalKey`s that are assigned to `DocumentComponent`s. I can't remember why I did this. It's possible that the original issue is gone...or not.

Once the package is capable of rendering complex documents that exceed the visible area of the screen, see what happens if we don't cache the keys. If it still works, then get rid of the caching. If it breaks something, add comments explaining why we need to cache the keys.

matthew-carroll:
I'm going to close this. It seems reasonable that we would cache `GlobalKey`s in the layout because we want the document layout to retain as many component `Element`s as possible, to avoid unnecessary tree restructuring during the build phase.

PR #839, opened 2022-10-27
[SuperEditor] Fix typing lag in large documents (Resolves #751)

### LINKED ISSUE #751: Typing lag
When a few hundred nodes (i noticed it with 150 nodes) are created it seems the whole view gets updated and not only the widget tree of the single node making the typing impossibile.

angelosilvestre:
[SuperEditor] Fix typing lag in large documents. Resolves #751

There is a lag when typing in a document with hundreds of nodes.

This PR is based on the investigation of @jmatth  and @Jei-sKappa.

Some analysis after testing with the Frankenstein document posted by @jmatth  (added as a gist [here](https://gist.github.com/angelosilvestre/cbec8875a67f1b916fcbe09dcbfb06d7)):

* `getNodeIndex` takes some hundred microseconds to complete. It basically walks the entire list of nodes doing a deep comparison. Typing a single letter causes `getNodeIndex` to be called more than 3000 times. This function is called by  `getNodeBefore` , `getNodeAfter` and others.
* `getNodeById` takes some microseconds, but far less than `getNodeIndex`.
* `ParagraphComponentViewModel.==` is broken. Checking `textStyleBuilder == other.textStyleBuilder` always evaluates to `false`. Typing a single letter in a paragraph causes all the components to be considered changed.

This PR changes `MutableDocument`, adding a `Map` to quickly lookup a node index by its id. This `Map` is eagerly updated when the document is created and in each operation which affects the node indices. A `Map` to lookup a node by its id was also added. Because of this `Map's`, I had to change `nodes` to return as an unmodifiable List, to avoid the node list being out of sync with the node maps.

This PR also removes `textStyleBuilder` from the `==` operators of the view models. However, we still need to figure out how to handle the case in which `textStyleBuilder` do change.

After this, I experienced another issue: toggling an attribution wasn't causing the style to be updated immediately. This happens because toggling an attribution just changes `AttributedText` internally. When the viewmodel checks if the text changed, it is the same instance, so it's considered as not changed. This bug was hidden because comparing `textStyleBuilder` always causes the viewmodel to be considered as changed.

This PR changes the attribution commands to create new `AttributedText's`.

After these changes:

https://user-images.githubusercontent.com/7597082/198163767-1941f893-53e5-4da9-a4f5-5059e739e532.mp4

Dragging to select has still some lag:

https://user-images.githubusercontent.com/7597082/198163946-7567f6c3-052a-4564-ba2f-da854be108f0.mp4

Most of the time seems to be spent on `getDocumentSelectionInRegion`, more specifically in a call to `localToGlobal` at:

https://github.com/superlistapp/super_editor/blob/d5e7460956fc6d60abd65991d48ca2c073c1da38/super_editor/lib/src/default_editor/layout_single_column/_layout.dart#L418-L429

When testing with this document I found another bug that is present in the latest main. Typing `1. ` to create a list item it throws this exception: `type 'ListItemNode' is not a subtype of type 'ParagraphNode' of 'element'`. This exception doesn't happen with the default example document.

### DIFF
--- a/super_editor/lib/src/core/document_editor.dart
+++ b/super_editor/lib/src/core/document_editor.dart
@@ -266,8 +266,8 @@ class MutableDocument with ChangeNotifier implements Document {
     }
   }
 
-  /// Inserts [node] at the end of the document.
-  void insertNode(DocumentNode node) {
+  /// Adds [node] to the end of the document.
+  void add(DocumentNode node) {
     _nodes.insert(_nodes.length, node);
     node.addListener(_forwardNodeChange);
 
--- a/super_editor/lib/src/core/document_editor.dart
+++ b/super_editor/lib/src/core/document_editor.dart
@@ -228,7 +228,7 @@ class MutableDocument with ChangeNotifier implements Document {
       node.addListener(_forwardNodeChange);
 
       // The node list changed, we need to update the map to consider the new indices.
-      _updateNodeIdMappings();
+      _refreshNodeIdCaches();
 
       notifyListeners();
     }
@@ -244,7 +244,7 @@ class MutableDocument with ChangeNotifier implements Document {
     newNode.addListener(_forwardNodeChange);
 
     // The node list changed, we need to update the map to consider the new indices.
-    _updateNodeIdMappings();
+    _refreshNodeIdCaches();
 
     notifyListeners();
   }
@@ -260,10 +260,21 @@ class MutableDocument with ChangeNotifier implements Document {
       newNode.addListener(_forwardNodeChange);
 
       // The node list changed, we need to update the map to consider the new indices.
-      _updateNodeIdMappings();
+      _refreshNodeIdCaches();
 
       notifyListeners();
     }
+  }
+
+  /// Inserts [node] at the end of the document.
+  void insertNode(DocumentNode node) {
+    _nodes.insert(_nodes.length, node);
+    node.addListener(_forwardNodeChange);
+
+    // The node list changed, we need to update the map to consider the new indices.
+    _refreshNodeIdCaches();
+
+    notifyListeners();
   }
 
   /// Deletes the node at the given [index].
@@ -273,7 +284,7 @@ class MutableDocument with ChangeNotifier implements Document {
       removedNode.removeListener(_forwardNodeChange);
 
       // The node list changed, we need to update the map to consider the new indices.
-      _updateNodeIdMappings();
+      _refreshNodeIdCaches();
 
       notifyListeners();
     } else {
@@ -290,7 +301,7 @@ class MutableDocument with ChangeNotifier implements Document {
     isRemoved = _nodes.remove(node);
 
     // The node list changed, we need to update the map to consider the new indices.
-    _updateNodeIdMappings();
+    _refreshNodeIdCaches();
 
     notifyListeners();
 
@@ -310,8 +321,8 @@ class MutableDocument with ChangeNotifier implements Document {
     if (_nodes.remove(node)) {
       _nodes.insert(targetIndex, node);
 
-      // The node list changed, we need to update the map to consider the new indices.
-      _updateNodeIdMappings();
+      // An existing node's index changed. Update our Node ID -> Node Index cache.
+      _refreshNodeIdCaches();
 
       notifyListeners();
     }
@@ -331,8 +342,8 @@ class MutableDocument with ChangeNotifier implements Document {
       newNode.addListener(_forwardNodeChange);
       _nodes.insert(index, newNode);
 
-      // The node list changed, we need to update the map to consider the new indices.
-      _updateNodeIdMappings();
+      // An existing node's index changed. Update our Node ID -> Node Index cache.
+      _refreshNodeIdCaches();
 
       notifyListeners();
     } else {
@@ -369,7 +380,7 @@ class MutableDocument with ChangeNotifier implements Document {
   /// Updates all the maps which use the node id as the key.
   ///
   /// All the maps are cleared and re-populated.
-  void _updateNodeIdMappings() {
+  void _refreshNodeIdCaches() {
     _nodeIndicesById.clear();
     _nodesById.clear();
     for (int i = 0; i < _nodes.length; i++) {
--- a/super_editor/lib/src/core/document_editor.dart
+++ b/super_editor/lib/src/core/document_editor.dart
@@ -174,7 +174,7 @@ class MutableDocument with ChangeNotifier implements Document {
     }
 
     if (_nodes[index] != node) {
-      // We found a node by id, but the contents aren't equal.
+      // We found a node by id, but it wasn't the node we expected. Therefore, we couldn't find the requested node.
       return -1;
     }
 
--- a/super_editor/lib/src/core/document_editor.dart
+++ b/super_editor/lib/src/core/document_editor.dart
@@ -168,7 +168,17 @@ class MutableDocument with ChangeNotifier implements Document {
 
   @override
   int getNodeIndex(DocumentNode node) {
-    return _nodeIndicesById[node.id] ?? -1;
+    final index = _nodeIndicesById[node.id] ?? -1;
+    if (index < 0) {
+      return -1;
+    }
+
+    if (_nodes[index] != node) {
+      // We found a node by id, but the contents aren't equal.
+      return -1;
+    }
+
+    return index;
   }
 
   @override
--- a/super_editor/lib/src/core/document_editor.dart
+++ b/super_editor/lib/src/core/document_editor.dart
@@ -167,6 +167,7 @@ class MutableDocument with ChangeNotifier implements Document {
   }
 
   @override
+  @Deprecated("Use getNodeIndexById() instead")
   int getNodeIndex(DocumentNode node) {
     final index = _nodeIndicesById[node.id] ?? -1;
     if (index < 0) {
@@ -183,19 +184,18 @@ class MutableDocument with ChangeNotifier implements Document {
 
   @override
   int getNodeIndexById(String nodeId) {
-    final node = getNodeById(nodeId);
-    return node != null ? getNodeIndex(node) : -1;
+    return _nodeIndicesById[nodeId] ?? -1;
   }
 
   @override
   DocumentNode? getNodeBefore(DocumentNode node) {
-    final nodeIndex = getNodeIndex(node);
+    final nodeIndex = getNodeIndexById(node.id);
     return nodeIndex > 0 ? getNodeAt(nodeIndex - 1) : null;
   }
 
   @override
   DocumentNode? getNodeAfter(DocumentNode node) {
-    final nodeIndex = getNodeIndex(node);
+    final nodeIndex = getNodeIndexById(node.id);
     return nodeIndex >= 0 && nodeIndex < _nodes.length - 1 ? getNodeAt(nodeIndex + 1) : null;
   }
 
@@ -217,13 +217,13 @@ class MutableDocument with ChangeNotifier implements Document {
     if (node1 == null) {
       throw Exception('No such position in document: $position1');
     }
-    final index1 = getNodeIndex(node1);
+    final index1 = getNodeIndexById(node1.id);
 
     final node2 = getNode(position2);
     if (node2 == null) {
       throw Exception('No such position in document: $position2');
     }
-    final index2 = getNodeIndex(node2);
+    final index2 = getNodeIndexById(node2.id);
 
     final from = min(index1, index2);
     final to = max(index1, index2);
@@ -249,7 +249,7 @@ class MutableDocument with ChangeNotifier implements Document {
     required DocumentNode existingNode,
     required DocumentNode newNode,
   }) {
-    final nodeIndex = getNodeIndex(existingNode);
+    final nodeIndex = getNodeIndexById(existingNode.id);
     _nodes.insert(nodeIndex, newNode);
     newNode.addListener(_forwardNodeChange);
 
@@ -264,7 +264,7 @@ class MutableDocument with ChangeNotifier implements Document {
     required DocumentNode existingNode,
     required DocumentNode newNode,
   }) {
-    final nodeIndex = getNodeIndex(existingNode);
+    final nodeIndex = getNodeIndexById(existingNode.id);
     if (nodeIndex >= 0 && nodeIndex < _nodes.length) {
       _nodes.insert(nodeIndex + 1, newNode);
       newNode.addListener(_forwardNodeChange);
--- a/super_editor/lib/src/core/document_editor.dart
+++ b/super_editor/lib/src/core/document_editor.dart
@@ -168,9 +168,6 @@ class MutableDocument with ChangeNotifier implements Document {
 
   @override
   int getNodeIndex(DocumentNode node) {
-    // We always update _nodeIndicesById when the node list changes.
-    // If the id isn't present in the map it means the node isn't present
-    // in the document.
     return _nodeIndicesById[node.id] ?? -1;
   }
 
--- a/super_editor/lib/src/core/document_editor.dart
+++ b/super_editor/lib/src/core/document_editor.dart
@@ -1,7 +1,6 @@
 import 'dart:math';
 
 import 'package:collection/collection.dart';
-import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:super_editor/src/core/document_selection.dart';
 import 'package:super_editor/src/infrastructure/_logging.dart';
@@ -133,22 +132,29 @@ class MutableDocument with ChangeNotifier implements Document {
   MutableDocument({
     List<DocumentNode>? nodes,
   }) : _nodes = nodes ?? [] {
-    // Register listeners for all initial nodes.
-    for (final node in _nodes) {
+    // Register listeners for all initial nodes and populates the node maps.
+    for (int i = 0; i < _nodes.length; i++) {
+      final node = _nodes[i];
       node.addListener(_forwardNodeChange);
+      _nodeIndicesById[node.id] = i;
+      _nodesById[node.id] = node;
     }
   }
 
   final List<DocumentNode> _nodes;
 
   @override
-  List<DocumentNode> get nodes => _nodes;
+  List<DocumentNode> get nodes => List.unmodifiable(_nodes);
+
+  /// Maps a node id to its index in the node list.
+  final Map<String, int> _nodeIndicesById = {};
+
+  /// Maps a node id to its node.
+  final Map<String, DocumentNode> _nodesById = {};
 
   @override
   DocumentNode? getNodeById(String nodeId) {
-    return _nodes.firstWhereOrNull(
-      (element) => element.id == nodeId,
-    );
+    return _nodesById[nodeId];
   }
 
   @override
@@ -162,7 +168,10 @@ class MutableDocument with ChangeNotifier implements Document {
 
   @override
   int getNodeIndex(DocumentNode node) {
-    return _nodes.indexOf(node);
+    // We always update _nodeIndicesById when the node list changes.
+    // If the id isn't present in the map it means the node isn't present
+    // in the document.
+    return _nodeIndicesById[node.id] ?? -1;
   }
 
   @override
@@ -180,12 +189,11 @@ class MutableDocument with ChangeNotifier implements Document {
   @override
   DocumentNode? getNodeAfter(DocumentNode node) {
     final nodeIndex = getNodeIndex(node);
-    return nodeIndex >= 0 && nodeIndex < nodes.length - 1 ? getNodeAt(nodeIndex + 1) : null;
+    return nodeIndex >= 0 && nodeIndex < _nodes.length - 1 ? getNodeAt(nodeIndex + 1) : null;
   }
 
   @override
-  DocumentNode? getNode(DocumentPosition position) =>
-      _nodes.firstWhereOrNull((element) => element.id == position.nodeId);
+  DocumentNode? getNode(DocumentPosition position) => getNodeById(position.nodeId);
 
   @override
   DocumentRange getRangeBetween(DocumentPosition position1, DocumentPosition position2) {
@@ -202,13 +210,13 @@ class MutableDocument with ChangeNotifier implements Document {
     if (node1 == null) {
       throw Exception('No such position in document: $position1');
     }
-    final index1 = _nodes.indexOf(node1);
+    final index1 = getNodeIndex(node1);
 
     final node2 = getNode(position2);
     if (node2 == null) {
       throw Exception('No such position in document: $position2');
     }
-    final index2 = _nodes.indexOf(node2);
+    final index2 = getNodeIndex(node2);
 
     final from = min(index1, index2);
     final to = max(index1, index2);
@@ -218,9 +226,13 @@ class MutableDocument with ChangeNotifier implements Document {
 
   /// Inserts the given [node] into the [Document] at the given [index].
   void insertNodeAt(int index, DocumentNode node) {
-    if (index <= nodes.length) {
-      nodes.insert(index, node);
+    if (index <= _nodes.length) {
+      _nodes.insert(index, node);
       node.addListener(_forwardNodeChange);
+
+      // The node list changed, we need to update the map to consider the new indices.
+      _updateNodeIdMappings();
+
       notifyListeners();
     }
   }
@@ -230,9 +242,13 @@ class MutableDocument with ChangeNotifier implements Document {
     required DocumentNode existingNode,
     required DocumentNode newNode,
   }) {
-    final nodeIndex = nodes.indexOf(existingNode);
-    nodes.insert(nodeIndex, newNode);
+    final nodeIndex = getNodeIndex(existingNode);
+    _nodes.insert(nodeIndex, newNode);
     newNode.addListener(_forwardNodeChange);
+
+    // The node list changed, we need to update the map to consider the new indices.
+    _updateNodeIdMappings();
+
     notifyListeners();
   }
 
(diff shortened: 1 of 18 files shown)


angelosilvestre:
I made an attempt to optimize mouse dragging: 

https://user-images.githubusercontent.com/7597082/198168459-9867ddc2-9690-4add-901e-1023b37ebad7.mp4

But it doesn't help when selecting nodes at the end of the document.

super_editor/lib/src/core/document_editor.dart
```diff
@@ -162,7 +168,10 @@ class MutableDocument with ChangeNotifier implements Document {
 
   @override
   int getNodeIndex(DocumentNode node) {
-    return _nodes.indexOf(node);
+    // We always update _nodeIndicesById when the node list changes.
```
matthew-carroll, line 171: I think we can do without this comment.

super_editor/test/super_editor/supereditor_style_test.dart
```diff
@@ -0,0 +1,40 @@
...
+import 'package:super_editor/src/default_editor/attributions.dart';
+import 'package:super_editor/super_editor_test.dart';
+import 'package:super_text_layout/super_text_layout.dart';
+
+import '../test_tools.dart';
+import 'document_test_tools.dart';
+
+void main() {
+  group('SuperEditor', () {
+    testWidgetsOnArbitraryDesktop('applies style when toggling attributions', (tester) async {
```
matthew-carroll, line 12: "applies style when toggling attributions" -> "changes visual text style when attributions change"

matthew-carroll (commented):
Generally LGTM. As of this point, what remaining problems do we have?

angelosilvestre:
@matthew-carroll Right now, the only thing I think is missing is a way to detect when `textStyleBuilder` actually changes.

BTW. I replaced `List.unmodifiable` with `UnmodifiableListView` because `List.unmodifiable` copies all elements. With this change, the performance improved a little bit.

matthew-carroll:
@Jei-sKappa @jmatth @Mijawel - Can each of you give this PR a try and verify that it resolves your performance problems?

Mijawel:
Looks good. No obvious lag at 500 nodes/50,000 words. At 1000 nodes/100,000 words there is a decent amount of lag but it is still usable. 

Much better than before. Was getting lag at 50 to 100 nodes.

Jei-sKappa:
For me the performance is good, btw I also notice what @Mijawel said, but it's totally fine unless you type at 200+ wpm.

Unfortunately while testing the editor I noticed a bug when pasting text, I want to make clear that it could be totally unrelated but I wanted to point it out.
Steps to reproduce:
- Run the example app
- Add "some" empty nodes (see the video, it occurred not only with those lines)
- Select all text (in my case with CMD+A)
- Copy it
- Paste it at the bottom



https://user-images.githubusercontent.com/46776108/198833074-3e34dee3-3bfa-49cc-8874-094434eaab72.mov

The error output is:
The following StateError was thrown building TextComponent-[GlobalKey#2a2a8](dirty, state:
TextComponentState#86374):
Bad state: No element

  TextComponent-[GlobalKey#2a2a8]
  TextComponent:file:///Users/jac/Downloads/super_editor-751_typing_lag/super_editor/lib/src/default_editor/paragraph.dar
  t:91:12

#0      List.last (dart:core-patch/growable_array.dart:348:5)
#1      AttributedSpans.collapseSpans (package:attributed_text/src/attributed_spans.dart:913:24)
#2      ComputeTextSpan.computeTextSpan (package:super_editor/src/infrastructure/attributed_text_styles.dart:32:34)
#3      TextComponentState.build (package:super_editor/src/default_editor/text.dart:783:31)
#4      StatefulElement.build (package:flutter/src/widgets/framework.dart:4992:27)
#5      ComponentElement.performRebuild (package:flutter/src/widgets/framework.dart:4878:15)
#6      StatefulElement.performRebuild (package:flutter/src/widgets/framework.dart:5050:11)
#7      Element.rebuild (package:flutter/src/widgets/framework.dart:4604:5)
#8      ComponentElement._firstBuild (package:flutter/src/widgets/framework.dart:4859:5)
#9      StatefulElement._firstBuild (package:flutter/src/widgets/framework.dart:5041:11)
#10     ComponentElement.mount (package:flutter/src/widgets/framework.dart:4853:5)
...     Normal element mounting (24 frames)
#34     Element.inflateWidget (package:flutter/src/widgets/framework.dart:3863:16)
#35     MultiChildRenderObjectElement.inflateWidget (package:flutter/src/widgets/framework.dart:6435:36)
#36     Element.updateChild (package:flutter/src/widgets/framework.dart:3592:18)
#37     RenderObjectElement.updateChildren (package:flutter/src/widgets/framework.dart:5964:32)
#38     MultiChildRenderObjectElement.update (package:flutter/src/widgets/framework.dart:6460:17)
#39     Element.updateChild (package:flutter/src/widgets/framework.dart:3570:15)
#40     SingleChildRenderObjectElement.update (package:flutter/src/widgets/framework.dart:6307:14)
#41     Element.updateChild (package:flutter/src/widgets/framework.dart:3570:15)
#42     ComponentElement.performRebuild (package:flutter/src/widgets/framework.dart:4904:16)
#43     StatefulElement.performRebuild (package:flutter/src/widgets/framework.dart:5050:11)
#44     Element.rebuild (package:flutter/src/widgets/framework.dart:4604:5)
#45     BuildOwner.buildScope (package:flutter/src/widgets/framework.dart:2667:19)
#46     WidgetsBinding.drawFrame (package:flutter/src/widgets/binding.dart:882:21)
#47     RendererBinding._handlePersistentFrameCallback (package:flutter/src/rendering/binding.dart:378:5)
#48     SchedulerBinding._invokeFrameCallback (package:flutter/src/scheduler/binding.dart:1175:15)
#49     SchedulerBinding.handleDrawFrame (package:flutter/src/scheduler/binding.dart:1104:9)
#50     SchedulerBinding._handleDrawFrame (package:flutter/src/scheduler/binding.dart:1015:5)
#51     _invoke (dart:ui/hooks.dart:148:13)
#52     PlatformDispatcher._drawFrame (dart:ui/platform_dispatcher.dart:318:5)
#53     _drawFrame (dart:ui/hooks.dart:115:31)

matthew-carroll:
@Jei-sKappa can you file a new issue for that copy bug?

matthew-carroll:
BTW @Mijawel and @Jei-sKappa - would you two mind telling us a bit about how you're using Super Editor? If you'd like to send that info privately, you're welcome to come over to the Super Editor Slack channel. I'm curious, because the more we know about community use-cases, the better we can focus our roadmap.

Jei-sKappa:
> @Jei-sKappa can you file a new issue for that copy bug?

Issue just created!

> BTW @Mijawel and @Jei-sKappa - would you two mind telling us a bit about how you're using Super Editor? If you'd like to send that info privately, you're welcome to come over to the Super Editor Slack channel. I'm curious, because the more we know about community use-cases, the better we can focus our roadmap.

Of course, I’d be happy to share how I use the editor.
I'll reach you in Slack.

super_editor/lib/src/core/document_editor.dart
```diff
@@ -282,8 +307,12 @@ class MutableDocument with ChangeNotifier implements Document {
       throw Exception('Could not find node with nodeId: $nodeId');
     }
 
-    if (nodes.remove(node)) {
-      nodes.insert(targetIndex, node);
+    if (_nodes.remove(node)) {
+      _nodes.insert(targetIndex, node);
+
+      // The node list changed, we need to update the map to consider the new indices.
```
matthew-carroll, line 313: Let's say "// An existing node's index changed. Update our Node ID -> Node Index cache.

super_editor/lib/src/core/document_editor.dart
```diff
@@ -302,6 +331,9 @@ class MutableDocument with ChangeNotifier implements Document {
       newNode.addListener(_forwardNodeChange);
       _nodes.insert(index, newNode);
 
+      // The node list changed, we need to update the map to consider the new indices.
```
matthew-carroll, line 334: Let's update this comment, too

super_editor/lib/src/core/document_editor.dart
```diff
@@ -320,19 +352,33 @@ class MutableDocument with ChangeNotifier implements Document {
...
       }
     }
 
     return true;
   }
 
+  /// Updates all the maps which use the node id as the key.
+  ///
+  /// All the maps are cleared and re-populated.
+  void _updateNodeIdMappings() {
```
matthew-carroll, line 372: Let's call this `_refreshNodeIdCaches()`

super_editor/lib/src/default_editor/layout_single_column/_layout.dart
```diff
@@ -320,7 +317,11 @@ class _SingleColumnDocumentLayoutState extends State<SingleColumnDocumentLayout>
     dynamic bottomNodeBasePosition;
     dynamic bottomNodeExtentPosition;
 
-    for (final componentKey in _topToBottomComponentKeys) {
+    final startOffset = min(baseOffset.dy, extentOffset.dy);
```
matthew-carroll, line 320: Let's call this `selectionRegionTopOffset`

super_editor/lib/src/default_editor/layout_single_column/_layout.dart
```diff
@@ -320,7 +317,11 @@ class _SingleColumnDocumentLayoutState extends State<SingleColumnDocumentLayout>
...
+    final startIndex = max(_findComponentIndexAtOffset(startOffset), 0);
```
matthew-carroll, line 321: Let's call this `componentSearchStartIndex`

super_editor/lib/src/default_editor/layout_single_column/_layout.dart
```diff
@@ -320,7 +317,11 @@ class _SingleColumnDocumentLayoutState extends State<SingleColumnDocumentLayout>
...
+
+    for (int i = startIndex; i < _topToBottomComponentKeys.length; i++) {
```
matthew-carroll, line 323: Let's add a comment before this for loop:

```
// Find the top and bottom nodes in the selection region. We do this by finding the component
// at the top of the selection, then we iterate down the document until we find the bottom
// component in the selection region. We obtain the document nodes from the components.
```

Also, given that the 2 variables above this are only used for the loop condition, let's move them directly above the for loop, and let's place this comment above those two lines.

super_editor/lib/src/default_editor/layout_single_column/_layout.dart
```diff
@@ -705,6 +719,60 @@ class _SingleColumnDocumentLayoutState extends State<SingleColumnDocumentLayout>
...
+  /// Performs a binary search starting from [minIndex] to [maxIndex] to find
+  /// a component whose bounds contains the offset [dy].
+  ///
+  /// Returns the index of the component, from top to bottom order.
+  int _binarySearchComponentIndexAtOffset(double dy, int minIndex, int maxIndex) {
+    if (minIndex > maxIndex) {
+      return -1;
+    }
+
+    final midleIndex = ((minIndex + maxIndex) / 2).floor();
```
matthew-carroll, line 739: "midleIndex" -> "middleIndex"

super_editor/lib/src/default_editor/layout_single_column/_layout.dart
```diff
@@ -705,6 +719,60 @@ class _SingleColumnDocumentLayoutState extends State<SingleColumnDocumentLayout>
...
+  /// Returns the index of the component, from top to bottom order.
+  int _findComponentIndexAtOffset(double dy) {
+    return _binarySearchComponentIndexAtOffset(dy, 0, _topToBottomComponentKeys.length);
+  }
+
+  /// Performs a binary search starting from [minIndex] to [maxIndex] to find
+  /// a component whose bounds contains the offset [dy].
+  ///
+  /// Returns the index of the component, from top to bottom order.
+  int _binarySearchComponentIndexAtOffset(double dy, int minIndex, int maxIndex) {
```
matthew-carroll, line 734: To confirm, have you tried selecting in a document with 0, 1, even, and odd node counts? Any of those might blow up a binary search that isn't quite right.

super_editor/lib/src/default_editor/layout_single_column/_layout.dart
```diff
@@ -705,6 +719,60 @@ class _SingleColumnDocumentLayoutState extends State<SingleColumnDocumentLayout>
...
+    if (minIndex > maxIndex) {
+      return -1;
+    }
+
+    final midleIndex = ((minIndex + maxIndex) / 2).floor();
+    final componentBounds = _getComponentBoundsByIndex(midleIndex);
+
+    if (dy >= componentBounds.top && dy <= componentBounds.bottom) {
```
matthew-carroll, line 742: FYI - if-statements like this are often easier to read if written like this:

```
componentBounds.top <= dy && dy <= componentBounds.bottom
```

super_editor/lib/src/default_editor/layout_single_column/_layout.dart
```diff
@@ -705,6 +719,60 @@ class _SingleColumnDocumentLayoutState extends State<SingleColumnDocumentLayout>
...
+      return midleIndex;
```
matthew-carroll, line 743: Let's add a comment:

```
// The component in the middle of the search region is the one we're looking for. Return its index.
```

super_editor/lib/src/default_editor/layout_single_column/_layout.dart
```diff
@@ -705,6 +719,60 @@ class _SingleColumnDocumentLayoutState extends State<SingleColumnDocumentLayout>
...
+    }
+
+    if (dy > componentBounds.bottom) {
+      if (midleIndex + 1 < _topToBottomComponentKeys.length) {
```
matthew-carroll, line 747: Let's add a comment:

```
// The component we're looking for is somewhere in the bottom half of the current search region
```

super_editor/lib/src/default_editor/layout_single_column/_layout.dart
```diff
@@ -705,6 +719,60 @@ class _SingleColumnDocumentLayoutState extends State<SingleColumnDocumentLayout>
...
+        // Check the gap between two components.
+        final nextComponentBounds = _getComponentBoundsByIndex(midleIndex + 1);
+        if (dy > componentBounds.bottom && dy < nextComponentBounds.top) {
+          return midleIndex;
+        }
+      }
+      return _binarySearchComponentIndexAtOffset(dy, midleIndex + 1, maxIndex);
+    } else {
+      if (midleIndex - 1 >= 0) {
```
matthew-carroll, line 756: Let's add a comment:

```
// The component we're looking for is somewhere in the top half of the current search region
```

super_editor/lib/src/default_editor/layout_single_column/_layout.dart
```diff
@@ -705,6 +719,60 @@ class _SingleColumnDocumentLayoutState extends State<SingleColumnDocumentLayout>
...
+
+    if (dy >= componentBounds.top && dy <= componentBounds.bottom) {
+      return midleIndex;
+    }
+
+    if (dy > componentBounds.bottom) {
+      if (midleIndex + 1 < _topToBottomComponentKeys.length) {
+        // Check the gap between two components.
+        final nextComponentBounds = _getComponentBoundsByIndex(midleIndex + 1);
+        if (dy > componentBounds.bottom && dy < nextComponentBounds.top) {
```
matthew-carroll, line 750: Shouldn't we divide these gaps in half? Otherwise, the component that happens to be in the middle of a search receives special treatment with regard to when it's considered selected.

super_editor/lib/src/default_editor/text.dart
```diff
@@ -887,9 +887,19 @@ class AddTextAttributionsCommand implements EditorCommand {
         final node = entry.key;
         final range = entry.value.toSpanRange();
         editorDocLog.info(' - adding attribution: $attribution. Range: $range');
-        node.text.addAttribution(
-          attribution,
-          range,
+
+        // During style phase, we check which components changed, so we only re-style those components.
```
matthew-carroll, line 891: Can you file an issue for this? We should probably have a way to compare content.

Also, please add a TODO to each of these comments that references the issue that you file.

super_editor/test/src/default_editor/text_test.dart
```diff
@@ -107,12 +107,13 @@ void main() {
         final editContext = _createEditContext();
 
         // Add a paragraph to the document.
-        (editContext.editor.document as MutableDocument).nodes.add(
-              ParagraphNode(
-                id: 'paragraph',
-                text: AttributedText(text: 'This is some text'),
-              ),
-            );
+        (editContext.editor.document as MutableDocument).insertNodeAt(
```
matthew-carroll, line 110: It looks like the `add()` calls have been replaced by `insertNodeAt()`, which isn't the same behavior. `insertNodeAt()` requires knowledge of how many nodes are in the document. The `add()` method just says "put this at the end, wherever that may be". Can we recover the original behavior without creating a performance problem?

super_editor/test/super_editor/supereditor_style_test.dart
```diff
@@ -0,0 +1,124 @@
...
+          .pump();
+
+      await tester.placeCaretInParagraph(SuperEditorInspector.findDocument()!.nodes.last.id, 0);
+
+      final presenter = tester.state<SuperEditorState>(find.byType(SuperEditor)).presenter;
+      presenter.addChangeListener(SingleColumnLayoutPresenterChangeListener(
+        onViewModelChange: ({required addedComponents, required changedComponents, required removedComponents}) {
+          // The listener is called two times. The first one for the text change,
+          // and the second one for the selection change.
+          componentChangedCount = max(componentChangedCount, changedComponents.length);
```
matthew-carroll, line 76: It's not obvious to me what this `max()` is trying to accomplish. Can we check the behavior with hard-coded values to make things super clear?

super_editor/lib/src/default_editor/text.dart
angelosilvestre: Maybe this will change in https://github.com/superlistapp/super_editor/issues/748 ? I will reference this issue then.

angelosilvestre: Nevermind, I was confused with another statement.

super_editor/test/src/default_editor/text_test.dart
angelosilvestre: This document created by  `_createEditContext` is an empty one, so, in practice, inserting the node will have the same effect. We can't directly mutate the node list because it will cause the cache to be incorrect, and we currently don't have a method in the document public API to add a node to the end.

super_editor/test/super_editor/supereditor_style_test.dart
angelosilvestre: It's because typing a letter cause 2 calls to `onViewModelChange`, so the selection change overrides the value we care about. I will try to change to make it more obvious.

This might change after https://github.com/superlistapp/super_editor/issues/748 is closed.

super_editor/test/src/default_editor/text_test.dart
matthew-carroll: Then let's go ahead and add that method. I can't think of a good reason to prevent users from doing that.

super_editor/test/super_editor/supereditor_style_test.dart
matthew-carroll: Can you use an if-statement, or some explicit control structure, rather than calculate a max number?

angelosilvestre: Updated.

super_editor/lib/src/default_editor/text.dart
angelosilvestre: We have a way to compare content. The `AttributedText` equality operator is defined as:

```dart
@override
  bool operator ==(Object other) {
    return identical(this, other) ||
        other is AttributedText && runtimeType == other.runtimeType && text == other.text && spans == other.spans;
  }
```

If we just toggle the attributions in the same instance, the first condition will always be true.

super_editor/lib/src/default_editor/layout_single_column/_layout.dart
angelosilvestre: I made a couple of small adjustments.

super_editor/lib/src/default_editor/layout_single_column/_layout.dart
angelosilvestre: Updated

super_editor/lib/src/default_editor/text.dart
matthew-carroll: Can you reframe this as an action item? I'm not sure what to do with the above information...

super_editor/test/src/default_editor/text_test.dart
```diff
@@ -107,12 +107,12 @@ void main() {
         final editContext = _createEditContext();
 
         // Add a paragraph to the document.
-        (editContext.editor.document as MutableDocument).nodes.add(
-              ParagraphNode(
-                id: 'paragraph',
-                text: AttributedText(text: 'This is some text'),
-              ),
-            );
+        (editContext.editor.document as MutableDocument).insertNode(
```
matthew-carroll, line 110: This method should be called `add()` - "insert" implies control over the index, or otherwise placing an item between other items.

super_editor/lib/src/default_editor/text.dart
angelosilvestre: `ParagraphComponentViewModel` compares its `AttributedText` as part of its equality operator (`text == other.text`).

Before the change, we were only modifying the contents of the same `AttributedText` that was referenced in the viewmodel. The contents changed, but as it's referencing the same object, the equality operator returns true, as `identical(this, other)` is `true`.

matthew-carroll: Ok. Let's change the comment to:

```
// Create a new AttributedText with updated attribution spans, so that the presentation system can
// see that we made a change, and re-renders the text in the document.
```

Let's change the implementation to:

```
node.text = AttributedText(
  text: node.text.text,
  spans: node.text.spans.copy()..addAttribution(
    newAttribution: attribution,
    start: range.start,
    end: range.end,
  ),
);
```

jmatth:
This resolves the performance issues for me.

super_editor/lib/src/core/document_editor.dart
```diff
@@ -162,7 +168,7 @@ class MutableDocument with ChangeNotifier implements Document {
 
   @override
   int getNodeIndex(DocumentNode node) {
-    return _nodes.indexOf(node);
+    return _nodeIndicesById[node.id] ?? -1;
```
jmatth, line 171: This is changing the behavior of the method slightly. It will now return an index if you pass a node that shares an `id` with one in the document even if they are otherwise unequal. Additionally, if someone implemented custom nodes with overridden `==` behavior, this method may now return `-1` instead of a valid index.

In my implementation I used the cache to quickly look up the node and then performed the comparison to maintain the existing behavior. [Code here][1].

[1]: https://github.com/jmatth/super_editor/blob/79f1032afbff19d7302899ea4e0906a9e48ba708/super_editor/lib/src/core/document_editor.dart#L171-L177

angelosilvestre: I updated to add the equality check.

super_editor/lib/src/core/document_editor.dart
```diff
@@ -162,7 +168,17 @@ class MutableDocument with ChangeNotifier implements Document {
...
   @override
   int getNodeIndex(DocumentNode node) {
-    return _nodes.indexOf(node);
+    final index = _nodeIndicesById[node.id] ?? -1;
+    if (index < 0) {
+      return -1;
+    }
+
+    if (_nodes[index] != node) {
+      // We found a node by id, but the contents aren't equal.
```
matthew-carroll, line 177: I think I would word this a bit differently: `// We found a node by id, but it wasn't the node we expected. Therefore, we couldn't find the requested node`

angelosilvestre: Updated.

super_editor/lib/src/core/document_editor.dart
matthew-carroll: I thought about this a bit more. Why would there ever be 2 nodes with the same ID and different content? The point of the ID is to uniquely refer to a node within a document...

matthew-carroll: @jmatth can you let me know when/why this change would be a problem?

jmatth: I could trivially construct two nodes with the same ID that are otherwise unequal, store one in a `MutableDocument`, then pass the second one to `getNodeIndex` and get a valid index back. I find that result unexpected. Best example I can think of for why this would happen is converting between node types, such as a paragraph to a list item or similar, but I don't actually have any real world examples. If you don't think that's a reasonable use case to accommodate then this method should just be removed entirely, because `getNodeIndexById` also exists, its signature makes it much clearer what the criteria for a match are, and without the equality check the two methods serve the exact same purpose.

matthew-carroll: @angelosilvestre what do you think about getting rid of this method? Can you take a look at the downstream consequences and see if there's any meaningful reason for us to keep it?

matthew-carroll: That said, if removal looks acceptable, there's probably no reason to make that breaking change now. We're trying to land a semi-stable release after this PR. So @angelosilvestre if you think removal looks OK, then add a "deprecated" annotation to the method and we'll merge it in as-is.

angelosilvestre: I will take a look.

angelosilvestre: @matthew-carroll We are using this method in a number of places, but we could change them to use `getNodeIndexById`. Should I add the deprecated annotation and change the usages?

matthew-carroll: If you've inspected all of those locations, and you're comfortable that changing them won't be a big deal, then sure, you can go ahead and deprecate this method. Please be sure to include a deprecation message that points to the other method.

angelosilvestre: I marked `getNodeIndex` as deprecated and changed the places which use this method.

matthew-carroll (approved):
LGTM

Issue #842, opened 2022-10-31
SuperTextField doesn't work if FocusNode is already focused.


knopp:
We have a use-case where SuperTextField replaces another widget, re-using the focus node. When the focus node is already focused, text-input doesn't work. The reason is controller selection being initialized in `SuperDesktopTextFieldState._onFocusChange`, but that function is not called since the node is already focused when state is initialized.

[Changing the initialisation logic slightly](https://github.com/knopp/super_editor/commit/c7a0072323018759c03b6b815c0ec1c952da52a7) resolves the issue.

knopp:
I added PR for desktop. The fix for mobile seems to be more involved and I'm not very familiar with the codebase.

matthew-carroll:
@angelosilvestre let's have you take this over and implement the mobile adjustments

knopp:
Is there any update on this?

matthew-carroll:
I thought this was fixed. @angelosilvestre did we merge the fix for this, or am I misremembering?

angelosilvestre:
The fix was already merged and cherry-picked to stable.

angelosilvestre:
This was fixed in https://github.com/superlistapp/super_editor/pull/846

PR #843, opened 2022-11-01
SuperTextField: Initialize controller selection if focused during initState

knopp:
Partially fixes https://github.com/superlistapp/super_editor/issues/842 (for desktop platforms)


matthew-carroll:
Can you describe what "partially" means? Where does this fail, and why do you think it's failing?

knopp:
It only fixes the problem with SuperTextField on desktop. Android and iOS versions still need to be fixed. You can see the failure when you enable the test for all devices (currently is only enabled for desktop).

matthew-carroll:
> It only fixes the problem with SuperTextField on desktop

Ok. I'd like to avoid partial solutions. Can you add the fixes for mobile? If you don't have time, would you like @angelosilvestre to take this over?

matthew-carroll:
I'm gonna close this PR in favor of this one: https://github.com/superlistapp/super_editor/pull/846

PR #844, opened 2022-11-01
[SuperEditor] Change TextInputClient implementations to use it as a mixin (Resolves #840)

### LINKED ISSUE #840: [SuperEditor] - switch TextInputClient implementations to use "with" instead of "implements"
We've implemented `TextInputClient` in a number of places. We `implements` the interface, because the interface is a fundamental contract between all platforms, Flutter, and Flutter apps about how text is edited. As such, it's important to implement all responsibilities.

Unfortunately, Flutter has established a pattern of throwing new methods into this interface that don't belong there. As a result, Flutter has broken us multiple times by adding methods that we don't need to implement.

This ticket is to switch our implementations from `implements` to mixins with `with`. This change creates the possibility that we don't know about new methods in the interface until long after they're added. This also means that we won't notice the change in time to push back on the Flutter team for adding the new method. But the Flutter team hasn't taken much of our push back seriously in the past, so it's more important that we stop letting them break us. It's an unfortunate situation all the way around, but it's out of our control.

angelosilvestre:
[SuperEditor] Change `TextInputClient` implementations to use it as a mixin. Resolves #840

`TextInputClient` was changed to be a mixin. That way, we can use it with the `with` keyword to inherit the default implementations and avoid breaking the build when new methods with default implementations are added.

This PR changes the places where we implemented `DeltaTextInputClient` and `TextInputClient` to use them with the `with` keyword.

### DIFF
--- a/super_editor/example/lib/demos/debugging/simple_deltas_input.dart
+++ b/super_editor/example/lib/demos/debugging/simple_deltas_input.dart
@@ -12,7 +12,7 @@ class SimpleDeltasInputDemo extends StatefulWidget {
   _SimpleDeltasInputState createState() => _SimpleDeltasInputState();
 }
 
-class _SimpleDeltasInputState extends State<SimpleDeltasInputDemo> implements DeltaTextInputClient {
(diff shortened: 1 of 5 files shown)


matthew-carroll (approved):
LGTM

PR #846, opened 2022-11-03
[SuperTextField] Fix selection and input when initializing with a focused FocusNode (Resolves #843)

angelosilvestre:
[SuperTextField] Fix selection and input when initializing with a focused FocusNode. Resolves #843

When `SuperTextField` replaces another widget using an already focused `FocusNode` it doesn't show the caret and connect to IME.

This PR adds a post-frame callback to update selection/IME connection.

We already have a test which gives a focused `FocusNode` to the textfield  but, apparently, the issue only happens if the `FocusNode` was previously used.

https://github.com/superlistapp/super_editor/blob/47d2b4a492aca04b69274871a8880394821b18e8/super_editor/test/super_textfield/super_textfield_test.dart#L172-L183

This PR is based on https://github.com/superlistapp/super_editor/pull/843

### DIFF
--- a/super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
+++ b/super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
@@ -120,11 +120,13 @@ class SuperDesktopTextFieldState extends State<SuperDesktopTextField> implements
     super.initState();
 
     _focusNode = (widget.focusNode ?? FocusNode())..addListener(_onFocusChange);
-    _hasFocus = _focusNode.hasFocus;
 
     _controller = (widget.textController ?? AttributedTextEditingController())
       ..addListener(_onSelectionOrContentChange);
     _scrollController = ScrollController();
+
+    // Check if we need to update the selection.
+    _onFocusChange();
   }
 
   @override
@@ -137,7 +139,9 @@ class SuperDesktopTextFieldState extends State<SuperDesktopTextField> implements
         _focusNode.dispose();
       }
       _focusNode = (widget.focusNode ?? FocusNode())..addListener(_onFocusChange);
-      _hasFocus = _focusNode.hasFocus;
+
+      // Check if we need to update the selection.
+      _onFocusChange();
     }
 
     if (widget.textController != oldWidget.textController) {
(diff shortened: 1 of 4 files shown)


super_editor/test/super_textfield/super_textfield_test.dart
```diff
@@ -200,6 +200,34 @@ void main() {
 
         expect(_isCaretPresent(tester), isTrue);
       });
+
+      testWidgetsOnAllPlatforms(
+          "is inserted automatically when the field is initialized with a focused node used by another widget",
```
matthew-carroll, line 205: Does "used by another widget" actually have any relevance here? Is any other widget required to create this situation, or does this situation only require that `SuperTextField` be given a `FocusNode` that already has focus?

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -120,11 +120,13 @@ class SuperDesktopTextFieldState extends State<SuperDesktopTextField> implements
...
 
     _focusNode = (widget.focusNode ?? FocusNode())..addListener(_onFocusChange);
-    _hasFocus = _focusNode.hasFocus;
 
     _controller = (widget.textController ?? AttributedTextEditingController())
       ..addListener(_onSelectionOrContentChange);
     _scrollController = ScrollController();
+
+    // Check if we need to update the selection.
+    _onFocusChange();
```
matthew-carroll, line 129: Ideally, `_onFocusChange()` should only be called when focus changes

super_editor/test/super_textfield/super_textfield_test.dart
angelosilvestre: It seems to be relevant. When removing the first `pumpWidget` the issue doesn't happen. We already have another test that gives `SuperTextField` an already focused `FocusNode` and it didn't catch the bug.

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
angelosilvestre: Renaming it to `_updateSelecion` or `_updateSelecionOnFocusChange` would be better?

matthew-carroll: Let's go with `_updateSelectionOnFocusChange()`. If the focus change behavior accumulates any responsibilities that aren't about selection, the name will need to change again.

angelosilvestre: Updated.

matthew-carroll (approved):
LGTM

PR #849, opened 2022-11-05
[SuperEditor] Add selection reason to selection changes (Resolves #847)

### LINKED ISSUE #847: [SuperEditor] - Add a "reason" for selection changes
A document selection might change because the user directly interacts with the selection, e.g., clicks and drags, or drags a handle. Or, a document selection might change because of background work, e.g., a multiplayer user deletes a character, which causes the primary user's selection to contract by one character.

When the user's selection changes for reasons beyond the user's interaction, we don't want things like auto-scrolling to happen.

For this ticket, introduce the concept of a "reason" for every selection change in the `DocumentComposer`. That "reason" can be a `String`, and super_editor can provide two initial options: `userInteraction` and `contentChange`.

Once the reason concept is added, adjust `SuperEditor` so that only `userInteraction` selection changes cause auto-scrolling.

Further details for the motivation of this ticket can be found here: https://github.com/superlistapp/super_editor/issues/816

angelosilvestre:
[SuperEditor] Add selection reason to selection changes. Resolves #847

This PR introduces the concept of a selection reason, which represents what caused the selection to change.

A class called `DocumentSelectionChange`  was added, which holds the `DocumentSelection` and the reason.

All places that used `ValueNotifier<DocumentSelection?>` now use `ValueNotifier<DocumentSelectionChange>`. To change the selection using this new `ValueNotifier` the following code is needed:

```dart
 selectionChangeNotifier.value = DocumentSelectionChange(
    selection: DocumentSelection(...),
    reason: CommonSelectionChangeReasons.userInteraction, // the reason is optional and defaults to CommonSelectionChangeReasons.userInteraction
 );
```
Two selection reasons are provided: `CommonSelectionChangeReasons.userInteraction` and `CommonSelectionChangeReasons.contentChange`. Users can extend `SelectionChangeReason` to create their own reasons.

We could also pass `DocumentComposer` around instead of the `ValueNotifier`, so user would be able to just use `composer.selection = ...;` for the common use-case.

This PR also changes the editor to avoid auto-scroll when selection changes by reasons other than `CommonSelectionChangeReasons.userInteraction`.

I used a class with static consts so users can discover all the provided reasons. We could also just use normal constants and avoid the `CommonSelectionChangeReasons` class.


super_editor/lib/src/core/document_composer.dart
```diff
@@ -33,21 +33,32 @@ class DocumentComposer with ChangeNotifier {
...
+  /// Sets the current [selectionChange] for a [Document].
+  set selectionChange(DocumentSelectionChange change) {
+    if (change != selectionChange) {
+      selectionChangeNotifier.value = change;
       notifyListeners();
     }
   }
 
-  final selectionNotifier = ValueNotifier<DocumentSelection?>(null);
+  final selectionChangeNotifier = ValueNotifier<DocumentSelectionChange>(const DocumentSelectionChange.empty());
```
matthew-carroll, line 57: I'm thinking that we should change this to a `Stream<DocumentSelectionChange>` and the name of the property should be `selectionChanges`.

super_editor/lib/src/core/document_composer.dart
```diff
@@ -33,21 +33,32 @@ class DocumentComposer with ChangeNotifier {
...
-    if (newSelection != selectionNotifier.value) {
-      selectionNotifier.value = newSelection;
+    if (newSelection != selectionChange.selection) {
+      selectionChangeNotifier.value = DocumentSelectionChange(selection: newSelection);
+      notifyListeners();
+    }
+  }
+
+  /// Sets the current [selectionChange] for a [Document].
+  set selectionChange(DocumentSelectionChange change) {
```
matthew-carroll, line 50: Instead of having two methods, and instead of forcing clients to construct one of these change objects, let's have one method with two parameters:

```dart
void setSelection(DocumentSelection? newSelection, [Object reason = userInteraction]) {...}
```

super_editor/lib/src/core/document_composer.dart
```diff
@@ -33,21 +33,32 @@ class DocumentComposer with ChangeNotifier {
   }
 
   /// Returns the current [DocumentSelection] for a [Document].
-  DocumentSelection? get selection => selectionNotifier.value;
+  DocumentSelection? get selection => selectionChange.selection;
```
matthew-carroll, line 36: After the change to turn the `ValueNotifier` into a `Stream`, you won't be able to use the `Stream` to get the current value. Instead, we should cache the latest selection and reason, privately.

super_editor/lib/src/core/document_composer.dart
```diff
@@ -33,21 +33,32 @@ class DocumentComposer with ChangeNotifier {
...
+
+  /// Returns the current [DocumentSelectionChange] for a [Document].
+  DocumentSelectionChange get selectionChange => selectionChangeNotifier.value;
```
matthew-carroll, line 39: I don't think clients will care about the `DocumentSelectionChange` object. They'll either want the selection, in which case they'll ask for `selection`, or they'll want the reason. So let's change this accessor to one that's called `latestSelectionChangeReason`.

super_editor/lib/src/core/document_selection.dart
```diff
@@ -208,6 +208,62 @@ class DocumentSelection {
...
 
+/// Represents a change of a [DocumentSelection].
+///
+/// The [reason] represents what caused the selection to change.
+/// For example, [CommonSelectionChangeReasons.userInteraction] represents
+/// a selection change caused by the user interacting with the editor.
+class DocumentSelectionChange {
+  const DocumentSelectionChange({
+    this.selection,
+    this.reason = CommonSelectionChangeReasons.userInteraction,
```
matthew-carroll, line 219: It's OK to have a default `reason` on the composer API, but whatever instantiates this object should probably be forced to specify the `reason`.

super_editor/lib/src/core/document_selection.dart
```diff
@@ -208,6 +208,62 @@ class DocumentSelection {
...
+
+  final DocumentSelection? selection;
+  final SelectionChangeReason reason;
+
+  @override
+  String toString() => '[DocumentSelectionChange]: selection: $selection, reason: $reason';
+}
+
+/// Represents what caused a [DocumentSelection] to change.
+class SelectionChangeReason {
```
matthew-carroll, line 235: Is there a reason that these class structures are introduced? If we want to allow any `Object` to be a `reason`, then I think all we need are a couple global constants:

```dart
/// doc comment
class SelectionReason {
  /// doc comment
  static const userInteraction = "userInteraction";

  /// doc comment
  static const contentChange = "contentChange";
}
```

Also, is there a reason that these concepts were defined in the core `document_selection.dart`? Are these artifacts referenced by any code outside of the `default_editor`?

matthew-carroll:
I'm currently working on getting us back in the test registry, and then I'm going to release a version of super_editor against Flutter's `master`. I want to wait on this breaking change until after that. If you don't get a review from me after that happens, then please remind me.

matthew-carroll:
@miguelcmedeiros @roughike - Would you like to try this PR in Superlist to see if it provides what you need?

Also, a question for you two and @angelosilvestre - this PR shows a number of places where we've adjusted behavior, despite gaining nothing from the new information. Like the example app. I'm wondering if we retain the original selection notifier for use-cases that don't care about the reason? The benefit would be fewer forced changes to existing apps that don't care about the reason, as well as the ability to use callback listeners instead of managing stream subscriptions. The downside is that developers might be confused about why we have multiple representations of the same thing.

super_editor/test/super_reader/reader_test_tools.dart
```diff
@@ -208,10 +208,11 @@ class TestDocumentConfigurator {
     assert(_document != null);
 
     final layoutKey = GlobalKey();
+    final composer = DocumentComposer(initialSelection: _selection);
     final documentContext = ReaderContext(
       document: _document!,
       getDocumentLayout: () => layoutKey.currentState as DocumentLayout,
-      selection: ValueNotifier<DocumentSelection?>(_selection),
+      composer: composer,
```
matthew-carroll, line 215: `SuperReader` should never be given a `DocumentComposer`. A `SuperReader` can't edit a document, so the composer wouldn't make sense there. That would be like taking the editing toolbar at the top of Google Docs and adding it to a PDF reader.

Also, in general, we don't want to pass a `DocumentComposer` to any location that previously worked directly with selections. This is because `DocumentComposer` is a high-level object that's narrowly designed for the `SuperEditor` editing experience.

super_editor/lib/src/core/document_composer.dart
```diff
@@ -33,21 +35,33 @@ class DocumentComposer with ChangeNotifier {
...
+    );
+    _streamController.sink.add(_latestSelectionChange);
   }
 
-  final selectionNotifier = ValueNotifier<DocumentSelection?>(null);
+  /// Emits a [DocumentSelectionChange] every time the selection changes.
+  Stream<DocumentSelectionChange> get selectionChanges => _streamController.stream;
+  late StreamController<DocumentSelectionChange> _streamController;
+
+  /// Holds the reason of the last selection change.
```
matthew-carroll, line 55: Method docs should describe behaviors. For example:

```dart
/// Returns the reason for the most recent selection change in the composer.
///
/// For example, a selection might change as a result of user interaction, or as
/// a result of another user editing content, or some other reason.
```

Please adjust any other docs with similar issues.

super_editor/lib/src/core/document_composer.dart
```diff
@@ -33,21 +35,33 @@ class DocumentComposer with ChangeNotifier {
...
+  void setSelection(DocumentSelection? newSelection, [Object reason = SelectionReason.userInteraction]) {
+    _latestSelectionChange = DocumentSelectionChange(
+      selection: newSelection,
+      reason: reason,
+    );
+    _streamController.sink.add(_latestSelectionChange);
   }
 
-  final selectionNotifier = ValueNotifier<DocumentSelection?>(null);
+  /// Emits a [DocumentSelectionChange] every time the selection changes.
```
matthew-carroll, line 51: This doc comment is mixing up the behavior of `get selectionChanges` with the behavior of listening to a stream. This "getter" method doesn't "emit" anything.

```dart
/// A stream of document selection changes.
///
/// Each new [DocumentSelectionChange] includes the most recent document selection,
/// along with the reason that the selection changed.
```

angelosilvestre:
> @miguelcmedeiros @roughike - Would you like to try this PR in Superlist to see if it provides what you need?
> 
> Also, a question for you two and @angelosilvestre - this PR shows a number of places where we've adjusted behavior, despite gaining nothing from the new information. Like the example app. I'm wondering if we retain the original selection notifier for use-cases that don't care about the reason? The benefit would be fewer forced changes to existing apps that don't care about the reason, as well as the ability to use callback listeners instead of managing stream subscriptions. The downside is that developers might be confused about why we have multiple representations of the same thing.

I think we could have both options and document when to use one or another.

angelosilvestre:
@matthew-carroll I updated this PR to add the `selectinNotifier` back.

miguelcmedeiros:
> @miguelcmedeiros @roughike - Would you like to try this PR in Superlist to see if it provides what you need?

Currently our project is built against Flutter stable channel while this PR seems to be built against Flutter master. In order to test it out, it would be helpful if this PR is targets `stable` branch.

matthew-carroll:
@miguelcmedeiros I retargeted the PR at stable

miguelcmedeiros:
> I retargeted the PR at stable

@matthew-carroll, if you try to build the demo app using Flutter stable it fails. 
Just retargeting this PR seems to not be enough, since the code in this branch is already migrated to work with flutter master (e.g. flutter stable does not know about methods like `Scrollable.maybeOf`).

matthew-carroll:
@angelosilvestre I switched the base back to `main` so that we can keep our merge target consistent. Can you return this PR to a state that works with `master`? Once this is merged, your cherry-pick can make adjustments for `stable`.

super_editor/lib/src/core/document_composer.dart
```diff
@@ -29,25 +32,75 @@ class DocumentComposer with ChangeNotifier {
...
+  ///
+  /// Listen to this [Stream] when the selection reason is needed. Otherwise, use [selectionNotifier].
+  Stream<DocumentSelectionChange> get selectionChanges => _streamController.stream;
+  late StreamController<DocumentSelectionChange> _streamController;
+
+  /// Returns the reason for the most recent selection change in the composer.
+  ///
+  /// For example, a selection might change as a result of user interaction, or as
+  /// a result of another user editing content, or some other reason.
+  Object? get latestSelectionChangeReason => _latestSelectionChange.reason;
```
matthew-carroll, line 87: Let's move this up beneath `set selection`

super_editor/lib/src/core/document_composer.dart
```diff
@@ -29,25 +32,75 @@ class DocumentComposer with ChangeNotifier {
...
+
+  /// Returns the most recent selection change in the composer.
+  ///
+  /// The [DocumentSelectionChange] includes the most recent document selection,
+  /// along with the reason that the selection changed.
+  DocumentSelectionChange get latestSelectionChange => _latestSelectionChange;
```
matthew-carroll, line 93: Let's move this up beneath `latestSelectionChangeReason`, after you move `latestSelectionChangeReason` up.

super_editor/lib/src/core/document_composer.dart
```diff
@@ -29,25 +32,75 @@ class DocumentComposer with ChangeNotifier {
...
     if (newSelection != selectionNotifier.value) {
       selectionNotifier.value = newSelection;
       notifyListeners();
     }
   }
 
+  /// Sets the current [selection] for a [Document].
+  ///
+  /// [reason] represents what caused the selection change to happen.
+  void setSelection(DocumentSelection? newSelection, [Object reason = SelectionReason.userInteraction]) {
```
matthew-carroll, line 53: Let's call this `setSelectionWithReason` so that it's clearly disambiguated from `set selection`

super_editor/lib/src/core/document_composer.dart
```diff
@@ -29,25 +32,75 @@ class DocumentComposer with ChangeNotifier {
...
+    _latestSelectionChange = DocumentSelectionChange(
+      selection: newSelection,
+      reason: reason,
+    );
+    _streamController.sink.add(_latestSelectionChange);
+
+    if (_updatingSelection) {
```
matthew-carroll, line 60: Is there any way to implement both behaviors without using a transient flag value like this? It doesn't take much for uses like this to become intractable...

super_editor/lib/src/core/document_composer.dart
```diff
@@ -29,25 +32,75 @@ class DocumentComposer with ChangeNotifier {
...
+      // The selection was changed by the selectionNotifier.
+      return;
+    }
+
+    _updatingSelection = true;
+    try {
```
matthew-carroll, line 66: Why do we need a try/catch here?

super_editor/lib/src/core/document_composer.dart
```diff
@@ -29,25 +32,75 @@ class DocumentComposer with ChangeNotifier {
...
+
+  /// Notifies whenever the current [DocumentSelection] changes.
+  ///
+  /// If the selection change reason is needed, use [selectionChanges] instead.
   final selectionNotifier = ValueNotifier<DocumentSelection?>(null);
 
   /// Clears the current [selection].
   void clearSelection() {
-    selection = null;
+    setSelection(null);
```
matthew-carroll, line 103: This looks like it should still be `selection = null`.

super_editor/lib/src/core/document_composer.dart
```diff
@@ -56,6 +109,28 @@ class DocumentComposer with ChangeNotifier {
 
   /// Returns the composition preferences for this composer.
   ComposerPreferences get preferences => _preferences;
+
+  /// Indicates wheter or not we are in the process of updating the selection.
```
matthew-carroll, line 113: "wheter" -> "whether"

super_editor/lib/src/core/document_composer.dart
```diff
@@ -56,6 +109,28 @@ class DocumentComposer with ChangeNotifier {
...
+  ///
+  /// The selection can be changed by [selectionNotifier] or by [setSelection].
+  bool _updatingSelection = false;
+
+  void _onSelectionChangedBySelectionNotifier() {
```
matthew-carroll, line 118: This selection behavior should be moved up with all other selection behavior. Otherwise, we're hiding `imeConfiguration`, `_preferences` and `get preferences` in the middle of selection behavior.

super_editor/lib/src/core/document_composer.dart
```diff
@@ -123,3 +198,35 @@ class ComposerPreferences with ChangeNotifier {
...
+  @override
+  bool operator ==(Object other) =>
+      identical(this, other) ||
+      other is DocumentSelectionChange && selection == other.selection && reason == other.reason;
+
+  @override
+  int get hashCode => (selection?.hashCode ?? 0) ^ reason.hashCode;
+}
+
+/// Holds common reasons for selection changes.
```
matthew-carroll, line 225: Let's add a second line that says:

```
/// Developers aren't limited to these selection change reasons. Any object can be passed as
/// a reason for a selection change. However, some Super Editor behavior is based on [userInteraction].
```

super_editor/lib/src/default_editor/common_editor_operations.dart
```diff
@@ -5,6 +5,7 @@ import 'dart:ui';
 import 'package:attributed_text/attributed_text.dart';
 import 'package:collection/collection.dart';
 import 'package:flutter/foundation.dart';
+import 'package:flutter/rendering.dart';
```
matthew-carroll, line 8: The only changes in this file are imports. Are these imports needed?

super_editor/lib/src/default_editor/document_gestures_mouse.dart
```diff
@@ -36,7 +38,7 @@ class DocumentMouseInteractor extends StatefulWidget {
     this.focusNode,
     required this.document,
     required this.getDocumentLayout,
-    required this.selection,
+    required this.composer,
```
matthew-carroll, line 41: We don't want to pass the `DocumentComposer` if we can avoid it. Is there a strong reason why it's needed here?

super_editor/lib/src/default_editor/document_gestures_mouse.dart
```diff
@@ -134,9 +138,14 @@ class _DocumentMouseInteractorState extends State<DocumentMouseInteractor>
           RawKeyboard.instance.keysPressed.contains(LogicalKeyboardKey.shiftRight) ||
           RawKeyboard.instance.keysPressed.contains(LogicalKeyboardKey.shift)) &&
       // TODO: this condition doesn't belong here. Move it to where it applies
-      widget.selection.value != null;
+      widget.composer.selection != null;
 
-  void _onSelectionChange() {
+  void _onSelectionChange(DocumentSelectionChange selectionChange) {
+    if (selectionChange.reason != SelectionReason.userInteraction) {
```
matthew-carroll, line 144: Please make sure this behavior is documented in the appropriate places. I don't think I noticed any Dart Doc adjustments in this file.

super_editor/test/super_editor/supereditor_scrolling_test.dart
```diff
@@ -143,12 +143,12 @@ void main() {
 
       // Place the caret at the end of the document, which causes the editor to
       // scroll to the bottom.
-      docContext.editContext.composer.selection = DocumentSelection.collapsed(
+      docContext.editContext.composer.setSelection(DocumentSelection.collapsed(
```
matthew-carroll, line 146: Why is this change needed?

super_editor/test/super_editor/supereditor_scrolling_test.dart
```diff
@@ -200,12 +200,12 @@ void main() {
 
       // Place the caret at the end of the document, which should cause the
       // editor to scroll to the bottom.
-      docContext.editContext.composer.selection = DocumentSelection.collapsed(
+      docContext.editContext.composer.setSelection(DocumentSelection.collapsed(
```
matthew-carroll, line 203: Why is this change needed?

super_editor/test/super_editor/supereditor_scrolling_test.dart
angelosilvestre: I added this change when the selectionNotifier was removed in the first commits. Will revert it.

super_editor/lib/src/default_editor/document_gestures_mouse.dart
angelosilvestre: To remove the `DocumentComposer` we need to pass both `selectionChanges` to listen for the document changes with a reason and the `selectionNotifier`, because we clear the selection at some places.

super_editor/lib/src/core/document_composer.dart
angelosilvestre: I added this to avoid some listener throwing an exception and preventing `_updatingSelection` to be reset.

super_editor/lib/src/core/document_composer.dart
angelosilvestre: I think the other option is to remove the listener `_onSelectionChangedBySelectionNotifier` before updating the `selectionNotifier`, and changing `_onSelectionChangedBySelectionNotifier` to also emit `DocumentSelectionChange`s. I'm not sure if this would be better.

super_editor/lib/src/default_editor/document_gestures_mouse.dart
matthew-carroll: Let's see what that looks like

super_editor/lib/src/core/document_composer.dart
matthew-carroll: If something is throwing exceptions then we should fix the exceptions, rather than silently hide them.

angelosilvestre: I mean exceptions in user code which could listen for selection changes. BTW, this is a try/finally so it won't hide any exception.

super_editor/lib/src/default_editor/document_gestures_mouse.dart
angelosilvestre: I'm not sure where is the place that I should document this. Should it be in the `DocumentMouseInteractor` dart doc?

super_editor/lib/src/core/document_composer.dart
matthew-carroll: What leads you to believe that there will be exceptions in user code? I don't think we wrap callbacks with try/catch anywhere else. I don't think I've seen any special reason to do that here...

super_editor/lib/src/default_editor/document_gestures_mouse.dart
matthew-carroll: Yeah, the class doc seems like a reasonable place.

super_editor/lib/src/core/document_composer.dart
angelosilvestre: Ok, I think we can just remove it.

super_editor/lib/src/default_editor/document_gestures_mouse.dart

super_editor/lib/src/default_editor/document_gestures_mouse.dart
angelosilvestre: Updated.

super_editor/lib/src/default_editor/document_gestures_mouse.dart
```diff
@@ -46,7 +52,8 @@ class DocumentMouseInteractor extends StatefulWidget {
 
   final Document document;
   final DocumentLayoutResolver getDocumentLayout;
-  final ValueNotifier<DocumentSelection?> selection;
+  final Stream<DocumentSelectionChange> selectionChanges;
+  final ValueNotifier<DocumentSelection?> selectionNotifier;
```
matthew-carroll, line 56: This property isn't actually being used to receive notifications, right? If so, we shouldn't broaden the contract to include unnecessary behaviors. If this property is only being used to update the selection, then we can simplify it to the following:

```dart
final void Function(DocumentSelection?) updateSelection;
```

Also, please add appropriate dart docs to the new properties.

super_editor/lib/src/default_editor/common_editor_operations.dart
```diff
@@ -22,7 +22,9 @@ import 'package:super_editor/src/infrastructure/_logging.dart';
 import 'attributions.dart';
 import 'horizontal_rule.dart';
 import 'image.dart';
+import 'list_items.dart';
```
matthew-carroll, line 25: I commented on this before but I didn't notice a response. Is there a reason that we need new imports in a file that's unchanged?

super_editor/lib/src/core/document_composer.dart
```diff
@@ -29,27 +32,97 @@ class DocumentComposer with ChangeNotifier {
...
+      // We are already emitted a [DocumentSelectionChange].
+      return;
+    }
+
+    _updatingSelection = true;
+
+    // The selection was changed using the selectionNotifier.
+    // We need to emit a new DocumentSelectionChange and sync the selectionNotifier
+    // selection with _latestSelectionChange selection.
+    setSelectionWithReason(selectionNotifier.value);
```
matthew-carroll, line 121: Can't we avoid the `_updatingSelection` flag by directly updating the selection, instead of calling `setSelectionWithReason()`? This method is called when the notifier is given a new selection. All we need to do at this point is emit a change event on the change stream, right? If so, why not do that here, directly, and avoid any concern about calling ourselves again?

super_editor/lib/src/default_editor/document_gestures_mouse.dart
angelosilvestre: We also need the `selectionNotifier` to pass it to `DocumentSelectionOnFocusMixin` and also for reading the current selection.

I forgot to mention this before, sorry.

super_editor/lib/src/default_editor/common_editor_operations.dart
angelosilvestre: Updated. Not sure why this was imported in the first place.

super_editor/lib/src/core/document_composer.dart
angelosilvestre: Updated.

super_editor/lib/src/default_editor/document_gestures_mouse.dart
matthew-carroll: It looks like we're still missing docs on these properties

matthew-carroll (approved):
There are still a couple docs missing, but I'm going to merge this so we can move forward.

matthew-carroll:
@angelosilvestre don't forget to cherry-pick this to `stable`.

PR #863, opened 2022-11-19
[super_editor_markdown] Add header alignment

vursin:
I know the plugin supports custom parsers but I think the header alignment should be added by default to be compatible with the super_editor, so I made this PR.


matthew-carroll:
@vursin  - Please start with an issue ticket that describes your goal and where `super_editor` is stopping you, as mentioned our contributing guide: https://github.com/superlistapp/super_editor/blob/main/super_editor/CONTRIBUTING.md#feature-requests-must-include-the-motivation

PR #866, opened 2022-11-21
[SuperEditor] Avoid scrolling when dragging over an image (Resolves #812)

### LINKED ISSUE #812: Scroll jumps to the image on selection
Scroll jumps to the image when there's a big one inside the content that I'm trying to select. I tried it on the current master branch (macos desktop)

angelosilvestre:
[SuperEditor] Avoid scrolling when dragging over an image (Resolves #812)

When dragging over an image, the editor scrolls to display the whole image.

This PR changes the mouse interactor to avoid scrolling when the selection extent is an `ImageNode`.

The downside is that when a user changes the selection manually, it won't scroll also.

We could also extract the `_FakeImageComponentBuilder` to a shared file where we could reuse it.

### DIFF
--- a/super_editor/lib/src/default_editor/document_gestures_mouse.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_mouse.dart
@@ -10,6 +10,7 @@ import 'package:super_editor/src/core/document_layout.dart';
 import 'package:super_editor/src/core/document_selection.dart';
 import 'package:super_editor/src/default_editor/document_scrollable.dart';
 import 'package:super_editor/src/default_editor/document_selection_on_focus_mixin.dart';
+import 'package:super_editor/src/default_editor/image.dart';
 import 'package:super_editor/src/default_editor/selection_upstream_downstream.dart';
 import 'package:super_editor/src/default_editor/text_tools.dart';
 import 'package:super_editor/src/document_operations/selection_operations.dart';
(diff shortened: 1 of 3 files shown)


abidex4yemi:
@matthew-carroll Will this be reviewed and merged anytime soon?

super_editor/lib/src/default_editor/document_gestures_mouse.dart
```diff
@@ -150,24 +151,36 @@ class _DocumentMouseInteractorState extends State<DocumentMouseInteractor>
...
-        final globalExtentRect = _getSelectionExtentAsGlobalRect();
-        if (globalExtentRect != null) {
-          widget.autoScroller.ensureGlobalRectIsVisible(globalExtentRect);
-        }
-      });
+
+    final selection = widget.selectionNotifier.value;
+    if (selection != null) {
+      final node = widget.document.getNodeById(selection.extent.nodeId);
+      if (node is ImageNode) {
```
matthew-carroll, line 167: I think we should generalize this, if we can. What do you think about about never auto-scrolling to the bottom of block components, which includes things like an HR, and probably videos, etc? Can you think of any downsides?

super_editor/test/super_editor/supereditor_scrolling_test.dart
```diff
@@ -253,5 +253,91 @@ void main() {
...
+        ),
+        false,
+      );
+    });
   });
 }
+
+/// A [ComponentBuilder] which builds an [ImageComponent] that always renders
+/// images as a [SizedBox] with the given [size].
+class _FakeImageComponentBuilder implements ComponentBuilder {
```
matthew-carroll, line 313: You mentioned making this re-usable. We might as well do that. See if there's an existing test tools file that makes sense for fake components and test nodes. If not, let's create one.

super_editor/lib/src/default_editor/document_gestures_mouse.dart
angelosilvestre: I think we could do this for all block components without causing issues.

matthew-carroll: Ok. Let's give that a try.

super_editor/test/super_editor/supereditor_scrolling_test.dart

matthew-carroll (approved):
LGTM

PR #867, opened 2022-11-24
[SuperEditor][DemoApp] Fix toolbar dismissal on copy (Resolves #818)

### LINKED ISSUE #818: Cut/Copy/Paste Toolbar Dismissal
After using any of these actions, the toolbar remains visible and an additional click on the document is required in order to dismiss it. Expected behavior would be that the toolbar is dismissed as soon as one of the actions is pressed.

angelosilvestre:
[SuperEditor][DemoApp] Fix toolbar dismissal on copy. Resolves #818 

Currently, tapping on the copy button doesn't close the toolbar. Tapping on cut or paste closes it, but this is incidental, the toolbar is closed because the document changes and the touch interactors hide the toolbar when this happens. 

It isn't possible for a package user to control the visibility of the toolbar. The toolbar is controlled by `AndroidDocumentGestureEditingController` and `IosDocumentGestureEditingController`, which aren't exposed in the public API. These controllers are created inside the touch interactors and can't be created elsewhere, as they hold `LayerLink's` to position the toolbar.

Both `AndroidDocumentGestureEditingController` and `IosDocumentGestureEditingController` extends `MagnifierAndToolbarController`, which is the class that actually controls the visibility of the toolbar and the magnifier.

This PR changes `AndroidDocumentGestureEditingController` and `IosDocumentGestureEditingController` to don't extend `MagnifierAndToolbarController` anymore. Instead, they will receive one in the constructor.

`SuperEditor` will expose the `MagnifierAndToolbarController` in its public API. That way, users can control the visibility of the toolbar.

### DIFF
--- a/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
@@ -39,6 +39,7 @@ class AndroidDocumentTouchInteractor extends StatefulWidget {
     required this.popoverToolbarBuilder,
     this.createOverlayControlsClipper,
     this.showDebugPaint = false,
+    this.toolbarController,
     required this.child,
   }) : super(key: key);
 
@@ -50,6 +51,10 @@ class AndroidDocumentTouchInteractor extends StatefulWidget {
   final ValueNotifier<DocumentSelection?> selection;
 
   final ScrollController? scrollController;
+
+  /// [MagnifierAndToolbarController] that governs the display and position of
+  /// the magnifier and the floating toolbar.
+  final MagnifierAndToolbarController? toolbarController;
 
   /// The closest that the user's selection drag gesture can get to the
   /// document boundary before auto-scrolling.
@@ -86,6 +91,10 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
   // The alternative case is the one in which this interactor defers to an
   // ancestor scrollable.
   late ScrollController _scrollController;
+
+  /// [MagnifierAndToolbarController] that governs the display and position of
+  /// the magnifier and the floating toolbar.
+  late MagnifierAndToolbarController _toolbarController;
   // The ScrollPosition attached to the _ancestorScrollable, if there's an ancestor
   // Scrollable.
   ScrollPosition? _ancestorScrollPosition;
@@ -144,9 +153,12 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
     // TODO: rely solely on a ScrollPosition listener, not a ScrollController listener.
     _scrollController.addListener(_onScrollChange);
 
+    _toolbarController = widget.toolbarController ?? MagnifierAndToolbarController();
+
     _editingController = AndroidDocumentGestureEditingController(
       documentLayoutLink: _documentLayoutLink,
       magnifierFocalPointLink: _magnifierFocalPointLink,
+      toolbarController: _toolbarController,
     );
 
     widget.document.addListener(_onDocumentChange);
@@ -211,6 +223,11 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
 
     if (widget.getDocumentLayout != oldWidget.getDocumentLayout) {
       onDocumentLayoutResolverReplaced(widget.getDocumentLayout);
+    }
+
+    if (widget.toolbarController != oldWidget.toolbarController) {
+      _toolbarController = widget.toolbarController ?? MagnifierAndToolbarController();
+      _editingController.toolbarController = _toolbarController;
     }
   }
 
--- a/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
@@ -39,7 +39,7 @@ class AndroidDocumentTouchInteractor extends StatefulWidget {
     required this.popoverToolbarBuilder,
     this.createOverlayControlsClipper,
     this.showDebugPaint = false,
-    this.toolbarController,
+    this.overlayController,
     required this.child,
   }) : super(key: key);
 
@@ -52,9 +52,8 @@ class AndroidDocumentTouchInteractor extends StatefulWidget {
 
   final ScrollController? scrollController;
 
-  /// [MagnifierAndToolbarController] that governs the display and position of
-  /// the magnifier and the floating toolbar.
-  final MagnifierAndToolbarController? toolbarController;
+  /// Shows, hides, and positions a floating toolbar and magnifier.
+  final MagnifierAndToolbarController? overlayController;
 
   /// The closest that the user's selection drag gesture can get to the
   /// document boundary before auto-scrolling.
@@ -92,9 +91,8 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
   // ancestor scrollable.
   late ScrollController _scrollController;
 
-  /// [MagnifierAndToolbarController] that governs the display and position of
-  /// the magnifier and the floating toolbar.
-  late MagnifierAndToolbarController _toolbarController;
+  /// Shows, hides, and positions a floating toolbar and magnifier.
+  late MagnifierAndToolbarController _overlayController;
   // The ScrollPosition attached to the _ancestorScrollable, if there's an ancestor
   // Scrollable.
   ScrollPosition? _ancestorScrollPosition;
@@ -153,12 +151,12 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
     // TODO: rely solely on a ScrollPosition listener, not a ScrollController listener.
     _scrollController.addListener(_onScrollChange);
 
-    _toolbarController = widget.toolbarController ?? MagnifierAndToolbarController();
+    _overlayController = widget.overlayController ?? MagnifierAndToolbarController();
 
     _editingController = AndroidDocumentGestureEditingController(
       documentLayoutLink: _documentLayoutLink,
       magnifierFocalPointLink: _magnifierFocalPointLink,
-      toolbarController: _toolbarController,
+      overlayController: _overlayController,
     );
 
     widget.document.addListener(_onDocumentChange);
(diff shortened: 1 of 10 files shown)


abidex4yemi:
@angelosilvestre Please any update on this PR, thanks.

angelosilvestre:
@abidex4yemi This PR is waiting the atomic commands PR to be merged first.

abidex4yemi:
Alright, thank you.

matthew-carroll:
@angelosilvestre we can go ahead and review this because we're still waiting on client feedback for atomic commands.

Are there any further changes you wanna make before review?

Can you elaborate on why it was important to make sure that the controllers don't extend the magnifier controller? I don't have a problem with that change, but I didn't understand why it was relevant here.

angelosilvestre:
@matthew-carroll You can review the PR.

It wasn't really important to change the controllers to don't extend the magnifier controller, but as they will now receive a `MagnifierAndToolbarController` in the constructor, I thought it would be simpler to avoid extending `MagnifierAndToolbarController`.

I could revert that change if needed.

matthew-carroll:
> but as they will now receive a MagnifierAndToolbarController in the constructor, I thought it would be simpler to avoid extending MagnifierAndToolbarController.

I think I'm still missing the point. If the existing controllers already extended `MagnifierAndToolbarController`, then why would you need to pass one in the constructor? Don't they already inherit everything they need?

angelosilvestre:
The users won't instantiate the existing controllers because they require a layer link.

matthew-carroll:
I'm not sure what that means. Can you show a snippet of code that demonstrates why the existing class hierarchy can't be used to accomplish the goal of not dismissing a toolbar?

angelosilvestre:
For example, the controller is created inside the iOS touch interactor:

https://github.com/superlistapp/super_editor/blob/a435690023a24c4874061cd2594170c7cf8be7c6/super_editor/lib/src/default_editor/document_gestures_touch_ios.dart#L158-L163

Because this controller requires `LayerLink`, the user can't create this controller outside the interactor.

With these changes, users will be able to create a `MagnifierAndToolbarController` from outside the editor.

angelosilvestre:
It could work, but then users will need to use the specific controller (`AndroidDocumentGestureEditingController` or `IosDocumentGestureEditingController`) depending on the platform.

matthew-carroll:
> but then users will need to use the specific controller

Use them where? And why wouldn't someone use a platform-specific controller? What happens if a non-Android controller is given to an Android-specific widget?

As you can see, I'm struggling to follow your points here. It would be helpful if you could take a step back and frame the context, and provide small snippets, which illustrate the issue you're dealing with, and the decision you're trying to make.

angelosilvestre:
To be able to control the visibility of the toolbar, package users need to give `SuperEditor` a controller.

Currently, we have `MagnifierAndToolbarController`, which is the base class for toolbar controllers. This class isn't used directly. Instead, the mobile interactors have their own specialized controllers (`IosDocumentGestureEditingController` and `AndroidDocumentGestureEditingController`). These controllers have other responsibilities, like auto-hiding the collapsed drag handles after a period of time and controlling upstream/downstream handle offsets. They aren't visible outside of the interactors.

With this PR, I'm trying to separate `MagnifierAndToolbarController` from the specialized controllers. `IosDocumentGestureEditingController` and  `AndroidDocumentGestureEditingController` will continue to exist and be instantiated inside the mobile interactors, but the `MagnifierAndToolbarController` will be the source of truth for the toolbar visibility. The specialized controllers will receive a `MagnifierAndToolbarController` and will use its toolbar visibility.

That way, package users will be able to create a `MagnifierAndToolbarController` to manually show/hide the toolbar.

matthew-carroll:
Ok. Thanks. That makes sense.

It raises some important questions. It's true that what we're concerned with right now is controlling toolbar appearance. But why shouldn't clients be able to make the Android drag handle re-appear whenever they want? In other words, is it appropriate to expose certain overlay behaviors that we deem controllable, but then hide others? I don't know about this particular case, but one of our biggest problems with the Flutter framework is that they do that stuff all the time. Then we need the stuff they've hidden, and we can't get to it.

I suppose we can make this change for now, but we should come back and offer a holistic solution to this problem. I think `androidHandleColor`, `androidToolbarBuilder`, `iOSHandleColor`, and `iOSToolbarBuilder` are all a mistake from an API standpoint. It's far too overly specific and specialized.

I'll do another review pass.

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
```diff
@@ -51,6 +52,10 @@ class AndroidDocumentTouchInteractor extends StatefulWidget {
 
   final ScrollController? scrollController;
 
+  /// [MagnifierAndToolbarController] that governs the display and position of
```
matthew-carroll, line 55: Let's try not to use the word "govern" any more. 

```
/// Shows, hides, and positions a floating toolbar and magnifier.
```

If you force yourself to avoid catch-all terms, it will push you to dig deeper into the role and responsibilities of classes, properties, and methods.

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
```diff
@@ -51,6 +52,10 @@ class AndroidDocumentTouchInteractor extends StatefulWidget {
...
+  /// the magnifier and the floating toolbar.
+  final MagnifierAndToolbarController? toolbarController;
```
matthew-carroll, line 57: Would it make more sense to call this property `overlayController`?

super_editor/lib/src/infrastructure/platforms/android/android_document_controls.dart
```diff
@@ -1,21 +1,27 @@
...
 
-import 'package:flutter/rendering.dart';
+import 'package:flutter/widgets.dart';
 import 'package:super_editor/src/infrastructure/platforms/mobile_documents.dart';
 
 /// Controls the display of drag handles, a magnifier, and a
 /// floating toolbar, assuming Android-style behavior for the
 /// handles.
-class AndroidDocumentGestureEditingController extends MagnifierAndToolbarController {
+class AndroidDocumentGestureEditingController extends ChangeNotifier {
```
matthew-carroll, line 9: We should never extend `ChangeNotifier`. Only mix it in.

super_editor/lib/src/infrastructure/platforms/android/android_document_controls.dart
```diff
@@ -1,21 +1,27 @@
...
   AndroidDocumentGestureEditingController({
     required LayerLink documentLayoutLink,
     required LayerLink magnifierFocalPointLink,
+    required MagnifierAndToolbarController toolbarController,
   })  : _documentLayoutLink = documentLayoutLink,
-        super(magnifierFocalPointLink: magnifierFocalPointLink);
+        _toolbarController = toolbarController,
+        _magnifierFocalPointLink = magnifierFocalPointLink,
+        super() {
```
matthew-carroll, line 17: Do you need this `super()` call?

super_editor/lib/src/infrastructure/platforms/ios/ios_document_controls.dart
```diff
@@ -599,6 +626,75 @@ class IosDocumentGestureEditingController extends MagnifierAndToolbarController
       notifyListeners();
     }
   }
+
+  /// Whether the toolbar currently has a designated display position.
+  ///
+  /// The toolbar should not be displayed if this is `false`, even if
+  /// [shouldDisplayToolbar] is `true`.
+  bool get isToolbarPositioned => _toolbarController.isToolbarPositioned;
```
matthew-carroll, line 634: I'm seeing what looks like a large amount of duplication between the iOS and Android controllers.

super_editor/lib/src/infrastructure/platforms/mobile_documents.dart
```diff
@@ -5,14 +5,7 @@ import 'package:super_editor/src/infrastructure/document_gestures.dart';
...
 class MagnifierAndToolbarController with ChangeNotifier {
-  MagnifierAndToolbarController({
-    required LayerLink magnifierFocalPointLink,
-  }) : _magnifierFocalPointLink = magnifierFocalPointLink;
-
-  /// A `LayerLink` whose top-left corner sits at the location where the
-  /// magnifier should magnify.
-  LayerLink get magnifierFocalPointLink => _magnifierFocalPointLink;
-  final LayerLink _magnifierFocalPointLink;
+  MagnifierAndToolbarController();
```
matthew-carroll, line 8: Do you need a constructor here at all?

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
angelosilvestre: Updated

super_editor/lib/src/infrastructure/platforms/android/android_document_controls.dart
angelosilvestre: Updated.

super_editor/lib/src/infrastructure/platforms/android/android_document_controls.dart
angelosilvestre: Isn't needed anymore.

super_editor/lib/src/infrastructure/platforms/mobile_documents.dart
angelosilvestre: Removed.

super_editor/lib/src/infrastructure/platforms/ios/ios_document_controls.dart
angelosilvestre: Should we re-introduce some base class for the common behavior? Or maybe pass the `overlayController` directly in the places where it's needed ?

matthew-carroll: A base class should work

angelosilvestre: Updated.

super_editor/lib/src/infrastructure/platforms/mobile_documents.dart
```diff
@@ -107,6 +98,109 @@ class MagnifierAndToolbarController with ChangeNotifier {
   }
 }
 
+/// Controls the display and position of a magnifier and a floating toolbar
+/// using a [MagnifierAndToolbarController] as the source of truth.
+class GestureEditingController with ChangeNotifier {
```
matthew-carroll, line 103: This isn't a great name, but I see it's based on the other controller names. I guess I gave those poor names, first. We can come back at a later time and look at renaming these things.

matthew-carroll (approved):
LGTM

PR #870, opened 2022-11-26
Make commands atomic (Resolves #748)

### LINKED ISSUE #748: [super_editor] - Make commands atomic

matthew-carroll:
This PR converts all existing document editing and selection change behavior to use a request/command/event triad.

A desired outcome begins with a "request".

Each "request" is passed down a chain of responsibility to find a compatible "command".

The first compatible "command" is executed, which changes the underlying document. One "command" might cause other "commands" to run. Each "command" accumulates "change events" as it mutates the document. Those "change events" are saved, in-order, until all commands are finished. Then, all of those "change events" are sent to document listeners as a "change log".

The reason that "requests" are separated from "commands" is because "commands" include implementation details for document alteration. For example, a "command" would know about a database that holds document information. As a result, the same request, such as "delete selected text", might be handled by different "commands", based on the specific implementation of the document.


matthew-carroll:
@miguelcmedeiros @roughike - Here's a first draft of the atomic command changes. I probably could have continued to massage the API, but I think it would be more effective to have you guys try to start integrating the API and see what makes sense and what doesn't. We can do some pair-programming, if that would help.

matthew-carroll:
@Jethro87 - FYI, you'll want to at least be aware of this PR. It's a foundational change to Super Editor, which is needed for undo/redo support, as well as concurrent editing support. Feel free to provide any opinions on the API.

@venkatd - I know you're working on other things these days, but just in case you're curious, this PR is the beginning of Super Editor's transition to undo/redo.

matthew-carroll:
@angelosilvestre - CC for visibility. No review needed, but this PR will change a large swathe of the codebase. Feel free to ask any questions.

zaynetro:
This is an excellent feature!

One thing that would be nice to have is the ability to inspect and alter the transaction changes before committing. Use case: disallow deleting certain nodes, like unselectable HR or an image.

Sorry, for if this isn't the right place or time for requests :/

matthew-carroll:
@zaynetro - That's a reasonable request. Once these changes are merged, I would recommend trying them for your use-case and if they don't do what you need, then file an issue describing what you're looking for. This API will likely evolve quite a bit. In fact, it's still evolving in this PR, which is why it hasn't been merged yet.

raulmabe-labhouse:
Hi 👋🏼  How is the state of this PR? I assume its low priority as its been inactive for very long. Do you have any insights on when will be merged?

miguelcmedeiros:
@matthew-carroll I guess we can close this one.

Issue #791, opened 2022-09-26
Platform Exception on Android


MenuItem207:
Steps to replicate:
1. Create a Super Editor:
```
SuperEditor(
  inputSource:
      DocumentInputSource.ime,
  editor: DocumentEditor(
    document: MutableDocument(
      nodes: [
        ParagraphNode(
          id: DocumentEditor
              .createNodeId(),
          text: AttributedText(
            text: '',
          ),
        ),
      ],
    ),
  ),
)
```

2. Type in 'Abc' 
3. Hit Enter
4. Type in 'a'
5. Backspace all the way until A
6. Repeat steps 2 - 5 once
7. Exception:

```
Exception has occurred.
PlatformException (PlatformException(error, invalid composing start: 0, null, java.lang.IndexOutOfBoundsException: invalid composing start: 0
	at io.flutter.embedding.engine.systemchannels.TextInputChannel$TextEditState.<init>(TextInputChannel.java:777)
	at io.flutter.embedding.engine.systemchannels.TextInputChannel$TextEditState.fromJson(TextInputChannel.java:738)
	at io.flutter.embedding.engine.systemchannels.TextInputChannel$1.onMethodCall(TextInputChannel.java:101)
	at io.flutter.plugin.common.MethodChannel$IncomingMethodCallHandler.onMessage(MethodChannel.java:262)
	at io.flutter.embedding.engine.dart.DartMessenger.invokeHandler(DartMessenger.java:295)
	at io.flutter.embedding.engine.dart.DartMessenger.lambda$dispatchMessageToQueue$0$io-flutter-embedding-engine-dart-DartMessenger(DartMessenger.java:319)
	at io.flutter.embedding.engine.dart.DartMessenger$$ExternalSyntheticLambda0.run(Unknown Source:12)
	at android.os.Handler.handleCallback(Handler.java:938)
	at android.os.Handler.dispatchMessage(Handler.java:99)
	at android.os.Looper.loop(Looper.java:236)
	at android.app.ActivityThread.main(ActivityThread.java:7889)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:600)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:967)
))
```

angelosilvestre:
This seems to be a Flutter bug.
I filed https://github.com/flutter/flutter/issues/113253

[...]

PR #877, opened 2022-12-05
[SuperEditor] Handle software keyboard gestures (Resolves #817)

### LINKED ISSUE #817: Can't move caret by swiping on spacebar (Gboard)
Normally when using Gboard you can swipe your finger backward and forward across the length of the spacebar to move the caret backwards and forwards in a precise manner from it's current position.  This doesn't appear to work, along with another gesture, swipe left on the delete key to delete whole words.

angelosilvestre:
[SuperEditor] Handle software keyboard gestures. Resolves #817

Some Android software keyboards, like GBoard, have some gestures:

* Swiping over the spacebar to change the selection.
* Swiping from right to left over the backspace key to delete a word.

The spacebar gesture generates a non-text delta and the backspace gesture generates a non-text delta to select the word followed by a deletion delta.

Currently, we don't perform any action when we receive a `TextEditingDeltaNonTextUpdate`, so these gestures don't work.

This PR changes the delta handling to change the document selection when we receive a `TextEditingDeltaNonTextUpdate`. This change use a pre-existing method to convert the selection from the delta to a `DocumentSelection`.

After the changes:

https://user-images.githubusercontent.com/7597082/205769007-888ac8e0-5667-40a5-ad38-28d55eaacadb.mp4

On the Samsung keyboard, the spacebar gesture is already working, because it generates key events instead of deltas.

During manual tests I noticed some exceptions, but these exceptions already happen without this change. The exceptions are related to https://github.com/superlistapp/super_editor/issues/791

### DIFF
--- a/super_editor/lib/src/default_editor/document_input_ime.dart
+++ b/super_editor/lib/src/default_editor/document_input_ime.dart
@@ -291,10 +291,7 @@ class _DocumentImeInteractorState extends State<DocumentImeInteractor>
     editorImeLog.fine("IME value before applying deltas: $imeValueBeforeChange");
 
     _isApplyingDeltas = true;
-    widget.softwareKeyboardHandler.applyDeltas(
-      textEditingDeltas,
-      onConvertSelection: (selection) => _currentImeSerialization?.imeToDocumentSelection(selection),
-    );
+    widget.softwareKeyboardHandler.applyDeltas(textEditingDeltas);
     _isApplyingDeltas = false;
 
     editorImeLog.fine("Done applying deltas. Serializing the document and sending to IME.");
@@ -528,19 +525,26 @@ class DocumentImeSerializer {
       editorImeLog.fine("Mapping the IME base/extent to their corresponding doc positions without modification.");
     }
 
+    final base = _imeToDocumentPosition(
+      imeSelection.base,
+      isUpstream: imeSelection.base.affinity == TextAffinity.upstream,
+    );
+    final extent = _imeToDocumentPosition(
+      imeSelection.extent,
+      isUpstream: imeSelection.extent.affinity == TextAffinity.upstream,
+    );
+
+    if (base == null || extent == null) {
+      return null;
+    }
+
     return DocumentSelection(
-      base: _imeToDocumentPosition(
-        imeSelection.base,
-        isUpstream: imeSelection.base.affinity == TextAffinity.upstream,
-      ),
-      extent: _imeToDocumentPosition(
-        imeSelection.extent,
-        isUpstream: imeSelection.extent.affinity == TextAffinity.upstream,
-      ),
+      base: base,
+      extent: extent,
     );
   }
 
-  DocumentPosition _imeToDocumentPosition(TextPosition imePosition, {required bool isUpstream}) {
+  DocumentPosition? _imeToDocumentPosition(TextPosition imePosition, {required bool isUpstream}) {
     for (final range in _imeRangesToDocTextNodes.keys) {
       if (imePosition.offset >= range.start && imePosition.offset <= range.end) {
         final node = _doc.getNodeById(_imeRangesToDocTextNodes[range]!)!;
@@ -570,7 +574,7 @@ class DocumentImeSerializer {
 
     editorImeLog.shout(
         "Couldn't map an IME position to a document position. IME position: $imePosition. The selected offset range is: ${_imeRangesToDocTextNodes.keys.last.start} -> ${_imeRangesToDocTextNodes.keys.last.end}");
-    throw Exception("Couldn't map an IME position to a document position. IME position: $imePosition");
+    return null;
   }
 
   TextSelection documentToImeSelection(DocumentSelection docSelection) {
@@ -835,10 +839,7 @@ class SoftwareKeyboardHandler {
   final CommonEditorOperations commonOps;
 
   /// Applies the given [textEditingDeltas] to the [Document].
-  void applyDeltas(
-    List<TextEditingDelta> textEditingDeltas, {
-    ImeSelectionConverter? onConvertSelection,
-  }) {
+  void applyDeltas(List<TextEditingDelta> textEditingDeltas) {
     editorImeLog.info("Applying ${textEditingDeltas.length} IME deltas to document");
 
     for (final delta in textEditingDeltas) {
@@ -850,7 +851,7 @@ class SoftwareKeyboardHandler {
       } else if (delta is TextEditingDeltaDeletion) {
         _applyDeletion(delta);
       } else if (delta is TextEditingDeltaNonTextUpdate) {
-        _applyNonTextChange(delta, onConvertSelection: onConvertSelection);
+        _applyNonTextChange(delta);
       } else {
         editorImeLog.shout("Unknown IME delta type: ${delta.runtimeType}");
       }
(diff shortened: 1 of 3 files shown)


super_editor/lib/src/default_editor/document_input_ime.dart
```diff
@@ -291,7 +291,10 @@ class _DocumentImeInteractorState extends State<DocumentImeInteractor>
     editorImeLog.fine("IME value before applying deltas: $imeValueBeforeChange");
 
     _isApplyingDeltas = true;
-    widget.softwareKeyboardHandler.applyDeltas(textEditingDeltas);
+    widget.softwareKeyboardHandler.applyDeltas(
+      textEditingDeltas,
+      onConvertSelection: (selection) => _currentImeSerialization?.imeToDocumentSelection(selection),
```
matthew-carroll, line 296: My initial thought is that I don't think we should pass this delegate here. It creates a weird two-way relationship that otherwise doesn't exist. Can you explain what's going on here and why you think this is the right approach?

angelosilvestre: I use this delegate because `DocumentImeSerializer` already has some logic to convert the `TextSelection` to a `DocumentSelection`. 

We could also subclass `TextEditingDeltaNonTextUpdate` so we could hold the `DocumentSelection` in the delta. This way we would need to replace the `TextEditingDeltaNonTextUpdate` from the framework with our custom delta.

super_editor/test/super_editor/supereditor_selection_test.dart
```diff
@@ -830,3 +870,38 @@ Finder _caretFinder() {
   }
   return find.byKey(primaryCaretKey);
 }
+
+/// Simulates receiving [TextEditingDelta]s from the platform.
```
matthew-carroll, line 874: Don't we already have tooling to simulate the dispatch of any given delta? Why do we need something new for selection reporting?

angelosilvestre: The `ImeSimulator` doesn't expose a method to dispatch arbitrary deltas. Currently, it exposes only `typeText` and `backspace`.

super_editor/lib/src/default_editor/document_input_ime.dart
```diff
@@ -926,13 +935,23 @@ class SoftwareKeyboardHandler {
...
+    ImeSelectionConverter? onConvertSelection,
+  }) {
     editorImeLog.fine("Non-text change:");
     // editorImeLog.fine("App-side selection - ${currentTextEditingValue.selection}");
     // editorImeLog.fine("App-side composing - ${currentTextEditingValue.composing}");
     editorImeLog.fine("OS-side selection - ${delta.selection}");
     editorImeLog.fine("OS-side composing - ${delta.composing}");
-    // currentTextEditingValue = _currentTextEditingValue.copyWith(composing: delta.composing);
+
+    final docSelection = onConvertSelection?.call(delta.selection);
```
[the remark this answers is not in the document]
angelosilvestre: The information needed is only in the `DocumentImeSerializer` class. 

We could write a custom `TextEditingDeltaNonTextUpdate`  so we can hold the document selection in the delta. For example: 

```dart

class DocumentEditingDeltaNonTextUpdate extends TextEditingDeltaNonTextUpdate {
  final DocumentSelection documentSelection;
  // constructor
}

@override
void updateEditingValueWithDeltas(List<TextEditingDelta> textEditingDeltas) {
    for(int i = 0; i < textEditingDeltas.length; i ++) {
      TextEditingDelta delta = textEditingDeltas[i];
      if (delta is TextEditingDeltaNonTextUpdate) {
        delta = DocumentEditingDeltaNonTextUpdate(
          oldText: delta.oldText,
          selection: delta.selection,
          composing: delta.composing,
          documentSelection: _currentImeSerialization?.imeToDocumentSelection(delta.selection),
        );
        textEditingDeltas[i] = delta;
      }
    }
}
```

super_editor/test/super_editor/supereditor_selection_test.dart
matthew-carroll: Ok. Can you look into a PR for `flutter_test_robots` to report arbitrary deltas?

super_editor/lib/src/default_editor/document_input_ime.dart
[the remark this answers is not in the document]
matthew-carroll: Looking down below, I see that in the `insert()` method we're instantiating a `DocumentImeSerializer`. If we instantiate one here, can that do the job instead of the new delegate that's being passed in?

super_editor/test/super_editor/supereditor_selection_test.dart
angelosilvestre: Opened https://github.com/Flutter-Bounty-Hunters/flutter_test_robots/pull/10

super_editor/lib/src/default_editor/document_input_ime.dart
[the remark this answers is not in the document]
angelosilvestre: You are right, it seems I missed that. Updated.

super_editor/lib/src/default_editor/document_input_ime.dart
```diff
@@ -567,7 +574,7 @@ class DocumentImeSerializer {
 
     editorImeLog.shout(
         "Couldn't map an IME position to a document position. IME position: $imePosition. The selected offset range is: ${_imeRangesToDocTextNodes.keys.last.start} -> ${_imeRangesToDocTextNodes.keys.last.end}");
-    throw Exception("Couldn't map an IME position to a document position. IME position: $imePosition");
+    return null;
```
matthew-carroll, line 577: Can you tell me why `null` become a possibility here?

super_editor/lib/src/default_editor/document_input_ime.dart
```diff
@@ -816,6 +823,9 @@ class ImeConfiguration {
       keyboardActionButton.hashCode;
 }
 
+/// A function that converts a [TextSelection] from the platform to a [DocumentSelection].
+typedef ImeSelectionConverter = DocumentSelection? Function(TextSelection selection);
```
matthew-carroll, line 827: Can we get rid of this?

super_editor/test/super_editor/supereditor_selection_test.dart
matthew-carroll: I merged that other PR. I need to release a new version, but I assume we're overriding localy with the git version, right?

angelosilvestre: We are not overriding the `flutter_test_robots` dependency. Should we target a specific commit while there isn't a new version?

super_editor/lib/src/default_editor/document_input_ime.dart
angelosilvestre: Removed.

super_editor/test/super_editor/supereditor_selection_test.dart
matthew-carroll: Can we set an override to latest on `main` like we do with other overrides?

super_editor/lib/src/default_editor/document_input_ime.dart
angelosilvestre: During my tests, sometimes when changing the selection from one paragraph to another we get an invalid selection. It might be related to the other IME issues.

matthew-carroll: Ok. Can you file a bug with reproduction steps, describe the error, and link to this comment thread?

super_editor/test/super_editor/supereditor_selection_test.dart
angelosilvestre: Setting to use a git dependency we get this warning:

```
Publishable packages can't have 'git' dependencies.
Try adding a 'publish_to: none' entry to mark the package as not for publishing or remove the git dependency
```

super_editor/lib/src/default_editor/document_input_ime.dart
angelosilvestre: This seems to be related to https://github.com/superlistapp/super_editor/issues/791, but I will investigate to confirm.

super_editor/test/super_editor/supereditor_selection_test.dart
angelosilvestre: Updated

matthew-carroll (approved):
LGTM - Please be sure to cherry pick this to `stable`, as well as the previous PR that was merged.

PR #878, opened 2022-12-06
[SuperEditor] Annotate custom test methods

angelosilvestre:
[SuperEditor] Annotate custom test methods

In VS Code, test methods and test groups are decorated with Run/Debug buttons, so users can easily run or debug a single test. Currently, for our custom test methods (`testWidgetsOnDesktop`, `testWidgetsOnMobile`, and others), these buttons are not displayed.

This PR annotates the custom test methods with `@isTestGroup` and `@isTest`, so VS Code displays the buttons as expected.

### DIFF
--- a/super_editor/pubspec.yaml
+++ b/super_editor/pubspec.yaml
@@ -50,6 +50,7 @@ dev_dependencies:
   super_editor_markdown:
     path: ../super_editor_markdown
   text_table: ^4.0.1
+  meta: ^1.8.0
 
 flutter:
   # no Flutter configuration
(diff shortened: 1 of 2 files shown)


matthew-carroll:
How does this work in practice? Those annotated methods aren't tests, themselves. They're test factories, used by our hundreds of actual tests.

angelosilvestre:
The buttons will be displayed when we use these methods to define the tests:
Captura de tela 2022-12-05 211059

Without the annotation, "Run | Debug" isn't displayed above the test method.

angelosilvestre:
For example: `testGoldens` uses the `@isTestGroup` annotation.

matthew-carroll (approved):
LGTM

PR #880, opened 2022-12-06
Hide Flutter's ListenableBuilder temporarily

gspencergoog:
## Description

In https://github.com/flutter/flutter/pull/116543, the Flutter team is planning to add a new `ListenableBuilder` widget that has an identical implementation to `AnimatedBuilder`. Unfortunately, this conflicts with SuperEditor's widget of the same name.

They're virtually identical, but SuperEditor's uses a `WidgetBuilder` callback instead of a `TransitionBuilder`, which has an additional nullable `child` widget argument for performance reasons.

This PR converts SuperEditor's `ListenableBuilder` to use a `TransitionBuilder`, and updates all of the call sites where it is used. In addition, to avoid lint warnings, I added a `hide ListenableBuilder` to the Flutter material library import in each of the files that use it.

Once the Flutter change lands, SuperEditor's `ListenableBuilder` can be removed, along with the `hide` directives.

## Related Issues
 - https://github.com/flutter/flutter/issues/19255

cc @matthew-carroll

### DIFF
--- a/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
@@ -1,6 +1,6 @@
 import 'dart:math';
 
-import 'package:flutter/material.dart';
+import 'package:flutter/material.dart' hide ListenableBuilder;
 import 'package:super_editor/src/core/document.dart';
 import 'package:super_editor/src/core/document_layout.dart';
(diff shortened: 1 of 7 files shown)


matthew-carroll:
@angelosilvestre would you like to give this PR a look?

angelosilvestre (approved):
LGTM

matthew-carroll (approved):
LGTM

PR #886, opened 2022-12-10
Implement missing layout methods for RenderSuperTextLayout

knopp:
We need to get intrinsic size or dry layout for `SuperTextField` but it's not currently possible because of unimplemented methods in `RenderSuperTextLayout`.


matthew-carroll:
Can you describe why you need that information?

matthew-carroll (approved):
LGTM - in the future, it would be helpful to start with an issue ticket so that we have a historical reference for who needed a change, and why they needed it.

PR #889, opened 2022-12-13
[SuperTextField] Expose caretStyle property (Resolves #884)

### LINKED ISSUE #884: [SuperEditor][SuperTextField] - Expose CaretStyle attributes in SuperTextField
If the user wants to style the caret on desktop, they have to use a `SuperDesktopTextField` instead of the platform adaptive `SuperTextField`.

Since all other attributes are already exposed, it might make sense to allow callers of `SuperTextField` to modify the `CaretStyle.width` and `CaretStyle.borderRadius` as well.

angelosilvestre:
[SuperTextField] Expose `caretStyle` property. Resolves #884

Currently, the caret customization is different between platforms:

* On desktop, a `CaretStyle` property is exposed, which configures color, width and border radius.
* On mobile, a `caretColor` property is exposed. The width and the border radius are the defaults defined in `CaretStyle` and can't be changed.

In `SuperTextField`'s public API, we only expose a `controlsColor` property, which defines the color for the caret and for the drag handles.

This PR adds an optional `caretStyle` property to `SuperTextField`. `SuperAndroidTextField` and `SuperIOSTextField` were changed to also accept a `CaretStyle`. 

IMO, we should deprecate the `caretColor` property in `SuperAndroidTextField` and `SuperIOSTextField` in favor of the new `caretStyle` property.

### DIFF
--- a/super_editor/lib/src/infrastructure/super_textfield/android/android_textfield.dart
+++ b/super_editor/lib/src/infrastructure/super_textfield/android/android_textfield.dart
@@ -35,14 +35,20 @@ class SuperAndroidTextField extends StatefulWidget {
     this.minLines,
     this.maxLines = 1,
     this.lineHeight,
-    required this.caretColor,
+    this.caretColor,
+    this.caretStyle,
     required this.selectionColor,
     required this.handlesColor,
     this.textInputAction = TextInputAction.done,
     this.popoverToolbarBuilder = _defaultAndroidToolbarBuilder,
     this.showDebugPaint = false,
     this.padding,
-  }) : super(key: key);
+  })  : assert(caretStyle != null || caretColor != null, 'A caretStyle or a caretColor is required.'),
+        assert(
+            caretStyle == null || caretColor == null,
+            'Cannot provide both caretStyle and caretColor.\n'
+            'Use "caretStyle: CaretStyle(color: color)"'),
+        super(key: key);
 
   /// [FocusNode] attached to this text field.
   final FocusNode? focusNode;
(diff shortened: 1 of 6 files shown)


super_editor/lib/src/infrastructure/super_textfield/android/android_textfield.dart
```diff
@@ -35,14 +35,20 @@ class SuperAndroidTextField extends StatefulWidget {
     this.minLines,
     this.maxLines = 1,
     this.lineHeight,
-    required this.caretColor,
+    this.caretColor,
```
matthew-carroll, line 38: Is there a reason that we need `caretColor` with `caretStyle`?

super_editor/lib/src/infrastructure/super_textfield/ios/ios_textfield.dart
```diff
@@ -537,9 +552,12 @@ class SuperIOSTextFieldState extends State<SuperIOSTextField>
           highlightStyle: SelectionHighlightStyle(
             color: widget.selectionColor,
           ),
-          caretStyle: CaretStyle(
-            color: _floatingCursorController.isShowingFloatingCursor ? Colors.grey : widget.caretColor,
-          ),
+          caretStyle: widget.caretStyle?.copyWith(
```
matthew-carroll, line 555: This property is pretty difficult to read and understand what decisions are being made. Since we alreaady have code above the `return` statement, let's move this calculation up there, and express it in a way that makes it clear what we're doing.

super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart
```diff
@@ -92,8 +93,13 @@ class SuperTextField extends StatefulWidget {
   final WidgetBuilder? hintBuilder;
 
   /// The color of the caret, drag handles, and other controls.
+  ///
+  /// If [caretStyle] is used, the color of the caret is the [caretStyle]'s color.
   final Color? controlsColor;
 
+  /// The visual representation of the caret.
```
matthew-carroll, line 100: Let's add the same line here that we have above about overrides.

super_editor/lib/src/infrastructure/super_textfield/android/android_textfield.dart
matthew-carroll: Yeah, let's go ahead and remove it

matthew-carroll (approved):
LGTM

PR #895, opened 2022-12-16
widen intl constraint

christopherfujino:
Part of https://github.com/flutter/flutter/issues/117163, an attempt to roll package:intl was blocked by the super_editor customer tests failing to resolve pub get. Widening the constraint to `any` will ensure that future rolls will not require an upstream PR to super_editor (to resolve dependencies, the super_editor example would be forced to use the exact pin from the Flutter SDK anyway).


matthew-carroll (approved):
LGTM

PR #898, opened 2022-12-19
[SuperTextField][Desktop] Implement IME input source (Resolves #892)

### LINKED ISSUE #892: [SuperTextfield] Special characers (ñ è ä ü ö) not working when triggered from US keyboard
Certain special characters can't be entered anymore from US keyboard layouts, for example
ñ
ä
è

If they are a combination of two key inputs (for example Option+n and n to create ñ)

angelosilvestre:
[SuperTextField][Desktop] Implement IME input source. Resolves #892

Currently, trying to type characters accented characters, like `ã`, `á`, `à`, `â` and `ä` isn't working as expected. On desktop, `SuperTextField` always uses the keyboard as the input source, and Flutter seems to be reporting key events incorrectly. For example, when typing `~` followed by an `a`, I expect Flutter to only report the `ã` character. Instead, it reports a `~` character followed by an `ã` character. I filed https://github.com/flutter/flutter/issues/117294 

This PR implements IME support for the `SuperDesktopTextField`, as it's already implemented for mobile platforms. `SuperDestopTextField` will always use an `ImeAttributedTextEditingController`, as `SuperTextField` does. When using the IME, these characters are handled correctly.

Depending on a given input source, `SuperDestopTextField` will now use either the existing `SuperTextFieldKeyboardInteractor` or a new `SuperTextFieldImeInteractor`. The `SuperTextFieldImeInteractor` implementation is based on the `SuperAndroidTextField's` current implementation.

I needed to change a couple of deletion methods in `AttributedTextEditingController`, because, when using IME, we need to update the text and the selection at the same time. Otherwise, we get an exception when deleting characters at the end of the text.

This PR also renames `super_textfield_mobile_keyboard_test` to `super_textfield_ime_test` and changes the tests to run on all platforms, using IME input source.


super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -178,6 +194,9 @@ class SuperDesktopTextFieldState extends State<SuperDesktopTextField> implements
   @override
   ProseTextLayout get textLayout => _textKey.currentState!.textLayout;
 
+  @override
+  DeltaTextInputClient get imeClient => _controller;
```
matthew-carroll, line 198: Should this have a "visibleForTesting" annotation, or is `imeClient` meant to be public?

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -339,6 +355,33 @@ class SuperDesktopTextFieldState extends State<SuperDesktopTextField> implements
     return widget.decorationBuilder != null ? widget.decorationBuilder!(context, child) : child;
   }
 
+  /// Builds the widget tree that applies user input, e.g., key
+  /// presses from a keyboard, or text deltas from the IME.
+  Widget _buildInputSystem({
```
matthew-carroll, line 360: Let's call this `_buildTextInputSystem()` to make sure we don't confuse it with gestures.

I think we can do without a doc comment on this method. The name tell us everything the comment says.

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -861,6 +904,157 @@ class _SuperTextFieldKeyboardInteractorState extends State<SuperTextFieldKeyboar
   }
 }
 
+/// Governs input that comes from the operating system's
```
matthew-carroll, line 907: "Governs" usually doesn't mean anything, and this doc comment is missing context as to what is done with the IME. 

After a quick overview of the code, I don't see any IME behavior at all. I only see physical keyboard behavior.

super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart
```diff
@@ -65,6 +65,7 @@ class SuperTextField extends StatefulWidget {
     this.lineHeight,
     this.keyboardHandlers = defaultTextFieldKeyboardHandlers,
     this.padding,
+    this.inputSource,
```
matthew-carroll, line 68: Let's place this property above `keyboardHandler`s. This property sets the most important detail about text input. `keyboardHandlers` sets the second most important detail about text input. So let's keep them near each other, and in that order. This includes the property declaration down below, too.

super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart
```diff
@@ -212,6 +218,26 @@ class SuperTextFieldState extends State<SuperTextField> {
     }
   }
 
+  /// Returns the [SuperTextFieldInputSource] which should be used.
```
matthew-carroll, line 221: Let's word this first line so it's a bit more clear:

```dart
/// Returns the desired [SuperTextFieldInputSource] for this text field.
```

super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart
```diff
@@ -212,6 +218,26 @@ class SuperTextFieldState extends State<SuperTextField> {
...
+  ///
+  /// If the `inputSource` is configured, it is used. Otherwise,
```
matthew-carroll, line 223: Let's replace "`inputSource`" with "[widget.inputSource]" so that there's a code reference.

super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart
```diff
@@ -317,3 +344,9 @@ enum SuperTextFieldPlatformConfiguration {
   android,
   iOS,
 }
+
+/// The mode of user text input.
+enum SuperTextFieldInputSource {
```
matthew-carroll, line 349: I assume that we have this same kind of enum defined for Super Editor. Is that right? If so, should we create a shared definition called `TextInputSource`?

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
angelosilvestre: It should have a "visibleForTesting". Added.

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
angelosilvestre: Updated.

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
angelosilvestre: Dart doc updated. I also added the same comments about the `keyboardHandlers` that the keyboard interactor has.

super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart
angelosilvestre: Updated.

super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart
angelosilvestre: Updated.

super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart
angelosilvestre: Updated.

super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart
angelosilvestre: I renamed the existing `DocumentInputSource` to `TextInputSource`.

super_editor/lib/src/core/document_interaction.dart
```diff
@@ -1,5 +1,5 @@
 /// The mode of user text input.
-enum DocumentInputSource {
+enum TextInputSource {
```
matthew-carroll, line 2: Given that we're now using this for documents and text fields, we should move this into a more appropriate file. How about a `text_input.dart` file in the infrastructure directory?

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -861,6 +904,162 @@ class _SuperTextFieldKeyboardInteractorState extends State<SuperTextFieldKeyboar
   }
 }
 
+/// Handles the connection with the platform IME through an [ImeAttributedTextEditingController] and dispatches key events.
```
matthew-carroll, line 907: This is still a very generic statement that doesn't really say what this object does. Try reading this from a random developer's perspective. What does it mean to "handle"? Dispatches key events where? For what reason?

The first line should clearly answer the question "should I use this class?" - I don't think that's obvious with this line.

super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart
```diff
@@ -144,6 +146,11 @@ class SuperTextField extends StatefulWidget {
   /// provided and used for all text field height calculations.
   final double? lineHeight;
 
+  /// The `SuperTextField` input source, e.g., keyboard or Input Method Engine.
```
matthew-carroll, line 149: Let's use square brackets "[]" when referencing classes that are available on the existing import paths. By doing that, they should be tied to refactoring tools.

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
angelosilvestre: How about `/// Connects to the Input Method Engine to receive text input from platform.` ? Then, the key events will be detailed only in the next sentence.

matthew-carroll: I think that line describes what a `TextInputClient` does. What does this widget do?

Also, maybe I'm misunderstanding something, but key events don't come from the IME. They come through `FocusNode`s. And they probably originate with a physical keyboard.

angelosilvestre: Basically, it uses `ImeAttributedTextEditingController` to connect/disconnect to the IME when focus changes. 

It uses a `FocusNode` to receive key events. These events are passed to the keyboard handlers, so keyboard actions (like ctrl-c) can work.

matthew-carroll: Can you point me to the mobile widget that this is based on?

angelosilvestre: This implementation is based on `SuperAndroidTextField`.

I used the same `_updateSelectionAndImeConnectionOnFocusChange`

https://github.com/superlistapp/super_editor/blob/527f1e165664caa4219c1612352c107c91aa3f98/super_editor/lib/src/infrastructure/super_textfield/android/android_textfield.dart#L294-L321

matthew-carroll: I see that method in `SuperAndroidTextField`, but I don't see any widget in that file that looks like the widget that you added. I thought you copied an existing widget from mobile and then adjusted it for desktop? Is this an entirely new widget?

angelosilvestre: This is almost the same implementation of the `SuperTextFieldKeyboardInteractor`, but with the parts that connect to ime from `SuperAndroidTextField`.

super_editor/lib/src/core/document_interaction.dart
matthew-carroll, line 2: Is this resolved? I still see this delta.

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
matthew-carroll: Maybe I'm missing something, but if this widget is doing all the same keyboard stuff, and then adding the IME connection on top, wouldn't it make more sense to always include the existing keyboard widget, and then optionally inject a new IME widget as a child of the keyboard widget, when IME is requested?

super_editor/lib/src/core/document_interaction.dart
angelosilvestre: I forgot to push, sorry.

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
angelosilvestre: This would work. However, there is a small difference between the keyboard handling of this widget and the existing keyboard widget:

```dart
 KeyEventResult _onKeyPressed(FocusNode focusNode, RawKeyEvent keyEvent) {
    if (keyEvent.character != null) {
      // Character keys are handled as deltas in the text controller.
      return KeyEventResult.ignored;
    }
    ...
 }
```

We ignore character keys, because they are already handled as deltas. If we use the same keyboard widget, we will need some property like `ignoreCharacterKeys`.

matthew-carroll: I think the expected solution in that situation is to avoid passing keyboard handlers that do things you shouldn't, right?

If the user doesn't pass character handlers, then it wouldn't double on deltas, would it?

As for default handlers, we can have one list for hardware keyboards, and another list for IME.

angelosilvestre: That makes sense.

angelosilvestre: Updated. With this change, it seems we could probably get rid of the `SuperTextFieldImeInteractor` widget.

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -861,6 +909,97 @@ class _SuperTextFieldKeyboardInteractorState extends State<SuperTextFieldKeyboar
   }
 }
 
+/// Handles the connection with the platform IME when focus changes.
+class SuperTextFieldImeInteractor extends StatefulWidget {
+  const SuperTextFieldImeInteractor({
+    Key? key,
+    required this.focusNode,
+    required this.textController,
+    required this.child,
```
matthew-carroll, line 918: `child` should be last

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -861,6 +909,97 @@ class _SuperTextFieldKeyboardInteractorState extends State<SuperTextFieldKeyboar
   }
 }
 
+/// Handles the connection with the platform IME when focus changes.
```
matthew-carroll, line 912:
```
/// Opens and closes an IME connection based on changes to focus and selection.
///
/// This widget watches [focusNode] for focus changes, and [textController] for
/// selection changes.
///
/// All IME commands are handled and applied to text field text by the given [textController].
///
/// When [focusNode] gains focus, if the [textController] doesn't have a selection, the caret is
/// placed at the end of the text.
///
/// When [focusNode] loses focus, the [textController]'s selection is cleared.
```

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -1257,6 +1396,35 @@ const defaultTextFieldKeyboardHandlers = <TextFieldKeyboardHandler>[
   DefaultSuperTextFieldKeyboardHandlers.insertCharacterWhenKeyIsPressed,
 ];
 
+/// The keyboard actions that a [SuperTextField] uses by default when using [TextInputSource.ime].
```
matthew-carroll, line 1399: Let's mention why this list exists. It's like the other one, except it leaves out handlers that input text, because the IME will do that.

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -339,6 +363,30 @@ class SuperDesktopTextFieldState extends State<SuperDesktopTextField> implements
...
+        child: keyboardChild,
+      );
+    }
+
+    return SuperTextFieldKeyboardInteractor(
+      focusNode: _focusNode,
+      textController: _controller,
+      textKey: _textKey,
+      keyboardActions: widget.keyboardHandlers,
+      child: keyboardChild,
```
matthew-carroll, line 386: We can do this all in one tree:

```dart
child: widget.inputSource == TextInputSource.ime
  ? SuperTextFieldImeInteractor(
        focusNode: _focusNode,
        textController: _controller,
        isMultiline: isMultiline,
        child: child,
      )
  : child,
```

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -57,8 +58,15 @@ class SuperDesktopTextField extends StatefulWidget {
     this.maxLines = 1,
     this.decorationBuilder,
     this.onRightClick,
-    this.keyboardHandlers = defaultTextFieldKeyboardHandlers,
-  }) : super(key: key);
+    this.inputSource = TextInputSource.keyboard,
+    List<TextFieldKeyboardHandler>? keyboardHandlers,
+  }) : super(key: key) {
+    final defaultKeyboardHandlers = inputSource == TextInputSource.keyboard
```
matthew-carroll, line 64: Is it possible to set this in the initializer list and avoid the `late` keyword?

Something like:
```dart
keyboardHandlers = keyboardHandlers ?? (inputSource == TextInputSource.keyboard ? defaultTextFieldKeyboardHandlers : defaultTextFieldImeKeyboardHandlers)
```

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -94,21 +102,24 @@ class SuperDesktopTextField extends StatefulWidget {
 
   final RightClickListener? onRightClick;
 
+  /// The `SuperDesktopTextField` input source, e.g., keyboard or Input Method Engine.
+  final TextInputSource inputSource;
+
   /// Priority list of handlers that process all physical keyboard
   /// key presses, for text input, deletion, caret movement, etc.
-  final List<TextFieldKeyboardHandler> keyboardHandlers;
+  late final List<TextFieldKeyboardHandler> keyboardHandlers;
```
matthew-carroll, line 110: We should probably add a note tot his dart doc that explains the relationship to the `inputSource` and what a developer needs to do to avoid double text entry.

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
angelosilvestre: It's possible. Updated.

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
angelosilvestre: Updated.

matthew-carroll (approved):
LGTM

PR #906, opened 2023-01-04
[SuperTextField][Mobile] Expose TextInputAction

jvelezos:
In order to modify the virtual keyboard action on Mobile, `TextInputAction` needs to be exposed and configurable.

At Superlist mobile we have a multiline SuperTextField, but pressing enter on the virtual keyboard should not add a line break but be done instead. This is not possible because TextInputAction currently depends only on maxLines.

### DIFF
--- a/super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart
+++ b/super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart
@@ -162,9 +162,9 @@ class SuperTextField extends StatefulWidget {
   /// scrollable viewport.
   final EdgeInsets? padding;
 
-  /// The main action for the virtual keyboard. When `null`, it depends on whether the text
-  /// field is multiline or single line.
+  /// The main action for the virtual keyboard, e.g. [TextInputAction.done].
   ///
+  /// When `null`, it depends on whether the text field is multiline or single line.
   /// Only used on mobile.
   final TextInputAction? textInputAction;
 
--- a/super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart
+++ b/super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart
@@ -162,9 +162,9 @@ class SuperTextField extends StatefulWidget {
   /// scrollable viewport.
   final EdgeInsets? padding;
 
-  /// The main action for the virtual keyboard, e.g. [TextInputAction.done].
+  /// The main action for the virtual keyboard. When `null`, it depends on whether the text
+  /// field is multiline or single line.
   ///
-  /// When `null`, it depends on whether the text field is multiline or single line.
   /// Only used on mobile.
   final TextInputAction? textInputAction;
 
@@ -208,12 +208,8 @@ class SuperTextFieldState extends State<SuperTextField> {
 
   bool get _isMultiline => (widget.minLines ?? 1) != 1 || widget.maxLines != 1;
 
-  TextInputAction get _textInputAction {
-    if (widget.textInputAction != null) {
-      return widget.textInputAction!;
-    }
-    return _isMultiline ? TextInputAction.newline : TextInputAction.done;
-  }
+  TextInputAction get _textInputAction =>
+      widget.textInputAction ?? (_isMultiline ? TextInputAction.newline : TextInputAction.done);
 
   SuperTextFieldPlatformConfiguration get _configuration {
     if (widget.configuration != null) {
--- a/super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart
+++ b/super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart
@@ -162,9 +162,9 @@ class SuperTextField extends StatefulWidget {
   /// scrollable viewport.
   final EdgeInsets? padding;
 
-  /// The main action for the virtual keyboard. When `null`, it depends on whether the text
-  /// field is multiline or single line.
+  /// The main action for the virtual keyboard, e.g. [TextInputAction.done].
   ///
+  /// When `null`, it depends on whether the text field is multiline or single line.
   /// Only used on mobile.
   final TextInputAction? textInputAction;
 
@@ -208,7 +208,12 @@ class SuperTextFieldState extends State<SuperTextField> {
 
   bool get _isMultiline => (widget.minLines ?? 1) != 1 || widget.maxLines != 1;
 
-  TextInputAction get _defaultTextInputAction => _isMultiline ? TextInputAction.newline : TextInputAction.done;
+  TextInputAction get _textInputAction {
+    if (widget.textInputAction != null) {
+      return widget.textInputAction!;
+    }
+    return _isMultiline ? TextInputAction.newline : TextInputAction.done;
+  }
 
   SuperTextFieldPlatformConfiguration get _configuration {
     if (widget.configuration != null) {
@@ -311,7 +316,7 @@ class SuperTextFieldState extends State<SuperTextField> {
             minLines: widget.minLines,
             maxLines: widget.maxLines,
             lineHeight: widget.lineHeight,
-            textInputAction: widget.textInputAction ?? _defaultTextInputAction,
+            textInputAction: _textInputAction,
             padding: widget.padding,
           ),
         );
@@ -335,7 +340,7 @@ class SuperTextFieldState extends State<SuperTextField> {
             minLines: widget.minLines,
             maxLines: widget.maxLines,
             lineHeight: widget.lineHeight,
-            textInputAction: widget.textInputAction ?? _defaultTextInputAction,
+            textInputAction: _textInputAction,
             padding: widget.padding,
           ),
         );
--- a/super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart
+++ b/super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart
@@ -162,9 +162,10 @@ class SuperTextField extends StatefulWidget {
   /// scrollable viewport.
   final EdgeInsets? padding;
 
-  /// Only used on mobile: sets the action for the virtual keyboard.
+  /// The main action for the virtual keyboard. When `null`, it depends on whether the text
+  /// field is multiline or single line.
   ///
-  /// When null, it depends on weather the text field is multiline or not.
+  /// Only used on mobile.
   final TextInputAction? textInputAction;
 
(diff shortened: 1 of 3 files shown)


matthew-carroll:
Hi @jvelezos can you please add appropriate tests for this behavior?

@angelosilvestre can you do a first pass review of this PR once the tests are added?

jvelezos:
> Hi @jvelezos can you please add appropriate tests for this behavior?
> 
> @angelosilvestre can you do a first pass review of this PR once the tests are added?

Sure, I've added tests 👍

super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart
```diff
@@ -161,6 +162,11 @@ class SuperTextField extends StatefulWidget {
   /// scrollable viewport.
   final EdgeInsets? padding;
 
+  /// Only used on mobile: sets the action for the virtual keyboard.
+  ///
+  /// When null, it depends on weather the text field is multiline or not.
+  final TextInputAction? textInputAction;
```
angelosilvestre, line 168: IMO the first line of the dart doc should mention only what this property is for. The fact that it's only used on mobile should be placed in another line.

Also, we should use \`null\` in the dart docs.

Also, "weather" -> "whether".

super_editor/test/super_textfield/super_textfield_test.dart
```diff
@@ -289,6 +289,38 @@ void main() {
...
+            _buildScaffold(
+              child: const SuperTextField(
+                textInputAction: TextInputAction.next,
+              ),
+            ),
+          );
+
+          final innerTextField = tester.widget<SuperAndroidTextField>(find.byType(SuperAndroidTextField).first);
+
+          expect(innerTextField.textInputAction, TextInputAction.next);
```
angelosilvestre, line 305: Instead of checking the widget property, we could also look at the platform channel messages to check if the `TextInput` is actually initialized with the correct `TextInputAction`, like the "applies keyboard appearance" test.

But looking at the other `TextInputAction` related tests, they are also checking the widget properties.

angelosilvestre (commented):
@matthew-carroll I left some comments.

super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart
jvelezos: 👍 fixed docs

super_editor/test/super_textfield/super_textfield_test.dart
jvelezos: As you've pointed out, I've followed the existing approach. Happy to revisit if needed.

super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart
```diff
@@ -161,6 +162,12 @@ class SuperTextField extends StatefulWidget {
   /// scrollable viewport.
   final EdgeInsets? padding;
 
+  /// The main action for the virtual keyboard. When `null`, it depends on whether the text
```
matthew-carroll, line 165: As a rule, for Dart docs, the first line should only consist of a single sentence. Anything beyond that should be moved down a couple lines.

super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart
```diff
@@ -201,6 +208,8 @@ class SuperTextFieldState extends State<SuperTextField> {
 
   bool get _isMultiline => (widget.minLines ?? 1) != 1 || widget.maxLines != 1;
 
+  TextInputAction get _defaultTextInputAction => _isMultiline ? TextInputAction.newline : TextInputAction.done;
```
matthew-carroll, line 211: Is there any reason to distinguish between the "default" input action and the actual input action? For example, why not:

```dart
TextInputAction get _textInputAction => widget.textInputAction ?? _isMultiline 
  ? TextInputAction.newline 
  : TextInputAction.done;
```

super_editor/test/super_textfield/super_textfield_test.dart
```diff
@@ -289,6 +289,38 @@ void main() {
       });
     });
 
+    group('textInputAction', () {
```
matthew-carroll, line 292: It looks like we have an existing test file called `super_textfield_ime_test.dart`. Is there any reason that this IME configuration test isn't in that file?

super_editor/test/super_textfield/super_textfield_test.dart
```diff
@@ -289,6 +289,38 @@ void main() {
...
+      group("is applied when configured", () {
+        testWidgetsOnAndroid('(on Android)', (tester) async {
```
matthew-carroll, line 294: These names will be confusing. I think you're going to end up with "is applied when configured (on Android) (on Android)". I'm pretty sure our specific platform test methods automatically add that. But maybe I'm misremembering.

In any event, the reason the name is weird is because this should really be one test, but it's written as two tests. We should find a way to fix that. Is there a reasonable way to run this as a `testWidgetsOnMobile()` test?

super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart
jvelezos: No reason. I've changed this. However I cannot use the `??` here but the full validation since `widget.textInputAction` is not static.

super_editor/test/super_textfield/super_textfield_test.dart
jvelezos: I've relocated this test to `super_textfield_ime_test.dart` which I was not aware of. However, I would like to point out that that one contains mostly events after using ime whereas `super_textfield_test.dart` contains different dependencies as textController or padding.

I would expect to find the test for this right after the one for padding.

super_editor/test/super_textfield/super_textfield_test.dart
jvelezos: Regarding names:
The resulting name is currently not duplicated since an automatic description is only added when you use `testWidgetsOnMobile`. In fact, I structured this test just as the one on line 100.

Regarding having two tests:
I've modified two different cases on the switch case for `SuperTextField` which will render two different widgets depending on the platform and call two different platform channels. The text expects two different things in each case, so even if the input dependencies are the same, to me it makes sense to have two tests.

Once again, this is also the case for the test on line 100 and I think the reasoning behind that was the same.

super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart
matthew-carroll: I still see 2 sentences on the first line...

super_editor/test/super_textfield/super_textfield_ime_test.dart
```diff
@@ -173,6 +173,38 @@ void main() {
         });
       });
     });
+
+    group('textInputAction', () {
+      group("is applied when configured", () {
+        testWidgetsOnAndroid('(on Android)', (tester) async {
```
matthew-carroll, line 179: Ideally, we should name tests based on the significance of the behavior, rather than the implementation details.

The current test name is "textInputAction is applied when configured". This doesn't really express the significance of the behavior.

Instead, consider the test name "configures the software keyboard action button". That test name describes the goal, rather than the implementation.

super_editor/test/super_textfield/super_textfield_ime_test.dart
```diff
@@ -173,6 +173,38 @@ void main() {
...
+          await tester.pumpWidget(
+            _buildScaffold(
+              child: const SuperTextField(
+                textInputAction: TextInputAction.next,
+              ),
+            ),
+          );
+
+          final innerTextField = tester.widget<SuperAndroidTextField>(find.byType(SuperAndroidTextField).first);
```
matthew-carroll, line 188: You mentioned in another comment that there was a meaningful difference between these tests, but the only difference that I see is that one test says `.widget<SuperAndroidTextField>` and the other says `.widget<SuperIOSTextField>`.

Am I missing something?

Also, while this test ensures that a property gets set on a widget, it doesn't ensure that the inner widget ever tells the IME to configure its action button accordingly.

What do you think about a test that pumps a `SuperTextField`, taps to place the caret in the text field, and then verifies that the desired platform channel message was sent?

super_editor/test/super_textfield/super_textfield_ime_test.dart
jvelezos: Done 👍 
And I accidentally reverted the docs change but now is back.

super_editor/test/super_textfield/super_textfield_ime_test.dart
jvelezos: No you are right, they are only different from the widget's perspective, from the platform channel message it is the same test.

I've changed the test to use `testWidgetsOnMobile`. The resulting description:

✓ SuperTextField configures the software keyboard action button (on Android)
✓ SuperTextField configures the software keyboard action button (on iOS)

super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart
```diff
@@ -161,6 +162,12 @@ class SuperTextField extends StatefulWidget {
   /// scrollable viewport.
   final EdgeInsets? padding;
 
+  /// The main action for the virtual keyboard, e.g. [TextInputAction.done].
+  ///
+  /// When `null`, it depends on whether the text field is multiline or single line.
```
matthew-carroll, line 167: This line is a bit confusing, and it doesn't really tell the reader what the behavior will be. 

The line should read something like "When `null`, and in single-line mode, the action will be XXXX. When in multi-line mode, the action will be XXXX."

Also, please add a blank line between the `null` section and the "Only used on mobile" line.

matthew-carroll (commented):
One last comment

super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart
jvelezos: done 👍

matthew-carroll (approved):
LGTM - Thanks for contributing. Please be sure to create a followup PR that cherry picks this change to `stable`.

matthew-carroll:
PS - I couldn't find an issue ticket for this. In the future, please be sure to start with an issue ticket so that we can mark the issue as resolved by a given PR. Thanks.

Issue #907, opened 2023-01-05
RawGestureDetector should respect MediaQuery gestureSettings


knopp:
Similar to how `GestureDetector` does it, `RawGestureDetector` should get the `MediaQuery` for gesture settings and apply it to gesture detectors.

Flutter has a weird quirk, where for some reason `panSlop` is by default defined as 2x `touchSlop`, which means that horizontal / vertical drag gestures (for which threshold is `touchSlop`) will always have priority to pan gesture (for which threshold is `panSlop`), even if the drag gesture is buried in the background.

This for example means that on iOS the selection dragging doesn't work (it is driven by pan gesture) if there is horizontal drag gesture below it (which we use to return to previous page).

We can get around this by using custom `DeviceGestureSettings` in top level `MediaQuery` for which `panSlop` equals `touchSlop`, but this is currently ignored by the `RawGestureDetector`s within SuperEditopr.

PR #908, opened 2023-01-05
Respect GestureSettings from MediaQuery

knopp:
Fixes https://github.com/superlistapp/super_editor/issues/907

### DIFF
--- a/super_editor/lib/src/default_editor/document_gestures_mouse.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_mouse.dart
@@ -641,6 +641,7 @@ Updating drag selection:
   Widget _buildGestureInput({
     required Widget child,
   }) {
+    final gestureSettings = MediaQuery.maybeOf(context)?.gestureSettings;
     return RawGestureDetector(
       behavior: HitTestBehavior.translucent,
       gestures: <Type, GestureRecognizerFactory>{
@@ -652,7 +653,8 @@ Updating drag selection:
               ..onDoubleTapDown = _onDoubleTapDown
               ..onDoubleTap = _onDoubleTap
               ..onTripleTapDown = _onTripleTapDown
-              ..onTripleTap = _onTripleTap;
+              ..onTripleTap = _onTripleTap
+              ..gestureSettings = gestureSettings;
           },
         ),
         PanGestureRecognizer: GestureRecognizerFactoryWithHandlers<PanGestureRecognizer>(
@@ -662,7 +664,8 @@ Updating drag selection:
               ..onStart = _onPanStart
               ..onUpdate = _onPanUpdate
               ..onEnd = _onPanEnd
-              ..onCancel = _onPanCancel;
+              ..onCancel = _onPanCancel
+              ..gestureSettings = gestureSettings;
           },
         ),
       },
--- a/super_editor/test/super_textfield/super_textfield_gestures_test.dart
+++ b/super_editor/test/super_textfield/super_textfield_gestures_test.dart
@@ -6,6 +6,7 @@ import 'package:super_editor/super_editor.dart';
 
 import '../test_tools.dart';
 import 'super_textfield_inspector.dart';
+import 'super_textfield_robot.dart';
 
 void main() {
   group('SuperTextField gestures', () {
@@ -273,8 +274,7 @@ void main() {
         );
 
         // Tap down and up so the field is focused.
-        await tester.tapAt(tester.getTopLeft(find.byType(SuperTextField)));
-        await tester.pumpAndSettle();
+        await tester.placeCaretInSuperTextField(0);
 
         expect(
           SuperTextFieldInspector.findSelection(),
--- a/super_editor/test/super_textfield/super_textfield_gestures_test.dart
+++ b/super_editor/test/super_textfield/super_textfield_gestures_test.dart
@@ -256,7 +256,8 @@ void main() {
                   // and panSlop
                   final data = MediaQuery.of(context).copyWith(
                     gestureSettings: const _GestureSettings(
-                      slop: 18,
+                      panSlop: 18,
+                      touchSlop: 18,
                     ),
                   );
                   return MediaQuery(
@@ -294,6 +295,50 @@ void main() {
         expect(
           SuperTextFieldInspector.findSelection(),
           const TextSelection.collapsed(offset: 1),
+        );
+
+        await tester.pumpWidget(
+          MaterialApp(
+            home: Scaffold(
+              body: GestureDetector(
+                onHorizontalDragStart: (d) {
+                  horizontalDragStartCalled = true;
+                },
+                child: Builder(builder: (context) {
+                  // Gesture settings that mimic flutter default where
+                  // panSlop = 2x touchSlop
+                  final data = MediaQuery.of(context).copyWith(
+                    gestureSettings: const _GestureSettings(
+                      touchSlop: 18,
+                      panSlop: 36,
+                    ),
+                  );
+                  return MediaQuery(
+                    data: data,
+                    child: SuperTextField(
+                      textController: AttributedTextEditingController(
+                        text: AttributedText(text: 'a b c'),
+                      ),
+                    ),
+                  );
+                }),
+              ),
+            ),
+          ),
+        );
+
+        final gesture2 = await tester.startGesture(tester.getTopLeft(find.byType(SuperTextField)));
+        addTearDown(() => gesture2.removePointer());
+        await gesture2.moveBy(const Offset(19, 0));
+        await gesture2.up();
+        await tester.pumpAndSettle();
+
+        // With default gesture settings the horizontal drag recognizer will
+        // win instead of the selection PanGestureRecognizer.
+        expect(horizontalDragStartCalled, isTrue);
+        expect(
+          SuperTextFieldInspector.findSelection(),
+          const TextSelection.collapsed(offset: 0),
         );
       });
 
@@ -385,9 +430,10 @@ Future<void> _pumpTestApp(
 // Custom gesture settings that ensure panSlop equal to touchSlop
 class _GestureSettings extends DeviceGestureSettings {
   const _GestureSettings({
-    required double slop,
-  }) : super(touchSlop: slop);
+    required double touchSlop,
+    required this.panSlop,
+  }) : super(touchSlop: touchSlop);
 
   @override
-  double? get panSlop => touchSlop;
+  final double panSlop;
 }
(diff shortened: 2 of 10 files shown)


super_editor/lib/src/default_editor/document_gestures_mouse.dart
```diff
@@ -641,6 +641,7 @@ Updating drag selection:
   Widget _buildGestureInput({
     required Widget child,
   }) {
+    final gestureSettings = MediaQuery.maybeOf(context)?.gestureSettings;
```
matthew-carroll, line 644: Should we cache this from `didChangeDependencies()`? In another project I'm working on, where we're solving a bunch of performance issues, we found that access `MediaQuery` can be surprisingly expensive...

matthew-carroll (commented):
Can you add tests that run gestures to demonstrate the different effect? For example, wherever you're running into the current gesture issues, can you test that without the `MediaQuery` setting one gesture is handled, but when you change the `MediaQuery` setting, a different gesture is handled?

super_editor/lib/src/default_editor/document_gestures_mouse.dart
knopp: I'd be curious to know more about the performance issues. Because if `MediaQuery` is expensive it means inherited widgets are expensive. And that'd be a problem since they're everywhere.

Last time I looked into that the query should have been cached and thus map operation, though there is also dependency tracking.

I'll look at the test.

matthew-carroll: Maybe it's a weird issue in the other project. We were seeing something like 4ms for some `MediaQuery` accesses.

knopp: I added test that will fail without custom gesture settings (triggering the background drag gesture instead of selection change).

knopp:
This got automatically closed by merging another PR, reopening.

super_editor/test/super_textfield/super_textfield_gestures_test.dart
```diff
@@ -7,6 +7,16 @@ import 'package:super_editor/super_editor.dart';
 import '../test_tools.dart';
 import 'super_textfield_inspector.dart';
 
+// Custom gesture settings that ensure panSlop equal to touchSlop
+class _GestureSettings extends DeviceGestureSettings {
```
matthew-carroll, line 11: Can you put this at the bottom of the file? Otherwise, readers will have to go through all the low-level details before they reach the actual tests.

super_editor/test/super_textfield/super_textfield_gestures_test.dart
```diff
@@ -240,6 +250,66 @@ void main() {
...
+                    ),
+                  );
+                }),
+              ),
+            ),
+          ),
+        );
+
+        // Tap down and up so the field is focused.
+        await tester.tapAt(tester.getTopLeft(find.byType(SuperTextField)));
```
matthew-carroll, line 287: We have actions defined on `WidgetTester` that will tap to place the caret at a specific text offset.

super_editor/test/super_textfield/super_textfield_gestures_test.dart
```diff
@@ -240,6 +250,66 @@ void main() {
...
+                  // and panSlop
+                  final data = MediaQuery.of(context).copyWith(
+                    gestureSettings: const _GestureSettings(
+                      slop: 18,
+                    ),
+                  );
+                  return MediaQuery(
+                    data: data,
+                    child: SuperTextField(
+                      textAlign: TextAlign.left,
```
matthew-carroll, line 274: I think this line is superfluous? If so, let's remove it so that readers don't think it's important.

super_editor/test/super_textfield/super_textfield_gestures_test.dart
```diff
@@ -240,6 +250,66 @@ void main() {
...
+                  );
+                }),
+              ),
+            ),
+          ),
+        );
+
+        // Tap down and up so the field is focused.
+        await tester.tapAt(tester.getTopLeft(find.byType(SuperTextField)));
+        await tester.pumpAndSettle();
```
matthew-carroll, line 288: When you switch to the SuperTextField methods in `WidgetTester`, you won't need to pump and settle. That's done for you.

super_editor/test/super_textfield/super_textfield_gestures_test.dart
```diff
@@ -240,6 +250,66 @@ void main() {
...
+          SuperTextFieldInspector.findSelection(),
+          const TextSelection.collapsed(offset: 0),
+        );
+
+        // The gesture should trigger the selection PanGestureRecognizer instead
+        // the HorizontalDragGestureRecognizer below.
+
+        final gesture = await tester.startGesture(tester.getTopLeft(find.byType(SuperTextField)));
+        addTearDown(() => gesture.removePointer());
+        await tester.pumpAndSettle();
```
matthew-carroll, line 300: Are you sure you want to be settling in each of these? Do you actually need to pump many frames? Typically, we'd like to account for every frame, so if `pump()` does the job, I would recommend using that.

super_editor/test/super_textfield/super_textfield_gestures_test.dart
```diff
@@ -240,6 +250,66 @@ void main() {
...
+        await gesture.moveBy(const Offset(19, 0));
+        await tester.pumpAndSettle();
+        await gesture.up();
+        await tester.pumpAndSettle();
+
+        expect(horizontalDragStartCalled, isFalse);
```
matthew-carroll, line 306: This expectation confirms that the drag start isn't called, when you don't want it to. But you should probably also check that the drag start *is* called, when you *do* want it. For example, if MediaQuery completely broke dragging, this test would still pass.

super_editor/test/super_textfield/super_textfield_gestures_test.dart
knopp: Which line? The MediaQuery specifies data with custom GestureSettings (with panSlop = touchSlop).

matthew-carroll: `textAlign: TextAlign.left` - you seem to be specifying a value which is already the default, no?

knopp: I see. I'll remove that.

super_editor/test/super_textfield/super_textfield_gestures_test.dart
knopp: Not sure what you mean. I just looked at other tests in the file and they all seem to do tapAt + pumpAndSettle.

super_editor/test/super_textfield/super_textfield_gestures_test.dart
knopp: All the other tests in file seem to be focusing the textfield like this though.

matthew-carroll: A quick inspection of a few of those tests show that we're verifying behaviors related to where and how we press. That's why they use direct gestures. In this case, you just want to place the caret. You're not testing the tap behavior.

Here's a test file that uses the SuperTextField test tooling: https://github.com/superlistapp/super_editor/blob/main/super_editor/test/super_textfield/super_textfield_ime_test.dart

In general, unless we're writing a test that's specific to an interaction, I'd like for tests to use the tooling that abstracts how a caret is places, or a selection is dragged.

super_editor/test/super_textfield/super_textfield_gestures_test.dart
matthew-carroll: Please see the response to your other question related to the caret placement tooling.

super_editor/test/super_textfield/super_textfield_gestures_test.dart
matthew-carroll: Based on your initial report, it sounded like the problem was that you expected to receive gesture callbacks that you never received. The fact that `MediaQuery` is involved, or that there's some particular implementation for `GestureDetector`, are implementation details. At the end of the day, you need to receive the expected gesture when you expect it. That's the part we want to lock down with a test.

This test currently checks that a given gesture doesn't trigger when slop is exceeded. But you haven't checked that the gesture does trigger when the slop isn't exceeded. Both of those details are equally critical to the expectations, no?

knopp: The gesture that is not received without the MediaQuery is the pan gesture that drives cursor movement. Since I don't have direct access to the gesture detector that does it, I test it indirectly by seeing that the cursor has moved (which does not happen without the MediaQuery override).

super_editor/test/super_textfield/super_textfield_gestures_test.dart
knopp: Thanks for the pointer. This makes sense. I replaced manual tap with the robot.

super_editor/test/super_textfield/super_textfield_gestures_test.dart
matthew-carroll: Here's the gesture callback:

```
onHorizontalDragStart: (d) {
                  horizontalDragStartCalled = true;
                }
```

Currently, you appear to check that the callback is never called.

I'm asking if you can now add a second gesture that causes it to be called. That callback is supposed to be reachable if you move less than the slop, right?

knopp: The callback should never be called. With overridden settings both the HorizontalDragGestureRecognizer and PanGestureRecognizer (that drives cursor movement) have same slop, which means recogniser that first handles the even wins. In this case the PanGesutreRecognizer is further up in the hit test path so it get first shot at handling the event. So as soon as the slop (which is same for both recognizers) is reached, pan gesture recognizer wins.

Without the media query override the background HorizontalDragGestureRecognizer would usually win (because of smaller slop), despite being further down in the hit test path.

matthew-carroll: Ok, so it's the changes to the `MediaQuery` slop value that produces different empirical results?

If so, can you pump again with a smaller `MediaQuery` slop that demonstrates the alternative execution path, with the other gesture trigger?

I'm just trying to add the other side of the coin to this test. Whatever form that takes. Because we're only verifying 1 out of 2 situations related to these gesture changes.

knopp: I can pump again with regular gesture settings (that define panSlop as 0.5 touchSlop) which should trigger the recognizer and not move cursor.

knopp: I updated the test. Is this what you meant?

matthew-carroll: Yeah, I think that looks good.

matthew-carroll (approved):
LGTM - Don't forget to cherry pick this commit from `main` to `stable` in a follow up PR

PR #876, opened 2022-12-05
Let apps open and close the keyboard on demand, and let selection exist when keyboard is closed (Resolves #875)

matthew-carroll:
This PR lets apps open and close the IME through a `SuperEditor` `DocumentComposer`. It also lets the document selection exist even after the keyboard has closed.

PR #910, opened 2023-01-08
Fixed SuperTextField selection glitch by never resending the previous TextEditingValue to the IME (Resolves #905)

### LINKED ISSUE #905: Selection glitch in SuperTextField when IME is used
The selection in `SuperTextField` jumps between old and current selection as shown in this recording:

https://user-images.githubusercontent.com/31278849/210235669-7d2d08b7-414b-4aed-b087-8c5e071b2013.mov

This is reproducible in demo app ("SuperTextField" -> "Interactive demo"). If `TextInputSource.keyboard` is used instead, then the issue is not reproducible.

matthew-carroll:
Fixed SuperTextField selection glitch by never resending the previous TextEditingValue to the IME (Resolves #905)


matthew-carroll:
This PR needs #876 to land first because we use a decorate introduced in that PR.

angelosilvestre (approved):
LGTM

PR #918, opened 2023-01-12
[SuperEditor][mac] Fix upstream selection using keyboard (Resolves #913)

### LINKED ISSUE #913: [SuperEditor] Shift+Up arrow to extend selection collapses selection
Steps to reproduce on `stable` brnach:
* Open Demo app
* On a paragraph with multiple lines, place the caret at one of the bottom lines
* Press Shift+Up arrow to extend the selection upwards

Expected: selection is extended accordingly
Actual:

https://user-images.githubusercontent.com/31278849/211541346-5f56d199-1ca1-4bdd-9b9f-fbe7dbfed10a.mov

angelosilvestre:
[SuperEditor][mac] Fix upstream selection using keyboard. Resolves #913 

Trying to expand the selection upstream using shift + up or shift + left arrow is causing the selection to be collapsed.

This was introduced in https://github.com/superlistapp/super_editor/pull/877 when non-text delta handling was implemented.

In macOS, after we send our text editing value, we always get back a non-text delta. When we convert a `DocumentSelection` to a `TextSelection` to be sent to IME, we are inverting the selection `base` and `extent`, if we have an upstream selection. We send it to the IME and we get back a non-text delta with a downstream selection. This causes the bug.

This PR changes the `DocumentImeSerializer` to always use the same `base`, `extent`, and `affinity` from the selection.

I tested in windows, mac, android and iOS (simulator) and everything seems to be working correctly.

### DIFF
--- a/super_editor/lib/src/default_editor/document_ime/document_serialization.dart
+++ b/super_editor/lib/src/default_editor/document_ime/document_serialization.dart
@@ -255,19 +255,15 @@ class DocumentImeSerializer {
   TextSelection documentToImeSelection(DocumentSelection docSelection) {
     editorImeLog.fine("Converting doc selection to ime selection: $docSelection");
     final selectionAffinity = _doc.getAffinityForSelection(docSelection);
-
-    final startDocPosition = selectionAffinity == TextAffinity.downstream ? docSelection.base : docSelection.extent;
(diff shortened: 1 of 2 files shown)


matthew-carroll (approved):
LGTM - please cherry pick to stable

PR #919, opened 2023-01-12
[SuperTextField] Fix selection glitch when using IME (Resolves #905)

### LINKED ISSUE #905: Selection glitch in SuperTextField when IME is used

angelosilvestre:
[SuperTextField] Fix selection glitch when using IME. Resolves #905

When using `SuperTextField` and typing a character, the selection jumps between the old and the new selected position. When using `SuperDesktopTextField` this isn't happening.

`SuperTextField` is always using the `defaultTextFieldKeyboardHandlers` by default. Because of this, typing is being handled by `insertCharacterWhenKeyIsPressed` instead of being handled using deltas.

This PR changes `SuperTextField` to use `defaultTextFieldImeKeyboardHandlers` when in IME mode, the same way `SuperDesktopTextField` does. With this change, the issue stopped.

### DIFF
--- a/super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart
+++ b/super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart
@@ -49,7 +49,7 @@ export 'styles.dart';
 ///  - [SuperAndroidTextField], which uses IME text input with Android-style handles
 ///  - [SuperIOSTextField], which uses IME text input with iOS-style handles
 class SuperTextField extends StatefulWidget {
-  const SuperTextField({
+  SuperTextField({
     Key? key,
     this.focusNode,
     this.configuration,
@@ -64,7 +64,7 @@ class SuperTextField extends StatefulWidget {
     this.minLines,
     this.maxLines = 1,
     this.lineHeight,
-    this.inputSource,
+    TextInputSource? inputSource,
     List<TextFieldKeyboardHandler>? keyboardHandlers,
     this.padding,
     this.textInputAction,
@@ -150,7 +150,7 @@ class SuperTextField extends StatefulWidget {
   /// The [SuperTextField] input source, e.g., keyboard or Input Method Engine.
   ///
   /// Only used on desktop. On mobile platforms, only [TextInputSource.ime] is available.
-  final TextInputSource? inputSource;
+  final TextInputSource inputSource;
 
   /// Priority list of handlers that process all physical keyboard
   /// key presses, for text input, deletion, caret movement, etc.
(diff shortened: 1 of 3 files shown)


angelosilvestre:
This caused some tests to fail. I will need to investigate.

angelosilvestre:
The `SuperTextField` initialization had to be modified.

super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart
```diff
@@ -334,6 +319,19 @@ class SuperTextFieldState extends State<SuperTextField> {
   }
 }
 
+TextInputSource platformDefaultInputSource() {
```
matthew-carroll, line 322: Is there a reason that this was moved? I don't see any functional differences.

super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart
```diff
@@ -64,10 +64,15 @@ class SuperTextField extends StatefulWidget {
...
     this.maxLines = 1,
     this.lineHeight,
-    this.inputSource,
-    this.keyboardHandlers = defaultTextFieldKeyboardHandlers,
+    TextInputSource? inputSource,
+    List<TextFieldKeyboardHandler>? keyboardHandlers,
     this.padding,
-  }) : super(key: key);
+  })  : inputSource = inputSource ?? platformDefaultInputSource(),
+        keyboardHandlers = keyboardHandlers ??
```
matthew-carroll, line 71: Can you do this in a local method, like how we dealt with the input source before?

super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart
angelosilvestre: We need this method to be outside of the state class, so we can use it in the `SuperTextField` initializer

super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart
angelosilvestre: We can't access local methods in an initializer. Also, the methods are in the state class...

super_editor/lib/src/infrastructure/super_textfield/super_textfield.dart
matthew-carroll: I guess what I'm confused about is why we need to access this from the widget initializer. Isn't the widget initializer just making a choice about which input source and which handlers to use? Why can't we make that decision in the `State` object?

angelosilvestre: It seems we can make it a lot simpler. As `SuperDesktopTextField` already chooses its default handlers, we could just make the property nullable in `SuperTextField`.

matthew-carroll (approved):
LGTM - Please cherry pick to stable

matthew-carroll:
Oops. @angelosilvestre looks like there are some conflicts, somehow

angelosilvestre:
@matthew-carroll Conflicts solved

PR #920, opened 2023-01-12
Cherry-pick "Respect GestureSettings from MediaQuery (Resolves #907) (#908)" to stable

### LINKED ISSUE #907: RawGestureDetector should respect MediaQuery gestureSettings

knopp:


matthew-carroll (approved):
LGTM

Issue #927, opened 2023-01-14
[SuperEditor] - Move mobile gesture recognition under document, like desktop


matthew-carroll:
On desktop, we recognize gestures from an area that sits under the document content. This allows inner component widgets, such as checkboxes, to receive and consume touches before the document gesture take over.

On mobile, we're still recognizing gestures above the document content, which means that tapping a checkbox activates the checkbox, and moves the caret. 

The mobile gesture system wasn't updated with the desktop gesture system, because the mobile layout is currently using a `SingleChildScrollView` to scroll the document. That widget consumes gestures and forces a particular structure on us.

Refactor mobile to implement our own scrolling, and then move all gesture handling beneath the document content.

Also, when that's complete, reactivate and refactor the Task component tests that are currently skipped for mobile.

robertodoering:
Hey @matthew-carroll,
is this related to the issue where a user goes into edit mode when they tap on the document while the scrollable is still scrolling (instead of the tap stopping the scrolling)?

matthew-carroll:
@robertodoering that might the be the cause but it's probably a good idea for you to file a new ticket for that particular problem so that we can track it, explicitly.

matthew-carroll:
@angelosilvestre can you verify whether or not the above scrolling issue is due to the location of the gesture system vs something else?

angelosilvestre:
@matthew-carroll It seems to be related to the location of the gesture system. The `onTapUp` of the gesture system is being called.

PR #928, opened 2023-01-14
Moved tasks into Super Editor and fixed bug that prevented creating new tasks after texisting tasks (Resolves #923)

### LINKED ISSUE #923: [SuperEditor] - Task component doesn't create new task on ENTER
We've had a regression where a task item in the example app doesn't do anything when `enter` is pressed at the end of the line. 

Return this behavior to its previous state.

Also, add a test for this under a new file location: `test/super_editor/components/task_test.dart`

matthew-carroll:
Moved tasks into Super Editor and fixed bug that prevented creating new tasks after texisting tasks (Resolves #923)

At some point we lost the ability to press ENTER at the end of a task in the example app and create a new task. This PR fixes that. I'm not sure how it was ever working. We have code in the ENTER handler that rejects any text node that isn't a `ParagraphNode`. I added a new key handler that specifically handles this case.

This PR also moves tasks into Super Editor as a standard content type. I also added tests to ensure that `TaskNode`s and `TaskComponent`s do the right thing within a `SuperEditor`.

I tried adding tests for tapping the checkbox. Those tests work on desktop, but not on mobile. At least part of the problem on mobile is that the mobile gesture recognizer still sits on top of the document, instead of beneath the document, like desktop. I filed https://github.com/superlistapp/super_editor/issues/927 to solve that problem. When it's solved, the mobile tests that are skipped in this PR should be combined with the desktop tests.

I changed how the demo app deals with brightness. Instead of making the setting a local `bool`, I switched to `Brightness` and applied it with a `Theme` widget. This way, any given component can lookup the current brightness and base its appearance on the right value.

I also discovered that the `ProxyDocumentComponent` implementation didn't deal with the possibility that the inner component might not take up all the space of the proxy component. I added offset calculations to deal with that.

### DIFF
--- a/super_editor/example/lib/demos/example_editor/example_editor.dart
+++ b/super_editor/example/lib/demos/example_editor/example_editor.dart
@@ -1,4 +1,3 @@
-import 'package:example/demos/example_editor/_task.dart';
 import 'package:example/logging.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
@@ -30,7 +29,7 @@ class _ExampleEditorState extends State<ExampleEditor> {
 
   final _darkBackground = const Color(0xFF222222);
   final _lightBackground = Colors.white;
-  bool _isLight = true;
+  Brightness _brightness = Brightness.light;
 
   OverlayEntry? _textFormatBarOverlayEntry;
   final _textSelectionAnchor = ValueNotifier<Offset?>(null);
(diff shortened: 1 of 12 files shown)


super_editor/example/lib/demos/example_editor/example_editor.dart
```diff
@@ -353,23 +360,23 @@ class _ExampleEditorState extends State<ExampleEditor> {
...
-            if (!_isLight) ..._darkModeStyles,
+            if (!isLight) ..._darkModeStyles,
             taskStyles,
           ],
         ),
         componentBuilders: [
-          ...defaultComponentBuilders,
           TaskComponentBuilder(_docEditor),
+          ...defaultComponentBuilders,
         ],
```
angelosilvestre, line 380: As tasks were moved into Super Editor, shouldn't `TaskComponentBuilder` be moved into `defaultComponentBuilders` ?

angelosilvestre (commented):
LGTM with a single comment.

super_editor/example/lib/demos/example_editor/example_editor.dart
matthew-carroll: It should, but I haven't found a way to create it without passing it the `DocumentEditor`, which isn't available, statically...

PR #931, opened 2023-01-14
Added newline key handler to IME SuperTextField to fix example app URL field issue (Resolves #925)

### LINKED ISSUE #925: [SuperEditor] - Example app URL insertion doesn't work with ENTER key
Repro: Select some text in the example editor. Open the URL textfield. Enter some text. Press ENTER.

Expected: Toolbar closes and URL is applied to text.

Actual: Toolbar closes, no URL is applied, and the editor splits the node wherever the caret was placed.

matthew-carroll:
Added newline key handler to IME SuperTextField to fix example app URL field issue (Resolves #925)


angelosilvestre (commented):
LGTM

PR #937, opened 2023-01-15
Don't clear editor selection when IME closes and editor still has non-primary focus. Also move focus selection policy from a mixin to a policy widget (Resolves #933, Resolves #935)

### LINKED ISSUE #933: [SuperEditor] - Limit focus and selection policies to "primary focus"
We recently refactored IME behavior by breaking down policies into small, targeted widgets.

The resulting policies include watching for changes to focus. However, the policies seem to only look at `FocusNode.hasFocus` without checking `FocusNode.hasPrimaryFocus`. This results in an open IME connection when the editor doesn't have primary focus, which shouldn't be the case.

For example, a toolbar might appear, and include a text field. When the user is typing into that text field, the editor should not have an open IME connection. In fact, the editor's IME connection will probably be stolen by the text field. But that behavior is non-deterministic and `SuperEditor` should instead explicitly close the connection when it loses primary focus. `SuperEditor` should also re-open the IME connection when it regains primary focus.

### LINKED ISSUE #935: [SuperEditor] - Move selection restoration behavior into a policy widget
We've introduced a couple "policy" widgets in `SuperEditor` for IME related behavior. Each such widget implements a very narrow set of behaviors related to focus, selection, and IME connections.

super_editor already has an implementation to restore a previous selection when the editor is refocused. I think that implementation is in a mixin. Try moving that behavior into a small policy widget, and then compose that policy widget into the `SuperEditor` tree.

matthew-carroll:
Don't clear editor selection when IME closes and editor still has non-primary focus. Also move focus selection policy from a mixin to a policy widget (Resolves #933, Resolves #935)

I'm working on debugging some issues in Superlist. This PR will hopefully take a step towards solving one of them.

`SuperEditor` was being too aggressive when clearing the editor's selection after closing its IME connection. When a popup toolbar is used to enter a URL for the current selection, that behavior requires that the editor close its IME, so the text field can open the IME. In this situation, we don't want to clear the editor's selection, because that's where the URL is supposed to apply. The difference between this situation and others is that in this situation the editor still has focus, but not primary focus. On the other hand, when the editor has primary focus, and the IME connection closes, then we want to clear the selection. This is the issue described in #933 

`SuperEditor` already had a mixin that would put the caret at the end of the document when `SuperEditor` gains focus. I moved that behavior into another policy widget, to match the pattern of the policy widgets we created recently. That solves #935 

There's a lot of complexity showing up at the intersection of focus, selection, and IME, so we may find additional issues to fix.


matthew-carroll:
@Jethro87 this PR will effect you because it re-organizes some of the policy properties that determine what `SuperEditor` does automatically for focus, selection, and IME.

Jethro87:
@matthew-carroll Thanks for the heads up.

Jethro87:
@matthew-carroll Thanks for the heads up.

super_editor/lib/src/default_editor/document_focus_and_selection_policies.dart
```diff
@@ -0,0 +1,140 @@
...
+    super.didUpdateWidget(oldWidget);
+
+    if (widget.focusNode != oldWidget.focusNode) {
+      oldWidget.focusNode.removeListener(_onFocusChange);
+      widget.focusNode.addListener(_onFocusChange);
+      _onFocusChange();
+    }
+
+    if (widget.selection != oldWidget.selection) {
+      oldWidget.selection.removeListener(_onFocusChange);
```
angelosilvestre, line 85: This should be `_onSelectionChange` instead of `_onFocusChange`.

angelosilvestre (commented):
LGTM with a single comment.

super_editor/lib/src/default_editor/document_focus_and_selection_policies.dart
matthew-carroll: Good find

PR #940, opened 2023-01-15
SuperTextField detach from IME in dispose() (Resolves #934)

### LINKED ISSUE #934: [SuperTextField] - on mobile it doesn't detach from IME when disposed with focus
`_textEditingController.detachFromIme();` is only called on focus node change. However we swap TextField to another widget while the focus node is still focused, so this is not called.

`_textEditingController.detachFromIme();` should unconditionally be called from `dispose`.

Same thing for goes from android SuperTextField.

Without this the on-screen keyboard keeps being shown after the text field is gone.

matthew-carroll:
SuperTextField detach from IME in dispose() (Resolves #934)

When a `SuperTextField` is disposed, it should close its IME connection so that the software keyboard doesn't stay up, and also to avoid conflicting with other widgets that want an IME connection.


angelosilvestre (commented):
LGTM

PR #950, opened 2023-01-16
Fixed fast-typing IME errors by re-combining responsibilities for both applying deltas and serializing the document for IME (Resolves #914)

### LINKED ISSUE #914: [SuperEditor] Typing fast can place the caret at the wrong offset
Steps to reproduce on `stable`:
* Open Demo app
* On an empty paragraph, start typing fast

Expected: characters are inserted and caret continues at the end of the paragraph
Actual:

https://user-images.githubusercontent.com/31278849/211542114-6d9c27fa-592b-4875-a2c1-480a76ffb6cf.mov

matthew-carroll:
Fixed fast-typing IME errors by re-combining responsibilities for both applying deltas and serializing the document for IME (Resolves #914)

In the recent refactor, I split up the responsibility of applying IME deltas to a document, and the responsibility of serializing a document and sending it to the IME. I made this split because there can many factors that alter a document: IME deltas, key presses, other users over the network, reactionary commands, etc. Therefore, I wanted to avoid a special relationship between the behavior that edits a document based on IME deltas, and behavior that updates the IME with the latest document structure.

Unfortunately, this split wasn't working out. Most of the problem was related to non-atomic changes with documents and selection. We should try this split again after atomic commands are introduced.

Here is a link to the relevant code with split responsibilities, which is undone by this PR:
https://github.com/superlistapp/super_editor/blob/19dd8cce862805668c75c9a659e92db23ec5952e/super_editor/lib/src/default_editor/document_ime/document_ime_communication.dart

### DIFF
--- a/super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
+++ b/super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
@@ -26,7 +26,7 @@ class TextDeltasDocumentEditor {
   void applyDeltas(List<TextEditingDelta> textEditingDeltas) {
     editorImeLog.info("Applying ${textEditingDeltas.length} IME deltas to document");
 
-    editorImeLog.fine("Serializing document to perform IME operation");
+    editorImeLog.fine("Serializing document to perform IME operations");
     final serializedDocBeforeDelta = DocumentImeSerializer(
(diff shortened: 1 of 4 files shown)


angelosilvestre (commented):
LGTM

PR #961, opened 2023-01-17
Don't try to send doc to IME when not attached, and ensure _isSendingToIme is always reset (Resolves #960)

### LINKED ISSUE #960: [SuperEditor] - Characters appear in wrong place after closing toolbar and typing
Repro: Run the example app. Select some text so that the popover toolbar appears. Place the caret somewhere else, so that the toolbar disappears. Press a character key.

Expected: The desired character is inserted at the caret.

Actual: Selection moves back to the beginning of the paragraph and the character is inserted there.

matthew-carroll:
Don't try to send doc to IME when not attached, and ensure _isSendingToIme is always reset (Resolves #960)


miguelcmedeiros (approved):
LGTM

PR #976, opened 2023-01-20
[SuperTextField] Implement missing methods in ImeAttributedTextEditingController

miguelcmedeiros:
Resolves #975


super_editor/lib/src/infrastructure/super_textfield/input_method_engine/_ime_text_editing_controller.dart
```diff
@@ -536,6 +543,57 @@ class ImeAttributedTextEditingController extends AttributedTextEditingController
...
+    _realController.insertCharacter(character);
+  }
+
+  @override
+  void moveCaretHorizontally({
+    required ProseTextLayout textLayout,
+    required bool expandSelection,
+    required bool moveLeft,
+    required MovementModifier? movementModifier,
+  }) {}
```
matthew-carroll, line 573: Is this missing an implementation?

matthew-carroll (commented):
Does this solve the token issue, or you're bringing this up to date because you know it's a problem?

PR #983, opened 2023-01-20
[SuperTextField][macOS] Fix compound characters insertion (Resolves #981)

### LINKED ISSUE #981: [SuperTextField] - Compound characters don't work in SuperTextField
Compound characters aren't working in `SuperTextField` because we aren't honoring the IME composing region.

Add appropriate composing region support to `SuperTextField`, including any tests that might be relevant, to get compound characters working.

angelosilvestre:
[SuperTextField][macOS] Fix compound characters insertion. Resolves #981

Compound characters, like `ã` and `á` aren't working in `SuperTextField` on mac. The reason is that we aren't honoring the composing region we get on the insertion deltas.

This PR changes `SuperTextField` to honor the composing region we get from the insertion delta.


matthew-carroll (approved):
LGTM

PR #989, opened 2023-01-21
[SuperEditor] Fix text attributions after tapping at an attributed range (Resolves #982)

### LINKED ISSUE #982: [SuperEditor] Text attribution is not always inherited
Steps to reproduce:
* Run the demo app
* Add a link to a piece of text
* Place the caret in the middle of that same piece of text
* Start typing

Expected: the new characters inherit the link attribution 
Actual:

https://user-images.githubusercontent.com/31278849/213818877-e1f1cc46-baa6-4894-a103-c8a3adeda4b9.mov

Similar when typing in the middle of a piece of text with other attributions:

https://user-images.githubusercontent.com/31278849/213819466-c84b6618-a77a-4ab4-b35a-78c42ba7d38a.mov

angelosilvestre:
[SuperEditor] Fix text attributions after tapping at an attributed range. Resolves #982

Tapping at an offset that contains attributions and typing a character isn't applying the attributions to the inserted text. When selecting using the keyboard this is working.

This was introduced after the selection reason PR. We aren't calling `notifyListeners` if the selection is changed using the `selectionNotifier`, which is how the selection is changed when tapping the editor.

This PR changes the `DocumentComposer` to call `notifyListeners` when the selection is changed using the `selectionNotifier`.

Link attributions are currently being ignored and there are tests ensuring that link attributions don't expand when typing at the beginning or end of a link.

This PR changes the `DocumentComposer` to allow expanding link attributions when typing at the middle of a link.

### DIFF
--- a/super_editor/lib/src/default_editor/super_editor.dart
+++ b/super_editor/lib/src/default_editor/super_editor.dart
@@ -429,13 +429,12 @@ class SuperEditorState extends State<SuperEditor> {
 
     final textPosition = _composer.selection!.extent.nodePosition as TextPosition;
 
-    int currentAttributionsOffset = -1;
+    if (textPosition.offset == 0 && (node.text.text.isEmpty)) {
+      return;
+    }
 
+    late int currentAttributionsOffset;
     if (textPosition.offset == 0) {
-      if (node.text.text.isEmpty) {
-        return;
-      }
-
       // Inserted text at the very beginning of a text blob assumes the
       // attributions immediately following it.
       currentAttributionsOffset = textPosition.offset + 1;
--- a/super_editor/lib/src/default_editor/super_editor.dart
+++ b/super_editor/lib/src/default_editor/super_editor.dart
@@ -429,17 +429,18 @@ class SuperEditorState extends State<SuperEditor> {
 
     final textPosition = _composer.selection!.extent.nodePosition as TextPosition;
 
-    if (textPosition.offset == 0 && (node.text.text.isEmpty)) {
+    if (textPosition.offset == 0 && node.text.text.isEmpty) {
       return;
     }
 
     late int currentAttributionsOffset;
     if (textPosition.offset == 0) {
-      // Inserted text at the very beginning of a text blob assumes the
-      // attributions immediately following it.
+      // The inserted text is at the very beginning of the text blob. Therefore, we should apply the
+      // same attributions to the inserted text, as the text that immediately follows the inserted text.
       currentAttributionsOffset = textPosition.offset + 1;
     } else {
-      // Inserted text assumes the attributions immediately preceding it.
+      // The inserted text is NOT at the very beginning of the text blob. Therefore, we should apply the
+      // same attributions to the inserted text, as the text that immediately precedes the inserted text.
       currentAttributionsOffset = textPosition.offset - 1;
     }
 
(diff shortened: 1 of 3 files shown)


super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -428,29 +429,42 @@ class SuperEditorState extends State<SuperEditor> {
 
     final textPosition = _composer.selection!.extent.nodePosition as TextPosition;
 
+    int currentAttributionsOffset = -1;
```
matthew-carroll, line 432: Let's re-order some of these lines to make it easier to read.

```dart
if (textPosition.offset == 0 && node.text.text.isEmpty) {
  return;
}

late int currentAttributionOffset;
if (textPosition.offset == 0) {
  // same stuff
} else {
  // same stuff
}
```

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -428,29 +429,42 @@ class SuperEditorState extends State<SuperEditor> {
...
       // Inserted text at the very beginning of a text blob assumes the
-      // attributions immediately following it (except links).
-      // TODO: attribution expansion policy should probably be configurable
-      final allStyles = node.text
-          .getAllAttributionsAt(textPosition.offset + 1)
-          .where((attribution) => attribution is! LinkAttribution)
-          .toSet();
-      _composer.preferences.addStyles(allStyles);
+      // attributions immediately following it.
+      currentAttributionsOffset = textPosition.offset + 1;
```
matthew-carroll, line 441: It looks like the lines that were removed from these if-branches were just moved down below into multiple lines. Was there any reason for that change? At first glance, it looks like we introduced a variable and spread out the lines without changing the behavior...

angelosilvestre: The only difference in the if-else branches was the offset we used to query the attributions, so I thought it would be simpler this way. I can revert it back if needed...

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -428,29 +429,41 @@ class SuperEditorState extends State<SuperEditor> {
 
     final textPosition = _composer.selection!.extent.nodePosition as TextPosition;
 
-    if (textPosition.offset == 0) {
-      if (node.text.text.isEmpty) {
-        return;
-      }
+    if (textPosition.offset == 0 && (node.text.text.isEmpty)) {
```
matthew-carroll, line 432: No need for the extra parentheses after the `&&`

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -428,29 +429,41 @@ class SuperEditorState extends State<SuperEditor> {
...
+      return;
+    }
 
+    late int currentAttributionsOffset;
+    if (textPosition.offset == 0) {
       // Inserted text at the very beginning of a text blob assumes the
```
matthew-carroll, line 438: Let's adjust this comment here and in the `else` block. The wording is very strange.

```
// The inserted text is at the very beginning of the text blob. Therefore, we should apply the
// same attributions to the inserted text, as the text that immediately follows the inserted text.
```

```
// The inserted text is NOT at the very beginning of the text blob. Therefore, we should apply the
// same attributions to the inserted text, as the text that immediately precedes the inserted text.
```

matthew-carroll (approved):
LGTM - with a couple final comments

angelosilvestre:
@matthew-carroll Updated.

PR #993, opened 2023-01-24
[SuperTextField] Add ability to update keyboard appearance (Resolves #991)

### LINKED ISSUE #991: [SuperTextField] Update keyboard appearance without reinstantiating ImeAttributedTextEditingController
Currently if an app wants to configure the keyboard appearance to match current theme's brightness (light or dark), it could look like this:
```dart
  ImeAttributedTextEditingController? _textController;
  Brightness? _keyboardBrightness;

  @override

    final newBrightness = SuperlistTheme.of(context).brightness;
    if (_textController == null || newBrightness != _keyboardBrightness) {
      _keyboardBrightness = newBrightness;
      _textController = ImeAttributedTextEditingController(
        controller: widget.controller,
        keyboardAppearance: _keyboardBrightness!,
        disposeClientController: false,
      );
    }
  }

  @override
  void dispose() {
    _textController?.dispose();
    super.dispose();
  }
```

In the code above the developer needs to be careful to not mess up the lifecycle of the `_textController` or the inner controller (`widget.controller`), and it could become more complicated once listeners get added to either.

The goal of this ticket is to have a way to decouple the update of keyboard appearance from the instantiation of the IME controller. Not sure if it would be the correct change, but if `updateTextInputConfiguration` method is extended with `bool keyboardAppearance` field, then the code above could become:
```dart
  late ImeAttributedTextEditingController _textController;
  Brightness? _keyboardBrightness;

  @override
  void initState() {
    super.initState();

    _textController = ImeAttributedTextEditingController(
      controller: widget.controller,
      disposeClientController: false,
    );
  }
  
  @override

    final newBrightness = SuperlistTheme.of(context).brightness;
    if (newBrightness != _keyboardBrightness) {
      _keyboardBrightness = newBrightness;
      _textController.updateTextInputConfiguration(
        keyboardAppearance: newBrightness,
      )
    }
  }

  @override
  void dispose() {
    _textController.dispose();
    super.dispose();
  }
```

angelosilvestre:
[SuperTextField] Add ability to update keyboard appearance. Resolves #991

Currently, it isn't possible to update the keyboard appearance after the `ImeAttributedTextEditingController` creation.

This PR adds the ability of changing the `keyboardAppearance` in the `updateTextInputConfiguration` method. 

We could also change the  `updateTextInputConfiguration` parameters to be nullable and use the previous values for the parameters that were not passed.

This PR also introduces a proposed class to facilitate intercepting platform messages.

### DIFF
--- a/super_editor/lib/src/infrastructure/super_textfield/input_method_engine/_ime_text_editing_controller.dart
+++ b/super_editor/lib/src/infrastructure/super_textfield/input_method_engine/_ime_text_editing_controller.dart
@@ -56,7 +56,7 @@ class ImeAttributedTextEditingController extends AttributedTextEditingController
   /// The appearance of the software keyboard.
   ///
   /// Only used for iOS devices.
-  final Brightness keyboardAppearance;
+  Brightness keyboardAppearance;
 
   final AttributedTextEditingController _realController;
 
@@ -122,11 +122,14 @@ class ImeAttributedTextEditingController extends AttributedTextEditingController
     bool enableSuggestions = true,
     TextInputAction textInputAction = TextInputAction.done,
     TextInputType textInputType = TextInputType.text,
+    Brightness keyboardAppearance = Brightness.light,
   }) {
     if (!isAttachedToIme) {
       // We're not attached to the IME, so there is nothing to update.
       return;
     }
+
+    this.keyboardAppearance = keyboardAppearance;
 
     // Close the current connection.
     _inputConnection?.close();
--- a/super_editor/lib/src/infrastructure/super_textfield/input_method_engine/_ime_text_editing_controller.dart
+++ b/super_editor/lib/src/infrastructure/super_textfield/input_method_engine/_ime_text_editing_controller.dart
@@ -35,12 +35,13 @@ class ImeAttributedTextEditingController extends AttributedTextEditingController
     AttributedTextEditingController? controller,
     bool disposeClientController = true,
     void Function(RawFloatingCursorPoint)? onIOSFloatingCursorChange,
-    this.keyboardAppearance = Brightness.light,
+    Brightness keyboardAppearance = Brightness.light,
     TextInputConnectionFactory? inputConnectionFactory,
   })  : _realController = controller ?? AttributedTextEditingController(),
         _disposeClientController = disposeClientController,
         _inputConnectionFactory = inputConnectionFactory,
-        _onIOSFloatingCursorChange = onIOSFloatingCursorChange {
+        _onIOSFloatingCursorChange = onIOSFloatingCursorChange,
+        _keyboardAppearance = keyboardAppearance {
     _realController.addListener(_onInnerControllerChange);
   }
 
@@ -56,7 +57,8 @@ class ImeAttributedTextEditingController extends AttributedTextEditingController
   /// The appearance of the software keyboard.
   ///
   /// Only used for iOS devices.
-  Brightness keyboardAppearance;
+  Brightness get keyboardAppearance => _keyboardAppearance;
+  Brightness _keyboardAppearance;
 
   final AttributedTextEditingController _realController;
 
@@ -129,7 +131,7 @@ class ImeAttributedTextEditingController extends AttributedTextEditingController
       return;
     }
 
-    this.keyboardAppearance = keyboardAppearance;
+    _keyboardAppearance = keyboardAppearance;
 
     // Close the current connection.
     _inputConnection?.close();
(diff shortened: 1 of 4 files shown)


super_editor/test/super_editor/document_test_tools.dart
```diff
@@ -611,3 +613,45 @@ class FakeImageComponentBuilder implements ComponentBuilder {
     );
   }
 }
+
+/// Extension on [WidgetTester] to easily intercept platform messages.
+extension TestMessageInterceptor on WidgetTester {
```
matthew-carroll, line 618: It doesn't look like this extension is related to documents, but its in `document_test_tools.dart`.

super_editor/test/super_textfield/super_textfield_test.dart
```diff
@@ -200,6 +201,48 @@ void main() {
...
+            keyboardAppearance = params['keyboardAppearance'];
+            return null;
+          },
+        );
+
+        // Change the keyboard appearance from light to dark.
+        controller.updateTextInputConfiguration(
+          keyboardAppearance: Brightness.dark,
+        );
+        await tester.pumpAndSettle();
```
matthew-carroll, line 241: Does this need to "settle", or would one pump be enough?

super_editor/test/super_textfield/super_textfield_test.dart
```diff
@@ -200,6 +201,48 @@ void main() {
...
+
+        // Ensure the given keyboardAppearance was applied.
+        expect(keyboardAppearance, 'Brightness.dark');
```
matthew-carroll, line 244: Should you verify the original brightness that's sent? For example, if we had a bug and "dark" was always sent, wouldn't this test still pass?

super_editor/test/super_editor/document_test_tools.dart
angelosilvestre: Moved into `test_tools.dart`

super_editor/test/super_textfield/super_textfield_test.dart
angelosilvestre: It seems that `pump` is enough. Updated

super_editor/test/super_textfield/super_textfield_test.dart
angelosilvestre: Makes sense. Updated.

super_editor/lib/src/infrastructure/super_textfield/input_method_engine/_ime_text_editing_controller.dart
```diff
@@ -56,7 +56,7 @@ class ImeAttributedTextEditingController extends AttributedTextEditingController
   /// The appearance of the software keyboard.
   ///
   /// Only used for iOS devices.
-  final Brightness keyboardAppearance;
+  Brightness keyboardAppearance;
```
matthew-carroll, line 59: Should this be a private property? Or maybe a private property with a getter?

Right now, clients can change this directly, but it won't update the IME, which probably isn't the client's intention...

super_editor/test/super_editor/document_test_tools.dart
matthew-carroll: It looks like this `TestMessageInterceptor` is still in the wrong file?

super_editor/lib/src/infrastructure/super_textfield/input_method_engine/_ime_text_editing_controller.dart
angelosilvestre: Updated.

super_editor/test/super_editor/document_test_tools.dart
```diff
@@ -1,13 +1,16 @@
 import 'dart:math';
 import 'dart:ui' as ui;
 
+import 'package:flutter/foundation.dart';
```
matthew-carroll, line 4: Are any of these new imports needed?

super_editor/test/super_textfield/super_textfield_test.dart
```diff
@@ -5,6 +5,7 @@ import 'package:flutter_test/flutter_test.dart';
 import 'package:super_editor/super_editor.dart';
 import 'package:super_text_layout/super_text_layout.dart';
 
+import '../super_editor/document_test_tools.dart';
```
matthew-carroll, line 8: I don't think we should be importing document tools in textfield tests...

super_editor/test/super_editor/document_test_tools.dart
angelosilvestre: Unused imports removed.

super_editor/test/super_textfield/super_textfield_test.dart
angelosilvestre: Unused imports removed.

matthew-carroll (approved):
LGTM

PR #995, opened 2023-01-24
[super_editor_markdown] Add justify alignment for markdown serialization (Resolves #879)

### LINKED ISSUE #879: [SuperEditorMarkdown] The markdown with MarkdownSyntax.superEditor mode should supports the justify alignment
The `super_editor` supports all the alignments but I'm curious why you don't add the `justify` alignment when adding others to the `super_editor_markdown`, is there any issue or something that I'm missing about it?

If it's just a missing feature, also for the header alignment issue #865, I would like to create PRs for it.

https://github.com/superlistapp/super_editor/blob/6c95044e7f83a43fc83d51ab0db313562b77ce6b/super_editor_markdown/lib/src/document_to_markdown_serializer.dart#L187-L198

https://github.com/superlistapp/super_editor/blob/6c95044e7f83a43fc83d51ab0db313562b77ce6b/super_editor_markdown/lib/src/markdown_to_document_parsing.dart#L499-L515

angelosilvestre:
[super_editor_markdown] Add justify alignment for markdown serialization. Resolves #879

Currently, SuperEditor supports left, right and center alignment when using `MarkdownSyntax.superEditor`.

This PR adds the `justify` alignment to super_editor_markdown. 

As there isn't a spec for a justify alignment token, this PR introduces the  `----` token.

### DIFF
--- a/super_editor_markdown/lib/src/document_to_markdown_serializer.dart
+++ b/super_editor_markdown/lib/src/document_to_markdown_serializer.dart
@@ -192,6 +192,8 @@ String? _convertAlignmentToMarkdown(String alignment) {
       return ':---:';
     case 'right':
       return '---:';
+    case 'justify':
+      return '----';
     default:
       return null;
   }
--- a/super_editor_markdown/lib/src/document_to_markdown_serializer.dart
+++ b/super_editor_markdown/lib/src/document_to_markdown_serializer.dart
@@ -193,7 +193,7 @@ String? _convertAlignmentToMarkdown(String alignment) {
     case 'right':
       return '---:';
     case 'justify':
-      return '----';
+      return '-::-';
     default:
       return null;
   }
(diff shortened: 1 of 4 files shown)


angelosilvestre:
@matthew-carroll I just saw that using `----` as the alignment token might be confusing, because the horizontal rule is represented with a similar token: `---`

Do you have any suggestions for another token? Maybe we could use `[---]` or `+---+`.

vursin:
@angelosilvestre I'm currently using `--:--` in my modified branch for the justify, I think using this token gives a more unified feel when compared to the rest of the tokens.

super_editor_markdown/lib/src/document_to_markdown_serializer.dart
```diff
@@ -192,6 +192,8 @@ String? _convertAlignmentToMarkdown(String alignment) {
       return ':---:';
     case 'right':
       return '---:';
+    case 'justify':
+      return '----';
```
matthew-carroll, line 196: Let's use "-::-". It doesn't make any sense, but at least that token won't be confused with HRs, and it's also the same length as the others. If you think there's any reason not to use that suggestion,  please let me knwo.

matthew-carroll (approved):
LGTM

PR #1008, opened 2023-01-27
[SuperEditor] Update mobile toolbar design (Resolves #963)

### LINKED ISSUE #963: [SuperEditor] Update design of mobile floating toolbar
EDIT (Matt): in addition to toolbar visuals below, also ensure the toolbars don't get too close to the screen.

### Current iOS
Simulator Screen Shot - iPhone 14 - 2023-01-17 at 09 04 52

### Desired iOS
IMG_1150

### Current Android
IMG_1149

### Desired Android
Screenshot_20230117-091546

### Extra
Additionally, the iOS toolbar shows additional options and scrolls between them, as per the video below:

https://user-images.githubusercontent.com/6467808/212968373-299f2b5d-87b5-4868-bb9c-c77908628ce8.MP4

While the above functionality is beyond a simple design refresh, it would be nice to eventually have it in super editor.

angelosilvestre:
[SuperEditor] Update mobile toolbar design. Resolves #963 

This PR updates the design of the mobile toolbars. For the iOS style toolbar I used the overlord package. https://github.com/Flutter-Bounty-Hunters/overlord/pull/13 needs to land first.

* I had to make a breaking change, in the toolbar builder we need the focal point to be able to draw the arrow pointing to the correct place. The Android toolbar doesn't use this by default, but I also changed it, so user can draw an arrow on the Android toolbar if they want to.
* I had to move the theme up. As the toolbars are displayed in an `Overlay`, if the `Theme` isn't above the `Overlay`, the toolbar won't see the correct theme.
* I added the screen edge padding configuration in the `MagnifierAndToolbarController`. I don't know if it's the right place for it. I added it there because the toolbar anchor is also there.

SuperEditor, Android with light theme:

editor_android_light

SuperEditor, Android with dark theme:

editor_android_dark

SuperEditor, iOS with light theme:

editor_ios_light

SuperEditor, iOS with dark theme:

editor_ios_dark

SuperTextField, Android:

textfield_android

SuperTextField, iOS:

textfield_ios

SuperEditor, Android theme switching:

https://user-images.githubusercontent.com/7597082/215135641-6727d317-a058-4d54-aee9-afad3618754e.mp4

SuperEditor, iOS theme switching:

https://user-images.githubusercontent.com/7597082/215135908-8a42e7eb-2bb4-4e87-8bc2-b8efc4ae42b2.mp4

Should the drag handle color on the dark theme be the same color as the toolbar color on Android?

### DIFF
--- a/super_editor/example/lib/demos/example_editor/example_editor.dart
+++ b/super_editor/example/lib/demos/example_editor/example_editor.dart
@@ -3,7 +3,6 @@ import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:super_editor/super_editor.dart';
 
-import '../../main.dart';
 import '_example_document.dart';
 import '_toolbar.dart';
 
@@ -30,6 +29,7 @@ class _ExampleEditorState extends State<ExampleEditor> {
 
   final _darkBackground = const Color(0xFF222222);
   final _lightBackground = Colors.white;
+  ValueNotifier<Brightness> _brightness = ValueNotifier<Brightness>(Brightness.light);
 
   SuperEditorDebugVisualsConfig? _debugConfig;
 
@@ -303,21 +303,34 @@ class _ExampleEditorState extends State<ExampleEditor> {
 
   @override
   Widget build(BuildContext context) {
-    return Stack(
-      children: [
-        Column(
-          children: [
-            Expanded(
-              child: _buildEditor(),
-            ),
-            if (_isMobile) _buildMountedToolbar(),
-          ],
-        ),
-        Align(
-          alignment: Alignment.bottomRight,
-          child: _buildCornerFabs(),
-        ),
-      ],
+    return ListenableBuilder(
+      listenable: _brightness,
+      builder: (context, _) {
+        return Theme(
+          data: ThemeData(brightness: _brightness.value),
+          child: Builder(
+            builder: (themedContext) {
+              // This builder captures the new theme
+              return Stack(
+                children: [
+                  Column(
+                    children: [
+                      Expanded(
+                        child: _buildEditor(themedContext),
+                      ),
+                      if (_isMobile) _buildMountedToolbar(),
+                    ],
+                  ),
+                  Align(
+                    alignment: Alignment.bottomRight,
+                    child: _buildCornerFabs(),
+                  ),
+                ],
+              );
+            },
+          ),
+        );
+      },
     );
   }
 
@@ -337,10 +350,9 @@ class _ExampleEditorState extends State<ExampleEditor> {
   }
 
   Widget _buildDebugVisualsToggle() {
-    final brightness = Theme.of(context).brightness;
     return FloatingActionButton(
-      backgroundColor: brightness == Brightness.light ? _darkBackground : _lightBackground,
-      foregroundColor: brightness == Brightness.light ? _lightBackground : _darkBackground,
+      backgroundColor: _brightness.value == Brightness.light ? _darkBackground : _lightBackground,
+      foregroundColor: _brightness.value == Brightness.light ? _lightBackground : _darkBackground,
       elevation: 5,
       onPressed: () {
         setState(() {
@@ -359,15 +371,14 @@ class _ExampleEditorState extends State<ExampleEditor> {
   }
 
   Widget _buildLightAndDarkModeToggle() {
-    final brightness = Theme.of(context).brightness;
     return FloatingActionButton(
-      backgroundColor: brightness == Brightness.light ? _darkBackground : _lightBackground,
-      foregroundColor: brightness == Brightness.light ? _lightBackground : _darkBackground,
+      backgroundColor: _brightness.value == Brightness.light ? _darkBackground : _lightBackground,
+      foregroundColor: _brightness.value == Brightness.light ? _lightBackground : _darkBackground,
       elevation: 5,
       onPressed: () {
-        AppSettings.of(context).toggleThemeBrightness();
+        _brightness.value = _brightness.value == Brightness.light ? Brightness.dark : Brightness.light;
       },
-      child: brightness == Brightness.light
+      child: _brightness.value == Brightness.light
           ? const Icon(
               Icons.dark_mode,
             )
@@ -377,7 +388,7 @@ class _ExampleEditorState extends State<ExampleEditor> {
     );
   }
 
-  Widget _buildEditor() {
+  Widget _buildEditor(BuildContext context) {
     final isLight = Theme.of(context).brightness == Brightness.light;
 
     return ColoredBox(
@@ -413,17 +424,33 @@ class _ExampleEditorState extends State<ExampleEditor> {
           gestureMode: _gestureMode,
           inputSource: _inputSource,
           keyboardActions: _inputSource == TextInputSource.ime ? defaultImeKeyboardActions : defaultKeyboardActions,
-          androidToolbarBuilder: (_, __) => AndroidTextEditingFloatingToolbar(
-            onCutPressed: _cut,
-            onCopyPressed: _copy,
-            onPastePressed: _paste,
-            onSelectAllPressed: _selectAll,
+          androidToolbarBuilder: (_, __) => ListenableBuilder(
+            listenable: _brightness,
+            builder: (context, _) {
+              return Theme(
+                data: ThemeData(brightness: _brightness.value),
+                child: AndroidTextEditingFloatingToolbar(
+                  onCutPressed: _cut,
+                  onCopyPressed: _copy,
+                  onPastePressed: _paste,
+                  onSelectAllPressed: _selectAll,
+                ),
+              );
+            },
           ),
-          iOSToolbarBuilder: (_, config) => IOSTextEditingFloatingToolbar(
-            onCutPressed: _cut,
-            onCopyPressed: _copy,
-            onPastePressed: _paste,
-            focalPoint: config.focalPoint,
+          iOSToolbarBuilder: (_, config) => ListenableBuilder(
+            listenable: _brightness,
+            builder: (context, _) {
+              return Theme(
+                data: ThemeData(brightness: _brightness.value),
+                child: IOSTextEditingFloatingToolbar(
+                  onCutPressed: _cut,
+                  onCopyPressed: _copy,
+                  onPastePressed: _paste,
+                  focalPoint: config.focalPoint,
+                ),
+              );
+            },
           ),
           overlayController: _overlayController,
         ),
--- a/super_editor/example/lib/demos/example_editor/example_editor.dart
+++ b/super_editor/example/lib/demos/example_editor/example_editor.dart
@@ -40,7 +40,7 @@ class _ExampleEditorState extends State<ExampleEditor> {
   final _imageSelectionAnchor = ValueNotifier<Offset?>(null);
 
   final _overlayController = MagnifierAndToolbarController() //
-    ..screenPadding = const EdgeInsets.all(8.0);
+    ..screenPadding = const EdgeInsets.all(20.0);
 
   @override
   void initState() {
--- a/super_editor/example/lib/demos/example_editor/example_editor.dart
+++ b/super_editor/example/lib/demos/example_editor/example_editor.dart
@@ -3,6 +3,7 @@ import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:super_editor/super_editor.dart';
 
+import '../../main.dart';
 import '_example_document.dart';
 import '_toolbar.dart';
 
@@ -29,7 +30,6 @@ class _ExampleEditorState extends State<ExampleEditor> {
 
   final _darkBackground = const Color(0xFF222222);
   final _lightBackground = Colors.white;
-  Brightness _brightness = Brightness.light;
 
   SuperEditorDebugVisualsConfig? _debugConfig;
 
@@ -39,7 +39,8 @@ class _ExampleEditorState extends State<ExampleEditor> {
   OverlayEntry? _imageFormatBarOverlayEntry;
   final _imageSelectionAnchor = ValueNotifier<Offset?>(null);
 
-  final _overlayController = MagnifierAndToolbarController();
+  final _overlayController = MagnifierAndToolbarController() //
+    ..screenPadding = const EdgeInsets.all(8.0);
 
   @override
   void initState() {
@@ -302,27 +303,21 @@ class _ExampleEditorState extends State<ExampleEditor> {
 
   @override
   Widget build(BuildContext context) {
-    return Theme(
-      data: ThemeData(brightness: _brightness),
-      child: Builder(builder: (themedContext) {
-        // This builder captures the new theme
-        return Stack(
+    return Stack(
+      children: [
+        Column(
           children: [
-            Column(
-              children: [
-                Expanded(
-                  child: _buildEditor(themedContext),
-                ),
-                if (_isMobile) _buildMountedToolbar(),
-              ],
+            Expanded(
+              child: _buildEditor(),
             ),
-            Align(
-              alignment: Alignment.bottomRight,
-              child: _buildCornerFabs(),
-            ),
+            if (_isMobile) _buildMountedToolbar(),
           ],
-        );
-      }),
+        ),
+        Align(
+          alignment: Alignment.bottomRight,
+          child: _buildCornerFabs(),
+        ),
+      ],
     );
   }
 
@@ -342,9 +337,10 @@ class _ExampleEditorState extends State<ExampleEditor> {
   }
 
   Widget _buildDebugVisualsToggle() {
+    final brightness = Theme.of(context).brightness;
     return FloatingActionButton(
-      backgroundColor: _brightness == Brightness.light ? _darkBackground : _lightBackground,
-      foregroundColor: _brightness == Brightness.light ? _lightBackground : _darkBackground,
+      backgroundColor: brightness == Brightness.light ? _darkBackground : _lightBackground,
+      foregroundColor: brightness == Brightness.light ? _lightBackground : _darkBackground,
       elevation: 5,
       onPressed: () {
         setState(() {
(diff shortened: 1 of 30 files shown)


super_editor/example/lib/demos/example_editor/example_editor.dart
```diff
@@ -302,27 +303,21 @@ class _ExampleEditorState extends State<ExampleEditor> {
 
   @override
   Widget build(BuildContext context) {
-    return Theme(
-      data: ThemeData(brightness: _brightness),
-      child: Builder(builder: (themedContext) {
-        // This builder captures the new theme
-        return Stack(
+    return Stack(
```
matthew-carroll, line 306: Is it possible for us to leave the `Theme` where it is, and then copy the `ThemeData` from here to the `Overlay` by adding a `Theme` widget around our stuff in the `Overlay`?

Themes can be overridden at various levels of the tree. It's probably not a good idea to force clients to apply their text editor `Theme` to their entire app just to get desired toolbar styling.

super_editor/example/lib/main.dart
```diff
@@ -535,3 +545,25 @@ class _DrawerButton extends StatelessWidget {
     );
   }
 }
+
+class AppSettings extends InheritedWidget {
```
matthew-carroll, line 549: Based on my comment about copying the `ThemeData` instead of moving the `Theme` widget, let's see if we can get rid of this widget, too.

super_editor/lib/src/infrastructure/touch_controls.dart
```diff
@@ -9,3 +11,17 @@ enum HandleType {
   /// Handle on the downstream side of an expanded selection.
   downstream,
 }
+
+typedef FloatingToolbarBuilder = Widget Function(BuildContext context, ToolbarConfig config);
```
matthew-carroll, line 15: Did we not have any toolbar arrow support previously? How are we positioning carets, drag handles, and magnifiers? Are we also pre-calculating the focal points for those and handing those focal points to those builders?

super_editor/pubspec.yaml
```diff
@@ -45,6 +46,9 @@ dependency_overrides:
   flutter_test_robots:
     git: 
       url: https://github.com/Flutter-Bounty-Hunters/flutter_test_robots
+  overlord:
```
matthew-carroll, line 49: Let's move this dependency declaration up above `flutter_test_robots` so that we can quickly read all the non-test dependencies before getting to the test dependencies.

super_editor/example/lib/demos/example_editor/example_editor.dart
angelosilvestre: Updated.

super_editor/lib/src/infrastructure/touch_controls.dart
angelosilvestre: We could position arrows if we use a `MagnifierAndToolbarController`. This object has offsets for drag handles and toolbars and magnifiers.

Should I just remove this and let users use `_overlayController.toolbarTopAnchor` to position the arrow?

matthew-carroll: For now, let's try that. The eventual answer is to use `follow_the_leader` to avoid tracking any of those details directly.

angelosilvestre: Updated `SuperEditor`. For `SuperTextField` we need a bigger refactor because we can't get the toolbar anchor where we need it to build the toolbar.

super_editor/lib/src/infrastructure/platforms/android/toolbar.dart
```diff
@@ -16,32 +16,38 @@ class AndroidTextEditingFloatingToolbar extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    final brightness = Theme.of(context).brightness;
     return Material(
-      borderRadius: BorderRadius.circular(4),
-      elevation: 3,
-      color: Colors.white,
+      borderRadius: BorderRadius.circular(1),
```
matthew-carroll, line 21: Is there really a 1px radius? Can you even notice a single pixel curve?

super_editor/lib/src/infrastructure/platforms/android/toolbar.dart
```diff
@@ -16,32 +16,38 @@ class AndroidTextEditingFloatingToolbar extends StatelessWidget {
...
+      elevation: 1,
+      color: brightness == Brightness.dark ? const Color(0xFF424242) : Colors.white,
```
matthew-carroll, line 23: Let's move the hard-coded color to a global constant with an appropriate name.

super_editor/lib/src/infrastructure/platforms/android/toolbar.dart
```diff
@@ -16,32 +16,38 @@ class AndroidTextEditingFloatingToolbar extends StatelessWidget {
...
+      type: MaterialType.card,
```
matthew-carroll, line 24: What does `card` give us?

super_editor/lib/src/infrastructure/platforms/android/toolbar.dart
```diff
@@ -16,32 +16,38 @@ class AndroidTextEditingFloatingToolbar extends StatelessWidget {
...
       child: Row(
         mainAxisSize: MainAxisSize.min,
         children: [
           if (onCutPressed != null)
             _buildButton(
               onPressed: onCutPressed!,
               title: 'Cut',
+              brightness: brightness,
```
matthew-carroll, line 32: Is there a widget that we can wrap around this `Row`, or around the whole toolbar, so that we're not repeating the brightness? For example, I think there's some kind of button theme widget.

super_editor/example/lib/demos/example_editor/example_editor.dart
```diff
@@ -418,16 +424,33 @@ class _ExampleEditorState extends State<ExampleEditor> {
...
+          iOSToolbarBuilder: (_) => ListenableBuilder(
+            listenable: _brightness,
+            builder: (context, _) {
+              return Theme(
+                data: ThemeData(brightness: _brightness.value),
+                child: IOSTextEditingFloatingToolbar(
+                  onCutPressed: _cut,
+                  onCopyPressed: _copy,
+                  onPastePressed: _paste,
+                  focalPoint: _overlayController.toolbarTopAnchor!,
```
matthew-carroll, line 450: How is this offset changing over time? It looks like we pass a single `Offset` and then never rebuild....

super_editor/lib/src/infrastructure/platforms/ios/toolbar.dart
```diff
@@ -1,49 +1,59 @@
...
-              title: 'Copy',
-            ),
-          if (onPastePressed != null)
-            _buildButton(
-              onPressed: onPastePressed!,
-              title: 'Paste',
-            ),
-        ],
-      ),
+    final brightness = Theme.of(context).brightness;
```
matthew-carroll, line 24: Same question here about brightness.

matthew-carroll (commented):
Added some comments. Also, you mentioned that you couldn't make the same change to SuperTextField as SuperEditor. Can you explain why? What are we missing in SuperTextField that prevents us from taking the same approach?

super_editor/lib/src/infrastructure/platforms/android/toolbar.dart
angelosilvestre: Wrapped with a `Theme` widget.

super_editor/lib/src/infrastructure/platforms/ios/toolbar.dart
angelosilvestre: Wrapped with a `Theme` widget.

super_editor/lib/src/infrastructure/platforms/android/toolbar.dart
angelosilvestre: It defines defaults to color and shape, but as we are defining manually it isn't needed. I removed.

super_editor/lib/src/infrastructure/platforms/android/toolbar.dart
angelosilvestre: It has a very small radius. I increased it to 2px. I didn't find a spec for this toolbar.

super_editor/example/lib/demos/example_editor/example_editor.dart
angelosilvestre: Currently, the toolbar only seems to follow the selection when we are scrolling...

angelosilvestre:
> Added some comments. Also, you mentioned that you couldn't make the same change to SuperTextField as SuperEditor. Can you explain why? What are we missing in SuperTextField that prevents us from taking the same approach?

In `SuperTextField` we don't have an object that holds the toolbar position as we have in `SuperEditor`. I'll see if I can change it.

angelosilvestre:
@matthew-carroll `SuperTextField` updated.

super_editor/example/lib/demos/example_editor/example_editor.dart
matthew-carroll: But how is it changing? We're passing a single `Offset` into `focalPoint`.

super_editor/lib/src/infrastructure/colors.dart
```diff
@@ -0,0 +1,7 @@
+import 'package:flutter/material.dart';
```
matthew-carroll, line 1: We have directories for android and iOS. Is there a reason that these colors are defined together?

super_editor/lib/src/infrastructure/platforms/android/toolbar.dart
```diff
@@ -1,5 +1,7 @@
 import 'package:flutter/material.dart';
 
+import '../../colors.dart';
```
matthew-carroll, line 3: Please ensure that all imports use fully specified paths. These relative paths break when we try to refactor file locations.

super_editor/example/lib/demos/example_editor/example_editor.dart
angelosilvestre: Whenever we scroll the document the toolbar is rebuilt.

super_editor/lib/src/infrastructure/colors.dart
angelosilvestre: Updated.

super_editor/lib/src/infrastructure/platforms/android/toolbar.dart
angelosilvestre: Updated.

matthew-carroll:
@angelosilvestre - I think I'm still waiting for an explanation as to how the focal point is working. I still don't see how passing a single `Offset` would give us, or is giving us, dynamic behavior. I'm either missing something, or the current approach has some blind spots.

angelosilvestre:
@matthew-carroll I thought I had answered that, sorry. 

Every time we scroll the document the toolbar builder is called, so we always get the new focal point.

matthew-carroll:
> Every time we scroll the document the toolbar builder is called, so we always get the new focal point.

How about when the user presses left/right arrows or up/down arrows to move the caret? Is there a selection listener that runs the builder?

Also, what if the `SuperEditor` widget, itself, moves? Is this offset local to the editor, or global? If it's global, I'm guessing that moving the `SuperEditor` widget would cause the focal point to mis-align.

matthew-carroll:
@Jethro87 can you try this PR and let me know if it meets your expectations?

Jethro87:
@matthew-carroll I did a quick test and it needs some work. There are some clear styling issues, like the non-standard `InkWell` effect on iOS (from the `TextButton`), and padding issues related to the `onPressed` state of the buttons.

I'll dig a bit deeper and add a longer comment.

matthew-carroll:
@angelosilvestre as soon as @Jethro87 provides more information on remaining issues, can you please adjust the PR accordingly?

angelosilvestre:
@matthew-carroll Sure

angelosilvestre:
@Jethro87 could you please try this PR again? All of the issues you reported should be solved now.

I also changed temporarily the `overlord` dependency to use a forked version. After merging these changes to the `overlord` repo, I will revert that dependency.

Jethro87:
@angelosilvestre I can't seem to build, as `LinearBorder` can't be found. I tried upgrading to `master`, as I see `LinearBorder` was merged in January, but it still won't build, `Error: Couldn't find constructor 'LinearBorder'`. I am referencing your branch as a dependency

```
  super_editor:
    git:
      url: https://github.com/angelosilvestre/super_editor.git
      path: super_editor
      ref: 963_mobile_toolbar_design
```

Any thoughts? I'll keep trying to get this built.

angelosilvestre:
@Jethro87 I'll take a look

angelosilvestre:
@Jethro87 It turns out that after the last changes, this shape isn't needed anymore.

I updated the PR

Jethro87:
@angelosilvestre This is looking good, thanks for updating. A couple of additional issues then I think we're good to go.

The left button highlight doesn't extend to the edges of the container:
IMG_1215

On Android, there is no padding between the toolbar and the edge of the screen. I'm not sure if there is supposed to be, but I thought I'd mention it:
IMG_1216

angelosilvestre:
@Jethro87 I update the button highlight. I steel need to take a look at the padding on android

Jethro87:
@angelosilvestre Thanks, the iOS part looks good.

angelosilvestre:
@Jethro87 Updated for Android.

After investigation, the screen padding on iOS wasn't there on purpose. The widget that displays the popover shape also allows displaying the arrow pointing to the left or right side of the popover. Because of that, it reserves space for the arrow on both x and y axis.

I changed this widget to avoid reserving space on the x-axis when the popover isn't allowed to point right or left. The screen padding should be configured in the `MagnifierAndToolbarController`s `screenPadding` property, as can be seen in the demo app. This way, the padding will be consistent in both Android and iOS.

Jethro87:
@angelosilvestre LGTM

angelosilvestre:
https://github.com/Flutter-Bounty-Hunters/overlord/pull/14 needs to land before we can merge this PR.

matthew-carroll:
@angelosilvestre it looks like we have a number of conflicts in this PR. Can you resolve those, and let me know if it's ready for review again after that?

angelosilvestre:
@matthew-carroll You can review it again.

super_editor/example/pubspec.yaml
```diff
@@ -62,6 +62,10 @@ dependency_overrides:
     path: ../../super_text_layout
   attributed_text:
     path: ../../attributed_text
+  overlord:
+    git:
+      url: https://github.com/angelosilvestre/overlord.git
+      ref: update_toolbar_design
```
matthew-carroll, line 68: You'll need to remember to remove this ref as soon as we merge the Overlord work. The only reason this is OK is because its for the `example` project, but we still want to get rid of it ASAP.

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -22,6 +22,7 @@ import 'package:super_editor/src/infrastructure/content_layers.dart';
 import 'package:super_editor/src/infrastructure/links.dart';
 import 'package:super_editor/src/infrastructure/platforms/ios/ios_document_controls.dart';
 import 'package:super_editor/src/infrastructure/text_input.dart';
+import 'package:super_editor/src/infrastructure/touch_controls.dart';
```
matthew-carroll, line 25: Does this need to be here?

super_editor/lib/src/infrastructure/super_textfield/android/_editing_controls.dart
```diff
@@ -12,6 +12,8 @@ import 'package:super_editor/src/infrastructure/toolbar_position_delegate.dart';
 import 'package:super_editor/src/infrastructure/touch_controls.dart';
 import 'package:super_text_layout/super_text_layout.dart';
 
+import '../metrics.dart';
```
matthew-carroll, line 15: Let's avoid relative imports. They get screwed up when we refactor.

super_editor/lib/src/infrastructure/super_textfield/android/android_textfield.dart
```diff
@@ -545,7 +546,8 @@ class SuperAndroidTextFieldState extends State<SuperAndroidTextField>
   }
 }
 
-Widget _defaultAndroidToolbarBuilder(BuildContext context, AndroidEditingOverlayController controller) {
+Widget _defaultAndroidToolbarBuilder(
+    BuildContext context, AndroidEditingOverlayController controller, ToolbarConfig config) {
```
matthew-carroll, line 550: Is there a reason we're taking in a `ToolbarConfig` but not using it?

super_editor/pubspec.yaml
```diff
@@ -43,10 +44,15 @@ dependency_overrides:
     path: ../attributed_text
   super_text_layout:
     path: ../super_text_layout
+  overlord:
+    git:
+      url: https://github.com/angelosilvestre/overlord.git
+      ref: update_toolbar_design
```
matthew-carroll, line 50: I see we also have this reference here. We can't have any git dependencies in our primary package pubspec. We'll need to release an updated `overlord` before we can merge this.

super_editor_markdown/pubspec.yaml
```diff
@@ -24,6 +24,10 @@ dependency_overrides:
     path: ../super_text_layout
   attributed_text:
     path: ../attributed_text
+  overlord:
```
matthew-carroll, line 27: Why do we need this here?

super_editor/lib/src/infrastructure/super_textfield/android/android_textfield.dart
angelosilvestre: We are taking the toolbar `ToolbarConfig` because the popover toolbar builder is declared as

```dart
final Widget Function(BuildContext, AndroidEditingOverlayController, ToolbarConfig) popoverToolbarBuilder;
```

The `ToolbarConfig` is only used on the iOS toolbar.

super_editor/pubspec.yaml
angelosilvestre: Could you release a new `overlord` version, so we can proceed with this PR?

matthew-carroll (approved):
LGTM

PR #1009, opened 2023-01-27
[SuperTextField][mobile] Update viewport height on first frame if possible (Resolves #939)

### LINKED ISSUE #939: [SuperTextField] flicker on initial display (iOS)

angelosilvestre:
[SuperTextField][mobile] Update viewport height on first frame if possible. Resolves #939 

When we switch a widget with a `SuperTextField`, the text isn't displayed in the first frame. The cause is that we defer the viewport height calculation until the layout runs for the first time. We do this because we need the text layout to estimate the line height and calculate the total height we need.

This PR changes `TextScrollView` to update its viewport height in the build method when possible. Currently, this is only possible when the `SuperTextField` has a `lineHeight` set. If this isn't the case, we defer the calculation until the end of the current frame.

We could also try to estimate the line height using a `Paragraph`, as we do in the desktop text field.

I replaced the `Opacity` with an `Offstage` as it makes it easier to test.


super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -465,20 +489,14 @@ class _TextScrollViewState extends State<TextScrollView>
...
     if (widget.textKey.currentContext == null || _needViewportHeight) {
-      // The text hasn't been laid out yet, which means our calculations
-      // for text height is probably wrong. Schedule a post frame callback
-      // to re-calculate the height after initial layout.
-      WidgetsBinding.instance.addPostFrameCallback((timeStamp) {
-        if (mounted) {
-          _updateViewportHeight();
-        }
-      });
+      // Try to update the viewport height synchronously.
```
matthew-carroll, line 492: Using the term "synchronously" is a bit misleading. This isn't really a "synchronous" vs "asynchronous" issue. Let's change this comment to the following:

```dart
/// The text hasn't been laid out yet, which means we don't know how tall to
/// make the viewport, based on lines of text. Try to calculate a viewport height
/// based on available information. If there's not enough information to choose
/// a viewport height, then schedule a post-frame callback to inspect the final
/// text layout and select a viewport height based on the number of actual lines
/// of text in the layout.
```

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -465,20 +489,14 @@ class _TextScrollViewState extends State<TextScrollView>
...
+      // If we can't, the viewport height is calculated at the end of the frame
+      // and the widget is rebuilt.
+      _updateViewportHeight();
     }
 
-    return Opacity(
-      opacity: (widget.maxLines != null && widget.maxLines! > 1 && _viewportHeight == null && _needViewportHeight)
-          ? 0.0
-          : 1.0,
+    return Offstage(
```
matthew-carroll, line 498: I see that you've replaced `Opacity` with `Offstage`. Maybe you asked me about this? I can't remember. Did we say that we didn't know why `Opacity` was used, and therefore we replaced it with `Offstage`? Can you tell me what `Offstage` is doing for you here?

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -441,7 +464,8 @@ class _TextScrollViewState extends State<TextScrollView>
     }
 
     if (widget.textKey.currentState == null) {
-      return 0;
+      // If we can't get a text layout, compute the line count by counting the line breaks.
+      return widget.textEditingController.text.text.split('\n').length;
```
matthew-carroll, line 468: Can you explain why this was added? It's very likely that the number of `\n` characters bare no resemblance to the actual number of lines in the layout. In fact, the most typical scenario for a text field is a single line of text that wraps many times. This value would be completely wrong...

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -402,20 +417,28 @@ class _TextScrollViewState extends State<TextScrollView>
...
       return didChange;
     }
 
     if (viewportHeight != null) {
-      setState(() {
-        _log.finer(' - new viewport height: $viewportHeight');
-        _needViewportHeight = false;
-        _viewportHeight = viewportHeight;
+      _log.finer(' - new viewport height: $viewportHeight');
+      // We set these values outside of setState, because this method
```
matthew-carroll, line 434: Seems like we should know which situation we're dealing with. Either, we know everything we need to know right now, in which case we shouldn't schedule a post-frame callback that then rebuilds everything. Or, we don't know everything we need to know right now, in which case calculations should be done at the end of the frame, instead of during the build phase. Right?

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -402,20 +417,28 @@ class _TextScrollViewState extends State<TextScrollView>
       _log.finer(
           ' - viewport height is null, but TextScrollView is unbounded or the content fits max height, so that is OK');
       final didChange = viewportHeight != _viewportHeight;
-      if (mounted) {
-        setState(() {
-          _needViewportHeight = false;
-          _viewportHeight = null;
-        });
-      }
+      // We set these values outside of setState, because this method
```
matthew-carroll, line 420: Same comment here as below.

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -354,6 +355,20 @@ class _TextScrollViewState extends State<TextScrollView>
         estimatedLineHeight = widget.textKey.currentState!.textLayout.estimatedLineHeight;
         _log.finer(' - estimated line height based on text styles: $estimatedLineHeight');
       }
+    } else {
+      estimatedLineHeight = 0;
+    }
+
+    if (estimatedLineHeight == 0) {
```
matthew-carroll, line 362: Let's not use magic numbers for this. Let's make `estimatedLineHeight` nullable, let's get rid of the `else` statement above, and then let's check for `null` here.

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -340,7 +341,7 @@ class _TextScrollViewState extends State<TextScrollView>
       _log.finer(' - explicit line height provided: ${widget.lineHeight}');
       // Use the line height that was explicitly provided by the widget.
       estimatedLineHeight = widget.lineHeight!;
-    } else {
+    } else if (widget.textKey.currentState != null) {
```
matthew-carroll, line 344: Can you tell me why this check for a null `currentState` wasn't here before, but we need it now?

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
angelosilvestre: The end result of `Opacity` and `Offstage` will be the same. I replaced it only because it makes testing easier. 

With `Opacity`, `find.text()` will find a widget even if it is completely transparent. With `Offstage`, the widget won't be found by default if `offstage` is `true`.

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
angelosilvestre: Without this, we would got zero as the lines of text in the first frame and we wouldn't be able to calculate the height we need...

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
angelosilvestre: Before this PR, `_updateViewportHeight` was called only after the text was laid out, so `widget.textKey.currentState` couldn't be null.

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
angelosilvestre: One alternative would be to use a `Paragraph` to calculate this.

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
angelosilvestre: This isn't only related to whether or not we can run the calculations on the current frame. 

`_updateViewportHeight` can be called when the text changes. Because of that, we need to call `setState` even if we can calculate the height on the same frame.

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
matthew-carroll: > Without this, we would got zero as the lines of text in the first frame and we wouldn't be able to calculate the height we need...

That might be true, but the value that you're gonna get from this line will be very wrong for most use-cases. So I'm wondering two things. First, is this situation related to the problem that we're trying to solve with this PR? I thought the situation we were interested in was one in which we're told the number of lines and the height of each line. Second, what's the value of this line given that this line will result in very wrong numbers for most use-cases?

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -356,6 +356,18 @@ class _TextScrollViewState extends State<TextScrollView>
       }
     }
 
+    if (estimatedLineHeight == null) {
+      _log.finer(' - could not calculate the estimated line height. Rescheduling calculation.');
+      WidgetsBinding.instance.addPostFrameCallback((timeStamp) {
+        if (mounted) {
+          setState(() {
```
matthew-carroll, line 363: Why do we need this `setState()` around the `_updateViewportHeight()` call?

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
matthew-carroll: I don't understand what you mean.

> _updateViewportHeight can be called when the text changes

It "can be called" or it "is called"? And from where? Are you talking about `didUpdateWidget()`? And why does that require that we call `setState()`?

angelosilvestre: It is called. It isn't about `didUpdateWidget()`. 

The viewport height depends on the number of lines of text. Therefore, whenever the text in the `AttributedTextEditingController` changes we need to recalculate the viewport height because the user might have added more lines.

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
angelosilvestre: It doesn't need to be around, but we need the `setState()` to rebuild the widget. 

If we call `_updateViewportHeight` in a post-frame callback, we need `setState` to schedule a new frame and build the widget using the calculated height.

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
angelosilvestre: Reviewing this again, it seems the only issue we have was some failing tests that use multi-line text fields and have a fixed `lineHeight`. Removing the `lineHeight` "fixes" the tests.

However, this indicates that we might break users that apply fixed `lineHeight`s in multi-line textfields.

Because we have a `lineHeight` and we can't get the number of lines of text, we end up with a textfield that is only one line tall.

matthew-carroll: Having a fixed `lineHeight` for a multiline field is a valid thing to do. So removing that value wouldn't be an appropriate solution.

I still don't see why this "\n" approach would be appropriate. I understand that it avoids a value of zero. You could also hard code `1000`, which would avoid a value of zero, but it wouldn't be an appropriate solution, either.

We already have a concept of "not having enough information to choose a viewport height", for which we then schedule another frame. Why aren't we doing that in this case?

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
matthew-carroll: What's the execution path, and why do we need to schedule another frame to invalidate the build? It's still not clear.

angelosilvestre:
@matthew-carroll I updated this PR to also compute the viewport on the first frame when not given a line height. Could you please review it again?

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -377,76 +405,115 @@ class _TextScrollViewState extends State<TextScrollView>
...
     if (widget.textKey.currentState == null) {
-      return 0;
+      return null;
     }
 
     return _textLayout.getLineCount();
   }
 
+  /// Computes information about the text layout using a [TextPainter].
+  _TextLayoutInfo _computeTextLayoutInfo() {
```
matthew-carroll, line 495: What's all of this? I'm pretty sure we shouldn't be doing this here. Can you explain why this was added and why this would be the appropriate place to collect such low level details?

angelosilvestre: I added this to be able to update the viewport height on the first frame for more cases. 

If the number of lines of text isn't available we can use this method to compute it.

matthew-carroll: I need a lot more information than that. This appears to re-implement text layout in a scrollview. So like I asked above, can you explain why this would be the appropriate place to do such a thing?

angelosilvestre: I thought it would be ok to do this text layout in the first frame... I reverted it.

matthew-carroll:
@angelosilvestre it looks like there are still unanswered questions in this thread. Can you either respond to those questions or explain why they're no longer relevant?

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
angelosilvestre: I reverted the `lineHeight` removal and also the `\n` approach.

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
angelosilvestre: `_updateViewportHeight` is called from the `build` method and also when the text changes. We need another frame because, when this method is called from the `build` method, we can't call `setState` directly.

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
angelosilvestre: Same as the previous answer.

angelosilvestre:
@matthew-carroll I answered the remaining questions.

super_editor/test/super_textfield/super_textfield_auto_scroll_test.dart
```diff
@@ -142,6 +142,7 @@ Future<void> _pumpTestApp(
       ),
     ),
   );
+  await tester.pumpAndSettle();
```
matthew-carroll, line 145: We do we need to settle here?

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -356,6 +356,37 @@ class _TextScrollViewState extends State<TextScrollView>
       }
     }
 
+    final hasLineConstraints = widget.maxLines != null || widget.minLines != null;
```
matthew-carroll, line 359: Is there any reason that this isn't run immediately after the first print statement in this method? Seems like the `estimatedLineHeight` stuff above is unnecessary in the case of `!hasLineConstraints`

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -356,6 +356,37 @@ class _TextScrollViewState extends State<TextScrollView>
...
+    if (!hasLineConstraints) {
```
matthew-carroll, line 360: Can you add an appropriate log statement in this conditional so that we know this condition happened?

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -356,6 +356,37 @@ class _TextScrollViewState extends State<TextScrollView>
...
+      // We don't have line constraints so we don't need to estimate the content height
+      // and compute a fixed viewport height. The viewport will size itself based on the
+      // text intrinsic height.
+
+      final didChange = _viewportHeight != null;
+      // We set these values outside of setState, because this method
+      // can be called during the build phase.
+      _needViewportHeight = false;
+      _viewportHeight = null;
+      WidgetsBinding.instance.addPostFrameCallback((timeStamp) {
```
matthew-carroll, line 370: Should we use the our tool for running a callback as soon as possible? It's possible that we can run this behavior immediately, right?

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -356,6 +356,37 @@ class _TextScrollViewState extends State<TextScrollView>
...
+        if (mounted) {
+          setState(() {});
```
matthew-carroll, line 372: Are we calling `setState()` because we changed `_needViewportHeight` and `_viewportHeight`? If so, shouldn't those variables be changed inside of `setState()`?

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -377,45 +408,53 @@ class _TextScrollViewState extends State<TextScrollView>
...
-          _needViewportHeight = false;
-          _viewportHeight = null;
-        });
-      }
+      final didChange = newViewportHeight != _viewportHeight;
+      // We set these values outside of setState, because this method
+      // can be called during the build phase.
+      _needViewportHeight = false;
+      _viewportHeight = null;
+      WidgetsBinding.instance.addPostFrameCallback((timeStamp) {
```
matthew-carroll, line 440: Same comments for running this callback, and for putting these variables in `setState()`

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
matthew-carroll: Please add a comment about why we're rebuilding. It's not clear why we need to update the tree.

super_editor/test/super_textfield/super_textfield_auto_scroll_test.dart
angelosilvestre: It seems `pump` will work

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
angelosilvestre: Moved up.

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
angelosilvestre: Added.

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
angelosilvestre: Comment added.

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
angelosilvestre: Changed to use `runAsSoonAsPossible`.

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
angelosilvestre: We need to set these values outside of the `setState` because when we call this function from the build method, these values need to be available immediately, as we use `_viewportHeight` to build the tree. If we don't do this, the `build` method will  'see' the old value.

The `setState` is used because `_updateViewportHeight` is called when the text changes. Because of this, we need to call `setState` to rebuild the widget.

angelosilvestre:
@matthew-carroll Rebased to fix the goldens

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -332,15 +333,36 @@ class _TextScrollViewState extends State<TextScrollView>
...
+      // We don't have line constraints so we don't need to estimate the content height
+      // and compute a fixed viewport height. The viewport will size itself based on the
+      // text intrinsic height.
+
+      final didChange = _viewportHeight != null;
+      // We set these values outside of setState, because this method
+      // can be called during the build phase.
+      _needViewportHeight = false;
+      _viewportHeight = null;
+      WidgetsBinding.instance.runAsSoonAsPossible(() {
```
matthew-carroll, line 349: It looks like the somewhat confusing behavior related to `setState()` is the result of this method being called from different phases, e.g., "idle" and "build".

What if we create another method called `_updateViewportHeightAndRebuild()` and we call that method from the places where a build isn't currently happening? Then this method would only make changes to values - no scheduling or requesting rebuilds?

super_editor/test/super_textfield/super_textfield_auto_scroll_test.dart
```diff
@@ -142,6 +142,7 @@ Future<void> _pumpTestApp(
       ),
     ),
   );
+  await tester.pump();
```
matthew-carroll, line 145: Why do we need to pump here? What does the 2nd pump accomplish?

super_editor/test/super_textfield/super_textfield_text_alignment_test.dart
```diff
@@ -227,8 +227,8 @@ Widget _buildSuperTextField({
       textController: controller,
       textAlign: textAlign,
       maxLines: maxLines,
-      minLines: 1,      
-      lineHeight: 20,      
+      lineHeight: 20,
```
matthew-carroll, line 230: Is there a reason we flipped the order of these properties?

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -465,20 +514,17 @@ class _TextScrollViewState extends State<TextScrollView>
...
+      /// of text in the layout.
+      _updateViewportHeight();
     }
 
-    return Opacity(
-      opacity: (widget.maxLines != null && widget.maxLines! > 1 && _viewportHeight == null && _needViewportHeight)
-          ? 0.0
-          : 1.0,
+    return Offstage(
+      offstage: (widget.maxLines != null && widget.maxLines! > 1 && _viewportHeight == null && _needViewportHeight),
```
matthew-carroll, line 527: Does `_needViewportHeight` not implicitly include `widget.maxLines != null && widget.maxLines! > 1`?

This seems like a lot of conditions for us to be throwing together here....

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
angelosilvestre: Updated.

super_editor/test/super_textfield/super_textfield_text_alignment_test.dart
angelosilvestre: It seems I moved by mistake. Reverted.

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
angelosilvestre: It seems that even if we `_needViewportHeight`, we allow the text to be displayed if we can grow indefinetely (`maxLines == null`) or if we have a single-line textfield.

This condition was already like this before this PR.

super_editor/test/super_textfield/super_textfield_auto_scroll_test.dart
angelosilvestre: In these tests, we aren't able to compute the viewport height in the first frame. 

Before this PR, this wasn't a problem for these tests, because even though the text field wasn't visible (0% opacity), the `finder` could still find it.

We could revert the `TextScrollView` to use the `Opacity` widget, but then the new tests for rendering on the first frame will need to check if the text field has a descendant `Opacity` widget with 100% opacity. Do you have another suggestion for this test? A golden test also wouldn't work, because `screenMatchesGolden` seems to `pumpAndSettle`.

matthew-carroll: Let's just add a comment above the pump that explains why we need it.

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
matthew-carroll: What is we set `_needViewportHeight` to false when `maxLines == null`? Does that interfere with anything else?

angelosilvestre: If `maxLines == null` and `minLines > 1` then we still need to perform the computation to determine the `minHeight`. 

We could set `_needViewportHeight`, but still re-schedule the viewport height computation. I'll try this.

super_editor/test/super_textfield/super_textfield_auto_scroll_test.dart
angelosilvestre: Updated.

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
angelosilvestre: Updated

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -144,16 +144,16 @@ class _TextScrollViewState extends State<TextScrollView>
     }
 
     if (widget.textEditingController != oldWidget.textEditingController) {
-      oldWidget.textEditingController.removeListener(_scheduleViewportHeightUpdate);
-      widget.textEditingController.addListener(_scheduleViewportHeightUpdate);
+      oldWidget.textEditingController.removeListener(_scheduleViewportHeightUpdateAndRebuild);
+      widget.textEditingController.addListener(_scheduleViewportHeightUpdateAndRebuild);
 
-      _scheduleViewportHeightUpdate();
+      _scheduleViewportHeightUpdateAndRebuild();
```
matthew-carroll, line 150: I don't think we should be calling the "rebuild" version here. We're not allowed to call `setState()` from `didUpdateWidget`, so I'm surprised this isn't crashing. But also `build()` always runs after `didUpdateWidget()`, anyway.

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -144,16 +144,16 @@ class _TextScrollViewState extends State<TextScrollView>
...
     }
 
     if (widget.minLines != oldWidget.minLines ||
         widget.maxLines != oldWidget.maxLines ||
         widget.lineHeight != oldWidget.lineHeight) {
-      _scheduleViewportHeightUpdate();
+      _scheduleViewportHeightUpdateAndRebuild();
```
matthew-carroll, line 156: Same here

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -1007,3 +1068,25 @@ enum _AutoScrollDirection {
   start,
   end,
 }
+
+/// The result of the viewport height computation.
```
matthew-carroll, line 1072: If this object can be `pending` then this object can't also be "the result". Those are mutually exclusive concepts.

Does this data structure help us somehow? This is a lot of extra stuff just to move two variables out of a state object. What's the goal here?

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -377,83 +394,130 @@ class _TextScrollViewState extends State<TextScrollView>
...
     }
 
     return _textLayout.getLineCount();
   }
 
-  void _scheduleViewportHeightUpdate() {
-    // The viewport height is calculated using the number of the lines of text.
-    // Therefore, when text changes we need to recalculate the viewport height
-    // to accommodate the new text.
+  void _scheduleViewportHeightUpdateAndRebuild() {
```
matthew-carroll, line 517: Let's place this method above `updateViewportHeightAndRebuild`

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -328,19 +328,28 @@ class _TextScrollViewState extends State<TextScrollView>
     }
   }
 
-  /// Returns true if the viewport height changed, false otherwise.
-  bool _updateViewportHeight() {
-    _log.finer('Updating viewport height...');
+  _ViewportComputationResult _computeNewViewportHeight() {
```
matthew-carroll, line 331: Let's place this method below `_updateViewportHeight()`

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
angelosilvestre: The `_scheduleViewportHeightUpdateAndRebuild` calls `_updateViewportHeightAndRebuild` in a post-frame callback, so it won't be calling `setState` during `didUpdateWidget`. 

But maybe we could just set `_needViewportHeight` to true, and the calculation will run again during `build`.

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
angelosilvestre: With the last commit, I extracted the computation to a different method, so we could get the computation in the `_updateViewportHeight` and `_updateViewportHeightAndRebuild` methods.

When we call `_computeNewViewportHeight` , we need two pieces of information:
1.  If the computation succeeded or if we need to re-schedule it.
2.  The new viewport height, if the computation succeeded.

Is `_ViewportComputationStatus` a better name?

matthew-carroll: If the computation succeeded, will the value always be non-null? If so, you can just return a nullable value. `null` means it failed, and non-null means it succeeded with the given value.

angelosilvestre: > If the computation succeeded, will the value always be non-null?

No. `null` is a valid value, which means the viewport should expand to fit its content.

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -1007,3 +1068,25 @@ enum _AutoScrollDirection {
   start,
   end,
 }
+
+/// The status of the viewport height computation.
+///
+/// When [pending] is `true`, the computation needs to be rescheduled to the next frame.
+///
+/// When [pending] is `false`, [viewportHeight] must be used as the new viewport height.
+class _ViewportComputationStatus {
```
matthew-carroll, line 1077: Let's do this:

```dart
// TODO: add appropriate dart docs
class _ViewportHeight {
  bool get isBounded => value == null;

  final double? value;
}
```

When the reference to `_ViewportHeight` is `null`, that means it hasn't been calculated yet. When `_viewportHeight.value` is `null`, it means the height isn't bounded (which is also made explicit by `isBounded`)

Does that solve for our various conditions without missing anything? This eliminates an explicit concept of "pending", and it doesn't confuse the concept of "status" with the concept of a "computed height".

angelosilvestre: It'll work. Updated.

angelosilvestre:
@matthew-carroll It seems the Flutter latest main broke one golden test with a slight rendering difference. I updated that golden file.

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -328,19 +328,96 @@ class _TextScrollViewState extends State<TextScrollView>
...
+  /// If the result is non `null`, the computation is completed.
+  _ViewportHeight? _computeNewViewportHeight() {
+    final hasLineConstraints = widget.maxLines != null || widget.minLines != null;
+    if (!hasLineConstraints) {
+      _log.finer(' - the widget does\'n have line number constraints. Sizing by intrinsic height.');
+
+      // We don't have line constraints so we don't need to estimate the content height
+      // and compute a fixed viewport height. The viewport will size itself based on the
+      // text intrinsic height.
+      return _ViewportHeight();
```
matthew-carroll, line 409: Can this a `const` constructor?

matthew-carroll (commented):
I think there's just this one last little detail

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
angelosilvestre: Yeah. Updated.

matthew-carroll (approved):
LGTM

PR #1013, opened 2023-01-31
[SuperEditor] Fix selection when the editor has autofocus (Resolves #883)

angelosilvestre:
[SuperEditor] Fix selection when the editor has autofocus. Resolves #883

When the editor has `autofocus`, the selection isn't set and we get the following exception:

```
flutter: Initializing logger: ExampleApp

The following NoSuchMethodError was thrown while dispatching notifications for FocusNode:
The method 'findLastSelectablePosition' was called on null.
Receiver: null
Tried calling: findLastSelectablePosition()

#0      Object.noSuchMethod (dart:core-patch/object_patch.dart:38:5)
#1      _EditorSelectionAndFocusPolicyState._onFocusChange
package:super_editor/…/default_editor/document_focus_and_selection_policies.dart:110
#2      ChangeNotifier.notifyListeners
package:flutter/…/foundation/change_notifier.dart:351
#3      FocusNode._notify
package:flutter/…/widgets/focus_manager.dart:1038
#4      FocusManager._applyFocusChange
package:flutter/…/widgets/focus_manager.dart:1804
(elided 2 frames from dart:async)
The FocusNode sending notification was: FocusNode#2c45c([PRIMARY FOCUS])
    context: Focus
    PRIMARY FOCUS
```

The cause is that the focus change listener is called before we can access the document layout. We have tests to ensure the selection is set when the editor has `autofocus`, but this doesn't seem to happen during tests.

This PR changes `EditorSelectionAndFocusPolicy` to check if we can access the document layout before we try to find the last selectable position. 

Other possible solutions: 

1. Change `EditorSelectionAndFocusPolicy` to receive a function that returns a `bool` indicating if the document has been laid out, instead of a `GlobalKey` to the `DocumentLayout`.
2. Change `DocumentLayoutResolver` to return a nullable `DocumentLayout`. This approach might be the more appropriate one, but it will require changes in many places.

This PR also overrides the `attributed_text` dependency, because we have a method that isn't present yet in the pub.dev released version.

### DIFF
--- a/super_editor/lib/src/default_editor/document_focus_and_selection_policies.dart
+++ b/super_editor/lib/src/default_editor/document_focus_and_selection_policies.dart
@@ -9,7 +9,7 @@ class EditorSelectionAndFocusPolicy extends StatefulWidget {
     Key? key,
     required this.focusNode,
     required this.selection,
-    required this.documentLayoutKey,
+    required this.isDocumentLayoutAvailable,
     required this.getDocumentLayout,
     this.placeCaretAtEndOfDocumentOnGainFocus = true,
     this.restorePreviousSelectionOnGainFocus = true,
@@ -17,8 +17,14 @@ class EditorSelectionAndFocusPolicy extends StatefulWidget {
     required this.child,
   }) : super(key: key);
 
-  /// [GlobalKey] to access the document layout.
-  final GlobalKey documentLayoutKey;
+  /// Returns whether or not we can access the document layout, which is needed for [placeCaretAtEndOfDocumentOnGainFocus].
+  ///
+  /// When [SuperEditor] has `autofocus`, the focus change callback is called before we can access
+  /// the document layout using [getDocumentLayout]. If [getDocumentLayout] is called before we can
+  /// access the document layout we get an exception.
+  ///
+  /// When this method returns `true`, we assume it's safe to call [getDocumentLayout].
+  final bool Function() isDocumentLayoutAvailable;
 
   /// The document editor's [FocusNode].
   ///
@@ -64,11 +70,6 @@ class EditorSelectionAndFocusPolicy extends StatefulWidget {
 class _EditorSelectionAndFocusPolicyState extends State<EditorSelectionAndFocusPolicy> {
   bool _wasFocused = false;
   DocumentSelection? _previousSelection;
-
-  /// Whether or not the document has been laid out.
-  ///
-  /// If `true`, we can access the document layout.
-  bool get _isDocumentLaidOut => widget.documentLayoutKey.currentContext != null;
 
   @override
   void initState() {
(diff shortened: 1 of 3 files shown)


matthew-carroll:
Can you try to get a test to fail with this? I know you said our existing tests don't. I'm wondering if that's due to the `pumpAndSettle()` when we pump a document. I'd like for us to have a failing test for the existing problem.

angelosilvestre:
> Can you try to get a test to fail with this? I know you said our existing tests don't. I'm wondering if that's due to the `pumpAndSettle()` when we pump a document. I'd like for us to have a failing test for the existing problem.

In the tests, even with a single `pump` we can access the document layout when the focus change callback is called.

The focus change callback seems to be called at a different moment in the tests. Maybe we should try to write a minimal sample and file an issue in flutter.

matthew-carroll:
> Maybe we should try to write a minimal sample and file an issue in flutter.

Sure. See if you can create a minimal code sample that tries to access a state object, which fails with regular Flutter, but passes in a test.

super_editor/lib/src/default_editor/document_focus_and_selection_policies.dart
```diff
@@ -9,13 +9,17 @@ class EditorSelectionAndFocusPolicy extends StatefulWidget {
...
+    required this.documentLayoutKey,
     required this.getDocumentLayout,
     this.placeCaretAtEndOfDocumentOnGainFocus = true,
     this.restorePreviousSelectionOnGainFocus = true,
     this.clearSelectionWhenEditorLosesFocus = true,
     required this.child,
   }) : super(key: key);
 
+  /// [GlobalKey] to access the document layout.
+  final GlobalKey documentLayoutKey;
```
matthew-carroll, line 21: Let's change this to `bool Function() isDocumentLayoutAvailable` so that we can minimize this widget's exposure to the implementation details.

Also, in the updated Dart Doc, include a secondary sentence that explains why this information is needed.

super_editor/lib/src/default_editor/document_focus_and_selection_policies.dart
```diff
@@ -107,6 +116,17 @@ class _EditorSelectionAndFocusPolicyState extends State<EditorSelectionAndFocusP
         // Place the caret at the end of the document.
         editorPoliciesLog
             .info("[${widget.runtimeType}] - placing caret at end of document because the editor gained focus");
+        if (!_isDocumentLaidOut) {
+          // We are focused, but the document hasn't been laid out yet. This could happen if SuperEditor has autofocus.
+          // Wait until the end of the frame, so we have access to the document layout.
+          editorPoliciesLog.info(
+              "[${widget.runtimeType}] - the document hasn't been laid out yet. Trying again at the end of the frame");
+          WidgetsBinding.instance.addPostFrameCallback((timeStamp) {
```
matthew-carroll, line 124: If you switch this to "schedule" the callback instead of "add", does it still work as expected?

When we `addPostFrameCallback` we're assuming that `_onFocusChange` was called in the middle of a frame. But that's not technically a guarantee. We don't really control the execution path that got us here. So it might be safer to schedule a frame.

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -507,6 +507,7 @@ class SuperEditorState extends State<SuperEditor> {
       child: EditorSelectionAndFocusPolicy(
         focusNode: _focusNode,
         selection: _composer.selectionNotifier,
+        documentLayoutKey: _docLayoutKey,
```
matthew-carroll, line 510: Based on my suggestion above, this line can become:

```dart
isDocumentLayoutAvailable: () => _docLayoutKey.currentContext != null,
```

super_editor/lib/src/default_editor/document_focus_and_selection_policies.dart
angelosilvestre: Updated.

matthew-carroll (approved):
LGTM

PR #1017, opened 2023-02-03
[SuperEditor] Fix performAction override for TextInputAction.newline (Resolves #1012)

### LINKED ISSUE #1012: [SuperEditor] [Android] Overriding performAction has no effect for newline
### Description
When providing an `imeConfiguration` setting newline as the `keyboardActionButton`, and also providing `imeOverrides` for `performAction`, such override will only be invoked if InputAction is not `newLine`.

### Expected behavior
I would expect to have the possibility to fully override the action being perform regardless of the keyboardActionButton that is provided.

### Minimal Example on demo_mobile_editing_android.dart

```

import 'keyboard_overlay_clipper.dart';

class _DeltaTextInputClientDecorator extends DeltaTextInputClientDecorator {
  @override
    print('performAction invoked');

    super.performAction(action);
  }
}

/// Mobile Android document editing demo.
///
/// This demo forces the editor into a mobile configuration,
/// no matter which platform or form factor you use.
class MobileEditingAndroidDemo extends StatefulWidget {
  @override
  _MobileEditingAndroidDemoState createState() =>
      _MobileEditingAndroidDemoState();
}

class _MobileEditingAndroidDemoState extends State<MobileEditingAndroidDemo> {
  final GlobalKey _docLayoutKey = GlobalKey();

  late Document _doc;
  late DocumentEditor _docEditor;
  late DocumentComposer _composer;
  late CommonEditorOperations _docOps;

  FocusNode? _editorFocusNode;

  @override
  void initState() {
    super.initState();
    _doc = _createInitialDocument();
    _docEditor = DocumentEditor(document: _doc as MutableDocument);
    _composer = DocumentComposer();
    _docOps = CommonEditorOperations(
      editor: _docEditor,
      composer: _composer,
      documentLayoutResolver: () =>
          _docLayoutKey.currentState as DocumentLayout,
    );
    _editorFocusNode = FocusNode();
  }

  @override
  void dispose() {
    _editorFocusNode!.dispose();
    _composer.dispose();
    super.dispose();
  }

  @override
    return SafeArea(
      child: Column(
        children: [
          Expanded(
            child: SuperEditor(
              focusNode: _editorFocusNode,
              documentLayoutKey: _docLayoutKey,
              editor: _docEditor,
              composer: _composer,
              imeOverrides: _DeltaTextInputClientDecorator(),
              imeConfiguration: SuperEditorImeConfiguration(
                keyboardActionButton: TextInputAction.newline,
              ),
              gestureMode: DocumentGestureMode.android,
              androidToolbarBuilder: (_) => AndroidTextEditingFloatingToolbar(
                onCutPressed: () => _docOps.cut(),
                onCopyPressed: () => _docOps.copy(),
                onPastePressed: () => _docOps.paste(),
                onSelectAllPressed: () => _docOps.selectAll(),
              ),
                documentPadding: const EdgeInsets.all(16),
              ),
              createOverlayControlsClipper: (_) =>
                  const KeyboardToolbarClipper(),
            ),
          ),
          MultiListenableBuilder(
            listenables: <Listenable>{
              _doc,
              _composer.selectionNotifier,
            },
            builder: (_) => SizedBox.shrink(),
          ),
        ],
      ),
    );
  }

  MutableDocument _createInitialDocument() {
    return MutableDocument(
      nodes: [
        ParagraphNode(
          text: AttributedText(
            text:
                'Etiam id lacus interdum, efficitur ex convallis, accumsan ipsum. Integer faucibus mollis mauris, a suscipit ante mollis vitae. Fusce justo metus, congue non lectus ac, luctus rhoncus tellus. Phasellus vitae fermentum orci, sit amet sodales orci. Fusce at ante iaculis nunc aliquet pharetra. Nam placerat, nisl in gravida lacinia, nisl nibh feugiat nunc, in sagittis nisl sapien nec arcu. Nunc gravida faucibus massa, sit amet accumsan dolor feugiat in. Mauris ut elementum leo.',
          ),
        ),
      ],
    );
  }
}

```

I would expect
` print('performAction invoked');`
to be invoked. However this only happens if I set something different as the keyboardAction

angelosilvestre:
[SuperEditor] Fix `performAction` override for `TextInputAction.newline`. Resolves #1012

When `SuperEditor` is configured to use `TextInputAction.newline`,  pressing the action button doesn't call `performAction` in the given `imeOverrides`. 

The cause is that, on Android, pressing an action button configured as `TextInputAction.newline` generates an insertion delta of `\n` instead of calling `performAction`. `TextDeltasDocumentEditor` translates the insertion into a `performAction` call. However, the call is handled inside `TextDeltasDocumentEditor` as it doesn't know about the overrides.

This PR changes `TextDeltasDocumentEditor` to accept a new property to override `performAction`.

### DIFF
--- a/super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
+++ b/super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
@@ -15,12 +15,16 @@ class TextDeltasDocumentEditor {
     required this.selection,
     required this.composingRegion,
     required this.commonOps,
+    this.onPerformAction,
   });
 
   final DocumentEditor editor;
   final ValueNotifier<DocumentSelection?> selection;
   final ValueNotifier<DocumentRange?> composingRegion;
   final CommonEditorOperations commonOps;
+
+  /// Override the handling of a [TextInputAction]s.
+  final void Function(TextInputAction action)? onPerformAction;
 
   /// Applies the given [textEditingDeltas] to the [Document].
   void applyDeltas(List<TextEditingDelta> textEditingDeltas) {
@@ -84,7 +88,7 @@ class TextDeltasDocumentEditor {
       // we forward the newline action to performAction.
       if (defaultTargetPlatform == TargetPlatform.android || kIsWeb) {
         editorImeLog.fine("Received a newline insertion on Android. Forwarding to newline input action.");
-        performAction(TextInputAction.newline);
+        _dispatchTextInputAction(TextInputAction.newline);
       } else {
         editorImeLog.fine("Skipping insertion delta because its a newline");
       }
@@ -128,7 +132,7 @@ class TextDeltasDocumentEditor {
       // we forward the newline action to performAction.
       if (defaultTargetPlatform == TargetPlatform.android || kIsWeb) {
         editorImeLog.fine("Received a newline replacement on Android. Forwarding to newline input action.");
-        performAction(TextInputAction.newline);
+        _dispatchTextInputAction(TextInputAction.newline);
       } else {
         editorImeLog.fine("Skipping replacement delta because its a newline");
       }
@@ -212,7 +216,7 @@ class TextDeltasDocumentEditor {
     editorImeLog.fine('With text: "$replacementText"');
 
     if (replacementText == "\n") {
-      performAction(TextInputAction.newline);
+      _dispatchTextInputAction(TextInputAction.newline);
       return;
     }
 
@@ -276,6 +280,14 @@ class TextDeltasDocumentEditor {
       case TextInputAction.unspecified:
         editorImeLog.warning("User pressed unhandled action button: $action");
         break;
+    }
+  }
+
+  void _dispatchTextInputAction(TextInputAction action) {
+    if (onPerformAction != null) {
+      onPerformAction!(action);
+    } else {
+      performAction(action);
     }
   }
 }
--- a/super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
+++ b/super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
@@ -23,7 +23,13 @@ class TextDeltasDocumentEditor {
   final ValueNotifier<DocumentRange?> composingRegion;
   final CommonEditorOperations commonOps;
 
-  /// Override the handling of a [TextInputAction]s.
+  /// Callback invoked when the client should handle a given [TextInputAction].
+  ///
+  /// Typically, [TextInputAction]s are reported through a different part of Flutter's IME
+  /// API. However, some actions, on some platforms, are converted into text editing
+  /// deltas, rather than reported as explicit actions. For example, on Android, the `newline`
+  /// action is reported as a text insertion with a `\n` character. That change is intercepted
+  /// by this editor and reported as a [TextInputAction.newline] instead.
   final void Function(TextInputAction action)? onPerformAction;
 
   /// Applies the given [textEditingDeltas] to the [Document].
--- a/super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
+++ b/super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
@@ -15,7 +15,7 @@ class TextDeltasDocumentEditor {
     required this.selection,
     required this.composingRegion,
     required this.commonOps,
-    this.onPerformAction,
+    required this.onPerformAction,
   });
 
   final DocumentEditor editor;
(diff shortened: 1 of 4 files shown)


super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
```diff
@@ -15,13 +15,17 @@ class TextDeltasDocumentEditor {
...
     required this.commonOps,
+    this.onPerformAction,
   });
 
   final DocumentEditor editor;
   final ValueNotifier<DocumentSelection?> selection;
   final ValueNotifier<DocumentRange?> composingRegion;
   final CommonEditorOperations commonOps;
 
+  /// Override the handling of a [TextInputAction]s.
```
matthew-carroll, line 26: Please change the dart doc to the following:

```
/// Callback invoked when the client should handle a given [TextInputAction].
///
/// Typically, [TextInputAction]s are reported through a different part of Flutter's IME
/// API. However, some actions, on some platforms, are converted into text editing
/// deltas, rather than reported as explicit actions. For example, on Android, the `newline`
/// action is reported as a text insertion with a `\n` character. That change is intercepted
/// by this editor and reported as a [TextInputAction.newline] instead.
```

super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
```diff
@@ -278,4 +282,12 @@ class TextDeltasDocumentEditor {
         break;
     }
   }
+
+  void _dispatchTextInputAction(TextInputAction action) {
```
matthew-carroll, line 286: Let's get rid of this method, and the existing `performAction()` implementation. Please move the existing newline action behavior such that it hooks into our general action handling system, instead of us maintaining a special implementation in this class just for that one action.

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
```diff
@@ -131,12 +131,14 @@ class SuperEditorImeInteractorState extends State<SuperEditorImeInteractor> impl
   void initState() {
     super.initState();
     _focusNode = (widget.focusNode ?? FocusNode());
+    _imeClient = DeltaTextInputClientDecorator();
```
matthew-carroll, line 134: You should keep the `_configureImeClientDecorators()` with this line. Otherwise it begs the question, what are we decorating?

super_editor/test/src/default_editor/document_input_ime_test.dart
```diff
@@ -126,6 +126,47 @@ void main() {
       expect(document.nodes.length, 1);
     });
 
+    testWidgetsOnAndroid('allows apps to handle performAction in their own way upon new line insertion',
```
matthew-carroll, line 129: This test name seems overly verbose: `allows app to handle newline action`

super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
angelosilvestre: `TextDeltasDocumentEditor` is the only place where we handle the actions in SuperEditor. `DocumentImeInputClient` just forwads the `performAction` call to this class.

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
angelosilvestre: It seems we can move this line back to where it was.

matthew-carroll:
Let's try this...

In the deltas document editor, change the existing `performAction()` method to be `insertNewline()` and only include the implementation for inserting a new line. Also, have the deltas document editor always call out to the `onPerformAction` callback that it's given. Make the `onPerformAction` callback required.

In the document ime client, change its implementation of `performAction()` to check for the newline action and then conditionally invoke `textDeltasDocumentEditor.insertNewline()`.

With these changes, we should see one of the following situations...

Normal:
```
IME -> _imeClient.performAction(action)
_imeClient.performAction -> TextDeltasDocumentEditor.insertNewline()
```

Normal with overrides:
```
IME -> _imeClient.performAction(action)
_imeClient.performAction -> MyOverrides.performAction()
```

Android "/n":
```
IME -> _imeClient.updateEditingValueWithDeltas()
_imeClient.updateEditingValueWithDeltas -> onPerformAction(action)
onPerformAction -> _imeClient.performAction(action)
_imeClient.performAction -> TextDeltasDocumentEditor.insertNewline()
```

Android "/n" with overrides:
```
IME -> _imeClient.updateEditingValueWithDeltas()
_imeClient.updateEditingValueWithDeltas -> onPerformAction(action)
onPerformAction -> _imeClient.performAction(action)
_imeClient.performAction -> MyOverrides.performAction()
```

With these changes we can always think of `performAction()` as handling IME actions. The only difference is how that method internally delegates to others.

angelosilvestre:
@matthew-carroll Updated.

super_editor/lib/src/default_editor/document_ime/document_ime_communication.dart
```diff
@@ -278,7 +278,17 @@ class DocumentImeInputClient extends TextInputConnectionDecorator with TextInput
   @override
   void performAction(TextInputAction action) {
     editorImeLog.fine("IME says to perform action: $action");
-    textDeltasDocumentEditor.performAction(action);
+    switch (action) {
```
matthew-carroll, line 281: I don't think we need this switch statement. I think we can just use an if-statement, because that's all this client cares about.

angelosilvestre: Updated.

matthew-carroll (approved):
LGTM

PR #1019, opened 2023-02-07
[SuperEditor] Fix image insertion at the middle of a paragraph (Resolves #1018)

### LINKED ISSUE #1018: [SuperEditor] Cannot add an ImageNode to the middle of a ParagraphNode
If a `ParagraphNode` is empty, I can add an `ImageNode` without a problem.

If a `ParagraphNode` contains text (and the selection is anywhere within the node) and an `ImageNode` is added, the following error occurs:

```
════════ Exception caught by scheduler library ═════════════════════════════════
The following _Exception was thrown during a scheduler callback:
Exception: Expected nodePosition of type UpstreamDownstreamNodePosition but received: TextPosition(offset: 0, affinity: TextAffinity.downstream)

#0      _BoxComponentState.getRectForPosition
box_component.dart:187
#1      _SingleColumnDocumentLayoutState.getRectForPosition
_layout.dart:220
#2      _IOSDocumentTouchInteractorState._positionCaret
document_gestures_touch_ios.dart:879
#3      _IOSDocumentTouchInteractorState._updateHandlesAfterSelectionOrLayoutChange
document_gestures_touch_ios.dart:373
#4      _IOSDocumentTouchInteractorState._onDocumentChange.<anonymous closure>
document_gestures_touch_ios.dart:343
#5      SchedulerBinding._invokeFrameCallback
binding.dart:1289
#6      SchedulerBinding.handleDrawFrame
binding.dart:1227
#7      SchedulerBinding._handleDrawFrame
binding.dart:1076
#8      _invoke (dart:ui/hooks.dart:145:13)
#9      PlatformDispatcher._drawFrame (dart:ui/platform_dispatcher.dart:338:5)
#10     _drawFrame (dart:ui/hooks.dart:112:31)

════════ Exception caught by scheduler library ═════════════════════════════════
Exception: Expected nodePosition of type UpstreamDownstreamNodePosition but received: TextPosition(offset: 0, affinity: TextAffinity.downstream)

════════ Exception caught by widgets library ═══════════════════════════════════
Multiple widgets used the same GlobalKey.
```

I've modified the [example editor](https://github.com/superlistapp/super_editor/blob/main/super_editor/example/lib/demos/example_editor/example_editor.dart) to add an "add image" button to the standard `KeyboardEditingToolbar` which calls the following method:

```
  void _addImage() {
    commonOps.insertImage('https://storage.googleapis.com/cms-storage-bucket/d406c736e7c4c57f5f61.png');
  }
```

angelosilvestre:
[SuperEditor] Fix image insertion at the middle of a paragraph. Resolves #1018 

When placing the caret at the middle of a paragraph and trying to insert an image, the following exception is thrown:

```
Exception: Expected nodePosition of type UpstreamDownstreamNodePosition but received: TextPosition(offset: 0, affinity: TextAffinity.downstream)
```

The cause is that, when inserting a block node, we are using the same node id as the selected paragraph.

This PR changes the insertion of images and horizontal rules to use a new node id.

During tests, I found other issues.

* The `_insertBlockLevelContent` dart doc has this sentence: 
```dart
/// If the selection extent sits in an empty paragraph, that paragraph
/// is converted into the given [blockNode] and a new empty paragraph
/// is inserted after the [blockNode].
```
This isn't the case, as a new empty paragraph isn't added after the node.

This PR implements the described behavior.

* To handle the image insertion at the end of a paragraph we are doing this check:

`if (paragraphPosition == endOfParagraph)`

This comparison may return `false` as it compares the text affinity. Sometimes we get an upstream selection when placing the caret at the end of a line.

This PR changes this check to compare only the offsets. It also changes the implementation of this if branch to add an empty paragraph after the inserted node, as this is the behavior described in the dart doc.

### DIFF
--- a/super_editor/lib/src/default_editor/common_editor_operations.dart
+++ b/super_editor/lib/src/default_editor/common_editor_operations.dart
@@ -1815,7 +1815,17 @@ class CommonEditorOperations {
       return false;
     }
 
-    return _insertBlockLevelContent(ImageNode(id: DocumentEditor.createNodeId(), imageUrl: url));
+    final node = editor.document.getNodeById(composer.selection!.base.nodeId);
+    if (node is! ParagraphNode) {
+      return false;
+    }
+
+    // When the selected node is empty, we convert it to an ImageNode.
+    final nodeId = node.text.text.isEmpty //
+        ? composer.selection!.base.nodeId
+        : DocumentEditor.createNodeId();
+
+    return _insertBlockLevelContent(ImageNode(id: nodeId, imageUrl: url));
   }
 
   /// Inserts horizontal rule at the current selection extent.
@@ -1844,7 +1854,17 @@ class CommonEditorOperations {
       return false;
     }
 
-    return _insertBlockLevelContent(HorizontalRuleNode(id: DocumentEditor.createNodeId()));
+    final node = editor.document.getNodeById(composer.selection!.base.nodeId);
+    if (node is! ParagraphNode) {
+      return false;
+    }
+
+    // When the selected node is empty, we convert it to a HorizontalRuleNode.
+    final nodeId = node.text.text.isEmpty //
+        ? composer.selection!.base.nodeId
+        : DocumentEditor.createNodeId();
+
+    return _insertBlockLevelContent(HorizontalRuleNode(id: nodeId));
   }
 
   /// Inserts the given [blockNode] after the caret.
--- a/super_editor/lib/src/default_editor/common_editor_operations.dart
+++ b/super_editor/lib/src/default_editor/common_editor_operations.dart
@@ -1806,7 +1806,7 @@ class CommonEditorOperations {
   ///
   /// If the selection extent sits in any other kind of node, nothing happens.
   ///
-  /// Returns [true] if an image was inserted, [false] if it wasn't.
+  /// Returns `true` if an image was inserted, `false` if it wasn't.
   bool insertImage(String url) {
     if (composer.selection == null) {
       return false;
@@ -1815,8 +1815,7 @@ class CommonEditorOperations {
       return false;
     }
 
-    final nodeId = composer.selection!.base.nodeId;
-    return _insertBlockLevelContent(ImageNode(id: nodeId, imageUrl: url));
+    return _insertBlockLevelContent(ImageNode(id: DocumentEditor.createNodeId(), imageUrl: url));
   }
 
   /// Inserts horizontal rule at the current selection extent.
@@ -1845,8 +1844,7 @@ class CommonEditorOperations {
       return false;
     }
 
-    final nodeId = composer.selection!.base.nodeId;
-    return _insertBlockLevelContent(HorizontalRuleNode(id: nodeId));
+    return _insertBlockLevelContent(HorizontalRuleNode(id: DocumentEditor.createNodeId()));
   }
 
   /// Inserts the given [blockNode] after the caret.
@@ -1866,7 +1864,7 @@ class CommonEditorOperations {
   ///
   /// If the selection extent sits in any other kind of node, nothing happens.
   ///
-  /// Returns [true] if the [blockNode] was inserted, [false] if it wasn't.
+  /// Returns `true` if the [blockNode] was inserted, `false` if it wasn't.
   bool _insertBlockLevelContent(DocumentNode blockNode) {
     if (composer.selection == null) {
       return false;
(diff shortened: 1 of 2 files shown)


matthew-carroll:
> When placing the caret at the middle of a paragraph and trying to insert an image

What does it mean to insert an image in the middle of a paragraph?

angelosilvestre:
> > When placing the caret at the middle of a paragraph and trying to insert an image
> 
> What does it mean to insert an image in the middle of a paragraph?

It means calling `commonOps.insertImage()` when the caret sits at the middle of a paragraph. It should split the paragraph into two and insert an image node between the two nodes.

matthew-carroll:
> It means calling commonOps.insertImage() when the caret sits at the middle of a paragraph.

Who is doing that? The example app doesn't do that, does it? From a UX perspective, how does one "insert an image in the middle of a paragraph"?

Jethro87:
@matthew-carroll Here are a few images showing how images are added in our app and what is expected.

IMG_1168

IMG_1169

IMG_1170

The `commonOps` function `insertImage` calls `_insertBlockLevelContent` and passes in the new `ImageNode`. `_insertBlockLevelContent` contains this description:

```
  /// Inserts the given [blockNode] after the caret.
  ///
  /// If the selection extent sits in an empty paragraph, that paragraph
  /// is converted into the given [blockNode] and a new empty paragraph
  /// is inserted after the [blockNode].
  ///
  /// If the selection extent sits at the end of a paragraph, the [blockNode]
  /// is inserted as a new node after that paragraph, and then a new
  /// empty paragraph is inserted after the [blockNode].
  ///
  /// If the selection extent sits in the middle of a paragraph then the
  /// paragraph is split into two at that position, the [blockNode] is
  /// inserted between the two paragraphs, the selection extent is placed
  /// at the beginning of the second paragraph.
  ///
  /// If the selection extent sits in any other kind of node, nothing happens.
  ///
  /// Returns [true] if the [blockNode] was inserted, [false] if it wasn't.
```

So it seems to me that the behavior shown in the images above is expected. Instead, an exception is thrown as per #1018 and Angelo's comments above.

matthew-carroll:
Weird. I have no memory of doing that. But that's what the dart doc says.

super_editor/lib/src/default_editor/common_editor_operations.dart
```diff
@@ -1815,8 +1815,7 @@ class CommonEditorOperations {
       return false;
     }
 
-    final nodeId = composer.selection!.base.nodeId;
-    return _insertBlockLevelContent(ImageNode(id: nodeId, imageUrl: url));
+    return _insertBlockLevelContent(ImageNode(id: DocumentEditor.createNodeId(), imageUrl: url));
```
matthew-carroll, line 1818: If we're converting a paragraph into another node type, then keeping the existing node ID is acceptable, right?

super_editor/test/super_editor/supereditor_content_insertion_test.dart
```diff
@@ -0,0 +1,270 @@
...
+            .withAddedComponents(
+          [const FakeImageComponentBuilder(size: Size(100, 100))],
+        ).pump();
+
+        // Place caret at "Before the image| after the image".
+        await tester.placeCaretInParagraph(context.editContext.editor.document.nodes.first.id, 16);
+
+        // Insert the image at the current selection.
+        context.editContext.commonOps.insertImage('http://image.fake');
+        await tester.pumpAndSettle();
```
matthew-carroll, line 34: Does this need to be a settle?

super_editor/lib/src/default_editor/common_editor_operations.dart
angelosilvestre: Do you mean we should add logic to `insertImage` and `insertHorizontalRule` to reuse the ID when the selection is at an empty paragraph and create a new one when it's not ?

super_editor/test/super_editor/supereditor_content_insertion_test.dart
angelosilvestre: It seems it doesn't. Updated.

super_editor/lib/src/default_editor/common_editor_operations.dart
angelosilvestre: But this is wrong for the cases where we insert the image at the middle or end of a non-empty paragraph, because we are creating a new node with the same id as the current selected node. We end up with two nodes with the same id.

matthew-carroll: Right, so we should choose the node ID based on which situation we're dealing with. Is there not enough information available in the command to tell the difference?

angelosilvestre: I think we have two options:
1. Check if the selected node is empty and choose the id based on that. We would need to change both `insertImage` and `insertHorizontalRule`.
2. Change `_insertBlockLevelContent` to receive a "node builder" and let `_insertBlockLevelContent` decide the node ID. For example:
```dart
bool _insertBlockLevelContent(DocumentNode Function(String id) nodeBuilder)
```

matthew-carroll: Why do we need to provide a node builder?

angelosilvestre: If we let `_insertBlockLevelContent` decide the ID, then we wouldn't know the ID until  `_insertBlockLevelContent` is called.

Or we can just check if the node is empty in `insertImage` and `insertHorizontalRule`...

matthew-carroll: Let's go the simpler route and make the decision in `insertImage` and `insertHorizontalRule`

angelosilvestre: Updated.

super_editor/lib/src/default_editor/common_editor_operations.dart
```diff
@@ -1815,7 +1815,16 @@ class CommonEditorOperations {
       return false;
     }
 
-    final nodeId = composer.selection!.base.nodeId;
+    final node = editor.document.getNodeById(composer.selection!.base.nodeId);
+    if (node is! ParagraphNode) {
+      return false;
+    }
+
+    // When the selected node is empty, we convert it to an ImageNode.
```
matthew-carroll, line 1823: This comment sounds like it's out of date with the new implementation

super_editor/lib/src/default_editor/common_editor_operations.dart
```diff
@@ -1845,7 +1854,16 @@ class CommonEditorOperations {
       return false;
     }
 
-    final nodeId = composer.selection!.base.nodeId;
+    final node = editor.document.getNodeById(composer.selection!.base.nodeId);
+    if (node is! ParagraphNode) {
+      return false;
+    }
+
+    // When the selected node is empty, we convert it to a HorizontalRuleNode.
```
matthew-carroll, line 1862: same with this comment

super_editor/lib/src/default_editor/common_editor_operations.dart
angelosilvestre: But when the selected node is empty we do "convert" it by replacing it with an ImageNode with the same Id as the existing node...

matthew-carroll: That comment appears before a ternary expression. It's misleading to say that one thing happens, when the statement that follows might do something else. At a minimum, the comment should reflect both possibilities.

angelosilvestre: Updated.

super_editor/lib/src/default_editor/common_editor_operations.dart
angelosilvestre: Updated.

PR #1023, opened 2023-02-09
[SuperEditor] Fix caret display when document layout changes due to animations (Resolves #1022)

### LINKED ISSUE #1022: [SuperEditor] Caret overlay is not updated when document layout changes due to animations
Steps to reproduce:
* Run the demo app
* Change the `build` method of `TaskComponent` so that it will animate in a subtitle when the node has selection:
```dart
  @override
    return Column(
      children: [
        Row(
          children: [
            Padding(
              padding: const EdgeInsets.only(left: 16, right: 4),
              child: Checkbox(
                value: widget.viewModel.isComplete,
                onChanged: (newValue) {
                  widget.viewModel.setComplete(newValue!);
                },
              ),
            ),
            Expanded(
              child: TextComponent(
                key: _textKey,
                text: widget.viewModel.text,
                  // Show a strikethrough across the entire task if it's complete.
                  final style = widget.viewModel.textStyleBuilder(attributions);
                  return widget.viewModel.isComplete
                      ? style.copyWith(
                          decoration: style.decoration == null
                              ? TextDecoration.lineThrough
                              : TextDecoration.combine([
                                  TextDecoration.lineThrough,
                                  style.decoration!
                                ]),
                        )
                      : style;
                },
                textSelection: widget.viewModel.selection,
                selectionColor: widget.viewModel.selectionColor,
                highlightWhenEmpty: widget.viewModel.highlightWhenEmpty,
                showDebugPaint: widget.showDebugPaint,
              ),
            ),
          ],
        ),
        AnimatedSize(
          child: widget.viewModel.selection != null
              ? const SizedBox(
                  height: 40,
                  child: Text('some metadata'),
                )
              : const SizedBox.shrink(),
        ),
      ],
    );
  }
```
* Move the selection downwards direction

Expected: the caret is painted in the correct place
Actual: the caret is painted where it was expected to be, if there was no subtitle

https://user-images.githubusercontent.com/31278849/217527882-2707112d-5c3b-478e-91d7-f6c2231085c2.mov

angelosilvestre:
[SuperEditor] Fix caret display when document layout changes due to animations. Resolves #1022

When we change the selection between components that change their sizes, the caret ends up at an incorrect position:

https://user-images.githubusercontent.com/31278849/217527882-2707112d-5c3b-478e-91d7-f6c2231085c2.mov

The cause is that when we change the selection, the caret offset is computed immediately and isn't updated again. After the animation ends, the selected component sits at a different offset but the caret is still positioned at the old offset.

This PR changes the interactors and the caret overlay to schedule caret updates until we get the same offset that we have.

This isn't a perfect solution, we might need a way for the components to signal that their layout changed.

### DIFF
--- a/super_editor/lib/src/core/edit_context.dart
+++ b/super_editor/lib/src/core/edit_context.dart
@@ -1,6 +1,5 @@
 import 'package:super_editor/src/core/document.dart';
 import 'package:super_editor/src/default_editor/common_editor_operations.dart';
-import 'package:super_text_layout/super_text_layout.dart';
 
 import 'document_composer.dart';
 import 'document_editor.dart';
@@ -21,9 +20,9 @@ class EditContext {
   EditContext({
     required this.editor,
     required DocumentLayout Function() getDocumentLayout,
+    required this.isDocumentLayoutAvailable,
     required this.composer,
     required this.commonOps,
-    required this.componentSizeNotifier,
   }) : _getDocumentLayout = getDocumentLayout;
 
   /// The editor of the [Document] that allows executing commands that alter the
@@ -36,6 +35,11 @@ class EditContext {
   DocumentLayout get documentLayout => _getDocumentLayout();
   final DocumentLayout Function() _getDocumentLayout;
 
+  /// Returns whether or not we can access the document layout.
+  ///
+  /// When this method returns `true`, we assume it's safe to access the [documentLayout].
+  final bool Function() isDocumentLayoutAvailable;
+
   /// The [DocumentComposer] that maintains selection and attributions to work
   /// in conjunction with the [editor] to apply changes to the document.
   final DocumentComposer composer;
@@ -44,6 +48,8 @@ class EditContext {
   /// the document.
   final CommonEditorOperations commonOps;
 
-  /// A [ChangeNotifier] that is triggered whenever a component changes its size.
-  final SignalListenable componentSizeNotifier;
+  /// Holds the [LayerLink]s linked to the caret and handle positions.
+  ///
+  /// Use a [CompositedTransformFollower] to display widgets at these positions.
+  LayoutLinks get layoutLinks => _getDocumentLayout().layerLinks;
 }
--- a/super_editor/lib/src/core/edit_context.dart
+++ b/super_editor/lib/src/core/edit_context.dart
@@ -1,5 +1,6 @@
 import 'package:super_editor/src/core/document.dart';
 import 'package:super_editor/src/default_editor/common_editor_operations.dart';
+import 'package:super_text_layout/super_text_layout.dart';
 
 import 'document_composer.dart';
 import 'document_editor.dart';
@@ -22,6 +23,7 @@ class EditContext {
     required DocumentLayout Function() getDocumentLayout,
     required this.composer,
     required this.commonOps,
+    required this.componentSizeNotifier,
   }) : _getDocumentLayout = getDocumentLayout;
 
   /// The editor of the [Document] that allows executing commands that alter the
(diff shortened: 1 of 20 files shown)


matthew-carroll:
Let's try to solve the root problem instead of the symptom. Where and how might a component notify the appropriate party of a change to layout?

angelosilvestre:
@matthew-carroll I implemented a possible solution, listening to component size changes via notifications and signaling the changes to other parts of the system.

super_editor/lib/src/core/edit_context.dart
```diff
@@ -22,6 +23,7 @@ class EditContext {
     required DocumentLayout Function() getDocumentLayout,
     required this.composer,
     required this.commonOps,
+    required this.componentSizeNotifier,
```
matthew-carroll, line 26: This is definitely way too high of a level to add such a deep implementation detail. We'll need to figure out a different approach.

matthew-carroll:
Is the caret overlay the exact same size as the document layout? Also, does the overall document layout change size when a component changes size?

If so, can we just add a `LayoutBuilder` inside the caret display, and any time the constraints change, we schedule a caret update?

angelosilvestre:
> Also, does the overall document layout change size when a component changes size?

It may not, in the sample code attached to the ticket, the document layout size only changes when we move from "no tasks selected" -> "a task selected" or from "a task selected" -> "no task selected".

When we already have a task selected and tap at another task, the layout size remains the same.

super_editor/lib/src/core/edit_context.dart
angelosilvestre: We could change `DocumentLayerBuilder` to pass the notifier to the layers. The current declaration is:

```dart
abstract class DocumentLayerBuilder {
  Widget build(BuildContext context, EditContext editContext);
}
```

We could add another parameter to the build method.

matthew-carroll: What if we have the `DocumentLayout` add an invisible widget at the selection position and then user a `CompositedTransformFollower` for the caret? Do you see any reason why that wouldn't work? If not, do you want to give that a try?

angelosilvestre:
@matthew-carroll I updated the PR to use  `CompositedTransformFollower`s to position the caret and drag handles. The toolbar positioning wasn't changed yet.

Could you take another look at this PR?

miguelcmedeiros:
@angelosilvestre 
With the proposed solution, does it imply that a super_editor client won't be able to create its own overlay builder that is able to rebuild once the size of a component changes?

In our case we have one that paints a blue line between 2 consecutive nodes and when the user taps on it, a new node is created.

angelosilvestre:
@miguelcmedeiros The solution using a `CompositedTransformFollower` is limited to caret and drag handles display.

As we are not calling any listener, overlays cannot know when the size of a component changes. Maybe we could use this solution but also provide a way to notify when the size of a component changes.

matthew-carroll:
Reacting to size changes in `RenderObject`s is really part of the core Flutter pipeline. If we think we need that, then I think we should look at a deeper solution than passing a callback or a controller around....

That said, moving these overlay things to use followers is something we should probably do, anyway.

Also, we can expose the anchors from `SuperEditor` so that other implementations can do the same thing.

matthew-carroll:
@angelosilvestre can you remind me where we left this?

angelosilvestre:
@matthew-carroll I'd like it if you take a look at the code to see if this is the ideal approach and if we should do the same to the toolbar positioning.

matthew-carroll:
I'm going to close this PR in favor of work in https://github.com/superlistapp/super_editor/pull/1064

PR #1026, opened 2023-02-14
[SuperEditor] Expose TextComponentState

### LINKED ISSUE #1025: [SuperEditor] Need access to TextComponentState when implementing a proxy component that returns a TextComponent
In order to implement a `ProxyDocumentComponent` that returns a `TextComponent` (e.g. TextWithHintComponent), we need to define a global key for the child component that is typed as `TextComponentState`.

Although `TextComponentState` is not private, it is not exported by `super_editor`, and so, can't be used without silencing `invalid_use_of_visible_for_testing_member` lint.

miguelcmedeiros:
Fixes #1025

### DIFF
--- a/super_editor/lib/src/default_editor/text.dart
+++ b/super_editor/lib/src/default_editor/text.dart
@@ -439,7 +439,6 @@ class TextComponent extends StatefulWidget {
   TextComponentState createState() => TextComponentState();
 }
 
-@visibleForTesting
 class TextComponentState extends State<TextComponent> with DocumentComponent implements TextComposable {
   final _textKey = GlobalKey<ProseTextState>();
 


matthew-carroll:
@miguelcmedeiros can you please change the target to the master branch?

matthew-carroll:
It looks like you'll need to rebase this branch onto `main`.

We don't ever expect to merge directly into `stable` unless there's a change or fix that only applies to Flutter's stable release.

miguelcmedeiros:
My bad, didn't notice I was using `stable` on my machine and then somehow thought it was a good idea to select `stable` instead of `main` 🤷‍♂️

PR #1028, opened 2023-02-15
[SuperEditor] Make links tappable when in "interaction mode" (Resolves #1024)

### LINKED ISSUE #1024: [SuperEditor][SuperReader] - Make links clickable, and also other attributions
Make it possible for users to tap/click on links to open a link. Also make it possible to the same with any other attribution that the developer might choose.

matthew-carroll:
Make links tappable when in "interaction mode" (Resolves #1024)

This PR introduces the concept of an "interaction mode", which is a special mode that results in different user interaction behaviors compared to normal.

The interaction mode concept is added to the `DocumentComposer`, because the `DocumentComposer` represents transient UI modes.

The mouse gesture interactor adds the concept of a content delegate, which can express a desired mouse cursor when hovering over a piece of content, and also take action on single, double, or triple taps, instead of the mouse gesture interactor changing the document selection.

The mouse gesture content delegate is set to a link click delegate, by default.

The mouse gesture interactor needed to be updated to report key up events, so that we can turn off "interaction mode" when the CMD or CTRL key is released. This also requires that every existing keyboard handler check whether the event is "up" or "down". Most of the changes in this PR are those handler updates.

### DIFF
--- a/super_editor/example/lib/main.dart
+++ b/super_editor/example/lib/main.dart
@@ -35,11 +35,11 @@ import 'demos/supertextfield/android/demo_superandroidtextfield.dart';
 /// Demo of a basic text editor, as well as various widgets that
 /// are available in this package.
 Future<void> main() async {
-  initLoggers(Level.FINEST, {
+  initLoggers(Level.FINE, {
     // editorScrollingLog,
     // editorGesturesLog,
     // editorImeLog,
-    // editorKeyLog,
+    editorKeyLog,
     // editorOpsLog,
     // editorLayoutLog,
     // editorDocLog,
--- a/super_editor/example/lib/main.dart
+++ b/super_editor/example/lib/main.dart
@@ -35,11 +35,11 @@ import 'demos/supertextfield/android/demo_superandroidtextfield.dart';
 /// Demo of a basic text editor, as well as various widgets that
 /// are available in this package.
 Future<void> main() async {
-  initLoggers(Level.FINE, {
+  initLoggers(Level.FINEST, {
     // editorScrollingLog,
     // editorGesturesLog,
     // editorImeLog,
-    editorKeyLog,
+    // editorKeyLog,
     // editorOpsLog,
     // editorLayoutLog,
     // editorDocLog,
--- a/super_editor/example/lib/demos/example_editor/_toolbar.dart
+++ b/super_editor/example/lib/demos/example_editor/_toolbar.dart
@@ -61,7 +61,8 @@ class _EditorToolbarState extends State<EditorToolbar> {
   void initState() {
     super.initState();
     _urlFocusNode = FocusNode();
-    _urlController = SingleLineAttributedTextEditingController(_applyLink);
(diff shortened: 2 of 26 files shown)


matthew-carroll:
@jmatth - Here's a PR for tappable links

matthew-carroll:
@miguelcmedeiros @Jethro87 - This PR changes key reporting so that up and down keys are both reported. I've updated all of our built-in key handlers to ignore up events, but you'll need to make similar changes to your custom key handlers after this merges. Otherwise, you'll take two actions for every key that's pressed.

super_editor/example/lib/main.dart
```diff
@@ -35,11 +35,11 @@ import 'demos/supertextfield/android/demo_superandroidtextfield.dart';
 /// Demo of a basic text editor, as well as various widgets that
 /// are available in this package.
 Future<void> main() async {
-  initLoggers(Level.FINEST, {
+  initLoggers(Level.FINE, {
     // editorScrollingLog,
     // editorGesturesLog,
     // editorImeLog,
-    // editorKeyLog,
+    editorKeyLog,
```
angelosilvestre, line 42: This was intended to be committed?

angelosilvestre (approved):
LGTM

super_editor/example/lib/main.dart
matthew-carroll: Nope. If any changes are requested for this PR then I'll undo that. Otherwise we can change it back in a followup PR.

matthew-carroll:
@jmatth - Please let me know if you're happy with the current API and implementation.

jmatth:
@matthew-carroll The API looks good to me.

I see a couple issues in the example app:

- Creating a link without a protocol specified causes an error when clicking on it. It might be nice to have the default tap handler auto-add `http://` or `https://` if a protocol is not present in the link it's trying to launch. Not a huge problem overall though
- Opening a link via cmd-click leaves the editor stuck in interaction mode. I think the sequence of events is cmd keydown -> click link -> browser opens and takes focus -> cmd keyup goes to the now focused browser and Flutter doesn't see it. This isn't necessarily a blocker since I'm sure I could work around it in my application code, but it seems like a bigger problem for the "out of the box" super_editor experience

matthew-carroll:
@jmatth - I investigated focus change, but it looks like Flutter doesn't tell us we lost focus when the window focus changes.

I then tried https://pub.dev/packages/window_manager - but for some reason I'm not getting any of the window callbacks from that package.

Any idea how we can figure out when our window changes? I thought about turning interactive mode off when a link is clicked, but that's a bad UX for web, because on web tapping a link usually opens a tab and stays on the current tab.

matthew-carroll:
We spoke offline. I got the `window_manager` plugin working for the purpose of resetting interaction mode on window blur. However, we decided that plugin is too heavy-weight to integrate for this purpose.

Instead, I added a comment to the `DocumentComposer` that explains how to use the `window_manager` plugin to solve that problem in apps that need it.

PR #1030, opened 2023-02-16
[SuperEditor][iOS] Fix caret display when floating cursor is visible (Resolves #992)

### LINKED ISSUE #992: [SuperEditor] 2 carets are painted when long-press the SPACE key on iOS
Steps to reproduce:
* Using `stable` branch, open the demo app on an iOS device
* Place the selection in the text
* On the software keyboard, long press the SPACE key

Expected: show only floating cursor (probably that looks the same as the selection caret)
Actual: shows a floating cursor (red) and the selection caret (blue) that follows closely the floating cursor

https://user-images.githubusercontent.com/31278849/214280623-8f5d2455-6508-4a03-aebd-da78ec3d6fc9.MP4

angelosilvestre:
[SuperEditor][iOS] Fix caret display when floating cursor is visible. Resolves #992

When the iOS floating cursor is visible, we always display the "normal" caret, and the caret color isn't changed to grey, as can be seen in the video:

https://user-images.githubusercontent.com/31278849/214280623-8f5d2455-6508-4a03-aebd-da78ec3d6fc9.MP4

We already have some code to change the caret color to grey, but it isn't working.

This PR changes the iOS interactor to avoid displaying the caret when the floating cursor is over text or less than 30 pixels away from the text.

This PR also fixes the code to change the caret color to grey when the floating cursor is visible.

I had to use the `textLayout` property from `TextComponentState`, which is marked as `@visibleForTesting`, to check if the floating cursor is over text. I can change that if there is another way to do it.

### DIFF
--- a/super_editor/lib/src/infrastructure/platforms/ios/ios_document_controls.dart
+++ b/super_editor/lib/src/infrastructure/platforms/ios/ios_document_controls.dart
@@ -120,8 +120,15 @@ class _IosDocumentTouchEditingControlsState extends State<IosDocumentTouchEditin
   late BlinkController _caretBlinkController;
   Offset? _prevCaretOffset;
 
+  /// Represents the maximum distance an offset can be from the end of a text
+  /// and still be considered to be near it.
+  static const _maximumDistanceToBeNearText = 30.0;
+
   static const _defaultFloatingCursorHeight = 20.0;
+  static const _defaultFloatingCursorWidth = 2.0;
+
   final _isShowingFloatingCursor = ValueNotifier<bool>(false);
+  final _isFloatingCursorOverOrNearText = ValueNotifier<bool>(false);
   final _floatingCursorKey = GlobalKey();
   Offset? _initialFloatingCursorOffset;
   final _floatingCursorOffset = ValueNotifier<Offset?>(null);
@@ -177,6 +184,7 @@ class _IosDocumentTouchEditingControlsState extends State<IosDocumentTouchEditin
 
         _caretBlinkController.startBlinking();
 
+        _isFloatingCursorOverOrNearText.value = false;
         _initialFloatingCursorOffset = null;
         _floatingCursorOffset.value = null;
         _floatingCursorHeight = _defaultFloatingCursorHeight;
@@ -204,6 +212,7 @@ class _IosDocumentTouchEditingControlsState extends State<IosDocumentTouchEditin
     if (_floatingCursorOffset.value == null) {
       // The floating cursor just started.
       widget.onFloatingCursorStart?.call();
+      _isShowingFloatingCursor.value = true;
     }
 
     _caretBlinkController.stopBlinking();
@@ -218,19 +227,63 @@ class _IosDocumentTouchEditingControlsState extends State<IosDocumentTouchEditin
     if (nearestDocPosition.nodePosition is TextNodePosition) {
       final nearestComponent = widget.documentLayout.getComponentByNodeId(nearestDocPosition.nodeId)!;
       _floatingCursorHeight = nearestComponent.getRectForPosition(nearestDocPosition.nodePosition).height;
+
+      if (_isFloatingCursorOverText(nearestComponent)) {
+        _isFloatingCursorOverOrNearText.value = true;
+      } else {
+        final distance = _floatingCursorDistanceFromSelectedPosition(nearestComponent, nearestDocPosition.nodePosition);
+        _isFloatingCursorOverOrNearText.value = distance.dx.abs() < _maximumDistanceToBeNearText;
+      }
     } else {
       final nearestComponent = widget.documentLayout.getComponentByNodeId(nearestDocPosition.nodeId)!;
       _floatingCursorHeight = (nearestComponent.context.findRenderObject() as RenderBox).size.height;
+      _isFloatingCursorOverOrNearText.value = false;
     }
 
     widget.onFloatingCursorMoved?.call(_floatingCursorOffset.value!);
   }
 
+  bool _isFloatingCursorOverText(DocumentComponent nearestComponent) {
+    if (nearestComponent is! TextComponentState) {
+      return false;
+    }
+
+    final selectionBoxes = nearestComponent.textLayout.getBoxesForSelection(
+      TextSelection(
+        baseOffset: 0,
+        extentOffset: nearestComponent.getEndPosition().offset,
+      ),
+    );
+
+    final renderBox = nearestComponent.context.findRenderObject() as RenderBox;
+
+    final globalFloatingCursorOffset =
+        widget.documentLayout.getGlobalOffsetFromDocumentOffset(_floatingCursorOffset.value!);
+    final localFloatingCursorOffset = renderBox.globalToLocal(globalFloatingCursorOffset);
+    final floatingCursorRect = localFloatingCursorOffset & Size(_defaultFloatingCursorWidth, _floatingCursorHeight);
+
+    for (final box in selectionBoxes) {
+      if (box.toRect().overlaps(floatingCursorRect)) {
+        return true;
+      }
+    }
+
+    return false;
+  }
+
+  Offset _floatingCursorDistanceFromSelectedPosition(DocumentComponent component, NodePosition position) {
+    final renderBox = component.context.findRenderObject() as RenderBox;
+    final selectedPositionOffet = component.getOffsetForPosition(position);
+    final selectedPositionLocalOffet =
+        renderBox.localToGlobal(widget.documentLayout.getGlobalOffsetFromDocumentOffset(selectedPositionOffet));
+    return _floatingCursorOffset.value! - selectedPositionLocalOffet;
+  }
+
   @override
   Widget build(BuildContext context) {
-    return ListenableBuilder(
-        listenable: widget.editingController,
-        builder: (context, _) {
+    return MultiListenableBuilder(
+        listenables: {widget.editingController, _isFloatingCursorOverOrNearText},
+        builder: (context) {
           return Padding(
             // Remove the keyboard from the space that we occupy so that
             // clipping calculations apply to the expected visual borders,
@@ -247,7 +300,7 @@ class _IosDocumentTouchEditingControlsState extends State<IosDocumentTouchEditin
                 child: Stack(
                   children: [
                     // Build caret or drag handles
-                    ..._buildHandles(),
+                    if (!_isFloatingCursorOverOrNearText.value) ..._buildHandles(),
                     // Build the floating cursor
                     _buildFloatingCursor(),
                     // Build the editing toolbar
@@ -401,7 +454,7 @@ class _IosDocumentTouchEditingControlsState extends State<IosDocumentTouchEditin
           handleKey: _floatingCursorKey,
           handleOffset: floatingCursorOffset - Offset(0, _floatingCursorHeight / 2),
           handle: Container(
-            width: 2,
+            width: _defaultFloatingCursorWidth,
             height: _floatingCursorHeight,
             color: Colors.red.withOpacity(0.75),
           ),
--- a/super_editor/lib/src/infrastructure/platforms/ios/ios_document_controls.dart
+++ b/super_editor/lib/src/infrastructure/platforms/ios/ios_document_controls.dart
@@ -100,6 +100,13 @@ class IosDocumentTouchEditingControls extends StatefulWidget {
 
 class _IosDocumentTouchEditingControlsState extends State<IosDocumentTouchEditingControls>
     with SingleTickerProviderStateMixin {
+  /// Represents the maximum distance an offset can be from the end of a text
+  /// and still be considered to be near it.
+  static const _maximumDistanceToBeNearText = 30.0;
+
+  static const _defaultFloatingCursorHeight = 20.0;
+  static const _defaultFloatingCursorWidth = 2.0;
+
   // These global keys are assigned to each draggable handle to
   // prevent a strange dragging issue.
   //
@@ -119,13 +126,6 @@ class _IosDocumentTouchEditingControlsState extends State<IosDocumentTouchEditin
 
   late BlinkController _caretBlinkController;
   Offset? _prevCaretOffset;
-
-  /// Represents the maximum distance an offset can be from the end of a text
-  /// and still be considered to be near it.
-  static const _maximumDistanceToBeNearText = 30.0;
-
-  static const _defaultFloatingCursorHeight = 20.0;
-  static const _defaultFloatingCursorWidth = 2.0;
 
   final _isShowingFloatingCursor = ValueNotifier<bool>(false);
   final _isFloatingCursorOverOrNearText = ValueNotifier<bool>(false);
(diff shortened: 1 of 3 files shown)


matthew-carroll:
@angelosilvestre can you please include a video of the working version in this PR?

super_editor/lib/src/infrastructure/platforms/ios/ios_document_controls.dart
```diff
@@ -247,7 +300,7 @@ class _IosDocumentTouchEditingControlsState extends State<IosDocumentTouchEditin
                 child: Stack(
                   children: [
                     // Build caret or drag handles
-                    ..._buildHandles(),
+                    if (!_isFloatingCursorOverOrNearText.value) ..._buildHandles(),
```
matthew-carroll, line 303: There's a lot of stuff built in this tree. We should avoid rebuilding everything just because the floating cursor moved. Please add the `ListenableBuilder` within `_buildHandles()`.

super_editor/lib/src/infrastructure/platforms/ios/ios_document_controls.dart
```diff
@@ -204,6 +212,7 @@ class _IosDocumentTouchEditingControlsState extends State<IosDocumentTouchEditin
     if (_floatingCursorOffset.value == null) {
       // The floating cursor just started.
       widget.onFloatingCursorStart?.call();
+      _isShowingFloatingCursor.value = true;
```
matthew-carroll, line 215: Is there any reason that the concept of "should we show the cursor" is separated from "is the cursor near text"?

super_editor/lib/src/infrastructure/platforms/ios/ios_document_controls.dart
```diff
@@ -120,8 +120,15 @@ class _IosDocumentTouchEditingControlsState extends State<IosDocumentTouchEditin
   late BlinkController _caretBlinkController;
   Offset? _prevCaretOffset;
 
+  /// Represents the maximum distance an offset can be from the end of a text
+  /// and still be considered to be near it.
+  static const _maximumDistanceToBeNearText = 30.0;
```
matthew-carroll, line 125: Please place static things above instance members. I'm not sure why `_defaultFloatingCursorHeight` was placed down here, but it shouldn't have been.

super_editor/lib/src/infrastructure/platforms/ios/ios_document_controls.dart
```diff
@@ -218,19 +227,63 @@ class _IosDocumentTouchEditingControlsState extends State<IosDocumentTouchEditin
     if (nearestDocPosition.nodePosition is TextNodePosition) {
       final nearestComponent = widget.documentLayout.getComponentByNodeId(nearestDocPosition.nodeId)!;
       _floatingCursorHeight = nearestComponent.getRectForPosition(nearestDocPosition.nodePosition).height;
+
+      if (_isFloatingCursorOverText(nearestComponent)) {
```
matthew-carroll, line 231: Can the `_isFloatingCursorOverText()` method be replaced by a comparison to the `getRectForPosition()` up above? Seems like we already have the relevant rectangle, and we already know it corresponds to text, so all we need to do is calculate the distance from that rectangle to the floating cursor...

super_editor/test/super_editor/supereditor_floating_cursor_test.dart
```diff
@@ -0,0 +1,156 @@
...
+import 'package:flutter/services.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:super_editor/src/infrastructure/blinking_caret.dart';
+import 'package:super_editor/src/test/super_editor_test/supereditor_inspector.dart';
+import 'package:super_editor/src/test/super_editor_test/supereditor_robot.dart';
+
+import '../test_tools.dart';
+import 'document_test_tools.dart';
+
+void main() {
```
matthew-carroll, line 11: Do we not have any existing floating cursor tests?

super_editor/test/super_editor/supereditor_floating_cursor_test.dart
```diff
@@ -0,0 +1,156 @@
...
+        await tester.placeCaretInParagraph(SuperEditorInspector.findDocument()!.nodes.first.id, 0);
+
+        // Ensure the caret is displayed.
+        expect(_caretFinder(), findsOneWidget);
+
+        // Moves the floating cursor to a position that is over text.
+        final floatingCursor = _FloatingCursorSimulator();
+        await floatingCursor.start();
+        await tester.pump();
+        await floatingCursor.moveTo(const Offset(10, 0));
```
matthew-carroll, line 31: I think you should probably move further than this. We should probably move a distance that equates to multiple characters to the right, and we should also probably pump and settle at multiple locations along the way, and make sure at every stop that we're not showing the regular caret.

super_editor/test/super_editor/supereditor_floating_cursor_test.dart
```diff
@@ -0,0 +1,156 @@
...
+import 'package:super_editor/src/test/super_editor_test/supereditor_inspector.dart';
+import 'package:super_editor/src/test/super_editor_test/supereditor_robot.dart';
+
+import '../test_tools.dart';
+import 'document_test_tools.dart';
+
+void main() {
+  group('SuperEditor', () {
+    group('floating cursor', () {
+      testWidgetsOnIos('hides caret when over text (on iOS)', (tester) async {
```
matthew-carroll, line 14: You don't need "(on iOS)" for any of these tests because you're using `testWidgetsOnIos()`.

super_editor/test/super_editor/supereditor_floating_cursor_test.dart
```diff
@@ -0,0 +1,156 @@
...
+
+      testWidgetsOnIos('hides caret when near text (on iOS)', (tester) async {
+        await tester //
+            .createDocument()
+            .fromMarkdown('This is a paragraph')
+            .withEditorSize(const Size(300, 300))
+            .pump();
+
+        // Place caret at "This is a| paragraph".
+        // This is the last position of the first line.
```
matthew-carroll, line 53: Can you add this knowledge of text layout in a comment above `fromMarkdown()` in each of the tests?

super_editor/test/super_editor/supereditor_floating_cursor_test.dart
```diff
@@ -0,0 +1,156 @@
...
+        await tester.placeCaretInParagraph(SuperEditorInspector.findDocument()!.nodes.first.id, 9);
+
+        // Ensure the caret is displayed.
+        expect(_caretFinder(), findsOneWidget);
+
+        // Moves the floating cursor to a position that is close to the text.
+        final floatingCursor = _FloatingCursorSimulator();
+        await floatingCursor.start();
+        await tester.pump();
+        await floatingCursor.moveTo(const Offset(10, 0));
```
matthew-carroll, line 63: Let's move the maximum distance that we're supposed to be able to move.

super_editor/test/super_editor/supereditor_floating_cursor_test.dart
```diff
@@ -0,0 +1,156 @@
...
+            .pump();
+
+        // Place caret at "This is a paragraph|".
+        await tester.placeCaretInParagraph(SuperEditorInspector.findDocument()!.nodes.first.id, 9);
+
+        // Moves the floating cursor to a position that is far from text.
+        final floatingCursor = _FloatingCursorSimulator();
+        await floatingCursor.start();
+        await tester.pump();
+        await floatingCursor.moveTo(const Offset(60, 0));
```
matthew-carroll, line 91: Let's move 1 unit more than the maximum we're supposed to be able to move.

super_editor/test/super_editor/supereditor_floating_cursor_test.dart
```diff
@@ -0,0 +1,156 @@
...
+
+        // Ensure the caret is blue.
+        caret = tester.widget<BlinkingCaret>(_caretFinder());
+        expect(caret.color, Colors.blue);
+      });
+    });
+  });
+}
+
+Finder _caretFinder() {
```
matthew-carroll, line 116: Don't we already have a caret finder?

super_editor/test/super_editor/supereditor_floating_cursor_test.dart
```diff
@@ -0,0 +1,156 @@
...
+  return find.byType(BlinkingCaret);
+}
+
+class _FloatingCursorSimulator {
```
matthew-carroll, line 120: We should integrate this ability into the Super Editor robot. Please see what makes sense from an API perspective to make floating cursor control available to all Super Editor tests.

super_editor/lib/src/infrastructure/platforms/ios/ios_document_controls.dart
angelosilvestre: `_buildHandles()` returns a list of widgets, should `ListenableBuilder` return another `Stack` containing those widgets?

matthew-carroll: I would expect:

_buildHandles() {
  return ListenableBuilder(
    builder: (...) {
      return Stack(
        children: [
          // handles
        ],
      );
    },
  );
}

angelosilvestre:
https://github.com/superlistapp/super_editor/pull/1023 needs to land before updating this PR due to changes in `ios_document_controls.dart`.

matthew-carroll:
@angelosilvestre can this PR be updated now that #1023 is complete?

super_editor/test/super_editor/supereditor_floating_cursor_test.dart
angelosilvestre: Updated

super_editor/lib/src/infrastructure/platforms/ios/ios_document_controls.dart
angelosilvestre: Currently, "should we show the cursor" is determined by checking if we have `collapsedHandleOffset` set.

This value is updated in another widget.

super_editor/test/super_editor/supereditor_floating_cursor_test.dart
angelosilvestre: I didn't find any other floating cursor test.

super_editor/test/super_editor/supereditor_floating_cursor_test.dart
angelosilvestre: Do you mean in `SuperEditorInspector`?

super_editor/test/super_editor/supereditor_floating_cursor_test.dart
angelosilvestre: Added `startFloatingCursorGesture`, `updateFloatingCursorGesture` and `stopFloatingCursorGesture`.

Should these methods call `pump`?

super_editor/test/super_editor/supereditor_floating_cursor_test.dart
angelosilvestre: Updated

super_editor/test/super_editor/supereditor_floating_cursor_test.dart
angelosilvestre: Updated.

super_editor/test/super_editor/supereditor_floating_cursor_test.dart
angelosilvestre: Updated.

angelosilvestre:
@matthew-carroll I updated the tests to be single-line. When using a multi-line editor and moving the floating cursor far to the right, sometimes the selection is moved to the second line.

matthew-carroll:
Does this PR need to be rebased for goldens to pass?

angelosilvestre:
@matthew-carroll Rebased

super_editor/lib/src/infrastructure/platforms/ios/ios_document_controls.dart
```diff
@@ -100,6 +100,13 @@ class IosDocumentTouchEditingControls extends StatefulWidget {
 
 class _IosDocumentTouchEditingControlsState extends State<IosDocumentTouchEditingControls>
     with SingleTickerProviderStateMixin {
+  /// Represents the maximum distance an offset can be from the end of a text
```
matthew-carroll, line 103: There are a few details about this comment that can be improved:

"Represents the" - This is superfluous. The number doesn't represent something. The number IS something. So no need for that qualifier.

"maximum distance an offset can be" - An "offset" is a distance. This begs the question "maximum distance that WHAT can be?"

"from the end of a text" - "a text" isn't a thing. There are character bounds, paragraph bounds, document bounds, etc. Also, given a particular set of bounds, it's unclear what "end" means. For example, what's the "end" of a paragraph's bounds?

It looks to me like what you probably mean is the following:

```dart
/// The maximum horizontal distance from the bounds of selectable text, for which we want to render 
/// the floating cursor.
///
/// Beyond this distance, no floating cursor is rendered.
```

Does that cover the relevant details?

super_editor/lib/src/infrastructure/platforms/ios/ios_document_controls.dart
```diff
@@ -273,20 +286,35 @@ class _IosDocumentTouchEditingControlsState extends State<IosDocumentTouchEditin
...
-    if (!widget.editingController.shouldDisplayCollapsedHandle &&
-        !widget.editingController.shouldDisplayExpandedHandles) {
-      editorGesturesLog.finer('Not building overlay handles because they aren\'t desired');
-      return [];
-    }
+  Widget _buildHandles() {
+    return ValueListenableBuilder<bool>(
+      valueListenable: _isFloatingCursorOverOrNearText,
+      builder: (context, isNearText, __) {
+        if (isNearText) {
```
matthew-carroll, line 293: Please add a comment about why the handles care what's going on with the floating cursor.

matthew-carroll:
@miguelcmedeiros I'd like to know if this PR works as expected, but I don't think we want to spend the effort re-targeting the PR. Once it's merged, can you give it a try and report back?

miguelcmedeiros:
> Once it's merged, can you give it a try and report back?

@matthew-carroll 👍

super_editor/lib/src/infrastructure/platforms/ios/ios_document_controls.dart
angelosilvestre: Updated.

angelosilvestre:
@matthew-carroll  This is a bit laggy because it was recorded from macincloud.

https://user-images.githubusercontent.com/7597082/228381483-6c417596-6695-48af-8956-4616ad2d32cf.mp4

matthew-carroll:
Looking back at the original video, and demo'ing iOS myself, there are two details we haven't captured here.

1. The gray caret animates to/from the floating cursor
2. The floating cursor slightly snaps to line boundaries

@miguelcmedeiros are these details that you think we should spend time on?

miguelcmedeiros:
The current implementation is definitely an improvement. I am fine with it and leave those 2 details for later.

matthew-carroll:
@angelosilvestre can you file a ticket for the couple of missing features that I mentioned above?

matthew-carroll (approved):
LGTM

PR #1036, opened 2023-02-18
Don't change selection when the user taps to stop scrolling momentum on mobile (Resolves #1035)

### LINKED ISSUE #1035: [SuperEditor][iOS][Android] - Tap should stop scrolling
When scrolling `SuperEditor` on mobile, and scroll momentum takes over, tapping on the document places the caret, instead of stopping the scroll momentum.

An example of the behavior was posted in #927

matthew-carroll:
Don't change selection when the user taps to stop scrolling momentum on mobile (Resolves #1035)

On Android, tapping on a `SingleChildScrollView` seems to stop the scroll momentum automatically. We don't get any direct notification of this, and our tap handlers are executed even though the user's action stopped the momentum. This PR adds some careful accounting to prevent us from taking further action when the user taps to stop scrolling.

On iOS, at least in the simulator, tapping on a `SingleChildScrollView` doesn't automatically stop the scrolling momentum. But it should. This PR forces the `SingleChildScrollView` to go idle when the user taps down.

### DIFF
--- a/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
@@ -89,6 +89,7 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
   // The alternative case is the one in which this interactor defers to an
   // ancestor scrollable.
   late ScrollController _scrollController;
+  bool _isScrolling = false;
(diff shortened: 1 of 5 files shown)


miguelcmedeiros (approved):
Code looks good 👍 
Is it possible to cover these changes with widget tests?

PR #1041, opened 2023-02-21
Implemented beginning of paragraph case for _insertBlockLevelContent

Jethro87:
The recent work on #1019 fixed image and horizontal rule insertion in the middle of a non-empty paragraph node. Both images and horizontal rules call `_insertBlockLevelContent`, which handles the following cases:
- Empty paragraph
- Selection at the end of a non-empty paragraph
- Selection in the middle of a non-empty paragraph

This PR adds the following case to the `_insertBlockLevelContent` function:
- Selection at the beginning of a non-empty paragraph

Before this PR, adding a horizontal rule or image when the selection is at the beginning of a non-empty paragraph would end up using the *Selection in the middle of a non-empty paragraph* case, which would end up adding an extra empty paragraph node above the original paragraph node. 

Now, the horizontal rule or image node is inserted before the paragraph node, no additional empty paragraph nodes are inserted, and selection remains at the beginning of the paragraph node.

#### Here's the behavior before this PR. Note the additional empty paragraph nodes inserted above the paragraph node:

https://user-images.githubusercontent.com/6467808/220417525-abf11a78-779a-4070-ba0f-2790fcba7bd4.mov

#### Here's the behavior after this PR:

https://user-images.githubusercontent.com/6467808/220417643-d3a101b1-e2b0-49d6-aa56-b37ec7d51287.MP4

### DIFF
--- a/super_editor/lib/src/default_editor/common_editor_operations.dart
+++ b/super_editor/lib/src/default_editor/common_editor_operations.dart
@@ -1875,6 +1875,9 @@ class CommonEditorOperations {
   /// is converted into the given [blockNode] and a new empty paragraph
   /// is inserted after the [blockNode].
   ///
+  /// If the selection extent sits at the beginning of a non-empty paragraph,
+  /// the [blockNode] is inserted as a new node before that paragraph.
+  ///
   /// If the selection extent sits at the end of a paragraph, the [blockNode]
   /// is inserted as a new node after that paragraph, and then a new
   /// empty paragraph is inserted after the [blockNode].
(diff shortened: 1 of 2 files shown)


Jethro87:
cc @matthew-carroll

matthew-carroll (approved):
LGTM

Jethro87:
Thanks @matthew-carroll. Can you cherry pick this to stable?

matthew-carroll:
@angelosilvestre can you put up the cherry pick PR for this to stable?

PR #1042, opened 2023-02-21
[SuperTextField][Android] Fix emoji deletion (Resolves #1031)

### LINKED ISSUE #1031: Emoji Delete Error - Only Android
Hi,

Firstly, thanks for this awesome package and extensive work.

I'm using latest main branch with this code:

```
            SuperTextField(
              controlsColor: AppColors.appPink,
              selectionColor: AppColors.appPink50,
              textAlign: TextAlign.center,
              textController: viewModel.textController,
              focusNode: viewModel.focusNode,
              lineHeight: viewModel.lineHeight,
              textStyleBuilder: (Set<Attribution> attributions) {
                return viewModel.textStyleBuilder(attributions, viewModel.textStyle, false);
              },
              maxLines: null,
            ),
```

It is working on iOS without a problem, but on Android when you type an emoji and click delete, displays unknown square character. If you click delete again, it fixes and completes emoji delete.

Example video:

https://user-images.githubusercontent.com/2773578/219391055-b800483d-d794-4023-90d1-0ad0cd14d70a.MOV

Is there any workaround or fix you can provide?

Thanks,

angelosilvestre:
[SuperTextField][Android] Fix emoji deletion. Resolves #1031

Using the Samsung keyboard, pressing the backspace button generates a key event instead of a deletion delta.

In the `deletePreviousCharacter` method, we aren't handling glyphs with multiple codepoints correctly.

This PR fixes the deletion of characters with multiple codepoints.

### DIFF
--- a/super_editor/lib/src/infrastructure/super_textfield/infrastructure/attributed_text_editing_controller.dart
+++ b/super_editor/lib/src/infrastructure/super_textfield/infrastructure/attributed_text_editing_controller.dart
@@ -515,10 +515,12 @@ class AttributedTextEditingController with ChangeNotifier {
       return;
     }
 
+    final deleteFromOffset = getCharacterStartBounds(_text.text, selection.extentOffset);
+
     delete(
-      from: selection.extentOffset - 1,
+      from: deleteFromOffset,
       to: selection.extentOffset,
-      newSelection: TextSelection.collapsed(offset: selection.extentOffset - 1),
+      newSelection: TextSelection.collapsed(offset: deleteFromOffset),
       newComposingRegion: newComposingRegion,
     );
   }
@@ -537,9 +539,11 @@ class AttributedTextEditingController with ChangeNotifier {
       return;
     }
 
+    final deleteToOffset = getCharacterEndBounds(_text.text, selection.extentOffset);
+
     delete(
       from: selection.extentOffset,
-      to: selection.extentOffset + 1,
+      to: deleteToOffset,
       newSelection: TextSelection.collapsed(offset: selection.extentOffset),
       newComposingRegion: newComposingRegion,
     );
--- a/super_editor/test/super_textfield/super_textfield_ime_test.dart
+++ b/super_editor/test/super_textfield/super_textfield_ime_test.dart
@@ -354,6 +354,22 @@ void main() {
       // Ensure text is deleted
       expect(controller.text.text, 'This is a');
     });
+
+    testWidgetsOnAndroid('deletes emojis with BACKSPACE (on Android)', (tester) async {
+      final controller = AttributedTextEditingController(
+        text: AttributedText(text: 'This is a text with an emoji 🐢'),
+      );
+      await _pumpScaffoldForBuggyKeyboards(tester, controller: controller);
+
+      // Place the caret at the end of the text field.
+      await tester.placeCaretInSuperTextField(controller.text.text.length);
+
+      // Press backspace to delete the previous character.
+      await tester.pressBackspace();
+
+      // Ensure the emoji is deleted.
+      expect(controller.text.text, 'This is a text with an emoji ');
+    });
   });
 }
 


super_editor/test/super_textfield/super_textfield_ime_test.dart
```diff
@@ -354,6 +354,22 @@ void main() {
       // Ensure text is deleted
       expect(controller.text.text, 'This is a');
     });
+
```
matthew-carroll, line 357: Don't we already have some emoji tests? Where are they?

super_editor/test/super_textfield/super_textfield_ime_test.dart
```diff
@@ -354,6 +354,22 @@ void main() {
...
+    testWidgetsOnAndroid('deletes emojis with BACKSPACE (on Android)', (tester) async {
+      final controller = AttributedTextEditingController(
+        text: AttributedText(text: 'This is a text with an emoji 🐢'),
+      );
+      await _pumpScaffoldForBuggyKeyboards(tester, controller: controller);
```
matthew-carroll, line 362: can you remind what pumping for "buggy keyboards" does?

super_editor/lib/src/infrastructure/super_textfield/infrastructure/attributed_text_editing_controller.dart
```diff
@@ -515,10 +515,12 @@ class AttributedTextEditingController with ChangeNotifier {
       return;
     }
 
+    final deleteFromOffset = getCharacterStartBounds(_text.text, selection.extentOffset);
```
matthew-carroll, line 518: Is `deleteFromOffset` the same name used in the other methods where we implemented emoji deletion support? I'm thinking `baseDeletionOffset` would be more clear, to match with `selection.extentOffset`. But I'm curious what the other implementation uses.

super_editor/test/super_textfield/super_textfield_ime_test.dart
angelosilvestre: We have some emoji tests in `super_editor\test\super_textfield\super_textfield_text_layout_test.dart`. Should I move these tests there?

super_editor/test/super_textfield/super_textfield_ime_test.dart
angelosilvestre: It just pumps a regular widget tree containing a `SuperTextField`...

super_editor/lib/src/infrastructure/super_textfield/infrastructure/attributed_text_editing_controller.dart
angelosilvestre: In `SuperEditor` we use `previousCharacterOffset`.

super_editor/test/super_textfield/super_textfield_ime_test.dart
matthew-carroll: So what makes it "buggy"?

super_editor/lib/src/infrastructure/super_textfield/infrastructure/attributed_text_editing_controller.dart
matthew-carroll: Does that work here? Or are we deleting multiple characters in this case?

super_editor/test/super_textfield/super_textfield_ime_test.dart
matthew-carroll: We have tests for keyboard emoji deletions in a layout test suite?

super_editor/test/super_textfield/super_textfield_ime_test.dart
angelosilvestre: The widget tree itself doesn't have anything to do with "buggy" keyboards. The tests that use this method simulate "buggy" keyboards that send key events for backspace instead of deletion deltas.

super_editor/test/super_textfield/super_textfield_ime_test.dart
angelosilvestre: I don't remember why that file is named like this, but it isn't related to layout. It's related to selection when the textfield contains emojis.

angelosilvestre: We could rename it to `super_textfield_emoji_test.dart`

super_editor/lib/src/infrastructure/super_textfield/infrastructure/attributed_text_editing_controller.dart
angelosilvestre: Yeah, it will work.

super_editor/test/super_textfield/super_textfield_ime_test.dart
matthew-carroll: Ok. For now let's rename it to `super_textfield_emoji_test.dart` and let's move these tests into there. Maybe later we'll combine those emoji tests back in with regular text editing tests.

matthew-carroll (approved):
LGTM

PR #1043, opened 2023-02-22
[SuperEditor] Fix toggling attributions on first insertion (Resolves #1007)

### LINKED ISSUE #1007: [SuperEditor] - Toggling an attribution is not possible on first selection of document
### Description

When trying to toggle an attribution either through a keyboard shortcut (like the default `cmd + b` for **bold**) or by calling `ComposerPreferences.toggleStyle(boldAttribution)` directly, the attribution doesn't become active if it was called on the first selected that a user made.

i.e.
1. User focuses the document
2. User presses `cmd + b`
3. User writes text
-> actual: written text has no attribution
-> expected: text is written in **bold**

Tested on the latest stable `3.7.0` using super_editor `0.2.3+1`

### Details

What might be the cause of this issue:
When `ComposerPreferences.toggleStyle(boldAttribution)` is called it causes `SuperEditorState._updateComposerPreferencesAtSelection` to be called, which in turn calls `_composer.preferences.clearStyles();` and immediately clears the attribution that we just enabled.

### Showcase:

<details><summary>minimal example</summary><p>

```dart

void main() {
  runApp(const MyApp());
}

  const MyApp({super.key});

  @override
    return const MaterialApp(
      home: HomePage(),
    );
  }
}

class HomePage extends StatefulWidget {
  const HomePage({super.key});

  @override
  State<HomePage> createState() => _HomePageState();
}

class _HomePageState extends State<HomePage> {
  late final DocumentComposer _composer;
  late final DocumentEditor _editor;

  @override
  void initState() {
    super.initState();

    _composer = DocumentComposer();
    _editor = DocumentEditor(
      document: MutableDocument(
        nodes: [
          ParagraphNode(
            id: '1',
            text: AttributedText(text: 'example text'),
          ),
        ],
      ),
    );
  }

  @override
    return Scaffold(
      body: SafeArea(
        child: Column(
          children: [
            ElevatedButton(
              onPressed: () {
                setState(() {
                  _composer.preferences.toggleStyle(boldAttribution);
                });
              },
              child: const Text('toggle attribute bold'),
            ),
            Expanded(
              child: SuperEditor(
                editor: _editor,
                composer: _composer,
              ),
            ),
          ],
        ),
      ),
    );
  }
}
```

</p></details>

https://user-images.githubusercontent.com/20045287/214872670-374060f6-7c00-4c4b-8059-f34ed046ea35.mov

angelosilvestre:
[SuperEditor] Fix toggling attributions on first insertion. Resolves #1007

After placing the caret at the end of a paragraph for the first time and toggling an attribution, the attribution isn't preserved when the user types:

https://user-images.githubusercontent.com/20045287/214872670-374060f6-7c00-4c4b-8059-f34ed046ea35.mov

The cause is that, upon the first insertion, the IME is sending us a `TextEditingDeltaNonTextUpdate` changing only the selection affinity. As the selection changed, we clear the current attributions.

This PR changes the `_updateComposerPreferencesAtSelection`  method to avoid updating the attributions when the selection is at the same node and the same offset as the previous selection.

### DIFF
--- a/super_editor/lib/src/default_editor/super_editor.dart
+++ b/super_editor/lib/src/default_editor/super_editor.dart
@@ -358,6 +358,8 @@ class SuperEditorState extends State<SuperEditor> {
   void dispose() {
     _contentTapDelegate?.dispose();
 
+    _composer.removeListener(_updateComposerPreferencesAtSelection);
+
     if (widget.composer == null) {
       _composer.dispose();
     }
@@ -434,6 +436,21 @@ class SuperEditorState extends State<SuperEditor> {
     if (_composer.selection?.extent == _previousSelectionExtent) {
       return;
     }
+
+    final selectedNodePosition = _composer.selection?.extent.nodePosition;
+    final previousSelectedNodePosition = _previousSelectionExtent?.nodePosition;
+
+    if (_composer.selection != null &&
+        _previousSelectionExtent != null &&
+        _composer.selection!.extent.nodeId == _previousSelectionExtent!.nodeId &&
+        (selectedNodePosition is TextNodePosition) &&
+        (previousSelectedNodePosition is TextNodePosition) &&
+        selectedNodePosition.offset == previousSelectedNodePosition.offset) {
+      // The selection changed, but the selected node and offset are the same.
+      // It might be the case that the OS sent us a selection with a different affinity.
+      return;
+    }
+
     _previousSelectionExtent = _composer.selection?.extent;
 
     _composer.preferences.clearStyles();
--- a/super_editor/lib/src/default_editor/super_editor.dart
+++ b/super_editor/lib/src/default_editor/super_editor.dart
@@ -437,18 +437,22 @@ class SuperEditorState extends State<SuperEditor> {
       return;
     }
 
-    final selectedNodePosition = _composer.selection?.extent.nodePosition;
-    final previousSelectedNodePosition = _previousSelectionExtent?.nodePosition;
-
-    if (_composer.selection != null &&
+    final selectionExtent = _composer.selection?.extent;
+    if (selectionExtent != null &&
+        selectionExtent.nodePosition is TextNodePosition &&
         _previousSelectionExtent != null &&
-        _composer.selection!.extent.nodeId == _previousSelectionExtent!.nodeId &&
-        (selectedNodePosition is TextNodePosition) &&
-        (previousSelectedNodePosition is TextNodePosition) &&
-        selectedNodePosition.offset == previousSelectedNodePosition.offset) {
-      // The selection changed, but the selected node and offset are the same.
-      // It might be the case that the OS sent us a selection with a different affinity.
-      return;
+        _previousSelectionExtent!.nodePosition is TextNodePosition) {
+      // The current and previous selections are text positions. Check for the situation where the two
+      // selections are functionally equivalent, but the affinity changed.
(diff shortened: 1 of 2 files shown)


super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -414,6 +416,21 @@ class SuperEditorState extends State<SuperEditor> {
     if (_composer.selection?.extent == _previousSelectionExtent) {
       return;
     }
+
+    final selectedNodePosition = _composer.selection?.extent.nodePosition;
```
matthew-carroll, line 420: Let's replace all this with the following, which makes it easier to read the meaning of what we're checking.

```dart
final selectionExtent = _composer.selection?.extent;
if (selectionExtent != null && selectionExtent.nodePosition is TextNodePosition 
    && _previousSelectionExtent != null && _previousSelectionExtent!.nodePosition is TextNodePosition) {
  // The current and previous selections are text positions. Check for the situation where the two
  // selections are functionally equivalent, but the affinity changed.
  final selectedNodePosition = selectionExtent.nodePosition as TextNodePosition;
  final previousSelectedNodePosition = _previousSelectionExtent!.nodePosition as TextNodePosition;

  if (selectionExtent.nodeId == _previousSelectionExtent!.nodeId &&
        selectedNodePosition.offset == previousSelectedNodePosition.offset) {
      // The text selection changed, but only the affinity is different. An affinity change doesn't alter
      // the selection from the user's perspective, so don't alter any preferences. Return.
      return;
    }
}
```

super_editor/test/super_editor/supereditor_attributions_test.dart
```diff
@@ -199,5 +203,50 @@ void main() {
...
+
+        final doc = context.editContext.editor.document;
+        final composer = context.editContext.composer;
+
+        // Place the caret at the end of the paragraph.
+        await tester.placeCaretInParagraph(doc.nodes.first.id, 19);
+
+        // Toggle the bold attribution.
+        composer.preferences.toggleStyle(boldAttribution);
+        await tester.pumpAndSettle();
```
matthew-carroll, line 223: Does this need to settle?

super_editor/test/super_editor/supereditor_attributions_test.dart
```diff
@@ -199,5 +203,50 @@ void main() {
...
+
+        // Invert the selection affinity.
+        final newAffinity =
```
matthew-carroll, line 226: Instead of selecting an affinity like this, I think I'd replace this code with an `expect()` that the affinity is `downstream`, and then down below, I would hard code a value of `upstream` for the delta.

super_editor/lib/src/default_editor/super_editor.dart

super_editor/test/super_editor/supereditor_attributions_test.dart
angelosilvestre: `pump` already works. Updated.

super_editor/test/super_editor/supereditor_attributions_test.dart
angelosilvestre: Updated, but when tapping at the end of the paragraph we get an upstream affinity.

matthew-carroll (approved):
LGTM

matthew-carroll:
@angelosilvestre looks like there are conflicts. Can you resolve them?

angelosilvestre:
@matthew-carroll Conflicts resolved.

PR #1045, opened 2023-02-24
[SuperReader] Fix setState() called after dispose (Resolves #1029)

### LINKED ISSUE #1029: [SuperReader] setState() called after dispose
I use `SuperReader` widgets in order to display a document, but the document itself is tappable and pushes a new route. The following issue has been happening quite often recently:

```
════════ Exception caught by scheduler library ═════════════════════════════════
The following assertion was thrown during a scheduler callback:
setState() called after dispose(): _ReadOnlyIOSDocumentTouchInteractorState#5e160(lifecycle state: defunct, not mounted, ticker inactive)

This error happens if you call setState() on a State object for a widget that no longer appears in the widget tree (e.g., whose parent widget no longer includes the widget in its build). This error can occur when code calls setState() from a timer or an animation callback.

The preferred solution is to cancel the timer or stop listening to the animation in the dispose() callback. Another solution is to check the "mounted" property of this object before calling setState() to ensure the object is still in the tree.
This error might indicate a memory leak if setState() is being called because another object is retaining a reference to this State object after it has been removed from the tree. To avoid memory leaks, consider breaking the reference to this object during dispose().

#0      State.setState.<anonymous closure>
framework.dart:1097
#1      State.setState
framework.dart:1132
#2      _ReadOnlyIOSDocumentTouchInteractorState.didChangeDependencies.<anonymous closure>
read_only_document_ios_touch_interactor.dart:194
```

I think there should be a `mounted` check before `setState` is called. This is in `didChangeDependencies` in `ReadOnlyIOSDocumentTouchInteractor`.

jasimameen:
[SuperReader] Fix setState() called after dispose Resolves #1029 

This PR checks Whether this State object is currently in a tree. and setsState() only executes only when the state is mounted

### DIFF
--- a/super_editor/lib/src/super_reader/read_only_document_ios_touch_interactor.dart
+++ b/super_editor/lib/src/super_reader/read_only_document_ios_touch_interactor.dart
@@ -248,6 +248,13 @@ class _ReadOnlyIOSDocumentTouchInteractorState extends State<ReadOnlyIOSDocument
       });
     }
   }
+  
+  @override
+  void setState(VoidCallback fn) {
+    if (mounted) {
+      super.setState(fn);
+    }
+  }
 
   @override
   void dispose() {


matthew-carroll:
If `setState()` is called at a bad time then the appropriate change is to avoid calling it when the target isn't mounted. This PR is treating the symptom rather than the cause.

Can you please link me to the line of code that's calling `setState() `at a bad time?

PR #1056, opened 2023-03-11
Add a demo where the height of the content (e.g. task) can animate

miguelcmedeiros:
This new demo has a custom task component that shows/hides content under the task title with a size animation, depending on whether the task has selection or not. 

In this demo it is a custom implementation of `TaskComponent`, but it could be other types of nodes, like an image nodes that animates from a placeholder to the actual image once it's fully loaded.

As described in #1022, the animation of the height of components will not trigger the caret offset to be recalculated, since it currently only reacts to changes to the selections or to the document content (i.e. nodes). Same applies for any other overlay builder that attempts to render something above the document.


matthew-carroll (approved):
LGTM

PR #1057, opened 2023-03-11
[SuperEditor][SuperTextField] Implement font scaling (Resolves #1055)

### LINKED ISSUE #1055: [SuperEditor][SuperReader] Support Dynamic Type font scaling
`TextField` and `CupertinoTextField` will scale font sizes based on the system-level Dynamic Type accessibility settings. super_editor doesn't seem to scale at all.

In the following screenshots, the top text is part of a `CupertinoTextField` and the bottom text is part of super_editor.

### Standard-sized text
IMG_1238

### Scaled text
IMG_1239

### Dynamic Type settings
IMG_1240

I noticed that `RenderLayoutAwareParagraph` in `super_text.dart` has `textScaleFactor` set to 1.0, which I think *might* be disabling dynamic type:

```
/// A [RenderParagraph] that publicly reports whether or not its
/// layout is valid, and also accepts a callback to notify a listener
/// when the layout is marked invalid.
class RenderLayoutAwareParagraph extends RenderParagraph {
  RenderLayoutAwareParagraph(
    InlineSpan text, {
    TextAlign textAlign = TextAlign.start,
    required TextDirection textDirection,
    bool softWrap = true,
    TextOverflow overflow = TextOverflow.clip,
    double textScaleFactor = 1.0,
    int? maxLines,
    Locale? locale,
    StrutStyle? strutStyle,
    TextWidthBasis textWidthBasis = TextWidthBasis.parent,
    TextHeightBehavior? textHeightBehavior,
    List<RenderBox>? children,
    VoidCallback? onMarkNeedsLayout,
  })  : _onMarkNeedsLayout = onMarkNeedsLayout,
        super(
          text,
          textAlign: textAlign,
          textDirection: textDirection,
          softWrap: softWrap,
          overflow: overflow,
          textScaleFactor: textScaleFactor,
          maxLines: maxLines,
          locale: locale,
          strutStyle: strutStyle,
          textWidthBasis: textWidthBasis,
          textHeightBehavior: textHeightBehavior,
          children: children,
        );
...
```

@matthew-carroll Is there a way to support dynamic type in super_editor?

angelosilvestre:
[SuperEditor][SuperTextField] Implement font scaling. Resolves #1055

Currently, SuperEditor doesn't scale the text when the user changes the text scaling settings in the OS. This is caused because we are always using a `textScaleFactor` of `1.0` in `RenderLayoutAwareParagraph`.

This PR adds a `textScaleFactor` property in both `RenderLayoutAwareParagraph` and `SuperTextWithSelection`, so it can be customized. 

This PR also changes SuperEditor components and SuperTextField to get the `textScaleFactor` from the `MediaQuery`.

https://user-images.githubusercontent.com/7597082/224502926-d5cafdd9-d88b-40b2-8015-c63ba7d943e0.mp4

There are some open questions:

* Should we expose the `textScaleFactor` in SuperEditor and SuperTextField public API's?
* Instead of changing the components and SuperTextField, should `SuperTextWithSelection` get the `textScaleFactor` from the `MediaQuery` ?
* Do we need some golden tests for SuperEditor and SuperTextField with a different `textScaleFactor`?


matthew-carroll:
@Jethro87 @miguelcmedeiros what are your thoughts on being able to control this scale?

I think at a minimum, the fundamental text layout widgets should probably defer to `MediaQuery`, with an optional property that would override the `MediaQuery` value.

Do you think `SuperEditor`, `SuperReader`, and `SuperTextField` should offer some kind of property for this at that top level? If we do offer a property, it's one more in a long list. If we don't offer a property then developers would need to introduce a `MediaQuery` right above those widgets to set the scale factor back.

I'll think more about it, but wanted to get your 2 cents.

miguelcmedeiros:
@matthew-carroll 
Maybe there are use cases for overriding this property per component, but I don't really have one in mind.
As long as super_editor gets the value from MediaQuery, I would be happy about and then depending on valid use cases (maybe @Jethro87 has some in mind) we can extend the public APIs to support overriding it at a more fine level.

Jethro87:
@matthew-carroll I haven't been able to think of any use cases for Clearful where we would want to modify the text scale. As this is modifiable using MediaQuery, perhaps we should release this without adding a top-level property and revisit this if legitimate use cases come up.

matthew-carroll:
Ok, so then responding to open questions:

 > Should we expose the textScaleFactor in SuperEditor and SuperTextField public API's?
No. We'll avoid exposing it, for now, in `SuperEditor`, `SuperReader`, and `SuperTextField`.

 > Instead of changing the components and SuperTextField, should SuperTextWithSelection get the textScaleFactor from the MediaQuery ?
`SuperEditor`, `SuperReader`, and `SuperTextField` shouldn't change at all, I don't think. But the text layout widget should take an optional property for the text scale, and if it's `null`, the text layout widget should defer to `MediaQuery`.

 > Do we need some golden tests for SuperEditor and SuperTextField with a different textScaleFactor?
That seems like a good idea.

super_editor/lib/src/default_editor/list_items.dart
```diff
@@ -285,7 +285,7 @@ class UnorderedListItemComponent extends StatelessWidget {
   Widget build(BuildContext context) {
     final textStyle = styleBuilder({});
     final indentSpace = indentCalculator(textStyle, indent);
-    final lineHeight = textStyle.fontSize! * (textStyle.height ?? 1.25);
+    final lineHeight = textStyle.fontSize! * (textStyle.height ?? 1.25) * MediaQuery.textScaleFactorOf(context);
```
matthew-carroll, line 288: Hm. This is a problem. The actual text scale hasn't been chosen at this point. Maybe the text layout chooses to use the `MediaQuery`, or maybe not...

super_editor/lib/src/default_editor/text.dart
```diff
@@ -435,6 +436,15 @@ class TextComponent extends StatefulWidget {
   final bool highlightWhenEmpty;
   final bool showDebugPaint;
 
+  /// The number of font pixels for each logical pixel.
```
matthew-carroll, line 439: Is this copied from the official docs? Parts of this doc comment sound a little strange.

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -296,7 +296,7 @@ class SuperDesktopTextFieldState extends State<SuperDesktopTextField> implements
       return lineHeight;
     }
     final defaultStyle = widget.textStyleBuilder({});
-    return _estimatedLineHeight.calculate(defaultStyle);
+    return _estimatedLineHeight.calculate(defaultStyle, MediaQuery.textScaleFactorOf(context));
```
matthew-carroll, line 299: This is another instance of the earlier issue. We're making decisions based on knowledge that we're not supposed to have at this point...

super_editor/test/super_editor/supereditor_text_layout_test.dart
```diff
@@ -0,0 +1,34 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:super_text_layout/super_text_layout.dart';
+
+import '../test_tools.dart';
+import 'document_test_tools.dart';
+
+void main() {
+  group('SuperEditor', () {
+    testWidgetsOnAllPlatforms('textScaleFactor defaults to OS configuration', (tester) async {
```
matthew-carroll, line 10: This test name is overly focused on implementation details - namely the reference to `textScaleFactor`. Instead, we should describe the feature:

```
SuperEditor respects the OS text scaling preference
```

super_editor/test/super_editor/supereditor_text_layout_test.dart
```diff
@@ -0,0 +1,34 @@
...
+                  data: const MediaQueryData(textScaleFactor: 1.5),
+                  child: superEditor,
+                ),
+              ),
+            ),
+          )
+          .pump();
+
+      // Find the widget used to render the paragraph.
+      final superText = tester.widget<SuperTextWithSelection>(find.byType(SuperTextWithSelection));
```
matthew-carroll, line 28: This isn't really a `SuperEditor` test, it's a `SuperTextWithSelection` test.

And `SuperTextWithSelection` just forwards the property to the layout widget, right? If so, this test is very low value. If we're going to verify that a widget property holds a particular value, we should at least write that test against the widget that truly uses that property value.

super_text_layout/lib/src/text_layout.dart
```diff
@@ -246,7 +248,8 @@ class RenderParagraphProseTextLayout implements ProseTextLayout {
     // If no text is currently displayed, we can't use a character box
     // to measure, but we may be able to use related metrics.
     if (_textLength == 0) {
-      final estimatedLineHeight = _renderParagraph.getFullHeightForCaret(position) ?? _richText.style?.fontSize ?? 0.0;
+      final estimatedLineHeight =
+          _renderParagraph.getFullHeightForCaret(position) ?? (_richText.style?.fontSize ?? 0.0) * textScaleFactor;
```
matthew-carroll, line 252: Is there a reason that `textScaleFactor` is only multiplied by the value after the `??`?

super_editor/lib/src/default_editor/list_items.dart
angelosilvestre: Should we keep the new `textScaleFactor` property in `TextComponent`?  That way we could make sure that, in every place we use a `TextComponent`, we are getting the `textScaleFactor` from the `MediaQuery`.

matthew-carroll: > That way we could make sure that, in every place we use a TextComponent, we are getting the textScaleFactor from the MediaQuery.

I mentioned this in another comment, but it's not particularly valuable to test middleman configuration. It doesn't matter what `TextComponent` does, if the eventual value in the `SuperText` widget is wrong. At the same time, if the eventual value in the `SuperText` widget is correct, then we still don't care what `TextComponent` does. We need to verify results with the widget, or render object that actually makes the final decision. 

> Should we keep the new textScaleFactor property in TextComponent?

As for this question in general, I'm not sure. I don't remmeber what other details are in this PR. But I see that a lot of my questions are still unanswered, so I'm guessing that there's more work to do on this PR based on the existing comments.

super_editor/lib/src/default_editor/text.dart
angelosilvestre: Yes, this is from the official docs.

super_editor/test/super_editor/supereditor_text_layout_test.dart
angelosilvestre: Updated.

super_editor/test/super_editor/supereditor_text_layout_test.dart
angelosilvestre: I changed to get the value from the `RenderParagraph`. If `SuperTextWithSelection` will default to the value obtained from the `MediaQueryData`, then we could remove this test and write a golden test instead.

super_editor/lib/src/default_editor/list_items.dart
angelosilvestre: We could get the value from the underlying `RenderParagraph`, but we won't be able to access it on the first build. We would need to schedule rebuilds until it's available.

matthew-carroll: Why not get it from the `SuperText`?

angelosilvestre: We could, but we will have the same issue. The `textScaleFactor` will be resolved from the `MediaQuery` during the `SuperText` build, so during the first build of `UnorderedListItemComponent` it won't be available.

Jethro87:
@matthew-carroll Do you think we could get this into stable in the next 2 weeks?

matthew-carroll:
> Do you think we could get this into stable in the next 2 weeks?

Yeah

super_editor/lib/src/default_editor/list_items.dart
matthew-carroll: Why wouldn't it be available? `MediaQuery` should definitely be available during the first `build()` call to widgets lower down in the tree.

matthew-carroll: > Should we keep the new textScaleFactor property in TextComponent?

Yeah, I guess the best we can do in this case is have the list item components explicitly pass the scale factor to the `TextComponent` so that the list item component can be sure that its own line height calculation is accurate.

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
matthew-carroll: I think we can live with this approach for two reasons:

1. This text field doesn't let clients specify their own text scale factor, so there's no possible override
2. This text field explicitly tells the selectable text widget to use the MediaQuery text scale factor, therefore we're maintaining consistency

But let's make one minor adjustment. Let's define a private method getter called `double get _textScaleFactor` and let's use that getter in both places, to enforce consistency at the compilation level.

super_editor/test/super_editor/supereditor_text_layout_test.dart
```diff
@@ -0,0 +1,36 @@
...
+                  data: const MediaQueryData(textScaleFactor: 1.5),
+                  child: superEditor,
+                ),
+              ),
+            ),
+          )
+          .pump();
+
+      // Find the widget used to render the paragraph.
+      final renderParagraph = tester.firstRenderObject(find.byType(LayoutAwareRichText)) as RenderLayoutAwareParagraph;
```
matthew-carroll, line 30: Can we instead introduce a `SuperTextInspector` in the `super_text_layout` package, similar to our `SuperEditorInspector`, and give it the ability to query the text scale factor?

super_text_layout/lib/src/super_text.dart
```diff
@@ -27,6 +27,7 @@ class SuperText extends StatefulWidget {
     required this.richText,
     this.textAlign = TextAlign.left,
     this.textDirection = TextDirection.ltr,
+    this.textScaleFactor = 1.0,
```
matthew-carroll, line 30: Maybe I said to make this `1.0`, I can't remember, but was there a reason we didn't attempt to defer to `MediaQuery` here?

super_text_layout/test_goldens/super_text_test.dart
```diff
@@ -10,16 +10,25 @@ void main() {
         await _pumpThreeLinePlainText(tester);
         await screenMatchesGolden(tester, "SuperText-reference-render");
       });
+
+      testGoldens("applies textScaleFactor", (tester) async {
```
matthew-carroll, line 14: The difficulty with this golden is that when a human needs to verify the output, it's unclear whether it's right or wrong. I don't have any sense for a 1.0 scale factor would look like in a test, so I have no frame of reference to verify a 2.0 scale factor. Is there something we can do about this? Can you pump them side-by-side? Maybe use the golden toolkit to do that?

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
angelosilvestre: Updated

super_editor/lib/src/default_editor/list_items.dart
angelosilvestre: Updated

super_text_layout/lib/src/super_text.dart
angelosilvestre: Updated to get the value from `MediaQuery`

super_editor/test/super_editor/supereditor_text_layout_test.dart
angelosilvestre: Created a `SuperTextInspector`. The downside is that we need to put this file in the `lib` folder in order to export it.

super_text_layout/test_goldens/super_text_test.dart
angelosilvestre: Updated

angelosilvestre:
@matthew-carroll I think I addressed all the comments.

Should I write other golden tests comparing side by side 1.0 scale factor with another scaleFactor? For example: collapsed selection, expanded selection, list items, etc...

https://github.com/superlistapp/super_editor/pull/1065 should land first, so we can update the goldens.

matthew-carroll:
> https://github.com/superlistapp/super_editor/pull/1065 should land first, so we can update the goldens.

We just landed that, so let's see what's next for this PR

angelosilvestre:
@matthew-carroll It seems Flutter introduced a `const` constructor for `Row` and this is causing the analyzer to complain and fail the CI workflow. For now, I added an ignore for this warning.

matthew-carroll:
@angelosilvestre is there a reason not to update our code to satisfy the lint rule?

angelosilvestre:
@matthew-carroll The const constructor doesn't exist on Flutter stable, so we need to wait until the next stable version.

matthew-carroll:
> The const constructor doesn't exist on Flutter stable, so we need to wait until the next stable version.

This means we should create a followup PR that makes this change just for `main`. Because `main` is supposed to be up to date with `master`. But we can skip that in this PR.

angelosilvestre:
@matthew-carroll PR updated with golden tests for SuperEditor.

matthew-carroll (approved):
LGTM

PR #1050, opened 2023-03-03
Added IME deserialization logs to help debug a Samsung issue.

matthew-carroll:
Added IME deserialization logs to help debug a Samsung issue.

PR #1062, opened 2023-03-13
Deserialize IME to document on every delta, instead of only before the first delta (Partial fix for #1039)

matthew-carroll:
Deserialize IME to document on every delta, instead of only before the first delta (Partial fix for #1039)


matthew-carroll:
@angelosilvestre can you take over this PR and add at least one test that sends multiple deltas at the same time, in a way that's similar to what's described in the issue ticket, and over at https://github.com/superlistapp/super_editor/pull/1050 ?

When you add the test, please make sure that the test fails without this change, and passes with this change.

angelosilvestre:
@matthew-carroll Added a test for multiple deltas.

Initially, I added all the deltas reported in the logs, but the only deltas relevant to the issue are the last ones. 

This test fails in the main branch. Should we add similar tests for deletion and replacement deltas?

matthew-carroll:
> Should we add similar tests for deletion and replacement deltas?

Sure, if you have some real-world deltas that would make sense for those, let's go ahead and test them.

super_editor/test/super_editor/supereditor_ime_test.dart
```diff
@@ -0,0 +1,138 @@
...
+import 'package:flutter_test_robots/flutter_test_robots.dart';
+import 'package:super_editor/src/infrastructure/text_input.dart';
+import 'package:super_editor/src/test/ime.dart';
+import 'package:super_editor/src/test/super_editor_test/supereditor_inspector.dart';
+import 'package:super_editor/src/test/super_editor_test/supereditor_robot.dart';
+
+import '../test_tools.dart';
+import 'document_test_tools.dart';
+
+void main() {
```
matthew-carroll, line 12: Do we not already have a test suite that runs deltas?

angelosilvestre: We do, moved to the correct file.

super_editor/test/src/default_editor/document_input_ime_test.dart
```diff
@@ -166,6 +166,259 @@ void main() {
...
+          )
+        ],
+        getter: imeClientGetter,
+      );
+
+      // Send the insertion of '.'.
+      // The current text is 'Goin.'.
+      await tester.ime.sendDeltas(
+        const [
+          TextEditingDeltaInsertion(
```
matthew-carroll, line 205: Let's move this delta into the list of deltas below.

super_editor/test/src/default_editor/document_input_ime_test.dart
```diff
@@ -166,6 +166,259 @@ void main() {
...
+          )
+        ],
+        getter: imeClientGetter,
+      );
+
+      // Send the insertion of the auto-corrected word,
+      // followed by the insertion of the '.' that were typed.
+      await tester.ime.sendDeltas(
+        const [
+          TextEditingDeltaReplacement(
```
matthew-carroll, line 220: I think I saw the IME use a deletion instead of a replacement. Let's delete all the content here, followed by the insertion with the corrected word.

super_editor/test/src/default_editor/document_input_ime_test.dart
```diff
@@ -166,6 +166,259 @@ void main() {
...
+      );
+
+      // Ensure the text was inserted.
+      expect(
+        SuperEditorInspector.findTextInParagraph('1').text,
+        'Going.',
+      );
+    });
+
+    testWidgetsOnAllPlatforms('applies replacement and insertion deltas at once', (tester) async {
```
matthew-carroll, line 245: I think we can get rid of these other tests becuase the test above this already combines multiple deltas, of different types, into a single change.

angelosilvestre: You mean to keep only "applies different kinds of deltas at once" ?

matthew-carroll: I mean "applies multiple insertion deltas at once", except with the two modifications I mentioned in the comments above. And that name should probably be changed to "applies list of deltas the way some IMEs report them"

super_editor/test/src/default_editor/document_input_ime_test.dart
angelosilvestre: Updated

super_editor/test/src/default_editor/document_input_ime_test.dart
angelosilvestre: Updated

super_editor/test/src/default_editor/document_input_ime_test.dart
angelosilvestre: Updated

angelosilvestre:
@matthew-carroll Test updated

matthew-carroll:
Thanks. Can you cherry-pick this to stable?

PR #1064, opened 2023-03-13
Rebuild document layers when document layout components change (Resolves #1022)

### LINKED ISSUE #1022: [SuperEditor] Caret overlay is not updated when document layout changes due to animations

matthew-carroll:
Rebuild document layers when document layout components change (Resolves #1022)


matthew-carroll:
This initial approach uses `SizeLayoutChangeNotifier` to cause an overlay layer rebuild. This is a proof of concept, which comes with an extra frame cost on rebuilding.

I'm going to iterate on this solution and try to introduce a custom `RenderObject` that will let us rebuild the overlay layers during the same frame that the document layout changes.

@angelosilvestre when I get the final solution in place, I'd like for you to take over the PR to add tests.

There are at least 3 unique testing conditions to make sure layers rebuild when expected:

- When a single component resizes somewhere above the bottom of the document, pushing everything down
- When two components resize at the same time, and resize the exact same amount, so that the overall document size doesn't change, but the positions of individual components does change
- When a single component resizes at the end of a document, which doesn't push any other components down, but it does change the content layout of that final component

We'll want tests for those situations that fail before this PR, and pass after this PR.

matthew-carroll:
This change broke a bunch of tests. I fixed most of them, but there are still a handful of failures.

I'm not sure that there's a reasonable solution to the timing problems in the tests without introducing a custom `RenderObject` that only builds the layers after the document. So I'm gonna go ahead and build that. It's been on my list of things we should build for a while, anyway.

matthew-carroll:
I have a custom `RenderObject` with underly and overlay layers that compiles and runs the editor as desired. Tomorrow I'll clean it up and see if we can get it merged in.


https://user-images.githubusercontent.com/7259036/224955328-14109f45-fc65-40fb-9f60-ea0ff1407da4.mov

matthew-carroll:
@miguelcmedeiros can you put this PR through the paces in Superlist? I've introduced a new `RenderObject` widget called `ContentLayers` which replaces our duct-tap layer solution that used a `Stack`.

But the challenge of a custom `RenderObject` with children is that there's about 1,000 ways to screw it up. So we should put it through a stress test, and also collect clear repro steps for anything that goes wrong.

miguelcmedeiros:
> can you put this PR through the paces in Superlist?

@matthew-carroll, in order to test this change with Superlist it needs to target stable. Would it be possible to change the target branch of this PR to stable?

super_editor/lib/src/infrastructure/content_layers.dart
```diff
@@ -0,0 +1,541 @@
...
+  void removeChild(RenderBox child, Object slot) {
+    print("Removing $slot - $child");
+    assert(_isContentLayersSlot(slot));
+
+    if (slot == _contentSlot) {
+      _content = null;
+    } else if (slot is _UnderlaySlot) {
+      _underlays.remove(child);
+    } else if (slot is _OverlaySlot) {
+      _overlays.remove(child);
```
angelosilvestre, line 368: Can we end up with invalid indexed slots when removing a child which is in the middle of the list? 

For example, if we have indices 1,2,3,4,5, and then remove the child at index 3. Would the remaining indices be 1,2,4,5?

matthew-carroll:
@angelosilvestre this PR probably wasn't ready for review. I should have added some explanation for that. Once @miguelcmedeiros lets me know that it works for Superlist, I'll clean it up, and then I can hand it to you for tests.

matthew-carroll:
We have golden failures in super_text_layout, but this PR didn't touch that, so I'm gonna merge this in. We should follow up with fixes to those goldens.

matthew-carroll:
I filed this Flutter issue to solve a fundamental limitation in Flutter that kept us from achieving our goals in this PR: https://github.com/flutter/flutter/issues/123305

PR #1065, opened 2023-03-14
[SuperEditor][SuperTextLayout] Change golden tests to run in a container (Resolves #1061)

### LINKED ISSUE #1061: Make goldens runnable by all platforms
Add `/bin` scripts to each project in this repo that allows developers on any platform to run goldens based on Linux rendering.

Re-render all goldens with Linux rendering.

Update CI to run goldens on Linux servers instead of Mac servers.

angelosilvestre:
[SuperEditor][SuperTextLayout] Change golden tests to run in a container. Resolves #1061.

Golden files aren't portable across platforms, so goldens generated in one platform can't be used to test on another platform.

This PR adds scripts in the `/bin` directory to run the tests and update the goldens in a docker container. This PR also updates the CI to run the golden tests on Linux. 

Some tests failed due to the selection `TextAffinity`. I updated those tests.

I noticed that we are not running `super_editor` golden tests in the CI, only `super_text_layout`. I don't know if this is intentional, but I added the `super_editor` golden tests to the CI. Some of these goldens are being rendered differently (different colors for drag handles and a missing debug paint).

## Usage

**Running the tests**

```
flutter pub run super_editor:goldens test <options> <directory>
```
Options:

-  `--plain-name "test-name"`: Runs only the tests containing the given value in the test name.

- The directory is optional.

**Updating golden files**

```
flutter pub run super_editor:goldens update <options> <directory>
```
Options:

-  `--plain-name "test-name"`: Updates only the tests containing the given value in the test name.

- The directory is optional.

To run the tests for `super_text_layout`, just replace `super_editor` with `super_text_layout`.


angelosilvestre:
@matthew-carroll It seems we can run non-activated packages with just `flutter pub run <package>:<executable>` without the `global` command.

super_text_layout/pubspec.yaml
```diff
@@ -2,6 +2,8 @@ name: super_text_layout
 description: Configurable, composable, extensible text display for Flutter.
 version: 0.1.4
 homepage: https://github.com/superlistapp/super_editor
+executables:
```
matthew-carroll, line 5: Can we move this section down, just above "dev_dependencies"?

super_editor/pubspec.yaml
```diff
@@ -2,6 +2,8 @@ name: super_editor
 description: Configurable, composable, extensible text editor and document renderer for Flutter.
 version: 0.2.3-dev.1
 homepage: https://github.com/superlistapp/super_editor
+executables:
```
matthew-carroll, line 5: Same here for placement of "executables"

.github/workflows/pr_validation.yaml
```diff
@@ -137,6 +137,27 @@ jobs:
       # Run all golden tests
       - run: flutter test test_goldens
 
+  test_goldens_super_editor:
```
matthew-carroll, line 140: Please declare super_editor tasks above other package tasks

matthew-carroll (commented):
In addition to the PR comments, can you please add a `README_TESTS.md` to `super_editor` and `super_text_layout` and describe how to run and update goldens on those READMEs?

matthew-carroll:
Do you know why goldens are failing on this PR? Shouldn't they work now that we moved to Ubuntu and you regenerated all of them?

angelosilvestre:
@matthew-carroll It's weird that the first run succeded and now the tests are failing. Maybe github's ubuntu-latest differs from docker's ubuntu-latest? Will investigate

angelosilvestre:
@matthew-carroll Flutter seems to be rendering the text slightly differently in the current master. There is another PR with golden tests failing.

This might be a problem for us in the stable branch.

super_text_layout/pubspec.yaml
angelosilvestre: Yeah, updated.

super_editor/pubspec.yaml
angelosilvestre: Updated

.github/workflows/pr_validation.yaml
angelosilvestre: Updated.

angelosilvestre:
It seems that the debug banner is also causing some tests to fail...

matthew-carroll:
> This might be a problem for us in the stable branch.

Can you phrase this as a problem statement and a suggestion for what to do about it?

angelosilvestre:
@matthew-carroll If flutter is rendering differently between master and stable branches and we update the goldens in master, then the golden tests will fail on stable.

Maybe we want to run the golden tests on master only?

matthew-carroll:
> Maybe we want to run the golden tests on master only?

Sure we can do that. If that's the approach, should we merge this in as-is? Or do we need more adjustments? Maybe update the CI config to only run that job when targeting `main`?

angelosilvestre:
@matthew-carroll Updated CI config to run the golden tests on main only

matthew-carroll (approved):
LGTM

PR #1071, opened 2023-03-24
Use dart test for attributed_text

natebosch:
This package does not have a dependency on `flutter_test` and so there is no guarantee that the version of `test_api` from the pub solve is compatible with the `flutter test` command.

See https://github.com/flutter/flutter/issues/91107


zanderso:
@matthew-carroll ptal

matthew-carroll (approved):
LGTM - Thanks

PR #1082, opened 2023-04-14
Cherry-pick "[SuperTextField][mobile] Update viewport height on first frame if possible (Resolves #939) (#1009)" to stable

### LINKED ISSUE #939: [SuperTextField] flicker on initial display (iOS)

angelosilvestre:
This PR cherry-picks "[SuperTextField][mobile] Update viewport height on first frame if possible (Resolves #939) (#1009)" to stable


angelosilvestre:
@matthew-carroll It seems our `ListenableBuilder` is different in the stable branch

matthew-carroll:
@angelosilvestre can you please elaborate on what that means, and what you think we should do about it?

angelosilvestre:
@matthew-carroll On `main`, the `ListenableBuilder` takes a `TransitionBuilder`. On `stable`, it takes a `WidgetBuilder`.

I guess we should cherry-pick https://github.com/superlistapp/super_editor/pull/880

matthew-carroll:
That change was probably made by the Flutter team due to their `master` - does Flutter's `stable` now contain the new `ListenableBuilder`?

angelosilvestre:
No, it doesn't seem to be on stable yet. 

That PR changes all places where we use the `ListenableBuilder`. 

I could adapt the current PR to our `ListenableBuilder` from the stable branch.

matthew-carroll:
> I could adapt the current PR to our ListenableBuilder from the stable branch.

That's what we'll need to do. We can't do anything to our `stable` that doesn't compile against Flutter `stable`

matthew-carroll (approved):
LGTM

PR #1086, opened 2023-04-18
[SuperEditor][Android] Fix text input after a new line insertion (Resolves #1081)

### LINKED ISSUE #1081: IME Error after insert: Couldn't map an IME position to a document position
We are running into this new issue on the latest stable SE and Flutter.

Repro steps:

1. Enter some text into a new document (a sentence or so)
2. Insert text into the document via `CommonEditorOperations.insertPlainText`
3. Tap and place cursor just before the inserted text, press enter (new line) once
4. Tap and place cursor just after the inserted text (now on a new line), press enter (go to next line) once
5. Attempt to enter additional text into the document, throws with aforementioned error

Logs:

══╡ EXCEPTION CAUGHT BY SERVICES LIBRARY ╞══════════════════════════════════════════════════════════
Exception: Couldn't map an IME position to a document position. IME position: TextPosition(offset:
80, affinity: TextAffinity.downstream)

#0      DocumentImeSerializer._imeToDocumentPosition (package:super_editor/src/default_editor/document_ime/document_serialization.dart:271:5)
#1      DocumentImeSerializer.imeToDocumentSelection (package:super_editor/src/default_editor/document_ime/document_serialization.dart:164:18)
#2      TextDeltasDocumentEditor._applyNonTextChange (package:super_editor/src/default_editor/document_ime/document_delta_editing.dart:180:7)
#3      TextDeltasDocumentEditor.applyDeltas (package:super_editor/src/default_editor/document_ime/document_delta_editing.dart:58:9)
#4      DocumentImeInputClient.updateEditingValueWithDeltas (package:super_editor/src/default_editor/document_ime/document_ime_communication.dart:213:30)
#5      DeltaTextInputClientDecorator.updateEditingValueWithDeltas (package:super_editor/src/default_editor/document_ime/ime_decoration.dart:119:14)
#6      TextInput._handleTextInputInvocation (package:flutter/src/services/text_input.dart:1867:63)
#7      TextInput._loudlyHandleTextInputInvocation (package:flutter/src/services/text_input.dart:1753:20)
#8      MethodChannel._handleAsMethodCall (package:flutter/src/services/platform_channel.dart:555:55)
#9      MethodChannel.setMethodCallHandler.<anonymous closure> (package:flutter/src/services/platform_channel.dart:548:34)
#10     _DefaultBinaryMessenger.setMessageHandler.<anonymous closure> (package:flutter/src/services/binding.dart:393:35)
#11     _invoke2.<anonymous closure> (dart:ui/hooks.dart:186:15)
#15     _invoke2 (dart:ui/hooks.dart:185:10)
#16     _ChannelCallbackRecord.invoke (dart:ui/channel_buffers.dart:40:5)
#17     _Channel.push (dart:ui/channel_buffers.dart:130:31)
#18     ChannelBuffers.push (dart:ui/channel_buffers.dart:326:17)
#19     PlatformDispatcher._dispatchPlatformMessage (dart:ui/platform_dispatcher.dart:664:22)
#20     _dispatchPlatformMessage (dart:ui/hooks.dart:86:31)
(elided 3 frames from dart:async)

call:
  MethodCall(TextInputClient.updateEditingStateWithDeltas, [1, {deltas: [{oldText: It works. Now I
  add one prompt and a new line and it breaksWho was your first true friend?
  , deltaText:
  , deltaStart: 79, deltaEnd: 79, selectionBase: 80, selectionExtent: 80, composingBase: -1,
  composingExtent: -1}, {oldText: It works. Now I add one prompt and a new line and it breaks
  Who was your first true friend?
  , deltaText: , deltaStart: -1, deltaEnd: -1, selectionBase: 80, selectionExtent: 80,
  composingBase: -1, composingExtent: -1}, {oldText: It works. Now I add one prompt and a new line and it breaks
  Who was your first true friend?
  , deltaText: , deltaStart: -1, deltaEnd: -1, selectionBase: 80, selectionExtent: 80,
  composingBase: 80, composingExtent: 83}]}])
════════════════════════════════════════════════════════════════════════════════════════════════════
I/TRuntime.CctTransportBackend(24996): Making request to: https://firebaselogging-pa.googleapis.com/v1/firelog/legacy/batchlog
I/TRuntime.CctTransportBackend(24996): Status Code: 200
D/InputMethodManager(24996): showSoftInput() view=io.flutter.embedding.android.FlutterView{6d84cc8 VFED..... .F...... 0,0-1080,2287 #2 aid=1073741824} flags=0 reason=SHOW_SOFT_INPUT
D/InsetsController(24996): show(ime(), fromIme=true)

Another exception was thrown: Exception: Couldn't map an IME position to a document position. IME position: TextPosition(offset: 81, affinity: TextAffinity.downstream)

angelosilvestre:
[SuperEditor][Android] Fix text input after a new line insertion. Resolves #1081 

On Android, placing the caret at the middle of a line and pressing the new line button causes the text input to crash with the exception:

```
Couldn't map an IME position to a document position
```

This is another issue when we get multiple deltas in one batch. 

In this case, we get an insertion delta followed by a non-text delta. The selection offset of the non-text delta becomes invalid after the insertion delta is applied.

For example, we have the following text, where | means the caret:

```
Before the line break |new line
```

After we add a line break we get an insertion delta of `\n` followed by a non-text delta, which has:

```
old text: Before the line break \nnew line
selection at offset: 23
```

The problem is that when we get the `\n`, we split the paragraph into two and our editing text changes to `new line`. With this new text, 23 is an invalid offset.

I'm not sure what's the correct approach to handle this issue.

This PR changes the delta handling to ignore non-text deltas that are in the same batch of text deltas. Whenever we get a text delta, we are already sending back to the OS the new selection. So, in theory, it should be safe to ignore these non-text deltas.

### DIFF
--- a/super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
+++ b/super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
@@ -1,4 +1,7 @@
+import 'dart:math';
+
 import 'package:flutter/foundation.dart';
+import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 import 'package:super_editor/src/core/document.dart';
 import 'package:super_editor/src/core/document_editor.dart';
@@ -10,7 +13,7 @@ import 'document_serialization.dart';
 
 /// Applies software keyboard text deltas to a document.
 class TextDeltasDocumentEditor {
-  const TextDeltasDocumentEditor({
+  TextDeltasDocumentEditor({
     required this.editor,
     required this.selection,
     required this.composingRegion,
@@ -22,6 +25,21 @@ class TextDeltasDocumentEditor {
   final ValueNotifier<DocumentSelection?> selection;
   final ValueNotifier<DocumentRange?> composingRegion;
   final CommonEditorOperations commonOps;
+
+  /// The composing region in the IME.
+  ///
+  /// Typically, this will be the same composing region we get from the
+  /// last [TextEditingDelta] received.
+  ///
+  /// However, in the cases where we receive a non-text delta in the same
+  /// batch where we receive a text delta, this range might be adjusted
+  /// to match our new content after the text delta is applied. This
+  /// can happen, for example, in a paragraph split or a paragraph merge,
+  /// where our changes to the document causes our editing text
+  /// to be different from what the OS thinks it is.
+  ///
+  /// The composing region is adjusted only when we get non-text deltas.
+  TextRange _imeComposingRegion = const TextRange(start: -1, end: -1);
 
   /// Callback invoked when the client should handle a given [TextInputAction].
   ///
@@ -55,7 +73,7 @@ class TextDeltasDocumentEditor {
       } else if (delta is TextEditingDeltaDeletion) {
         _applyDeletion(delta, serializedDocBeforeDelta);
       } else if (delta is TextEditingDeltaNonTextUpdate) {
-        _applyNonTextChange(delta);
+        _applyNonTextChange(delta, serializedDocBeforeDelta);
       } else {
         editorImeLog.shout("Unknown IME delta type: ${delta.runtimeType}");
       }
@@ -77,12 +95,14 @@ class TextDeltasDocumentEditor {
           ? PrependedCharacterPolicy.include
           : PrependedCharacterPolicy.exclude,
     );
-    editorImeLog.fine("Raw IME delta composing region: ${textEditingDeltas.last.composing}");
-    composingRegion.value = finalSerializedDoc.imeToDocumentRange(textEditingDeltas.last.composing);
+    editorImeLog.fine("Raw IME delta composing region: $_imeComposingRegion");
+    composingRegion.value = finalSerializedDoc.imeToDocumentRange(_imeComposingRegion);
     editorImeLog.fine("Document composing region: ${composingRegion.value}");
   }
 
   void _applyInsertion(TextEditingDeltaInsertion delta, DocumentImeSerializer docSerializer) {
+    _imeComposingRegion = delta.composing;
+
     editorImeLog.fine('Inserted text: "${delta.textInserted}"');
     editorImeLog.fine("Insertion offset: ${delta.insertionOffset}");
     editorImeLog.fine("Selection: ${delta.selection}");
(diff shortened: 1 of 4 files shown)


matthew-carroll:
> This PR changes the delta handling to ignore non-text deltas that are in the same batch of text deltas. Whenever we get a text delta, we are already sending back to the OS the new selection. So, in theory, it should be safe to ignore these non-text deltas.

This is only true of the non-text delta merely repeats existing information. But, it's possible that such a delta really does want  to move the selection somewhere else. In that case, it would be inappropriate to ignore such a delta.

Does this same problem exist in iOS? This can't be the first time someone has inserted a newline in the middle of text. How was this ever working?

matthew-carroll:
@angelosilvestre we'll definitely have to go in the direction of "2". We can't ignore any input.

As for that direction, can you think of any form of input, other than newlines, which might be relevant for these situations with multiple deltas at one time? Would any "replacement" deltas cause similar issues? Or any other combination of deltas?

angelosilvestre:
@matthew-carroll Technically, any delta that causes us to end up with an editing text with a different length than the OS thinks it is can cause problems if we get non-text deltas in the same batch.

For example, if the selection is at the beginning of a paragraph and we press backspace, the paragraph will merge with the previous one. If we get a non-text delta in the same batch as the deletion delta, the selection might jump to the beginning of the previous paragraph.

angelosilvestre:
@matthew-carroll I changed the implementation to transform the offsets using a simple diff algorithm.

There is still some work to be done, but I'd like a first-round review to see if I'm in the right direction.

matthew-carroll:
There seems to be a lot of transformation logic in here. Before I go through line by line, can you describe how this new PR code attacks the problem? What should I expect to see in the code?

angelosilvestre:
@matthew-carroll  The issue we are solving is that deltas that change our content might invalidate the offsets used in non-text deltas that are in the same batch.

To address this issue, whenever we get a non-text delta, before we try to change the editor selection, we check the diff between our editing text and the text in the delta. 

We detect if:
 - Text was inserted/deleted at the start/middle/end of the text
 - Text was replaced (as a deletion followed by an insertion)
 
After computing the diff, we add/subtract from the selection offset to account for the changes in the text. For example, given the prior example:

Before adding a new line:

```
Before the line break |new line
selection at offset: 22
``` 

After adding a new line:

```
Before the line break \n|new line
selection at offset: 23
``` 

With the "\n" insertion, we split the paragraph into two, and now our editing text is:

```
new line
```

We compute the diff between `Before the line break \nnew line` and `new line`. The diff detects that the range 0-23 was deleted. As the deleted range appears before the selection offset, we subtract the deleted length, and now our selection is set to offset `0`  

If we merge two paragraphs, it will detect an insertion at the beginning of the text and will add the inserted length to the selection offset.

matthew-carroll:
Ok. I think we'll want to take a different approach. We've already introduced a structure between our `Document` and the IME deltas. All of this text diff'ing is essentially another layer that's now added between our previous layer and the IME deltas. That seems overly complex, and likely to be difficult to maintain over time.

I'll try to think of an alternative approach that adjusts our existing translation layer.

angelosilvestre:
Closing as this was fixed in another PR.

PR #1088, opened 2023-04-18
Add new 'Editor' class and edit pipeline without breaking anything (Related to #748)(Resolves #1087)

### LINKED ISSUE #1087: [SuperEditor] - Add new `Editor` framework
For #748 we're replacing the `DocumentEditor` system with a new edit pipeline. At the center of that pipeline is a class called `Editor`, which essentially replaces `DocumentEditor`.

For this ticket, add `Editor` to the codebase without altering the existing system. We'll do this to get eyeballs on the core API without breaking anything.

matthew-carroll:
Add new 'Editor' class and edit pipeline without breaking anything (Related to #748)(Resolves #1087)

This is the first PR that moves us towards atomic commands, content reactions, and undo/redo.

This PR adds all the core concepts for the edit pipeline without changing any existing behavior. The purpose of this PR is to put the core API in front of other developers and see if there are any objections.

This API was created in a separate branch where I've implemented a handful of content reactions, including linkafication, image replacement, and user tagging.

I have a test suite for this `Editor` class, but I can't integrate that test until we're ready to break `Document` and `DocumentComposer`.

You can read a wiki article about how the edit pipeline concepts are meant to work: https://github.com/superlistapp/super_editor/wiki/How-it-works:-The-DocumentEditor

### DIFF
--- a/super_editor/lib/src/core/editor.dart
+++ b/super_editor/lib/src/core/editor.dart
@@ -1,7 +1,7 @@
 import 'dart:math';
+import 'dart:ui';
 
 import 'package:collection/collection.dart';
-import 'package:flutter/material.dart';
 import 'package:super_editor/src/infrastructure/_logging.dart';
 import 'package:uuid/uuid.dart';
 
@@ -55,8 +55,6 @@ class Editor implements RequestDispatcher {
         _reactionPipeline = reactionPipeline ?? [],
         _changeListeners = listeners ?? [] {
     _context = EditContext(editables);
-    assert(_context.findMaybe<Document>(Editor.documentKey) != null,
-        "Expected a Document in the 'editables' map but it wasn't there");
 
     _commandExecutor = _DocumentEditorCommandExecutor(_context);
   }
@@ -290,7 +288,7 @@ class EditContext {
   final Map<String, Editable> _resources;
 
   /// Finds an object of type [T] within this [EditContext], which is identified by the given [id].
-  T find<T>(String id) {
+  T find<T extends Editable>(String id) {
     if (!_resources.containsKey(id)) {
       editorLog.shout("Tried to find an editor resource for the ID '$id', but there's no resource with that ID.");
       throw Exception("Tried to find an editor resource for the ID '$id', but there's no resource with that ID.");
@@ -307,7 +305,7 @@ class EditContext {
 
   /// Finds an object of type [T] within this [EditContext], which is identified by the given [id], or
   /// returns `null` if no such object is in this [EditContext].
-  T? findMaybe<T>(String id) {
+  T? findMaybe<T extends Editable>(String id) {
     return _resources[id] as T?;
   }
 }
--- a/super_editor/lib/src/core/editor.dart
+++ b/super_editor/lib/src/core/editor.dart
@@ -27,11 +27,11 @@ class Editor implements RequestDispatcher {
   static const Uuid _uuid = Uuid();
 
   /// Service locator key to obtain a [Document] from [find], if a [Document]
-  /// is available in the [EditorContext].
+  /// is available in the [EditContext].
   static const documentKey = "document";
 
   /// Service locator key to obtain a [DocumentComposer] from [find], if a
-  /// [DocumentComposer] is available in the [EditorContext].
+  /// [DocumentComposer] is available in the [EditContext].
   static const composerKey = "composer";
 
   /// Generates a new ID for a [DocumentNode].
@@ -54,7 +54,7 @@ class Editor implements RequestDispatcher {
   })  : _requestHandlers = requestHandlers,
         _reactionPipeline = reactionPipeline ?? [],
         _changeListeners = listeners ?? [] {
-    _context = EditorContext(editables);
+    _context = EditContext(editables);
     assert(_context.findMaybe<Document>(Editor.documentKey) != null,
         "Expected a Document in the 'editables' map but it wasn't there");
 
@@ -70,7 +70,7 @@ class Editor implements RequestDispatcher {
   final List<EditRequestHandler> _requestHandlers;
 
   /// Service Locator that provides all resources that are relevant for document editing.
-  late final EditorContext _context;
+  late final EditContext _context;
 
   /// Executes [EditCommand]s and collects a list of changes.
   late final _DocumentEditorCommandExecutor _commandExecutor;
@@ -217,7 +217,7 @@ class Editor implements RequestDispatcher {
 class _DocumentEditorCommandExecutor implements CommandExecutor {
   _DocumentEditorCommandExecutor(this._context);
 
-  final EditorContext _context;
+  final EditContext _context;
 
   final _commandsBeingProcessed = EditorCommandQueue();
 
@@ -279,17 +279,17 @@ abstract class RequestDispatcher {
 /// A command that alters something in a [Editor].
 abstract class EditCommand {
   /// Executes this command and logs all changes with the [executor].
-  void execute(EditorContext context, CommandExecutor executor);
+  void execute(EditContext context, CommandExecutor executor);
 }
 
 /// All resources that are available when executing [EditCommand]s, such as a document,
 /// composer, etc.
-class EditorContext {
-  EditorContext(this._resources);
+class EditContext {
+  EditContext(this._resources);
 
   final Map<String, Editable> _resources;
 
-  /// Finds an object of type [T] within this [EditorContext], which is identified by the given [id].
+  /// Finds an object of type [T] within this [EditContext], which is identified by the given [id].
   T find<T>(String id) {
     if (!_resources.containsKey(id)) {
       editorLog.shout("Tried to find an editor resource for the ID '$id', but there's no resource with that ID.");
@@ -305,8 +305,8 @@ class EditorContext {
     return _resources[id] as T;
   }
 
-  /// Finds an object of type [T] within this [EditorContext], which is identified by the given [id], or
-  /// returns `null` if no such object is in this [EditorContext].
+  /// Finds an object of type [T] within this [EditContext], which is identified by the given [id], or
+  /// returns `null` if no such object is in this [EditContext].
   T? findMaybe<T>(String id) {
     return _resources[id] as T?;
   }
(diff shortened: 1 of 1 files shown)


matthew-carroll:
CC @jmatth @Jethro87

super_editor/lib/src/core/editor.dart
```diff
@@ -0,0 +1,1052 @@
...
+
+/// All resources that are available when executing [EditCommand]s, such as a document,
+/// composer, etc.
+class EditContext {
+  EditContext(this._resources);
+
+  final Map<String, Editable> _resources;
+
+  /// Finds an object of type [T] within this [EditContext], which is identified by the given [id].
+  T find<T>(String id) {
```
miguelcmedeiros, line 293:
```suggestion
  T find<T extends Editable>(String id) {
```

super_editor/lib/src/core/editor.dart
```diff
@@ -0,0 +1,1052 @@
...
+      throw Exception(
+          "Tried to find an editor resource of type '$T' for ID '$id', but the resource with that ID is of type '${_resources[id].runtimeType}");
+    }
+
+    return _resources[id] as T;
+  }
+
+  /// Finds an object of type [T] within this [EditContext], which is identified by the given [id], or
+  /// returns `null` if no such object is in this [EditContext].
+  T? findMaybe<T>(String id) {
```
miguelcmedeiros, line 310:
```suggestion
  T? findMaybe<T extends Editable>(String id) {
```

super_editor/lib/src/core/editor.dart
```diff
@@ -0,0 +1,1052 @@
+import 'dart:math';
+
+import 'package:collection/collection.dart';
+import 'package:flutter/material.dart';
```
miguelcmedeiros, line 4: nit: do we need to import material?

super_editor/lib/src/core/editor.dart
```diff
@@ -0,0 +1,1052 @@
...
+          runtimeType == other.runtimeType &&
+          const DeepCollectionEquality().equals(_nodes, other.nodes);
+
+  @override
+  int get hashCode => _nodes.hashCode;
+}
+
+// TODO: move the following document stuff into document.dart when we start breaking things.
+
+abstract class Document {
```
miguelcmedeiros, line 729: Should `Document` extend `Editable` instead of having `MutableDocument` (and other implementations) doing so?

The assert in `Editor`'s constructor expects to find an `Editable` of type `Document`, but that should not work.

super_editor/lib/src/core/editor.dart
```diff
@@ -0,0 +1,1052 @@
...
+  bool operator ==(Object other) =>
+      identical(this, other) || other is NodeChangeEvent && runtimeType == other.runtimeType && nodeId == other.nodeId;
+
+  @override
+  int get hashCode => nodeId.hashCode;
+
+  @override
+  String toString() => "[NodeChangeEvent] - Node: $nodeId";
+}
+
```
miguelcmedeiros, line 904: Will the event for selection changes be done in a follow up PR or should we add them here?
(assuming that we need something like `DocumentSelectionChange`...)

super_editor/lib/src/core/editor.dart
```diff
@@ -0,0 +1,1052 @@
...
+    required List<EditRequestHandler> requestHandlers,
+    List<EditReaction>? reactionPipeline,
+    List<EditListener>? listeners,
+  })  : _requestHandlers = requestHandlers,
+        _reactionPipeline = reactionPipeline ?? [],
+        _changeListeners = listeners ?? [] {
+    _context = EditContext(editables);
+    assert(_context.findMaybe<Document>(Editor.documentKey) != null,
+        "Expected a Document in the 'editables' map but it wasn't there");
+
```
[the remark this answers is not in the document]
matthew-carroll: No, because `DocumentComposer` is specific to `SuperEditor`. Other experiences might want to bring their own. Granted, many of our default commands will need it, but I didn't think we should make it a permanent requirement. I'm not even sure requiring a `Document` is a great idea, but I thought it might prevent mistakes.

super_editor/lib/src/core/editor.dart
matthew-carroll: `Document` can be used in a read-only situation, in which case there wouldn't be an `Editor`. I don't think we want to pull `Editor` concepts into `Document`. `MutableDocument`, on the other hand, is intended to be edited, which is why it implements `Editable`.

If we enforce the presence of a `MutableDocument`, specifically, would that prevent you from using your OT version?

super_editor/lib/src/core/editor.dart
matthew-carroll: Yeah. There's a lot more to come. But I'm not sure that we can go further than this before we start breaking things.

super_editor/lib/src/core/editor.dart
```diff
@@ -0,0 +1,1052 @@
...
+  EditCommand _findCommandForRequest(EditRequest request) {
+    EditCommand? command;
+    for (final handler in _requestHandlers) {
+      command = handler(request);
+      if (command != null) {
+        return command;
+      }
+    }
+
+    throw Exception(
```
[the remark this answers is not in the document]
matthew-carroll: Wouldn't the exception halt execution up to the nearest try/catch or zone? If this exception is thrown, I don't think the code that called this method would continue executing, would it?

miguelcmedeiros (approved):
LGTM

super_editor/lib/src/core/editor.dart
```diff
@@ -0,0 +1,1050 @@
...
+
+  /// Tracks the number of request executions that are in the process of running.
+  int _activeCommandCount = 0;
+
+  /// Executes the given [request].
+  ///
+  /// Any changes that result from the given [request] are reported to listeners as a series
+  /// of [EditEvent]s.
+  @override
+  void execute(List<EditRequest> requests) {
```
angelosilvestre, line 125: `[request]`-> `[requests]`

super_editor/lib/src/core/editor.dart
```diff
@@ -0,0 +1,1050 @@
...
+      // their own requests, to make additional changes.
+      _reactToChanges(_activeChangeList!);
+
+      // Notify all listeners that care about changes, but won't spawn additional requests.
+      _notifyListeners(_activeChangeList!);
+
+      // This is the end of a transaction.
+      for (final editable in _context._resources.values) {
+        editable.onTransactionEnd(_activeChangeList!);
+      }
```
angelosilvestre, line 159: Shouldn't the listeners be notified after the transition is marked as completed? Would it make a difference?

super_editor/lib/src/core/editor.dart
```diff
@@ -0,0 +1,1050 @@
...
+
+  @override
+  void prependCommand(command) {
+    _commandsBeingProcessed.prepend(command);
+  }
+
+  @override
+  void appendCommand(command) {
+    _commandsBeingProcessed.append(command);
+  }
```
angelosilvestre, line 248: The parameter should be typed.

super_editor/lib/src/core/editor.dart
```diff
@@ -0,0 +1,1050 @@
...
+
+  /// A transaction that was previously started with [onTransactionStart] has now ended, this
+  /// [Editable] should notify interested parties of changes.
+  void onTransactionEnd(List<EditEvent> edits);
+}
+
+/// An object that processes [EditRequest]s.
+abstract class RequestDispatcher {
+  /// Pushes the given [request] through a [Editor] pipeline.
+  void execute(List<EditRequest> request);
```
angelosilvestre, line 274: As the parameter is a list, shouldn't it be named `requests`?

super_editor/lib/src/core/editor.dart
```diff
@@ -0,0 +1,1050 @@
...
+          "Tried to find an editor resource of type '$T' for ID '$id', but the resource with that ID is of type '${_resources[id].runtimeType}");
+    }
+
+    return _resources[id] as T;
+  }
+
+  /// Finds an object of type [T] within this [EditContext], which is identified by the given [id], or
+  /// returns `null` if no such object is in this [EditContext].
+  T? findMaybe<T extends Editable>(String id) {
+    return _resources[id] as T?;
```
angelosilvestre, line 309: If the object isn't `null`, we should check its type, as in `find`.

super_editor/lib/src/core/editor.dart
```diff
@@ -0,0 +1,1050 @@
...
+
+    // Set the active command to the next command in the backlog.
+    _activeCommand = _commandBacklog.removeAt(0);
+  }
+
+  EditCommand? get activeCommand => _activeCommand;
+
+  void expandActiveCommand(List<EditCommand> replacementCommands) {
+    _activeCommandExpansionQueue.addAll(replacementCommands);
+  }
```
angelosilvestre, line 361: Why is the parameter named `replacementCommands` if we are adding the contents of this list to `_activeCommandExpansionQueue`?

super_editor/lib/src/core/editor.dart
```diff
@@ -0,0 +1,1050 @@
...
+
+/// A [DocumentChange] that impacts a single, specified [DocumentNode] with [nodeId].
+abstract class NodeDocumentChange implements DocumentChange {
+  String get nodeId;
+}
+
+/// A new [DocumentNode] was inserted in the [Document].
+class NodeInsertedEvent implements NodeDocumentChange {
+  const NodeInsertedEvent(this.nodeId);
+
```
angelosilvestre, line 827: Would it make sense to include the index where the node was inserted?

angelosilvestre (commented):
I left a few comments.

super_editor/lib/src/core/editor.dart
matthew-carroll: It shouldn't make a difference. Listeners aren't supposed to submit additional requests. In the future we might add an explicit tracker for that and blow up if a listener does try to do that.

matthew-carroll:
This PR has been up for a little while. I've made some minor adjustments based on feedback. Now I'm going to merge in so that we can move to the breaking changes.

PR #1089, opened 2023-04-18
Move golden tools to /tool directory (Resolves #1085)

### LINKED ISSUE #1085: Move golden tools to /tools directory
Recently we adjusted how goldens are generated. We now run a Dart script that runs Docker on a Linux image.

This tooling was added under `/bin`. 

This new tooling requires certain dependencies that aren't needed by our package runtimes, only by developers at build time. Pub won't let us publish the package if `/bin` dependencies are under `dev_dependencies`.

For this ticket, move the `/bin` tools to a `/tools` directory so that we don't need to add CLI deps to the package.

Also, checkout the `sidekick` package and see if there's any reason for us to integrate that in `super_editor`.

angelosilvestre:
Move golden tools to `/tool` directory. Resolves #1085

This PR moves the existing golden tools from /bin to /tool, to avoid issues with dependencies when releasing the packages.

I used `/tool` instead of `/tools` because this is the convention described in https://dart.dev/tools/pub/package-layout#public-tools

Now we will run the command as the following:

```
flutter pub run tool/goldens test
```

For these simples tasks I think we don't need to integrate the sidekick package.


matthew-carroll (approved):
LGTM

PR #1092, opened 2023-04-20
Make super_text_layout releasable again (Resolves #1083)

### LINKED ISSUE #1083: [SuperText] - Make super_text_layout releasable again
We can't release `super_text_layout` right now because Pub says our new golden test dependencies under `dev_dependencies` have to into regular `dependencies`, instead. But we don't want to ship with those dependencies because they're unrelated to actual package behavior.

Figure out how to get around this and then publish a new version.

matthew-carroll:
Make super_text_layout releasable again (Resolves #1083)

I added `flutter_test` to the regular package deps. I'd prefer not to, but we ship an "inspector", which uses test inspection APIs.

I also released v0.1.5 of `super_text_layout` with the code from this PR.


angelosilvestre (commented):
LGTM

PR #1097, opened 2023-04-24
[SuperEditor][DemoApp] Fix overlayController configuration on mobile editing demos (Resolves #1095)

### LINKED ISSUE #1095: [SuperEditor] the SuperEditor demo on iOS is using MagnifierAndToolbarController properly
The instance of `MagnifierAndToolbarController` is not passed to `SuperEditor` as `overlayController` in the "Mobile Editing - iOS" demo.

angelosilvestre:
[SuperEditor][DemoApp] Fix `overlayController` configuration on mobile editing demos. Resolves #1095

In the "Mobile Editing - iOS" demo, the `overlayController` was not being is not been passed to `SuperEditor`.  This causes a crash when we try to select something, because we access `_overlayController.toolbarTopAnchor!` when we build the toolbar.

This PR changes the iOS and Android demos to pass `overlayController` to `SuperEditor` to fix the crash. It also adds methods to copy, paste and cut, so we properly call `hideToolbar` on the `overlayController`.


matthew-carroll:
@angelosilvestre looks like we have failing goldens

matthew-carroll:
@miguelcmedeiros can you confirm, as per the ticket, that you just wanted a demo fix, and that there wasn't an actual package issue that's getting in the way for Superlist?

miguelcmedeiros:
@matthew-carroll the ticket was only about the demo app. Because of this issue I didn't check if the there was any issue with mobile vs. SuperEditor vs. selection...

matthew-carroll (approved):
LGTM

matthew-carroll:
@angelosilvestre looks like we still have golden failures

angelosilvestre:
@matthew-carroll It seems Flutter master branch broke the goldens slightly again. I updated the goldens.

PR #1099, opened 2023-04-24
[SuperTextField][iOS] Fix toolbar arrow direction (Resolves #1094)

### LINKED ISSUE #1094: [SuperTextfield] Clipboard toolbar on iOS is incorrect
Steps to reproduce:
* Run the demo app and go to "Super iOS Textfield" or "Super Android Textfield"
* Use "single line" configuration 
* Select some text

**Expected**: there's a visual indicator at the bottom side of the toolbar that indicates that selected content is below the toolbar
**Actual**: the indicator is at the top of the toolbar
Simulator Screenshot - iPhone 14 Pro Max - 2023-04-24 at 15 56 15

angelosilvestre:
[SuperTextField][iOS] Fix toolbar arrow direction. Resolves #1094

On iOS, the textfield toolbar is always pointing up, even when it's positioned above the text field.

The cause is that we are passing local offsets to `overlayController`, which expects global offsets.

This PR changes the iOS editing controls to pass global offsets and adds golden tests.

I noticed we have golden tests for super_textfield in the "test" dir instead of "test_goldens" (it seems it was my fault).  I moved those tests to "test_goldens" dir.


super_editor/test_goldens/super_textfield/super_textfield_toolbar_test.dart
```diff
@@ -0,0 +1,88 @@
...
+/// Pumps a widget tree which displays the [child] inside a [Stack].
+Future<void> _pumpSuperTextfieldToolbarTestApp(
+  WidgetTester tester, {
+  required Widget child,
+}) async {
+  await tester.pumpWidget(
+    MaterialApp(
+      debugShowCheckedModeBanner: false,
+      home: Scaffold(
+        body: Stack(
```
matthew-carroll, line 82: Why are we using a `Stack` to display a single `child`?

super_editor/test_goldens/super_textfield/super_textfield_toolbar_test.dart
```diff
@@ -0,0 +1,88 @@
...
+
+      // Double tap to select "pointing".
+      await tester.doubleTapAtSuperTextField(6);
+
+      await screenMatchesGolden(tester, 'super_textfield_ios_toolbar_pointing_up');
+    });
+  });
+}
+
+Widget _buildSuperTextField({
```
matthew-carroll, line 47: Let's move this below the other build method so that they're declared in the same order that they're used within the tests, for easier reading.

super_editor/test_goldens/super_textfield/super_textfield_toolbar_test.dart
```diff
@@ -0,0 +1,88 @@
...
+        child: Positioned(
+          bottom: 50,
+          child: _buildSuperTextField(
+            text: 'Arrow pointing down',
+            configuration: SuperTextFieldPlatformConfiguration.iOS,
+          ),
+        ),
+      );
+
+      // Double tap to select "pointing".
```
matthew-carroll, line 23: Let's write comments that relay the relevant information, rather than merely the literal information:

```dart
// Select a word so that the popover toolbar appears.
```

super_editor/test_goldens/super_textfield/super_textfield_toolbar_test.dart
angelosilvestre: It was just to use the `Positioned` widget to place the textfield at the bottom of the screen.

matthew-carroll (approved):
LGTM

PR #1101, opened 2023-04-25
[SuperTextField][Mobile] Fix scrolling when changing the selection to a visible position (Resolves #1096)

### LINKED ISSUE #1096: [SuperTextfield] On mobile, changing selection on a multi-line field can scroll the content
Steps to reproduce:
* Run the demo app and go to "Super iOS Textfield" or "Super Android Textfield"
* Use "short" configuration
* Place the selection at the end of the content
* Tap on the content, but no the last visible line

**Expected**: caret is placed where the tap happened and there is no scrolling
**Actual**: content scrolls so that line with the caret is placed at the bottom of the visible content

https://user-images.githubusercontent.com/31278849/234040900-ba2d20e6-a826-474d-9a6e-95a5c0826337.mov

angelosilvestre:
[SuperTextField][Mobile] Fix scrolling when changing ths selection to a visible position. Resolves #1096

When changing the selection in a multi-line textfield, if the selected position offset is greater than the text field height the text field is being scrolled, even if this offset is already visible on screen:

https://user-images.githubusercontent.com/31278849/234040900-ba2d20e6-a826-474d-9a6e-95a5c0826337.mov

The issue is that we are only comparing the offset with the viewport height, ignoring the fact that we might already have scrolled to a position that makes the selection visible.

This PR changes the comparison to consider the current scroll offset.

I tried using the robot to change the selection, but it doesn't seem to work when the textfield scrolls. 

I copied `_findInnerPlatformTextField` from `SuperTextFieldRobot`. Is there any shared file where we could de-dup this?


super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -995,7 +995,7 @@ class TextScrollController with ChangeNotifier {
         // Scroll the content down.
         _scrollOffset = rect.top;
         _log.finer(' - updated _scrollOffset to $_scrollOffset');
-      } else if (rect.bottom > _delegate!.viewportHeight!) {
+      } else if ((rect.bottom - _scrollOffset) > _delegate!.viewportHeight!) {
```
matthew-carroll, line 998: We should rename this `rect` and the incoming `extentCharacterRect` to reflect the space that the rectangle sits in.

`rectInContentSpace`, `extentCharacterRectInContentSpace`

This way we hopefully won't overlook details like this in the future.

Also, can you please confirm in the `_delegate` that this rect definitely is in content space, and not viewport space? Related to that, if the rect is in content space, then how is `rect.top < 0` working? If `rect` is in content space, then `rect.top` would never be less than zero. So something seems inconsistent here.

matthew-carroll:
> I tried using the robot to change the selection, but it doesn't seem to work when the textfield scrolls.

When saying that something doesn't work, please describe exactly what isn't working. What did you try? What was expected? What was the actual result?

I can't review a decision to replicate behavior if I can't see exactly what issue you're sidestepping.

angelosilvestre:
@matthew-carroll 

`placeCaretInSuperTextField` calls a method that contains:

```dart
if (!textFieldBox.size.contains(adjustedOffset)) {  
  print("Couldn't tap at $adjustedOffset in text field with size ${textFieldBox.size}");
  return false;
}
```

This code doesn't take into account the cases where the content height is bigger than the textfield height.

matthew-carroll:
@angelosilvestre that explanation still requires me to go read some source code to contextualize what you're saying. How are you trying to use the tool, what's the expected output, what's the actual output?

angelosilvestre:
I'm trying to place the selection at the end of the text. This offset isn't visible, because the content height is bigger than the viewport height.

I call `tester.placeCaretInSuperTextField(offset)`.

Expected result: the selection is placed at `offset`

Actual result: An exception is thrown with the message `"The desired text offset wasn't tappable in SuperTextField: $offset"`

The exception also happens if we already scrolled to the end.

matthew-carroll:
Ok, that's a much more helpful description. Thanks.

I'll think about that.

matthew-carroll:
> I copied _findInnerPlatformTextField from SuperTextFieldRobot. Is there any shared file where we could de-dup this?

Do we have an `Inspector` for `SuperTextField`? If so, we should probably have the `Inspector` answer questions about current scroll offset, max scroll offset, etc, rather than implement inspection behavior in the robot or other tests.

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
angelosilvestre: The rect is in content space, we get it from: 
```dart
final extentCharacterRectInContentSpace =  _delegate!.getCharacterRectAtPosition(TextPosition(offset: characterIndex));
 _ensureRectIsVisible(extentCharacterRectInContentSpace);
```

 `getCharacterRectAtPosition` just calls `_textLayout.getCharacterBox`: 

```dart
@override
  Rect getCharacterRectAtPosition(TextPosition position) {
    return _textLayout.getCharacterBox(position)?.toRect() ?? Rect.fromLTRB(0, 0, 0, _textLayout.estimatedLineHeight);
  }
```

> Related to that, if the rect is in content space, then how is rect.top < 0 working?

Indeed, this doesn't seem to be working.

matthew-carroll: >  `getCharacterRectAtPosition` just calls `_textLayout.getCharacterBox`

What if there's padding in the text field? Does everything still work?

super_editor/test/super_textfield/super_textfield_scroll_test.dart
```diff
@@ -68,6 +69,78 @@ void main() {
       expect(textBottom, lessThanOrEqualTo(viewportBottom));
     });
 
+    testWidgetsOnAllPlatforms("multi-line jump scroll position vertically up when selection extent changes",
```
matthew-carroll, line 72: Please rename test to: "multi-line jumps scroll position vertically when selection extent moves above or below the visible viewport area"

super_editor/test/super_textfield/super_textfield_scroll_test.dart
```diff
@@ -147,4 +221,12 @@ Future<void> _pumpTestApp(
...
     ),
   );
+
+  // The textfield scrolling takes into account its viewport height to compute the scroll offset.
+  //
+  // In the first frame, it's possible that we get zero as the viewport height. This causes our scroll offset
+  // to be wrong.
+  //
+  // Request another frame, so we can get the correct viewport height.
+  await tester.pumpAndSettle();
```
matthew-carroll, line 231: > // Request another frame...

You're requesting many frames. Can this be a `pump()`?

Also, let's reword this comment to be more concise:

```dart
// The first frame might have a zero viewport height. Pump a second frame to account for the final viewport size.
```

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -196,6 +196,9 @@ class _TextScrollViewState extends State<TextScrollView>
   @override
   double get startScrollOffset => 0.0;
 
+  @override
+  EdgeInsets? get padding => widget.padding;
```
matthew-carroll, line 200: Does this need to be public?

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -981,36 +984,44 @@ class TextScrollController with ChangeNotifier {
     final characterIndex = _textController.selection.extentOffset >= _textController.text.text.length
         ? _textController.text.text.length - 1
         : _textController.selection.extentOffset;
-    final extentCharacterRect = _delegate!.getCharacterRectAtPosition(TextPosition(offset: characterIndex));
-    _ensureRectIsVisible(extentCharacterRect);
+
+    final padding = _delegate!.padding ?? const EdgeInsets.all(0.0);
```
matthew-carroll, line 988: There's a `widget.padding`, which you also exposed as `padding`, and now there's `_delegate!.padding`? Which padding is the real padding? Or am I looking at two different classes? I can't tell with the way that GitHub is showing the code.

Also, my initial reaction is that I don't think we want to expose arbitrary layout details like padding. What happens if there's also a border, or margin, or anything else that impacts position? It seems to me that the `getCharacterRectAtPosition` should just return the actual character rect. To me, this looks like pieces of the layout implementation are being pulled into the scrolling system.

angelosilvestre: > Or am I looking at two different classes?

There are different classes

Updated `getCharacterRectAtPosition` to return the adjusted rect and removed the padding from the delegate.

super_editor/test/super_textfield/super_textfield_scroll_test.dart
```diff
@@ -107,6 +107,43 @@ void main() {
...
+
+      // Place the caret at the beginning of the text.
+      controller.selection = const TextSelection.collapsed(offset: 0);
+      await tester.pumpAndSettle();
+
+      // Ensure the text field scrolled to the beginning of the text.
+      expect(
+        SuperTextFieldInspector.findScrollOffset(),
+        10.0,
+      );
```
matthew-carroll, line 144: I don't see anything in this test that shows what it has to do with padding. If the test of padding is incidental, then please update the comments to focus on what we're proving with this behavior.

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -1185,15 +1185,29 @@ class SuperTextFieldScrollviewState extends State<SuperTextFieldScrollview> with
     const gutterExtent = 0; // _dragGutterExtent
     final extentLineIndex = (extentOffset.dy / widget.estimatedLineHeight).round();
 
+    final firstCharY = _textLayout.getCharacterBox(const TextPosition(offset: 0))?.top ?? 0.0;
+    final lastCharY =
+        _textLayout.getCharacterBox(TextPosition(offset: widget.textController.text.text.length - 1))?.top ??
```
matthew-carroll, line 1190: Shouldn't we use the `bottom` of the last character, not the `top`?

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -1185,15 +1185,29 @@ class SuperTextFieldScrollviewState extends State<SuperTextFieldScrollview> with
...
+    final isAtLastLine = extentOffset.dy == lastCharY;
+
+    // If we are at the first or last line, add some space between the line
+    // and the textfield's edge.
+    final extraSpace = isAtFirstLine || isAtLastLine ? _textLayout.getLineHeightAtPosition(selection.extent) / 2 : 0;
+
     final myBox = context.findRenderObject() as RenderBox;
-    final beyondTopExtent = min<double>(extentOffset.dy - widget.scrollController.offset - gutterExtent, 0).abs();
+    final beyondTopExtent =
+        min<double>(extentOffset.dy - widget.scrollController.offset - gutterExtent - extraSpace, 0).abs();
```
matthew-carroll, line 1202: Please format this line to match the beyond bottom extent line

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -1185,15 +1185,29 @@ class SuperTextFieldScrollviewState extends State<SuperTextFieldScrollview> with
...
+    final lastCharY =
+        _textLayout.getCharacterBox(TextPosition(offset: widget.textController.text.text.length - 1))?.top ??
+            _textLayout.estimatedLineHeight;
+
+    final isAtFirstLine = extentOffset.dy == firstCharY;
+    final isAtLastLine = extentOffset.dy == lastCharY;
+
+    // If we are at the first or last line, add some space between the line
+    // and the textfield's edge.
+    final extraSpace = isAtFirstLine || isAtLastLine ? _textLayout.getLineHeightAtPosition(selection.extent) / 2 : 0;
```
matthew-carroll, line 1198: I don't think there's any reason to combine the first and last line possibilities. It's kind of confusing to read, because we know that only one of those conditions can ever be true at one time. I think I would move the conditional into the extent calculations below: `isAtFirstLine ? _textLayout.getLineHeightAtPosition(selection.extent) / 2 : 0.0`

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -1185,15 +1185,29 @@ class SuperTextFieldScrollviewState extends State<SuperTextFieldScrollview> with
     const gutterExtent = 0; // _dragGutterExtent
     final extentLineIndex = (extentOffset.dy / widget.estimatedLineHeight).round();
 
+    final firstCharY = _textLayout.getCharacterBox(const TextPosition(offset: 0))?.top ?? 0.0;
```
matthew-carroll, line 1188: Can you confirm why we need this method? As I'm reading through the PR, I noticed that the delegate has a method to ensure that a rect is visible. Is there a reason that we don't just calculate a rect and send it to the delegate?

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -981,36 +990,51 @@ class TextScrollController with ChangeNotifier {
...
+      final firstCharRect = _delegate!.getCharacterRectAtPosition(const TextPosition(offset: 0));
+      final lastCharRect =
+          _delegate!.getCharacterRectAtPosition(TextPosition(offset: _textController.text.text.length - 1));
+
+      final isAtFirstLine = rectInContentSpace.top == firstCharRect.top;
+      final isAtLastLine = rectInContentSpace.top == lastCharRect.top;
+
+      // If we are at the first or last line, add some space between the line
+      // and the textfield's edge.
+      final extraSpace = isAtFirstLine || isAtLastLine ? rectInContentSpace.height / 2 : 0;
```
matthew-carroll, line 1014: Same point here about this calculation. I think it would be easier to read if we clearly differentiate between top and bottom calculations.

super_editor/test_goldens/super_textfield/goldens/super_textfield_scrolled_down_android.png
matthew-carroll, line 1: It's not obvious what details we care about in these goldens. At a minimum, can you make the text field not as wide as the parent, center it, add a border around the viewport? Also, can you take a few minutes to see if it's easy for us to use the golden toolkit to render multiple of these images, together, along with a brief description of each one?

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
angelosilvestre: I think we should use the `top`, because the `extentOffset`  refers to the top-left. I made an adjustment to use zero when  `_textLayout.getCharacterBox` returns  `null`.

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
angelosilvestre: The desktop implementation doesn't use the same `TextScrollView`. The desktop textfield doesn't use the delegate.

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
matthew-carroll: I didn't understand that explanation. We're trying to determine the bottom most y-value that should be visible, right? I mean, imagine that only the top pixel of a line were visible - that means that nearly the entire line is below the bottom of the viewport, right?

Correct me if I'm wrong, but for the top line we essentially do "topCharacter.top - halfLineHeight", right?

I would expect for the bottom would do "bottomCharacter.bottom + halfLineHeight".

Am I getting something wrong?

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -1185,15 +1185,29 @@ class SuperTextFieldScrollviewState extends State<SuperTextFieldScrollview> with
...
+    final lastCharY =
+        _textLayout.getCharacterBox(TextPosition(offset: widget.textController.text.text.length - 1))?.top ?? 0.0;
+
+    final isAtFirstLine = extentOffset.dy == firstCharY;
+    final isAtLastLine = extentOffset.dy == lastCharY;
```
matthew-carroll, line 1193: Let's also separate these lines based on where they're used. Notice below that there's clearly one line that deals with top, followed by a line that deals with bottom.

It seems to me that we might as well calculate `firstCharY` and `IsAtFirstLine`, right before we need them. And same for the bottom lines. We should also now add a blank line between the top calculation and the bottom because each calculation is now more than one line long.

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -981,36 +990,49 @@ class TextScrollController with ChangeNotifier {
...
-    _log.finer('Ensuring rect is visible: $rect');
+    _log.finer('Ensuring rect is visible: $rectInContentSpace');
     if (_delegate!.isMultiline) {
-      if (rect.top < 0) {
+      final firstCharRect = _delegate!.getCharacterRectAtPosition(const TextPosition(offset: 0));
+      final lastCharRect =
+          _delegate!.getCharacterRectAtPosition(TextPosition(offset: _textController.text.text.length - 1));
+
+      final isAtFirstLine = rectInContentSpace.top == firstCharRect.top;
+      final isAtLastLine = rectInContentSpace.top == lastCharRect.top;
```
matthew-carroll, line 1010: Same comment here about breaking up top and bottom line calculations.

super_editor/lib/src/infrastructure/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -981,36 +990,49 @@ class TextScrollController with ChangeNotifier {
...
+
+      if (rectInContentSpace.top - (isAtFirstLine ? rectInContentSpace.height / 2 : 0) - _scrollOffset < 0) {
         // The character is entirely or partially above the top of the viewport.
         // Scroll the content down.
-        _scrollOffset = rect.top;
+        _scrollOffset = max(rectInContentSpace.top - (isAtFirstLine ? rectInContentSpace.height / 2 : 0), 0);
         _log.finer(' - updated _scrollOffset to $_scrollOffset');
-      } else if (rect.bottom > _delegate!.viewportHeight!) {
+      } else if (rectInContentSpace.bottom - _scrollOffset + (isAtLastLine ? rectInContentSpace.height / 2 : 0) >
```
matthew-carroll, line 1017: I guess in this case, given that we're using the same calculation, for the same purpose, multiple times, lets go ahead and assign it to a variable like you originally had it. But a different variable for top vs bottom.

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
angelosilvestre: But we only use `lastCharY` to determine if we are at the last line of the textfield, not to determine the rect that should be visible.

super_editor/test_goldens/super_textfield/goldens/super_textfield_scrolled_down_android.png
angelosilvestre: Updated to use the golden toolkit.

super_editor/test_goldens/super_textfield/goldens/super_textfield_scrolled_down.png
matthew-carroll, line 1: These descriptions state what's painted, but not what we're testing. Why do we care that we're at the end? If a developer eventually got a different golden result, and had to re-evaluate it, what should they look for?

Also, should there a caret visible in here?

angelosilvestre: Updated.

The caret isn't visible because we are not focusing the textfield. Also, using the robot to place the selection is throwing an exception when the textfield is scrollable, as we mentioned before.

matthew-carroll: Remind me how this situation is happening to users, if we can't simulate it with user interaction in the test?

angelosilvestre: Well, this test was added to ensure the textfield is being displayed correctly when we scroll it to make the selected line visible. This also happens if the selection is changed without direct interaction with the textfield...

matthew-carroll: What if you alter the test to place the caret where visible, and then press the down arrow enough times to get to the end, and then press up enough times to get back to the start? Then we'd test real user behavior, and we'd see the caret...

angelosilvestre: As now we are using the three text fields in the same golden test, when we focus one textfield to press down it will unfocus the others. When `SuperTextField` loses focus it clears the selection...

matthew-carroll (approved):
LGTM

PR #1103, opened 2023-04-25
[SuperEditor] Avoid restoring selection upon re-focus when selected nodes were deleted (Resolves #1074)

### LINKED ISSUE #1074: [SuperEditor] Don't restore selection on refocus if selection nodes disappeared from document
I have the following use case:
1. I have several nodes in a document,
2. I select some of the nodes,
3. I open a menu that takes focus from `SuperEditor`, and use a menu item to remove the selected nodes.
4. The menu closes and focus goes back to `SuperEditor`.

Expected outcome:
- The editor is focused with no selection 👍 .

Actual outcome:
- The editor crashes 💥 .

angelosilvestre:
[SuperEditor] Avoid restoring selection upon re-focus when selected nodes were deleted. Resolves #1074

`EditorSelectionAndFocusPolicy` restores the previous selection when the editor re-gains focus. If the node at the selection base or extent was removed while the editor didn't have focus, restoring the selection causes the editor to crash, because we are applying a selection with an inexistent node.

This PR changes `EditorSelectionAndFocusPolicy` to check if the nodes at base and offset are still present before restoring the selection.

### DIFF
--- a/super_editor/lib/src/default_editor/document_focus_and_selection_policies.dart
+++ b/super_editor/lib/src/default_editor/document_focus_and_selection_policies.dart
@@ -8,6 +8,7 @@ class EditorSelectionAndFocusPolicy extends StatefulWidget {
   const EditorSelectionAndFocusPolicy({
     Key? key,
     required this.focusNode,
+    required this.document,
     required this.selection,
     required this.isDocumentLayoutAvailable,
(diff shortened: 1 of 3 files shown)


matthew-carroll (approved):
LGTM

PR #1107, opened 2023-04-30
[SuperEditor][SuperTextField] Fix emoji panel positioning (Resolves #1105)

### LINKED ISSUE #1105: Emoji & Symbols panel doesn't snap to caret position
Steps to reproduce:
* Place the selection on a `SuperEditor` or `SuperTextField`
* Press CONTROL+COMMAND+SPACE shortcut

**Expected:** opens the Emoji & Symbols panel close to the caret position

Screenshot 2023-04-27 at 09 32 40

**Actual:** opens the Emoji & Symbols panel in the middle of the window

Screenshot 2023-04-27 at 09 33 30

angelosilvestre:
[SuperEditor][SuperTextField] Fix emoji panel positioning. Resolves #1105

Currently, when using the CONTROL+COMMAND+SPACE shortcut, the emoji & symbols panel is always displayed at the middle of the screen, instead of near the selection.

The reason is that we need to call the following methods on the input connection to position this panel:

- `setEditableSizeAndTransform`: This method tells the IME our size and how to map our coordinates to the root node.
- `setCaretRect`:  This method tells the IME the caret rect, in local coordinates.

This PR changes `SuperEditorImeInteractor` and introduces a new class to add this behavior to `SuperTextField`. 

This PR is based on Flutter's `EditableText` implementation. However, in `EditableText`, these methods are called at every frame and why this is needed isn't documented. This PR implementation calls these methods when the text, selection or focus changes and seems to work correctly.

https://user-images.githubusercontent.com/7597082/235372537-942d056f-72d4-4e3d-a211-db6ad2cbaf54.mov

https://user-images.githubusercontent.com/7597082/235372563-0d657c8a-3af6-4592-9af5-e4ff69da511c.mov


matthew-carroll:
@angelosilvestre what happens in if you scroll the content with the popover displayed? Does it stay with the caret?

matthew-carroll:
@miguelcmedeiros would you like to try this PR with Superlist, before I review the specifics in the code?

angelosilvestre:
@matthew-carroll Scrolling causes the panel to disappear. This is the same behavior from `EditableText`.

miguelcmedeiros:
> would you like to try this PR with Superlist, before I review the specifics in the code?

@matthew-carroll  I would but Superlist is pointing at `stable` while this PR is points at `main`. 
I'll test it once it's retargeted or after being available in `stable`.

matthew-carroll:
@angelosilvestre can you duplicate this PR on another branch that targets stable so that Superlist can try this before we merge?

angelosilvestre:
@matthew-carroll  Created https://github.com/superlistapp/super_editor/pull/1108

matthew-carroll:
@miguelcmedeiros can you try #1108 ?

miguelcmedeiros:
@angelosilvestre I tested it out and found the following issues:
* Single line SuperTextField with panel below it, it anchors it at the top of the caret
Screenshot 2023-05-02 at 09 27 02
* In SuperEditor it does not update the anchor offset if the content is scrolled

https://user-images.githubusercontent.com/31278849/235618278-50ee5be2-416c-4039-a5c4-3f518c8305ec.mov

* If the window is moved and the editor or field does not lose focus, then the panel is not anchored to the window at all

angelosilvestre:
@miguelcmedeiros Could you please try https://github.com/superlistapp/super_editor/pull/1108 again?

angelosilvestre:
@matthew-carroll The `SuperTextField` issue was an easy fix. However, we have bigger problems.

The issue is that scrolling changes the caret position on screen, but the document itself didn't change, so we aren't notified. We could listen to scroll notifications, but there will be many other problems to solve...

There are many other operations the influence the caret position, like resizing the screen, resizing the editor, moving the editor around the screen, a document component changing its size, etc. I think we will have the same issues we faced in https://github.com/superlistapp/super_editor/issues/1022

We could try use the document overlay, but I'm not sure if it will work for all the cases. We would also need a solution for `SuperTextField`.

I think I understand now why Flutter's `EditableText` calls `setEditableSizeAndTransform` and `setCaretRect` at every frame...

Changing the implementation to call `setEditableSizeAndTransform` and `setCaretRect` at every frame, as `EditableText` does, solves the issue. It doesn't look like a good solution, but I don't know what other options we have right now.

Please let me know what you think about this.

matthew-carroll:
Updating on every frame is probably the least bad solution right now. Please include appropriate comments that explain why we're doing it so that we can undo it in the future, if possible.

angelosilvestre:
@matthew-carroll Already added some comments in the appropriate places.

miguelcmedeiros:
@angelosilvestre tested it again and those issues are fixed. No new issues found.
🙌

matthew-carroll:
@angelosilvestre do you feel that this is ready for review? Is all the code cleaned up, commented, and tested?

angelosilvestre:
@matthew-carroll I think it's ready for review.

I'm just not sure how to write tests for this. I could write a test that checks we are calling the necessary IME methods, but I don't think there will be much value in a test like this.

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
```diff
@@ -147,6 +147,10 @@ class SuperEditorImeInteractorState extends State<SuperEditorImeInteractor> impl
       floatingCursorController: widget.floatingCursorController,
     );
 
+    widget.editContext.editor.document.addListener(_scheduleUpdateImeVisualInformation);
```
matthew-carroll, line 150: It seems like there's a lot of extra stuff going on here just to update on every frame. Is there any reason that we don't just call a single method to start reporting viewport and caret info, and then let that method keep running itself? All these listeners seem superfluous, and also splitting the reporting into two separate methods looks like it's  just adding more accounting. If we're gonna report both details every frame, we might as well do it, together.

super_editor/lib/src/infrastructure/super_textfield/infrastructure/ime_controls.dart
```diff
@@ -0,0 +1,164 @@
+import 'package:flutter/material.dart';
+
+import 'package:super_editor/src/infrastructure/super_textfield/super_textfield.dart';
+import 'package:super_text_layout/super_text_layout.dart';
+
+/// A widget that positions the OS emoji & symbols toolbar at the text field's selection.
+class SuperTextFieldImeControls extends StatefulWidget {
```
matthew-carroll, line 7: Is there any reason that we don't have the viewport report its bounds each frame, and have the caret layer report its bounds each frame? It seems like we're adding an extra widget, and extra complexity elsewhere, to do what they could do directly...

angelosilvestre: I added a new widget because I thought we could need to do similar things on the Android and iOS text fields, but I guess we only need this on desktop, so we could move the implementation to the desktop textfield widget

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
angelosilvestre: Updated

super_editor/lib/src/infrastructure/super_textfield/android/android_textfield.dart
```diff
@@ -473,52 +473,52 @@ class SuperAndroidTextFieldState extends State<SuperAndroidTextField>
...
+          child: AndroidTextFieldTouchInteractor(
+            focusNode: _focusNode,
             textKey: _textContentKey,
-            textEditingController: _textEditingController,
-            textAlign: widget.textAlign,
-            minLines: widget.minLines,
-            maxLines: widget.maxLines,
-            lineHeight: widget.lineHeight,
-            perLineAutoScrollDuration: const Duration(milliseconds: 100),
+            textFieldLayerLink: _textFieldLayerLink,
```
matthew-carroll, line 479: With this diff I can't really tell what's changing here?

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
```diff
@@ -202,6 +202,7 @@ class SuperEditorImeInteractorState extends State<SuperEditorImeInteractor> impl
     } else {
       _configureImeClientDecorators();
       _documentImeConnection.value = _documentImeClient;
+      _updateImeVisualInformation();
```
matthew-carroll, line 205: Is there a reason that we don't call this method `_startReportingBoundsToIme()` and then in the if-statement branch above have a `_stopReportingBoundsToIme()`?

It's probably also a good idea to call stop from `dispose()`, even though there's already a check for attachment in the method. It's generally a good idea for developers to explicitly show when things are expected to stop.

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
```diff
@@ -214,6 +215,62 @@ class SuperEditorImeInteractorState extends State<SuperEditorImeInteractor> impl
     _imeClient.client = widget.imeOverrides ?? _documentImeClient;
   }
 
+  /// Update our size, transform to the root node coordinates, and caret rect on the IME.
```
matthew-carroll, line 218: We're not really updating "our size" or any part of us. We're reporting information to the IME.

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
```diff
@@ -214,6 +215,62 @@ class SuperEditorImeInteractorState extends State<SuperEditorImeInteractor> impl
...
+  /// This is needed to display the OS emoji & symbols panel at the editor selected position.
+  ///
+  /// This methods is re-scheduled to run at the end of every frame while we are attached to the IME.
+  void _updateImeVisualInformation() {
+    if (!isAttachedToIme) {
+      return;
+    }
+
+    final renderBox = context.findRenderObject() as RenderBox;
+    _imeConnection.value!.setEditableSizeAndTransform(renderBox.size, renderBox.getTransformTo(null));
```
matthew-carroll, line 229: Do you actually need to explicitly pass `null` in `getTransformTo()`? Is that not the default value?

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
```diff
@@ -214,6 +215,62 @@ class SuperEditorImeInteractorState extends State<SuperEditorImeInteractor> impl
...
+    WidgetsBinding.instance.addPostFrameCallback((timeStamp) {
+      _updateImeVisualInformation();
+    });
+  }
+
+  /// Compute the caret rect in the editor's content space.
+  ///
+  /// Returns `null` if we don't have a selection or if we can't get the caret rect
+  /// from the document layout.
+  Rect? _computeCaretRectInContentSpace() {
```
matthew-carroll, line 249: At the end of this method you say you're mapping to viewport space. If that's true, then this method isn't computing a rect in "content space".

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -958,6 +964,8 @@ class _SuperTextFieldImeInteractorState extends State<SuperTextFieldImeInteracto
     super.initState();
     widget.focusNode.addListener(_updateSelectionAndImeConnectionOnFocusChange);
 
+    widget.textController.onImeConnectionChange = _updateImeVisualInformation;
```
matthew-carroll, line 967: How do you know that someone else didn't already register this callback? If they did, you're silently disconnecting it.

super_editor/lib/src/infrastructure/super_textfield/input_method_engine/_ime_text_editing_controller.dart
```diff
@@ -36,10 +36,12 @@ class ImeAttributedTextEditingController extends AttributedTextEditingController
     void Function(RawFloatingCursorPoint)? onIOSFloatingCursorChange,
     Brightness keyboardAppearance = Brightness.light,
     TextInputConnectionFactory? inputConnectionFactory,
+    VoidCallback? onImeConnectionChange,
```
matthew-carroll, line 39: Taking in a single callback is generally a mistake, for the reason I mentioned above, where you use it.

At a minimum, you should be able to add and remove listeners for this purpose.

How have we managed to avoid listening for IME connection changes up to this point? Does nothing else care about that in our current code?

super_editor/lib/src/infrastructure/super_textfield/input_method_engine/_ime_text_editing_controller.dart
```diff
@@ -173,6 +190,16 @@ class ImeAttributedTextEditingController extends AttributedTextEditingController
     _inputConnection?.close();
   }
 
+  /// Sends the text field size and tranform to the root coordinates to the IME.
+  void setEditableSizeAndTransform(Size editableBoxSize, Matrix4 transform) {
+    _inputConnection?.setEditableSizeAndTransform(editableBoxSize, transform);
```
matthew-carroll, line 195: How about we expose the input connection publicly, and then avoid adding the listener, and avoid adding these methods?

With this pattern, we could end up replicating any number of Flutter IME methods, and I'm not sure there's any reason for us to do so.

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
angelosilvestre: We need to pass `null` because `getTransformTo` doesn't have a default value.

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
angelosilvestre: Renamed.

super_editor/lib/src/infrastructure/super_textfield/input_method_engine/_ime_text_editing_controller.dart
angelosilvestre: We can remove these methods, but I don't think we can avoid the listener, as the `TextInputConnection` itself doesn't expose any listener

angelosilvestre: Do you mean exposing a `ValueNotifier<TextInputConnection?>`  ?

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
angelosilvestre: `_updateImeVisualInformation` reports the information to the IME and then re-schedules itself. I'm not sure what would be the `_stopReportingBoundsToIme` implementation...

super_editor/lib/src/infrastructure/super_textfield/input_method_engine/_ime_text_editing_controller.dart
angelosilvestre: Changed to use a `ValueNotifier<TextInputConnection?>`.

In `SuperTextField`, it seems nothing outside `ImeAttributedTextEditingController` cares about the text input connection.

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
angelosilvestre: Changed to use a ValueNotifier<TextInputConnection?> instead.

super_editor/lib/src/infrastructure/super_textfield/android/android_textfield.dart
angelosilvestre: It seems that just the formatting was changed. I reverted this file.

matthew-carroll:
Where did we leave this? I see that a review was requested, but there's also another PR that targets stable. And there's a merge conflict on this PR.

Do you believe that this code is still ready for a review?

angelosilvestre:
@matthew-carroll Rebase onto main.

I believe this is still ready for review.

The PR targeting main was opened so miguel could test the PR. That PR will be closed in favor of this one.

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -1011,6 +1020,65 @@ class _SuperTextFieldImeInteractorState extends State<SuperTextFieldImeInteracto
...
+    final renderBox = context.findRenderObject() as RenderBox;
+    widget.textController.inputConnectionNotifier.value!
+        .setEditableSizeAndTransform(renderBox.size, renderBox.getTransformTo(null));
+
+    final caretRect = _computeCaretRectInContentSpace();
+    if (caretRect != null) {
+      widget.textController.inputConnectionNotifier.value!.setCaretRect(caretRect);
+    }
+
+    // Without showing the keyboard, the panel is always positioned at the screen center after the first time.
```
matthew-carroll, line 1042: I think this is something we need to investigate. I guess it might not be worth holding up this PR, but it's such a weird requirement that I'm not comfortable leaving this to a comment that says we don't know why. Please file a ticket with ample context and references to both our editor behavior and other mobile app behaviors, as they compare to this weird requirement.

super_editor/lib/src/infrastructure/super_textfield/input_method_engine/_ime_text_editing_controller.dart
```diff
@@ -91,6 +91,11 @@ class ImeAttributedTextEditingController extends AttributedTextEditingController
   }
 
   TextInputConnection? _inputConnection;
+
+  /// Notifies whenever the current [TextInputConnection] changes.
+  ValueNotifier<TextInputConnection?> get inputConnectionNotifier => _inputConnectionNotifier;
```
matthew-carroll, line 96: Should clients be able to change this value? If not, you should expose a `ValueListenable`, not a `ValueNotifier`.

super_editor/lib/src/infrastructure/super_textfield/input_method_engine/_ime_text_editing_controller.dart
```diff
@@ -91,6 +91,11 @@ class ImeAttributedTextEditingController extends AttributedTextEditingController
   }
 
   TextInputConnection? _inputConnection;
+
```
matthew-carroll, line 94: ^ If we have a notifier, is there a reason to also have a standalone `_inputConnection` variable? That seems like an opportunity to get a mismatch between the two.

super_editor/lib/src/infrastructure/super_textfield/input_method_engine/_ime_text_editing_controller.dart
```diff
@@ -167,10 +175,15 @@ class ImeAttributedTextEditingController extends AttributedTextEditingController
     _inputConnection?.close();
 
     _osCurrentTextEditingValue = const TextEditingValue();
+    _inputConnectionNotifier.value = null;
   }
 
   void showKeyboard() {
     _isKeyboardDisplayDesired = true;
+    if (!(_inputConnection?.attached ?? false)) {
```
matthew-carroll, line 183: Any idea how we missed this condition all this time? I would think that if calling `showKeyboard()` when not attached causes an error, we would have seen that by now....

super_editor/lib/src/infrastructure/super_textfield/input_method_engine/_ime_text_editing_controller.dart
angelosilvestre: Updated.

super_editor/lib/src/infrastructure/super_textfield/input_method_engine/_ime_text_editing_controller.dart
angelosilvestre: I think that in most places we are checking if we are attached to the IME before calling`showKeyboard`.

super_editor/lib/src/infrastructure/super_textfield/desktop/desktop_textfield.dart
angelosilvestre: Ok, will do.

matthew-carroll (approved):
LGTM

PR #1122, opened 2023-05-09
Prepare super_editor_markdown for v0.1.4+2 release

matthew-carroll:
Prepare super_editor_markdown for v0.1.4+2 release

This PR copies some changes that I think we only have on `stable`. This is probably because the last release was cut directly from `stable`, so we let `main` fall behind.

After bringing `main` up to date with `stable`, I updated the version number and added to the CHANGELOG.


angelosilvestre (commented):
LGTM

Issue #1127, opened 2023-05-11
ListenableBuilder duplicate prevents building with flutter stable


olavemil:
Despite https://github.com/superlistapp/super_editor/pull/880, `_listenable_builder.dart` itself imports flutter, which currently results in a compile error for newly released Flutter stable, 3.10.

Resolved by chaning to `import 'package:flutter/widgets.dart' hide ListenableBuilder;` on line one.

To reproduce:
1. Have the following in `pubspec.yaml` for any app
```
environment:
  sdk: ">=3.0.0 <4.0.0"
  flutter: ^3.10.0

dependencies:
  flutter:
    sdk: flutter
  super_editor: ^0.2.4
```
2. try building any platform.
3. results in various compile errors 
`Error: 'ListenableBuilder' is imported from both 'package:flutter/src/widgets/transitions.dart' and 'package:super_editor/src/infrastructure/_listenable_builder.dart'`

matthew-carroll:
Can you please describe which versions of `super_editor` are failing to compile?

- v0.2.4-dev.1 on Pub
- v0.2.4 on Pub
- GitHub main branch
- GitHub stable branch

matthew-carroll:
Also, can you please file an issue for this PR so that we have documented history for a fix?

matthew-carroll:
I'm going to close this issue in favor of https://github.com/superlistapp/super_editor/issues/1131 - there are multiple things we need to fix, and we need to run new releases.

PR #1128, opened 2023-05-11
Hide Flutter's ListenableBuilder in _listenable_builder.dart

olavemil:
https://github.com/superlistapp/super_editor/issues/1127

This allows super_editor 0.2.4 to build with flutter 3.10 (current stable)


matthew-carroll:
Thanks for the PR - there are multiple things we need to fix, and we need to run new releases. I'm going to handle all of that in https://github.com/superlistapp/super_editor/issues/1131

PR #1160, opened 2023-06-08
[SuperEditor] - Hash (#) header conversions (Resolves 1159)

matthew-carroll:
[SuperEditor] - Hash (#) header conversions (Resolves 1159)

Supports from 1 hash (#) to 6 hashes (######)

I also added variant support to our custom test methods.

https://github.com/superlistapp/super_editor/assets/7259036/cd9e2d1c-a6d8-465a-ab22-56929012ce6c


miguelcmedeiros:
Would it be possible to define which header levels are supported?

By default it can be all, but let's say an app only wants to support level 1 and 2.

matthew-carroll:
@miguelcmedeiros - I made the header reaction configurable. I also extracted a base class for all paragraph prefix reactions.

miguelcmedeiros (approved):

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -1,14 +1,14 @@
 import 'dart:io';
 
 import 'package:attributed_text/attributed_text.dart';
-import 'package:characters/characters.dart';
 import 'package:collection/collection.dart';
+import 'package:flutter/cupertino.dart';
```
angelosilvestre, line 5: Do we need this import?

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -19,36 +19,130 @@ import 'package:super_editor/src/infrastructure/_logging.dart';
 
 import 'multi_node_editing.dart';
 
+/// Converts a [ParagraphNode] from a regular paragraph to a header when the
+/// user types "# " (or similar) at the start of the paragraph.
+class HeaderConversionReaction extends ParagraphPrefixConversionReaction {
+  static final _headerRegExp = RegExp(r'^#{1,6}\s+$');
```
angelosilvestre, line 25: This regex will match up to 6 hashes, so I think we should either document that the maximum number of levels is 6 or generate the regex depending on `maxLevel`.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -19,36 +19,130 @@ import 'package:super_editor/src/infrastructure/_logging.dart';
...
+    }
+  }
+
+  const HeaderConversionReaction([
+    this.maxLevel = 6,
+    this.mapping = _getHeaderAttributionForLevel,
+  ]);
+
+  /// The highest level of header that this reaction will recognize, e.g., `3` -> "### ".
+  final int maxLevel;
```
angelosilvestre, line 53: It seems we are not using `maxLevel` anywhere, am I missing something?

super_editor/test/super_editor/text_entry/paragraph_conversions_test.dart
```diff
@@ -0,0 +1,241 @@
...
+
+      testWidgetsOnAllPlatforms("does not convert with 7 or more #", (tester) async {
+        final context = await tester //
+            .createDocument()
+            .withSingleEmptyParagraph()
+            .withInputSource(TextInputSource.ime)
+            .autoFocus(true)
+            .pump();
+
+        // Type a header token that's longer than the smallest supported header
```
angelosilvestre, line 44: smallest -> largest

super_editor/test/super_editor/text_entry/paragraph_conversions_test.dart
```diff
@@ -0,0 +1,241 @@
...
+        // Ensure that we now have two nodes, and the first one is an HR.
+        final document = context.editContext.document;
+        expect(document.nodes.length, 2);
+
+        expect(document.nodes.first, isA<HorizontalRuleNode>());
+        expect(document.nodes.last, isA<ParagraphNode>());
+        expect((document.nodes.last as ParagraphNode).text.text.isEmpty, isTrue);
+      });
+
+      testWidgetsOnAllPlatforms('does not convert', (tester) async {
```
angelosilvestre, line 174: I think it's missing what does not convert

angelosilvestre (commented):
I left some comments.

PR #1164, opened 2023-06-09
[SuperEditor] - Hash (#) header conversions (Resolves 1159) (#1160)

matthew-carroll:
[SuperEditor] - Hash (#) header conversions (Resolves 1159) (#1160)


angelosilvestre (commented):
LGTM

Issue #1168, opened 2023-06-12
[super_editor][IOS]There is an error when using the Chinese nine-grid input method.


Dboy233:
super_editor: 0.2.6

Please add keyboard input tests for other countries during testing，For example, the Chinese Nine-square Input Method.

> The Chinese Nine-square Input Method is an input method for typing Chinese characters. It mainly relies on users pressing the corresponding numeric keys on a smartphone or computer keyboard to enter the pinyin letters. The input system will automatically match possible Chinese characters based on the user's input and display them as candidate options in the input box. Users can choose the corresponding Chinese characters by scrolling the cursor up and down or pressing the corresponding number key. When the user confirms the input, the input system will display the selected Chinese character in the input box. Therefore, as the user enters the pinyin letters and matches them with the Chinese characters, the text in the input box changes continuously until the user confirms the input is complete.***Input methods with similar input features also include Japanese and Korean.***

Please take a look at my input test.

https://github.com/superlistapp/super_editor/assets/37604230/d8129a13-b54f-42ba-b0da-b5fc140321fc

Process of reproducing the issue：Press the 'MNO' button twice

Error occurs when pressing 'MNO' for the second time. 'DocumentImeInputClient.updateEditingValueWithDeltas' will be executed twice.

This is the error log. ➡️ [error.log](https://github.com/superlistapp/super_editor/files/11721220/error.log)

| Keyboard operations.      | Step 1 | Step 2｜
| ----------- | ----------- | ----------- |
| <img width="874" alt="图片" src="https://github.com/superlistapp/super_editor/assets/37604230/25602d6c-8fa4-42bd-9f59-9d1df2f92a6b">  | <img width="1053" alt="图片" src="https://github.com/superlistapp/super_editor/assets/37604230/8b9184bf-ca74-456c-924b-06f039fda373">| <img width="1023" alt="图片" src="https://github.com/superlistapp/super_editor/assets/37604230/c9d8c525-5b9b-4a02-92b9-4bf2bb57ce3b">|

error:

<img width="1327" alt="图片" src="https://github.com/superlistapp/super_editor/assets/37604230/4ff177ad-9751-4aad-b09b-38d18c9cc400">

Dboy233:
#1225

PR #1170, opened 2023-06-13
FixBug:#1168,An exception occurred during the text replacement process.

Dboy233:
https://github.com/superlistapp/super_editor/issues/1168

In Chinese, Korean, and Japanese, sometimes the input of the software keyboard does not match the display on the screen, similar to character merging. 

In Chinese input, when 'ni' is entered on the software keyboard, it actually represents the character '你' (you), replaced two characters with one character.

The software keyboard performs replacement, and after the replacement, the range of the text also changes.

Therefore, in this PR, the data changed after text replacement is resynchronized by the DocumentImeSerializer.


Dboy233:
#1225

PR #1172, opened 2023-06-14
Clean up analysis issues stable

rutvik110:
Updates to pending clean up on analysis issues due to upgrade to Flutter 3.10 and Dart 3


matthew-carroll:
@rutvik110 ideally these PRs are a direct cherry-pick from `main` and that's all, so that we maintain as close a match to `main` as possible. Additionally, we don't squash these PRs, we rebase them, which means all 3 of your "Update" commits would appear in the global history.

Is there a reason that this wasn't just a cherry pick? Were there changes we made against Flutter `master` that don't apply to Flutter `stable`?

rutvik110:
hey Matt, initially I did the cherry pick from the `main`, but the pub update on it after introduced a few changes and the analytics showed up one linting issue in the project which was needed to be fixed.

matthew-carroll:
What was the lint issue that appeared on this branch, but didn't appear on `main`?

rutvik110:
There were some unused imports[ in here](https://github.com/superlistapp/super_editor/pull/1172/commits/6b1b75154e1c6937d3f4c97addf147bdd167d37b). I might've missed those on `main` somehow.

matthew-carroll:
Can you check `main`? If they were missed on `main` then we should do this cherry pick without making that change, and we should fix it on `main` in some future PR. Whenever possible, we want `stable` to be downstream from `main`. The merge direction is always from `main` to `stable`, never the reverse.

rutvik110:
Just checked out main, this was resolved on main a long time back. 
Seems like it was present on `stable` for a while.

matthew-carroll:
Ok. We can keep it then. Can you clean up the commit history, as mentioned above?

rutvik110:
Just did a clean up on the commit history and pushed the updates.

PR #1173, opened 2023-06-14
Update one floating cursor test for the M3 default transition

HansMuller:
Prepare for the change to the default value of ThemeData.useMaterial3 coming in https://github.com/flutter/flutter/issues/127064. Currently the 'shows grey caret when far from text' test depends on the UI being configured for Material2. Changed the test so that it checks the theme's primary color; the cursor's default color. This works for M2 and M3.

### DIFF
--- a/super_editor/test/super_editor/supereditor_floating_cursor_test.dart
+++ b/super_editor/test/super_editor/supereditor_floating_cursor_test.dart
@@ -128,6 +128,11 @@ void main() {
 
         // Ensure the caret has the default color.
         caret = tester.widget<BlinkingCaret>(_caretFinder());
+        // The default caret color is the theme's primary color. Currently
+        // this test will succeed for both Material2 and Material3.
+        // TODO(hansmuller) when the default for ThemeData.useMaterial3 is
+        // changed to true, change this test to check for the default
+        // M3 theme's primary color's value, Color(0xff6750a4).
         expect(caret.color, Theme.of(tester.firstElement(_caretFinder())).primaryColor);
       });
     });
--- a/super_editor/test/super_editor/supereditor_floating_cursor_test.dart
+++ b/super_editor/test/super_editor/supereditor_floating_cursor_test.dart
@@ -126,9 +126,9 @@ void main() {
         // Ensure the caret is displayed.
         expect(_caretFinder(), findsOneWidget);
 
-        // Ensure the caret is blue.
+        // Ensure the caret has the default color.
         caret = tester.widget<BlinkingCaret>(_caretFinder());
-        expect(caret.color, Colors.blue);
+        expect(caret.color, Theme.of(tester.firstElement(_caretFinder())).primaryColor);
       });
     });
   });


super_editor/test/super_editor/supereditor_floating_cursor_test.dart
```diff
@@ -126,9 +126,9 @@ void main() {
         // Ensure the caret is displayed.
         expect(_caretFinder(), findsOneWidget);
 
-        // Ensure the caret is blue.
+        // Ensure the caret has the default color.
         caret = tester.widget<BlinkingCaret>(_caretFinder());
-        expect(caret.color, Colors.blue);
+        expect(caret.color, Theme.of(tester.firstElement(_caretFinder())).primaryColor);
```
matthew-carroll, line 131: We try to assert exactly what we expect in tests, so that when our assumptions fail to align with new behavior, we catch the difference. Given this change to theming, what is the specific color, within this test, that we should expect to find?

HansMuller: The default primary color is different for Material2 and Material3 and the caret just ends up with the default primary color. When we make Material3 the default, this test will break. I can make the test conditional on ThemeData.useMaterial3 and check the exact color if you want.

matthew-carroll: What will the new default color be?

HansMuller: The default primary color is Color(0xff6750a4). That's _colorSchemeLightM3.primary in theme_data.dart.

matthew-carroll: Is there a `Colors` value associated with that, or only the hex value? I think we can just replace `Colors.blue` with that new primary color value. If there's a constant, that's preferable, but if not, then you can add the hex directly. If you add the hex, please add a comment on the line above that states we expect the caret color to the theme's primary color.

HansMuller: It's just a hex value.  

I need to change your test so that it passes now, and after we change the default for `ThemeData.useMaterial3` to true. Once the `useMaterial3` change lands, I'd be happy to remove the code that handles the (Colors.blue) Material2 case, and just check for the new default primary color.

matthew-carroll: So you're saying the reason for the color lookup is because that same access just so happens to work for Material 2 and Material 3, which keeps the test passing now, and after the Material 3 change?

If I'm understanding that correctly, then can you please add a comment in the test code that makes this clear, so that we know we intend to return to a hard-coded color, and we intend to return to the previous direct caret property access? I think it would also be good to phrase that in a TODO with your name on it, for future reference.

HansMuller: Will do

HansMuller: Done

PR #1179, opened 2023-06-16
[SuperEditor] Fix deletion of first character of a header with backspace (Resolves #1176)

### LINKED ISSUE #1176: [SuperEditor] Deleting the first character of a heading node using backspace, converts it to a paragraph
Steps to reproduce:
* Open the demo app
* Place the caret in front the first character of a heading node (doesn't matter the level)
* Press backspace

**Expected:** deletes the character 
**Actual:** deletes the character and converts the node to a paragraph

https://github.com/superlistapp/super_editor/assets/31278849/3f27c21e-58bb-4d55-8162-b856bd96b24c

angelosilvestre:
[SuperEditor] Fix deletion of first character of a header with backspace (Resolves #1176)

When the caret is after the first character of a header, pressing deleting is causing the header to be converted to a paragraph.

The cause is that, after deleting the character, we are handling the backspace KeyUpEvent in `backspaceToClearParagraphBlockType`. As the first character was already removed when `backspaceToClearParagraphBlockType` the selection sits at the beginning of the text and thus the header metadata is cleared.

This PR changes `backspaceToClearParagraphBlockType` to ignore events other than `RawKeyDownEvent`.

### DIFF
--- a/super_editor/lib/src/default_editor/paragraph.dart
+++ b/super_editor/lib/src/default_editor/paragraph.dart
@@ -529,6 +529,10 @@ ExecutionInstruction backspaceToClearParagraphBlockType({
   required SuperEditorContext editContext,
   required RawKeyEvent keyEvent,
 }) {
+  if (keyEvent is! RawKeyDownEvent) {
+    return ExecutionInstruction.continueExecution;
+  }
+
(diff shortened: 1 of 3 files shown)


matthew-carroll (approved):
LGTM

PR #1182, opened 2023-06-17
Added reference to old DocumentNode within NodeRemovedEvent (Resolves #1181)

### LINKED ISSUE #1181: [SuperEditor] - Deletion events need to know what was deleted
Currently deletion events don't include what they deleted. This creates problems for reactions that should only run when certain types or forms of content are deleted.

For example: an app has multiple documents. Any of the documents are allowed to refer to shared tasks. When a task node is deleted, the app needs to delete the corresponding task. But the corresponding reaction can't tell that a task node was removed. It only knows the ID of the node that was removed.

matthew-carroll:
Added reference to old DocumentNode within NodeRemovedEvent (Resolves #1181)

This change was made so that editor reactions can base their decisions on content in a removed node.

This addition isn't ideal because we're passing a reference to a now-deleted, mutable `DocumentNode`, but it's unclear what else we should do. We'll probably revisit issues like this later.

### DIFF
--- a/super_editor/example/analysis_options.yaml
+++ b/super_editor/example/analysis_options.yaml
@@ -1,5 +1,3 @@
-include: package:flutter_lints/flutter.yaml
-
 linter:
   rules:
     avoid_print: false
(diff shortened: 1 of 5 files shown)


angelosilvestre (approved):
LGTM

miguelcmedeiros:
@matthew-carroll 
It would also be helpful if `NodeChangedEvent` also references to the old document node as well, for the same reasons that `NodeRemovedEvent`. What do you think?

miguelcmedeiros (approved):

PR #1183, opened 2023-06-17
[SuperEditor] Linkify URLs without www (Resolves #1178)

### LINKED ISSUE #1178: [SuperEditor] Linkify a URL without "www."
Following how Apple Notes detect a URL and converts the text into a link, `SuperEditor` should be able to linkify "google.com" without using "www."

angelosilvestre:
[SuperEditor] Linkify URLs without www (Resolves #1178).

By default, the package used to extract URLs from the text doesn't recognize URLs without www.

This PR enables the option to allow "loose" URLs.


matthew-carroll (approved):
LGTM

PR #1118, opened 2023-05-09
Make Android IME work when newlines are in deltas list (Resolves #1081)

matthew-carroll:
Make Android IME work when newlines are in deltas list (Resolves #1081)

Our existing IME delta implementation assumes that the overall structure of a `Document` stays consistent through a given list of deltas. In some cases, such as an Android device inserting a newline through deltas (instead of a hardware ENTER button), the structure of the document changes in the middle of the deltas, e.g., a single paragraph becomes two paragraphs.

This PR reworks the IME delta application process so that our internal mapping from `TextEditingValue` ranges, to `Document` ranges, updates itself with each delta, if needed.

PR #1184, opened 2023-06-19
[SuperEditor][Mobile] Fix task and list item splitting on ENTER (Resolves #1157)

### LINKED ISSUE #1157: [IOS] The enter key of the soft keyboard is not triggered properly
In the example program, moving the cursor to the end of the TaskNode and pressing the enter key will not create a new TaskNode, nor will it execute the `enterToInsertNewTask` logic in the `defaultImeKeyboardActions` list.

The 'newline' logic was executed in the "DeltaTextInputClientDecorator" class.

On the Android side, it is exactly the opposite.

图片

图片

图片

图片

angelosilvestre:
[SuperEditor][Mobile] Fix task and list item splitting on ENTER. Resolves #1157 

On mobile, pressing ENTER on a task doesn't split the task. The cause is that we weren't handling the `TaskNode` in `insertBlockLevelNewline` and `_insertNewlineInDeltas`.

After the fix, I found the same issue reported in https://github.com/superlistapp/super_editor/issues/1081, so I applied the same solution.

During the tests, I noticed that the list item splitting wasn't working on Android. I also fixed that.

There are 3 ways that the ENTER key is reported:
- On Desktop we receive key presses.
- On Android we receive an insertion delta of `\n`.
- On iOS we receive a `newline` input action.

Because of that I added 3 tests for each case. Maybe we could add another `pressEnter` method that generates different signals for each platform.

### DIFF
--- a/super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
+++ b/super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
@@ -7,9 +7,11 @@ import 'package:super_editor/src/core/document_layout.dart';
 import 'package:super_editor/src/core/document_selection.dart';
 import 'package:super_editor/src/core/editor.dart';
 import 'package:super_editor/src/default_editor/common_editor_operations.dart';
+import 'package:super_editor/src/default_editor/list_items.dart';
 import 'package:super_editor/src/default_editor/multi_node_editing.dart';
 import 'package:super_editor/src/default_editor/paragraph.dart';
 import 'package:super_editor/src/default_editor/selection_upstream_downstream.dart';
+import 'package:super_editor/src/default_editor/tasks.dart';
 import 'package:super_editor/src/default_editor/text.dart';
 import 'package:super_editor/src/infrastructure/_logging.dart';
 
@@ -478,7 +480,40 @@ class TextDeltasDocumentEditor {
 
     final newNodeId = Editor.createNodeId();
 
-    if (extentNode is ParagraphNode) {
+    if (extentNode is ListItemNode) {
+      if (extentNode.text.text.isEmpty) {
+        // The list item is empty. Convert it to a paragraph.
+        editorOpsLog.finer(
+            "The current node is an empty list item. Converting it to a paragraph instead of inserting block-level newline.");
+        editor.execute([
+          ConvertTextNodeToParagraphRequest(nodeId: extentNode.id),
+        ]);
+        return;
+      }
+
+      // Split the list item into two.
+      editorOpsLog.finer("Splitting list item in two.");
+      editor.execute([
+        SplitListItemRequest(
+          nodeId: extentNode.id,
+          splitPosition: caretPosition.nodePosition as TextNodePosition,
+          newNodeId: newNodeId,
+        ),
+        ChangeSelectionRequest(
+          DocumentSelection.collapsed(
+            position: DocumentPosition(
+              nodeId: newNodeId,
+              nodePosition: const TextNodePosition(offset: 0),
+            ),
+          ),
+          SelectionChangeType.insertContent,
+          SelectionReason.userInteraction,
+        ),
+      ]);
+      final newListItemNode = document.getNodeById(newNodeId)!;
+
+      _updateImeRangeMappingAfterNodeSplit(originNode: extentNode, newNode: newListItemNode);
+    } else if (extentNode is ParagraphNode) {
       // Split the paragraph into two. This includes headers, blockquotes, and
       // any other block-level paragraph.
       final currentExtentPosition = caretPosition.nodePosition as TextNodePosition;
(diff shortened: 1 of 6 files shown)


super_editor/test/src/default_editor/list_items_test.dart
```diff
@@ -161,6 +161,221 @@ void main() {
         // Ensure the caret is being displayed at the correct position.
         expect(SuperEditorInspector.findCaretOffsetInDocument(), offsetMoreOrLessEquals(computedOffsetAfterUnindent));
       });
+
+      testWidgetsOnDesktop("inserts new item on ENTER at end of existing item", (tester) async {
```
matthew-carroll, line 165: This test should probably run on all platforms - what if a user plugs in a Bluetooth keyboard to an Android tablet?

super_editor/test/src/default_editor/list_items_test.dart
```diff
@@ -161,6 +161,221 @@ void main() {
...
+          DocumentSelection.collapsed(
+            position: DocumentPosition(
+              nodeId: document.nodes.last.id,
+              nodePosition: const TextNodePosition(offset: 0),
+            ),
+          ),
+        );
+      });
+
+      testWidgetsOnAndroid("inserts new item on ENTER at end of existing item with mobile keyboard (on Android)",
```
matthew-carroll, line 202: If, by some chance, we were to receive a "/n" from the Mac OS IME, would we not want this same behavior?

super_editor/test/src/default_editor/list_items_test.dart
```diff
@@ -161,6 +161,221 @@ void main() {
...
+          DocumentSelection.collapsed(
+            position: DocumentPosition(
+              nodeId: document.nodes.last.id,
+              nodePosition: const TextNodePosition(offset: 0),
+            ),
+          ),
+        );
+      });
+
+      testWidgetsOnIos("inserts new item on ENTER at end of existing item with mobile keyboard (on iOS)",
```
matthew-carroll, line 240: Seems like the difference between these tests is mostly about the newline signal, and not necessarily the platform. For example, if a new version of iOS suddenly switched from a newline action to a "/n", we'd still want our code to work, and we'd still want the test names to make sense, right?

super_editor/lib/src/default_editor/tasks.dart
```diff
@@ -403,20 +403,24 @@ class SplitExistingTaskRequest implements EditRequest {
   const SplitExistingTaskRequest({
     required this.nodeId,
     required this.splitOffset,
+    this.newNodeId,
   });
 
   final String nodeId;
```
matthew-carroll, line 409: For clarity, if we're going to have a `newNodeId`, then I think it makes sense to change `nodeId` -> `existingNodeId`.

super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
```diff
@@ -574,4 +609,35 @@ class TextDeltasDocumentEditor {
     }
     commonOps.insertBlockLevelNewline();
   }
+
+  /// Updates mappings from Document nodes to IME ranges and IME ranges to Document nodes.
+  void _updateImeRangeMappingAfterNodeSplit({
```
matthew-carroll, line 614: I haven't read through each line of this, but it's worrying to see an entirely new mapping method. What was our existing implementation missing?

angelosilvestre: This is the same change introduced in https://github.com/superlistapp/super_editor/pull/1118. I just extracted into a new method.

super_editor/test/src/default_editor/list_items_test.dart
angelosilvestre: Currently, we only have this behavior for Android. On iOS, for example, pressing ENTER generates both a new line insertion and a new line input action, so we ignore the `\n` insertion.

super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
matthew-carroll: I just looked at the change log for #1118 - where did you extract this method from? Why did you extract it?

angelosilvestre: I extracted from here: https://github.com/superlistapp/super_editor/blob/c70dd49ce08fdc28c1738e9aa89332196d21069a/super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart#L481-L534

We need the same code when we split list items and tasks.  Otherwise, we get the same "Couldn't map an IME position to a document position" exception. So I extracted this code to a method to be used when splitting paragraphs, list items and tasks.

matthew-carroll: If you take `if (extentNode is ParagraphNode) {` and change it to `if (extentNode is TextNode) {` does that fix everything?

angelosilvestre: It doesn't. We need to execute different requests depending on the node type.

matthew-carroll: Can you be more specific? What needs to be different between a `ParagraphNode` and a `TextNode`?

angelosilvestre: If we change it to `if (extentNode is TextNode) { `, then the if branch will be executed for both `ListItemNode` and `TaskNode`.  

We don't want to execute a `SplitParagraphRequest`  for these nodes because it would generate a new `ParagraphNode` instead of a `ListItemNode` / `TaskNode`.

matthew-carroll (approved):
LGTM

PR #1196, opened 2023-06-20
[SuperEditor] Fix text insertion when the caret sits at a horizontal rule (Resolves #1187)

### LINKED ISSUE #1187: [SuperEditor] Typing text when horizontal rule node has selection, introduces empty text nodes in document
Steps to reproduce:
* Open demo app
* Add a horizontal rule node
* Place selection at the start or end of the node
* Type text

**Expected:** Taking Google docs as a reference, a new paragraph node is added below or above (depending where the selection is) with the inserted character and the selection moves to the end of the newly inserted node
**Actual:** Empty text nodes are added above (if selection is at the start) or below (if selection is at the end)

https://github.com/superlistapp/super_editor/assets/31278849/ec93c429-21eb-47fc-bd35-2f4c2b6ad1e1

(not sure why KeyCastr did not capture all keys that I was typing, but for each character that I would type, a new empty text node would be added)

angelosilvestre:
[SuperEditor] Fix text insertion when the caret sits at a horizontal rule. Resolves #1187

Typing text when a horizontal rule node has selection, introduces empty text nodes in the document. The issue is that we are inserting a new paragraph but not updating the insertion position before trying to insert the text.

This PR changes `_insertPlainText` to consider the end position of the new node to insert the text.

https://github.com/superlistapp/super_editor/assets/7597082/df5f983a-5e18-427c-861d-c22713b9a7ee

### DIFF
--- a/super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
+++ b/super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
@@ -268,26 +268,34 @@ class TextDeltasDocumentEditor {
   ) {
     editorOpsLog.fine('Attempting to insert "$text" at position: $insertionPosition');
 
-    final extentNodePosition = insertionPosition.nodePosition;
+    DocumentPosition effectiveInsertionPosition = insertionPosition;
+    NodePosition extentNodePosition = insertionPosition.nodePosition;
+    String extentNodeId = insertionPosition.nodeId;
+
     if (extentNodePosition is UpstreamDownstreamNodePosition) {
       editorOpsLog.fine("The selected position is an UpstreamDownstreamPosition. Inserting new paragraph first.");
-      commonOps.insertBlockLevelNewline();
+      extentNodeId = Editor.createNodeId();
+      commonOps.insertBlockLevelNewline(nodeId: extentNodeId);
+
+      final node = document.getNodeById(extentNodeId)!;
+      extentNodePosition = node.endPosition;
+      effectiveInsertionPosition = DocumentPosition(nodeId: extentNodeId, nodePosition: extentNodePosition);
     }
 
-    final extentNode = document.getNodeById(insertionPosition.nodeId)!;
+    final extentNode = document.getNodeById(extentNodeId)!;
     if (extentNode is! TextNode || extentNodePosition is! TextNodePosition) {
       editorOpsLog.fine(
           "Couldn't insert text because Super Editor doesn't know how to handle a node of type: $extentNode, with position: $extentNodePosition");
       return false;
     }
 
-    final textNode = document.getNodeById(insertionPosition.nodeId) as TextNode;
+    final textNode = document.getNodeById(extentNodeId) as TextNode;
 
     editorOpsLog.fine("Executing text insertion command.");
     editorOpsLog.finer("Text before insertion: '${textNode.text.text}'");
     editor.execute([
       InsertTextRequest(
-        documentPosition: insertionPosition,
+        documentPosition: effectiveInsertionPosition,
         textToInsert: text,
         attributions: composerPreferences.currentAttributions,
       ),
--- a/super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
+++ b/super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
@@ -271,6 +271,7 @@ class TextDeltasDocumentEditor {
     DocumentNode? insertionNode = document.getNodeById(insertionPosition.nodeId);
     if (insertionNode == null) {
       editorOpsLog.warning('Attempted to insert text using a non-existing node');
+      return false;
     }
 
     if (insertionPosition.nodePosition is UpstreamDownstreamNodePosition) {
--- a/super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
+++ b/super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
@@ -268,41 +268,37 @@ class TextDeltasDocumentEditor {
   ) {
     editorOpsLog.fine('Attempting to insert "$text" at position: $insertionPosition');
 
-    DocumentPosition effectiveInsertionPosition = insertionPosition;
-    NodePosition extentNodePosition = insertionPosition.nodePosition;
-    String extentNodeId = insertionPosition.nodeId;
+    DocumentNode? insertionNode = document.getNodeById(insertionPosition.nodeId);
+    if (insertionNode == null) {
+      editorOpsLog.warning('Attempted to insert text using a non-existing node');
+    }
 
-    if (extentNodePosition is UpstreamDownstreamNodePosition) {
+    if (insertionPosition.nodePosition is UpstreamDownstreamNodePosition) {
       editorOpsLog.fine("The selected position is an UpstreamDownstreamPosition. Inserting new paragraph first.");
       commonOps.insertBlockLevelNewline();
 
-      // After inserting a block level new line, the selection might change to another node.
+      // After inserting a block level new line, the selection changes to another node.
       // Therefore, we need to update the insertion position.
-      extentNodeId = selection.value!.extent.nodeId;
-      final node = document.getNodeById(extentNodeId)!;
-      extentNodePosition = node.endPosition;
-      effectiveInsertionPosition = DocumentPosition(nodeId: extentNodeId, nodePosition: extentNodePosition);
+      insertionNode = document.getNodeById(selection.value!.extent.nodeId)!;
+      insertionPosition = DocumentPosition(nodeId: insertionNode.id, nodePosition: insertionNode.endPosition);
     }
 
-    final extentNode = document.getNodeById(extentNodeId)!;
-    if (extentNode is! TextNode || extentNodePosition is! TextNodePosition) {
+    if (insertionNode is! TextNode || insertionPosition.nodePosition is! TextNodePosition) {
       editorOpsLog.fine(
-          "Couldn't insert text because Super Editor doesn't know how to handle a node of type: $extentNode, with position: $extentNodePosition");
+          "Couldn't insert text because Super Editor doesn't know how to handle a node of type: $insertionNode, with position: ${insertionPosition.nodePosition}");
       return false;
     }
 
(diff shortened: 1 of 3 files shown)


matthew-carroll:
@angelosilvestre was this always broken, or did it break recently? If it broke recently, do you know why?

angelosilvestre:
@matthew-carroll I think it was always broken.

matthew-carroll:
@angelosilvestre can you go back a ways and check? I'd be surprised if this was missed.

angelosilvestre:
@matthew-carroll It was working before indeed.

After running git bisect it seems it was broken after 5c8de5de565b9377cbc2e5cba89837f36473766a

matthew-carroll:
@angelosilvestre can you see where/why it broke? This PR adds information to a high level public API that we weren't passing before. I'm not convinced that this PR gets to the root cause of the issue. Can you take a deeper a look?

angelosilvestre:
@matthew-carroll I took a deeper look. When this was working, the code was like this: 

```dart
// Here the selection sits at the horizontal rule.
final extentNodePosition = composer.selection!.extent.nodePosition;
if (extentNodePosition is UpstreamDownstreamNodePosition) {
    editorOpsLog.fine("The selected position is an UpstreamDownstreamPosition. Inserting new paragraph first.");
    insertBlockLevelNewline();
 }

// After inserting the block level new line, the selection sits at the newly created paragraph.
// So when we access composer.selection!.extent.nodeId we are accessing the paragraph node id.
final extentNode = editor.document.getNodeById(composer.selection!.extent.nodeId)!;
if (extentNode is! TextNode) {
   return false;
}
```

With the new code: 

```dart
// Here the selection sits at the horizontal rule.
final extentNodePosition = insertionPosition.nodePosition;
if (extentNodePosition is UpstreamDownstreamNodePosition) {
  editorOpsLog.fine("The selected position is an UpstreamDownstreamPosition. Inserting new paragraph first.");
  commonOps.insertBlockLevelNewline();
 }

// Even though we created a new paragraph node, we are still trying to insert the text in the horizontal rule, because we are using the same insertionPosition.
final extentNode = document.getNodeById(insertionPosition.nodeId)!;
if (extentNode is! TextNode || extentNodePosition is! TextNodePosition) {
   return false;
}
```

I was able to update the PR to avoid adding information to the public API.

super_editor/test/super_editor/components/horizontal_rule_test.dart
```diff
@@ -0,0 +1,108 @@
+import 'package:flutter_test/flutter_test.dart';
+import 'package:flutter_test_robots/flutter_test_robots.dart';
+import 'package:super_editor/src/test/super_editor_test/supereditor_inspector.dart';
+import 'package:super_editor/src/test/super_editor_test/supereditor_robot.dart';
+import 'package:super_editor/super_editor.dart';
+
+import '../../test_tools.dart';
+import '../document_test_tools.dart';
+
+void main() {
```
matthew-carroll, line 10: Do we not have existing tests that create horizontal rules?

super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
```diff
@@ -268,26 +268,36 @@ class TextDeltasDocumentEditor {
   ) {
     editorOpsLog.fine('Attempting to insert "$text" at position: $insertionPosition');
 
-    final extentNodePosition = insertionPosition.nodePosition;
+    DocumentPosition effectiveInsertionPosition = insertionPosition;
```
matthew-carroll, line 271: The word "effective" is confusing. If this "insertion position" is different from `insertionPosition`, please ensure that the name says why and/or how.

super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
```diff
@@ -268,26 +268,36 @@ class TextDeltasDocumentEditor {
...
+    NodePosition extentNodePosition = insertionPosition.nodePosition;
```
matthew-carroll, line 272: Is "extent" the right term here, and below? You're accessing the `insertionPosition`'s information, but then you're referring to it as the "extent". I understand that in practice the insertion position happens to be the selection extent, but is that what's important here? And will that position still be the "extent" after you insert the block level newline?

super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
```diff
@@ -268,26 +268,36 @@ class TextDeltasDocumentEditor {
...
+    String extentNodeId = insertionPosition.nodeId;
+
     if (extentNodePosition is UpstreamDownstreamNodePosition) {
       editorOpsLog.fine("The selected position is an UpstreamDownstreamPosition. Inserting new paragraph first.");
       commonOps.insertBlockLevelNewline();
+
+      // After inserting a block level new line, the selection might change to another node.
```
matthew-carroll, line 279: Aren't we, in fact, guaranteed that the selection has changed?

super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
```diff
@@ -268,26 +268,36 @@ class TextDeltasDocumentEditor {
...
-    final extentNode = document.getNodeById(insertionPosition.nodeId)!;
+    final extentNode = document.getNodeById(extentNodeId)!;
     if (extentNode is! TextNode || extentNodePosition is! TextNodePosition) {
       editorOpsLog.fine(
           "Couldn't insert text because Super Editor doesn't know how to handle a node of type: $extentNode, with position: $extentNodePosition");
       return false;
     }
 
-    final textNode = document.getNodeById(insertionPosition.nodeId) as TextNode;
+    final textNode = document.getNodeById(extentNodeId) as TextNode;
```
matthew-carroll, line 294: Can't you just say `final textNode = extentNode as TextNode`?

super_editor/test/super_editor/components/horizontal_rule_test.dart
```diff
@@ -0,0 +1,108 @@
...
+---
+
+Paragraph 2
+''')
+          .withInputSource(TextInputSource.ime)
+          .pump();
+
+      final document = testContext.editContext.document;
+
+      // Place caret at the beginning of the last node.
```
matthew-carroll, line 27: Let's put these two actions together, with a comment about the overall goal. Otherwise, it's still confusing to the reader why we're doing these things:

```dart
// Place the caret at the end of the horizontal rule, by first placing the caret in the paragraph after the
// horizontal rule, and then pressing the left arrow to move it up.
await tester.placeCaretInParagraph(...);
await tester.pressLeftArrow();
```

super_editor/test/super_editor/components/horizontal_rule_test.dart
```diff
@@ -0,0 +1,108 @@
...
+      await tester.placeCaretInParagraph(document.nodes.last.id, 0);
+
+      // Press left arrow to move the selection to the end of the horizontal rule.
+      await tester.pressLeftArrow();
+
+      // Type at the end of the horizontal rule
+      await tester.typeImeText('new paragraph');
+
+      // Ensure a new node was created.
```
matthew-carroll, line 36: All of these expectations can be placed in a single block, with a single overarching comment.

```dart
// Ensure that the new text was inserted in a new paragraph after the horizontal rule.
expect(...);
final insertedNode...
expect(insertedNode, ...);
...
```

super_editor/test/super_editor/components/horizontal_rule_test.dart
angelosilvestre: We have some places were we convert from `---` or create manually with `context.editContext.commonOps.insertHorizontalRule();`

We have a `task_test.dart`, so I thought it would be a good idea to create this new file.

super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
angelosilvestre: In fact, we can just use the `extentNode` because we already checked that `extentNode` is a `TextNode`.

super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
angelosilvestre: Updated.

super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
angelosilvestre: Updated.

super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
```diff
@@ -268,31 +268,37 @@ class TextDeltasDocumentEditor {
   ) {
     editorOpsLog.fine('Attempting to insert "$text" at position: $insertionPosition');
 
-    final extentNodePosition = insertionPosition.nodePosition;
-    if (extentNodePosition is UpstreamDownstreamNodePosition) {
+    DocumentNode? insertionNode = document.getNodeById(insertionPosition.nodeId);
+    if (insertionNode == null) {
+      editorOpsLog.warning('Attempted to insert text using a non-existing node');
```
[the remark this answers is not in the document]
angelosilvestre: Probably not. Updated.

super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
```diff
@@ -268,31 +268,37 @@ class TextDeltasDocumentEditor {
...
+    }
+
+    if (insertionPosition.nodePosition is UpstreamDownstreamNodePosition) {
       editorOpsLog.fine("The selected position is an UpstreamDownstreamPosition. Inserting new paragraph first.");
       commonOps.insertBlockLevelNewline();
+
+      // After inserting a block level new line, the selection changes to another node.
+      // Therefore, we need to update the insertion position.
+      insertionNode = document.getNodeById(selection.value!.extent.nodeId)!;
```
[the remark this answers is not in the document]
angelosilvestre: They are the node and the position where the text will be inserted. Maybe it would be clearer if we call them `textInsertionPosition` and `nodeAtTextInsertionPosition`...

matthew-carroll (approved):
LGTM

PR #1199, opened 2023-06-21
Let the user add attributions without auto-merging with nearby attributions (Resolves #1198)

### LINKED ISSUE #1198: [attributed_text] - Make it possible to add attributions without expanding existing attributions
Currently, whenever an attribution is added to an `AttributedText`, if the new attribution is immediately preceded or followed by another attribution, whose "can merge with" query returns true, those attributions are combined into one.

This is useful for styles like bold, italics, etc, which have no reason to subdivide themselves when characters appear in sequence.

However, there are other attributions, such as hash tags, which should sometimes merge, and sometimes shouldn't. For example:

```
// Should merge
#
#j
#jo
#joh
#john

// Shouldn't merge
#john
#john#

// Should merge again
#john#
#john#s
```

Make auto-merging optional when adding attributions to an `AttributedText`.

matthew-carroll:
Let the user add attributions without auto-merging with nearby attributions (Resolves #1198)

This change supports use-cases like hash tags where sometimes we want the attribution to expand, and other times we don't.

Example:
```
// We want to expand
#jo
#joh

// We don't want to expand
#john
#john#
```

### DIFF
--- a/attributed_text/CHANGELOG.md
+++ b/attributed_text/CHANGELOG.md
@@ -1,3 +1,6 @@
+## [X.Y.Z] - ???
+ * `AttributedText` now allows you to `addAttribution()` without auto-merging with preceding and following attributions (#1198)
+
 ## [0.2.2] - May, 2023
 Upgrade Dart constraints to explicitly include Dart 3. Make `markers` public on `AttributedSpans`.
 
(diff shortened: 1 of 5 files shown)


angelosilvestre (commented):
LGTM

Issue #1201, opened 2023-06-23
[attributed_text] AttributedSpans.hasAttributionsWithin incorrectly throws as it incorrectly iterates over a Set of attributes


quaaantumdev:
The method AttributedSpans.hasAttributionsWithin iterates over a Set of attributes it want's to check and removes removes entries from this Set while iterating. This throws an exception (ConcurrentModificationError) as the iterator cannot be used after removing entries from the underlying it. This is a simple algorithmic bug / invalid usage of a set and it's iterator.

Here is the fix for this issue: https://github.com/superlistapp/super_editor/pull/1202
(previous pull request without a unit test was: https://github.com/superlistapp/super_editor/pull/1188)

PR #1202, opened 2023-06-23
Fix ConcurrentModificationError in AttributedSpans.hasAttributionsWithin (with Unit-Test)

quaaantumdev:
(second pull request for this issue, this time with tests, closing the previous one https://github.com/superlistapp/super_editor/pull/1188)

Fixes Issue: https://github.com/superlistapp/super_editor/issues/1201

The method AttributedSpans.hasAttributionsWithin iterates over a Set of attributes it want's to check and removes removes entries from this Set while iterating. This throws an exception (ConcurrentModificationError) as the iterator cannot be used after removing entries from the underlying it. This is a simple algorithmic bug / invalid usage of a set and it's iterator which will be fixed by this pull request

### DIFF
--- a/attributed_text/lib/src/attributed_spans.dart
+++ b/attributed_text/lib/src/attributed_spans.dart
@@ -65,11 +65,14 @@ class AttributedSpans {
   }) {
     final attributionsToFind = Set.from(attributions);
     for (int i = start; i <= end; ++i) {
+      final foundOnCharacter = <Attribution>{};
       for (final attribution in attributionsToFind) {
         if (hasAttributionAt(i, attribution: attribution)) {
-          attributionsToFind.remove(attribution);
+          foundOnCharacter.add(attribution);
         }
-
+      }
+      if (foundOnCharacter.isNotEmpty) {
+        attributionsToFind.removeAll(foundOnCharacter);
         if (attributionsToFind.isEmpty) {
           return true;
         }
--- a/attributed_text/lib/src/attributed_spans.dart
+++ b/attributed_text/lib/src/attributed_spans.dart
@@ -65,17 +65,15 @@ class AttributedSpans {
   }) {
     final attributionsToFind = Set.from(attributions);
     for (int i = start; i <= end; ++i) {
-      final foundOnCharacter = <Attribution>{};
+      final foundAttributions = <Attribution>{};
       for (final attribution in attributionsToFind) {
         if (hasAttributionAt(i, attribution: attribution)) {
-          foundOnCharacter.add(attribution);
+          foundAttributions.add(attribution);
         }
       }
-      if (foundOnCharacter.isNotEmpty) {
-        attributionsToFind.removeAll(foundOnCharacter);
-        if (attributionsToFind.isEmpty) {
-          return true;
-        }
+      attributionsToFind.removeAll(foundAttributions);
+      if (attributionsToFind.isEmpty) {
+        return true;
       }
     }
     return false;
--- a/attributed_text/test/attributed_spans_test.dart
+++ b/attributed_text/test/attributed_spans_test.dart
@@ -537,6 +537,40 @@ void main() {
         ]).expectSpans(spans);
       });
 
+      test('looking for multiple', () {
+        final spans = AttributedSpans()
+          ..addAttribution(newAttribution: ExpectedSpans.bold, start: 0, end: 8)
+          ..addAttribution(newAttribution: ExpectedSpans.italics, start: 1, end: 5)
+          ..addAttribution(newAttribution: ExpectedSpans.strikethrough, start: 5, end: 9);
+
+        ExpectedSpans([
+          'bbbbbbbbb_',
+          '_iiiii____',
+          '_____sssss',
+        ]).expectSpans(spans);
+
+        expect(spans.hasAttributionsWithin(attributions: {
+          ExpectedSpans.bold,
+          ExpectedSpans.italics,
+          ExpectedSpans.strikethrough,
+        }, start: 0, end: 9), true);
+
+        expect(spans.hasAttributionsWithin(attributions: {
+          ExpectedSpans.bold,
+          ExpectedSpans.italics,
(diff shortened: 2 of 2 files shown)


matthew-carroll:
@angelosilvestre can you do a first-review pass?

attributed_text/lib/src/attributed_spans.dart
```diff
@@ -65,11 +65,14 @@ class AttributedSpans {
   }) {
     final attributionsToFind = Set.from(attributions);
     for (int i = start; i <= end; ++i) {
+      final foundOnCharacter = <Attribution>{};
```
angelosilvestre, line 68: IMO `foundAttributions` would be clearer than `foundOnCharacter`.

attributed_text/test/attributed_spans_test.dart
```diff
@@ -537,6 +537,40 @@ void main() {
...
+          'bbbbbbbbb_',
+          '_iiiii____',
+          '_____sssss',
+        ]).expectSpans(spans);
+
+        expect(spans.hasAttributionsWithin(attributions: {
+          ExpectedSpans.bold,
+          ExpectedSpans.italics,
+          ExpectedSpans.strikethrough,
+        }, start: 0, end: 9), true);
```
angelosilvestre, line 556: Please add commas at the end to reformat.

attributed_text/test/attributed_spans_test.dart
```diff
@@ -537,6 +537,40 @@ void main() {
         ]).expectSpans(spans);
       });
 
+      test('looking for multiple', () {
```
angelosilvestre, line 540: I think we could find a better name for this test.

angelosilvestre (commented):
I left some comments.

matthew-carroll:
@quaaantumdev 

> If you want to change these things, please let me know when all your comments are done

Please make changes when requested. The review process may move through multiple stages.

> or even better, push a fixing commit

Our bandwidth is consumed elsewhere. If this were the current priority, the work would already be done. We appreciate your contribution, but we need you to deliver the end result if you'd like for it to be merged.

> We already found a few more minor issues, not relevant to us, but they may haunt other people.

Have you filed tickets for these issues?

> I think making such contributions/fixes little effort may enhance the overall code quality more than discussing the perfect variable name for a local variable name used on three lines of code

This is a false choice. I haven't seen any onerous requests of you in this PR, it shouldn't be tough to comply. As for code quality, the only way to avoid a complete disaster is to always uphold as high a bar as we can. Once we start making exceptions, the project rots.

@angelosilvestre - Please provide specific direction in PR reviews whenever possible. Assume that any time you leave options open to the implementer, the implementer will choose the option you don't want them to. So provide context as to why something isn't ideal, and then direct the efforts towards what you want.

attributed_text/lib/src/attributed_spans.dart
matthew-carroll, line 68: Please call this `foundAttributions`

attributed_text/lib/src/attributed_spans.dart
```diff
@@ -65,11 +65,14 @@ class AttributedSpans {
...
       for (final attribution in attributionsToFind) {
         if (hasAttributionAt(i, attribution: attribution)) {
-          attributionsToFind.remove(attribution);
+          foundOnCharacter.add(attribution);
         }
-
+      }
+      if (foundOnCharacter.isNotEmpty) {
```
matthew-carroll, line 74: No need to check if the accumulated set is empty. Let's remove this if-statement.

attributed_text/test/attributed_spans_test.dart
matthew-carroll, line 540: As Angelo mentioned, this test name doesn't convey any particular goal or verification. What part of this test would fail without the modifications to `attributed_spans`?

attributed_text/test/attributed_spans_test.dart
matthew-carroll: I think all we're looking for here is a "," after `end` and `true`. Generally speaking, if everything easily fits on a line then we usually don't break it up with trailing ",". However, in most cases where we need to split across lines, we use trailing ",". You can see this formatting with `addAttribution` in the test right below this one. The `throwsA` line doesn't quite follow this pattern, but the `addAttribution` does.

The project format is standard `dartfmt` but with line length set to 120.

attributed_text/test/attributed_spans_test.dart
matthew-carroll: I took a deeper look at all of our tests. I think this test is probably in the wrong group.

All the tests above this one are testing that adding multiple attributions will result in the expected marker spans. But this test appears to be testing `hasAttributionsWithin()`.

We have a group of tests at the top called `attribution queries`. That seems like a much more appropriate place for this test.

You can add this test to the end of that group and call it `hasAttributionsWithin() can look for multiple attributions at the same time'

You mentioned elsewhere that you're hoping to contribute more in the future. To that end, it's important to keep in mind that clear communication is critical. If this test ever fails, for example, it will probably be 1+ years down the road, and the developer who has to deal with that failure won't be you. When a test fails, a developer must understand whether the source code is doing something it shouldn't, or whether the test should be altered for new requirements. This first requires understanding the intention of the test. All of that requires clear communication in test descriptions, test implementations, source code implementations, and source code documentation. 

That said, clear communication doesn't mean "write more words". We want to use the fewest words possible, while still communicating the critical information. Getting PRs merged quick within Flutter Bounty Hunter projects is tied heavily to learning how to document the right details, effectively. It's an art, and it takes time and focus.

matthew-carroll:
It looks like this PR is failing in CI. Perhaps it needs to rebase against `main`?

matthew-carroll:
@angelosilvestre can you take over this PR so we can clear it from the backlog?

angelosilvestre:
Closing as this was resolved in https://github.com/superlistapp/super_editor/pull/1401

PR #1206, opened 2023-06-26
chore: Relax http dependency constraint by including v1.x.x

llfbandit:
`http` package has finally reached stable version.
There is no compatiblity issue between version `0.13.1` and `1.0.0`.


Jethro87:
+1 on this. Packages are starting to depend on http ^1.0.0 and it's resulting in version incompatibilities.

cc @matthew-carroll

matthew-carroll:
@angelosilvestre can you cherry pick this to stable?

PR #1228, opened 2023-07-10
[SuperTextField] Fix ImeAttributedTextEditingController not removing inner controller listener (Resolves #1227)

### LINKED ISSUE #1227: [SuperTextField] ImeAttributedTextEditingController does not remove listener from the inner controller
`ImeAttributedTextEditingController` (let's call it the IME controller) takes a real controller (a.k.a. inner controller) and it attaches a listener to when it changes in the constructor.

In case the IME controller is disposed, but the real controller isn't, and then the real controller is passed to a new instance of the IME controller, then we get in a situation where the listener previously added to the real controller will try to notify the listeners of the disposed instance of the IME controller, which leads to the following exception:
```
flutter: exception: A _TextEditingController was used after being disposed.
Once you have called dispose() on a _TextEditingController, it can no longer be used.
stackTrace: #0      ChangeNotifier.debugAssertNotDisposed.<anonymous closure> (package:flutter/src/foundation/change_notifier.dart:179:9)
#1      ChangeNotifier.debugAssertNotDisposed (package:flutter/src/foundation/change_notifier.dart:186:6)
#2      ChangeNotifier.notifyListeners (package:flutter/src/foundation/change_notifier.dart:382:27)
#3      ImeAttributedTextEditingController._onInnerControllerChange (package:super_editor/src/infrastructure/super_textfield/input_method_engine/_ime_text_editing_controller.dart:199:5)
#4      ChangeNotifier.notifyListeners (package:flutter/src/foundation/change_notifier.dart:403:24)
#5      AttributedTextEditingController.update (package:super_editor/src/infrastructure/super_textfield/infrastructure/attributed_text_editing_controller.dart:655:5)
#6      AttributedTextEditingController.insertAtCaret (package:super_editor/src/infrastructure/super_textfield/infrastructure/attributed_text_editing_controller.dart:219:5)
#7      ImeAttributedTextEditingController.insertAtCaret (package:super_editor/src/infrastructure/super_textfield/input_method_engine/_ime_text_editing_controller.dart:465:21)
#8      ImeAttributedTextEditingController.updateEditingValueWithDeltas (package:super_editor/src/infrastructure/super_textfield/input_method_engine/_ime_text_editing_controller.dart:293:11)
#9      TextInput._handleTextInputInvocation (package:flutter/src/services/text_input.dart:1857:63)
#10     TextInput._loudlyHandleTextInputInvocation (package:flutter/src/services/text_input.dart:1746:20)
#11     MethodChannel._handleAsMethodCall (package:flutter/src/services/platform_channel.dart:547:55)
#12     MethodChannel.setMethodCallHandler.<anonymous closure> (package:flutter/src/services/platform_channel.dart:540:34)
#13     _DefaultBinaryMessenger.setMessageHandler.<anonymous closure> (package:flutter/src/services/binding.dart:479:35)
...
```

angelosilvestre:
[SuperTextField] Fix ImeAttributedTextEditingController not removing inner controller listener. Resolves #1227 

In `ImeAttributedTextEditingController` we are adding a listener to the inner controller, but we aren't removing it upon disposal. This causes a crash if the inner controller is reused in a new `ImeAttributedTextEditingController`.

This PR changes `ImeAttributedTextEditingController` to remove the listener when `dispose` is called.

I'm not sure if we need a test for this, but I added anyway.

### DIFF
--- a/super_editor/test/super_textfield/ime_attributed_text_editing_controller_test.dart
+++ b/super_editor/test/super_textfield/ime_attributed_text_editing_controller_test.dart
@@ -401,6 +401,63 @@ void main() {
       expect(controller.selection, equals(const TextSelection.collapsed(offset: 8)));
     });
   });
+
+  testWidgetsOnAllPlatforms('allows reusing inner controller', (tester) async {
+    final innerController = AttributedTextEditingController(
+      text: AttributedText(
+        text: 'some text',
+      ),
+    );
+
+    final controllerA = ImeAttributedTextEditingController(
+      controller: innerController,
+      disposeClientController: false,
+    );
+
+    late ImeAttributedTextEditingController controllerB;
+
+    final rebuildNotifier = ValueNotifier<bool>(true);
+
+    await tester.pumpWidget(
+      MaterialApp(
+        home: Scaffold(
+          body: SizedBox(
+            width: 300,
+            child: ValueListenableBuilder(
+              valueListenable: rebuildNotifier,
+              builder: (context, shouldUseTextFieldA, _) {
+                return shouldUseTextFieldA //
+                    ? SuperTextField(
+                        inputSource: TextInputSource.ime,
+                        textController: controllerA,
+                      )
+                    : SuperTextField(
+                        inputSource: TextInputSource.ime,
+                        textController: controllerB,
+                      );
+              },
+            ),
+          ),
+        ),
+      ),
+    );
+
+    // Rebuild the widget tree, attaching the inner controller to another IME controller
+    // and disposing the old one.
+    controllerB = ImeAttributedTextEditingController(
+      controller: innerController,
+      disposeClientController: false,
+    );
+    controllerA.dispose();
+    rebuildNotifier.value = false;
+    await tester.pump();
+
+    // Change the text of the inner controller to notify the listeners.
+    innerController.text = AttributedText(text: 'New text');
+    await tester.pump();
+
+    // Reaching this point means that disposing the old controller didn't cause a crash.
+  });
 }
 
 Future<void> _pumpSuperTextField(
--- a/super_editor/test/super_textfield/ime_attributed_text_editing_controller_test.dart
+++ b/super_editor/test/super_textfield/ime_attributed_text_editing_controller_test.dart
@@ -402,7 +402,7 @@ void main() {
     });
   });
 
-  testWidgetsOnAllPlatforms('allows reusing inner controller', (tester) async {
+  testWidgetsOnAllPlatforms('allows reusing inner controller after disposal', (tester) async {
     final innerController = AttributedTextEditingController(
       text: AttributedText(
         text: 'some text',
@@ -416,7 +416,7 @@ void main() {
 
     late ImeAttributedTextEditingController controllerB;
 
-    final rebuildNotifier = ValueNotifier<bool>(true);
+    final useControllerA = ValueNotifier<bool>(true);
 
     await tester.pumpWidget(
       MaterialApp(
@@ -424,7 +424,7 @@ void main() {
           body: SizedBox(
             width: 300,
             child: ValueListenableBuilder(
-              valueListenable: rebuildNotifier,
+              valueListenable: useControllerA,
               builder: (context, shouldUseTextFieldA, _) {
                 return shouldUseTextFieldA //
                     ? SuperTextField(
@@ -449,12 +449,10 @@ void main() {
       disposeClientController: false,
     );
     controllerA.dispose();
-    rebuildNotifier.value = false;
-    await tester.pump();
+    useControllerA.value = false;
 
     // Change the text of the inner controller to notify the listeners.
     innerController.text = AttributedText(text: 'New text');
-    await tester.pump();
 
     // Reaching this point means that disposing the old controller didn't cause a crash.
   });
--- a/super_editor/test/super_textfield/ime_attributed_text_editing_controller_test.dart
+++ b/super_editor/test/super_textfield/ime_attributed_text_editing_controller_test.dart
@@ -402,57 +402,34 @@ void main() {
     });
   });
 
-  testWidgetsOnAllPlatforms('allows reusing inner controller after disposal', (tester) async {
+  testWidgets('isn\'t notified by inner controller after disposal', (tester) async {
     final innerController = AttributedTextEditingController(
       text: AttributedText(
         text: 'some text',
       ),
     );
 
-    final controllerA = ImeAttributedTextEditingController(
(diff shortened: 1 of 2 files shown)


super_editor/test/super_textfield/ime_attributed_text_editing_controller_test.dart
```diff
@@ -401,6 +401,63 @@ void main() {
...
+              },
+            ),
+          ),
+        ),
+      ),
+    );
+
+    // Rebuild the widget tree, attaching the inner controller to another IME controller
+    // and disposing the old one.
+    controllerB = ImeAttributedTextEditingController(
```
matthew-carroll, line 447: What's the purpose of creating another `controllerB` instance outside of the widget tree?

super_editor/test/super_textfield/ime_attributed_text_editing_controller_test.dart
```diff
@@ -401,6 +401,63 @@ void main() {
...
+      controller: innerController,
+      disposeClientController: false,
+    );
+    controllerA.dispose();
+    rebuildNotifier.value = false;
```
matthew-carroll, line 452: Why are we tracking rebuilds?

super_editor/test/super_textfield/ime_attributed_text_editing_controller_test.dart
```diff
@@ -401,6 +401,63 @@ void main() {
...
+    await tester.pump();
```
matthew-carroll, line 453: What's this pump supposed to do?

super_editor/test/super_textfield/ime_attributed_text_editing_controller_test.dart
```diff
@@ -401,6 +401,63 @@ void main() {
...
+
+    // Change the text of the inner controller to notify the listeners.
```
matthew-carroll, line 455: Reading through this test it feels like you're trying to write a generic test for a very specific issue. The test title is "allows reusing inner controller" which is very generic. But your test is specifically looking for a crash when notifying listeners after disposal. There's no need to be overly generic when there's something specific that you're verifying. In fact, being overly generic makes it very difficult for future developers to understand why you did what you did in this test.

super_editor/test/super_textfield/ime_attributed_text_editing_controller_test.dart
angelosilvestre: This is used to simulate the situation mentioned in the ticket, where the inner controller is attached to a new IME controller instance.

super_editor/test/super_textfield/ime_attributed_text_editing_controller_test.dart
angelosilvestre: This is just to force a rebuild and tell the builder we should use `controllerB` instead of `controllerA`. 

I renamed the variable.

super_editor/test/super_textfield/ime_attributed_text_editing_controller_test.dart
angelosilvestre: The pump doesn't seem to be needed. Removed.

super_editor/test/super_textfield/ime_attributed_text_editing_controller_test.dart
angelosilvestre: Is "allows reusing inner controller after disposal" specif enough?

We could also use a regular dart test instead of a widget test, as the issue doesn't depend on a widget tree.

matthew-carroll: That's still a generalization of the problem, and a generalization of what you're verifying with this test. You wrote this test to make sure that a listener isn't notified after disposal. Do we have any reason to believe that this test would ever verify anything other than that condition? The name implies that there might be any number of things that could go wrong when "re-using an inner controller" and this test makes sure that none of those bad things happen. But there's only one thing we know that this test will catch.

super_editor/test/super_textfield/ime_attributed_text_editing_controller_test.dart
```diff
@@ -401,6 +401,61 @@ void main() {
...
+      controller: innerController,
+      disposeClientController: false,
+    );
+    controllerA.dispose();
+    useControllerA.value = false;
+
+    // Change the text of the inner controller to notify the listeners.
+    innerController.text = AttributedText(text: 'New text');
+
+    // Reaching this point means that disposing the old controller didn't cause a crash.
```
matthew-carroll, line 457: This statement will be a surprise for the reader, because you haven't mentioned at any point what might crash. If this test ever becomes out of date and needs adjustment, what problem is the developer supposed to adjust for? We've lost the root problem that this test is meant to prevent.

super_editor/test/super_textfield/ime_attributed_text_editing_controller_test.dart
```diff
@@ -401,6 +401,61 @@ void main() {
...
+
+    await tester.pumpWidget(
+      MaterialApp(
+        home: Scaffold(
+          body: SizedBox(
+            width: 300,
+            child: ValueListenableBuilder(
+              valueListenable: useControllerA,
+              builder: (context, shouldUseTextFieldA, _) {
+                return shouldUseTextFieldA //
```
matthew-carroll, line 429: I think we can simplify this build in a couple ways. First, I don't think we need two controller variables. You can just re-assign a single `imeController` variable at the appropriate time.

Second, there's then no need for a ternary. Maybe you can use a `ValueNotifier<int>` to cause rebuilds:

```dart
ValueListenableBuilder(
  valueListenable: buildSignal,
  builder: (context, _, __) {
    return SuperTextField(
      textController: imeController,
    );
  },
);
```

super_editor/test/super_textfield/ime_attributed_text_editing_controller_test.dart
```diff
@@ -401,6 +401,61 @@ void main() {
...
+                      );
+              },
+            ),
+          ),
+        ),
+      ),
+    );
+
+    // Rebuild the widget tree, attaching the inner controller to another IME controller
+    // and disposing the old one.
```
matthew-carroll, line 446: I'm wondering if this requires a widget tree build at all? The problem is a listener firing when it shouldn't, right? Can't we test that with a lot less code without the widget tree?

angelosilvestre:
@matthew-carroll Updated the test to avoid pumping a widget tree. We still need to use a widget test because, otherwise, the exception isn't thrown.

matthew-carroll (approved):
LGTM

PR #1239, opened 2023-07-19
Re-use ContentLayers overlay and underlay Elements instead of always re-inflating (Resolves #1236)

### LINKED ISSUE #1236: Performance regressions
I was recently profiling energy usage and have noticed that there are some significant performance regressions in super_editor.

A while ago https://github.com/superlistapp/super_editor/pull/285 got merged to fix blinking cursor performance, but later it was replaced by [BlinkController](https://github.com/superlistapp/super_editor/blob/main/super_text_layout/lib/src/infrastructure/blink_controller.dart), which uses a `Ticker`. Unlike custom timer `Ticker` on desktop runs at constant 60 fps (each tick requesting new frame), causing unnecessary repaints given that cursor blinking interval is much longer.

This is further exacerbated by the fact that super editor forces repaint even if cursor is not visible. This is caused by [DocumentCaretOverlay](https://github.com/superlistapp/super_editor/blob/c6e2cfe5a0f0f0a097ab30707d6e337502cbc8b8/super_editor/lib/src/default_editor/document_caret_overlay.dart#L37) calling `startBlinking` on `BlinkController` unconditionally in `initState`. Regardless of whether there is any selection or whether the editor is focused).

Initially I assumed that removing `startBlinking` from `initState` would fix this, but that's not the case because `DocumentCaretOverlay` state gets torn down and re-inflated every time content layer gets rebuilt (i.e. on selection change). Which seems like a strange behavior caused by [ContentLayers](https://github.com/superlistapp/super_editor/blob/main/super_editor/lib/src/infrastructure/content_layers.dart). Every time layers are rebuilt, the child elements are [inflated](https://github.com/superlistapp/super_editor/blob/c6e2cfe5a0f0f0a097ab30707d6e337502cbc8b8/super_editor/lib/src/infrastructure/content_layers.dart#L146), which throws away the previous element state recreating the element tree from scratch. This seems at odds with how rest of Flutter works, where instead of always calling `inflate`, the element would normally call [updateChild](https://github.com/flutter/flutter/blob/b2e22d3558e8334d4427ed9b78d21a347375fad2/packages/flutter/lib/src/widgets/framework.dart#L3649) which allows preserving state of elements in tree if widgets are compatible.

matthew-carroll:
Re-use ContentLayers overlay and underlay Elements instead of always re-inflating (Resolves #1236)

### DIFF
--- a/super_editor/lib/src/infrastructure/content_layers.dart
+++ b/super_editor/lib/src/infrastructure/content_layers.dart
@@ -137,13 +137,8 @@ class ContentLayersElement extends RenderObjectElement {
 
   void buildLayers() {
     contentLayersLog.finer("ContentLayersElement - (re)building layers");
-    // FIXME: To get the layers to rebuild, we have to deactivate the existing layer Element and re-inflate
-    //        the layer's widget. This probably creates a lot of extra work for layers that don't
-    //        need to be rebuilt. Create a way for layers to opt-in to this behavior.
 
     owner!.buildScope(this, () {
-      // _deactivateLayers();
-
       final List<Element> underlays = List<Element>.filled(widget.underlays.length, _NullElement.instance);
       for (int i = 0; i < underlays.length; i += 1) {
         late final Element child;
--- a/super_editor/lib/src/infrastructure/content_layers.dart
+++ b/super_editor/lib/src/infrastructure/content_layers.dart
@@ -31,7 +31,7 @@ class ContentLayers extends RenderObjectWidget {
   ///
   /// These layers are placed at the same (x,y) as [content], and they're forced to layout
   /// at the exact same size as [content].
-  final List<Widget> underlays;
+  final List<WidgetBuilder> underlays;
 
   /// The primary content displayed in this widget, which determines the size and location
   /// of all [underlays] and [overlays].
@@ -41,7 +41,7 @@ class ContentLayers extends RenderObjectWidget {
   ///
   /// These layers are placed at the same (x,y) as [content], and they're forced to layout
   /// at the exact same size as [content].
-  final List<Widget> overlays;
+  final List<WidgetBuilder> overlays;
 
   @override
   RenderObjectElement createElement() {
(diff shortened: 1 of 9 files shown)


knopp:
This seems like a good way forward. I haven't looked at the failed tests yet, but I was able to remove `..startBlinking()` from `_CaretDocumentOverlayState.initState()` (which cases repaints even when not actively editing)/

knopp:
So the problem here seems to be that you're calling `child.rebuild(force: true);` during `buildLayers`, but the child is a `SingleChildRenderObjectElement` which already has a child and rebuild doesn't do anything useful. It definitely doesn't traverse down to children - that would normally be a performance issue given that Flutter already has finer grained dirty tracking.

Given that it seems that you want to avoid regular dirty element tracking here and just force rebuild of entire subtree, you probably needs bigger hammer than `rebuild`. Something like `reassemble` should do the trick, with the caveat that this is not what it is intended for (and is marked as protected - though calling it from ContentLayersElement the analyzer won't complain).

Changing `child.rebuild(force: true);` into `child.reassemble();` should make the tests pass.

matthew-carroll:
Do you have any recommendation for alternative approaches? I'm not married to anything in particular. I just want to make sure that the overlay and underlay layers are given an opportunity to build their subtree after the content is laid out...

knopp:
So, without using reassemble (which is quite hacky), you could change `ContentLayers` to take list of build methods instead of widgets. That way you will get new widget instances for every `buildLayer` call and Flutter will update nested elements in the tree.

I tried changing `final List<Widget> overlays` to `final List<WidgetBuilder> overlays` (same for underlays) and it seems to work fine.

knopp:
https://github.com/superlistapp/super_editor/pull/1240

knopp:
We may need a PR for main to replace `textScaleFactorOf` with `textScalerOf`

matthew-carroll:
@knopp I switched the layers to builders. I also updated the caret overlay to only start blinking when there's a non-null selection. Let me know if this PR looks good to you.

super_editor/lib/src/infrastructure/content_layers.dart
```diff
@@ -132,19 +142,29 @@ class ContentLayersElement extends RenderObjectElement {
     //        need to be rebuilt. Create a way for layers to opt-in to this behavior.
 
     owner!.buildScope(this, () {
-      _deactivateLayers();
+      // _deactivateLayers();
```
knopp, line 145: Remove this?

super_editor/pubspec.yaml
```diff
@@ -37,13 +37,13 @@ dependencies:
     sdk: flutter
   flutter_test_robots: ^0.0.17
 
-#dependency_overrides:
+dependency_overrides:
 #  # Override to local mono-repo path so devs can test this repo
 #  # against changes that they're making to other mono-repo packages
 #  attributed_text:
 #    path: ../attributed_text
-#  super_text_layout:
```
knopp, line 45: Is this committed on purpose?

knopp (approved):
Just some nits otherwise LGTM.

super_editor/pubspec.yaml
matthew-carroll: Yeah - I changed the blink controller API, which is in super_text_layout. I'll need to cut a new release of super_text_layout to be able to remove this override.

PR #1251, opened 2023-07-24
[SuperEditor] Fix image insertion bugs (Partially resolves #1209)

### LINKED ISSUE #1209: [SuperEditor] Image issues (atomic commands)
@matthew-carroll I've found several image issues since the atomic commands branch was merged. All issues are present on Android (I will test iOS shortly), and one is present on Mac.

All of the following were run on the `stable` branch of the example app. The only change I made to the code was to add an icon button that calls the following:

```
  void _insertImage() {
    commonOps.insertImage('https://storage.googleapis.com/cms-storage-bucket/c823e53b3a1a7b0d36a9.png');
  }
```

### 1. When an image is first added, the carat doesn't move to the ParagraphNode inserted below the image
Present on **Android**

However, if the same image is added again, this doesn't occur. I'm assuming it has something to do with fetching the image from the network the first time.

https://github.com/superlistapp/super_editor/assets/6467808/146372b4-dd8a-4f09-91f3-647bcbb1db08

### 2. When the carat is to the right of an image and enter/return is pressed, new ParagraphNode's are inserted below (as expected), but the carat does not move
Present on **Android**

https://github.com/superlistapp/super_editor/assets/6467808/0c3cb5f3-dfb1-4dd6-add8-5e8ce7b6d0d5

### 3. Can't add an image in a non-empty paragraph *after a previous image was removed*
Present on **Android** and **Mac**

Add an image, remove it, then add it again where the carat now rests. The following error occurs:

```
The following _TypeError was thrown while dispatching notifications for PausableValueNotifier<DocumentSelection?>:
type 'UpstreamDownstreamNodePosition' is not a subtype of type 'TextPosition' in type cast

#0      SuperEditorState._updateComposerPreferencesAtSelection
super_editor.dart:468
#1      ChangeNotifier.notifyListeners
change_notifier.dart:403
#2      ValueNotifier.value=
change_notifier.dart:530
#3      PausableValueNotifier.resumeNotifications
pausable_value_notifier.dart:49
#4      MutableDocumentComposer.onTransactionEnd
document_composer.dart:156
#5      Editor.execute
editor.dart:164
#6      CommonEditorOperations._insertBlockLevelContent
common_editor_operations.dart:1787
#7      CommonEditorOperations.insertImage
common_editor_operations.dart:1711
#8      KeyboardEditingToolbar._insertImage
mobile_toolbar.dart:193
#9      _InkResponseState.handleTap
ink_well.dart:1154
#10     GestureRecognizer.invokeCallback
recognizer.dart:275
#11     TapGestureRecognizer.handleTapUp
tap.dart:654
#12     BaseTapGestureRecognizer._checkUp
tap.dart:311
#13     BaseTapGestureRecognizer.acceptGesture
tap.dart:281
#14     GestureArenaManager.sweep
arena.dart:167
#15     GestureBinding.handleEvent
binding.dart:469
#16     GestureBinding.dispatchEvent
binding.dart:445
#17     RendererBinding.dispatchEvent
binding.dart:331
#18     GestureBinding._handlePointerEventImmediately
binding.dart:400
#19     GestureBinding.handlePointerEvent
binding.dart:363
#20     GestureBinding._flushPointerEventQueue
binding.dart:320
#21     GestureBinding._handlePointerDataPacket
binding.dart:293
#22     _invoke1 (dart:ui/hooks.dart:158:13)
#23     PlatformDispatcher._dispatchPointerDataPacket (dart:ui/platform_dispatcher.dart:382:7)
#24     _dispatchPointerDataPacket (dart:ui/hooks.dart:91:31)
The PausableValueNotifier<DocumentSelection?> sending notification was: PausableValueNotifier<DocumentSelection?>#9983b([DocumentSelection] - 
  base: ([DocumentPosition] - node: "44a3dbf9-65c6-4068-98dd-a585c901ea50", position: ([UpstreamDownstreamNodePosition] - TextAffinity.downstream)),
  extent: ([DocumentPosition] - node: "44a3dbf9-65c6-4068-98dd-a585c901ea50", position: ([UpstreamDownstreamNodePosition] - TextAffinity.downstream)))
```

https://github.com/superlistapp/super_editor/assets/6467808/9965ccb2-d20c-4995-8755-defb9657af07

https://github.com/superlistapp/super_editor/assets/6467808/8dd8ad71-292e-4ea1-9f2a-c50266ebdfa7

angelosilvestre:
[SuperEditor] Fix image insertion bugs. Partially resolves #1209

We have some issues related to images:

-  Inserting a paragraph when the caret sits after an image doesn't change the selection:

https://github.com/superlistapp/super_editor/assets/6467808/0c3cb5f3-dfb1-4dd6-add8-5e8ce7b6d0d5

The root cause is that we aren't executing a `ChangeSelectionRequest` after inserting the image.

This PR changes the image insertion to execute a `ChangeSelectionRequest`.

- Inserting an image when the caret sits at the end of a paragraph (with upstream affinity) throws an exception:

https://github.com/superlistapp/super_editor/assets/6467808/9965ccb2-d20c-4995-8755-defb9657af07

We are comparing the text positions as `paragraphPosition == endOfParagraph`, which returns `false` if the affinity is different. 

We already have tests for image insertion. However, in the tests, we are always producing downstream selections, and the issue only happens with upstream affinity. May we should have tests for both affinities?

This PR changes the comparison to only care about the offset, ignoring the affinity.

The other issue reported in the ticket is caused because the image component resizes itself after the image is loaded, and we aren't rebuilding the caret. I think this should be handled by refactoring the mobile widget tree to use the same layer structure as the desktop.

### DIFF
--- a/super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
+++ b/super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
@@ -575,6 +575,16 @@ class TextDeltasDocumentEditor {
               text: AttributedText(text: ''),
             ),
           ),
+          ChangeSelectionRequest(
+            DocumentSelection.collapsed(
+              position: DocumentPosition(
+                nodeId: newNodeId,
+                nodePosition: const TextNodePosition(offset: 0),
+              ),
+            ),
+            SelectionChangeType.insertContent,
+            SelectionReason.userInteraction,
+          ),
         ]);
       } else {
         // The caret sits on the upstream edge of block-level content. Insert
--- a/super_editor/lib/src/default_editor/multi_node_editing.dart
+++ b/super_editor/lib/src/default_editor/multi_node_editing.dart
@@ -164,7 +164,7 @@ class InsertNodeAtCaretCommand extends EditCommand {
           nodePosition: selectedNode.beginningPosition,
         ),
       );
-    } else if (paragraphPosition == beginningOfParagraph) {
+    } else if (paragraphPosition.offset == beginningOfParagraph.offset) {
       // Insert block item after the paragraph.
       document.insertNodeAt(document.getNodeIndexById(selectedNode.id), newNode);
       executor.logChanges([
@@ -179,19 +179,26 @@ class InsertNodeAtCaretCommand extends EditCommand {
           nodePosition: selectedNode.beginningPosition,
         ),
       );
-    } else if (paragraphPosition == endOfParagraph) {
-      // Insert block item after the paragraph.
-      document.insertNodeAfter(existingNode: selectedNode, newNode: newNode);
+    } else if (paragraphPosition.offset == endOfParagraph.offset) {
+      final emptyParagraph = ParagraphNode(id: Editor.createNodeId(), text: AttributedText());
+
+      // Insert block item after the paragraph and insert a new empty paragraph.
+      document
+        ..insertNodeAfter(existingNode: selectedNode, newNode: newNode)
+        ..insertNodeAfter(existingNode: newNode, newNode: emptyParagraph);
       executor.logChanges([
         DocumentEdit(
           NodeInsertedEvent(newNode.id, document.getNodeIndexById(newNode.id)),
-        )
+        ),
+        DocumentEdit(
+          NodeInsertedEvent(emptyParagraph.id, document.getNodeIndexById(emptyParagraph.id)),
+        ),
       ]);
 
       newSelection = DocumentSelection.collapsed(
         position: DocumentPosition(
-          nodeId: selectedNodeId,
-          nodePosition: newNode.endPosition,
+          nodeId: emptyParagraph.id,
+          nodePosition: emptyParagraph.endPosition,
         ),
       );
     } else {
--- a/super_editor/test/super_editor/supereditor_content_insertion_test.dart
+++ b/super_editor/test/super_editor/supereditor_content_insertion_test.dart
@@ -1,15 +1,11 @@
 import 'dart:ui';
 
 import 'package:flutter_test/flutter_test.dart';
-import 'package:super_editor/src/core/document.dart';
-import 'package:super_editor/src/core/document_selection.dart';
-import 'package:super_editor/src/default_editor/horizontal_rule.dart';
-import 'package:super_editor/src/default_editor/image.dart';
-import 'package:super_editor/src/default_editor/paragraph.dart';
-import 'package:super_editor/src/default_editor/text.dart';
+import 'package:flutter_test_robots/flutter_test_robots.dart';
 
 import 'package:super_editor/src/test/super_editor_test/supereditor_inspector.dart';
 import 'package:super_editor/src/test/super_editor_test/supereditor_robot.dart';
+import 'package:super_editor/super_editor.dart';
 
 import '../test_tools.dart';
 import 'document_test_tools.dart';
@@ -104,13 +100,18 @@ void main() {
         // Pump a widget with an arbitrary size for the images.
         final context = await tester //
             .createDocument()
-            .fromMarkdown("First paragraph")
-            .withAddedComponents(
+            .fromMarkdown("""First paragraph
+            
+Second paragraph"""). //
+            withAddedComponents(
           [const FakeImageComponentBuilder(size: Size(100, 100))],
         ).pump();
 
-        // Place caret at the end of the paragraph.
-        await tester.placeCaretInParagraph(context.editContext.document.nodes.first.id, 15);
+        // Place caret at the end of the first paragraph by selecting the second paragraph and pressing left.
+        //
+        // This results in an upstream text affinity.
+        await tester.placeCaretInParagraph(context.editContext.document.nodes.last.id, 0);
+        await tester.pressLeftArrow();
 
         // Insert the image at the current selection.
         context.editContext.commonOps.insertImage('http://image.fake');
@@ -119,7 +120,7 @@ void main() {
         final doc = SuperEditorInspector.findDocument()!;
 
         // Ensure that two nodes were inserted.
-        expect(doc.nodes.length, 3);
+        expect(doc.nodes.length, 4);
 
         // Ensure that the first node remains unchanged.
         expect(doc.nodes[0], isA<ParagraphNode>());
@@ -132,12 +133,12 @@ void main() {
         expect(doc.nodes[2], isA<ParagraphNode>());
         expect((doc.nodes[2] as ParagraphNode).text.text, '');
 
-        // Ensure the selection was placed at the beginning of the last paragraph.
+        // Ensure the selection was placed at the beginning of the newly created paragraph.
         expect(
           SuperEditorInspector.findDocumentSelection(),
           DocumentSelection.collapsed(
             position: DocumentPosition(
-              nodeId: doc.nodes.last.id,
+              nodeId: doc.nodes[2].id,
               nodePosition: const TextNodePosition(offset: 0),
             ),
           ),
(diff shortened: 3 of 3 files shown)


matthew-carroll:
> The other issue reported in the ticket is caused because the image component resizes itself after the image is loaded, and we aren't rebuilding the caret. I think this should be handled by refactoring the mobile widget tree to use the same layer structure as the desktop.

Please add a reference to that other ticket that mentions this image sizing rebuild issue.

Also, is it possible for you to write a failing test for that image resize rebuild behavior?

super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
```diff
@@ -575,6 +575,16 @@ class TextDeltasDocumentEditor {
...
+          ChangeSelectionRequest(
+            DocumentSelection.collapsed(
+              position: DocumentPosition(
+                nodeId: newNodeId,
+                nodePosition: const TextNodePosition(offset: 0),
+              ),
+            ),
+            SelectionChangeType.insertContent,
+            SelectionReason.userInteraction,
+          ),
```
matthew-carroll, line 587: If this execution was wrong because it didn't have a `ChangeSelectionRequest`, then is the `execute()` call below this one also wrong?

super_editor/lib/src/default_editor/multi_node_editing.dart
```diff
@@ -179,19 +179,26 @@ class InsertNodeAtCaretCommand extends EditCommand {
           nodePosition: selectedNode.beginningPosition,
         ),
       );
-    } else if (paragraphPosition == endOfParagraph) {
-      // Insert block item after the paragraph.
-      document.insertNodeAfter(existingNode: selectedNode, newNode: newNode);
+    } else if (paragraphPosition.offset == endOfParagraph.offset) {
+      final emptyParagraph = ParagraphNode(id: Editor.createNodeId(), text: AttributedText());
```
matthew-carroll, line 183: Why did the inside of this `else if` block change so much?

super_editor/test/super_editor/supereditor_content_insertion_test.dart
```diff
@@ -104,13 +100,18 @@ void main() {
...
+            withAddedComponents(
           [const FakeImageComponentBuilder(size: Size(100, 100))],
         ).pump();
 
-        // Place caret at the end of the paragraph.
-        await tester.placeCaretInParagraph(context.editContext.document.nodes.first.id, 15);
+        // Place caret at the end of the first paragraph by selecting the second paragraph and pressing left.
+        //
+        // This results in an upstream text affinity.
+        await tester.placeCaretInParagraph(context.editContext.document.nodes.last.id, 0);
```
matthew-carroll, line 113: This change takes an existing test and alters the test, and does so without altering the name of the test, which means we're describing the same intention while testing at least a slightly different behavior.

What if an issue showed up with a downstream selection? Now we've lost that test.

If you believe that testing affinity is important, instead of cannibalizing an existing test, you should update the name of the current test to mention a downstream selection, and then you should write another test that mentions an upstream selection.

super_editor/test/super_editor/supereditor_content_insertion_test.dart
```diff
@@ -341,5 +347,180 @@ void main() {
         );
       });
     });
+
+    group('inserts a paragraph', () {
+      testWidgetsOnDesktop('on ENTER at the end of an image', (tester) async {
```
matthew-carroll, line 352: These test names are difficult to read with the way that they begin, e.g., "on ENTER at", "upon new line insertion". Let's use an active tense for these descriptions, rather than a passive tense.

"when the user presses ENTER at the end of an image"

"when the user presses the newline button on the software keyboard at the end of an image (on Android)"

Also you've defined an Android-specific test, and then a generic mobile test. It seems to me like the Android test covers the Android case, and the test that you've written as a generic mobile test is really just a reflection of iOS behavior.

super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
angelosilvestre: I'm not sure what is the expected behavior for this case. Currently, when the selection sits before the image, pressing ENTER inserts a paragraph before the image and keeps the selection where it is.

Should the selection move to the newly created paragraph? As the paragraph is inserted before, I think it makes sense keeping the selection at the image.

super_editor/lib/src/default_editor/multi_node_editing.dart
angelosilvestre: This is a change that was made before and I think it got lost in the atomic commands migration. 

Previously, we have this dart doc:

```dart
/// If the selection extent sits at the end of a paragraph, the [blockNode]
/// is inserted as a new node after that paragraph, and then a new
/// empty paragraph is inserted after the [blockNode].
```

Which isn't happening right now.

matthew-carroll: So this code was here before I migrated to atomic commands?

super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
matthew-carroll: Ok, that makes sense why one changes the selection and the other doesn't.

super_editor/lib/src/default_editor/multi_node_editing.dart
matthew-carroll: Also, if this was the behavior before the migration, it indicates that we must be missing some tests for this, given that I missed the code and nothing failed.

angelosilvestre: Yeah, I think it was written while the atomic commands PR was in progress.

super_editor/test/super_editor/supereditor_content_insertion_test.dart
matthew-carroll: This comment was marked as resolved, but I don't see corresponding changes. I still see existing tests which were modified. Why are we altering existing behavior expectations?

angelosilvestre: I forgot to push the changes 🤦‍♂️.

angelosilvestre:
> Please add a reference to that other ticket that mentions this image sizing rebuild issue.
> 
> Also, is it possible for you to write a failing test for that image resize rebuild behavior?

I think this is probably the same issue as https://github.com/superlistapp/super_editor/issues/1022.

matthew-carroll:
> > Please add a reference to that other ticket that mentions this image sizing rebuild issue.
> > Also, is it possible for you to write a failing test for that image resize rebuild behavior?
> 
> I think this is probably the same issue as #1022.

The issue that you linked is closed. So no current problem should be tied to that. Either that issue wasn't really solved, or this is a different problem, right?

angelosilvestre:
> The issue that you linked is closed. So no current problem should be tied to that. Either that issue wasn't really solved, or this is a different problem, right?

I think this is related to this open PR: https://github.com/superlistapp/super_editor/pull/1256, where we will use `ContentLayers`. After this is merged, we should re-build when the child size changes, right?

matthew-carroll:
Hopefully that's the case. You should probably file an issue to come back and check, just in case that doesn't happen.

angelosilvestre:
OK. Will do.

matthew-carroll:
Should I review this PR again?

angelosilvestre:
@matthew-carroll Yes, please

angelosilvestre:
> The other issue reported in the ticket is caused because the image component resizes itself after the image is loaded, and we aren't rebuilding the caret. I think this should be handled by refactoring the mobile widget tree to use the same layer structure as the desktop.

https://github.com/superlistapp/super_editor/pull/1256#event-10051096376 Solved this issue.

@matthew-carroll

matthew-carroll (approved):
LGTM

Issue #1140, opened 2023-05-16
Abnormal cursor position and component click events caused by gesture recognition


Dboy233:
# Bug Description

super_editor: ^0.2.5

 When I click on the CheckBox, the editor should not gain focus. The click event should only be handled by the CheckBox. I looked at the source code and found that it was the "TapSequenceGestureRecognizer" that intercepted all click events. So I investigated this and compared it to the implementation of "TextField". I found that the problem could be solved by simply following the implementation of "TextField".

https://github.com/superlistapp/super_editor/assets/37604230/ae64e18d-f692-40dd-89dc-6b5b3f4d3fee

raulmabe-labhouse:
@Dboy233 Testing your solution on iOS, stable branch, Flutter 3.10.0, editor selection is broken (can't focus).

Dboy233:
@raulmabe-labhouse 

Since I didn't have an iPhone, I robbed a friend's phone to test it.

After I tested it, I found that using `TextSelectionGestureDetector` does not work properly on ios devices, so I found a replacement - `TapAndPanGestureRecognizer`. 

'TapDragDownDetails' has a member variable that counts the number of presses --- `consecutiveTapCount`.

`TextField` also uses this variable to judge multiple clicks.

iOS 'Scrollbar' doesn't seem to work properly. The reason may be that vertical scrolling events are being intercepted。

<img width="732" alt="图片" src="https://github.com/superlistapp/super_editor/assets/37604230/80d95aea-a8f6-4af9-bce8-4e5d7ee20c80">

<img width="757" alt="图片" src="https://github.com/superlistapp/super_editor/assets/37604230/6ae3555c-bb06-4930-a2ac-2bdea502bef9">

This is my test video：

iOS:

https://github.com/superlistapp/super_editor/assets/37604230/62f458e3-4a34-4e6f-a862-e6b11940427d

Android:

https://github.com/superlistapp/super_editor/assets/37604230/aeb6e077-d78b-4669-89b2-cb9eca5a5666


This zip file contains two files that I modified.

[change_code.zip](https://github.com/superlistapp/super_editor/files/11499096/change_code.zip)

Dboy233:
I think this issue should be taken seriously, as gesture operation is particularly important for mobile platforms.

matthew-carroll:
It's not clear to me what specific problem or problems are being described here.

I see a lot of reference to implementation code, but that's not particularly useful. I also see some mention of response time, but nothing specific.

[...]

PR #1256, opened 2023-07-26
Unified SuperEditor mobile and desktop gesture and content layers (Resolves #927)

### LINKED ISSUE #927: [SuperEditor] - Move mobile gesture recognition under document, like desktop

matthew-carroll:
When we implemented `ContentLayers` for desktop, we didn't apply it in the same way to mobile, because it required some mobile rework. This PR reworks mobile such that nearly all of the widget tree is the same for all platforms. The only difference now is the actual gesture handler widget for each platform.

This PR also changes the way we set the gesture area's bounds. Previously we used a `LayoutBuilder` to find out how big the viewport was, and then we passed that info to the gesture area inside of the viewport. But `LayoutBuilder` blows up when trying to calculate intrinsic height. This PR introduces an approach that should work with intrinsic height. It's explained, in detail, in comments in the code.

Angelo found that mobile handles weren't re-positioning themselves when document components change size, e.g., animate larger or smaller. This was because only document content layers rebuild when that happens, not the mobile handle overlays. I introduced a document content layer that positions invisible widgets near the base, extent, and around the entire selection so that handles can anchor near them. I then converted the mobile overlays to use the layer links.

This PR also  converts `SuperReader` to use `ContentLayers` and layer links for drag handle positioning.


super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -606,135 +610,330 @@ class SuperEditorState extends State<SuperEditor> {
...
+    required this.contentConstraints,
+    required super.child,
+  });
+
+  /// The layout constraints that apply outside of the scrollable viewport.
+  ///
+  /// This widget is expected to build around the scrollable viewport. This
+  /// widget then reports its layout constraints to this notifier to be used
+  /// by a descendant [_SuperEditorGestureBounds].
+  final ValueNotifier<BoxConstraints> contentConstraints;
```
angelosilvestre, line 864: As the value is set in `performLayout`, will it break if `contentContraints` is used in a `ListenableBuilder`? If so, I think it would be good to add to the docs that we shouldn't use it that way.

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -606,135 +610,330 @@ class SuperEditorState extends State<SuperEditor> {
...
+  }
+
+  @override
+  void updateRenderObject(BuildContext context, _RenderSuperEditorBounds renderObject) {
+    renderObject.contentConstraints = contentConstraints;
+  }
+}
+
+class _RenderSuperEditorBounds extends RenderProxyBox {
+  late ValueNotifier<BoxConstraints> contentConstraints;
```
angelosilvestre, line 879: If `contentConstraints` changes, should we call `markNeedsLayout` ?

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -606,135 +610,330 @@ class SuperEditorState extends State<SuperEditor> {
...
+      ..contentConstraints = contentConstraints;
+  }
+
+  @override
+  void updateRenderObject(BuildContext context, _RenderSuperEditorBounds renderObject) {
+    renderObject.contentConstraints = contentConstraints;
+  }
+}
+
+class _RenderSuperEditorBounds extends RenderProxyBox {
```
angelosilvestre, line 878: By convention, shouldn't we call this `_RenderSuperEditorViewportBounds`?

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -606,135 +610,330 @@ class SuperEditorState extends State<SuperEditor> {
...
+    return _RenderSuperEditorContentBounds()..contentConstraints = contentConstraints;
+  }
+
+  @override
+  void updateRenderObject(BuildContext context, _RenderSuperEditorContentBounds renderObject) {
+    renderObject.contentConstraints = contentConstraints;
+  }
+}
+
+class _RenderSuperEditorContentBounds extends RenderProxyBox {
```
angelosilvestre, line 928: By convention, shouldn't we call this `_SuperEditorGestureBounds`?

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -606,135 +610,330 @@ class SuperEditorState extends State<SuperEditor> {
...
+  late ValueNotifier<BoxConstraints> contentConstraints;
```
angelosilvestre, line 929: Same question here about calling `markNeedsLayout`.

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -819,18 +1018,32 @@ class FunctionalDocumentLayerBuilder implements DocumentLayerBuilder {
...
+
   @override
   Widget build(BuildContext context, SuperEditorContext editContext) {
+    // By default, don't show a caret on mobile because SuperEditor displays
+    // mobile carets and handles elsewhere. This can be overridden by settings
+    // `displayOnAllPlatforms` to true.
+    if (!displayOnAllPlatforms &&
+        (defaultTargetPlatform == TargetPlatform.android || defaultTargetPlatform == TargetPlatform.iOS)) {
+      return const SizedBox();
+    }
```
angelosilvestre, line 1045: What do you think of passing the gesture mode instead of checking the `defaultTargetPlatform`?

If, for whathever reason, I run the editor on desktop and use `DocumentGestureMode.android` as the gesture mode, I will see two carets.

angelosilvestre (commented):
I left a few comments.

angelosilvestre:
This seems to also fix https://github.com/superlistapp/super_editor/issues/1140.

Unfortunately, it doesn't solve this: https://github.com/superlistapp/super_editor/pull/1251#issuecomment-1663106888

The "Animated task height demo" also doesn't work as expected. The caret stays at the wrong offset:

Screenshot_20230803-193929

super_editor/lib/src/default_editor/super_editor.dart
matthew-carroll: I'll add a comment, but the constraints are expected to be set by an ancestor and then used by a descendent. If the ancestor runs layout, then so will the descendent. If only the descendent runs layout, it means the ancestor hasn't changed size.

super_editor/lib/src/default_editor/super_editor.dart
matthew-carroll: This builder is created by the app developer, which would make it a bit strange to expect those developers to also pass a `DocumentGestureMode` to the builder, too. But I'll add the ability to pass an override `TargetPlatform`.

matthew-carroll:
I think the reason the caret and handle are in the wrong place is because on mobile those are implemented in an app overlay, not a document layer. The document layers rebuild, but the app overlay doesn't rebuild.

I think that probably the answer is to use a leader/follower for caret and handle positions, and place the leaders in the document layer. I'll try to make that change.

angelosilvestre:
Double tapping causes the drag handle to quickly flash at the top left of the screen:

https://github.com/superlistapp/super_editor/assets/7597082/ce464c03-df51-4f4c-96d8-3b294692cf71

matthew-carroll:
Ok. I'll look at that later. Any other issues?

angelosilvestre:
This is the only issue I've found so far.

matthew-carroll:
@angelosilvestre can you verify that the handle flickering is fixed? Also can you continue with the review to see if there's anything else that needs to be addressed?

matthew-carroll:
I'm going to merge this because all indications are that everything is working and I'd like to move forward, but @angelosilvestre should still do a final test of this behavior after the merge.

angelosilvestre (commented):
LGTM

PR #1258, opened 2023-07-26
[SuperEditor] Moves caret on downstream when right arrow's pressed while block is selected (Resolves #1247)

### LINKED ISSUE #1247: BlockNode.selectDownstreamPosition() returns the incorrect order (simple fix)
Every DocumentNode has to implement select**Downstream**Position but the base class BlockNodes does that incorrectly. It looks like the code was copied from select**Upstream**Position, but the changes applied are not correct.

Example: Select an image, then press the right arrow key. One would expect the selection to move to the right side of the image but it moves to the left side. The issue is the BlockNode.selectDownstreamPosition.

Fix: Reverse the two return parts of the function (see the comment in the following code).

```dart
@override
  UpstreamDownstreamNodePosition selectUpstreamPosition(NodePosition position1, NodePosition position2) {
    if (position1 is! UpstreamDownstreamNodePosition) {
      throw Exception(
          'Expected a UpstreamDownstreamNodePosition for position1 but received a ${position1.runtimeType}');
    }
    if (position2 is! UpstreamDownstreamNodePosition) {
      throw Exception(
          'Expected a UpstreamDownstreamNodePosition for position2 but received a ${position2.runtimeType}');
    }

    if (position1.affinity == TextAffinity.upstream || position2.affinity == TextAffinity.upstream) {
      return const UpstreamDownstreamNodePosition.upstream();
    } else {
      return const UpstreamDownstreamNodePosition.downstream();
    }
  }

  @override
  UpstreamDownstreamNodePosition selectDownstreamPosition(NodePosition position1, NodePosition position2) {
    if (position1 is! UpstreamDownstreamNodePosition) {
      throw Exception(
          'Expected a UpstreamDownstreamNodePosition for position1 but received a ${position1.runtimeType}');
    }
    if (position2 is! UpstreamDownstreamNodePosition) {
      throw Exception(
          'Expected a UpstreamDownstreamNodePosition for position2 but received a ${position2.runtimeType}');
    }

    if (position1.affinity == TextAffinity.downstream || position2.affinity == TextAffinity.downstream) {
      return const UpstreamDownstreamNodePosition.upstream();  // <-- wrong, should be downstream()
    } else {
      return const UpstreamDownstreamNodePosition.downstream();  // <-- wrong, should be upstream()
    }
  }
```

Had no luck with my previous pull request, maybe somebody else is interested in fixing this :)

rutvik110:
Resolves #1247

**Description:**

The current implementation of the `selectDownstreamPosition` on the `BlockNode` returns the wrong `UpstreamDownstreamNodePosition` which results in a bug where the caret moves to upstream on the selected block when right arrow key is pressed, as seen in the video below

https://github.com/superlistapp/super_editor/assets/65209850/c084456d-6992-4e13-9c8c-b496322123bf

This affected the Nodes based on `BlockNode`, `ImageNode` and `HorizontalRuleNode`.

**Solution:**

Correct behaviour is to move the caret position on the downstream on the block. This PR resolves this by returning the `UpstreamDownstreamNodePosition` with appropriate `TextAffinity`, which moves the caret position on the downstream when right arrow key is pressed with the block selected.

https://github.com/superlistapp/super_editor/assets/65209850/77838d2c-bab1-41e4-93c6-37544ebe6238

Also adds Tests to ensure the behaviour remains consistent.

### DIFF
--- a/super_editor/test/src/default_editor/upstream_downstream_selection_test.dart
+++ b/super_editor/test/src/default_editor/upstream_downstream_selection_test.dart
@@ -278,7 +278,8 @@ void main() {
         expect(composer.selection!.extent.nodePosition, const UpstreamDownstreamNodePosition.upstream());
       });
 
-      testWidgets("right arrow moves caret to downstream edge of the selection made on a block node", (tester) async {
+      testWidgets("right arrow collapses the expanded selection around block node to a caret on the downstream edge",
+          (tester) async {
         await tester
             .createDocument()
             .withCustomContent(paragraphThenHrThenParagraphDoc())
--- a/super_editor/test/src/default_editor/upstream_downstream_selection_test.dart
+++ b/super_editor/test/src/default_editor/upstream_downstream_selection_test.dart
@@ -277,6 +277,26 @@ void main() {
       });
     });
 
+    group("selection", () {
+      testWidgets("keeps caret downstream when right arrow is pressed while a block is selected", (tester) async {
+        final document = paragraphThenHrThenParagraphDoc();
+        final composer = MutableDocumentComposer(
+          initialSelection: const DocumentSelection(
+            base: DocumentPosition(nodeId: "2", nodePosition: UpstreamDownstreamNodePosition.upstream()),
+            extent: DocumentPosition(nodeId: "2", nodePosition: UpstreamDownstreamNodePosition.downstream()),
+          ),
+        );
+        await tester.pumpWidget(_buildHardwareKeyboardEditor(document, composer));
+
+        await tester.sendKeyEvent(LogicalKeyboardKey.arrowRight);
+        await tester.pump();
+
+        expect(composer.selection!.isCollapsed, true);
+        expect(composer.selection!.extent.nodeId, "2");
+        expect(composer.selection!.extent.nodePosition, const UpstreamDownstreamNodePosition.downstream());
+      });
+    });
+
     group("deletion", () {
       testWidgets("backspace moves caret to node above when caret is on upstream edge", (tester) async {
         final document = paragraphThenHrThenParagraphDoc();
--- a/super_editor/test/src/default_editor/upstream_downstream_selection_test.dart
+++ b/super_editor/test/src/default_editor/upstream_downstream_selection_test.dart
@@ -285,19 +285,17 @@ void main() {
             .withEditorSize(const Size(300, 300))
             .pump();
 
-        final document = SuperEditorInspector.findDocument()!;
-        final hrNode = document.getNodeById("2") as HorizontalRuleNode;
-
-        await tester.tapAtDocumentPosition(DocumentPosition(
-          nodeId: hrNode.id,
-          nodePosition: hrNode.beginningPosition,
+        await tester.tapAtDocumentPosition(const DocumentPosition(
+          nodeId: "2",
+          nodePosition: UpstreamDownstreamNodePosition.upstream(),
         ));
+        // Wait till the tap event is complete and the latest selection is available.
         await tester.pumpAndSettle();
 
         final testerGesture = await tester.startDocumentDragFromPosition(
-          from: DocumentPosition(
-            nodeId: hrNode.id,
-            nodePosition: hrNode.beginningPosition,
+          from: const DocumentPosition(
+            nodeId: "2",
+            nodePosition: UpstreamDownstreamNodePosition.upstream(),
           ),
         );
 
@@ -310,14 +308,14 @@ void main() {
 
         expect(
           SuperEditorInspector.findDocumentSelection(),
-          DocumentSelection(
+          const DocumentSelection(
             base: DocumentPosition(
-              nodeId: hrNode.id,
-              nodePosition: hrNode.beginningPosition,
+              nodeId: "2",
+              nodePosition: UpstreamDownstreamNodePosition.upstream(),
             ),
             extent: DocumentPosition(
-              nodeId: hrNode.id,
-              nodePosition: hrNode.endPosition,
+              nodeId: "2",
+              nodePosition: UpstreamDownstreamNodePosition.downstream(),
             ),
           ),
         );
@@ -328,7 +326,7 @@ void main() {
         final selection = SuperEditorInspector.findDocumentSelection();
 
         expect(selection!.isCollapsed, true);
-        expect(selection.extent.nodeId, hrNode.id);
+        expect(selection.extent.nodeId, "2");
         expect(selection.extent.nodePosition, const UpstreamDownstreamNodePosition.downstream());
       });
     });
--- a/super_editor/test/src/default_editor/upstream_downstream_selection_test.dart
+++ b/super_editor/test/src/default_editor/upstream_downstream_selection_test.dart
@@ -285,26 +285,13 @@ void main() {
             .withEditorSize(const Size(300, 300))
             .pump();
 
-        await tester.tapAtDocumentPosition(const DocumentPosition(
-          nodeId: "2",
-          nodePosition: UpstreamDownstreamNodePosition.upstream(),
-        ));
-        // Wait till the tap event is complete and the latest selection is available.
-        await tester.pumpAndSettle();
-
-        final testerGesture = await tester.startDocumentDragFromPosition(
-          from: const DocumentPosition(
+        for (var i = 0; i < 2; i++) {
+          await tester.tapAtDocumentPosition(const DocumentPosition(
             nodeId: "2",
             nodePosition: UpstreamDownstreamNodePosition.upstream(),
-          ),
-        );
-
-        // Move by sufficient distance to select the hr.
-        await testerGesture.moveBy(const Offset(300, 0));
-        await tester.pump();
-
-        await tester.endDocumentDragGesture(testerGesture);
-        await tester.pump();
+          ));
+          await tester.pump(kTapMinTime + const Duration(milliseconds: 1));
+        }
 
         expect(
           SuperEditorInspector.findDocumentSelection(),
--- a/super_editor/test/src/default_editor/upstream_downstream_selection_test.dart
+++ b/super_editor/test/src/default_editor/upstream_downstream_selection_test.dart
@@ -286,13 +286,18 @@ void main() {
             .withEditorSize(const Size(300, 300))
             .pump();
 
-        for (var i = 0; i < 2; i++) {
-          await tester.tapAtDocumentPosition(const DocumentPosition(
-            nodeId: "2",
-            nodePosition: UpstreamDownstreamNodePosition.upstream(),
-          ));
-          await tester.pump(kTapMinTime + const Duration(milliseconds: 1));
-        }
+        // TODO: Move this to Super Editor tooling.
+        await tester.tapAtDocumentPosition(const DocumentPosition(
+          nodeId: "2",
+          nodePosition: UpstreamDownstreamNodePosition.upstream(),
+        ));
+        await tester.pump(kTapMinTime + const Duration(milliseconds: 1));
+
+        await tester.tapAtDocumentPosition(const DocumentPosition(
+          nodeId: "2",
+          nodePosition: UpstreamDownstreamNodePosition.upstream(),
+        ));
+        await tester.pump(kTapMinTime + const Duration(milliseconds: 1));
 
         expect(
           SuperEditorInspector.findDocumentSelection(),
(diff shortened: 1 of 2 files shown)


super_editor/test/src/default_editor/upstream_downstream_selection_test.dart
```diff
@@ -277,6 +277,26 @@ void main() {
       });
     });
 
+    group("selection", () {
+      testWidgets("keeps caret downstream when right arrow is pressed while a block is selected", (tester) async {
+        final document = paragraphThenHrThenParagraphDoc();
+        final composer = MutableDocumentComposer(
+          initialSelection: const DocumentSelection(
```
matthew-carroll, line 284: When possible, we prefer to use simulated gestures to place selections, so that we ensure real users can do the same. Is it possible to achieve your desired selection with our testing selection APIs?

super_editor/test/src/default_editor/upstream_downstream_selection_test.dart
```diff
@@ -277,6 +277,26 @@ void main() {
       });
     });
 
+    group("selection", () {
+      testWidgets("keeps caret downstream when right arrow is pressed while a block is selected", (tester) async {
+        final document = paragraphThenHrThenParagraphDoc();
```
matthew-carroll, line 282: Do you need to create the `document` here? Look at many of our other tests that use our test builder APIs to constructor a `SuperEditor` experience.

super_editor/test/src/default_editor/upstream_downstream_selection_test.dart
```diff
@@ -277,6 +277,26 @@ void main() {
...
+        final composer = MutableDocumentComposer(
+          initialSelection: const DocumentSelection(
+            base: DocumentPosition(nodeId: "2", nodePosition: UpstreamDownstreamNodePosition.upstream()),
+            extent: DocumentPosition(nodeId: "2", nodePosition: UpstreamDownstreamNodePosition.downstream()),
+          ),
+        );
+        await tester.pumpWidget(_buildHardwareKeyboardEditor(document, composer));
```
matthew-carroll, line 289: Is there a reason that you're using `_buildHardwareKeyboardEditor` in this test?

super_editor/test/src/default_editor/upstream_downstream_selection_test.dart
```diff
@@ -277,6 +277,26 @@ void main() {
       });
     });
 
+    group("selection", () {
```
matthew-carroll, line 280: I can't see what groups are above this, but given that we're in the `upstream_downstream_selection_test` is there really not an existing group with an appropriate name?

super_editor/test/src/default_editor/upstream_downstream_selection_test.dart
```diff
@@ -277,6 +277,26 @@ void main() {
...
+      testWidgets("keeps caret downstream when right arrow is pressed while a block is selected", (tester) async {
```
matthew-carroll, line 281: This test description is misleading. If I were to take this test description at face value, it sounds like when the caret is on the downstream edge of a block node, pressing the right arrow will never move the caret. But that's not true. If there's any content beyond that block node, then pressing the right arrow will move the caret to the next node.

Can you look carefully at the behavior in question and see if you can come up with a test description that's more specific to the situation that you're testing?

super_editor/test/src/default_editor/upstream_downstream_selection_test.dart
rutvik110: I'll check on it.

super_editor/test/src/default_editor/upstream_downstream_selection_test.dart
rutvik110: I see. Will have it updated to something more appropriate to the behaviour.

super_editor/test/src/default_editor/upstream_downstream_selection_test.dart
rutvik110: Not really. There's one group `move caret horizontally` but it deals with moving the caret only in collapsed selection cases. And our behaviour involves proper selection of the node so I created a new group that better suits it. 

The group name's a little ambiguous rn, will've to update it to something more appropriate.

super_editor/test/src/default_editor/upstream_downstream_selection_test.dart
rutvik110: Nothing specific. I saw it being used within this test suite so decided to use it.

super_editor/test/src/default_editor/upstream_downstream_selection_test.dart
rutvik110: Similar approach is used across all the tests within this test suite so I did the same for newly added test. I could switch it to what you're suggesting but just want to confirm once if it's necessary in this particular test suite.

rutvik110:
@matthew-carroll Updated the test based on the feedback. Let me know if need any further updates.

super_editor/test/src/default_editor/upstream_downstream_selection_test.dart
```diff
@@ -275,6 +277,60 @@ void main() {
...
+            .createDocument()
+            .withCustomContent(paragraphThenHrThenParagraphDoc())
+            .withEditorSize(const Size(300, 300))
+            .pump();
+
+        final document = SuperEditorInspector.findDocument()!;
+        final hrNode = document.getNodeById("2") as HorizontalRuleNode;
+
+        await tester.tapAtDocumentPosition(DocumentPosition(
+          nodeId: hrNode.id,
```
matthew-carroll, line 292: Using the `hrNode` reference isn't really necessary. Your code shows above that you already know the node ID, which is "2". And you know the position where you want to begin, which is something like UpstreamDownstreamDocumentPosition.beginning or something like that. So I think you can avoid access the document, or querying the node, which is generally preferable because we want to tests to rely as little on implementation details as possible.

In fact, to verify, I assume we're missing a test method to tap on a block component? I know we have a test method for tapping into a paragraph, but maybe we don't have one for blocks?

super_editor/test/src/default_editor/upstream_downstream_selection_test.dart
```diff
@@ -275,6 +277,60 @@ void main() {
...
+          nodePosition: hrNode.beginningPosition,
+        ));
+        await tester.pumpAndSettle();
```
matthew-carroll, line 295: You should only ever `pumpAndSettle()` when absolutely necessary, and when you do, you should almost always add a comment explaining why. Otherwise you should just `pump()` a single frame.

super_editor/test/src/default_editor/upstream_downstream_selection_test.dart
```diff
@@ -275,6 +277,60 @@ void main() {
...
+
+        final testerGesture = await tester.startDocumentDragFromPosition(
```
matthew-carroll, line 297: Why is this gesture necessary? You already tapped at the beginning of the block up above...

super_editor/test/src/default_editor/upstream_downstream_selection_test.dart
```diff
@@ -275,6 +277,60 @@ void main() {
...
+        await testerGesture.moveBy(const Offset(300, 0));
+        await tester.pump();
+
+        await tester.endDocumentDragGesture(testerGesture);
+        await tester.pump();
+
+        expect(
+          SuperEditorInspector.findDocumentSelection(),
+          DocumentSelection(
+            base: DocumentPosition(
```
matthew-carroll, line 314: This is an expanded selection - did the original ticket mention an expanded selection? I thought the original issue was about moving the caret?

super_editor/test/src/default_editor/upstream_downstream_selection_test.dart
rutvik110: That's right. A test method as well as an update on the test builder api to support building blocks would be helpful.

super_editor/test/src/default_editor/upstream_downstream_selection_test.dart
rutvik110: I'll update it with the comment on it's used there.

super_editor/test/src/default_editor/upstream_downstream_selection_test.dart
rutvik110: I tried a few other options but they weren't able to get the expected result. I'll check on this once more .

super_editor/test/src/default_editor/upstream_downstream_selection_test.dart
rutvik110: yes. To reproduce the issue shared in the issue #1247 requires to make a non-expanded selection on the block. I've made sure that the issue is in fact in that case only and isn't in any other scenario.

super_editor/test/src/default_editor/upstream_downstream_selection_test.dart
matthew-carroll: Add a separate test for what? Are you asking if you should add a new testing tool to select block components?

super_editor/test/src/default_editor/upstream_downstream_selection_test.dart
matthew-carroll: Why is it required?

super_editor/test/src/default_editor/upstream_downstream_selection_test.dart
matthew-carroll: This detail should certainly be captured in the test description. Probably something like "expanded selection around block node collapses to a caret on the downstream edge when the right arrow is pressed"

super_editor/test/src/default_editor/upstream_downstream_selection_test.dart
```diff
@@ -275,6 +277,58 @@ void main() {
...
+            .createDocument()
+            .withCustomContent(paragraphThenHrThenParagraphDoc())
+            .withEditorSize(const Size(300, 300))
+            .pump();
+
+        await tester.tapAtDocumentPosition(const DocumentPosition(
+          nodeId: "2",
+          nodePosition: UpstreamDownstreamNodePosition.upstream(),
+        ));
+        // Wait till the tap event is complete and the latest selection is available.
```
matthew-carroll, line 292: When I said that we add a comment on every `pumpAndSettle`, I meant that we comment *why* we need to settle. Why isn't a single `pump` sufficient? What's scheduling the extra work? My guess is that this really should be a `pump`. If we need to settle just to place the caret, that seems like a bug.

By the way, could we simplify this whole test by double tapping on the HR instead of tapping, placing the caret, and then dragging?

super_editor/test/src/default_editor/upstream_downstream_selection_test.dart
rutvik110: ah my bad. I assumed you were suggesting to add a new test for tap on a block. I'll check into the adding test method for that.

super_editor/test/src/default_editor/upstream_downstream_selection_test.dart
rutvik110: Seems like that. Atm the test fails when I just do `pump` with the selection returning null but doing `pumpAndSettle` resolves that. I haven't found the src of this issue but I'll see if it's indeed a bug or something else.

Sure! Will update it.

super_editor/test/src/default_editor/upstream_downstream_selection_test.dart
matthew-carroll: I don't think there's any need to add a test tool for selection at this point. You can just file an issue for that.

rutvik110:
@matthew-carroll Updated tests to use double-tap for selection and test description is updated to reflect the expected behaviour.

Let me know if any further updates are needed.

super_editor/test/src/default_editor/upstream_downstream_selection_test.dart
```diff
@@ -275,6 +277,46 @@ void main() {
...
+
+      testWidgets("right arrow collapses the expanded selection around block node to a caret on the downstream edge",
+          (tester) async {
+        await tester
+            .createDocument()
+            .withCustomContent(paragraphThenHrThenParagraphDoc())
+            .withEditorSize(const Size(300, 300))
+            .pump();
+
+        for (var i = 0; i < 2; i++) {
```
matthew-carroll, line 289: Please add a TODO that says to move this to Super Editor tooling.

Also, no need to loop just for two taps. Go ahead and just duplicate the code for two explicit taps. That will be more clear to the reader.

rutvik110: Updated.

matthew-carroll (approved):
LGTM

PR #1260, opened 2023-07-27
Moved IME upstream deletion in empty nodes to Editor requests instead of direct implementation, for the purpose of extensability (Resolves #1259)

### LINKED ISSUE #1259: [SuperEditor] - Try dispatching an EditRequest when the user presses backspace at beginning of node
IME editing results in a weird situation when the user presses BACKSPACE at the beginning of a node, or the beginning of content in general. When there's no content, BACKSPACE doesn't delete anything, and therefore no IME event is generated. Yet, applications often need to know when BACKSPACE is pressed at the beginning of empty content, so that the app can delete the paragraph, or the task, etc.

We've long used a hack where we insert an invisible "." at the beginning of any IME input so that we can detect when the user presses BACKSPACE at the beginning of a node. This behavior is currently entirely encapsulated within our IME delta processing system. Internalizing this behavior prevents apps from taking custom actions when the user presses BACKSPACE at the beginning of an empty node. For example, Superlist needs to convert a Task to a Paragraph. Also, when this happens at the beginning of a document, Superlist needs to move the selection to a different area of the UI. It's unclear how this can be accomplished with the current, private implementation.

For this ticket, within the IME delta system, instead of automatically taking an action when the user presses BACKSPACE in an empty node, try submitting an `EditRequest` to the `Editor`. This approach allows the Chain of Responsibility of `EditRequestHandler`s to decide what to do. `SuperEditor` should ship with sane defaults for deleting a paragraph, converting list items to paragraphs, etc.

matthew-carroll:
Moved IME upstream deletion in at the beginning of nodes to Editor requests instead of direct implementation, for the purpose of extensability (Resolves #1259)

### DIFF
--- a/super_editor/test/super_editor/text_entry/paragraph_conversions_test.dart
+++ b/super_editor/test/super_editor/text_entry/paragraph_conversions_test.dart
@@ -1,4 +1,6 @@
+import 'package:flutter/services.dart';
 import 'package:flutter_test/flutter_test.dart';
+import 'package:flutter_test_robots/flutter_test_robots.dart';
 import 'package:super_editor/super_editor.dart';
 import 'package:super_editor/super_editor_test.dart';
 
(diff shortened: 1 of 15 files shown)


matthew-carroll:
@miguelcmedeiros  let me know if this approach to deleting upstream at the beginning of nodes works for your needs. The PR includes 3 examples of request -> command pairs that respond to upstream deletions at the beginning of nodes. Those should give you an idea about how to add your own.

matthew-carroll:
@miguelcmedeiros @angelosilvestre this is the real PR for this change. It's ready for review.

miguelcmedeiros (approved):

super_editor/test/super_editor/text_entry/paragraph_conversions_test.dart
```diff
@@ -207,6 +209,118 @@ void main() {
...
+            .createDocument()
+            .fromMarkdown("# My Header")
+            .withInputSource(TextInputSource.ime)
+            .pump();
+        final headerNode = context.editContext.document.nodes.first;
+
+        await tester.placeCaretInParagraph(headerNode.id, 0);
+
+        // Ensure that we're starting with a header.
+        headerNode.metadata["blockType"] = header1Attribution;
```
angelosilvestre, line 225: Shouldn't this be an `expect()` call ?

super_editor/test/super_editor/text_entry/paragraph_conversions_test.dart
```diff
@@ -207,6 +209,118 @@ void main() {
...
+              oldText: ". My Blockquote",
+              selection: TextSelection.collapsed(offset: 1),
+              deletedRange: TextRange(start: 1, end: 2),
+              composing: TextRange.empty,
+            ),
+          ],
+          getter: imeClientGetter,
+        );
+
+        // Ensure that the header became a paragraph.
```
angelosilvestre, line 281: header -> blockquote

angelosilvestre (approved):
LGTM with a few comments.

PR #1264, opened 2023-07-27
Migrates from our custom ListenableBuilder to Flutter's new ListenableBuilder (Resolves #1132)

### LINKED ISSUE #1132: Migrate from our custom ListenableBuilder to Flutter's new ListenableBuilder
We defined a widget called `ListenableBuilder` so that we could rebuild when a `Listenable` changed. This could always be done with an `AnimatedBuilder`, but that name was confusing, because we weren't watching animations.

Flutter recently made a similar change in the framework. As a result, there's a naming conflict in `super_editor`. Up to this point, we've explicitly hidden Flutter's own `ListenableBuilder` in favor of our own. 

For this ticket, get rid of our implementation of `ListenableBuilder` in favor of Flutter's implementation.

rutvik110:
Removed our custom `ListenableBuilder` in favour of newly introduced `ListenableBuilder` in Flutter. (Resolves https://github.com/superlistapp/super_editor/issues/1132)


matthew-carroll:
@rutvik110 in the future, please use a branch name like `cherry-pick-to-stable` so that we don't confuse cherry pick PRs with the real, original work.

PR #1266, opened 2023-07-27
Moved IME upstream deletion in empty nodes to Editor requests instead of direct implementation, for the purpose of extensability (Resolves #1259)

### LINKED ISSUE #1259: [SuperEditor] - Try dispatching an EditRequest when the user presses backspace at beginning of node

matthew-carroll:
Moved IME upstream deletion in empty nodes to Editor requests instead of direct implementation, for the purpose of extensability (Resolves #1259)

This is a temporary PR so that Superlist can try this behavior on Flutter `stable`.

### DIFF
--- a/super_editor/example/lib/demos/demo_empty_document.dart
+++ b/super_editor/example/lib/demos/demo_empty_document.dart
@@ -40,7 +40,6 @@ class _EmptyDocumentDemoState extends State<EmptyDocumentDemo> {
         document: _doc,
         composer: _composer,
         gestureMode: DocumentGestureMode.mouse,
-        inputSource: TextInputSource.keyboard,
       ),
     );
   }
--- a/super_editor/example/lib/main.dart
+++ b/super_editor/example/lib/main.dart
@@ -40,6 +40,7 @@ Future<void> main() async {
     // editorScrollingLog,
     // editorGesturesLog,
     // editorImeLog,
+    // editorImeDeltasLog,
     // editorKeyLog,
     // editorOpsLog,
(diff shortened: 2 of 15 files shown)


miguelcmedeiros (approved):

matthew-carroll:
I forgot this was the stable branch PR. I'm going to move the review to the master branch PR.

PR #1267, opened 2023-07-28
[SuperEditor][SuperTextLayout] Fix local golden runner (Resolves #1265)

### LINKED ISSUE #1265: Fix local golden runner
We have a number of issues with our local golden runner that we need to resolve:

- The runner caches the Flutter version - it needs to always use the latest `master`
- The runner doesn't output any information - it should send to stdout and stderr
- When running the golden tool, it does something to local project cache that causes the IDE to show syntax errors everywhere. Running another `pub get` locally fixes the issue, but it's annoying.
- The runner, at least on Mac, produces different output than GitHub. The purpose of the runner is to be able to generate the same output from any platform, so this is a critical issue. I've observed that when I run the tool, I get thicker text than what's currently in the goldens.
- We haven't thought previously about which platform the golden tests target. Flutter and Super Editor make various decisions based on the target platform. At the moment we're running `testGoldens` without any reference to the platform, so we're using whatever the Flutter test runner chooses as the default. We probably need to introduce golden versions of our platform tests, like `testGoldensOnMac`, `testGoldensOnLinux`, `testGoldensOnAndroid`, etc, so that we can get the golden UI we expect for a given test.

angelosilvestre:
[SuperEditor][SuperTextLayout] Fix local golden runner. Resolves #1265

We have some issues when running the golden tests in a container:

- The runner caches the Flutter version.
- The runner doesn't output any information until the process ends.
- It messes with the project cache in the host machine.
- It doesn't specify the target platform
- It's generating different output on Mac.

This PR changes the following:

- Invalidates the Flutter version when a new commit is pushed to Flutter master.
- Streams the process output to stdout.
- Copies the whole repo to the image.
- Adds custom golden test methods to specify the target platform.

**Remarks:**

1. As we don't map the whole repository inside the container anymore, we need to map the `test_goldens/failures` directories, so when a failure happens the image will be written to the host machine. The tool will list all of the subdirectories of the test folder and will map a `test_dir/failures`.

1. I needed to move the dockerfile to the repo root because it's the only way we can copy the whole repo. If the dockerfile is located in `super_editor/super_editor` it isn't possible to copy `super_editor/super_text_layout` and the other directories: 

```
 COPY obeys the following rules:

The <src> path must be inside the context of the build; you cannot COPY ../something /something, because the first step of a docker build is to send the context directory (and subdirectories) to the docker daemon.
```

3. The Flutter docs mention that in tests, the default target platform is always Android:

```dart
/// In a test environment, the platform returned is [TargetPlatform.android]
/// regardless of the host platform. (Android was chosen because the tests were
/// originally written assuming Android-like behavior, and we added platform
/// adaptations for iOS later). Tests can check iOS behavior by using the
/// platform override APIs (such as [ThemeData.platform] in the material
/// library) or by setting [debugDefaultTargetPlatformOverride].
```

Because of that, I changed all the existing golden tests to use the new  `testGoldensOnAndroid` method, to explicitly reflect what is the target platform.

4. I'm not sure yet why it is generating a different output on Mac. I think it might be possible that it is related to mapping the whole repository. We need to test these changes on mac to see if it will solve this issue.


matthew-carroll:
Quick note as I'm working on this PR: we need to support a file path specification in our runner. Currently, apparently, we only support specifying a test directory. We should be able to run:

 * No specified directory
 * Top-level test directory: `test_goldens`
 * Test sub-directory: `test_goldens/super_editor`
 * Test file: `test_goldens/super_editor/super_editor_text_layout.dart`

Ideally, we should support all the various ways to specify tests that `flutter test` supports.

matthew-carroll:
I'm noticing that when I run the Docker golden tests, `failures` directories are created even when there aren't any failures for the corresponding tests.

matthew-carroll:
I did work on this branch, but when I pushed, GitHub created another branch: https://github.com/superlistapp/super_editor/pull/1268

@angelosilvestre do you wanna pull those changes into this PR?

angelosilvestre:
@matthew-carroll Done.

golden_tester.Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+RUN git clone https://github.com/flutter/flutter.git ${FLUTTER_HOME}
+
+RUN flutter doctor
+
+# Copy the whole repo.
+# We need this because we use local dependencies.
+COPY ./ /super_editor
+
+# Print the Ubuntu version. Useful when there are failing tests.
+RUN cat /etc/lsb-release
```
matthew-carroll, line 24: Can we move this to the top, before we start doing a bunch of Flutter stuff?

super_editor/test_goldens/editor/components/_components_test_utils.dart
```diff
@@ -1,8 +1,10 @@
 import 'package:flutter/material.dart';
 import 'package:golden_toolkit/golden_toolkit.dart';
 
+import '../../../test/test_tools.dart';
+
 void testComponentGolden(String description, Widget componentBuilder, String fileName) {
-  testGoldens(description, (tester) async {
+  testGoldensOnAndroid(description, (tester) async {
```
matthew-carroll, line 7: Can you file an issue to come back and audit the golden tests to select the appropriate platforms for each test? This change preserves the existing goldens, but in reality, we should probably have goldens for multiple platforms, and some tests might only apply to mobile or only apply to desktop.

super_editor/test/test_tools.dart
```diff
@@ -325,6 +326,96 @@ void testWidgetsOnIos(
   }, skip: skip, variant: variant);
 }
 
+/// A golden test that configures itself as a Android platform before executing the
+/// given [test], and nullifies the Android configuration when the test is done.
+@isTest
+void testGoldensOnAndroid(
```
matthew-carroll, line 332: Let's put these tools in `test_tools_goldens.dart` so that the tests don't need to cross top-level testing directories to get to the tools.

super_editor/tool/goldens.dart
```diff
@@ -1,229 +1,301 @@
...
+
+/// A [Command] which runs golden tests.
+///
+/// Usage: `flutter pub run super_editor:goldens test <options> <directory>`
+///
+/// Options:
+///
+/// `--plain-name "test-name"`: Runs only the tests containing the given value in the test name.
+///
+/// The directory is optional.
```
matthew-carroll, line 28: The "directory" isn't necessarily a directory, is it?

super_editor/tool/goldens.dart
```diff
@@ -1,229 +1,301 @@
...
+
+    // Other arguments passed at the end of the command.
+    // For example, the test directory.
+    final rest = [...args.rest];
+
+    late String testDirOrTestFileName;
+    late String testBaseDirectory;
+
+    if (rest.isNotEmpty) {
+      // An argument was passed after the command options. For example, tool/goldens test my_test_dir.
```
matthew-carroll, line 72: It's unclear in this comment what "tool/goldens test my_test_dir" means. Is that meant to be read as "tool/goldens", "test", "my_test_dir"? Or is that supposed to be one command? And what's the "/" mean?

super_editor/tool/goldens.dart
```diff
@@ -1,229 +1,301 @@
...
+      // Use the first argument after the command options as the test directory and remove it from the rest.
```
matthew-carroll, line 73: It's not necessarily a "directory", right?

super_editor/tool/goldens.dart
```diff
@@ -1,229 +1,301 @@
...
+      testDirOrTestFileName = rest.removeAt(0);
+
+      if (testDirOrTestFileName.endsWith('.dart')) {
```
matthew-carroll, line 76: Why do we need to break this apart and inspect it? The Flutter test tool should accept whatever the argument happens to be, right?

If you really need to inspect directories and file names and extensions, consider using the `path` tools, which were made for that.

super_editor/tool/goldens.dart
```diff
@@ -1,229 +1,301 @@
...
+    // Remove all the empty failure directories.
+    for (final dirName in dirs) {
+      final dir = Directory('$dirName/failures');
+      if (dir.existsSync() && dir.listSync().isEmpty) {
+        dir.deleteSync();
+      }
+    }
+  }
+
+  /// Returns a list with all volume mappings for the test failure directories.
```
matthew-carroll, line 125: It returns a `List<String>` so it's unclear what a "volume mapping" is. Can you clarify what this method returns?

Also, given that `testDirs` is also a `List<String>`, please specify the expects format of those items. Are they absolute file system paths? Are they relative paths? Are they just names of directories without a location?

super_editor/tool/goldens.dart
```diff
@@ -1,229 +1,301 @@
...
+  ///
+  /// This mappings are used so when a failure happens, the failure images are save in the host OS.
+  List<String> _generateFailureDirMappings(List<String> testDirs) {
+    final mapppings = <String>[];
```
matthew-carroll, line 129: "mapppings" -> "mappings"

super_editor/tool/goldens.dart
```diff
@@ -1,229 +1,301 @@
...
+      if (dir.existsSync() && dir.listSync().isEmpty) {
+        dir.deleteSync();
+      }
+    }
+  }
+
+  /// Returns a list with all volume mappings for the test failure directories.
+  ///
+  /// This mappings are used so when a failure happens, the failure images are save in the host OS.
+  List<String> _generateFailureDirMappings(List<String> testDirs) {
```
matthew-carroll, line 128: Let's try to avoid shortening words when it's not too bad. So let's say directory and directories instead of dir and dirs.

super_editor/tool/goldens.dart
```diff
@@ -1,229 +1,301 @@
...
+
+    for (final dir in testDirs) {
+      mapppings.add('-v');
+      mapppings.add('${Directory.current.path}/$dir/failures:/super_editor/super_editor/$dir/failures');
+    }
+
+    return mapppings;
+  }
+
+  /// Returns all sub-directories inside a root test directory.
```
matthew-carroll, line 139: Please provide an example of what to expect as a return value. Are they absolute paths or relative paths? If they're relative, they're relative to what?

super_editor/tool/goldens.dart
```diff
@@ -1,229 +1,301 @@
...
+  ///
+  /// Ignores "goldens" and "failures" directories.
+  List<String> _findAllTestDirs(String rootTestDir) {
+    final dir = Directory(rootTestDir);
+    final subDirs = dir
+        .listSync(recursive: true) //
+        .whereType<Directory>()
+        // Ensure we use linux path separator.
```
matthew-carroll, line 147: There's probably no reason to assume Linux. In general, when working with paths in Dart, you should use `Platform.pathSeparator`

super_editor/tool/goldens.dart
```diff
@@ -1,229 +1,301 @@
...
+        .map((e) => e.path.replaceAll(path.separator, '/'))
+        .where((e) => !e.endsWith('goldens') && !e.endsWith('failures'))
```
matthew-carroll, line 149: Why are we avoiding directories named "goldens"? That seems brittle.

super_editor/tool/goldens.dart
```diff
@@ -1,229 +1,301 @@
...
+  );
+}
+
+/// Runs [exe] with the given [arguments].
+///
+/// The child process stdout and stderr are written to the current process stdout.
+///
+/// Throws and exception if the process exists with a non-zero exit code.
+Future<void> _runProcess({
+  required String exe,
```
matthew-carroll, line 282: Let's call this `executable`

super_editor/tool/goldens.dart
```diff
@@ -1,229 +1,301 @@
...
+    ],
+    // We need to use the repository root as the working directory to be able to copy all of the files
+    // in this repository, not just the super_editor sub-directory.
+    workingDirectory: '../',
+    description: 'Image build',
+  );
+}
+
+/// Runs [exe] with the given [arguments].
+///
```
matthew-carroll, line 277: This Dart Doc should include a specification for the expected format of the various `String` arguments. For example, is `executable` an aboslute file path? a relative file path? What should it look like?

Should the `arguments` include modifiers like `-` and `--`, such as `--platforms`?

What does the `description` do?

What impact does the `workingDirectory` have?

super_text_layout/test_goldens/test_tools.dart
```diff
@@ -8,6 +11,96 @@ import 'package:super_text_layout/super_text_layout.dart';
 
 const defaultSelectionColor = Color(0xFFACCEF7);
 
+/// A golden test that configures itself as a Android platform before executing the
+/// given [test], and nullifies the Android configuration when the test is done.
+@isTest
+void testGoldensOnAndroid(
```
matthew-carroll, line 17: We shouldn't duplicate the definition of all of these test tools

super_text_layout/tool/goldens.dart
```diff
@@ -61,38 +63,92 @@ class GoldenTestCommand extends Command {
 
     // Other arguments passed at the end of the command.
     // For example, the test directory.
-    final rest = args.rest;
+    final rest = [...args.rest];
```
matthew-carroll, line 66: Is this all duplicated from the other runner? If so, we really shouldn't duplicate the same tool. What can we do to de-duplicate it?

super_editor/tool/goldens.dart
angelosilvestre: We need to inspect it, because we need to list the sub-directories to map the failure directories.

Updated.

super_editor/tool/goldens.dart
angelosilvestre: We need to assume Linux because it's the platform of the container. When running the tool on windows, the path separator is "\". If we don't replace it, the "flutter test" process won't find the files/directories.

super_editor/tool/goldens.dart
angelosilvestre: We are ignoring because I'm assuming directories name "goldens" will contain only the golden files, not tests.

super_editor/test_goldens/editor/components/_components_test_utils.dart
angelosilvestre: Will do.

super_text_layout/test_goldens/test_tools.dart
angelosilvestre: To avoid duplication, I think we would need to create a top-level package where we could add these test tools.

super_text_layout/tool/goldens.dart
angelosilvestre: Same as above.

super_text_layout/test_goldens/test_tools.dart
matthew-carroll: What would happen if we created another package in the monorepo called golden_runner? Can we neatly package all this stuff in there and still make use of it?

super_editor/tool/goldens.dart
matthew-carroll: Is there any particular reason that we need to ignore it? Even if it's a directory filled with golden images, does that cause a problem for us?

Other than this code here, I don't think anything prevents developers from writing tests inside of "goldens" directories, even if it's a weird choice to make.

super_editor/tool/goldens.dart
matthew-carroll: What do you mean the path separator is ""?

super_text_layout/test_goldens/test_tools.dart
angelosilvestre: I think so, I'll give it a shot.

super_editor/tool/goldens.dart
angelosilvestre: Sorry, it should be: on windows, the path separator is `\`.

matthew-carroll: Ok, so the answer here is that we can't use `Platform.pathSeparator` because this code isn't running inside the Docker container, it's running outside on the host system, which will have a different separator. Therefore, we have to include the knowledge that the Docker container is running on Linux.

super_editor/tool/goldens.dart
angelosilvestre: No particular reason. Updated.

angelosilvestre:
@matthew-carroll Created a `golden_runner` package to dedup the golden commands.

Unfortunately, I wasn't able yet to move the custom test methods to this package. When I try to run the tool using the test methods from the package,  I get the following compilation error;

```
/c:/git/flutter/packages/flutter/lib/src/foundation/binding.dart:9:8: Error: Dart library 'dart:ui' is not available on this platform.
import 'dart:ui' as ui show Brightness, PlatformDispatcher, SingletonFlutterWindow, window; 
Context: The unavailable library 'dart:ui' is imported through these packages:
```

I tried with the `flutter_test` dependency in `dev_dependencies` and also tried with `dependencies`.

matthew-carroll:
> When I try to run the tool using the test methods from the package

Can you be more specific? What do you mean by "using the test methods from the package"?

Have you tried writing a few tests within the new package as a sanity check, to make sure tests can run there, before trying to use the package to run tests elsewhere?

Also, the error message you showed looks like it's about to list some packages. Is there a list of packages that weren't included in that error message?

angelosilvestre:
@matthew-carroll 

> Can you be more specific? What do you mean by "using the test methods from the package"?

Using the `testGoldensOnAndroid` method declared in the package.

> Have you tried writing a few tests within the new package as a sanity check, to make sure tests can run there, before trying to use the package to run tests elsewhere?

Yeah, using `flutter test test_goldens` with tests in the `golden_runner` package works as expected.

> Also, the error message you showed looks like it's about to list some packages. Is there a list of packages that weren't included in that error message?

I get errors lik `Error: 'Offset' isn't a type` for pretty much every Flutter class. It seems that the tool cannot access Flutter classes.

matthew-carroll (approved):
LGTM

PR #1270, opened 2023-08-01
[SuperEditor] Add chinese input test (Resolves #1230)

### LINKED ISSUE #1230: [SuperEditor][IOS]The input text disappeared.
SuperEditor Version: 0.2.6 

Test video using the main branch code for demonstration.

1. Choose an iOS device.
2. Use Chinese full keyboard input method.
3. Edit at the beginning of an empty document.
4. Continuously input 'aaaa'.
5. **Click the '换行(Enter)' or '确认(Done)' button.**
6. Repeat steps 4 and 5.

The second scenario of disappearing.

1. Choose an iOS device.
2. Use Chinese full keyboard input method.
3. Edit at the beginning of an empty document.
4. Continuously input 'aaaa'.
5. **Click on the suggested text, and it disappears from the screen after being output.**
6. Repeat steps 4 and 5.

https://github.com/superlistapp/super_editor/assets/37604230/d984c2d3-26ef-48c1-8e91-8e1bfb138f76

angelosilvestre:
[SuperEditor] Add chinese input test. Resolves #1230

Before #1245 was merged, when the user uses a chinese keyboard, typing a characterd after accepting a suggestion causes the whole text to be replaced by the last character typed.

Looking at the logs, the following is happening:

1. We receive the replacement delta
2. We send [0,3) as the composing region back to the IME.
3. We receive [-1,-1] as the new composing region (in the same frame)

It seems that, after step 3, the IME changes the composing region to the value we sent in step 2. As we had code to ignore non-text deltas in the same frame as we send the editing value to the IME, we don't clear the composing region after step 3.

Because of that, the whole text becomes part of the composing region, which is why the whole text is replaced.

Relevant logs: 

```console
flutter: (14.982) editor.ime > INFO: Applying delta: TextEditingDeltaReplacement#70159(oldText: a a a, textReplaced: a a a, replacementText: 呵呵呵, replacedRange: TextRange(start: 0, end: 5), selection: TextSelection.collapsed(offset: 3, affinity: TextAffinity.upstream, isDirectional: false), composing: TextRange(start: 0, end: 3))

flutter: (15.015) editor.ime > FINE: Sending forceful update to IME because our local TextEditingValue didn't change, but the IME may have:
flutter: (15.015) editor.ime > FINE: TextEditingValue(text: ┤呵呵呵├, selection: TextSelection.collapsed(offset: 3, affinity: TextAffinity.downstream, isDirectional: false), composing: TextRange(start: 0, end: 3))

flutter: (15.017) editor.ime > FINE: TextEditingDeltaNonTextUpdate#e0ae5(oldText: 呵呵呵, selection: TextSelection.collapsed(offset: 3, affinity: TextAffinity.upstream, isDirectional: false), composing: TextRange(start: -1, end: -1))
flutter: (15.017) editor.ime > FINE: Ignoring this delta because it's a non-text update that came in right after we setEditingState()

flutter: (16.431) editor.ime > FINE: Received edit deltas from platform: 1 deltas
flutter: (16.435) editor.ime > FINE: TextEditingDeltaReplacement#9043f(oldText: 呵呵呵, textReplaced: 呵呵呵, replacementText: d, replacedRange: TextRange(start: 0, end: 3), selection: TextSelection.collapsed(offset: 1, affinity: TextAffinity.upstream, isDirectional: false), composing: TextRange(start: 0, end: 1))
```

This PR adds a test to ensure we are sending the composing region back to the IME after we receive the non-text delta in the same frame. I'm not sure this test is valuable enough though...

### DIFF
--- a/super_editor/test/src/default_editor/document_input_ime_test.dart
+++ b/super_editor/test/src/default_editor/document_input_ime_test.dart
@@ -685,15 +685,20 @@ Paragraph two
 
         // Simulate the user typing "a a a".
         await tester.ime.sendDeltas(const [
+          TextEditingDeltaNonTextUpdate(
+            oldText: '. ',
+            selection: TextSelection.collapsed(offset: 2),
+            composing: TextRange(start: -1, end: -1),
+          ),
           TextEditingDeltaInsertion(
-            oldText: '',
+            oldText: '. ',
             textInserted: 'a',
-            insertionOffset: 0,
+            insertionOffset: 2,
             selection: TextSelection.collapsed(
-              offset: 1,
+              offset: 3,
               affinity: TextAffinity.upstream,
             ),
-            composing: TextRange(start: 0, end: 1),
+            composing: TextRange(start: 2, end: 3),
           ),
         ], getter: imeClientGetter);
         await tester.ime.sendDeltas(const [
--- a/super_editor/test/src/default_editor/document_input_ime_test.dart
+++ b/super_editor/test/src/default_editor/document_input_ime_test.dart
@@ -669,6 +669,116 @@ Paragraph two
       });
     });
 
+    group('on iPhone 11 (iOS 13.7) with chinese keyboard', () {
+      testWidgetsOnIos('applies keyboard suggestions', (tester) async {
+        // Holds the composing region that we sent to the IME.
+        TextRange? composingRegion;
+
+        await tester //
+            .createDocument()
+            .withSingleEmptyParagraph()
+            .withInputSource(TextInputSource.ime)
+            .pump();
+
+        // Place the caret at the start of the paragraph.
+        await tester.placeCaretInParagraph('1', 0);
+
+        // Simulate the user typing "a a a".
+        await tester.ime.sendDeltas(const [
+          TextEditingDeltaInsertion(
+            oldText: '',
+            textInserted: 'a',
+            insertionOffset: 0,
+            selection: TextSelection.collapsed(
+              offset: 1,
+              affinity: TextAffinity.upstream,
+            ),
+            composing: TextRange(start: 0, end: 1),
+          ),
+        ], getter: imeClientGetter);
+        await tester.ime.sendDeltas(const [
+          TextEditingDeltaInsertion(
+            oldText: 'a',
+            textInserted: ' a',
+            insertionOffset: 1,
+            selection: TextSelection.collapsed(
+              offset: 3,
+              affinity: TextAffinity.upstream,
+            ),
+            composing: TextRange(start: 0, end: 3),
+          ),
+        ], getter: imeClientGetter);
+        await tester.ime.sendDeltas(const [
+          TextEditingDeltaInsertion(
+            oldText: 'a a',
+            textInserted: ' a',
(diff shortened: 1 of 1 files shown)


super_editor/test/src/default_editor/document_input_ime_test.dart
```diff
@@ -669,6 +669,116 @@ Paragraph two
...
+            replacedRange: TextRange(start: 0, end: 5),
+            selection: TextSelection.collapsed(
+              offset: 3,
+              affinity: TextAffinity.upstream,
+            ),
+            composing: TextRange(start: 0, end: 3),
+          ),
+        ]);
+
+        // Intercept the setEditingState message sent to the platform.
```
matthew-carroll, line 742: Please specify *why* we're intercepting it. The most important detail of anything we do is the intention.

super_editor/test/src/default_editor/document_input_ime_test.dart
```diff
@@ -669,6 +669,116 @@ Paragraph two
...
+            composingRegion = TextRange(
+              start: params['composingBase'],
+              end: params['composingExtent'],
+            );
+            return null;
+          },
+        );
+
+        // After receiving the previous delta and before receiving the next delta,
+        // we send [0, 3) as the composing region.
```
matthew-carroll, line 758: Where do we send it? Are we sending it with the code below this comment? If not, this comment is in the wrong place. In general, code comments shouldn't refer to things across time, because each line of code runs at a moment in time. We want to comment what we're doing, just before we do it, and explain why we're doing it.

Comments about multi-step sequences belong above entire methods, or classes, or at the beginning of a test.

super_editor/test/src/default_editor/document_input_ime_test.dart
```diff
@@ -669,6 +669,116 @@ Paragraph two
...
+        //
+        // The whole text is part of the composing region again. If the composing region stays like this,
```
matthew-carroll, line 760: With these two following comments, we have another problem with very confusing references to time and place. I can't tell which pieces of these descriptions refer to things that have already happened vs are going to happen. I can't tell which pieces refer to what we want to happen vs what is actually happening. We need to be crystal clear along those dimensions.

matthew-carroll:
@angelosilvestre I wanted to make sure you saw my comments in this PR

super_editor/test/src/default_editor/document_input_ime_test.dart
angelosilvestre: Updated

angelosilvestre:
@matthew-carroll I tried to make the last comment more clear. Please let me know if you think it needs more changes.

matthew-carroll:
@angelosilvestre it looks like this test is now failing

angelosilvestre:
@matthew-carroll Solved.

matthew-carroll:
@angelosilvestre what was the issue?

angelosilvestre:
@matthew-carroll I needed to add the placeholder `. ` at the beginning of the first delta.

This change was already made to an existing test.

matthew-carroll (approved):
LGTM

Issue #1253, opened 2023-07-25
SuperEditor refreshes at 60FPS, even though cursor only changes every 500ms.


knopp:
https://github.com/superlistapp/super_editor/issues/1236 was closed but one issue still remains:

A while ago https://github.com/superlistapp/super_editor/pull/285 got merged to fix blinking cursor performance, but later it was replaced by [BlinkController](https://github.com/superlistapp/super_editor/blob/main/super_text_layout/lib/src/infrastructure/blink_controller.dart), which uses a `Ticker`. Unlike custom timer `Ticker` on desktop runs at constant 60 fps (each tick requesting new frame), causing unnecessary repaints given that cursor blinking interval is much longer.

matthew-carroll:
@knopp do you have any ideas for how to resolve this? The reason we use `Ticker`s throughout Super Editor, and not `Timer`s is because `Timer`s are a Dart construct that can't work in Flutter tests. `Ticker`s on the other hand are completely controlled by Flutter's test runner, which allows us to make dependable assertions related to their behavior in tests.

Is there a way to continue using `Ticker`s without forcing 60FPS?

knopp:
I think that what ticker does is that it schedules a frame immediately after current. I haven't found any way to change the behaviour. 

As for the timer, wouldn't something like [fake_async](https://pub.dev/packages/fake_async) help with this?

matthew-carroll:
Maybe that package would help, but it looks like it requires us to retrofit pretty much every test in our project. It looks like it's attempting to replicate Flutter's existing time control but for non-Flutter classes, like `Timer`.

I think I'll file a Flutter issue and see what they say.

matthew-carroll:
I filed this: https://github.com/flutter/flutter/issues/131357

PR #1281, opened 2023-08-09
Temporarily offer a Timer-based blink controller to avoid excessive FPS (Bandaid #1253)

matthew-carroll:
Temporarily offer a Timer-based blink controller to avoid excessive FPS (Bandaid #1253)

Our hope is that the Flutter team solves https://github.com/flutter/flutter/issues/131357 to give us a singular approach that works in production and in tests.


brian-superlist (approved):
Looks like a reasonable bandaid to improve battery performance while avoiding refactoring thousands of tests.

super_text_layout/lib/src/infrastructure/blink_controller.dart
```diff
@@ -20,19 +22,28 @@ class BlinkController with ChangeNotifier {
     _ticker = tickerProvider.createTicker(_onTick);
   }
 
+  BlinkController.withTimer({
+    Duration flashPeriod = const Duration(milliseconds: 500),
```
brian-superlist, line 26: Not an important improvement to make here, but should this `flashPeriod` duration be configurable by the options passed to `DefaultCaretOverlayBuilder` as well?

matthew-carroll: Probably. If anyone ever asks for it, we can add it. Although, so far, no one has.

matthew-carroll:
@knopp would you like to review this before I merge, to ensure it solves your FPS problem?

knopp (approved):
Thanks!

LGTM!

matthew-carroll:
Note: Apparently I didn't expose the necessary Timer controls for a SuperTextField. Add that API and then merge in.

angelosilvestre (commented):
LGTM

PR #1284, opened 2023-08-10
[SuperEditor] Fix task deletion (Resolves #1283)

### LINKED ISSUE #1283: [SuperEditor] In the latest main branch, deleting a Task node causes an exception
In the latest main branch, deleting a Task node causes an exception, and then the entire editor becomes unresponsive.

SuperEditor: main

flutter doctor:
[✓] Flutter (Channel master, 3.13.0-14.0.pre.5, on macOS 13.4.1 22F770820d darwin-arm64, locale zh-Hans-CN)
[✓] Android toolchain - develop for Android devices (Android SDK version 33.0.0)
[✓] Xcode - develop for iOS and macOS (Xcode 14.3.1)
[✓] Chrome - develop for the web
[✓] Android Studio (version 2022.3)
[✓] IntelliJ IDEA Ultimate Edition (version 2022.3.3)
[✓] VS Code (version 1.81.0)

 Video:

https://github.com/superlistapp/super_editor/assets/37604230/c5d3717c-1585-4ed9-9253-d6067f448445

Error:
```
======== Exception caught by services library ======================================================
Exception: Could not handle EditorRequest. DocumentEditor doesn't have a handler that recognizes the request: Instance of 'DeleteUpstreamAtBeginningOfNodeRequest'

#0      Editor._findCommandForRequest (package:super_editor/src/core/editor.dart:189:5)
#1      Editor.execute (package:super_editor/src/core/editor.dart:147:23)
#2      TextDeltasDocumentEditor.delete (package:super_editor/src/default_editor/document_ime/document_delta_editing.dart:370:14)
#3      TextDeltasDocumentEditor._applyDeletion (package:super_editor/src/default_editor/document_ime/document_delta_editing.dart:219:5)
#4      TextDeltasDocumentEditor.applyDeltas (package:super_editor/src/default_editor/document_ime/document_delta_editing.dart:83:9)
#5      DocumentImeInputClient.updateEditingValueWithDeltas (package:super_editor/src/default_editor/document_ime/document_ime_communication.dart:196:30)
#6      DeltaTextInputClientDecorator.updateEditingValueWithDeltas (package:super_editor/src/default_editor/document_ime/ime_decoration.dart:129:14)
#7      TextInput._handleTextInputInvocation (package:flutter/src/services/text_input.dart:1864:63)
#8      TextInput._loudlyHandleTextInputInvocation (package:flutter/src/services/text_input.dart:1753:20)
#9      MethodChannel._handleAsMethodCall (package:flutter/src/services/platform_channel.dart:552:55)
#10     MethodChannel.setMethodCallHandler.<anonymous closure> (package:flutter/src/services/platform_channel.dart:545:34)
#11     _DefaultBinaryMessenger.setMessageHandler.<anonymous closure> (package:flutter/src/services/binding.dart:567:35)
#12     _invoke2 (dart:ui/hooks.dart:202:13)
#13     _ChannelCallbackRecord.invoke (dart:ui/channel_buffers.dart:45:5)
#14     _Channel.push (dart:ui/channel_buffers.dart:135:31)
#15     ChannelBuffers.push (dart:ui/channel_buffers.dart:343:17)
#16     PlatformDispatcher._dispatchPlatformMessage (dart:ui/platform_dispatcher.dart:745:22)
#17     _dispatchPlatformMessage (dart:ui/hooks.dart:114:31)
call: MethodCall(TextInputClient.updateEditingStateWithDeltas, [2, {deltas: [{selectionBase: 1, oldText: . Cmmatic control over the user's selection and styles, create a DocumentComposer., selectionAffinity: TextAffinity.downstream, deltaEnd: -1, deltaText: , deltaStart: -1, composingExtent: -1, selectionIsDirectional: false, selectionExtent: 2, composingBase: -1}, {selectionBase: 1, oldText: . Cmmatic control over the user's selection and styles, create a DocumentComposer., selectionAffinity: TextAffinity.downstream, deltaEnd: 2, deltaText: , deltaStart: 1, composingExtent: -1, selectionIsDirectional: false, selectionExtent: 1, composingBase: -1}]}])
====================================================================================================
```

angelosilvestre:
[SuperEditor] Fix task deletion. Resolves #1283 

Pressing backspace at the beginning of a task causes the editor to crash with the exception:

```console
Exception: Could not handle EditorRequest. DocumentEditor doesn't have a handler that recognizes the request: Instance of 'DeleteUpstreamAtBeginningOfNodeRequest'
```

The cause is that we don't have a handler registered to handle the `DeleteUpstreamAtBeginningOfNodeRequest` with a `TaskNode`.

This PR duplicates `DeleteUpstreamAtBeginningOfParagraphCommand` and adapts it to work with the `TaskNode`.

It would be  probably better if we dedup the common behavior.


matthew-carroll:
> It would be probably better if we dedup the common behavior.

Then let's do that.

angelosilvestre:
@matthew-carroll Deduped `mergeTextNodeWithUpstreamTextNode` and `mergeTextNodeWithUpstreamTextNode`.

super_editor/lib/src/default_editor/paragraph.dart
```diff
@@ -492,83 +492,83 @@ class DeleteUpstreamAtBeginningOfParagraphCommand implements EditCommand {
...
+  if (node == null) {
+    return;
+  }
+
+  final nodeBefore = document.getNodeBefore(node);
+  if (nodeBefore == null) {
+    return;
+  }
+
+  executor.executeCommand(
```
matthew-carroll, line 560: I thought the original PR comment was about de-duping more than just this command.

But this command really isn't about paragraphs, which means it's not in the right file. Also, configuring a single command to place the caret is a reasonable thing for each operation to choose on its own, without centralizing this selection configuration. 

Do you know why `moveSelectionToEndOfPrecedingNode` was publicly declared in this file to begin with? Is it being used outside this fiel?

super_editor/lib/src/default_editor/tasks.dart
```diff
@@ -487,3 +487,57 @@ class SplitExistingTaskCommand implements EditCommand {
...
+    final document = context.find<MutableDocument>(Editor.documentKey);
+    final composer = context.find<MutableDocumentComposer>(Editor.composerKey);
+    final documentLayoutEditable = context.find<DocumentLayoutEditable>(Editor.layoutKey);
+
+    final nodeBefore = document.getNodeBefore(node);
+    if (nodeBefore == null) {
+      return;
+    }
+
+    if (nodeBefore is TextNode) {
```
matthew-carroll, line 516: Shouldn't we be turning a `TaskNode` into a regular `ParagraphNode` when the user presses backspace at the beginning?

super_editor/test/super_editor/components/task_test.dart
```diff
@@ -369,5 +371,97 @@ void main() {
         ),
       );
     });
+
+    testWidgetsOnAllPlatforms("combines with previous paragraph when the user presses BACKSPACE at the beginning",
```
matthew-carroll, line 375: What about merging into a task above? Is that already covered by a test in here?

angelosilvestre: Added a test for that.

super_editor/lib/src/default_editor/tasks.dart
angelosilvestre: Is this the expected behavior? I tried with the pre-atomic commands version and this isn't whats happening.

Also, unlike List Items, pressing backspace on an empty task item removes the task instead of turning it into a `TaskNode`.

super_editor/lib/src/default_editor/paragraph.dart
angelosilvestre: As I copied `DeleteUpstreamAtBeginningOfParagraphCommand` I just extracted the functions that both commands use (`moveSelectionToEndOfPrecedingNode` and `mergeTextNodeWithUpstreamTextNode`).

`DeleteUpstreamAtBeginningOfParagraphCommand` has some paragraph specific behavior, like changing the paragraph block type. If I should dedup the whole command then I would need some type checks like:

```dart
  if (node is ParagraphNode) {
   // paragraph behavior.
  }
```

matthew-carroll: But do you know why `moveSelectionToEndOfPrecedingNode` and `mergeTextNodeWithUpstreamTextNode` were public in the first place? Who else was using those methods?

super_editor/lib/src/default_editor/tasks.dart
matthew-carroll: Unless there's a specific reason to keep it that way, I think that pressing Backspace at the start of a task should convert the task to a regular paragraph. I think that would make more sense from a UX perspective.

angelosilvestre: Would this happen only if the task is empty? Or always?

matthew-carroll: I think always. In my mind I'm not sure why a task would be treated differently from a list item, or a block quote, or a header. If you Backspace at the start, you end up with a regular paragraph.

This behavior should be overridable. Because, for example, Superlist might need to remove the task from the database and maybe they want to delete the whole thing. But hopefully they could just register their own request/command pair. For the default behavior, I'd say always convert to a paragraph.

angelosilvestre: Ok, so we won't be merging a task with a preceding paragraph or task. I'm gonna close this PR and create another one to convert the task to a paragraph.

angelosilvestre:
Closing in favor of https://github.com/superlistapp/super_editor/pull/1322.

PR #1286, opened 2023-08-10
Page-up,page-down,home,end scrolls SuperEditor (Resolves #351)

### LINKED ISSUE #351: Page up/down and home/end introduce box-with-question-mark-character

rutvik110:
Page-up,page-down,home,end scrolls SuperEditor (Resolves #351)


super_editor/lib/src/default_editor/document_hardware_keyboard/document_keyboard_actions.dart
```diff
@@ -11,6 +14,189 @@ import 'package:super_editor/src/default_editor/text.dart';
...
+
+  if (defaultTargetPlatform == TargetPlatform.macOS) {
+    if (keyEvent.logicalKey.keyId != LogicalKeyboardKey.home.keyId || !keyEvent.isMetaPressed) {
+      return ExecutionInstruction.continueExecution;
+    }
+  } else {
+    if (keyEvent.logicalKey.keyId != LogicalKeyboardKey.home.keyId || !keyEvent.isControlPressed) {
+      return ExecutionInstruction.continueExecution;
+    }
+  }
```
rutvik110, line 99: @matthew-carroll Updated both scroll-to-top and scroll-to-bottom handlers to respond to the correct key presses. 
This handlers now look up for `Cmd+Home` and `Cmd+End` key presses on mac and `Ctrl+Home` and `Ctrl+End` for all other platforms.  

Wanted your opinion on one thing though. Currently, for all platforms except mac, we're looking up for `ctrl+home/end` key press to scroll the document which is implied directly with the use of `else` statement.  Is that fine? 
Or do you want me to explicitly specify all the other platforms that would be responding to `ctrl+home/end` key presses.

rutvik110:
Also, lot of references to `scrollController` and other implementation details related to it will be removed once [this](https://github.com/superlistapp/super_editor/pull/1292) updates are in.

matthew-carroll:
@rutvik110 I just merged the scrolling API. Please rebase your PR branch on top of `main`.

rutvik110:
> @rutvik110 I just merged the scrolling API. Please rebase your PR branch on top of `main`.

On it!

rutvik110:
@matthew-carroll Updated this PR to use the `DocumentScroller` introduced in recent updates instead of `ScrollController` and removed any previous references related to usage of `ScrollController` and implementation details related to it. 
Let me know if anything needs to be updated.

super_editor/test/test_tools.dart
```diff
@@ -326,6 +326,19 @@ void testWidgetsOnIos(
   }, skip: skip, variant: variant);
 }
 
+@isTestGroup
+void testWidgetsOnAllPlatformsExceptMac(
```
matthew-carroll, line 330: I don't think we want to introduce the concept of negative sets. We already have a lot of variations with positive sets of platforms.

If you want to test non-Mac, then use the existing `testWidgetsOnWindowsAndLinux`. We don't care about Fuschia, and if some other desktop platform emerges, it might not even have the same policy.

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
```diff
@@ -978,6 +978,550 @@ void main() {
...
+          await tester.sendKeyEvent(LogicalKeyboardKey.pageDown);
+
+          // Let the scrolling system auto-scroll, as desired.
+          await tester.pumpAndSettle();
+
+          // Ensure we didn't scrolled past the bottom of the viewport.
+          expect(scrollState.position.pixels, equals(scrollState.position.maxScrollExtent));
+        });
+
+        testWidgetsOnAllPlatforms('PAGE DOWN scrolls down by the viewport height', (tester) async {
```
matthew-carroll, line 1008: In terms of reading a test suite, it would be nice if the normal case came before the exceptional case. For example, this test should come before the test that makes sure PAGE DOWN doesn't exceed the bottom of the viewport.

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
```diff
@@ -978,6 +978,550 @@ void main() {
...
+          await _pumpPageScrollTestSetup(tester);
+
+          await tester.tapAtDocumentPosition(
+            const DocumentPosition(
+              nodeId: '1',
+              nodePosition: TextNodePosition(offset: 0),
+            ),
+          );
+          await tester.pump();
```
matthew-carroll, line 1017: Doesn't `tapAtDocumentPosition` already have a `pumpAndSettle()`? Do you need this extra `pump()`?

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
```diff
@@ -978,6 +978,550 @@ void main() {
...
+
+          // Scroll to the bottom of the viewport.
+          scrollState.position.jumpTo(scrollState.position.maxScrollExtent);
+
+          await tester.sendKeyEvent(LogicalKeyboardKey.pageDown);
+
+          // Let the scrolling system auto-scroll, as desired.
+          await tester.pumpAndSettle();
+
+          // Ensure we didn't scrolled past the bottom of the viewport.
```
matthew-carroll, line 1004: "scrolled" -> "scroll"

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
```diff
@@ -978,6 +978,550 @@ void main() {
           ),
         );
       });
+
+      group("page scrolling", () {
+        testWidgetsOnAllPlatforms('PAGE DOWN does not scroll past bottom of the viewport', (tester) async {
+          await _pumpPageScrollTestSetup(tester);
+
+          await tester.tapAtDocumentPosition(
```
matthew-carroll, line 986: When placing the caret in a paragraph, you can use the much shorter `tester.placeCaretInParagraph()`.

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
```diff
@@ -978,6 +978,550 @@ void main() {
...
+            const DocumentPosition(
+              nodeId: '1',
+              nodePosition: TextNodePosition(offset: 0),
+            ),
+          );
+          await tester.pump();
+
+          final scrollState = tester.state<ScrollableState>(find.byType(Scrollable));
+
+          await _pressCmdHome(tester);
```
matthew-carroll, line 1096: We should add this combination to our testing robots package so that nobody has to do this themselves. Can you put up a PR that adds any of these missing keyboard combinations to that package?

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
```diff
@@ -978,6 +978,550 @@ void main() {
...
+              nodeId: '1',
+              nodePosition: TextNodePosition(offset: 0),
+            ),
+          );
+          await tester.pump();
+
+          final scrollState = tester.state<ScrollableState>(find.byType(Scrollable));
+
+          // Scroll to the bottom of the viewport.
+          scrollState.position.jumpTo(scrollState.position.maxScrollExtent);
```
matthew-carroll, line 1141: When testing the scroll bounds, both for tests at the top and bottom, this probably isn't the most robust test. It's easy to imagine that there's an if-statement that won't scroll if we're already at the destination. The more interesting question is "What happens if I'm very close to the destination and THEN I scroll?"

In each of these cases, how about you move us very close to the start, or the end, but not all the way. And then press the key. When you make that change, please include a short comment in each location that explains why.

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
```diff
@@ -978,6 +978,550 @@ void main() {
...
+
+          await _pressCtrlEnd(tester);
+
+          // Ensure we didn't scrolled past the bottom of the viewport.
+          expect(scrollState.position.pixels, equals(scrollState.position.maxScrollExtent));
+        });
+
+        testWidgetsOnAllPlatformsExceptMac('CTRL + END scrolls to bottom of viewport', (tester) async {
```
matthew-carroll, line 1234: I'm noticing that we're essentially repeating test implementations just so that we can press slightly different keys. Consider creating a test variant that you pass into a single test and then use that test variant to press the desired keys for the platform. That way, you specify the expected behavior only once, but the test runs twice, once for each variant value.

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
```diff
@@ -978,6 +978,550 @@ void main() {
...
+          final scrollState = tester.state<ScrollableState>(find.byType(Scrollable));
+
+          await _pressCtrlEnd(tester);
+
+          // Ensure we scrolled to the bottom of the viewport.
+          expect(scrollState.position.pixels, equals(scrollState.position.maxScrollExtent));
+        });
+      });
+
+      group("page scrolling in presense of an external scrollable", () {
```
matthew-carroll, line 1254: Let's replace this long description with "in an ancestor scrollable"

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
```diff
@@ -978,6 +978,550 @@ void main() {
...
+        testWidgetsOnAllPlatforms('PAGE DOWN does not scroll past bottom of the viewport', (tester) async {
```
matthew-carroll, line 1255: With these tests, we again appear to be repeating almost the entire set of setup, behavior, and expectations. It looks like the only difference is that we call a different setup method. Consider adding these variations with the platform variations to avoid repeating the test implementation.

I'm not entirely sure if the following is the optimal approach, but you could create a variant that has values like:

macWithSuperEditorScrollable
macWithAncestorScrollable
windowsAndLinuxWithSuperEditorScrollable
windowsAndLinuxWithAncestorScrollable

super_editor/lib/src/default_editor/document_hardware_keyboard/document_keyboard_actions.dart
```diff
@@ -11,6 +14,168 @@ import 'package:super_editor/src/default_editor/text.dart';
...
+ExecutionInstruction scrollOnCtrlOrCmdAndHomeKeyPress({
+  required SuperEditorContext editContext,
+  required RawKeyEvent keyEvent,
+}) {
+  if (keyEvent is! RawKeyDownEvent) {
+    return ExecutionInstruction.continueExecution;
+  }
+
+  if (defaultTargetPlatform == TargetPlatform.macOS) {
+    if (keyEvent.logicalKey.keyId != LogicalKeyboardKey.home.keyId || !keyEvent.isMetaPressed) {
```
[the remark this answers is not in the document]
rutvik110: yes, that's much better.

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
rutvik110: That's much concise. Now using `tester.placeCaretInParagraph()` instead of `tester. tapAtDocumentPosition()` wherever needed.

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
rutvik110: yah sure! Will put up a PR once I get done with other suggestions on this PR.

super_editor/test/test_tools.dart
rutvik110: Are we not considering mobile platforms then? On mobile, we might've an external keyboard attached. So, in that case we would want to make sure this works across mobile platforms as well. 

Without this negative test, we'll have to repeat tests for mobile platforms separately.

super_editor/test/src/default_editor/document_keyboard_actions_test.dart
rutvik110: But in this case, the different key presses are required for the different platforms. For mac, we have cmd+home/end and for others ctrl+home/end. 

I'm not sure how using key press variant would help here. Could you please clarify a bit on this?

super_editor/test/test_tools.dart
matthew-carroll: In that case, the least bad solution is probably to test against all platforms and then in the test check if we're on Mac and adjust the behavior. Though I wonder, if someone plugged a keyboard into an iPad, would that keyboard not also have a CMD key?

matthew-carroll:
@rutvik110 did you push your changes? I don't think I see any of the fixes that are marked as resolved.

rutvik110:
> @rutvik110 did you push your changes? I don't think I see any of the fixes that are marked as resolved.

Sry abt that. Just pushed those updates.

super_editor/test/test_tools.dart
rutvik110: They would! In that case, I believe I should fix the handlers and tests to check for cmd key for ios as well along side mac.

Also just checked abt what if a windows keyboard is used for iPad/iPhone. And found [this](https://edu.gcfglobal.org/en/osxbasics/using-a-windows-keyboard-with-a-mac/1/) article mentioning that the Windows key would act as a modifier just like command key. So this scenario won't require any changes from our side.

rutvik110:
@matthew-carroll Pushed a few updates related to refactoring the tests to reduce duplication of tests for slightly different behaviours and adjusted handles and tests to check for `cmd` key press instead of `ctrl` for `iOS`. 

Updates :
* Added a variant that pumps up a `SuperEditor` experience with and without an ancestor scrollable to test the page scrolling behaviour in both cases but without repetition of tests.
* Scroll to `top` and `bottom` handler updated to check for `cmd` key press instead of `ctrl` for `iOS` alongside `macOS`. 
* Page scrolling tests now uses `testWidgetsOnAllPlatforms` for testing scroll to top and bottom behaviour and checks the `defaultTargetPlatform` to press either `cmd + home/end` for `macOS` and `iOS`, or `ctrl + home/end` for all other platforms .

super_editor/lib/src/default_editor/document_hardware_keyboard/document_keyboard_actions.dart
```diff
@@ -11,6 +14,167 @@ import 'package:super_editor/src/default_editor/text.dart';
...
+}) {
+  if (keyEvent is! RawKeyDownEvent) {
+    return ExecutionInstruction.continueExecution;
+  }
+
+  if (keyEvent.logicalKey.keyId != LogicalKeyboardKey.pageUp.keyId) {
+    return ExecutionInstruction.continueExecution;
+  }
+
+  final scrollController = editContext.scroller;
```
matthew-carroll, line 31: Let's call this `scroller` to remain consistent, and to avoid people thinking that it's actually a `ScrollController`.

super_editor/lib/src/default_editor/document_hardware_keyboard/document_keyboard_actions.dart
```diff
@@ -11,6 +14,168 @@ import 'package:super_editor/src/default_editor/text.dart';
...
+  return ExecutionInstruction.haltExecution;
+}
+
+/// Halt execution of the current key event if the [SuperEditorContext.scrollController]
+/// is null and the key pressed is one of the page up, page down, home or end key.
+///
+/// Without this action in place pressing the above mentioned keys in absense of
+/// [SuperEditorContext.scrollController] would display an unknown '?' character in
+/// the document.
+ExecutionInstruction doNothingWhenPageUpOrPageDownOrHomeOrEndKeyPressed({
```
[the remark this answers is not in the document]
matthew-carroll: How about we combine this handler with the Fn key handler and we call it `blockControlKeys`?

super_editor/test/super_editor/supereditor_input_keyboard_actions_test.dart
```diff
@@ -1264,6 +1266,229 @@ void main() {
         ),
       );
     });
+
+    group("page scrolling", () {
+      testWidgetsOnAllPlatforms(
+        'PAGE DOWN scrolls down by the viewport height',
+        (tester) async {
+          await _pumpPageScrollTestWidgetVariant.currentValue!(tester);
```
matthew-carroll, line 1274: The naming here is a bit confusing. I think it would make more sense to refer to the variant as an object, and then call a behavior on it, such as:

```dart
await _scrollingVariant.currentValue!.pumpEditor(tester);
```

super_editor/test/super_editor/supereditor_input_keyboard_actions_test.dart
```diff
@@ -1264,6 +1266,229 @@ void main() {
...
+
+          await tester.placeCaretInParagraph(
+            '1',
+            0,
```
matthew-carroll, line 1278: Let's keep very short argument lists on a single line.

super_editor/test/super_editor/supereditor_input_keyboard_actions_test.dart
```diff
@@ -1264,6 +1266,229 @@ void main() {
...
+          );
+
+          final scrollState = tester.state<ScrollableState>(find.byType(Scrollable));
+
+          // Scroll to the bottom of the viewport.
```
matthew-carroll, line 1309: Looks like my comment about testing these boundary cases hasn't been incorporated. Please take a look at that previous comment.

super_editor/test/super_editor/supereditor_input_keyboard_actions_test.dart
```diff
@@ -1264,6 +1266,229 @@ void main() {
         ),
       );
     });
+
+    group("page scrolling", () {
+      testWidgetsOnAllPlatforms(
+        'PAGE DOWN scrolls down by the viewport height',
```
matthew-carroll, line 1272: I think you need to incorporate the variant in the test name so that we can tell which one fails. For example: `'PAGE DOWN scrolls down by the viewport height ($_scrollVariant.currentValue!.description)'`

And `description` might say `'inner viewport'` and `'ancestor viewport'`

super_editor/test/super_editor/supereditor_input_keyboard_actions_test.dart
```diff
@@ -1264,6 +1266,229 @@ void main() {
...
+            '1',
+            0,
+          );
+
+          final scrollState = tester.state<ScrollableState>(find.byType(Scrollable));
+
+          // Scroll to the bottom of the viewport.
+          scrollState.position.jumpTo(scrollState.position.maxScrollExtent);
+
+          final currentPlatform =
```
matthew-carroll, line 1390: It doesn't look like you need to assign this value to a variable. You can place it directly in the if-statement, right?

super_editor/test/super_editor/supereditor_input_keyboard_actions_test.dart
```diff
@@ -1317,3 +1542,97 @@ Future<TestDocumentContext> _pumpExplicitLineBreakTestSetup(
...
 }
+
+/// Variant to pump up a [SuperEditor] experience with and without a parent [Scrollable].
+final _pumpPageScrollTestWidgetVariant = ValueVariant<Future<void> Function(WidgetTester)>({
+  _pumpPageScrollTestSetup,
+  _pumpPageScrollSliverTestSetup,
+});
+
+Future<TestDocumentContext> _pumpPageScrollTestSetup(WidgetTester tester) async {
+  return await tester.createDocument().withLongDoc().withInputSource(TextInputSource.ime).pump();
```
matthew-carroll, line 1553: Please check other examples about how to spread this declaration across multiple lines for readability.

Also, is there a reason that we're explicitly setting the input source to IME? I'm not sure that should be necessary. In fact, we probably want to run test variants for IME and keyboard input to make sure both configurations work with these keys.

super_editor/test/super_editor/supereditor_input_keyboard_actions_test.dart
```diff
@@ -1317,3 +1542,97 @@ Future<TestDocumentContext> _pumpExplicitLineBreakTestSetup(
       .withEditorSize(size)
       .pump();
 }
+
+/// Variant to pump up a [SuperEditor] experience with and without a parent [Scrollable].
```
matthew-carroll, line 1546: We don't want to say "parent scrollable" we want to say "ancestor scrollable". Please adjust the various uses of that term.

Also, when making the adjustment I mentioned above, please change this description to "Variant for an editor experience with an internal scrollable and an ancestor scrollable."

super_editor/test/super_editor/supereditor_input_keyboard_actions_test.dart
```diff
@@ -1317,3 +1542,97 @@ Future<TestDocumentContext> _pumpExplicitLineBreakTestSetup(
...
+final _pumpPageScrollTestWidgetVariant = ValueVariant<Future<void> Function(WidgetTester)>({
+  _pumpPageScrollTestSetup,
+  _pumpPageScrollSliverTestSetup,
+});
+
+Future<TestDocumentContext> _pumpPageScrollTestSetup(WidgetTester tester) async {
```
matthew-carroll, line 1552: Please add a Dart Doc that tells us what this method does.

super_editor/test/super_editor/supereditor_input_keyboard_actions_test.dart
```diff
@@ -1317,3 +1542,97 @@ Future<TestDocumentContext> _pumpExplicitLineBreakTestSetup(
...
+/// content above the [SuperEditor] and additional content on top of [Scrollable].
+///
+/// By including content above the [SuperEditor], it doesn't have the same origin as the parent [Scrollable].
+///
+/// By including content on top of [Scrollable], it doesn't have the origin at [Offset.zero].
+///
+/// This setup is intended for testing page scrolling actions behaviour in presense of an
+/// ancestor [Scrollable].
+Future<void> _pumpPageScrollSliverTestSetup(WidgetTester tester) async {
+  final doc = longDoc();
```
matthew-carroll, line 1566: I believe this setup can be re-configured to use our `WidgetTester` configuration tools. One of our configuration methods allows us to provide a custom widget builder around the super editor.

super_editor/lib/src/default_editor/document_hardware_keyboard/document_keyboard_actions.dart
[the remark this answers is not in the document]
rutvik110: That's much better I think. Will update it.

super_editor/test/super_editor/supereditor_input_keyboard_actions_test.dart
rutvik110: yah, it was just there to reduce writing the full check twice within if-statements.

rutvik110:
@matthew-carroll Pushed the remaining updates. Let me know if anything else is needed.

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -923,6 +932,7 @@ final defaultImeKeyboardActions = <DocumentKeyboardAction>[
   deleteWordDownstreamWithAltDeleteOnMac,
   deleteWordDownstreamWithControlDeleteOnWindowsAndLinux,
   deleteDownstreamContentWithDelete,
+  blockControlKeys,
```
matthew-carroll, line 935: Placing this handler at the bottom of the list won't have any effect.

super_editor/test/super_editor/supereditor_input_keyboard_actions_test.dart
matthew-carroll: Can you verify the result? Try forcing the test to fail and check the output. I think what Flutter uses is the `toString()` of the variant.

super_editor/test/super_editor/supereditor_input_keyboard_actions_test.dart
```diff
@@ -1264,6 +1266,231 @@ void main() {
...
+          );
+
+          await tester.placeCaretInParagraph('1', 0);
+
+          final scrollState = tester.state<ScrollableState>(find.byType(Scrollable));
+
+          // Scroll very close to the top but not all the way to avoid explicit 
+          // checks comparing scroll offset directly against `minScrollExtent` 
+          // and test scrolling behaviour in more realistic manner.
+          scrollState.position.jumpTo(scrollState.position.minScrollExtent+10);
```
matthew-carroll, line 1369: Please put spaces between math operations => "Extent + 10"

rutvik110:
> Can you verify the result? Try forcing the test to fail and check the output. I think what Flutter uses is the toString() of the variant.

Yup I did. Here's the attached screenshot of test failure putting out a proper description of the variant.
Screenshot 2023-08-18 at 10 02 29 AM

super_editor/lib/src/default_editor/super_editor.dart
rutvik110: That's right. Though we are supposed to place this handler before any character insertion handlers. But we don't have those added here unlike in the `defaultKeyboardActions`. Is there a reason as to why we don't have those here?

rutvik110:
@matthew-carroll I'm setting up a PR for the new key combinations we added in this PR and move them to flutter_test_robot package. Could you please add me as an contributor to that repo?

super_editor/lib/src/default_editor/super_editor.dart
matthew-carroll: These are the handlers for IME input, which means the input comes from the OS, instead of us handling key presses. I'm thinking we probably don't need this handler at all in this list. Can you try removing it and then see if the IME test variants still pass in your test suite?

super_editor/test/super_editor/supereditor_input_keyboard_actions_test.dart
```diff
@@ -1317,3 +1544,147 @@ Future<TestDocumentContext> _pumpExplicitLineBreakTestSetup(
...
+    textInputSource: TextInputSource.keyboard,
+    ),
+  
+});
+
+/// Pumps a [SuperEditor] experience with the default [Scrollable].
+Future<TestDocumentContext> _pumpPageScrollTestSetup(
+  WidgetTester tester,
+  TextInputSource textInputSource,
+  ) async {
```
matthew-carroll, line 1578: I don't think there should be leading spaces on this line.

Also down below, the methods after `tester` should be indented by two spaces compared to the `return` line

super_editor/test/super_editor/supereditor_input_keyboard_actions_test.dart
```diff
@@ -1317,3 +1544,147 @@ Future<TestDocumentContext> _pumpExplicitLineBreakTestSetup(
...
+///
+/// By including content on top of [Scrollable], it doesn't have the origin
+/// at [Offset.zero].
+///
+/// This setup is intended for testing page scrolling actions behaviour in presense
+/// of an ancestor [Scrollable].
+Future<TestDocumentContext> _pumpPageScrollSliverTestSetup(
+  WidgetTester tester,
+  TextInputSource textInputSource,
+  ) async {
```
matthew-carroll, line 1600: Same formatting issues here.

super_editor/test/super_editor/supereditor_input_keyboard_actions_test.dart
```diff
@@ -1317,3 +1544,147 @@ Future<TestDocumentContext> _pumpExplicitLineBreakTestSetup(
...
+  @override
+  String toString() {
+    return "PageScrollSetup: $description, ${textInputSource.toString()}";
+  }
+}
+
+typedef _PumpEditorWidget = Future<TestDocumentContext> Function(
+  WidgetTester tester,
+  TextInputSource textInputSource,
+  );
```
matthew-carroll, line 1690: Same formatting issue here. Are you running the Dart formatter on your code?

rutvik110: yah but I think I had it disabled for a moment before I pushed this changes which resulted in this.

super_editor/lib/src/default_editor/super_editor.dart
rutvik110: Everything works as usual after removing this handler.

rutvik110:
Requested updates are in. We should be good to go now🙌

matthew-carroll (approved):
LGTM. As a reminder, after I merge this, please go through the standard cherry pick process.

rutvik110:
Yup!

PR #1290, opened 2023-08-10
Fixes for the new tagging system

matthew-carroll:
Added export for composer_reactions.dart


matthew-carroll:
Closing this because I brought the fix in another PR.

PR #1291, opened 2023-08-10
Fixes for the new tagging system (stable)

matthew-carroll:
See: https://github.com/superlistapp/super_editor/pull/1290


matthew-carroll:
Closing this because I brought the fix into another PR

PR #1301, opened 2023-08-12
Replaced widget post frame scheduling with a new extension method - also fixes SuperTextField hot reload bug (Resolves #1262)

### LINKED ISSUE #1262: [super_editor] While performing a hot reload and changing the position of the SuperTextField in the widget tree, an exception is thrown
During development if the SuperTextField is moved around the widget tree, and a hot reload is issued, a flutter assertion is thrown:
```
The following assertion was thrown during a scheduler callback:
This widget has been unmounted, so the State no longer has a context (and should be considered defunct).

Consider canceling any active work during "dispose" or using the "mounted" getter to determine if the State is still active.
```

I believe the issue is here, in `reassemble`:
https://github.com/superlistapp/super_editor/blob/5227a938611632f3241a6cca36c4d5f9833b26f8/super_editor/lib/src/infrastructure/super_textfield/android/android_textfield.dart#L237-L241

I'm not sure if text fields other than android also have this problem.

matthew-carroll:
Replaced widget post frame scheduling with a new extension method - also fixes SuperTextField hot reload bug (Resolves #1262)

We have a ridiculous number of post frame callbacks. At some point we should see if a change in approach could eliminate most of them.

### DIFF
--- a/super_editor/lib/src/default_editor/document_gestures_mouse.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_mouse.dart
@@ -15,6 +15,7 @@ import 'package:super_editor/src/default_editor/selection_upstream_downstream.da
 import 'package:super_editor/src/default_editor/text_tools.dart';
 import 'package:super_editor/src/document_operations/selection_operations.dart';
 import 'package:super_editor/src/infrastructure/_logging.dart';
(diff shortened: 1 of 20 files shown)


angelosilvestre (approved):
LGTM with a single comment:

We are not technically calling the method at the next frame, we are calling at the end of the current frame. But I think it doesn't matter.

PR #1308, opened 2023-08-13
Moved SuperTextField into a top-level /src directory and created a dedicated library export (Resolves #688)

### LINKED ISSUE #688: [SuperTextField] - Move the supertextfield package from /infrastructure to the root
Currently the `supertextfield` package in `super_editor` sits at `lib/src/infrastructure/supertextfield`. This came from a time when it was viewed as infrastructure within the broader `super_editor` package. However, `SuperTextField` is becoming more popular in its own right.

To help people quickly find `SuperTextField`, and to represent `SuperTextField` as a sibling tool alongside `SuperEditor`, move the package to `lib/src/supertextfield`.

This should be done at a time when it won't break too many open PRs.

matthew-carroll:
Moved SuperTextField into a top-level /src directory and created a dedicated library export (Resolves #688)

The goal is to move to a /src directory structure that looks like:
 * super_editor
 * super_reader
 * super_text_field
 * infrastructure


angelosilvestre (approved):
LGTM

PR #1310, opened 2023-08-13
Use TapRegions in SuperTextField to support unfocusing when tapping or dragging outside (Resolves #1263)

### LINKED ISSUE #1263: [feature-request] Have SuperTextField work with TapRegion
Currently, it is not possible to easily control the focus of a `SuperTextField` using a `TapRegion` since the handles and tooltip do not use a `TapRegion`. Would it be possible to allow the user of a `SuperTextField` to set a `TapRegion.groupId` for the handles and all the different widgets displayed in the overlay, thereby making possible to easily control the focus, such as:
```dart
  return TapRegion(
      groupId: 'group-id',
      onTapInside: (_) => _focusNode.requestFocus(),
      onTapOutside: (_) => _focusNode.unfocus(),
      child: SuperTextField(
        tapRegionGroupId: 'group-id',
        focusNode: _focusNode,
        // additional properties
      )
    );
}
```

In other words, flutter includes a `TextFieldTapRegion` for its selection handles and the selection toolbar, which enables a developer to use a `TextFieldTapRegion` to expand the focus of a text field to related fields or to close it preemptively. The example above might not necessarily be the best one. But I was wondering if, in general, exposing these settings would make sense for the project or not?

matthew-carroll:
Use TapRegions in SuperTextField to support unfocusing when tapping or dragging outside (Resolves #1263)

This PR adds support for passing a `TapRegion` "group ID" to all widgets associated with a `SuperTextField`. This includes the area of the text field, itself, as well as handles and toolbars.

Out of the box, this change has no impact. However, by providing a group ID, developers can surround a `SuperTextField` with their own `TapRegion` and implement "tap outside to unfocus". I updated one of our example app demos to show this. We should probably rework all the `SuperTextField` demos to be more consistent and appealing, but I'll leave that for another ticket.

### DIFF
--- a/super_editor/example/lib/demos/supertextfield/_emojis_demo.dart
+++ b/super_editor/example/lib/demos/supertextfield/_emojis_demo.dart
@@ -1,3 +1,4 @@
+import 'package:example/demos/supertextfield/demo_text_styles.dart';
 import 'package:flutter/material.dart';
 import 'package:super_editor/super_editor.dart';
 
@@ -121,6 +122,7 @@ class _EmojisTextFieldDemoState extends State<EmojisTextFieldDemo> with TickerPr
                       );
                     },
                     hintBehavior: HintBehavior.displayHintUntilTextEntered,
+                    textStyleBuilder: demoTextStyleBuilder,
                     minLines: 1,
                     maxLines: 1,
                   ),
--- a/super_editor/example/lib/demos/supertextfield/_expanding_multi_line_demo.dart
+++ b/super_editor/example/lib/demos/supertextfield/_expanding_multi_line_demo.dart
@@ -1,3 +1,4 @@
+import 'package:example/demos/supertextfield/demo_text_styles.dart';
 import 'package:flutter/material.dart';
 import 'package:super_editor/super_editor.dart';
 
@@ -110,6 +111,7 @@ class _ExpandingMultiLineTextFieldDemoState extends State<ExpandingMultiLineText
                       );
                     },
                     hintBehavior: HintBehavior.displayHintUntilTextEntered,
+                    textStyleBuilder: demoTextStyleBuilder,
                     minLines: 1,
                     maxLines: 5,
                   ),
--- a/super_editor/example/lib/demos/supertextfield/_interactive_demo.dart
+++ b/super_editor/example/lib/demos/supertextfield/_interactive_demo.dart
@@ -1,10 +1,8 @@
+import 'package:example/demos/supertextfield/demo_text_styles.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 import 'package:super_editor/super_editor.dart';
 import 'package:super_text_layout/super_text_layout.dart';
-
-const brandAttribution = NamedAttribution('brand');
-const flutterAttribution = NamedAttribution('flutter');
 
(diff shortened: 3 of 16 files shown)


angelosilvestre (approved):
LGTM

super_editor/lib/src/infrastructure/super_textfield/ios/ios_textfield.dart
```diff
@@ -470,62 +475,65 @@ class SuperIOSTextFieldState extends State<SuperIOSTextField>
...
-          textScrollController: _textScrollController,
-          isMultiline: _isMultiline,
-          handleColor: widget.handlesColor,
-          showDebugPaint: widget.showDebugPaint,
-          child: TextScrollView(
-            key: _scrollKey,
+    return TapRegion(
+      groupId: widget.tapRegionGroupId,
+      child: NonReparentingFocus(
+        key: _textFieldKey,
```
jellynoone, line 481: Great work, thank you for working on this.

Just one question, does it make sense to actually wrap the text field in the tap region? The drag handles and toolbar indeed have to be for this to work. But for the text field, with the current setup the user will have to wrap their text field in another TapRegion and the same id in order to attach the desired listeners.

While it's also true that with the current implementation if the user would provide a groupId, that would by default also make all the other parts work together.

Looking at how Flutter uses TapRegions, it does the following:
- It uses it with the CupertinoTextField and TextField
- It uses it with the drag handles and toolbar
- It uses in some other elements to ensure platform-like behaviour, for instance, the CupertinoTabBar
- It also uses it with the EditableText, however, the onTapOutside callback can also be passed to the EditableText

With SuperTextField, since the developer is already responsible for wrapping the field in order to attach the listeners, would it make more sense to not wrap the text field and avoid needless double registration?

jellynoone: In addition, the reason I think flutter also wraps its `EditableText` in the `TextRegion` is because this way they implement some default unfocusing behaviour i.e. unfocus on tap outside for non-mobile platforms. With `SuperTextField` the developer would have to have done that either by doing what's done in the examples i.e. use a `GestureDetector` or now with the use of `TapRegion`s.

jellynoone: My thoughts are probably not. Since Flutter uses the `TapRegion` in the `EditableText` and then also in the `TextField` and `CupertinoTextField` which both use the `EditableText`. So such double registration seems to work fine there.

Looking at the implementation of `TapRegion`. The way it makes its registration is through `TapRegionRegistry.of`, which finds the registry through `findAncestorRenderObjectOfType`.

I cannot provide any other practical data, but I think what's done here is probably fine. And in case `SuperTextField` would want to one day, for instance, support similar default gestures handling as `Flutter` does i.e. unfocus on tap outside the text field on non-mobile platforms, then it would simply need to provide the callback together with the `TapRegion` and be done.

matthew-carroll: That makes sense. I think we can go ahead and move forward with this approach. I need to rebase this PR against some SuperTextField refactoring I just merged. Then I'll merge this.

PR #1311, opened 2023-08-13
[SuperEditor] Create IME visualizer (Resolves #1120)

### LINKED ISSUE #1120: Create visualizer for IME events and document changes
We have a track record of bugs related to converting IME deltas to document changes. This behavior is very tricky. IME deltas inherently mask user behavior. They also get transformed internally by Flutter. Flutter hasn't proven its delta API against meaningful editor experiences.

To help us debug and fix IME delta issues, create some kind of visualizer that shows IME deltas and associated document changes.

This visualizer might be hidden in a right-side drawer of the example app, or placed anywhere else that's reasonably accessible.

As a starting point, consider displaying two lists side-by-side: IME deltas and document changes. Visually connect a delta with its associated document changes.

angelosilvestre:
[SuperEditor] Create IME visualizer. Resolves #1120.

This PR is a starting point for #1120.

This PR introduces a `TextInputDebugger` class, which collects events related to text input. It collects events about call to `updateEditingValueWithDeltas`, `performAction`, `setEditingState` and key events.

We could also introduce a `SuperEditorDebugTools` class which can hold debuggers for other areas, for example, commands, requests, etc.

https://github.com/superlistapp/super_editor/assets/7597082/18a24ffc-08d7-4f6c-8bcb-9042de60bdc0


matthew-carroll:
Let's start with the UX. I think we definitely want to reduce the default amount of information shown. We can include small icons on each list item to expand deeper areas of info about each item in the list, but I think we should only show the minimal relevant info for each item.

We can include a small toolbar at the top. That toolbar can include buttons to expand a type of information for all list items, or collapse that information. We can also move that "clear" button into an icon in that toolbar at the top. I'd also like to see the total number of logged events in the toolbar, too.

Looking at the video, I don't know what the up and down arrows mean on the left.

Also, I'm not sure if this is happening or not, but can we mark divisions between group of deltas so that we know what we got in a single batch?

From a brightness standpoint, let's make those whole panel dark to help communicate that it's a developer tool and not part of the editor.

angelosilvestre:
@matthew-carroll  Updated.

> Looking at the video, I don't know what the up and down arrows mean on the left.

The key pointing down means it's an event generated by Flutter, e.g, a key event or a delta. The arrow pointing up means it's an event we generated, like a call to `setEditingState`.

> Also, I'm not sure if this is happening or not, but can we mark divisions between group of deltas so that we know what we got in a single batch?

If we get multiple deltas in the same batch, we will see one expansion tile for the `updateEditingStateWithDeltas` call and each delta is displayed when the tile is expanded.

https://github.com/superlistapp/super_editor/assets/7597082/c1b85bb8-ba2c-4b56-802e-eb6528f8ed2f

PR #1319, opened 2023-08-15
[SuperTextField][Mobile] Always compute the viewport height in the first frame (Resolves #939)

### LINKED ISSUE #939: [SuperTextField] flicker on initial display (iOS)

angelosilvestre:
[SuperTextField][Mobile] Always compute the viewport height in the first frame. Resolves #939

Currently, there are cases where we compute the textfield size after the first frame. This causes some flickering when using it inside a scrollable.

This PR moves the existing viewport height computation to a new widget, which performs the computation during layout. With this change, we always size the viewport since the first frame.

The downside is that we perform layout twice. The first one is needed so the text layout is performed so we can access the text layout information and the second one uses the information retrieved from the first layout to size the viewport. We probably could introduce some caching to avoid performing two layouts in some situations.

### DIFF
--- a/super_editor/lib/src/super_textfield/infrastructure/text_scrollview.dart
+++ b/super_editor/lib/src/super_textfield/infrastructure/text_scrollview.dart
@@ -1,10 +1,9 @@
 import 'dart:math';
 
 import 'package:flutter/material.dart';
+import 'package:flutter/rendering.dart';
 import 'package:flutter/scheduler.dart';
-import 'package:super_editor/src/infrastructure/_logging.dart';
-import 'package:super_editor/src/infrastructure/flutter/flutter_pipeline.dart';
-import 'package:super_editor/src/super_textfield/super_textfield.dart';
+import 'package:super_editor/super_editor.dart';
 import 'package:super_text_layout/super_text_layout.dart';
 
 final _log = scrollingTextFieldLog;
@@ -118,9 +117,6 @@ class _TextScrollViewState extends State<TextScrollView>
 
   final _scrollController = ScrollController();
 
-  bool _needViewportHeight = true;
-  double? _viewportHeight;
-
   @override
   void initState() {
     super.initState();
@@ -129,7 +125,7 @@ class _TextScrollViewState extends State<TextScrollView>
       ..delegate = this
       ..addListener(_onTextScrollChange);
 
-    widget.textEditingController.addListener(_scheduleViewportHeightUpdateAndRebuild);
+    widget.textEditingController.addListener(_scheduleRebuild);
   }
 
   @override
@@ -147,17 +143,13 @@ class _TextScrollViewState extends State<TextScrollView>
     }
 
     if (widget.textEditingController != oldWidget.textEditingController) {
-      oldWidget.textEditingController.removeListener(_scheduleViewportHeightUpdateAndRebuild);
-      widget.textEditingController.addListener(_scheduleViewportHeightUpdateAndRebuild);
-
-      _updateViewportHeight();
+      oldWidget.textEditingController.removeListener(_scheduleRebuild);
+      widget.textEditingController.addListener(_scheduleRebuild);
     }
 
     if (widget.minLines != oldWidget.minLines ||
         widget.maxLines != oldWidget.maxLines ||
-        widget.lineHeight != oldWidget.lineHeight) {
-      _updateViewportHeight();
-    }
+        widget.lineHeight != oldWidget.lineHeight) {}
   }
 
   @override
@@ -166,7 +158,7 @@ class _TextScrollViewState extends State<TextScrollView>
       ..delegate = null
       ..removeListener(_onTextScrollChange);
 
-    widget.textEditingController.removeListener(_scheduleViewportHeightUpdateAndRebuild);
+    widget.textEditingController.removeListener(_scheduleRebuild);
 
     super.dispose();
   }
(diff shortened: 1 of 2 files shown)


matthew-carroll:
> Currently, there are cases where we compute the textfield size after the first frame. This causes some flickering when using it inside a scrollable.

Can you please provide a deeper description of the situation? In which cases do we do that, and why? And do you mean that we compute the size on the first frame AND after the first frame? Or are you saying that we don't compute it on the first frame, and instead we compute the initial size in a later frame?

angelosilvestre:
When we have `minLines` or `maxLines`, we might need to set the textfield size to a fixed size in the following situations:

- If `minLines != null`, we check if the current text height is smaller than the amount of pixels of `minLines` of text. If it is, we set the textfield size to be `minLines * lineHeight`.
- If `maxLines != null`, we check if the current text height is greater than the amount of pixels of `maxLines` of text. If it is, we set the textfield size to be `maxLines * lineHeight`. 

Padding was omitted for simplicity.

To be able to measure the height of the text, text must be laid out first. In the first `build` call the text still isn't laid out, so we defer to the end of the frame.

The goal of this PR is to remove this computations from `TextScroller`, so during the build phase, it doesn't matter if the text is laid out or not.

super_editor/lib/src/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -338,191 +330,10 @@ class _TextScrollViewState extends State<TextScrollView>
...
-      return 0;
-    }
-
-    if (widget.textKey.currentState == null) {
-      return null;
-    }
-
-    return _textLayout.getLineCount();
+  void _scheduleRebuild() {
+    setState(() {
```
matthew-carroll, line 334: How do we know that `textEditingController` won't call this method during a build? If it does, this `setState()` is going to blow up.

super_editor/lib/src/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -1095,24 +897,260 @@ enum _AutoScrollDirection {
...
-/// or unbounded, which means it should expand to fit its content.
-class _ViewportHeight {
-  const _ViewportHeight({
-    this.value,
+/// If the [child]'s desired height is smaller than [minLines], the size is set to acommodate [minLines] of
+/// text, plus [padding].
+///
+/// If the [child]'s desired height is bigger than [maxLines], the size is set to acommodate [maxLines] of
+/// text, plus [padding].
+class _TextViewport extends SingleChildRenderObjectWidget {
```
matthew-carroll, line 909: A "viewport" means the bounds of the visible region of a scrolling experience. This doesn't seem to be a viewport?

super_editor/lib/src/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -1095,24 +897,260 @@ enum _AutoScrollDirection {
   end,
 }
 
-/// The height of the viewport.
+/// Sizes a [child] containing a [SuperText] to respect the given [minLines] and [maxLines].
```
matthew-carroll, line 900: I'm confused by a lot of references `[child]` in this vs the concept of a descendent `SuperText`. If this render object truly handles any subtree that happens to contain a `SuperText` somewhere in it, then surely there's more to consider than just the text height and the padding, right? On the other hand, if the `child` is expected to be a `SuperText` then that changes how these docs should be written.

super_editor/lib/src/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -1095,24 +897,260 @@ enum _AutoScrollDirection {
...
+    required this.textKey,
+    this.minLines,
+    this.maxLines,
+    this.lineHeight,
+    this.padding,
+    super.child,
   });
 
-  /// Whether or not the viewport height is bounded.
+  /// The text that the [child] will display.
```
matthew-carroll, line 920: If the child contains this text, then shouldn't this render object base itself on teh child's layout size? This render object seems to be replicating a bunch of implementation details of text layout, which the `child` should already be doing on our behalf.

super_editor/lib/src/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -1095,24 +897,260 @@ enum _AutoScrollDirection {
...
+      return;
+    }
+
+    _padding = value;
+    markNeedsLayout();
+  }
+
+  @override
+  void performLayout() {
+    if (child == null) {
```
matthew-carroll, line 1047: Is there any reasonable use of this render object where no child would be provided? If not, let's make it required and not worry about thi.

super_editor/lib/src/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -1095,24 +897,260 @@ enum _AutoScrollDirection {
...
+      size = Size.zero;
+      return;
+    }
+
+    // Perform a first layout so we can access information about the text layout.
+    child!.layout(constraints, parentUsesSize: true);
+
+    final viewportHeight = _computeViewportHeight();
+    if (viewportHeight == null) {
```
matthew-carroll, line 1056: I'm not a fan of using a `null` value for this. It's very strange to read. Why would our height be null? Something should probably change here with regard to naming or responsibility breakdown.

super_editor/lib/src/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -531,29 +342,20 @@ class _TextScrollViewState extends State<TextScrollView>
...
-        height: _viewportHeight,
-        child: Stack(
-          children: [
-            _buildScrollView(
-              child: widget.child,
-            ),
-            if (widget.showDebugPaint) ..._buildDebugScrollRegions(),
-          ],
-        ),
+    return _TextViewport(
```
matthew-carroll, line 345: There's no great place to leave this comment, so I'll put it here. Can you write a test that demonstrates the desired effect? It should fail for the current case where we need to shrink, and another test where we need to expand. Then it should pass with these changes.

super_editor/lib/src/super_textfield/infrastructure/text_scrollview.dart
angelosilvestre: The child is a subtree like `Stack`-> `SingleChildScrollView`-> Text. When laying out the child without aditional constraints, it's taking all the available height.

matthew-carroll: What if you surround `SingleChildScrollView` with an `IntrinsicHeight` widget? Or, if this needs to be done imperatively during layout, what if you `computeIntrinsicHeight()` on the `child`?

super_editor/lib/src/super_textfield/infrastructure/text_scrollview.dart
angelosilvestre: Updated.

super_editor/lib/src/super_textfield/infrastructure/text_scrollview.dart
angelosilvestre: Updated.

super_editor/lib/src/super_textfield/infrastructure/text_scrollview.dart
angelosilvestre: Updated.

super_editor/lib/src/super_textfield/infrastructure/text_scrollview.dart
angelosilvestre: Updated.

super_editor/lib/src/super_textfield/infrastructure/text_scrollview.dart
angelosilvestre: Added tests.

super_editor/lib/src/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -1,20 +1,20 @@
 import 'dart:math';
 
 import 'package:flutter/material.dart';
+import 'package:flutter/rendering.dart';
 import 'package:flutter/scheduler.dart';
-import 'package:super_editor/src/infrastructure/_logging.dart';
 import 'package:super_editor/src/infrastructure/flutter/flutter_pipeline.dart';
-import 'package:super_editor/src/super_textfield/super_textfield.dart';
+import 'package:super_editor/super_editor.dart';
```
matthew-carroll, line 7: A piece of `super_editor` shouldn't import `super_editor`.

super_editor/lib/src/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -129,7 +126,7 @@ class _TextScrollViewState extends State<TextScrollView>
       ..delegate = this
       ..addListener(_onTextScrollChange);
 
-    widget.textEditingController.addListener(_scheduleViewportHeightUpdateAndRebuild);
+    widget.textEditingController.addListener(scheduleBuildAfterBuild);
```
matthew-carroll, line 129: Even though you can call this method directly, doing so removes our opportunity to explain why we're doing that. Which means no one will risk touching it again. Can you add back a method and include an explanation in that method for why we need to schedule another build after every text editing change?

super_editor/lib/src/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -147,17 +144,13 @@ class _TextScrollViewState extends State<TextScrollView>
...
+      oldWidget.textEditingController.removeListener(scheduleBuildAfterBuild);
+      widget.textEditingController.addListener(scheduleBuildAfterBuild);
     }
 
     if (widget.minLines != oldWidget.minLines ||
         widget.maxLines != oldWidget.maxLines ||
-        widget.lineHeight != oldWidget.lineHeight) {
-      _updateViewportHeight();
-    }
+        widget.lineHeight != oldWidget.lineHeight) {}
```
matthew-carroll, line 153: This is an empty if-statement.

super_editor/lib/src/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -338,222 +331,26 @@ class _TextScrollViewState extends State<TextScrollView>
...
-        child: Stack(
-          children: [
-            _buildScrollView(
-              child: widget.child,
-            ),
-            if (widget.showDebugPaint) ..._buildDebugScrollRegions(),
-          ],
-        ),
+    return _TextLinesLimiter(
+      text: widget.textEditingController.text,
```
matthew-carroll, line 341: It looks like we're still depending on text from this new widget. Our discussion about intrinsic height was intended to remove text-specific knowledge - especially outside a scrollview. The other conversation just had an "Updated" response. Can you please let me know what you're thinking at this point with this approach?

super_editor/lib/src/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -1,20 +1,20 @@
...
+import 'package:super_editor/super_editor.dart';
 import 'package:super_text_layout/super_text_layout.dart';
 
 final _log = scrollingTextFieldLog;
 
 /// A scrollable that positions its [child] based on text metrics.
 ///
-/// The [child] must contain a [SuperSelectableText] in its tree,
-/// [textKey] must refer to that [SuperSelectableText], and the
+/// The [child] must contain a [SuperTextWithSelection] in its tree,
```
matthew-carroll, line 14: Was this a correction of an old name, or did we actually change the requirements?

super_editor/lib/src/super_textfield/infrastructure/text_scrollview.dart
angelosilvestre: We don't need the `text` property anymore, I forgot to remove it. However, we still need to access the text layout to be able to query the line height so we are able to determine how many pixels `minLines` and `maxLines` of text are.

super_editor/lib/src/super_textfield/infrastructure/text_scrollview.dart
angelosilvestre: A correction of an old name.

super_editor/lib/src/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -1095,24 +894,223 @@ enum _AutoScrollDirection {
   end,
 }
 
-/// The height of the viewport.
+/// Sizes a [child] to respect the given [minLines] and [maxLines].
+///
```
matthew-carroll, line 898: Let's add a line here:

```
/// The height of the [child] is calculated as the [child]'s `intrinsicHeight`, plus the 
/// given [padding] above and below the [child].
```

super_editor/lib/src/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -1095,24 +894,223 @@ enum _AutoScrollDirection {
...
+
+    // Perform a first layout so we can access information about the text layout.
+    child!.layout(constraints, parentUsesSize: true);
+
+    final lineHeight = _computeLineHeight();
+    final minHeight = _computeMinHeight(lineHeight);
+    final maxHeight = _computeMaxHeight(lineHeight);
+
+    // The height we need to enforce if the child doesn't already respects the line restrictions.
+    double? enforcedHeight;
```
matthew-carroll, line 1046: `enforcedHeight` -> `adjustedChildHeight`

super_editor/lib/src/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -1095,24 +894,223 @@ enum _AutoScrollDirection {
...
+
+    final childIntrinsicHeight = child!.getMinIntrinsicHeight(constraints.maxWidth);
+    if (childIntrinsicHeight <= minHeight) {
```
matthew-carroll, line 1049: Should this `<=` be `<`?

super_editor/lib/src/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -1095,24 +894,223 @@ enum _AutoScrollDirection {
...
+    );
+
+    // Perform the second layout with the adjusted constraints.
+    child!.layout(childConstraints, parentUsesSize: true);
+
+    size = child!.size;
+  }
+
+  double _computeMinHeight(double lineHeight) {
+    final totalVerticalPadding = _padding?.vertical ?? 0.0;
```
matthew-carroll, line 1077: Does the word `total` mean something here? This looks like just a normal `verticalPadding`

super_editor/lib/src/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -1095,24 +894,223 @@ enum _AutoScrollDirection {
...
+  EdgeInsets? _padding;
+  set padding(EdgeInsets? value) {
+    if (value == _padding) {
+      return;
+    }
+
+    _padding = value;
+    markNeedsLayout();
+  }
+
```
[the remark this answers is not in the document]
angelosilvestre: This render object extends `RenderProxyBox`, so it already has the implementations and the mixins.

super_editor/test/super_textfield/super_textfield_rendering_test.dart
```diff
@@ -31,6 +32,68 @@ void main() {
...
+        controller: AttributedTextEditingController(
+          text: AttributedText('1'),
+        ),
+        showTextField: showTextField,
+        minLines: 5,
+      );
+
+      // Switch to display the SuperTextField.
+      // As our text has only one line, we should expand the text field height
+      // to acommodate 5 lines.
```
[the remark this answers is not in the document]
angelosilvestre: I think that `// As our text has only one line, we should expand the text field height to 5 lines.` should be enough to communicate what's happening. What do you think?

super_editor/test/super_textfield/super_textfield_rendering_test.dart
```diff
@@ -31,6 +32,68 @@ void main() {
...
+        tester,
+        controller: AttributedTextEditingController(
+          text: AttributedText('1'),
+        ),
+        showTextField: showTextField,
+        minLines: 5,
+      );
+
+      // Switch to display the SuperTextField.
+      // As our text has only one line, we should expand the text field height
```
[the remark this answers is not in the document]
angelosilvestre: > I'm guessing that the reason you're flipping this value is to show the text field and inspect that text field on its very first frame.

Yeah, I will include this information.

angelosilvestre:
@matthew-carroll 

After changing from  `if (childIntrinsicHeight <= minHeight)` to `if (childIntrinsicHeight < minHeight)` the following test failed:

```dart
testWidgetsOnAllPlatforms('recalculates its viewport height when text changes for text bigger than maxLines',
        (tester) async {
      final controller = AttributedTextEditingController();

      await tester.pumpWidget(
        _buildScaffold(
          child: SuperTextField(
            minLines: 1,
            maxLines: 2,
            textController: controller,
          ),
        ),
      );
      await tester.pumpAndSettle();

      final textFieldSizeBefore = tester.getSize(find.byType(SuperTextField));

      // Change the text, so the content height is greater
      // than the initial content height.
      controller.text = AttributedText(
        """
This is
a
multi-line
SuperTextField
""",
      );
      await tester.pumpAndSettle();

      // Ensure the text field height has increased.
      expect(tester.getSize(find.byType(SuperTextField)).height, greaterThan(textFieldSizeBefore.height));
    });
```

As the text is initially empty, the `child` intrinsic height was already between `minLines` and `maxLines`, so we don't enforced any specific height. Because of that, the `SingleChildScrollView` expanded to fill all the available height.

I modified the implementation so that, if the `child`'s intrinsic size already respects `minLines` and `maxLines`, we force it to be exactly the height of its intrinsic height.

matthew-carroll (approved):
LGTM

PR #1321, opened 2023-08-15
Switched SuperEditor from LayerLinks to LeaderLinks so we can add boundary behaviors (Resolves #669)

### LINKED ISSUE #669: [Example App] - Toolbar can move off-screen
The popover toolbar in the example app is not confined to the screen's visual area. We should prevent the toolbar from moving off screen.

You can see a screenshot of the problem here: https://github.com/superlistapp/super_editor/issues/380

This ticket should probably wait for the introduction of a custom transform follower solution (a Flutter Bounty Hunters package).

matthew-carroll:
Switched SuperEditor from LayerLinks to LeaderLinks so we can add boundary behaviors (Resolves #669)

Initially, we were computing an (x,y) position for the toolbar on every frame. Then we recently switched to `LayerLink`s to avoid scheduling an extra frame for every selection change. The `LayerLink` followed an invisible widget in a document overlay. But `LayerLink`s can't avoid obstaces, so this PR switches up to use `LeaderLink`s from `follow_the_leader`, which can be configured to avoid screen boundaries.

Original problem:

Screenshot 2023-08-15 at 11 46 22 AM

New behavior:

https://github.com/superlistapp/super_editor/assets/7259036/3f07fe3e-a97b-4025-a97e-83952fe43f0d

### DIFF
--- a/super_editor/example/lib/demos/example_editor/_toolbar.dart
+++ b/super_editor/example/lib/demos/example_editor/_toolbar.dart
@@ -459,40 +459,8 @@ class _EditorToolbarState extends State<EditorToolbar> {
 
   @override
   Widget build(BuildContext context) {
-    return Stack(
+    return BuildInOrder(
       children: [
-        // Conditionally display the URL text field below
-        // the standard toolbar.
-        // if (_showUrlField)
-        // Positioned(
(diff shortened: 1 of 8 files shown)


angelosilvestre (approved):
LGTM

PR #1325, opened 2023-08-16
Converted AttributedText to use optional position constructor parameters instead of named parameters, for brevity, and also added ability to create an AttributedText directly from a Markdown string (Resolves #488)

### LINKED ISSUE #488: Add AttributedText factories for convenience
Add an `AttributedText` factory method that deserializes from Markdown (place it in super_editor_markdown):

Possibilities:

```dart
// Extension method on String
myMarkdownString.toMarkdownAttributedText();

// Global factory
attributedTextFromMarkdown(myMarkdownString);
```
Markdown support makes it much easier to quickly assemble `AttributedText` for tests, demos, etc.

Reduce the verbosity of plain text `AttributedText`:

```dart
final myText = AttributedText.plain("some text");
```

matthew-carroll:
Converted AttributedText to use optional position constructor parameters instead of named parameters, for brevity, and also added ability to create an AttributedText directly from a Markdown string (Resolves #488)

We create a lot of `AttributedText` instances and doing so is very verbose. There are multiple ways in which its verbose, but one way is that we always require named constructor parameters of "text" and "spans". This PR reduces that verbosity by making both parameters optional and ordered. The following are now legitimate:

```dart
var text = AttributedText();
text = AttributedText("some text");
text = AttributedText("some text", AttributedSpans(...));
```

Additionally, sometimes we write tests where we want an `AttributedText` instance that includes some bold, or italics, or links. Up to this point, we've needed to explicitly specify the start and end markers for each of those attributions. This PR adds a top-level function to the `super_editor_markdown` package called `attributedTextFromMarkdown(String markdown)`. That top-level function parses the markdown and returns an appropriate `AttributedText` with whatever styles are in the markdown string.

This parser certainly isn't efficient compared to directly marker declarations. But we don't care very much about that efficiency in our test suites. We can use this parser to easily create `AttributedText` instances with desired styles by specifying the Markdown instead of a bunch of individual markers.

This PR is a breaking change. All `AttributedText` constructors will be broken by this, but those constructor calls can be fixed by simply removing the "text" and "spans" parameter names.

### DIFF
--- a/attributed_text/lib/src/attributed_text.dart
+++ b/attributed_text/lib/src/attributed_text.dart
@@ -17,10 +17,11 @@ final _log = attributionsLog;
 // TODO: there is a mixture of mutable and immutable behavior in this class.
 //       Pick one or the other, or offer 2 classes: mutable and immutable (#113)
 class AttributedText {
-  AttributedText({
-    this.text = '',
+  AttributedText([
+    String? text,
     AttributedSpans? spans,
-  }) : spans = spans ?? AttributedSpans();
+  ])  : text = text ?? "",
+        spans = spans ?? AttributedSpans();
 
   void dispose() {
     _listeners.clear();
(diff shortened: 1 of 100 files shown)


matthew-carroll:
CC @miguelcmedeiros @brian-superlist - This PR will almost certainly break you, but the fix is very simple. You just need to remove the constructor argument names for `AttributedText`.

brian-superlist:
Thanks @matthew-carroll. It looks like we only create about 20 or so instances directly in our app, should be a quick fix on our side.

angelosilvestre (approved):
LGTM

PR #1336, opened 2023-08-21
[SuperEditor][web] Fix new line duplication (Resolves #1219)

### LINKED ISSUE #1219: [Web] Paragraph new line via Enter key duplication
Reproduces on `super_editor 0.2.6`

### Steps to reproduce

- Launch example app on web
- Put cursor on any paragraph
- Press "Enter" on the keyboard

#### Actual result

- New line gets created inserted twice

https://github.com/superlistapp/super_editor/assets/31796433/5da3c616-a860-4142-8936-210b54636843

angelosilvestre:
[SuperEditor][web] Fix new line duplication. Resolves #1219.

In the Sliver Editor Demo in the Demo App, pressing ENTER is still causing the editor to duplicate the new lines.

The cause is that, if the `inputSource` isn't passed in the constructor, the editor uses IME as the input source, but still uses the keyboard handlers for the desktop input source.

This PR inverts the condition, as a `null` `inputSource` in the constructor will default to IME.

We could also write it as the following: 

```dart
keyboardActions = keyboardActions ??
            (inputSource == null || inputSource == TextInputSource.ime
                ? defaultImeKeyboardActions
                : defaultKeyboardActions),
```

or also

```dart
((inputSource ?? TextInputSource.ime) == TextInputSource.ime
                ? defaultImeKeyboardActions
                : defaultKeyboardActions),
```

I'm happy with updating the PR to match the prefered style.

### DIFF
--- a/super_editor/lib/src/default_editor/super_editor.dart
+++ b/super_editor/lib/src/default_editor/super_editor.dart
@@ -106,7 +106,7 @@ class SuperEditor extends StatefulWidget {
     this.imePolicies = const SuperEditorImePolicies(),
     this.imeConfiguration = const SuperEditorImeConfiguration(),
     this.imeOverrides,
-    List<DocumentKeyboardAction>? keyboardActions,
+    this.keyboardActions,
     this.gestureMode,
     this.contentTapDelegateFactory = superEditorLaunchLinkTapHandlerFactory,
     this.androidHandleColor,
@@ -122,8 +122,6 @@ class SuperEditor extends StatefulWidget {
     this.debugPaint = const DebugPaintConfig(),
   })  : stylesheet = stylesheet ?? defaultStylesheet,
         selectionStyles = selectionStyle ?? defaultSelectionStyle,
-        keyboardActions = keyboardActions ??
-            (inputSource == TextInputSource.keyboard ? defaultKeyboardActions : defaultImeKeyboardActions),
         componentBuilders = componentBuilders != null
             ? [...componentBuilders, const UnknownComponentBuilder()]
             : [...defaultComponentBuilders, const UnknownComponentBuilder()],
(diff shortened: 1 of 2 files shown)


matthew-carroll:
I actually just ran into this myself recently and I think I already made a code change for this in one of my branches. I'll see what I did over there.

brian-superlist:
What's interesting: In Superlist, we use the editor inside a custom scroll view. We also pass an explicit parameter for `inputSource`:

```
SuperEditor(
```

Yet we're still seeing this duplication.

angelosilvestre:
@brian-superlist Are you using a custom list of keyboard actions? On web, we should prevent the editor from handling some key events.

brian-superlist:
My bad, @angelosilvestre -- thanks for the suggestion! I'd missed those new actions and that we don't apply the defaults to our editor. 

Applying all of the keyboard actions in the previous PR fixed both the deletion errors and newline errors in our codebase.

matthew-carroll:
@angelosilvestre what I did was make `keyboardActions` optional as a widget property. Then I created a property in the state object that resolves the actions:

```dart
  /// Returns the key handlers that respond to keyboard events within [SuperEditor].
  List<DocumentKeyboardAction> get _keyboardActions =>
      widget.keyboardActions ??
      (inputSource == TextInputSource.ime ? defaultImeKeyboardActions : defaultKeyboardActions);
```

That property is then used by the key handler and the IME handler, like

```dart
  keyboardActions: [
            for (final plugin in widget.plugins) //
              ...plugin.keyboardActions,
            ...keyboardActions,
          ],
```

angelosilvestre:
@matthew-carroll Updated.

I also removed the following sentence from the `keyboardActions` docs:

```dart
/// These actions are only used when in [TextInputSource.keyboard]
/// mode.
```

We use the `keyboardActions` even in IME mode.

matthew-carroll (approved):
LGTM

PR #1338, opened 2023-08-23
Cherry-pick "[SuperTextField][Mobile] Always compute the viewport height in the first frame (Resolves #939) (#1319)" to stable

### LINKED ISSUE #939: [SuperTextField] flicker on initial display (iOS)

angelosilvestre:
This PR cherry-picks "[SuperTextField][Mobile] Always compute the viewport height in the first frame (Resolves #939) (#1319)" to stable.

### DIFF
--- a/super_editor/lib/src/super_textfield/infrastructure/text_scrollview.dart
+++ b/super_editor/lib/src/super_textfield/infrastructure/text_scrollview.dart
@@ -1,20 +1,21 @@
 import 'dart:math';
 
 import 'package:flutter/material.dart';
+import 'package:flutter/rendering.dart';
 import 'package:flutter/scheduler.dart';
 import 'package:super_editor/src/infrastructure/_logging.dart';
 import 'package:super_editor/src/infrastructure/flutter/flutter_pipeline.dart';
-import 'package:super_editor/src/super_textfield/super_textfield.dart';
 import 'package:super_text_layout/super_text_layout.dart';
+import 'package:super_editor/src/super_textfield/super_textfield.dart';
 
 final _log = scrollingTextFieldLog;
 
 /// A scrollable that positions its [child] based on text metrics.
 ///
-/// The [child] must contain a [SuperSelectableText] in its tree,
-/// [textKey] must refer to that [SuperSelectableText], and the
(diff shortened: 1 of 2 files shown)


knopp:
Sorry, missed this. Calling `child!.getMinIntrinsicHeight` from performLayout might be tricky. For example RenderFlex will throw exception when getting min intrinsic height with CrossAxisAlignment.baseline.

matthew-carroll:
@angelosilvestre can you file a new ticket based on @knopp's comment here and then see what we can do about that? Maybe we have to catch an exception or something. Let's see if we can create a test that fails and then fix it.

knopp:
Would it be possible to layout the text vertically unconstrained and then just use the height? Is intrinsicHeight necessary?

matthew-carroll:
Isn't intrinsic height the same thing as layout without a vertical constraint?

knopp:
In the case of RenderParagraph I believe it is same. But generally speaking it is implemented as separate call (I think the reason is that for some render objects it may be cheaper than calculating actual layout) and in some cases such as RenderFlex with baseline alignment it is not always supported.

matthew-carroll:
What does a baseline RenderFlex produce with infinite vertical constraints?

knopp:
It just lays out as expected. It's only intrinsicHeight that doesn't work.

matthew-carroll:
Ok @angelosilvestre let's see if things still work with two full passes of layout. It's not great for performance, but maybe it will avoid issues with intrinsic height not being supported.

angelosilvestre:
@knopp Even though we take an arbitrary `child`, I wouldn't expect a `Column` / `Row` / `Flex`  in the `_TextLinesLimiter` subtree. Do you have an use case for that?

knopp:
> @knopp Even though we take an arbitrary `child`, I wouldn't expect a `Column` / `Row` / `Flex` in the `_TextLinesLimiter` subtree. Do you have an use case for that?

I think we have a `Row` with two `Text` widgets aligned at baseLine returned by `hintBuilder`.

angelosilvestre:
Ok, that makes sense.

Issue #1274, opened 2023-08-03
Accent panel is not closed when the composing character is deleted


miguelcmedeiros:
Steps to reproduce:
* Setup the keyboard with a language that has accents (e.g. Portuguese or German)
* Run `SuperEditor` or `SuperTextField` demos
* Press and hold letter "a" (wait for the accent panel to be shown)
* Press backspace to delete the "a"

**Expected:** the accent panel is closed
**Actual:** the accent panel remains open

https://github.com/superlistapp/super_editor/assets/31278849/6c872d94-3462-4b89-b738-511beb5afef4

miguelcmedeiros:
Cc @matthew-carroll

angelosilvestre:
The accent panel should be closed automatically by the OS. However, when we handle the BACKSPACE key event (`deleteUpstreamContentWithBackspace`), we prevent the OS from doing so. I haven't found any way to close the panel.

It's possible that we need to avoid handling the BACKSPACE key event, and instead implement the `performSelector` method. https://api.flutter.dev/flutter/services/TextInputClient/performSelector.html 

I filed https://github.com/flutter/flutter/issues/132272#issue-1844169343.

angelosilvestre:
It seems that we need to implement `performSelector` to fix this.

https://github.com/superlistapp/super_editor/issues/836

PR #1339, opened 2023-08-23
[SuperEditor][SuperTextField] Implement performSelector (Resolves #836, #1274)

### LINKED ISSUE #836: [SuperEditor][SuperTextField] - Implement new performSelector IME method
We added a no-op declaration for `performSelector` when Flutter 3.3.4 was in Beta. We did this so that developers could begin to work with `super_editor` on Beta. However, we left the implementation as a no-op (https://github.com/superlistapp/super_editor/pull/835).

Once Flutter 3.3.4 moves to Stable, implement the appropriate behavior for `performSelector`.

angelosilvestre:
[SuperEditor][SuperTextField] Implement performSelector. Resolves #836, #1274.

## The issue

Steps to reproduce:

1. Setup the keyboard with a language that has accents (e.g. Portuguese or German)
2. Run `SuperEditor` or `SuperTextField` demos
3. Press and hold letter "a" (wait for the accent panel to be shown)
4. Press backspace to delete the "a"

The letter is deleted but the accent panel is not closed:

https://github.com/superlistapp/super_editor/assets/31278849/6c872d94-3462-4b89-b738-511beb5afef4

This panel is entirely managed by the IME. As we are handling the key events, we prevent the IME from handling them and closing the panel. The navigation between the characters in the accent panel also doesn't work.

## The solution

We need to change the way we handle shortcuts. On mac, we need to let the key events propagate to the IME and then we handle the selectors the IME generates.

These selectors are like system-level shortcuts. The OS maps the shortcuts to a string, which might be seen as an intent, like `moveLeft:`, `deleteBackward:`, `scrollToBeginningOfDocument:` and others. The user can override the keybinding and the only thing we know is the selector name.

 For each selector,  the `performSelector` method  is called on the `TextInputClient`. By handling `performSelection` instead of the key events, the interaction with the accent panel works as expected: 

https://github.com/superlistapp/super_editor/assets/7597082/d4a9e8d0-20cc-485d-a462-2d0453937591

## Changes in this PR

This PR makes the following changes:

- Implements `performSelector` handling for both SuperEditor and SuperTextField. I used a map to configure the handlers for each selector. We can change it to be a list, like the list of keyboard handlers.
- Implements missing shortcuts: 
  - OPTION + UP ARROW: Move the selection to the beginning of the paragraph. 
  - OPTION + DOWN ARROW: Move the selection to the end of the paragraph.
  - CMD + UP ARROW: Move the selection to the beginning of the document.
  - CMD + DOWN ARROW: Move the selection to the end of the document.
  - SHIFT + OPTION + UP ARROW: Expand the selection to the beginning of the paragraph. 
  - SHIFT + OPTION + DOWN ARROW: Expand the selection to the end of the paragraph.
  - SHIFT + CMD + UP ARROW: Expand the selection to the beginning of the document.
  - SHIFT + CMD + DOWN ARROW: Expand the selection to the end of the document.
- Adds a new `MovementModifier` to move the selection to the beginning or end of a paragraph.
- Wraps the text input handler widgets with an `Actions` widget. This is used to prevent the framework from using the arrow keys to move focus between widgets.
- Changes the new line insertion in `SuperTextField` to be handled by the `performAction` call instead of key handlers. This is needed because, if the accent panel is opened and the user presses ENTER, handling the key event causes a new line to be inserted instead of confirming the select character and closing the panel.
- Changes `SuperTextField` replacement delta handling to consider the incoming composing region. Previously, we were ignoring the composing region. This was causing us to clear the composing region on the IME. Because of that, when trying to navigate through the suggestions in the accent panel, the panel was closed immediately.

## Remarks 	

- Obviously, not all of our shortcuts have a corresponding selector, se we keep handling the key events for these shortcuts.
- I declared private methods for the selector handlers, like `_moveCaretUpstream`, because we already have public methods with conflicting names.
- I didn't added a public API yet. After confirming if we should use a map, list or other approach to map the handlers we should add a public API to override the selector handlers.
- Most of the tests worked without any changes. The flutter test package maps common macOS shortcuts to selectors. During tests, I've noticed the `deleteBackwardByDecomposingPreviousCharacter:` wasn't mapped, so I opened https://github.com/flutter/flutter/pull/132919. This PR is already merged in Flutter master. When we cherry-pick this PR to stable, we will need to include a workaround to generate the platform channel message manually.


angelosilvestre:
@brian-superlist After this PR is merged, you'll need to update your list of keyboard handlers. Most notably, `doNothinOnMac` must be added above the handlers that perform caret movement and content deletion.

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -905,15 +905,25 @@ final defaultKeyboardActions = <DocumentKeyboardAction>[
...
+  doNothingWithBackspaceOnWeb,
+  backspaceToConvertTaskToParagraph,
+  backspaceToUnIndentListItem,
+  backspaceToClearParagraphBlockType,
+  deleteDownstreamCharacterWithCtrlDeleteOnMac,
+  scrollOnCtrlOrCmdAndHomeKeyPress,
+  scrollOnCtrlOrCmdAndEndKeyPress,
+  shiftEnterToInsertNewlineInBlock,
+  deleteToEndOfLineWithCmdDeleteOnMac,
+  doNothinOnMac,
```
matthew-carroll, line 923: It looks like all of these handler movements was related to short-circuiting at this line on Mac, is that right? If so, I'm wondering if we should just have two separate lists. One for Mac and one for non-Mac. Maybe that doesn't make sense, because maybe a keyboard plugged into an iPad would also want this approach? But I'm worried about how easily it will be for future developers to not appreciate the significance of this careful ordering.

super_editor/lib/src/core/document_layout.dart
```diff
@@ -440,6 +440,13 @@ class MovementModifier {
   ///  * [word], which moves text selection word-by-word.
   static const line = MovementModifier('line');
 
+  /// Move text selection paragraph-by-paragraph.
+  ///
+  /// See also:
+  ///
+  ///  * [word], which moves text selection word-by-word.
```
matthew-carroll, line 447: Either all of these "See also"s should reference all the others, or none, but let's not do a half-way reference.

super_editor/lib/src/default_editor/document_hardware_keyboard/document_keyboard_actions.dart
```diff
@@ -199,6 +199,44 @@ ExecutionInstruction doNothingWhenThereIsNoSelection({
   }
 }
 
+ExecutionInstruction doNothinOnMac({
```
matthew-carroll, line 202: To make the purpose of this method even more clear, let's name it `sendKeyEventToMacOs()`

super_editor/lib/src/default_editor/document_hardware_keyboard/document_keyboard_actions.dart
```diff
@@ -199,6 +199,44 @@ ExecutionInstruction doNothingWhenThereIsNoSelection({
...
+    // For the full list of selectors handled by SuperEditor, see the MacOsSelectors class.
+    //
+    // This is needed for the interaction with the accent panel to work.
+    return ExecutionInstruction.blocked;
+  }
+
+  return ExecutionInstruction.continueExecution;
+}
+
+ExecutionInstruction deleteDownstreamCharacterWithCtrlDeleteOnMac({
```
matthew-carroll, line 219: Were we missing this shortcut? And are you sure it's Ctrl and not Cmd?

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
```diff
@@ -163,6 +164,7 @@ class SuperEditorImeInteractorState extends State<SuperEditorImeInteractor> impl
       textDeltasDocumentEditor: _textDeltasDocumentEditor,
       imeConnection: _imeConnection,
       floatingCursorController: widget.floatingCursorController,
+      editContext: widget.editContext,
```
matthew-carroll, line 167: We really don't want to pass the `EditContext` any place that we don't have to. It essentially represents the universe. So `DocumentImeInputClient` now depends on the entire universe for construction. What does it need now that it didn't need before?

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
```diff
@@ -291,33 +293,41 @@ class SuperEditorImeInteractorState extends State<SuperEditorImeInteractor> impl
   Widget build(BuildContext context) {
     return SuperEditorImeDebugVisuals(
       imeConnection: _imeConnection,
-      child: SuperEditorHardwareKeyHandler(
-        focusNode: _focusNode,
-        editContext: widget.editContext,
-        keyboardActions: widget.hardwareKeyboardActions,
-        autofocus: widget.autofocus,
-        child: DocumentSelectionOpenAndCloseImePolicy(
+      child: Actions(
```
matthew-carroll, line 296: It's difficult to read this diff because we changed stuff in the middle of the tree. Is the `Actions` widget the only thing that's changed in this `build()` method?

super_editor/lib/src/default_editor/document_ime/document_ime_communication.dart
```diff
@@ -281,3 +304,349 @@ class DocumentImeInputClient extends TextInputConnectionDecorator with TextInput
...
+/// A callback to handle a `performSelector` call.
+typedef SuperEditorSelectorHandler = void Function(SuperEditorContext context);
+
+/// MacOS selector names that are sent to [TextInputClient.performSelector].
+///
+/// These selectors express the user intent and are generated by shortcuts. For example,
+/// pressing SHIFT + Left Arrow key generates a moveLeftAndModifySelection selector.
+///
+/// The full list can be found on https://developer.apple.com/documentation/appkit/nsstandardkeybindingresponding?changes=_8&language=objc
+class MacOsSelectors {
```
matthew-carroll, line 596: Let's move these into infrastructure under a new platform directory for `mac` in `mac_ime.dart`

super_editor/lib/src/default_editor/document_ime/document_ime_communication.dart
```diff
@@ -281,3 +304,349 @@ class DocumentImeInputClient extends TextInputConnectionDecorator with TextInput
...
+  MacOsSelectors.deleteBackwardByDecomposingPreviousCharacter: _deleteUpstream,
+
+  // Scrolling.
+  MacOsSelectors.scrollToBeginningOfDocument: _scrollToBeginningOfDocument,
+  MacOsSelectors.scrollToEndOfDocument: _scrollToEndOfDocument,
+  MacOsSelectors.scrollPageUp: _scrollToStarOfPage,
+  MacOsSelectors.scrollPageDown: _scrollToEndOfPage,
+};
+
+void _moveCaretUpstream(SuperEditorContext context) {
```
matthew-carroll, line 360: Notice that all of these handlers bypass the entire request/command/reaction edit pipeline. That's no good. We'll need an approach that goes through the pipeline.

super_editor/lib/src/default_editor/document_ime/document_ime_communication.dart
```diff
@@ -281,3 +304,349 @@ class DocumentImeInputClient extends TextInputConnectionDecorator with TextInput
     editorImeLog.info("IME connection was closed");
   }
 }
+
+const defaultEditorSelectorHandlers = <String, SuperEditorSelectorHandler>{
```
matthew-carroll, line 308: My first thought is that we should expose this from `SuperEditor` the same way we do with keyboard handlers. What do you think?

In the case of handling selectors, we should be able to stick with a map because selectors are guaranteed to have unique names. That's not true for key combos so we had to use a priority queue of handlers.

super_editor/test/super_editor/supereditor_input_keyboard_actions_test.dart
```diff
@@ -91,6 +90,212 @@ void main() {
...
+            expect(
+              SuperEditorInspector.findDocumentSelection(),
+              const DocumentSelection(
+                base: DocumentPosition(
+                  nodeId: "1",
+                  nodePosition: TextNodePosition(offset: 0),
+                ),
+                extent: DocumentPosition(
+                  nodeId: "1",
+                  nodePosition: TextNodePosition(offset: 35),
```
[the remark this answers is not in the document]
angelosilvestre: `36` is for the second paragraph. Should I use paragraphs with more different lengths to avoid confusion?

super_editor/test/super_textfield/super_textfield_ime_test.dart
```diff
@@ -198,7 +245,25 @@ void main() {
           expect(SuperTextFieldInspector.findSelection(), const TextSelection.collapsed(offset: 9));
         });
 
-        testWidgetsOnDesktop('when ENTER is pressed at beginning of text', (tester) async {
+        testWidgetsOnMac('when ENTER is pressed in middle of text (on MAC)', (tester) async {
```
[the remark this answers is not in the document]
angelosilvestre: We have an existing test `when ENTER is pressed at beginning of text`, which will append  `(on Windows)` and `(on Linux)`. I thought it would be good to add `(on MAC)` to this test so we could easily find this test in the case of a failure.

super_editor/lib/src/default_editor/text.dart
```diff
@@ -607,6 +609,8 @@ class TextComponentState extends State<TextComponent> with DocumentComponent imp
       }
 
       return TextNodePosition(offset: newOffset);
+    } else if (movementModifier == MovementModifier.paragraph) {
```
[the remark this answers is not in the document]
angelosilvestre: Nope, removed.

super_editor/lib/src/default_editor/super_editor.dart
angelosilvestre: Yeah, this is related to short-circuiting. I'm not sure if an iPad with a keyboard will generate selectors, or if it is a macOS only thing.

We could add docs to inform that the order of the handlers is important. Maybe we could create static methods, or a builder to configure the keyboard handlers? Something like:

```dart
SuperEditorKeyboardHandlers
  .forMac()
  .withIme();

SuperEditorKeyboardHandlers
  .forMac()
  .onWeb()
  .withPrependedHandlers(...)
  .withAppendedHandlers(...)
  .withIme();
```

I'm not sure if it would make it easier or even more complicated.

super_editor/test/super_editor/supereditor_input_keyboard_actions_test.dart
```diff
@@ -1666,6 +1900,26 @@ Future<void> _pressCtrlEnd(WidgetTester tester) async {
   await tester.pumpAndSettle();
 }
 
+Future<void> _pressShiftAltUpArrow(WidgetTester tester) async {
```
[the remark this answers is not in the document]
angelosilvestre: Filed https://github.com/Flutter-Bounty-Hunters/flutter_test_robots/issues/12

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
angelosilvestre: Yep, I just wrapped `SuperEditorHardwareKeyHandler` with `Actions`.

super_editor/lib/src/default_editor/document_ime/document_ime_communication.dart
angelosilvestre: All of these `commonOps` methods seem to generate editor requests instead of manipulating the document/selection directly.

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
angelosilvestre: I'm passing the `EditContext` to provide it to the selector handlers. The majority of the handlers access the `commonOps` and the scrolling selectors need to access the `scroller`.

super_editor/lib/src/default_editor/document_ime/document_ime_communication.dart
angelosilvestre: We can do that, should the handlers still be declared in `document_ime_comunication.dart`. If so, we will need a pattern for the method names, so they don't conflict with other existing methods, like `moveCaretDownstream`.

super_editor/lib/src/default_editor/document_hardware_keyboard/document_keyboard_actions.dart
angelosilvestre: We have a test `control + delete: deletes a single downstream character`. This test started to fail because there isn't a selector being generated for this key combo.

super_editor/lib/src/default_editor/document_ime/document_ime_communication.dart
matthew-carroll: Are these handlers actually about the IME? The selector mapping might be, but these handlers are really just requests, right?

Would it make sense to have a request associated with each of these selectors? We could just dispatch the appropriate request and use the existing request/command pairing mechanism to run the change, right? What do you think about that?

super_editor/test/super_textfield/super_textfield_ime_test.dart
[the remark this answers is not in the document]
angelosilvestre: Nope, the test methods that add the platform name are `testWidgetsOnAllPlatforms`, `testWidgetsOnDesktop`, etc. The test methods that run the test on a single platform don't add any information.

super_editor/lib/src/default_editor/document_hardware_keyboard/document_keyboard_actions.dart
angelosilvestre: It seems that the tests are passing because `deleteDownstreamContentWithDelete` doesn't care if ctrl is pressed or not. As we were running `deleteDownstreamContentWithDelete` on mac, the tests were passing.

Now we are not running `deleteDownstreamContentWithDelete` for mac. This caused the tests to fail.

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
angelosilvestre: The handler implementation could come from outside, but even that way, we need to pass the `EditContext` to those handlers. The way you are suggesting, the handlers would receive just the selector names as an argument.

An alternative approach is like I did in SuperTextField, where I exposed a stream of `performSelector` calls.

super_editor/lib/src/default_editor/document_ime/document_ime_communication.dart
angelosilvestre: I guess it makes sense. I'll try to change it.

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
angelosilvestre: I'll try to do this in another way.

super_editor/test/super_textfield/super_textfield_ime_test.dart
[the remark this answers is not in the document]
matthew-carroll: Hm. That seems like an oversight. Any reason that we shouldn't add those postfixes?

super_editor/lib/src/default_editor/document_hardware_keyboard/document_keyboard_actions.dart
matthew-carroll: Ok. Makes sense.

super_editor/test/super_textfield/super_textfield_ime_test.dart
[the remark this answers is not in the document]
angelosilvestre: I can't think of any reason. Adding the postfixes directly in `testWidgetsOnMac` seems reasonable.

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
```diff
@@ -116,6 +123,11 @@ class SuperEditorImeInteractor extends StatefulWidget {
   /// a property on this IME interactor.
   final FloatingCursorController? floatingCursorController;
 
+  /// Map selector names to its handlers.
```
matthew-carroll, line 126:
```
/// Handlers for all Mac OS "selectors" reported by the IME.
///
/// The IME reports selectors as unique `String`s, therefore selector handlers are
/// defined as a mapping from selector names to handler functions.
```

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -575,6 +576,7 @@ class SuperEditorState extends State<SuperEditor> {
               ...plugin.keyboardActions,
             ..._keyboardActions,
           ],
+          selectorHandlers: defaultEditorSelectorHandlers,
```
matthew-carroll, line 579: Shouldn't this be configurable, like keyboard handlers?

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -905,15 +907,25 @@ final defaultKeyboardActions = <DocumentKeyboardAction>[
...
+  doNothingWithBackspaceOnWeb,
+  backspaceToConvertTaskToParagraph,
+  backspaceToUnIndentListItem,
+  backspaceToClearParagraphBlockType,
+  deleteDownstreamCharacterWithCtrlDeleteOnMac,
+  scrollOnCtrlOrCmdAndHomeKeyPress,
+  scrollOnCtrlOrCmdAndEndKeyPress,
+  shiftEnterToInsertNewlineInBlock,
+  deleteToEndOfLineWithCmdDeleteOnMac,
+  sendKeyEventToMacOs,
```
matthew-carroll, line 925: Let's call out this line with a comment above it.

```
// WARNING: No keyboard handlers below this point will run on Mac. On Mac, most
// common shortcuts are recognized by the OS. This line short circuits Super Editor
// handlers, passing the key combo to the OS on Mac. Place all custom Mac key
// combos above this handler.
```

super_editor/lib/src/super_textfield/input_method_engine/_ime_text_editing_controller.dart
```diff
@@ -62,6 +65,10 @@ class ImeAttributedTextEditingController extends AttributedTextEditingController
   Brightness get keyboardAppearance => _keyboardAppearance;
   Brightness _keyboardAppearance;
 
+  /// A stream of `performSelector` calls.
+  Stream<String> get onPerformSelector => _onPerformSelectorController.stream;
```
matthew-carroll, line 69: Why is this a `Stream`?

super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -380,20 +382,28 @@ class SuperDesktopTextFieldState extends State<SuperDesktopTextField> implements
...
+        textController: _controller,
+        textKey: _textKey,
+        keyboardActions: widget.keyboardHandlers,
+        child: widget.inputSource == TextInputSource.ime
+            ? SuperTextFieldImeInteractor(
+                textKey: _textKey,
+                focusNode: _focusNode,
+                textController: _controller,
+                isMultiline: isMultiline,
+                child: child,
```
matthew-carroll, line 403: Shouldn't selector handlers be passed in here, just like `keyboardActions` in the widget above?

super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -1043,6 +1075,7 @@ class _SuperTextFieldImeInteractorState extends State<SuperTextFieldImeInteracto
 
           widget.textController.attachToIme(
             textInputType: widget.isMultiline ? TextInputType.multiline : TextInputType.text,
+            textInputAction: widget.isMultiline ? TextInputAction.newline : TextInputAction.done,
```
matthew-carroll, line 1078: Shouldn't the action be configurable? In fact, I thought I already implemented that?

super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -1112,6 +1145,30 @@ class _SuperTextFieldImeInteractorState extends State<SuperTextFieldImeInteracto
...
+  }
+
+  /// Handles actions from the IME
+  void _onPerformActionPressed(TextInputAction action) {
+    switch (action) {
+      case TextInputAction.newline:
+        widget.textController.insertNewline();
+        break;
+      default:
+        _log.warning("User pressed unhandled action button: $action");
```
matthew-carroll, line 1168: Shouldn't we be dispatching this in some way so that the app can handle it? For example, if the user presses "Next", the app needs to move focus to the next form field. How would they do that?

super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -1546,6 +1603,8 @@ const defaultTextFieldImeKeyboardHandlers = <TextFieldKeyboardHandler>[
   DefaultSuperTextFieldKeyboardHandlers.copyTextWhenCmdCIsPressed,
   DefaultSuperTextFieldKeyboardHandlers.pasteTextWhenCmdVIsPressed,
   DefaultSuperTextFieldKeyboardHandlers.selectAllTextFieldWhenCmdAIsPressed,
+  DefaultSuperTextFieldKeyboardHandlers.doNothingOnMac,
```
matthew-carroll, line 1606: Please include a similar warning comment like Super Editor.

super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -1964,6 +2022,24 @@ class DefaultSuperTextFieldKeyboardHandlers {
     return TextFieldKeyboardHandlerResult.handled;
   }
 
+  static TextFieldKeyboardHandlerResult doNothingOnMac({
```
matthew-carroll, line 2025: Please rename this to match the Super Editor handler

super_editor/lib/src/super_textfield/input_method_engine/_ime_text_editing_controller.dart
angelosilvestre: We might get an already constructed `ImeAttributedTextEditingController` in the place where we need to set the `performSelector` handler.

We can define a setter for this handler, but we might end up silently overriding the handler if it's already set. But I guess it won't make sense to have more than one method listening for `performSelector` calls.

super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
angelosilvestre: We implemented that for mobile only.

super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
angelosilvestre: Currently, even on the mobile textfields we don't have a way to apps override the  `onPerformActionPressed`.

I changed the implementation to handle the same way as the mobile textfields do.

super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -1112,6 +1157,39 @@ class _SuperTextFieldImeInteractorState extends State<SuperTextFieldImeInteracto
...
+    }
+
+    handler(
+      controller: widget.textController,
+      textLayout: widget.textKey.currentState!.textLayout,
+    );
+  }
+
+  /// Handles actions from the IME.
+  void _onPerformActionPressed(TextInputAction action) {
```
matthew-carroll, line 1174: In terms of the naming conventions that you seem to be using here, this should probably be called `_onPerformAction()`

super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
matthew-carroll: Can you file an issue for this? I suppose we can take this action automatically in this PR, but we don't want to force any particular behavior on developers.

super_editor/lib/src/super_textfield/input_method_engine/_ime_text_editing_controller.dart
```diff
@@ -62,6 +66,16 @@ class ImeAttributedTextEditingController extends AttributedTextEditingController
   Brightness get keyboardAppearance => _keyboardAppearance;
   Brightness _keyboardAppearance;
 
+  /// Handles a selector generated by the IME.
+  ///
+  /// For the list of selectors, see [MacOsSelectors].
+  void Function(String selectorName)? get onPerformSelector => _onPerformSelector;
```
matthew-carroll, line 72: Does this need getters and setters? Can it just be a public property?

angelosilvestre: It can be public.

super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
angelosilvestre: Will do.

super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -974,6 +989,8 @@ class SuperTextFieldImeInteractor extends StatefulWidget {
     required this.focusNode,
     required this.textController,
     required this.isMultiline,
+    required this.selectorHandlers,
```
matthew-carroll, line 992: Should this be set to the defaults?

angelosilvestre: I'm passing the defaults where we are instantiating the `SuperTextFieldImeInteractor`

PR #1340, opened 2023-08-23
[SuperTextField][mobile] Avoid calling `getMinIntrinsicHeight` in `TextScrollview` (Resolves #939)

### LINKED ISSUE #939: [SuperTextField] flicker on initial display (iOS)

angelosilvestre:
[SuperTextField][mobile] Avoid calling `getMinIntrinsicHeight` in `TextScrollview. Resolves #939.

This PR fixes an issue pointed out by @knopp, where calling `getMinIntrinsicHeight` causes an exception in `RenderFlex` if it has a baseline `crossAxisAlignment`.

I was able to write a failing test with the previous implementation.

This PR removes the `getMinIntrinsicHeight` and instead, lays out the child with infinity height constraints.


matthew-carroll (approved):
LGTM

PR #1345, opened 2023-08-24
Don't delete content when control keys are pressed with a selection (Resolves #1331)

### LINKED ISSUE #1331: [SuperEditor] - Tab key deletes content
Reproduction: select some text, press TAB.

Expected: focus moves and content remains unchanged.

Actual: focus moves, selected content is deleted.

We need to adjust the keyboard handlers to keep non-content keys from deleting content.

matthew-carroll:
Don't delete content when control keys are pressed with a selection (Resolves #1331)

### DIFF
--- a/super_editor/lib/src/default_editor/document_hardware_keyboard/document_keyboard_actions.dart
+++ b/super_editor/lib/src/default_editor/document_hardware_keyboard/document_keyboard_actions.dart
@@ -149,16 +149,17 @@ ExecutionInstruction blockControlKeys({
   required SuperEditorContext editContext,
   required RawKeyEvent keyEvent,
 }) {
-  if (keyEvent.logicalKey.keyId != LogicalKeyboardKey.pageUp.keyId &&
-      keyEvent.logicalKey.keyId != LogicalKeyboardKey.pageDown.keyId &&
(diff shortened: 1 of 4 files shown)


rutvik110:
I'm not able to reproduce the issue this PR solves on my side on `main`?

rutvik110:
Now that I remember, I believe we address an similar issue earlier in this (https://github.com/superlistapp/super_editor/pull/1286) PR where pressing control keys would insert an `?` character while deleting the selected content.  I believe that's why I'm not able to reproduce this issue now.

rutvik110 (approved):
LGTM!

angelosilvestre:
After https://github.com/superlistapp/super_editor/pull/1339 is merged, we need to make sure this handler is placed after the short-circuit handler, to let the IME handle these keys.

matthew-carroll:
@angelosilvestre You can make that change, right? If I merge this in then you can rebase?

matthew-carroll:
Actually, @angelosilvestre, this handler is already merged in isn't it? I only changed what it filters. Either way, I'll merge this and you can update as needed.

angelosilvestre (commented):
LGTM

PR #1350, opened 2023-08-29
[super_editor_markdown] Add header alignments

### LINKED ISSUE #865: [SuperEditorMarkdown] Alignment for the header text
When using the `SuperEditorMarkdown` with `MarkdownSyntax.superEditor`, it can only save the alignment for the Paragraph (only `p` tag), I think that it would be nicer if it also supports the header text. I know that we can write a custom serializer to do it but I think it should be implemented by default to be compatible with the `SuperEditor`.

vnniz:
@matthew-carroll Here are my implementations for the header alignments.

Fixes #865


matthew-carroll:
@vnniz as I mentioned in the ticket, I'm still trying to determine what syntax you're proposing. Can you please return to the ticket and explain the Markdown syntax? We need to decide that this is something we want before taking time to review an implementation.

super_editor_markdown/lib/src/document_to_markdown_serializer.dart
```diff
@@ -348,3 +349,77 @@ class AttributedTextMarkdownSerializer extends AttributionVisitor {
     return "";
   }
 }
+
+/// [DocumentNodeMarkdownSerializer] for serializing [ParagraphNode]s as standard Markdown
+/// paragraphs.
+///
+/// Includes support for headers, blockquotes, and code blocks.
+class HeaderNodeSerializer extends NodeTypedDocumentNodeMarkdownSerializer<ParagraphNode> {
```
angelosilvestre, line 357: Please update the docs. It's mentioning that this is a serializer for all paragraph nodes, but it only serializes headers, right?

super_editor_markdown/lib/src/document_to_markdown_serializer.dart
```diff
@@ -348,3 +349,77 @@ class AttributedTextMarkdownSerializer extends AttributionVisitor {
...
+        blockType == header3Attribution ||
+        blockType == header4Attribution ||
+        blockType == header5Attribution ||
+        blockType == header6Attribution) {
+      // this node is a header node
+
+    } else {
+      // this node is not a header node
+      return null;
+    }
```
angelosilvestre, line 381: Can you try to remove this empty if block? You could try to invert the condition, like: 

```dart
if (!isHeaderNode) {
  return null;
}
```

super_editor_markdown/lib/src/document_to_markdown_serializer.dart
```diff
@@ -348,3 +349,77 @@ class AttributedTextMarkdownSerializer extends AttributionVisitor {
...
+      buffer.write('## ${node.text.toMarkdown()}');
+    } else if (blockType == header3Attribution) {
+      buffer.write('### ${node.text.toMarkdown()}');
+    } else if (blockType == header4Attribution) {
+      buffer.write('#### ${node.text.toMarkdown()}');
+    } else if (blockType == header5Attribution) {
+      buffer.write('##### ${node.text.toMarkdown()}');
+    } else if (blockType == header6Attribution) {
+      buffer.write('###### ${node.text.toMarkdown()}');
+    }
```
angelosilvestre, line 413: This block should be updated to serialize the alignment token after the `#`.

angelosilvestre (changes requested):
I discussed with Matt about the syntax and we agreed we should add the alignment notation in the same line of the header, after the `#`.

If you want to proceed with this PR, please update it to the following syntax:

`#:--- Header` or `# Header` => Left Alignment
`#:---: Header` => Center Alignment
`#---: Header` => Right Alignment
`#-::- Header` => Justify Alignment

Also, please add tests for parsing a header with alignment, also updating the `'example doc'` and `'example doc 1'` tests to include headers with all alignments. 

If you are not willing to make those changes, please let me know and I will create a new PR based on this one.

super_editor_markdown/lib/src/document_to_markdown_serializer.dart
```diff
@@ -348,3 +349,72 @@ class AttributedTextMarkdownSerializer extends AttributionVisitor {
     return "";
   }
 }
+
+/// [DocumentNodeMarkdownSerializer] for serializing [ParagraphNode]s as headers
```
angelosilvestre, line 353: Please end the docs with a `.`.

Also can you add an information explaining that this parser supports the alignment notation.

super_editor_markdown/lib/src/markdown_to_document_parsing.dart
```diff
@@ -644,6 +648,82 @@ class _LineBreakSeparatedElement extends md.Element {
...
 
+/// Parses a header preceded by an alignment token.
+class _HeaderWithAlignmentSyntax extends md.BlockSyntax {
+  /// This pattern matches the text alignment notation.
+  ///
+  /// Possible values are `:---`, `:---:`, `---:` and `-::-`.
+  static final _alignmentNotationPattern = RegExp(r'^:-{3}|:-{3}:|-{3}:|-::-$');
+
+  /// Use internal HeaderSyntax
+  final _headerSyntax = const md.HeaderSyntax();
```
angelosilvestre, line 659: Please end the docs with a `.`.

super_editor_markdown/lib/src/markdown_to_document_parsing.dart
```diff
@@ -644,6 +648,82 @@ class _LineBreakSeparatedElement extends md.Element {
...
+
+  @override
+  RegExp get pattern => RegExp('');
+
+  @override
+  bool canEndBlock(md.BlockParser parser) => false;
+
+  @override
+  bool canParse(md.BlockParser parser) {
+    //
```
angelosilvestre, line 669: Please remove the empty comment.

super_editor_markdown/lib/src/markdown_to_document_parsing.dart
```diff
@@ -644,6 +648,82 @@ class _LineBreakSeparatedElement extends md.Element {
...
+    if (!_alignmentNotationPattern.hasMatch(parser.current)) {
+      return false;
+    }
+
+    final nextLine = parser.peek(1);
+
+    // We found a match for a paragraph alignment token. However, the alignment token is the last
+    // line of content in the document. Therefore, it's not really a paragraph alignment token, and we
+    // should treat it as regular content.
+    if (nextLine == null) return false;
```
angelosilvestre, line 679: For readability, we prefer to always use brackets, even when the if body is a single line, e.g.:

```dart
if (nextLine == null) {
  return false;
}
```

super_editor_markdown/test/super_editor_markdown_test.dart
```diff
@@ -32,6 +32,64 @@ void main() {
...
+          ),
+        ]);
+        // Even when using superEditor markdown syntax, which has support
+        // for text alignment, we don't add an alignment token when
+        // the paragraph is left-aligned.
+        // Paragraphs are left-aligned by default, so it isn't necessary
+        // to serialize the alignment token.
+        expect(serializeDocumentToMarkdown(doc), '# Header1');
+      });
+      test('header with center alignment', () {
```
angelosilvestre, line 53: Please add a blank line between each test.

angelosilvestre (commented):
I left some comments. After addressing them you can tag @matthew-carroll for his review.

vnniz:
@matthew-carroll I have formatted the code and improved the comments. Please give it a look.

matthew-carroll:
@vnniz there seems to be a number of merge conflicts with your PR. Please rebase against main and resolve those.

vnniz:
@matthew-carroll done.

super_editor_markdown/lib/src/document_to_markdown_serializer.dart
```diff
@@ -362,3 +363,76 @@ class AttributedTextMarkdownSerializer extends AttributionVisitor {
     return "";
   }
 }
+
+/// [DocumentNodeMarkdownSerializer] for serializing [ParagraphNode]s as headers.
+///
+/// We already had a header parser in [ParagraphNodeSerializer] but without the alignment,
```
matthew-carroll, line 369: Let's not refer to the passage of time during the development of the code. Please describe these details in the present tense.

```dart
/// Headers are represented by `ParagraphNode`s and therefore this serializer must
/// run before a [ParagraphNodeSerializer], so that this serializer can process
/// header-specific details, such as header alignment.
```

super_editor_markdown/lib/src/document_to_markdown_serializer.dart
```diff
@@ -362,3 +363,76 @@ class AttributedTextMarkdownSerializer extends AttributionVisitor {
     return "";
   }
 }
+
+/// [DocumentNodeMarkdownSerializer] for serializing [ParagraphNode]s as headers.
```
matthew-carroll, line 367: Please adjust this first line:

```dart
/// [DocumentNodeMarkdownSerializer], which serializes Markdown headers to
/// [ParagraphNode]s with an appropriate header block type, and (optionally) a
/// block alignment.
```

super_editor_markdown/lib/src/document_to_markdown_serializer.dart
```diff
@@ -362,3 +363,76 @@ class AttributedTextMarkdownSerializer extends AttributionVisitor {
...
+  }
+
+  @override
+  String doSerialization(Document document, ParagraphNode node) {
+    final buffer = StringBuffer();
+
+    final Attribution? blockType = node.getMetadataValue('blockType');
+    final String? textAlign = node.getMetadataValue('textAlign');
+
+    // Left alignment is the default, so there is no need to add the alignment token.
```
matthew-carroll, line 406: This comment is in a strange location. It's unclear how it relates to the code. After reading this comment, I go down a few lines, and we're adding the alignment token, so it's confusing to see a comment that says "there is no need to add the alignment token". Please ensure that comments appear in places that clarify code behavior, rather than confuse it.

super_editor_markdown/lib/src/document_to_markdown_serializer.dart
matthew-carroll: This comment is no longer applicable, right?

super_editor_markdown/lib/src/markdown_to_document_parsing.dart
```diff
@@ -694,6 +698,86 @@ class _TaskSyntax extends md.BlockSyntax {
   }
 }
 
+/// Parses a header preceded by an alignment token.
```
matthew-carroll, line 701: Does this parser only parse headers that have an alignment token? Or does it parse all headers? If it parses all headers then please adjust this line to `Parses Markdown headers, possibly preceded by an optional alignment token.`

super_editor_markdown/lib/src/markdown_to_document_parsing.dart
```diff
@@ -694,6 +698,86 @@ class _TaskSyntax extends md.BlockSyntax {
...
+///
+/// We already had a header serializer in [_ParagraphWithAlignmentSyntax] but without the alignment,
```
matthew-carroll, line 703: As with the other comment, we shouldn't write time-based descriptions of source code changes. That's what the git history is for. Please describe things in present tense.

vnniz (commented):
@matthew-carroll  I have updated some comments and reviews.

super_editor_markdown/lib/src/document_to_markdown_serializer.dart
matthew-carroll: The primary issue is the location of the comment. The code directly below that comment doesn't implement "no need to add the alignment token". This comment actually refers to a non-existent `else` statement.

The code that actually exists is code that adds an alignment token. Not code that avoids adding an alignment token.

super_editor_markdown/lib/src/markdown_to_document_parsing.dart
matthew-carroll: Is there any reason that we shouldn't parse all headers in the header parser?

vnniz: We have a parser for `_ParagraphWithAlignmentSyntax` that only parses paragraphs with alignment, so we should expect the same behavior for the `_HeaderWithAlignmentSyntax`.

super_editor_markdown/lib/src/document_to_markdown_serializer.dart
vnniz:
``` dart
// Left alignment is the default, so there is no need to add the alignment token.
if (markdownSyntax == MarkdownSyntax.superEditor && textAlign != null && textAlign != 'left') {
   final alignmentToken = _convertAlignmentToMarkdown(textAlign);
   if (alignmentToken != null) {
      buffer.writeln(alignmentToken);
  }
}
```

I think it would be clear if the comment is "Left alignment is the default, so there is no need to add the **left** alignment token.".

matthew-carroll: You're still phrasing the comment in terms of what the code ISN'T doing, instead of what the code IS doing. Please write code comments in terms of what the code IS doing.

vnniz: This comment has been updated as you requested.

matthew-carroll (approved):
LGTM - Thanks for the contribution.

matthew-carroll:
@angelosilvestre can you please cherry pick this to stable?

PR #1351, opened 2023-08-29
[SuperEditor] Fix composer preferences after text replacement (Resolves #1167)

### LINKED ISSUE #1167: [super_editor][IOS]When the software keyboard automatically completes the text, it causes the loss of the text style.
super_editor: 0.2.6

As the title suggests, if I set the text to bold before entering text, and then after entering some content, the software keyboard pops up the recommended content, clicking on the recommended text content will cause the previously set bold effect to be lost.

https://github.com/superlistapp/super_editor/assets/37604230/ec610720-acb2-49e9-beae-c0f19b436934

angelosilvestre:
[SuperEditor] Fix composer preferences after text replacement. Resolves #1167 

Accepting keyboard suggestions can cause the text to lose the style, as can be seen in the following video:

https://github.com/superlistapp/super_editor/assets/37604230/ec610720-acb2-49e9-beae-c0f19b436934

When we receive the replacement delta generated by the IME, we first delete the old text and then insert the new one. Deleting the old text might clear the composer current attributions. This is what is causing the issue.

This PR changes `insertPlainText`, so when we are replacing text, we store the current attributions. After deleting the old text we insert the new one with the stored attributions.

### DIFF
--- a/super_editor/lib/src/default_editor/common_editor_operations.dart
+++ b/super_editor/lib/src/default_editor/common_editor_operations.dart
@@ -1417,10 +1417,18 @@ class CommonEditorOperations {
       return false;
     }
 
+    late Set<Attribution>? replacedTextAttributions;
+
     if (!composer.selection!.isCollapsed) {
       // The selection is expanded. Delete the selected content
       // and then insert the new text.
       editorOpsLog.fine("The selection is expanded. Deleting the selection before inserting text.");
+
+      // As we are replacing text by deleting the selection and then inserting the new text,
+      // we need to store the current attributions.
+      // This is required as deleting text can clear the composer current attributions.
+      // Without this, the new text doesn't preserve the attributions of the replaced text.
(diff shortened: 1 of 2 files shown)


matthew-carroll:
My thought is that `insertPlainText()` isn't the right place to make this adjustment. It seems that the very nature of this issue is that we're not trying to "insert plain text". We're trying to insert text with some set of attributions.

Two possible approaches come to mind:

- After deleting the text in question, add attributions to to the `DocumentComposer` so that text that's inserted at that caret position receive the desired attributions, or...
- Take the inserted text, turn it into attributed text, apply the desired attributions, and then insert it.

angelosilvestre:
@matthew-carroll I simplified the implementation. In `insertPlainText`, the plain text refers to the text received as argument. We already insert the text using the composer attributions.  

The deletion also happens inside `insertPlainText`.

matthew-carroll (approved):
LGTM

PR #1357, opened 2023-08-30
Adjusted tag indices to only notify listeners when tags actually change (Resolves #1355)

### LINKED ISSUE #1355: [SuperEditor] - Only notify tag listeners when tags change
Currently, our tagging implementation notifies listeners that the tag index has changed every time the reaction runs. Adjust the tagging implementations so that listeners are only notified of changes when a change actually occurs.

matthew-carroll:
Adjusted tag indices to only notify listeners when tags actually change (Resolves #1355)

Previously the pattern tag and stable tag indices were sending notifications during every reaction. Now they only send notifications when a tag is added, removed, or changed.


angelosilvestre (approved):
LGTM

PR #1359, opened 2023-08-30
[SuperEditor] Apply theme brightness to the software keyboard (Resolves #1003)

### LINKED ISSUE #1003: [SuperEditor] Does not respect the `Theme.brightness`
SuperEditor keyboard has brightness always set to dark. 

This means that when `Brightness.light` is set on the `ThemeData`, the keyboard on Flutter `TextField`s are bright whereas the keyboard on SuperEditor remains dark

angelosilvestre:
[SuperEditor] Apply theme brightness to the software keyboard. Resolves #1003

Currently, SuperEditor doesn't look at the theme to determine the keyboard appearance. The ticket mentioned that the keyboard is always dark, but it seems to be the contrary. Even when the the brightness is `dark`, the keyboard stays in `light` mode.

It's already possible to customize the keyboard appearance by passing a `SuperEditorImeConfiguration` to the editor. When a configuration isn't given, the editor uses a default `SuperEditorImeConfiguration`, which always use light keyboard appearance.

This PR makes `imeConfiguration` optional, and resolves the default configuration during build.

### DIFF
--- a/super_editor/test/super_editor/supereditor_input_ime_test.dart
+++ b/super_editor/test/super_editor/supereditor_input_ime_test.dart
@@ -1067,17 +1067,9 @@ Paragraph two
         // Holds the keyboard appearance sent to the platform.
         String? keyboardAppearance;
 
-        // Intercept the setClient message sent to the platform so we can check
-        // the keyboardAppearance that was sent.
-        tester
-            .interceptChannel(SystemChannels.textInput.name) //
-            .interceptMethod(
-          'TextInput.setClient',
-          (methodCall) {
-            final params = methodCall.arguments[1] as Map;
-            keyboardAppearance = params['keyboardAppearance'];
-            return null;
-          },
+        _interceptKeyboardAppearanceSentToPlatform(
+          tester,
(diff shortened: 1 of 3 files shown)


super_editor/test/super_editor/supereditor_input_ime_test.dart
```diff
@@ -1055,6 +1055,105 @@ Paragraph two
...
+            .withSingleEmptyParagraph()
+            .useAppTheme(ThemeData.dark())
+            .pump();
+
+        // Holds the keyboard appearance sent to the platform.
+        String? keyboardAppearance;
+
+        // Intercept the setClient message sent to the platform so we can check
+        // the keyboardAppearance that was sent.
+        tester
```
matthew-carroll, line 1072: Should this be moved to a local private method instead of repeating all this in each test?

angelosilvestre: Extracted to a private function.

matthew-carroll (approved):
LGTM

PR #1360, opened 2023-08-31
[super_editor_markdown] Add support for task syntax (Resolves #1294)

### LINKED ISSUE #1294: [SuperEditor] - Add support for task syntax in Markdown
GitHub style Markdown supports a task syntax:

- [ ] This is a task
- [x] This is a completed task

Given that `SuperEditor` now includes default support for tasks, let's add corresponding support to our Markdown parser/serializer.

angelosilvestre:
[super_editor_markdown] Add support for task syntax. Resolves #1294 

SuperEditor now includes support for tasks by default, but it doesn't support serializing/deserializing them.

This PR implements markdown serialization/deserialization for tasks using the github style syntax:

```
- [x] A completed task
- [ ] An incomplete task
```

Tasks can also be multiline, for example:

```
- [ ] A task
with multiple lines
- [ ] Another task
```
The previous input renders:

- [ ] A task
with multiple lines
- [ ] Another task

### DIFF
--- a/super_editor_markdown/lib/src/markdown_to_document_parsing.dart
+++ b/super_editor_markdown/lib/src/markdown_to_document_parsing.dart
@@ -28,6 +28,7 @@ MutableDocument deserializeMarkdownToDocument(
       if (syntax == MarkdownSyntax.superEditor) //
         _ParagraphWithAlignmentSyntax(),
       _EmptyLinePreservingParagraphSyntax(),
+      _TaskSyntax(),
     ],
   );
   final blockParser = md.BlockParser(markdownLines, markdownDoc);
@@ -148,6 +149,9 @@ class _MarkdownToDocument implements md.NodeVisitor {
         break;
       case 'hr':
         _addHorizontalRule();
+        break;
+      case 'task':
+        _addTask(element);
         break;
     }
 
@@ -282,6 +286,16 @@ class _MarkdownToDocument implements md.NodeVisitor {
         itemType: listItemType,
         indent: indent,
         text: _parseInlineText(element),
+      ),
+    );
+  }
+
+  void _addTask(md.Element element) {
+    _content.add(
+      TaskNode(
+        id: Editor.createNodeId(),
+        text: _parseInlineText(element),
+        isComplete: element.attributes['completed'] == 'x',
       ),
     );
   }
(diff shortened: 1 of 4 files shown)


super_editor_markdown/lib/src/markdown_to_document_parsing.dart
```diff
@@ -644,6 +658,57 @@ class _LineBreakSeparatedElement extends md.Element {
...
+  @override
+  md.Node? parse(md.BlockParser parser) {
+    final match = pattern.firstMatch(parser.current);
+    if (match == null) {
+      return null;
+    }
+
+    final completionToken = match.group(1)!;
+
+    // Add the first line of content to the buffer.
```
matthew-carroll, line 686: Does this mean the first line of the task description? If so, calling it the "first line of content" is a bit ambiguous. That sounds like it could mean the entire first task line.

super_editor_markdown/lib/src/markdown_to_document_parsing.dart
```diff
@@ -644,6 +658,57 @@ class _LineBreakSeparatedElement extends md.Element {
...
+    final buffer = StringBuffer(match.group(2)!);
```
matthew-carroll, line 687: Regex is really tough to read, so it's helpful to explicitly name what you're extracting.

I'd say that up above where you have `completionToken` it would be a good idea to declare a second variable called `taskDescriptionFirstLine` and set it to `match.group(2)`.

super_editor_markdown/lib/src/markdown_to_document_parsing.dart
```diff
@@ -644,6 +658,57 @@ class _LineBreakSeparatedElement extends md.Element {
...
+
+    // Consume the first line.
```
matthew-carroll, line 689: This statement is a little confusing because we just got done "consuming" the first line. How about `// Move to the second line.`

super_editor_markdown/lib/src/markdown_to_document_parsing.dart
```diff
@@ -644,6 +658,57 @@ class _LineBreakSeparatedElement extends md.Element {
...
+      buffer.write('\n');
+      buffer.write(parser.current);
+
+      parser.advance();
+    }
+
+    return md.Element(
+      'task',
+      [md.Text(buffer.toString())],
+    )..attributes['completed'] = completionToken;
```
matthew-carroll, line 708: Is this `Element` structure based on anything? If not, we should consider how a typical HTML element is likely to represent completion, which would be a boolean, not a string like "x".

super_editor_markdown/test/super_editor_markdown_test.dart
```diff
@@ -568,6 +605,16 @@ Paragraph3""");
             text: AttributedText('{\n  // This is some code.\n}'),
             metadata: {'blockType': codeAttribution},
           ),
+          TaskNode(
```
matthew-carroll, line 608: Given that the tasks test doesn't intermingle tasks with other content, we should probably do that here. Maybe add another paragraph, followed by another couple of tasks. Also, let's make the task above the paragraph a multi-line task.

super_editor_markdown/lib/src/markdown_to_document_parsing.dart
angelosilvestre: I changed to use the the boolean as a String, because the element `attributes` only accepts strings.

matthew-carroll (approved):
LGTM

PR #1364, opened 2023-09-01
Selected text can now display itself in a different color from surrounding text (Resolves #1332)

### LINKED ISSUE #1332: [SuperEditor] - Change text color within selection bounds
Text selection highlights can make it difficult to see the selected text. This is especially true in dark mode where we have a light blue highlight with white or light gray text.

It should be possible in `SuperEditor` to alter the text color for the selected text. This is a temporary state that shouldn't impact the document content. The related style decisions should be configurable via stylesheet.

matthew-carroll:
Selected text can now display itself in a different color from surrounding text (Resolves #1332)

When selected text changes color, we don't want that color change to be reflected in the document content. It's a transient visual detail that shouldn't be serialized or saved. Therefore, to accomplish this effect, we add an attribution to selected text within the presentation pipeline. This is the pipeline that applies the stylesheet, for example.

Also includes:

- `ColorAttribution` to change the color of spans of text.
- Minor adjustment to prevent an NPE related to `ContentLayers`
- A new "in the lab" demo area and scaffold. This will be a place for us to demo features that haven't stabilized yet. In a followup PR, I'm going to move the tagging demos into the lab area.

https://github.com/superlistapp/super_editor/assets/7259036/675d1ab7-4f88-41a8-a8e6-d8eb3f8dc1f8

### DIFF
--- a/super_editor/example/lib/demos/example_editor/example_editor.dart
+++ b/super_editor/example/lib/demos/example_editor/example_editor.dart
@@ -422,16 +422,12 @@ class _ExampleEditorState extends State<ExampleEditor> {
             selectionLayerLinks: _selectionLayerLinks,
             selectionStyle: isLight
                 ? SelectionStyles(
-                    // selectionColor: Color(0xFF0088F7),
-                    selectionColor: Color(0xFFACDDFF),
+                    selectionColor: Color(0xFF0088F7),
                   )
                 : SelectionStyles(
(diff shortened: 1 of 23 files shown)


matthew-carroll:
FYI @miguelcmedeiros @Jethro87 @jmatth @Dboy233

angelosilvestre (approved):
LGTM

PR #1369, opened 2023-09-02
[SuperTextField] Expose IME configuration (Resolves #1109) 

### LINKED ISSUE #1109: [SuperTextField][Mobile] - IME connection should support configuration
Flutter's IME connection accepts configuration to enable or disable autocorrection, suggestions, and other properties. 

This configuration is not currently exposed in any way by the IME text editing controller or `SuperTextField`.

Adjust the API so that users can control these configurations.

angelosilvestre:
[SuperTextField] Expose IME configuration. Resolves #1109 

Currently, we can configure the textfield's `textInputAction` but we can't configure other IME aspects, like keyboard type and autocorrection.

This PR  introduces a `SuperTextFieldImeConfiguration`, which users can use to configure the IME. As both `SuperTextFieldImeConfiguration` and `textInputAction` configures the input action, the `textInputAction` is being deprecated.

I added an `attachToImeWithConfig` method in `ImeAttributedTextEditingController` instead of changing the existing method to avoid a breaking change. If we should just change `attachToIme` then I can make this change.

### DIFF
--- a/super_editor/lib/src/super_textfield/input_method_engine/_ime_text_editing_controller.dart
+++ b/super_editor/lib/src/super_textfield/input_method_engine/_ime_text_editing_controller.dart
@@ -123,14 +123,24 @@ class ImeAttributedTextEditingController extends AttributedTextEditingController
       return;
     }
 
-    final imeConfig = TextInputConfiguration(
-      autocorrect: autocorrect,
-      enableDeltaModel: true,
+    final config = SuperTextFieldImeConfiguration(
+      enableAutocorrect: autocorrect,
       enableSuggestions: enableSuggestions,
-      inputAction: textInputAction,
-      inputType: textInputType,
-      keyboardAppearance: keyboardAppearance,
+      keyboardInputType: textInputType,
+      keyboardActionButton: textInputAction,
+      keyboardBrightness: _keyboardAppearance,
     );
+
+    attachToImeWithConfig(config);
+  }
+
+  void attachToImeWithConfig(SuperTextFieldImeConfiguration configuration) {
+    if (isAttachedToIme) {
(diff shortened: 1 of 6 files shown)


super_editor/lib/src/super_textfield/input_method_engine/_ime_text_editing_controller.dart
```diff
@@ -123,14 +123,24 @@ class ImeAttributedTextEditingController extends AttributedTextEditingController
       return;
     }
 
-    final imeConfig = TextInputConfiguration(
-      autocorrect: autocorrect,
-      enableDeltaModel: true,
+    final config = SuperTextFieldImeConfiguration(
```
matthew-carroll, line 126: I can't remember if we talked about this, but why did we need to change the type of this, and the names of parameters?

angelosilvestre: I used the same names that are already used in `SuperEditorImeConfiguration`. We don't expose `TextInputConfiguration` because we need to make sure `enableDeltaModel` is set to `true`.

We could also just take a `TextInputConfiguration` and `copyWith` to set `enableDeltaModel`.

matthew-carroll: Let's try it with the `copyWith` approach and see how it looks.

angelosilvestre: Updated.

matthew-carroll (approved):
LGTM

angelosilvestre:
@matthew-carroll Rebased to fix the failing tests.

PR #1374, opened 2023-09-04
Added substring and length APIs to AttributedText (Resolves #1200)

### LINKED ISSUE #1200: [attributed_text] - Return substrings given a SpanRange
We often work with `SpanRange`s when dealing with `AttributedText`. Currently, to get a substring, we have to access the inner plain text string, and then unpack the `SpanRange` into a start and end value.

Consider offering a plain text substring method on `AttributedText` that takes a `SpanRange` directly.

Also, consider offering a top-level `length` method on `AttributedText`.

matthew-carroll:
Added substring and length APIs to AttributedText (Resolves #1200)

### DIFF
--- a/attributed_text/test/attributed_text_test.dart
+++ b/attributed_text/test/attributed_text_test.dart
@@ -27,6 +27,82 @@ void main() {
       );
     });
 
+    group('fragments', () {
+      test('can be copied as attributed text with a SpanRange', () {
+        final text = AttributedText(
+          "this that other",
+          AttributedSpans(
+            attributions: [
+              SpanMarker(attribution: ExpectedSpans.bold, offset: 0, markerType: SpanMarkerType.start),
+              SpanMarker(attribution: ExpectedSpans.bold, offset: 6, markerType: SpanMarkerType.end),
+              SpanMarker(attribution: ExpectedSpans.italics, offset: 7, markerType: SpanMarkerType.start),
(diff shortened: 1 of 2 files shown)


attributed_text/test/attributed_text_test.dart
```diff
@@ -27,6 +27,82 @@ void main() {
...
+              SpanMarker(attribution: ExpectedSpans.bold, offset: 6, markerType: SpanMarkerType.end),
+              SpanMarker(attribution: ExpectedSpans.italics, offset: 7, markerType: SpanMarkerType.start),
+              SpanMarker(attribution: ExpectedSpans.italics, offset: 14, markerType: SpanMarkerType.end),
+            ],
+          ),
+        );
+
+        final slice = text.copyTextInRange(SpanRange(start: 5, end: 9));
+        expect(slice.text, "that");
+        expect(slice.length, 4);
```
angelosilvestre, line 46: Is this check necessary? We already compared the exact string in the line before.

angelosilvestre (commented):
LGTM with a single comment.

attributed_text/test/attributed_text_test.dart
matthew-carroll: There's an argument both ways. The argument for doing this check is that this `length` property is on the `AttributedText`, not the `String`, so it's technically possible that `length` could be wrong as compared to the `String` length.

PR #1375, opened 2023-09-04
Added collapsed span visitation in AttributedText, added tests for both versions of attribution visitors (Resolves #1215)

### LINKED ISSUE #1215: [attributed_text] - Add method that visits collapsed attributions
The visitor system currently built into `AttributedText` visits every character, reporting all open and closed markers at that character. This is a fine thing to do as a basis for processing markers and attributions. However, users might want to visit all attributions as if they're collapsed, only visiting the indices where the applied attributions change, such as when styling text.

We offer this visitation behavior in `super_editor` for styling purposes. I don't see any reason not to offer this same visitation in `attributed_text`, itself.

matthew-carroll:
Added collapsed span visitation in AttributedText, added tests for both versions of attribution visitors (Resolves #1215)


attributed_text/test/attributed_text_test.dart
```diff
@@ -277,5 +278,269 @@ void main() {
...
+              AttributedText fullText,
+              int index,
+              Set<Attribution> startingAttributions,
+              Set<Attribution> endingAttributions,
+            ) {
+              expect(_AttributionVisit(index, startingAttributions, endingAttributions), expectedVisits.first);
+              expectedVisits.removeAt(0);
+            },
+          ),
+        );
```
angelosilvestre, line 311: We could add an expect after calling `visitAttributions` to ensure `expectedVisits` is empty after the call.

Currently, if `visitAttributions` reports only the first attribution the test will still pass.

attributed_text/test/attributed_text_test.dart
```diff
@@ -277,5 +278,269 @@ void main() {
...
```
angelosilvestre, line 342: The previous comment applies to the other tests as well.

angelosilvestre (commented):
LGTM with some comments.

PR #1376, opened 2023-09-04
Added a 'range' property to AttributionSpan (Resolves #1192)

### LINKED ISSUE #1192: [attributed_text] - Return a SpanRange from an AttributionSpan
An `AttributionSpan` has a `start` and `end`, which are accessible individually. I'd like to obtain a `SpanRange` from an `AttributionSpan` instead of passing `start` and `end` into my own `SpanRange` constructor.

```dart
final span = SpanRange(start: attributionSpan.start, end: attributionSpan.end);

// vs

final span = attributionSpan.range;
```

matthew-carroll:
Added a 'range' property to AttributionSpan (Resolves #1192)

### DIFF
--- a/attributed_text/lib/src/attributed_spans.dart
+++ b/attributed_text/lib/src/attributed_spans.dart
@@ -1089,6 +1089,12 @@ class AttributionSpan {
   final int start;
   final int end;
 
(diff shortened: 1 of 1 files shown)


angelosilvestre (commented):
LGTM

PR #1391, opened 2023-09-05
[SuperEditor] Fix pasting multiple lines (Resolves #871)

### LINKED ISSUE #871: Single Line Breaks are Removed Between Paragraphs on Paste
When pasting multiple paragraphs into a document and when there is a single line break in between each paragraph, upon paste, the line breaks are removed. However, if there are two or more line breaks between paragraphs, all line breaks except one are removed.

Screenshot 2022-11-29 at 19 34 15

Illustrated above, copying and pasting the red text into a document results in empty lines between paragaphs being entirely removed.  Copying and pasting the blue text results in what would be the expected result of copying and pasting the red text, one empty line between remains between each paragraph.

Notably, if you copy the text from the same document after having pasted the blue block of text and then paste it directly back into the document, all line breaks are removed.  However, if you manually add the extra two lines between paragraphs from within the editor itself and then copy and paste the paragraphs with a single line break remaining, the line breaks are retained on subsequent pastes, but only if you add two extra lines. Adding a single empty line break between paragraphs within the editor itself and then copying and pasting that text again removes that single line breaks. I hope that's clear enough.

It appears as though when you paste paragraphs of text into the editor, it removes all line breaks between paragraphs unless you have two or more empty lines between paragraphs, in which case it removes all line breaks except one.  This feels like it may be intended behavior but it doesn't feel like standard or expected behavior for an editor. I wouldn't expect to have to advise users to add an extra line break between paragraphs.

Is there a suggested way of handling this?

angelosilvestre:
[SuperEditor] Fix pasting multiple lines. Resolves #871.

When pasting multiple lines, the editor is creating a single paragraph containing all the lines. We are requiring two line breaks to create a new paragraph.

This PR changes the paste behavior to split the paragraphs by a single line break.

This test is running on mac only because we use the mac shortcut to paste: `await tester.pressCmdV()`. We could also add methods in the robot to simulate common shortcuts depending on the platform. For example: `await tester.pressPasteCombo()`.

### DIFF
--- a/super_editor/test/super_editor/supereditor_copy_and_paste_test.dart
+++ b/super_editor/test/super_editor/supereditor_copy_and_paste_test.dart
@@ -1,3 +1,4 @@
+import 'package:flutter/foundation.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:flutter_test_robots/flutter_test_robots.dart';
 import 'package:flutter_test_runners/flutter_test_runners.dart';
@@ -58,7 +59,7 @@ void main() {
       expect(SuperEditorInspector.findTextInParagraph(nodeId).text, "Pasted text: This was pasted here");
     });
 
-    testAllInputsOnMac('pastes multiple paragraphs', (
+    testAllInputsOnDesktop('pastes multiple paragraphs', (
       tester, {
       required TextInputSource inputSource,
     }) async {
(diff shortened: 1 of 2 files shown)


super_editor/test/super_editor/supereditor_copy_and_paste_test.dart
```diff
@@ -56,5 +57,34 @@ void main() {
       final nodeId = doc.nodes.first.id;
       expect(SuperEditorInspector.findTextInParagraph(nodeId).text, "Pasted text: This was pasted here");
     });
+
+    testAllInputsOnMac('pastes multiple paragraphs', (
```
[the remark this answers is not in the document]
angelosilvestre: Added tests running on linux and windows instead

super_editor/test/super_editor/supereditor_copy_and_paste_test.dart
```diff
@@ -56,5 +57,63 @@ void main() {
...
+      // Place the caret at the empty paragraph.
+      await tester.placeCaretInParagraph('1', 0);
+
+      // Simulate pasting multiple lines.
+      tester
+        ..simulateClipboard()
+        ..setSimulatedClipboardContent('''This is a paragraph
+This is a second paragraph
+This is the third paragraph''');
+      await tester.pressCtlV();
```
matthew-carroll, line 109: How about we write one test and we select the shortcut based on the platform?

angelosilvestre: Ok, I usually try to avoid conditions in tests, but I think for this case it won't hurt.

angelosilvestre: Updated.

matthew-carroll (approved):
LGTM

PR #1401, opened 2023-09-05
[attributed_text] Fix exception in hasAttributionsWithin (Resolves #1201)

### LINKED ISSUE #1201: [attributed_text] AttributedSpans.hasAttributionsWithin incorrectly throws as it incorrectly iterates over a Set of attributes

angelosilvestre:
[attributed_text] Fix exception in hasAttributionsWithin. Resolves #1201.

This PR is built on top of https://github.com/superlistapp/super_editor/pull/1202

`hasAttributionsWithin` incorrectly removes attributions while iterating over a `Set`. This causes an exception when looking for multiple attributions at the same time.

This PR changes the method to only remove the attributions after iterating.

### DIFF
--- a/attributed_text/lib/src/attributed_spans.dart
+++ b/attributed_text/lib/src/attributed_spans.dart
@@ -65,14 +65,18 @@ class AttributedSpans {
   }) {
     final attributionsToFind = Set.from(attributions);
     for (int i = start; i <= end; ++i) {
+      final foundAttributions = <Attribution>{};
       for (final attribution in attributionsToFind) {
         if (hasAttributionAt(i, attribution: attribution)) {
-          attributionsToFind.remove(attribution);
+          // Add the attribution to be removed later.
+          // Removing from the set while iterating throws an exception.
+          foundAttributions.add(attribution);
         }
-
-        if (attributionsToFind.isEmpty) {
-          return true;
-        }
+      }
+      attributionsToFind.removeAll(foundAttributions);
+      if (attributionsToFind.isEmpty) {
+        // We found all attributions.
+        return true;
       }
     }
     return false;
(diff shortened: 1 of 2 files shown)


attributed_text/lib/src/attributed_spans.dart
```diff
@@ -65,15 +65,19 @@ class AttributedSpans {
...
     for (int i = start; i <= end; ++i) {
+      final foundAttributions = <Attribution>{};
       for (final attribution in attributionsToFind) {
         if (hasAttributionAt(i, attribution: attribution)) {
-          attributionsToFind.remove(attribution);
-        }
-
-        if (attributionsToFind.isEmpty) {
-          return true;
+          // Add the attribution to be removed later.
```
matthew-carroll, line 71: This comment doesn't make much sense to me. Can we get a better explanation?

attributed_text/test/attributed_spans_test.dart
```diff
@@ -157,6 +157,52 @@ void main() {
...
+            true);
+
+        expect(
+          spans.hasAttributionsWithin(attributions: {
+            ExpectedSpans.bold,
+            ExpectedSpans.strikethrough,
+          }, start: 0, end: 4),
+          false,
+        );
+      });
```
matthew-carroll, line 205: Please add a blank line after the test.

attributed_text/lib/src/attributed_spans.dart
angelosilvestre: Updated. Please take a look if it's more clear now.

matthew-carroll (approved):
LGTM

PR #1406, opened 2023-09-07
Fixed tagging NPEs - includes new ContentLayers layer widgets, IME focus policy adjustment (Resolves #1397)

### LINKED ISSUE #1397: [SuperEditor] - Tagging is broken with NPEs (again)
All tagging appears to be broken due to another resurgence of NPEs related to document layout access through `ContentLayers`. We've tried to address the `ContentLayers` issue a few times:

https://github.com/superlistapp/super_editor/pull/1354

https://github.com/superlistapp/super_editor/pull/1357

https://github.com/superlistapp/super_editor/pull/1364

Hash tags: errors generated on every character inserted into a hash tag (but not when typing regular characters)

User tags: throws error on every character typed, even when not in a tag. Also, after typing trigger, can't type any more characters. The character key seems to go unhandled - it generates an OS sound.

Action tags: Same problems as user tags.

matthew-carroll:
Fixed tagging NPEs - includes new ContentLayers layer widgets, IME focus policy adjustment (Resolves #1397)

The tangible goal of this PR is to fix issues found with the tagging system in the Example app. Before this PR, the tagging examples were producing NPEs on every character the user entered.

To fix the tagging experience, a few changes were made:

## Content Layers
Despite my best attempts to hack the Flutter build order, every time the user inserted a new character into a tag, it would trigger a build timing problem where the caret overlay would try to inspect the document layout before the document layout was run.

I was unable to further hack the Flutter build order to resolve this. At this point, I had to bring some level of new responsibility to the widgets that are displayed as layers.

As of this PR, any widget that's provided as a layer in `ContentLayers` must be a subclass of `ContentLayerWidget`. I've created a few options to facilitate different use-cases:

- `ContentLayerProxyWidget` - Let's users display regular widgets that don't care about the content layout. This is essentially an escape hatch for when you don't care about any of the rest of this.
- `ContentLayerStatelessWidget` - A stateless widget that wants to base its widget tree on the content layout render object.
- `ContentLayerStatefulWidget` - A stateful widget that wants to base its widget tree on the content layout render object. 

## IME and Focus
User tag and action tag demos display a popover, where the user can press up/down to select a tag. As a result, the popover needs to hold primary focus, meaning `SuperEditor` has non-primary focus. Previously, we were automatically closing the IME when `SuperEditor` loses primary focus. This focus policy breaks expected behavior because the user can't type anything once the popover appears. Therefore, this PR changes the default focus policy so that `SuperEditor` keeps the IME open as long as it has focus, even if it's not primary focus.

Also related to focus is Mac OS selectors. When `SuperEditor` moves to non-primary focus, and some other part of the UI takes primary focus, like a popover, Flutter jumps in and responds to certain keys, like arrow keys. For example, Flutter tries to move focus to other widgets when the user presses arrow keys. We don't want this. We want arrow keys to either move selection in the popover, or move the caret in `SuperEditor`. To prevent Flutter from taking over, when `SuperEditor` has non-primary focus, we no longer ignore the key press on Mac. Instead, we respond to all key presses on Mac when `SuperEditor` has non-primary focus so that we can block Flutter from doing things we don't want.

### DIFF
--- a/super_editor/lib/src/infrastructure/selection_leader_document_layer.dart
+++ b/super_editor/lib/src/infrastructure/selection_leader_document_layer.dart
@@ -5,6 +5,7 @@ import 'package:follow_the_leader/follow_the_leader.dart';
 import 'package:super_editor/src/core/document.dart';
 import 'package:super_editor/src/core/document_layout.dart';
 import 'package:super_editor/src/core/document_selection.dart';
+import 'package:super_editor/src/infrastructure/content_layers.dart';
 
 /// A document layer that positions leader widgets at the user's selection bounds.
 ///
@@ -14,7 +15,7 @@ import 'package:super_editor/src/core/document_selection.dart';
 /// on the document's self-reported caret height for the given document position.
 ///
 /// When no selection exists, no leaders are built in the layer's widget tree.
-class SelectionLeadersDocumentLayer extends StatefulWidget {
+class SelectionLeadersDocumentLayer extends ContentLayerStatefulWidget {
   const SelectionLeadersDocumentLayer({
     Key? key,
     required this.document,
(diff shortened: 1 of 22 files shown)


matthew-carroll:
CC @miguelcmedeiros @Jethro87 @jmatth - This PR will break any custom content layers/document layers. The fix should be pretty straightforward - it requires using a custom version of stateless and/or stateful widgets. Within those widgets you can include traditional widget trees.

There are also some minor focus policy changes that could conceivably impact you.

super_editor/lib/src/infrastructure/selection_leader_document_layer.dart
```diff
@@ -194,6 +193,20 @@ class _SelectionLeadersDocumentLayerState extends State<SelectionLeadersDocument
   }
 }
 
+class SelectionLeaderLayout {
```
angelosilvestre, line 196: Having "Layout" in the class name can be a bit confusing as this class just holds values.

angelosilvestre (commented):
I left just a couple comments.

PR #1413, opened 2023-09-07
Adjusted attributed_text tests to avoid false negatives (Resolves #1392)

### LINKED ISSUE #1392: [attributed_text] - Adjust conditions in a few tests
`attributed_text` has a few tests that visit attributions. Those tests aren't wrong, but they're incomplete. It's possible that the tested behavior could fail in a way that wouldn't trigger test failures.

Update the tests to comply with comments in this previous PR:
https://github.com/superlistapp/super_editor/pull/1375#pullrequestreview-1609976738

matthew-carroll:
Adjusted attributed_text tests to avoid false negatives (Resolves #1392)

### DIFF
--- a/attributed_text/test/attributed_text_test.dart
+++ b/attributed_text/test/attributed_text_test.dart
@@ -460,6 +460,8 @@ void main() {
             },
           ),
         );
+
+        expect(expectedVisits, isEmpty);
       });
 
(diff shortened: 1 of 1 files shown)


angelosilvestre (commented):
LGTM

PR #1415, opened 2023-09-08
Added convenience APIs for document selections within a single TextNode (Resolves #1195)

### LINKED ISSUE #1195: [super_editor] - Make it easy to generate a DocumentSelection and DocumentRange within a TextNode
There are various operations that operate against text within a `TextNode`, such as reactions that modify text segments within a single paragraph. These operations need to pass `DocumentSelection`s or `DocumentRange`s to editor requests.

It's tedious and verbose to construct these document selections. We should offer an easier way to create selections and ranges within a single text node.

```dart
final selection = DocumentSelection(
  base: DocumentPosition(
   nodeId: textNode.id,
   nodePosition: TextNodePosition(offset: 5),
  ),
  extent: DocumentPosition(
    nodeId: textNode.id,
    nodePosition: TextNodePosition(offset: 10),
  ),
);

// vs something like

final selection = textNode.selectionWithin(5, 10);
final range = textNode.rangeWithin(5, 10);
```

matthew-carroll:
Added convenience APIs for document selections within a single TextNode (Resolves #1195)

### DIFF
--- a/super_editor/lib/src/default_editor/common_editor_operations.dart
+++ b/super_editor/lib/src/default_editor/common_editor_operations.dart
@@ -190,15 +190,9 @@ class CommonEditorOperations {
 
     editor.execute([
       ChangeSelectionRequest(
-        DocumentSelection(
-          base: DocumentPosition(
-            nodeId: selectedNode.id,
-            nodePosition: wordNodeSelection.base,
-          ),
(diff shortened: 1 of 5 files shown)


angelosilvestre (commented):
LGTM

PR #1418, opened 2023-09-08
[SuperEditor][SuperTextField][Web] Fix IME panels positioning (Resolves #1356)

### LINKED ISSUE #1356: [Web] Japenese & Emoji IME popovers in wrong position
**Reproduction**

  1. Launch Super Editor Example on super_editor main branch / flutter master branch as a MacOS Desktop App
  3. Run main demo or super text field demo
  4. On MacOS, Open emoji popover. Popover appears in correct position
  5. On MacOS, enable Japanese Romanji Keyboard.
  6. From the Keyboard Selection menu, choose "Hiragana - Romanji"
  7. Type the letters "sa"
  8. Japanese IME popver should appear in correct position
  9. Launch example app as a Chrome Web App
  10. Repeat Steps 4-8.
  11. Emoji & Japanese IME popovers appear in the wrong part of the screen.

**macOS (working)**

https://github.com/superlistapp/super_editor/assets/90610075/4fbf6ea0-477e-47ca-ae2d-6e3228bf9ad4

**Chrome (not working)**

https://github.com/superlistapp/super_editor/assets/90610075/b1cfcef6-ea80-4cd6-8cff-bc8618c5789f

angelosilvestre:
[SuperEditor][SuperTextField][Web] Fix IME panels positioning. Resolves #1356

On web, IME panels like the emoji picker, accent panel or japanese input panel aren't displayed at the correct position.

The root cause is that calling `setCaretRect` does nothing on web. To work around that, we need two changes:

- Report the current text style to the IME
- Change `SuperEditor`'s size and transform reporting to report only the bounds of the selected node's visual component instead of the bounds of the whole editor

This PR changes the IME interactors so that, on web, we avoid calling `setCaretRect` and send the the selected component size and transform to the IME. 

This PR introduces a `TextInputComponent` interface to provide methods and properties to query the required information that we need to send to the IME. I'd prefer to avoid declaring this interface and just adding those methods to `TextComponentState`, but it wouldn't work for the task component as it doesn't bind the component key to the inner `TextComponent`. If we can change that, this new interface can be removed.

https://github.com/superlistapp/super_editor/assets/7597082/1801c9d2-d11d-4da3-a6f6-715ac1839677

https://github.com/superlistapp/super_editor/assets/7597082/96bdf4d2-35db-4afe-90b6-2a503ac4a3d3

https://github.com/superlistapp/super_editor/assets/7597082/f04839c0-ddb3-4fdf-8683-940bfac8fa3d


matthew-carroll:
@angelosilvestre you said that the task component doesn't link to its inner text component, but it does. The Task state object exposes both a `childDocumentComponentKey` and `childTextComposable`.

super_editor/lib/src/default_editor/document_hardware_keyboard/document_keyboard_actions.dart
```diff
@@ -517,6 +517,14 @@ ExecutionInstruction moveUpAndDownWithArrowKeys({
     return ExecutionInstruction.continueExecution;
   }
 
+  if (isWeb && (editContext.hasPrimaryFocus.value) && (editContext.composer.composingRegion.value != null)) {
+    // We are composing a character on web. It's possible that a native element is being displayed,
+    // like an emoji picker or a character selection panel.
+    // We need to let the OS handle the key so the user can navigate
+    // on the list of possible characters.
+    return ExecutionInstruction.blocked;
```
matthew-carroll, line 525: Please include links to any associated Flutter issue ticket related to this and mention that ideally we wouldn't have to run this check.

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
```diff
@@ -231,11 +242,18 @@ class SuperEditorImeInteractorState extends State<SuperEditorImeInteractor> impl
...
-      _configureImeClientDecorators();
-      _documentImeConnection.value = _documentImeClient;
-      _reportVisualInformationToIme();
+      return;
+    }
+
+    _configureImeClientDecorators();
+    _documentImeConnection.value = _documentImeClient;
+
+    if (!_imeConnection.value!.attached) {
```
matthew-carroll, line 251: What does it mean to not be attached here? Why would this happen?

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
```diff
@@ -257,13 +275,8 @@ class SuperEditorImeInteractorState extends State<SuperEditorImeInteractor> impl
...
 
-    final renderBox = context.findRenderObject() as RenderBox;
-    _imeConnection.value!.setEditableSizeAndTransform(renderBox.size, renderBox.getTransformTo(null));
-
-    final caretRect = _computeCaretRectInViewportSpace();
-    if (caretRect != null) {
-      _imeConnection.value!.setCaretRect(caretRect);
-    }
+    _reportSizeAndTransformToIme();
+    _reportCaretRectToIme();
```
matthew-carroll, line 279: Is there a reason that we don't report the style information here, too? Isn't the style info also "visual information" that the IME needs?

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
```diff
@@ -273,6 +286,104 @@ class SuperEditorImeInteractorState extends State<SuperEditorImeInteractor> impl
     onNextFrame((_) => _reportVisualInformationToIme());
   }
 
+  /// Report our size and transform to the root node coordinates to the IME.
```
matthew-carroll, line 289: This first line phrasing is confusing.

"Report the global size and transform of the editor to the IME"

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
```diff
@@ -273,6 +286,104 @@ class SuperEditorImeInteractorState extends State<SuperEditorImeInteractor> impl
...
+  ///
+  /// This is needed to display the OS emoji & symbols panel at the editor selected position.
+  void _reportSizeAndTransformToIme() {
+    //print('reporting size and transform');
```
matthew-carroll, line 293: Please remove `print`

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
```diff
@@ -273,6 +286,104 @@ class SuperEditorImeInteractorState extends State<SuperEditorImeInteractor> impl
...
+    late Size size;
+    late Matrix4 transform;
+
+    if (isWeb) {
+      // On web, we can't set the caret rect.
+      // To display the IME panels at the correct position,
+      // instead of reporting the whole editor size and transform,
+      // we report only the information about the selected node.
+      final sizeAndTransform = _computeSizeAndTransformOfSelectNode();
+      if (sizeAndTransform == null) {
```
matthew-carroll, line 303: Why might this happen?

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
```diff
@@ -273,6 +286,104 @@ class SuperEditorImeInteractorState extends State<SuperEditorImeInteractor> impl
...
+        return;
+      }
+
+      size = sizeAndTransform.$1;
```
matthew-carroll, line 307: Please always name properties. I don't know why Dart thought that referring to things as `$1` and `$2` was an acceptable UX decision, but it's awful.

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
```diff
@@ -273,6 +286,104 @@ class SuperEditorImeInteractorState extends State<SuperEditorImeInteractor> impl
...
+      size = renderBox.size;
+      transform = renderBox.getTransformTo(null);
+    }
+
+    _imeConnection.value!.setEditableSizeAndTransform(size, transform);
+  }
+
+  void _reportCaretRectToIme() {
+    if (isWeb) {
+      // On web, setting the caret rect isn't supported.
```
matthew-carroll, line 321: Rather than say that something isn't supported, please tell the reader what conclusion to draw. For example, does this comment mean that all popovers are misaligned on web because "setting the caret rect isn't supported"?

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
```diff
@@ -273,6 +286,104 @@ class SuperEditorImeInteractorState extends State<SuperEditorImeInteractor> impl
...
+      return;
+    }
+
+    final caretRect = _computeCaretRectInViewportSpace();
```
matthew-carroll, line 325: I see that we do this in viewport space. Can you check what happens when SuperEditor isn't aligned to the top-left of the window? Do we still report the correct offset?

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
```diff
@@ -273,6 +286,104 @@ class SuperEditorImeInteractorState extends State<SuperEditorImeInteractor> impl
...
+      _imeConnection.value!.setCaretRect(caretRect);
+    }
+  }
+
+  /// Report our text style to the IME.
+  ///
+  /// This is used on web to set the text style of the hidden native input,
+  /// to try to match the text size on the browser with our text size.
+  ///
+  /// As our content can have multiple styles, the sizes won't be 100% in sync.
```
matthew-carroll, line 336: Please include link to Flutter issue where we're trying to get this to work for our multi-style editor experience

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
```diff
@@ -273,6 +286,104 @@ class SuperEditorImeInteractorState extends State<SuperEditorImeInteractor> impl
...
+
+    final nodePosition = selection.extent.nodePosition;
+    if (nodePosition is! TextNodePosition) {
+      // The selected component doesn't contain text.
+      return;
+    }
+
+    final docLayout = widget.editContext.documentLayout;
+    final selectedComponent = docLayout.getComponentByNodeId(selection.extent.nodeId);
+    if (selectedComponent == null) {
```
matthew-carroll, line 357: Why might this happen?

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
```diff
@@ -273,6 +286,104 @@ class SuperEditorImeInteractorState extends State<SuperEditorImeInteractor> impl
...
+    _imeConnection.value!.setStyle(
+      fontFamily: style.fontFamily,
+      fontSize: style.fontSize,
+      fontWeight: style.fontWeight,
+      textDirection: textInputComponent.textDirection ?? TextDirection.ltr,
+      textAlign: textInputComponent.textAlign ?? TextAlign.left,
+    );
+  }
+
+  void _onSelectionChanged() {
```
matthew-carroll, line 377: Please put this above all these reporting methods.

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
```diff
@@ -273,6 +286,104 @@ class SuperEditorImeInteractorState extends State<SuperEditorImeInteractor> impl
...
+    if (!isWeb) {
+      return;
+    }
+
+    // Selection has changed.
+    // Report the text style of the current selection to the IME.
```
matthew-carroll, line 383: Please mention why this needs to happen on the next frame.

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
```diff
@@ -302,6 +413,32 @@ class SuperEditorImeInteractorState extends State<SuperEditorImeInteractor> impl
     return caretOffset & rectInDocLayoutSpace.size;
   }
 
+  /// Compute the size and transform of the selected node's visual component
```
matthew-carroll, line 416: There's probably no need to ever talk about "root node coordinates". I think what you're talking about are "global coordinates".

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
```diff
@@ -302,6 +413,32 @@ class SuperEditorImeInteractorState extends State<SuperEditorImeInteractor> impl
...
+  /// to the root node coordinates.
+  (Size size, Matrix4 transform)? _computeSizeAndTransformOfSelectNode() {
+    final selection = widget.editContext.composer.selection;
+    if (selection == null) {
+      return null;
+    }
+
+    final docLayout = widget.editContext.documentLayout;
```
matthew-carroll, line 424: Let's call this `documentLayout`, matching the thing you're accessing.

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
```diff
@@ -302,6 +413,32 @@ class SuperEditorImeInteractorState extends State<SuperEditorImeInteractor> impl
...
+    final selectedComponent = docLayout.getComponentByNodeId(selection.extent.nodeId);
+    if (selectedComponent == null) {
```
matthew-carroll, line 426: Why might this happen?

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
```diff
@@ -302,6 +413,32 @@ class SuperEditorImeInteractorState extends State<SuperEditorImeInteractor> impl
...
+      return null;
+    }
+
+    if (selectedComponent is! TextInputComponent) {
+      // The selected doesn't report text bounds.
+      return null;
+    }
+
+    final rect = (selectedComponent as TextInputComponent).getTextBounds();
```
matthew-carroll, line 435: Is there a reason that the text component doesn't report its global transform? I know we talked briefly about transforms, but I thought render objects could find their global transform?

super_editor/lib/src/default_editor/tasks.dart
```diff
@@ -230,6 +232,38 @@ class _TaskComponentState extends State<TaskComponent> with ProxyDocumentCompone
...
+    return _computeStyles(attributions);
+  }
+
+  @override
+  TextAlign? get textAlign => TextAlign.left;
+
+  @override
+  TextDirection? get textDirection => Directionality.of(context);
+
+  TextStyle _computeStyles(Set<Attribution> attributions) {
```
matthew-carroll, line 255: Please add Dart Doc so it's clear which portion of text these styles apply to, or how it's expected to be used.

super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -1016,6 +1022,8 @@ class SuperTextFieldImeInteractor extends StatefulWidget {
   /// the [ProseTextLayout] widget that paints the text for this text field.
   final GlobalKey<ProseTextState> textKey;
 
+  final AttributionStyleBuilder textStyleBuilder;
```
matthew-carroll, line 1025: Please add all appropriate Dart Docs

super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -997,9 +1000,12 @@ class SuperTextFieldImeInteractor extends StatefulWidget {
     required this.focusNode,
     required this.textController,
     required this.isMultiline,
+    required this.textStyleBuilder,
```
matthew-carroll, line 1003: Let's put this down with the other text style properties

super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -1128,6 +1143,15 @@ class _SuperTextFieldImeInteractorState extends State<SuperTextFieldImeInteracto
     }
   }
 
+  void _onImeConnectionChanged() {
+    if (!widget.textController.isAttachedToIme) {
+      return;
+    }
+
+    _reportTextStyleToIme();
```
matthew-carroll, line 1151: Here again, is there a reason that these two reportings are separated?

super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -1157,6 +1175,58 @@ class _SuperTextFieldImeInteractorState extends State<SuperTextFieldImeInteracto
     onNextFrame((_) => _reportVisualInformationToIme());
   }
 
+  void _reportSizeAndTransformToIme() {
```
matthew-carroll, line 1178: Please add Dart Doc

super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -1157,6 +1175,58 @@ class _SuperTextFieldImeInteractorState extends State<SuperTextFieldImeInteracto
...
+
+    final selection = widget.textController.selection;
+    if (selection.isValid) {
+      // We have a selection, compute the style based on the attributions present
+      // at the selection extent.
+      final text = widget.textController.text;
+      final attributions = text.getAllAttributionsAt(selection.extentOffset);
+      textStyle = widget.textStyleBuilder(attributions);
+    } else {
+      // We don't have a selection, use the default style.
```
matthew-carroll, line 1217: If we don't have a selection, should we be reporting anything?

super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -1789,6 +1859,15 @@ class DefaultSuperTextFieldKeyboardHandlers {
     if (!arrowKeys.contains(keyEvent.logicalKey)) {
       return TextFieldKeyboardHandlerResult.notHandled;
     }
+
+    if (isWeb && (controller.composingRegion.isValid)) {
+      // We are composing a character on web. It's possible that a native element is being displayed,
+      // like an emoji picker or a character selection panel.
+      // We need to let the OS handle the key so the user can navigate
+      // on the list of possible characters.
```
matthew-carroll, line 1867: Same comment here as super editor, please reference a Flutter issue and explain that we'd like to get rid of this.

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
angelosilvestre: `_onImeConnectionChange` is also called when the we detach from the IME, so we shouldn't try to report this information.

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
angelosilvestre: Updated.

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
angelosilvestre: Updated.

super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
angelosilvestre: Updated.

super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
angelosilvestre: Probably not. updated.

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
angelosilvestre: Added a dart doc to `_computeSizeAndTransformOfSelectNode` explaining why it might return null.

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
angelosilvestre: As `getComponentByNodeId` can return null, I added this guard clause. Should I just use a null assertion? 

```dart
final selectedComponent = docLayout.getComponentByNodeId(selection.extent.nodeId)!;
```

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
angelosilvestre: Same as before.

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
angelosilvestre: I tested the sliver demo and it seems we are reporting the correct offset.

angelosilvestre:
@matthew-carroll 

> The Task state object exposes both a childDocumentComponentKey and childTextComposable

The `childDocumentComponentKey` can be bound to an arbitrary component, not just a text component. It's true that in the majority of cases it will be a text component though. If we want to use `childDocumentComponentKey` then we need to check for a `ProxyDocumentComponent` and hard cast its child component to a `TextComponent`.

Please let me know what do you think.

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
angelosilvestre: I will come back for this comment after we discuss https://github.com/superlistapp/super_editor/pull/1418#issuecomment-1714731423

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
matthew-carroll: Ok. In general when I ask about why a condition might happen, it means that the code should clarify why it can happen. Please let the reader the know why we might not be attached to the IME at that point.

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
matthew-carroll: Can you try using a non-sliver SuperEditor in a location that isn't aligned with the top left? Slivers might do all sorts of weird things when it comes to geometry reporting.

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
matthew-carroll: At a minimum we should add a warning log in the if-statement that explains what went wrong. An if-statement like this, without a comment, and without a log, suggests that such a situation is expected and acceptable.

matthew-carroll:
For the case of a proxy component, how about we check if we're dealing with a proxy component. If we are, get the wrapped component, and then check if it's a text component, and then everything else continues as before.

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
angelosilvestre: Updated.

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
angelosilvestre: Updated.

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
angelosilvestre: I tested and it seems everything is correct.

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
matthew-carroll: This comment is still confusing. Why would the IME connection be closed and yet `_imeConnection.value` is non-null?

super_editor/lib/src/default_editor/text.dart
```diff
@@ -834,6 +834,17 @@ class TextComponentState extends State<TextComponent> with DocumentComponent imp
     );
   }
 
+  /// Return the [TextStyle] for the character at [offset].
```
matthew-carroll, line 837: Please also describe what this returns when the caret is at the beginning of the text, the end of the text, or when there is no text.

super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -1000,6 +1003,9 @@ class SuperTextFieldImeInteractor extends StatefulWidget {
     required this.selectorHandlers,
     this.textInputAction,
     this.imeConfiguration,
+    this.textAlign,
+    this.textDirection,
+    required this.textStyleBuilder,
```
matthew-carroll, line 1008: Let's place this above `textAlign`

super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -1016,6 +1022,10 @@ class SuperTextFieldImeInteractor extends StatefulWidget {
   /// the [ProseTextLayout] widget that paints the text for this text field.
   final GlobalKey<ProseTextState> textKey;
 
+  /// Text style factory that creates styles for the content in
```
matthew-carroll, line 1025: I'm not sure this Dart Doc covers relevant information....

The actual text styling is done somewhere else right? We have this IME interactor, which otherwise shouldn't know anything about text styling, and yet we have these properties. A user is going to be confused by that, right? Also, I assume that we expect the user to pass us the exact same styler that's used elsewhere? If so, these are the kinds of details that we need to capture.

Similar to some other recent Dart Docs in PRs, putting yourself in the shoes of a user, what do you need to know about this property to use it correctly? It doesn't seem to me that we've captured that.

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
angelosilvestre: I checked the places where we close the IME connection and we always set `_imeConnection.value` to `null`, so it seems we don't need this if statement.

super_editor/lib/src/default_editor/text.dart
angelosilvestre: Updated. I also adjusted the case where the selection is at the end of the text.

super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
angelosilvestre: Updated.

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
matthew-carroll: Won't this if-statement be triggered if the user selects a proxy component that doesn't include text? If so, should that really result in a warning?

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
```diff
@@ -302,6 +398,35 @@ class SuperEditorImeInteractorState extends State<SuperEditorImeInteractor> impl
     return caretOffset & rectInDocLayoutSpace.size;
   }
 
+  /// Compute the size and transform of the selected node's visual component
+  /// to the global coordinates.
+  ///
+  /// Returns `null` if the we don't have a selection, or if the selected
+  /// component can't report its inner text bounds.
```
matthew-carroll, line 405: Why wouldn't it be able to report its inner text bounds?

super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -1028,6 +1046,10 @@ class SuperTextFieldImeInteractor extends StatefulWidget {
   /// Preferences for how the platform IME should look and behave during editing.
   final TextInputConfiguration? imeConfiguration;
 
+  final TextAlign? textAlign;
```
matthew-carroll, line 1049: Please ensure that the order of properties matches the order of named parameters in the constructor.

angelosilvestre: Updated.

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
angelosilvestre: Updated.

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
angelosilvestre: No, it will only trigger if `selectedComponent.childDocumentComponentKey.currentState` returns `null`. It doesn't matter if it has text or not.

super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
matthew-carroll: Looks like this is still an issue?

super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
matthew-carroll: I was expecting that you'd move the style builder down here above the related text style proeprties, rather than alter the constructor to separate the properties. Please keep highly related properties together.

angelosilvestre: I misunderstood the comment. Sorry.

angelosilvestre: Updated.

matthew-carroll (approved):
LGTM

PR #1419, opened 2023-09-08
Replace selections with ranges in relevant requests and commands (Resolves #1194)

### LINKED ISSUE #1194: [super_editor] - Requests with ranges should take a DocumentRange
Editor requests, such as `AddAttributionRequest` and `RemoveAttributionRequest` should take `DocumentRange`s instead of `DocumentSelection`s. 

First, these requests have no implied affinity, so a range is semantically correct, and a selection isn't.

Second, it's easy to confuse users when referring to a "selection" because it makes it seem as if the user's selection needs to be adjusted to execute the behavior, which it doesn't.

matthew-carroll:
Replace selections with ranges in relevant requests and commands (Resolves #1194)


angelosilvestre (approved):
LGTM

PR #1420, opened 2023-09-08
Fixed broken document switching demo (Resolves #1410)

### LINKED ISSUE #1410: [SuperEditor] - Can't switch between different documents
In the Example app, the document switching demo is broken.

There are two buttons, one for each document. The user is supposed to be able to tap "Document 2" to switch to a different document. But pressing that document has no visual effect.

rutvik110:
Fixed broken document switching demo (Resolves #1410)

Pressing document buttons now updates the active document to the selected document in the editor.

https://github.com/superlistapp/super_editor/assets/65209850/55a34011-4b33-432b-b3ac-da6ae5a66453


rutvik110:
Beside this, I happen to come across an issue rn where switching between documents breaks the selection on the second doc.
I wasn't able to find any issues related to this behaviour. Is this something you're aware of?

https://github.com/superlistapp/super_editor/assets/65209850/333a40d2-0910-4ed4-9751-4a0b6c1f31b1

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -362,6 +362,10 @@ class SuperEditorState extends State<SuperEditor> {
   void didUpdateWidget(SuperEditor oldWidget) {
     super.didUpdateWidget(oldWidget);
 
+    if (widget.composer != oldWidget.composer) {
```
rutvik110, line 365: > Beside this, I happen to come across an issue rn where switching between documents breaks the selection on the second doc. I can neither select or edit the second documents content in this case. I wasn't able to find any issues related to this behaviour. Is this something you're aware of?
> 
>  Screen.Recording.2023-09-08.at.9.03.00.AM.mov

Was able to find to the root cause of this issue. The `_composer` created within `SuperEditor` wasn't being updated whenever the `composer` passed down to the `SuperEditor` is changed. 
This caused the second document to be non-interactive as we were still using the `composer` from the old document. 

With this change, you can see that the document switching works properly and the document editing behaves as usual.

https://github.com/superlistapp/super_editor/assets/65209850/233a7331-5007-49d4-a108-d2f303a7cd6a

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
```diff
@@ -191,6 +175,12 @@ class SuperEditorImeInteractorState extends State<SuperEditorImeInteractor> impl
   void didUpdateWidget(SuperEditorImeInteractor oldWidget) {
     super.didUpdateWidget(oldWidget);
 
+    if (widget.editContext != oldWidget.editContext) {
```
rutvik110, line 178: Also had to update the `_textInputConfiguration` and the `_documentImeClient` whenever the `editContext` changed as they depend on it.  Without this, text input events were not being handled within the switched document.

matthew-carroll:
@rutvik110 when putting up a PR about a bug fix, please be sure to describe in the PR description what the problem was, and how it was solved.

super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
```diff
@@ -153,31 +153,15 @@ class SuperEditorImeInteractorState extends State<SuperEditorImeInteractor> impl
...
-    _documentImeClient = DocumentImeInputClient(
-      selection: widget.editContext.composer.selectionNotifier,
-      composingRegion: widget.editContext.composer.composingRegion,
-      textDeltasDocumentEditor: _textDeltasDocumentEditor,
-      imeConnection: _imeConnection,
-      floatingCursorController: widget.floatingCursorController,
-      onPerformSelector: _onPerformSelector,
-    );
+    _createTextDeltasDocumentEditor();
+    _createDocumentImeClient();
```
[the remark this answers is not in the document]
rutvik110: Updated.

rutvik110:
@matthew-carroll Made the requested updates. And I believe we wanna lock down this expected behaviour when switching between docs, right? Should I get onto adding tests related to that?

matthew-carroll:
@rutvik110 yeah a test would be good

rutvik110:
@matthew-carroll What kind of tests do we want for these updates? 
As this update doesn't alter any behaviours specifically, then I assume we may only wanna check whether we are using proper composer and IME connection is reset every time we switch the document. 
Other than that, anything specific in your mind that should be tested as well?

matthew-carroll:
The test would do pretty much what the demo does. Display one editor, then tap a button, and switch to a different editor. You also demonstrated selection retention. So it looks like you can confirm that the desired document is displayed visually, and you can confirm that different selections are maintained when switching, and you can confirm that after switching you're able to interact with the editor to change the selection. All the stuff that you focused on as a person in this issue and PR are things that we can automatically verify in a test.

super_editor/test/super_editor/supereditor_demo_switch_document_content_test.dart
```diff
@@ -0,0 +1,301 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:flutter_test_robots/flutter_test_robots.dart';
+import 'package:flutter_test_runners/flutter_test_runners.dart';
+import 'package:super_editor/super_editor.dart';
+import 'package:super_editor/super_editor_test.dart';
+
```
matthew-carroll, line 7: This is a rather specific test to have it sown file. Did you look for any reasonable existing test file?

super_editor/test/super_editor/supereditor_demo_switch_document_content_test.dart
```diff
@@ -0,0 +1,301 @@
...
+void main() {
+  group("SuperEditor document content switching", () {
+    testWidgetsOnAllPlatforms("active document content is visible", (tester) async {
```
matthew-carroll, line 10: This test description doesn't make a lot of sense when combined with the group name. Also, this test description doesn't really make it clear what it's testing. What's an "active document" and "is visible" when? What's the behavior under test here?

super_editor/test/super_editor/supereditor_demo_switch_document_content_test.dart
```diff
@@ -0,0 +1,301 @@
...
+      // Switch to the second document.
+      await tester.tap(find.byKey(const ValueKey("Document2")));
+      await tester.pump();
+
+      // Ensure that the second documents content is visible.
+      expect(SuperEditorInspector.findWidgetForComponent<TextComponent>("Document2_Header"), isNotNull);
+      expect(SuperEditorInspector.findWidgetForComponent<TextComponent>("Document2_Para"), isNotNull);
+    });
+
+    testWidgetsOnAllPlatforms("each document maintains its selection when switching between documents", (tester) async {
```
matthew-carroll, line 26: Try to describe test behaviors in the present tense rather than passive tense. It's easy to lose track of actors, responsibilities, and time when you speak passively.

`restores selection when switching back to a previously visible editor`

super_editor/test/super_editor/supereditor_demo_switch_document_content_test.dart
```diff
@@ -0,0 +1,301 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:flutter_test_robots/flutter_test_robots.dart';
+import 'package:flutter_test_runners/flutter_test_runners.dart';
+import 'package:super_editor/super_editor.dart';
+import 'package:super_editor/super_editor_test.dart';
+
+void main() {
+  group("SuperEditor document content switching", () {
```
matthew-carroll, line 9: Your changes found that the composer and possibly some other things need to be switched, right? If so, we're not really switching a document. It's probably more appropriate to say that we're switching between editors.

super_editor/test/super_editor/supereditor_demo_switch_document_content_test.dart
```diff
@@ -0,0 +1,301 @@
...
+      await tester.pump();
+
+      // Ensure that the original selection is maintained for the first document.
+      expect(
+        SuperEditorInspector.findDocumentSelection(),
+        docSelection1,
+      );
+    });
+
+    testWidgetsOnDesktop("changes selection of the second document after switching", (tester) async {
```
matthew-carroll, line 80: Similar comment here about improving clarity about who can do what, and when.

`the user can select content after switching to a different editor`

super_editor/test/super_editor/supereditor_demo_switch_document_content_test.dart
```diff
@@ -0,0 +1,301 @@
...
+          ),
+          extent: DocumentPosition(
+            nodeId: paragraph.id,
+            nodePosition: paragraph.endPosition,
+          ),
+        ),
+      );
+    });
+
+    testWidgetsOnDesktop("able to edit the content of the active document after switching", (tester) async {
```
matthew-carroll, line 116: Similar comment here about who can do what.

`the user can edit content in the after switching to a different editor`

super_editor/test/super_editor/supereditor_demo_switch_document_content_test.dart
rutvik110: Could probably put it in the `supereditor_multi_editor_test.dart` file. This wasn't clear to me earlier but makes more sense now to have it there.

matthew-carroll:
@rutvik110 it looks like goldens are broken. It might be due to your work, or it might be due to Flutter `master` changes, but we should find out what it is, and also fix them.

super_editor/test/super_editor/supereditor_multi_editor_test.dart
```diff
@@ -38,6 +39,153 @@ void main() {
       );
     });
   });
+
+  group("SuperEditor switching between multiple editors", () {
+    testWidgetsOnAllPlatforms("selected editor is visible after switching from a different editor", (tester) async {
```
matthew-carroll, line 44: This test name still isn't clear. What's the "selected editor"? Please see my previous comments for examples of test names.

super_editor/test/super_editor/supereditor_multi_editor_test.dart
```diff
@@ -38,6 +39,153 @@ void main() {
...
+          ),
+          extent: DocumentPosition(
+            nodeId: paragraph.id,
+            nodePosition: paragraph.endPosition,
+          ),
+        ),
+      );
+    });
+
+    testWidgetsOnDesktop("the user can edit content in the selected editor after switching from a different editor",
```
matthew-carroll, line 150: Here again "selected editor" adds confusion to what you're trying to say.

super_editor/test/super_editor/supereditor_multi_editor_test.dart
```diff
@@ -84,3 +232,150 @@ Future<void> _buildTextScaleScaffold(
     ),
   );
 }
+
+/// Demo of an [SuperEditor] widget where the [DocumentEditor] changes.
+///
+/// This demo ensures that [SuperEditor] state resets where appropriate
+/// when its content is replaced.
+class _SwitchDocumentDemo extends StatefulWidget {
```
matthew-carroll, line 240: Please check for all references to "switching documents" and adjust to become "switching editors" and any similar situations.

rutvik110:
> @rutvik110 it looks like goldens are broken. It might be due to your work, or it might be due to Flutter `master` changes, but we should find out what it is, and also fix them.

I investigated this issue and the affected areas are the tests that are dealing with text rendering and text layout. The issue seems to be arising due to the text being rendered with extra boldness as seen in below image. 

Screenshot 2023-10-01 at 11 43 41 AM

I'm testing this on Flutter master `3.15.0-13.0.pre.28`. And the root cause seems like the `Text.rich` that's rendering the text.
But the suprising thing is that I wasn't able to reproduce the issue when running the example app or directly testing the Text.rich output in the running app.  It's only being reproduced through tests for reasons I'm not sure abt rn.

There's an [open issue](https://github.com/flutter/flutter/issues/134979) on Flutter that's tracking this behaviour. So this seems like a Flutter issue rather than something on our end. And Atm, this doesn't seem to be affecting any running examples of `SuperEditor` but I've to investigate why it's like that.

Should I file it as an separate issue here for us to track it?

matthew-carroll:
@rutvik110 did you run the goldens using the Docker image test runner?

rutvik110:
I'vent. I'll try running it through docker image.

matthew-carroll:
Ok. The goldens must always be run through the Docker image. That's how we stay consistent across all of our machines.

rutvik110:
I'm facing some issues while trying to run the golden tests locally. I've failed the issue for it #1479 

@matthew-carroll Beside the failing tests, I've made the requested updates to this PR related to tests naming and docs. The failure of tests doesn't seem to be related to this PR so should we go ahead with it? or do we want to wait till we fix the issue with the tests.

matthew-carroll (approved):
LGTM

PR #1421, opened 2023-09-09
[SuperEditor][SuperReader] - Lock scrolling when content fits within available space (Resolves #1409)

### LINKED ISSUE #1409: [SuperEditor][SuperReader] - Is scrollable when it doesn't need to be
`SuperEditor` and `SuperReader` are scrollable, even when all of their content fits the available space. The following demo wraps `SuperEditor` with `IntrinsicHeight`.

https://github.com/superlistapp/super_editor/assets/7259036/7965cec1-8b61-4af9-a7c4-13b3f2442ef6

rutvik110:
[SuperEditor][SuperReader] - Lock scrolling when content fits within available space (Resolves #1409)

Currently in the `SuperEditor` or `SuperReader`, the document is scrollable even if the content fits within the available space as seen in the below demo. 

This demo wraps the `SuperEditor` within an `IntrinsicHeight` widget to demonstrate the issue,

https://github.com/superlistapp/super_editor/assets/7259036/7965cec1-8b61-4af9-a7c4-13b3f2442ef6

This PR tries to address this issue by locking the scrolling for the document in such cases. We lock the scrolling when the content of the document fits within the available viewport space and there isn't any unconsumed content within the `Scrollable`. 

Scroll lock is introduced within both `DocumentScroller` and `AutoScrollController`.  When the scroll lock is enabled, any calls/actions related to scrolling are ignored to prevent the document from scrolling. 

After this latest updates, the scrolling is locked whenever content fits within available space as seen in the following demo,

https://github.com/superlistapp/super_editor/assets/65209850/a5c83f3b-a2db-4c0e-b96f-0d5288ca27a6

### DIFF
--- a/super_editor/example/lib/demos/demo_switch_document_content.dart
+++ b/super_editor/example/lib/demos/demo_switch_document_content.dart
@@ -50,7 +50,8 @@ class _SwitchDocumentDemoState extends State<SwitchDocumentDemo> {
       child: Column(
         children: [
           _buildDocSelector(),
-          Expanded(
+          IntrinsicHeight(
+            // height: 300,
             child: SuperEditor(
               editor: _activeDocumentEditor,
               document: _activeDocument,
--- a/super_editor/example/lib/demos/demo_switch_document_content.dart
+++ b/super_editor/example/lib/demos/demo_switch_document_content.dart
@@ -50,8 +50,7 @@ class _SwitchDocumentDemoState extends State<SwitchDocumentDemo> {
       child: Column(
         children: [
           _buildDocSelector(),
-          IntrinsicHeight(
-            //  height: 400,
+          Expanded(
             child: SuperEditor(
               editor: _activeDocumentEditor,
               document: _activeDocument,
--- a/super_editor/example/lib/demos/demo_switch_document_content.dart
+++ b/super_editor/example/lib/demos/demo_switch_document_content.dart
@@ -51,7 +51,7 @@ class _SwitchDocumentDemoState extends State<SwitchDocumentDemo> {
         children: [
           _buildDocSelector(),
           IntrinsicHeight(
-            // height: 300,
+            //  height: 400,
             child: SuperEditor(
               editor: _activeDocumentEditor,
               document: _activeDocument,
--- a/super_editor/lib/src/default_editor/document_scrollable.dart
+++ b/super_editor/lib/src/default_editor/document_scrollable.dart
@@ -387,6 +387,12 @@ class AutoScrollController with ChangeNotifier {
       return;
     }
 
+    // There's no content to scroll. So, we can't go ballistic as it would result
+    // in overscroll.
+    if (pos.maxScrollExtent == 0) {
+      return;
+    }
+
     if (pos is ScrollPositionWithSingleContext) {
       pos.goBallistic(pixelsPerSecond);
       pos.context.setIgnorePointer(false);
--- a/super_editor/lib/src/default_editor/document_scrollable.dart
+++ b/super_editor/lib/src/default_editor/document_scrollable.dart
@@ -197,10 +197,9 @@ class _DocumentScrollableState extends State<DocumentScrollable> with SingleTick
   }
 
   void _recomputeShoulDisableScrollingIfContentFitsWithinViewport() {
-    final contentAfter = (widget.scroller!.extentAfter - widget.docPadding.bottom);
+    final contentAfter = (_scrollPosition.extentAfter - widget.docPadding.bottom);
     final isContentFit = //
-        (widget.scroller!.extentTotal - widget.scroller!.extentAfter + contentAfter) <=
-            widget.scroller!.viewportDimension;
+        (_scrollPosition.extentTotal - _scrollPosition.extentAfter + contentAfter) <= _scrollPosition.viewportDimension;
 
     widget.scroller!.isScrollingEnabled = !isContentFit;
     widget.autoScroller.isScrollingEnabled = !isContentFit;
--- a/super_editor/lib/src/default_editor/document_scrollable.dart
+++ b/super_editor/lib/src/default_editor/document_scrollable.dart
@@ -373,6 +373,11 @@ class AutoScrollController with ChangeNotifier {
     }
 
     final scrollPosition = _getScrollPosition!();
+
+    if (scrollPosition.maxScrollExtent == 0) {
+      return;
+    }
+
     scrollPosition.jumpTo(
       (scrollPosition.pixels + delta).clamp(0.0, scrollPosition.maxScrollExtent),
     );
@@ -387,10 +392,12 @@ class AutoScrollController with ChangeNotifier {
       return;
     }
 
+    if (pos.maxScrollExtent == 0) {
+      return;
+    }
+
     if (pos is ScrollPositionWithSingleContext) {
-      if (pos.maxScrollExtent > 0) {
-        pos.goBallistic(pixelsPerSecond);
-      }
+      pos.goBallistic(pixelsPerSecond);
       pos.context.setIgnorePointer(false);
     }
   }
--- a/super_editor/lib/src/default_editor/document_scrollable.dart
+++ b/super_editor/lib/src/default_editor/document_scrollable.dart
@@ -392,12 +392,10 @@ class AutoScrollController with ChangeNotifier {
       return;
     }
 
-    if (pos.maxScrollExtent == 0) {
-      return;
-    }
-
     if (pos is ScrollPositionWithSingleContext) {
-      pos.goBallistic(pixelsPerSecond);
+      if (pos.maxScrollExtent > 0) {
+        pos.goBallistic(pixelsPerSecond);
+      }
       pos.context.setIgnorePointer(false);
     }
   }
--- a/super_editor/lib/src/default_editor/document_scrollable.dart
+++ b/super_editor/lib/src/default_editor/document_scrollable.dart
@@ -31,6 +31,7 @@ class DocumentScrollable extends StatefulWidget {
     this.scroller,
     this.scrollingMinimapId,
     this.showDebugPaint = false,
+    required this.docPadding,
     required this.child,
   }) : super(key: key);
 
@@ -56,6 +57,8 @@ class DocumentScrollable extends StatefulWidget {
   /// Paints some extra visual ornamentation to help with
   /// debugging, when `true`.
   final bool showDebugPaint;
+
+  final EdgeInsets docPadding;
 
   /// This widget's child, which should include a document.
   final Widget child;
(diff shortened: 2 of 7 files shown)


rutvik110:
@matthew-carroll Could you please take a look at the approach to fix this issue? It's still WIP but wanted your opinion on the approach and see if whether this is the direction that we want to move in with this issue.

matthew-carroll:
@rutvik110 before evaluating your choice of solution, can you describe what you investigated? For example, why is the scrollable letting us scroll when there's no content to scroll?

rutvik110:
> @rutvik110 before evaluating your choice of solution, can you describe what you investigated? For example, why is the scrollable letting us scroll when there's no content to scroll?

There seems to be two reasons here that are causing the overscroll.

The first reason is related to our handling of the scrolling within `Scrollable`. In case of auto-scrolling in response to mouse events, currently we are always trying to scroll irrespective of whether there's any content left to scroll. 
This calls to scroll is leading to the overscroll/scroll-bounce that we are noticing in demo you shared. 

Second reason is bit more ambiguous and it may be totally different issue that I may've confused with this one but here it's. It seems to be particularly related to usage of `IntrinisicHeight`. I noticed that when we have large chunks of paragraphs within `SuperEditor`, the final height calculated by the `IntrinisicHeight` to layout the `SuperEditor` isn't correct and is smaller by some magnitude. So the `SuperEditor` is laid out with the height smaller than it's contents actual height which results in introducing some content within `SuperEditor` beyond its scrollable viewport.

https://github.com/superlistapp/super_editor/assets/65209850/630ebbd6-bb09-4fb5-a776-a3ae818fa49c

super_editor/lib/src/default_editor/document_scrollable.dart
```diff
@@ -387,6 +387,12 @@ class AutoScrollController with ChangeNotifier {
       return;
     }
 
+    // There's no content to scroll. So, we can't go ballistic as it would result
```
rutvik110, line 390: @matthew-carroll I've reverted the previous additions I made to this issue. I think those additions were targeting a different issue than the one we're trying to solve here. 

Wrt the first reason I mentioned above for overscrolling, this update resolves that. 
Currently, we do a final scroll call with `goBallistic` to animate to a new scroll position at the end of the mouse events that results in scrolling the content.
I added a check to prevent making this call if there's no content to scroll to or `maxScrollExtent` is 0. 

This should take care of the overscroll that was happening when the content fits within the space.

super_editor/lib/src/default_editor/document_scrollable.dart
```diff
@@ -388,7 +388,9 @@ class AutoScrollController with ChangeNotifier {
     }
 
     if (pos is ScrollPositionWithSingleContext) {
-      pos.goBallistic(pixelsPerSecond);
+      if (pos.maxScrollExtent > 0) {
```
matthew-carroll, line 391: Doesn't this only check a boundary condition in one scroll direction?

Also, what about when the user is using the scroll wheel? I was able to overscroll using the scroll wheel which I believe is unrelated to going ballistic.

super_editor/example/lib/demos/demo_switch_document_content.dart
```diff
@@ -50,7 +50,8 @@ class _SwitchDocumentDemoState extends State<SwitchDocumentDemo> {
       child: Column(
         children: [
           _buildDocSelector(),
-          Expanded(
+          IntrinsicHeight(
```
matthew-carroll, line 53: Please revert this.

super_editor/lib/src/default_editor/document_scrollable.dart
rutvik110: It does. Though I'm not sure whether the check for upward direction is necessary. If we had to check within the upward direction, we'll be checking for `minScrollExtent`.  The docs for `minScrollExtent` states that,

```
This value is typically less than or equal to [maxScrollExtent]. It can be negative infinity, if the scroll is unbounded.
```
Going by that fact, in case where the content fits the available space, if it's not negative then it should be equal to `maxScrollExtent` which is 0. And if it's negative or goes to negative infinity, I'm not sure how that would affect the scrolling so can't really come around good way to handle it. 
May be I'm overthinking it and we can just do the similar ` > 0` check on `minScrollExtent`  if we're scrolling in upward direction? Wdyt?


On the scroll view issue, I wasn't aware of that. Will check on it.

rutvik110:
> Also, what about when the user is using the scroll wheel? I was able to overscroll using the scroll wheel which I believe is unrelated to going ballistic.

ah yah, had to add the same max extent check to `jumpBy` which resolved this.  @matthew-carroll Could you please re-check this on your end? 

Also, what are your thoughts on the previous comment I made regarding extent check in the reverse direction?

matthew-carroll:
I think the current changes are probably good as-is. Can you confirm that these changes work for SuperReader as well as SuperEditor? Also, can you write a test for each of these scenarios in the appropriate test file?

And please don't forget to revert the code that I mentioned in an earlier comment.

rutvik110:
Great! Will check on `SuperReader` once which I believe should work fine as well and get onto writing tests for this.

super_editor/test/super_editor/supereditor_scrolling_test.dart
```diff
@@ -682,11 +683,55 @@ void main() {
...
           // Ensure SuperEditor has scrolled
           expect(editorScrollController.offset, greaterThan(0));
-
-          // Ensure that scrolling didn't scroll the ListView
-          expect(listScrollController.position.pixels, equals(0));
         });
       });
+
+      group("doesn't scroll when content fits within the available space", () {
+        testWidgetsOnDesktop(
```
matthew-carroll, line 690: Did you verify that this test fails without your changes? Also, this looks like it tests a touch gesture, but what about the scroll wheel?

super_editor/test/super_editor/supereditor_scrolling_test.dart
```diff
@@ -682,11 +683,55 @@ void main() {
...
+
+            // Perform a fling on the editor to attemp scrolling.
+            await tester.trackpadFling(
+              find.byType(SuperEditor),
+              Offset(0.0, isScrollingUp ? 100 : -100),
+              300,
+            );
+
+            // Pump a few frames of momentum.
+            for (int i = 0; i < 5; i += 1) {
```
matthew-carroll, line 725: Is there a reason that you need to pump 5 frames? What if you only pump a single frame?

super_editor/test/super_editor/supereditor_scrolling_test.dart
rutvik110: Yah I did and it does fails without this changes.

Will add tests for scroll wheel as well.

super_editor/test/super_editor/supereditor_scrolling_test.dart
rutvik110: In case where checking for `ScrollPosition.pixels` to verify the expected outcome, this was needed to let it gain some momentum to report offset pixels values and verify test fails without this PR updates.

Works with a single frame push when checking for scroll activity. Though should we let it pump a few initial frames to avoid any cases where we don't have updated info from `Scrollable` like in case when checking for `.pixels`.

super_editor/test/super_editor/supereditor_scrolling_test.dart
```diff
@@ -682,11 +683,103 @@ void main() {
...
           expect(editorScrollController.offset, greaterThan(0));
-
-          // Ensure that scrolling didn't scroll the ListView
-          expect(listScrollController.position.pixels, equals(0));
         });
       });
+
+      group("doesn't scroll editor when content fits within the available space", () {
+        testWidgetsOnDesktop(
+          "attempts to scroll editor vertically in both directions using trackpad",
```
matthew-carroll, line 691: These test names aren't great. Look at what the final test name is:

"doesn't scroll editor when content fits within the available space attempts to scroll editor vertically in both directions using trackpad"

I think you can describe the group and test in a way that's much more concise, and makes sense when read as the final test description.

super_editor/test/super_editor/supereditor_scrolling_test.dart
matthew-carroll: I tried to look for this code but it seems to be gone. Did you decide that a single pump is enough? If so, did you discover something that caused you to rethink your reply above where you said you thought we needed multiple frames?

super_editor/test/super_editor/supereditor_scrolling_test.dart
```diff
@@ -682,11 +683,103 @@ void main() {
 
           // Ensure SuperEditor has scrolled
           expect(editorScrollController.offset, greaterThan(0));
-
-          // Ensure that scrolling didn't scroll the ListView
-          expect(listScrollController.position.pixels, equals(0));
         });
       });
+
+      group("in absence of scrollable content", () {
```
matthew-carroll, line 689: This can still be much more direct about what we care about: "when all content fits in the viewport"

super_editor/test/super_editor/supereditor_scrolling_test.dart
```diff
@@ -682,11 +683,103 @@ void main() {
...
+        testWidgetsOnDesktop(
+          "doesn't scroll on scroll attempts through trackpad",
```
matthew-carroll, line 691: "trackpad doesn't scroll content"

super_editor/test/super_editor/supereditor_scrolling_test.dart
```diff
@@ -682,11 +683,103 @@ void main() {
...
+          (tester) async {
+            const windowSize = Size(800, 600);
```
matthew-carroll, line 693: Is there a reason that `windowSize` is declared as a variable instead of directly setting the `physicalSize`?

super_editor/test/super_editor/supereditor_scrolling_test.dart
```diff
@@ -682,11 +683,103 @@ void main() {
...
+            await tester.pump();
+
+            // Ensure SuperEditor is not scrolling.
+            expect(scrollState.position.activity?.isScrolling, false);
+          },
+          variant: _scrollDirectionVariant,
+        );
+
+        testWidgetsOnDesktop(
+          "doesn't scroll on scroll attempts through scroll wheel",
```
matthew-carroll, line 733: "mouse scroll wheel doesn't scroll content"

super_editor/test/super_reader/super_reader_scrolling_test.dart
```diff
@@ -218,5 +219,110 @@ void main() {
         isTrue,
       );
     });
+
+    group("in absence of scrollable content", () {
```
matthew-carroll, line 223: Same naming and other test adjustments in this file.

super_editor/test/super_editor/supereditor_scrolling_test.dart
rutvik110: No particular reason. Just tried following it the same way as we are doing it within above tests.
But makes sense to just directly set it to physicalSize instead of creating an additional variable. Will update it.

matthew-carroll (approved):
LGTM

PR #1422, opened 2023-09-09
[SuperEditor][iOS] Fix magnifier (Resolves #1329)

### LINKED ISSUE #1329: [SuperTextField] Flutter 3.13 Bug - MagnifyingGlass Broke
With the new Flutter 3.13 update on iOS, when you select text and move the selection using the handles the magnifier flashes (disappears and reappears) as you move the mouse.

Not sure what is causing this, however I believe I've narrowed it down a bit. See the following file: https://github.com/superlistapp/super_editor/blob/main/super_editor/lib/src/super_textfield/infrastructure/magnifier.dart 

In the `MagnifyingGlass` class, the `offsetFromFocalPoint` usually is `Offset(0, -72)` however it does work when it is like `Offset(0,-40)`.

Everything else up the widget tree seems to not be affected as I can put a container width/hight: 40 color:blue above it and it doesn't disappear ever. The problem seems to be centered on the `BackdropFilter`.

angelosilvestre:
[SuperEditor][iOS] Fix magnifier. Resolves #1329.

On iOS, the magnifier is flickering when moving and it also doesn't magnify at all.

The reason is that we are applying a `1.0` magnification scale.

After adjusting the scale, the shadows are being displayed above the magnifier. This happens only on impeller. 

image

I changed the order between the `Container` which displays the shadows and the magnifier and the issue stopped:

https://github.com/superlistapp/super_editor/assets/7597082/f8586099-15bb-4191-b94d-5dbf1a160752

### DIFF
--- a/super_editor/lib/src/infrastructure/platforms/ios/magnifier.dart
+++ b/super_editor/lib/src/infrastructure/platforms/ios/magnifier.dart
@@ -55,7 +55,7 @@ Widget _circleMagnifierBuilder(BuildContext context, Offset offsetFromFocalPoint
     );
 
 class IOSRoundedRectangleMagnifyingGlass extends StatelessWidget {
-  static const _magnification = 1.0;
+  static const _magnification = 1.5;
 
   const IOSRoundedRectangleMagnifyingGlass({
     this.offsetFromFocalPoint = Offset.zero,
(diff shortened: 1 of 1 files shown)


matthew-carroll (approved):
LGTM

Issue #1288, opened 2023-08-10
[SuperEditor][SuperReader] - Expose scroll info and controls


matthew-carroll:
Scroll control is needed outside of the document scrollable widget, e.g., to scroll for Page Up, Page Down, Home, End. Therefore, some amount of scroll control needs to be provided through the `SuperEditorContext`.

Clients at least need access to:
- Viewport size (possibly the global rect)
- Min/max scroll extent
- Get/Set scroll offset
- Animate the scroll offset

Scroll control APIs need to be available on all platforms.

PR #1423, opened 2023-09-10
[SuperReader] Migrate to DocumentScroller API (Resolves #1306)

### LINKED ISSUE #1306: [SuperReader] - Migrate scrolling API to new DocumentScroller
We introduced a narrow scrolling API called `DocumentScroller` in https://github.com/superlistapp/super_editor/issues/1288. That change was aimed at `SuperEditor` for page up/down behavior, but it required passing `DocumentScroller` to `DocumentScrollable`, which is used by both `SuperEditor` and `SuperReader`.

I got `SuperReader` to compile with a dummy `DocumentScroller` but I didn't refactor `SuperReader` to attach itself to the `DocumentScroller`. For this ticket, complete the migration of `SuperReader` from its stand alone scrolling API to this shared API.

angelosilvestre:
[SuperReader] Migrate to DocumentScroller API. Resolves #1306

https://github.com/superlistapp/super_editor/issues/1288 introduced the `DocumentScroller` API, but it wasn't fully migrated within `SuperReader`.

`SuperReader` already uses a `DocumentScaffold`, which uses `DocumentScrollable` to attach/detach from the scroller.

The only things I found still left to do is to migrate `SuperReaderContext` to take a `DocumentScroller` instead of an `AutoScrollController` and adjust key up/down handlers.

### DIFF
--- a/super_editor/lib/src/infrastructure/documents/document_scroller.dart
+++ b/super_editor/lib/src/infrastructure/documents/document_scroller.dart
@@ -29,6 +29,11 @@ class DocumentScroller {
     _scrollPosition!.jumpTo(newScrollOffset);
   }
 
+  /// Immediately moves the [scrollOffset] by [delta] pixels.
+  void jumpBy(double delta) {
+    _scrollPosition!.jumpTo(_scrollPosition!.pixels + delta);
+  }
+
   /// Animates [scrollOffset] from its current offset to [to], over the given [duration]
   /// of time, following the given animation [curve].
   void animateTo(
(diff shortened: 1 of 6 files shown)


matthew-carroll:
@angelosilvestre you mentioned you replaced an `AutoScrollController` with the `DocumentScroller`. Does that mean that `SuperReader` can no longer auto-scroll? Did we lose anything?

matthew-carroll (approved):
LGTM

angelosilvestre:
@matthew-carroll `AutoScroller` still work as before. It's justo not accessible by `SuperReaderContext` anymore. The code related to auto-scroll receives `AutoScroller` via constructor.

PR #1430, opened 2023-09-12
[SuperEditor][SuperTextField] Use more specific intents to block Flutter default handlers (Resolves #1408)

### LINKED ISSUE #1408: [SuperTextField] ESCAPE key event is not bubbling up
Steps to reproduce:
* Wrap a `SuperTextField` with `Actions` widget that handles the `DismissIntent` (usually invoked when ESCAPE is pressed)
* Give focus to the field
* Press ESCAPE

**Expected:** the handler in `Actions` for the `DismissIntent` is invoked
**Actual:** nothing happens

Here's a code snippet with the changes that I made in the build method of `InteractiveTextFieldDemo`:
<details>
  <summary>Code snippet</summary>

  ### Before
  ```dart
  @override
    return TapRegion(
  ```
  ### After
  ```dart
  @override
    return Actions(
      actions: {
        DismissIntent: CallbackAction<DismissIntent>(
          onInvoke: (_) {
            print('DismissIntent');
            _focusNode!.unfocus();
            return () {};
          },
        ),
      },
      child: TapRegion(
  ```
</details>

From trying different commits, it seems that this was introduced in 3df65bf8d311ceb8c914b676bd37a1a7bf212cfb.

The ticket mentions explicitly the ESCAPE key, but this could be applicable for other keys, although I haven't come across other issues.

angelosilvestre:
[SuperEditor][SuperTextField] Use more specific intents to block Flutter default handlers. Resolves #1408

Flutter install default shortcuts for arrow keys, page up, page down, and others. This prevents the key events from bubbling up to the IME, which prevents selectors from being generate. 

Because of that, we are handling `DoNothingAndStopPropagationTextIntent` to stop these intents. However, by doing this we prevent apps from handling other intents, like `DismissIntent`.

This PR replaces `DoNothingAndStopPropagationTextIntent` with more specific intents.


super_editor/lib/src/infrastructure/platforms/mac/mac_ime.dart
```diff
@@ -61,3 +63,23 @@ class MacOsSelectors {
...
   static const String insertNewLine = 'insertNewline:';
 }
+
+/// Stop propagating intents that prevent [TextInputClient.performSelector] from being called.
+///
+/// These intents have default handlers installed by Flutter
+/// that prevents the key events from bubbling up to the IME.
+///
+/// Use this object in an [Actions] widget.
+final Map<Type, Action<Intent>> disabledMacIntents = {
```
matthew-carroll, line 73: Let's rename to "disabledMacIntents" otherwise it sounds like this is an action, rather than a collection.

angelosilvestre: It's already named disabledMacIntents :)

matthew-carroll: Weird. I didn't see a "d" there at all. Sounds good.

matthew-carroll (approved):
LGTM

PR #1442, opened 2023-09-12
Tap to open multiple links in the same paragraph (Resolves #1426)

### LINKED ISSUE #1426: [SuperEditor] Paragraphs with multiple links only open the first link when tapped
If a ParagraphNode contains more than one LinkAttribution, tapping any of the links in interaction mode will open the first link's URL in the browser. For example, the first link in the following video's URL is google.com, while the second link's URL is yahoo.com.

https://github.com/superlistapp/super_editor/assets/6467808/37be616c-e734-4e17-b5e1-e8a6118cf048

These lines in `_getLinkAtPosition` in `super_editor.dart` return the first LinkAttributon's URL, rather than the url that was tapped:

```
    final tappedAttributions = textNode.text.getAllAttributionsAt(nodePosition.offset);
    print('tappedAttributions: $tappedAttributions');
    for (final tappedAttribution in tappedAttributions) {
      if (tappedAttribution is LinkAttribution) {
        return tappedAttribution.url;
      }
    }
```

cc @matthew-carroll

matthew-carroll:
Tap to open multiple links in the same paragraph (Resolves #1426)

The root problem is that when looking up attributions in `AttributionSpans` at a specific location, we weren't differentiating between different URLs in link attributions. This is because the attribution differentiation was based only on the attribution `id`, which is always `link` for a link attribution.

I changed the attribution comparison to compare the overall attribution instead of just the attribution ID. This might break some other styling behaviors, but all the tests pass, so we'll have to handle those situations when they come up.

### DIFF
--- a/attributed_text/lib/src/attributed_spans.dart
+++ b/attributed_text/lib/src/attributed_spans.dart
@@ -265,7 +265,7 @@ class AttributedSpans {
     return _markers //
         .reversed // search from the end so its the nearest start marker
         .where((marker) {
-      return attribution == null || (marker.attribution.id == attribution.id);
+      return attribution == null || (marker.attribution == attribution);
     }).firstWhereOrNull((marker) => marker.isStart && marker.offset <= offset);
   }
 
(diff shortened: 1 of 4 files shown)


angelosilvestre (approved):
LGTM

PR #1443, opened 2023-09-13
[ExampleApp] Fix link popover (Resolves #1440)

### LINKED ISSUE #1440: [Example App][Mac] - Can't enter a link URL anymore
In the Example app, entering a URL requires that the user presses ENTER after typing the URL, which should apply the URL to the text and close the toolbar.

Currently, on Mac, pressing ENTER bypasses the toolbar and seems to be applied by Super Editor, which then deletes the selected text and inserts a new paragraph.

On Mac web, pressing ENTER dismisses the toolbar without changing the text at all.

angelosilvestre:
[ExampleApp] Fix link popover. Resolves #1440

The implementation of `performSelectors` broke the tag popover demos. The fix for that ended up breaking the  url popover in the example app.

For selectors work, we need to prevent default Flutter shortcuts that perform focus traversal. We currently do that by using an `Actions` widget in `SuperEditor` and `SuperTextField`. However, when the editor doesn't have primary focus, which is the case when a popover is displayed, `SuperEditor` actions are ignored.

To fix that, the popovers also need to prevent these shortcuts. To make it easier for app, this PR introduces a `SuperEditorPopover` widget, which adds an `Actions` widget in the tree with the intents that need to be blocked.

Another issue is that pressing ENTER on mac generates a `onPerformAction` call. The demo needed to be updated to handle `onPerformAction` to also submit the link.

### DIFF
--- a/super_editor/lib/src/infrastructure/popovers.dart
+++ b/super_editor/lib/src/infrastructure/popovers.dart
@@ -1,0 +1,22 @@
+import 'package:flutter/material.dart';
+import 'package:super_editor/src/infrastructure/platforms/mac/mac_ime.dart';
+
+/// Widget that displays a [child] and blocks Flutter [Intent]s
+/// that causes focus traversal.
+class SuperEditorPopover extends StatelessWidget {
+  const SuperEditorPopover({
+    super.key,
+    required this.child,
+  });
+
+  /// The popover to display.
+  final Widget child;
+
+  @override
+  Widget build(BuildContext context) {
+    return Actions(
+      actions: disabledMacIntents,
+      child: child,
+    );
+  }
+}
--- a/super_editor/lib/src/infrastructure/popovers.dart
+++ b/super_editor/lib/src/infrastructure/popovers.dart
@@ -1,4 +1,5 @@
 import 'package:flutter/material.dart';
+import 'package:super_editor/src/infrastructure/focus.dart';
 import 'package:super_editor/src/infrastructure/platforms/mac/mac_ime.dart';
 
 /// A popover that shares focus with a `SuperEditor`.
@@ -28,8 +29,22 @@ import 'package:super_editor/src/infrastructure/platforms/mac/mac_ime.dart';
 class SuperEditorPopover extends StatelessWidget {
   const SuperEditorPopover({
     super.key,
+    required this.popoverFocusNode,
+    required this.editorFocusNode,
+    this.onKeyEvent,
     required this.child,
   });
+
+  /// The [FocusNode] attached to the popover.
+  final FocusNode popoverFocusNode;
+
+  /// The [FocusNode] attached to the editor.
+  ///
+  /// The [popoverFocusNode] will be reparented with this [FocusNode].
+  final FocusNode editorFocusNode;
+
+  /// Callback that notifies key events.
+  final FocusOnKeyEventCallback? onKeyEvent;
 
   /// The popover to display.
   final Widget child;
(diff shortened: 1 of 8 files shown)


super_editor/lib/src/infrastructure/popovers.dart
```diff
@@ -0,0 +1,22 @@
+import 'package:flutter/material.dart';
+import 'package:super_editor/src/infrastructure/platforms/mac/mac_ime.dart';
+
+/// Widget that displays a [child] and blocks Flutter [Intent]s
```
matthew-carroll, line 4: We should be far more descriptive about when and why this is used. Try to think in terms of someone who is using Super Editor rather than implementing it. What are they likely to know? What questions are they likely to ask?

```dart
/// A popover that shares focus with a `SuperEditor`.
///
/// Popovers often need to handle keyboard input, such as arrow keys for
/// selection movement. But `SuperEditor` also needs to continue handling
/// keyboard input to move the caret, enter text, etc. In such a case, the
/// popover has primary focus, and `SuperEditor` has non-primary focus.
/// Due to the way that Flutter's `Actions` system works, along with
/// Flutter's default response to certain key events, a few careful
/// adjustments need to be made so that a popover works with 
/// `SuperEditor` as expected. This widget handles those adjustments.
///
/// Despite this widget being a "Super Editor popover", this widget can be
/// placed anywhere in the widget tree, so long as it's able to share focus
/// with `SuperEditor`.
///
/// This widget is purely logical - it doesn't impose any particular layout
/// or constraints. It's up to you whether this widget tightly hugs your
/// popover [child], or whether it expands to fill a space.
///
/// It's possible to create a `SuperEditor` popover without this widget.
/// This widget doesn't have any special access to `SuperEditor`
/// properties or behavior. But, if you choose to display a popover
/// without using this widget, you'll likely need to re-implement this
/// behavior to avoid unexpected user interaction results.
```

super_editor/lib/src/infrastructure/popovers.dart
```diff
@@ -0,0 +1,22 @@
...
+    super.key,
+    required this.child,
+  });
+
+  /// The popover to display.
+  final Widget child;
+
+  @override
+  Widget build(BuildContext context) {
+    return Actions(
```
matthew-carroll, line 17: Should we also add the `Focus` sharing stuff that's likely to be needed for any such popover? Take a look at the tagging demo in that regard.

angelosilvestre: In `PopoverList`, the focus sharing is configured by setting the `parentNode` in the popover `Focus` widget, right?

```dart
return SuperEditorPopover(
    child: GestureDetector(
      onTap: () => !_focusNode.hasPrimaryFocus ? _focusNode.requestFocus() : null,
      child: Focus(
        focusNode: _focusNode,
        parentNode: widget.editorFocusNode,
        onKeyEvent: _onKeyEvent,
        ...
```

Should we use a `FocusWithCustomParent` within  `SuperEditor` popover and remove the `parentNode` from `Focus`?

matthew-carroll: I think that's a question as to which one works in which one doesn't. I don't remember which situation fails with which one. If you believe that `FocusWithCustomParent` is the right answer in general then we can use that. Please adjust any and all popovers in the example app across all demos to use this new popover wrapper, and then please do a spot check on each one to make sure that focus seems to continue working as expected.

matthew-carroll (approved):
LGTM

PR #1444, opened 2023-09-13
[SuperTextField][Mac] Allow selector handlers configuration (Resolves #1439)

### LINKED ISSUE #1439: [SuperTextField] on Mac it is not possible to overwrite key events that get handled by the Mac OS selectors
Since #836 that on Mac the key events such as arrow keys, are not handled by keyboard handlers but by the Mac OS selectors.

With `SuperEditor` it is possible to pass a list of selector, similar to what we do for keyboard handlers, but not for `SuperTextField`. It would be great to have this capability there as well.

angelosilvestre:
[SuperTextField][Mac] Allow selector handlers configuration. Resolves #1439

Currently, we allow apps to provide custom selector handlers to `SuperEditor` but not for `SuperTextField`.

This PR exposes `selectorHandlers` in `SuperTextField`'s public API.


matthew-carroll (approved):
LGTM

PR #1455, opened 2023-09-18
[SuperEditor][Web] Fix entering spaces (Resolves #1446)

### LINKED ISSUE #1446: [SuperEditor] [Web] Cannot enter spaces
Repro: Run the example app targeting web and try to enter spaces. Nothing happens. This occurs in debug and profile mode.

I'm running Flutter 3.13.3 stable and super_editor stable.

cc @matthew-carroll @angelosilvestre

angelosilvestre:
[SuperEditor][Web] Fix entering spaces. Resolves #1446

On web, entering spaces is not possible. The reason is that Flutter is generating another `Intent` when the spacebar is pressed. This PR blocks this `Intent` from bubbling up.

At some point we will need integration tests for preventing regressions with keyboard handling on web.


matthew-carroll:
@angelosilvestre can you reference this ticket in the existing Flutter issue where we've requested that Flutter support text input in integration tests?

super_editor/lib/src/infrastructure/platforms/mac/mac_ime.dart
```diff
@@ -82,4 +82,48 @@ final Map<Type, Action<Intent>> disabledMacIntents = {
 
   // Generated by pressing HOME/END.
   ScrollToDocumentBoundaryIntent: DoNothingAction(consumesKey: false),
+
+  // Generated by pressing SPACE.
+  // PrioritizedIntents might contain intents that we don't want to prevent
+  // from bubbling up.
+  // So, we need to look into its inner intents.
+  PrioritizedIntents: _PreventPrioritizedIntentsFromBubblingUp(
+    intentFilter: (e) => e is ActivateIntent,
```
matthew-carroll, line 91: Space bar sends an "activate intent"? That seems like a very strange intent to send...

super_editor/lib/src/infrastructure/platforms/mac/mac_ime.dart
```diff
@@ -82,4 +82,48 @@ final Map<Type, Action<Intent>> disabledMacIntents = {
...
+  void invoke(Intent intent) {}
+
+  @override
+  bool isEnabled(PrioritizedIntents intent, [BuildContext? context]) {
+    final FocusNode? focus = primaryFocus;
+    if (focus == null || focus.context == null) {
+      return false;
+    }
+
+    for (final Intent candidateIntent in intent.orderedIntents) {
```
matthew-carroll, line 120: What does it mean if there are multiple intents? What does it mean if we block all intents when a single intent in that list matches our filter?

super_editor/lib/src/infrastructure/platforms/mac/mac_ime.dart
angelosilvestre: Yes. This is the definition of the shortcut:

```dart
  // Default shortcuts for the web platform.
  static const Map<ShortcutActivator, Intent> _defaultWebShortcuts = <ShortcutActivator, Intent>{
    // Activation
    SingleActivator(LogicalKeyboardKey.space): PrioritizedIntents(
      orderedIntents: <Intent>[
        ActivateIntent(),
        ScrollIntent(direction: AxisDirection.down, type: ScrollIncrementType.page),
      ],
    ),
```

I filed https://github.com/flutter/flutter/issues/134912 in the hopes we get better debug info for these shortcut activations.

super_editor/lib/src/infrastructure/platforms/mac/mac_ime.dart
angelosilvestre: Flutter iterates over the list of `Intent`s until it finds an enabled `Action`. It won't execute other actions after that.

What we can do is to also check if the `Action` found for the `Intent` is enabled before we block it.

matthew-carroll: That's probably a good change to make. It sounds like we want to have as little impact on the natural behavior as possible.

angelosilvestre: Updated.

matthew-carroll (approved):
LGTM

PR #1469, opened 2023-09-26
[SuperTextField] Fix new line duplication (Resolves #1468)

### LINKED ISSUE #1468: [SuperTextField] SHIFT+ENTER adds 2 new lines
Steps to reproduce:
* Run interactive SuperTextField demo 
* Place the selection somewhere in the content
* Press SHIFT+ENTER

**Expected:** a single new line is added
**Actual:** 2 new lines are added

angelosilvestre:
[SuperTextField] Fix new line duplication. Resolves #1468

On macOS, pressing ENTER or SHIFT + ENTER is inserting two lines. This is caused because pressing ENTER generates both an `onPerformAction` call and a `\n` insertion.

On Windows this issue doesn't happen because we are using key event handlers to insert new lines. If we remove the key handler that inserts new lines, the issue happens on Windows as well. This issue was introduced when the selector handlers were implemented.

This PR changes `_onPerformAction` to avoid inserting new lines.

### DIFF
--- a/super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
+++ b/super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
@@ -1308,7 +1308,13 @@ class _SuperTextFieldImeInteractorState extends State<SuperTextFieldImeInteracto
   void _onPerformAction(TextInputAction action) {
     switch (action) {
       case TextInputAction.newline:
-        _textController.insertNewline();
+        // On mac desktop, we don't insert new lines via keyboard events.
+        // Because of that, pressing ENTER causes both a "\n" insertion delta
+        // and an onPerformAction call.
+        // We handle the "\n" insertion and ignore the text input action.
+        // If we handle the text input action instead of the "\n" insertion,
+        // pressing ENTER while the textfield has an expanded selection would cause
+        // a new line to be inserted instead of replacing the selection with a new line.
         break;
       case TextInputAction.done:
         widget.focusNode.unfocus();
--- a/super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
+++ b/super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
@@ -1311,7 +1311,7 @@ class _SuperTextFieldImeInteractorState extends State<SuperTextFieldImeInteracto
         // Do nothing for IME newline actions.
         //
         // Mac: Key presses flow, unhandled, to the OS and turn into IME selectors. We handle newlines there.
-        // Windows/Linux: Newlines are inserted by SuperTextField key handlers, instead of IME actions.
+        // Windows/Linux: Key presses flow, unhandled, to the OS and turn into text deltas. We handle newlines there.
         // Android/iOS: This text field implementation is only for desktop, mobile is handled elsewhere.
         break;
       case TextInputAction.done:
--- a/super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
+++ b/super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
@@ -1765,7 +1765,6 @@ const defaultTextFieldImeKeyboardHandlers = <TextFieldKeyboardHandler>[
   // handlers, passing the key combo to the OS on Mac. Place all custom Mac key
   // combos above this handler.
   DefaultSuperTextFieldKeyboardHandlers.sendKeyEventToMacOs,
-  DefaultSuperTextFieldKeyboardHandlers.insertNewlineWhenEnterIsPressed,
   DefaultSuperTextFieldKeyboardHandlers.moveCaretToStartOrEnd,
   DefaultSuperTextFieldKeyboardHandlers.moveUpDownLeftAndRightWithArrowKeys,
   DefaultSuperTextFieldKeyboardHandlers.moveToLineStartWithHome,
--- a/super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
+++ b/super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
@@ -1308,13 +1308,11 @@ class _SuperTextFieldImeInteractorState extends State<SuperTextFieldImeInteracto
   void _onPerformAction(TextInputAction action) {
     switch (action) {
       case TextInputAction.newline:
-        // On mac desktop, we don't insert new lines via keyboard events.
-        // Because of that, pressing ENTER causes both a "\n" insertion delta
-        // and an onPerformAction call.
-        // We handle the "\n" insertion and ignore the text input action.
-        // If we handle the text input action instead of the "\n" insertion,
-        // pressing ENTER while the textfield has an expanded selection would cause
-        // a new line to be inserted instead of replacing the selection with a new line.
+        // Do nothing for IME newline actions.
+        //
+        // Mac: Key presses flow, unhandled, to the OS and turn into IME selectors. We handle newlines there.
+        // Windows/Linux: Newlines are inserted by SuperTextField key handlers, instead of IME actions.
+        // Android/iOS: This text field implementation is only for desktop, mobile is handled elsewhere.
         break;
       case TextInputAction.done:
         widget.focusNode.unfocus();
--- a/super_editor/test/super_editor/supereditor_test_tools.dart
+++ b/super_editor/test/super_editor/supereditor_test_tools.dart
@@ -748,21 +748,3 @@ class FakeSuperEditorScroller implements DocumentScroller {
   @override
   void detach() => throw UnimplementedError();
 }
-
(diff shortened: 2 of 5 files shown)


super_editor/test/super_editor/supereditor_test_tools.dart
```diff
@@ -738,3 +739,21 @@ class FakeSuperEditorScroller implements DocumentScroller {
   @override
   void detach() => throw UnimplementedError();
 }
+
+/// The platform to be used when simulating a keyboard event with `sendKeyEvent`,
```
matthew-carroll, line 743: Why would this platform be different than the `defaultTargetPlatform`? Also, where is this being used?

super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -1288,7 +1288,13 @@ class _SuperTextFieldImeInteractorState extends State<SuperTextFieldImeInteracto
   void _onPerformAction(TextInputAction action) {
     switch (action) {
       case TextInputAction.newline:
-        widget.textController.insertNewline();
+        // On mac desktop, we don't insert new lines via keyboard events.
```
matthew-carroll, line 1291: This says "On mac desktop" but I don't see any platform inspection nearby....

super_editor/test/super_textfield/super_textfield_ime_test.dart
```diff
@@ -683,6 +621,38 @@ void main() {
   });
 }
 
+Future<void> _pressEnterOnSuperTextField(WidgetTester tester) async {
```
matthew-carroll, line 624: Is this actually specific to `SuperTextField`? Why doesn't this apply to `SuperEditor`, too? Also, does this belong in the testing robots package for general use?

super_editor/test/super_textfield/super_textfield_ime_test.dart
```diff
@@ -683,6 +621,38 @@ void main() {
...
+  final handled = await tester.sendKeyEvent(LogicalKeyboardKey.enter, platform: testKeyEventPlatform);
+  if (handled) {
+    // The textfield handled the key event.
+    // It won't bubble up to the OS to generate text deltas or input actions.
+    return;
+  }
+
+  // The ENTER key event wasn't handled by the textfield.
+  // The OS generates both a "\n" insertion and a new line action.
+  await tester.ime.typeText('\n', getter: imeClientGetter);
```
matthew-carroll, line 634: Should we be checking for an open connection instead of assuming that there is one?

super_editor/test/super_editor/supereditor_test_tools.dart
angelosilvestre: It's used in `_pressEnterOnSuperTextField`. I extracted this from `pressEnter`.

super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
angelosilvestre: I'm not inspecting the platform because for other desktop platforms, the new line is inserted via keyboard handlers. Because of that, the IME doesn't reports input actions nor insertion deltas.

If, at some point, we remove the ENTER keyboard handler, we would be in the same situation as on mac...

Should I just update the comment?

super_editor/test/super_textfield/super_textfield_ime_test.dart
angelosilvestre: We could add this to the test robots. Should I open a PR for this?

super_editor/test/super_textfield/super_textfield_ime_test.dart
angelosilvestre: Updated

super_editor/test/super_editor/supereditor_test_tools.dart
matthew-carroll: Why did you extract it instead of use it? And also I'd still like to know why it doesn't use `defaultTargetPlatform`. I don't remember what we did for `pressEnter` originally, or why.

super_editor/test/super_textfield/super_textfield_ime_test.dart
matthew-carroll: Yeah you can open a PR for that. What about the first two questions?

super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
matthew-carroll: Is the purpose of this code comment to explain why we're not doing anything when we get a newline?

If so, does the following comment better capture the situation?

```dart
// Do nothing for IME newline actions
//
// Mac: Key presses flow, unhandled, to the OS and turn into IME selectors. We handle newlines there.
// Windows/Linux: Newlines are inserted by Super Editor key handlers, instead of IME actions, because Flutter doesn't dispatch newline IME actions for these platforms.
// Android/iOS: This text field implementation is only for desktop, mobile is handled elsewhere.
```

Also, can you remind me why we're using key handlers for Windows/Linux and not using the IME actions? Is that because Flutter doesn't send us IME actions for those?

super_editor/test/super_editor/supereditor_test_tools.dart
angelosilvestre: I didn't use `pressEnter` because I had to check whether or not the key was handled in order to generate the input action and the insertion delta.

super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
angelosilvestre: Previously, we used key handlers to insert new line on all desktop platforms. After the `performSelection` implementation, we stopped doing that on mac.

We could probably could do the same on the other desktop platforms. I tested on Windows and we also receive the input actions and insertion deltas if we don't handle the ENTER key.

super_editor/test/super_textfield/super_textfield_ime_test.dart
angelosilvestre: This will apply for `SuperEditor` also. I was looking into implementing it on testing robots `pressEnter`, but then we will need to add an optional `imeClientGetter` parameter. 

To generate deltas, we need to access `DeltaTextInputClient` in order to obtain the current editing value.

super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
matthew-carroll: Ok. Let's try removing the ENTER handler from the IME handler list for Windows and Linux.

super_editor/test/super_textfield/super_textfield_ime_test.dart
matthew-carroll: Can we introduce a `pressEnterOnIme()` to the robots package?

angelosilvestre: Opened https://github.com/Flutter-Bounty-Hunters/flutter_test_robots/pull/20

matthew-carroll: I just released v0.0.22

matthew-carroll:
@angelosilvestre is this ready for review? Or still doing work?

matthew-carroll:
Also, looks like tests are failing, FYI.

angelosilvestre:
@matthew-carroll This is ready for another review now.

matthew-carroll:
Looks like this PR has a merge conflict.

angelosilvestre:
@matthew-carroll Rebased.

super_editor/test/super_editor/supereditor_test_tools.dart
```diff
@@ -1,6 +1,7 @@
 import 'dart:math';
 import 'dart:ui' as ui;
 
+import 'package:flutter/foundation.dart';
```
matthew-carroll, line 4: Do you need this import?

matthew-carroll (approved):
LGTM with removal of one superfluous import

super_editor/test/super_editor/supereditor_test_tools.dart
angelosilvestre: Nope. Removed.

matthew-carroll:
FYI @angelosilvestre CI is going to automatically create a PR when we merge, but you'll have to explicitly close and re-open the PR to get other CI checks to run.

angelosilvestre:
@matthew-carroll I didn't find any PR created for after the merge

matthew-carroll:
Ugh - CI is failing for some reason....complaining that GitHub bot doesn't have permission, which doesn't make sense given that this has worked previously.

https://github.com/superlistapp/super_editor/actions/runs/6552954869/job/17799413806

Issue #1166, opened 2023-06-09
[SuperEditor] - Let users anchor popovers near text tokens


matthew-carroll:
Users should be able to display arbitrary popovers near a text token, e.g., "@someUser", "#someHashTag".

This ticket depends on a couple others:

- Tokenization reactions: https://github.com/superlistapp/super_editor/issues/1077
- Generic anchor system: https://github.com/superlistapp/super_editor/issues/424

matthew-carroll:
CC @miguelcmedeiros @jmatth

matthew-carroll:
CC @Jethro87

Issue #893, opened 2022-12-14
[SuperEditor] Carat, selection handles, and overlays should hide behind above-keyboard toolbars


Jethro87:
They are also visible above the "panel_behind_keyboard" in #876.

https://user-images.githubusercontent.com/6467808/207709646-d90d1f59-fd89-4858-b417-b88c8970fc8e.mov

angelosilvestre:
Currently, for Android and iOS we display the caret and the drag handles in an `OverlayEntry`. We will need to replace the mobile interactor with the layer system, which is already used on desktop.

Jethro87:
@matthew-carroll Now that #1256 has merged, is it possible to show the carat, selection handles, and overlays below a mobile keyboard toolbar?

matthew-carroll:
@Jethro87 we should have at least a couple options. I mentioned those in this ticket: https://github.com/superlistapp/super_editor/issues/398

I think the primary question is whether we want app-level overlays, document overlays, or both.

Issue #424, opened 2022-02-11
Add an anchor system to SuperEditor widget


matthew-carroll:
It should be easier than it is to display popover toolbars and other overlay anchored items.

Currently, developers need to provide a `ScrollController` to `SuperEditor` so that the developer can update the toolbar's location on every frame. This is very tedious and it's likely to fail when there's more separation between `SuperEditor` and the rest of the app behavior.

One idea that I have to deal with this is to pass a map of `DocumentSelection` -> `LayerLink`. The `SuperEditor` widget can maintain a set of invisible `SizedBox` widgets at the locations and sizes represented by the various `DocumentSelection`s and those boxes can be wrapped in a `LinkTarget`.

Then, builder widgets could be used to easily anchor other widgets in the overlay.

```dart
SuperEditorAnchoredBuilder(
  link: _selectionLink,
  overlayBuilder: (context) {
    return EditorToolbar(
      editContext: editContext,
    );
  },
  child: SuperEditor(
     //...
     anchors: {
       // Place an invisible anchor around the user's selection
       _selectionLink: composer.selectionNotifier,
     },
  ),
);
```

jonaird:
Another issue I'm having is that my toolbar is a bit large and it prevents the user from being able to scroll the document when the mouse is on top of the toolbar. This is because the toolbar isn't actually in the scrollable widget. afaik I don't think it's possible to collect raw scroll information?

Another use case is slash commands. For example in Notion, if you type "/h1" a filtered list of blocks shows up and hitting enter will create a new block of that type.

PR #1470, opened 2023-09-26
Update SuperEditor and SuperReader layer system to provide app-level composition (iOS)(Relates to #424, #893, #1166)(Resolves #1508)

### LINKED ISSUE #1508: [SuperEditor][SuperReader] - Make iOS overlay controls configurable through composition and use document layers
Currently, iOS overlay controls, like the caret, drag handles, magnifier, and toolbar, are all positioned and managed within `SuperEditor` and `SuperReader` by an iOS interactor widget. This interactor widget routinely schedules extra frames to calculate the overlay control positions based on the latest document layout. Also, the only way to configure the overlay controls is by way of `SuperEditor` and `SuperReader` widget properties, which is exploding the number of properties.

Update the implementation to accomplish the following:

- Where appropriate, add document layers that position `Leader` widgets for things like handles and toolbars to follow.
- Where appropriate, display controls within a document layer instead of in the app overlay. For example, the iOS caret and drag handles probably make more sense in a document layer than an app overlay.
- For remaining app overlay document controls, move their overlay management into dedicated management widgets within `SuperEditor`, instead of making it the responsibility of the existing touch interactor.
- Make the toolbar builder, magnifier builder, and other relevant iOS control details configurable via composition, instead of using `SuperEditor` and `SuperReader` properties.

Similar Android changes will be handled in a separate ticket.

matthew-carroll:
This PR includes changes to a number of different areas, which are all somewhat related to working with document layers in `SuperEditor` and `SuperReader`. We need to ensure that our layer system makes it possible to achieve obvious goals. This PR stress tests our layers by reworking our iOS caret, handles, magnifier, toolbar, and floating cursor display. This PR also includes some changes to the content layer system to achieve that goal.

## Proving Layer Capabilities by Adjusting iOS Editing Controls
Super Editor displays a number of controls on iOS: caret, drag handles, magnifier, toolbar, and a floating cursor. All of these controls were primarily managed by `IosDocumentTouchInteractor`. This interactor was constantly scheduling frames to inspect the document layout and figure out where to place the various controls. This coordination was complicated, convoluted, fragile, and non-performant. 

To prove the viability of our layer system, I aimed to do a few things:
  1. Move caret and handles into a document layer. They're based on document layer, so they make sense as a document layer. Also, we have a ticket that wants them beneath the onscreen keyboard, which happens automatically with this change - #893
  2. Make caret and drag handles configurable via widget composition instead of creating special `SuperEditor` properties for that purpose.
  3. Make the toolbar and magnifier configurable via widget composition instead of creating special `SuperEditor` properties to build it.

When breaking apart the iOS controls to be more composable, it became clear that relying on `SuperEditor` properties for all such controls was getting out of hand. We'd need to also expose similar controls for Android, and possibly other platforms. Therefore, I shifted the approach for shared control. I created a widget called `SuperEditorIosControlsScope`, which provides access to a `SuperEditorIosControlsController`, which holds the status of the magnifier, the toolbar, and the floating cursor. `SuperEditor` internally adds a `SuperEditorIosControlsScope` above the document layers in the widget tree, so that all document layers have access to a shared context. Also, an app can surround its `SuperEditor` with its own `SuperEditorIosControlsScope` to further share visibility into those controls. This allows apps, for example, to build their own toolbar widget.

### New Floating Cursor Architecture
The previous floating cursor behavior was spread through widgets in a confusing way. This was primarily due to two factors. First, the floating cursor changes originate within the IME system, which is otherwise dedicated to text editing. So it's a weird starting point (but not in our control). Second, our iOS touch interactor was calculating the visual size and offset for the floating cursor, as well as updating the document selection when it moved. But the floating cursor had nothing to do with Flutter gestures, so this was a weird widget to handle those responsibilities.

As of this PR, the responsibility breakdown is as follows:

```
SuperEditor
  SuperEditorIosControlsScope (owns the FloatingCursorController, available to all descendent widgets)
    SuperEditorImeInteractor (reports floating cursor start/move/end from IME)
      EditorFloatingCursor (decides where the floating cursor should appear, paints it, and updates the doc selection)
        IosDocumentTouchInteractor (when floating cursor is active - auto-scrolls up/down)
          IosHandlesDocumentLayer (when floating cursor is active - hides caret/makes it gray, stops it from blinking)
```

Ideally, the auto-scrolling behavior in `IosDocumentTouchInteractor` should probably be handled by `EditorFloatingCursor`. However, I think this change would require that we put the concept of `SuperEditor` scrolling in its own widget, so that all descendent widgets have an opportunity to auto-scroll, or to change the scroll offset. I filed a ticket for this: https://github.com/superlistapp/super_editor/issues/1495

### Composable toolbar and magnifier
The iOS toolbar and magnifier are now configurable.

Example of overriding the default iOS toolbar and magnifier:

```dart
  @override
  void initState() {
    super.initState();
    
    _iosControlsController = SuperEditorIosControlsController(
      toolbarBuilder: _buildIosFloatingToolbar,
      magnifierBuilder: _buildIosMagnifier,
    );
  }

  @override
  void dispose() {
    _iosControlsController.dispose();

    super.dispose();
  }

  @override
    return SuperEditorIosControlsScope(
      controlsContext: _iosControlsController,
      child: SuperEditor(
         // ...
      ),
    );
  }


  Widget _buildIosToolbar(BuildContext context) {
    return IOSTextEditingFloatingToolbar(
      onCutPressed: _cut,
      onCopyPressed: _copy,
      onPastePressed: _paste,
      focalPoint: _selectionLayerLinks.expandedSelectionBoundsLink,
    );
  }

  Widget _buildIosMagnifier(BuildContext context, LeaderLink focalPoint) {
    return Center(
      child: IOSFollowingMagnifier.roundedRectangle(
        leaderLink: focalPoint,
        offsetFromFocalPoint: const Offset(0, -72),
      ),
    );
  }
```

### Editors vs Reader Controls Scope
Editors have some controls that readers don't have, e.g., caret and floating cursor. Therefore, `SuperEditor` and `SuperReader` should use different types of controllers, and therefore different types of scopes for overlay controls.

Super Editor has:
```dart
SuperEditorIosControlsScope - widget that owns a controller and can be used to customized the controls.

SuperEditorIosControlsController - owns the state of overlay controls: handle color, floating cursor, toolbar builder, etc.

SuperEditorIosToolbarOverlayManager - widget inside of SuperEditor that adds and removes the toolbar from the app overlay.

SuperEditorIosMagnifierOverlayManager - widget inside of SuperEditor that adds and removes the magnifier from the app overlay.

SuperEditorIosToolbarFocalPointDocumentLayerBuilder - document layer widget that places a rectangle around the selection as a focal point for the toolbar overlay widget.

SuperEditorIosHandlesDocumentLayerBuilder - document layer widget that displays the caret and drag handles
```

Super Reader has:
```dart
SuperReaderIosControlsScope - same as above but for SuperReader

SuperReaderIosControlsController

SuperReaderIosToolbarOverlayManager

SuperReaderIosMagnifierOverlayManager

SuperReaderIosToolbarFocalPointDocumentLayerBuilder

SuperReaderIosHandlesDocumentLayerBuilder
```

## Additions and Changes to Help with Editor/Reader Document Layers
In theory, all apps could use existing content layer widgets, like `ContentLayerStatelessWidget` and `ContentLayerStatefulWidget` to implement document layers. However, all such layers would need to obtain a reference to the content `StatefulElement`, obtain its `State`, and then type cast it to a `DocumentLayout`. This PR introduces
 `DocumentLayoutLayerStatelessWidget` and `DocumentLayoutLayerStatefulWidget`, which do this work on the app's behalf and make it easy to quickly position content based on a `DocumentLayout`.

I utilized the new `DocumentLayoutLayerStatefulWidget` to simplify the existing `SelectionLeadersDocumentLayer` implementation.

## Content Layer Changes
During work on this ticket, I discovered that my `StatefulWidget` layers were constantly re-inflating their widget on every frame. This rendered the `State` object useless. I altered `ContentLayersElement` to use `forgetChild()` on the layers instead of `deactivateChild()`, which allows `ContentLayersElement` to avoid building those widgets, while also retaining the associated `Element` and `State` of each layer so they can be re-added to the tree later in the frame.

A `ContentLayerState` can query its cached layout information, instead of being forced to only receive it in `doBuild()`. This allows for event callbacks, such as user interactions, to operate on that same cached layout data.

`ContentLayerState` used to provide its associated `RenderObject` to `computeLayoutData()` but now it provides both its `Element` and `RenderObject`. This change is used in this PR, for example, to access the `State` of a `StatefulElement` and then interact with that `State` as a `DocumentLayout`.

## Changes to text selection bounds calculations
~Previously, calculating a selection rectangle at a collapsed selection (e.g., at a caret location) returned a zero-sized rectangle. In general, it's more useful to return a zero-width rectangle that's as tall as the nearest character. I changed the calculation to return the latter.~

The above statement is no longer true, because I copied that position rectangle behavior into a long-press selection PR that's already been merged. So the behavior exists, but this PR isn't the first place that it was added.

## Minor Tasks in this PR:
 * Added padding to Example App floating action buttons to keep them above mobile keyboard controls.
 * Migrated iOS magnifier placement from Flutter's `LayerLink` to `follow_the_leader` `LeaderLink`.
 * Adjusted control character test so that it correctly ensures focus doesn't move. As a result I removed the test for TAB because TAB probably should move focus by default. We may be missing TAB behavior, but we can add that later.
 * Added tests to ensure that toolbar and magnifier are hidden on web for iOS. That original work was done in https://github.com/superlistapp/super_editor/pull/1483 but no tests were added and this PR completely changed the toolbar and magnifier code, so we needed tests to ensure it wasn't broken by this PR.


matthew-carroll:
@angelosilvestre this is a really big PR. If you want to review it in an unusual way, let me know. We can review by behavior, or by small groups of files. We can also jump on a call at some point if you think that's necessary for the review.

I'd like to get this merged ASAP, but I also know that this review is gonna be a lot of work.

matthew-carroll:
CC @miguelcmedeiros @brian-superlist @Jethro87 @jmatth - this PR changes the architecture for `SuperEditor` and `SuperReader` configure and position the caret, handles, floating cursor, magnifier, and toolbar. Related to those changes, this PR also makes changes to the existing gesture interactor, because I moved all the non-gesture behavior out of there into other places, as mentioned in this PR description.

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
```diff
@@ -1,133 +1,262 @@
...
+  /// Whether the iOS magnifier should be displayed right now.
+  final shouldShowMagnifier = ValueNotifier<bool>(false);
+
+  /// Link to a location where a magnifier should be focused.
+  final magnifierFocalPoint = LeaderLink();
+
+  /// (Optional) Builder to create the visual representation of the magnifier.
+  ///
+  /// If [magnifierBuilder] is `null`, a default iOS magnifier is displayed.
+  final DocumentMagnifierBuilder? magnifierBuilder;
```
angelosilvestre, line 145: Do this and `toolbarBuilder` belong here?

When we see "Controller" in the name of a class is Flutter, usually we don't expect to have widget builders inside of it.

angelosilvestre (commented):
I didn't look through all of the code yet, but I left some initial comments.

super_editor/example/lib/demos/editor_configs/demo_mobile_editing_ios.dart
```diff
@@ -53,17 +67,23 @@ class _MobileEditingIOSDemoState extends State<MobileEditingIOSDemo> {
 
   void _cut() {
     _docOps.cut();
+    // TODO: get rid of overlay controller once Android is refactored to use a control scope (as follow up to: https://github.com/superlistapp/super_editor/pull/1470)
     _overlayController.hideToolbar();
+    _iosEditorControlsController.shouldShowToolbar.value = false;
```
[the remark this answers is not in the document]
matthew-carroll: I went ahead and switched all toolbar and magnifier changes to explicit show/hide/toggle methods instead of setting a boolean.

super_editor/example/lib/demos/interaction_spot_checks/toolbar_following_content_in_layer.dart
```diff
@@ -0,0 +1,207 @@
...
+  void didChangeDependencies() {
+    super.didChangeDependencies();
+
+    _toolbarEntry = OverlayEntry(builder: (_) {
+      return _buildToolbarOverlay();
+    });
+
+    WidgetsBinding.instance.addPostFrameCallback((timeStamp) {
+      Overlay.of(context).insert(_toolbarEntry);
+    });
```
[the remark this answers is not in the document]
matthew-carroll: I reworked the behavior and added a comment explanation. The answer is that the `Overlay` is a dependency and therefore it can change whenever our dependencies change.

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
```diff
@@ -1,133 +1,262 @@
...
   // The actual ScrollPosition that's used for the document layout, either
   // the Scrollable installed by this interactor, or an ancestor Scrollable.
   ScrollPosition? _activeScrollPosition;
 
-  // OverlayEntry that displays editing controls, e.g.,
-  // drag handles, magnifier, and toolbar.
-  OverlayEntry? _controlsOverlayEntry;
-  late IosDocumentGestureEditingController _editingController;
-  final _magnifierFocalPointLink = LayerLink();
+  SuperEditorIosControlsController? _controlsContext;
```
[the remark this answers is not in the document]
matthew-carroll: When dependencies update, this reference might change. If it does, we need to remove some listeners. But `didChangeDependencies` also gets called initially when we don't have any reference at all. The only way that we can remove listeners without triggering an NPE is if we can check of `_controlsContext` is `null`. If we declare it `late` and non-null, then we'll trigger an exception if we try to check null.

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
```diff
@@ -199,6 +295,15 @@ class _IOSDocumentTouchInteractorState extends State<IOSDocumentTouchInteractor>
   void didChangeDependencies() {
     super.didChangeDependencies();
 
+    if (_controlsContext != null) {
+      _controlsContext!.floatingCursorController.removeListener(_floatingCursorListener);
+      _controlsContext!.floatingCursorController.cursorGeometryInViewport
+          .removeListener(_onFloatingCursorGeometryChange);
+    }
+    _controlsContext = SuperEditorIosControlsScope.rootOf(context);
```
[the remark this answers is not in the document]
matthew-carroll: Ideally we'd use the nearest because that's more performant. You've described a solution for the case of two scopes, one inside `SuperEditor` and one outside `SuperEditor`. But what about when there's 3, 4, or 5 scopes? Which one do we use? The nearest or the root? I think the assumption of such a scope is that the outermost scope is the one that takes priority...

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
matthew-carroll: Yeah. A "controller" is any object that shares state and control between the inside of a widget, and the outside of a widget. These builders allow code that's outside of `SuperEditor` to control how something is built within `SuperEditor`. While those could be direct properties on `SuperEditor`, as I mentioned in the PR description, if we make all of these things direct properties then we're going to balloon the `SuperEditor` constructor with a mess of iOS, Android, Mac, etc. properties.

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
```diff
@@ -508,12 +505,16 @@ class _IOSDocumentTouchInteractorState extends State<IOSDocumentTouchInteractor>
     _globalTapDownOffset = details.globalPosition;
     _tapDownLongPressTimer?.cancel();
     _tapDownLongPressTimer = Timer(kLongPressTimeout, _onLongPressDown);
+
+    // Stop the caret from blinking, in case this tap down turns into a long-press drag,
+    // or a caret drag.
+    _controlsContext!.shouldCaretBlink.value = false;
```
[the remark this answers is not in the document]
matthew-carroll: I switched to a `blinkCaret()` and `dontBlinkCaret()`

super_editor/lib/src/infrastructure/platforms/ios/ios_document_controls.dart
```diff
@@ -656,16 +266,601 @@ class IosDocumentGestureEditingController extends GestureEditingController {
...
+          widget.document.selectUpstreamPosition(selection.base, selection.extent),
+        )!,
+        downstream: documentLayout.getRectForPosition(
+          widget.document.selectDownstreamPosition(selection.base, selection.extent),
+        )!,
+        expandedSelectionBounds: documentLayout.getRectForSelection(
+          selection.base,
+          selection.extent,
+        ),
+      );
```
angelosilvestre, line 695: Just a personal preference: if we are returning inside the `if` branch we don't need the `else`.

You can ignore this if you prefere this way :)

matthew-carroll: When we're dealing with a guard clause, where we weed out an exceptional situation and return, then I agree. However, in this case, both conditional branches are of equal weight. There's nothing exceptional about a collapsed selection. So I retained an if-else statement.

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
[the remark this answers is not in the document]
matthew-carroll: But how does that policy facilitate our users? Imagine that Package includes a `SuperEditor` and wraps it with an iOS controls scope to customize the controls. Then App uses Package which uses SuperEditor. Now the app wants to further change those controls by wrapping with another iOS controls scope. That won't have any effect if we look for the nearest ancestor.

What would App do to customize the controls within Package, which customized the controls within SuperEditor?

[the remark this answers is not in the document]
angelosilvestre: Understood, with the given example it makes sense to pick the outermost scope.

super_editor/lib/src/infrastructure/platforms/ios/toolbar.dart
```diff
@@ -67,6 +73,11 @@ class IOSTextEditingFloatingToolbar extends StatelessWidget {
     required String title,
     required VoidCallback onPressed,
   }) {
+    // return CupertinoPopoverToolbarMenuItem(
+    //   label: title,
+    //   onPressed: onPressed,
+    // );
+
```
angelosilvestre, line 80: Please remove the commented code.

super_editor/lib/src/super_reader/super_reader.dart
```diff
@@ -332,103 +365,296 @@ class SuperReaderState extends State<SuperReader> {
...
+          autofocus: widget.autofocus,
+          child: _buildPlatformSpecificViewportDecorations(
+            controlsScopeContext,
+            child: DocumentScaffold(
+              documentLayoutLink: _documentLayoutLink,
+              documentLayoutKey: _docLayoutKey,
+              gestureBuilder: _buildGestureInteractor,
+              scrollController: _scrollController,
+              autoScrollController: _autoScrollController,
+              // TODO: Finish integrating the DocumentScroller in SuperReader (https://github.com/superlistapp/super_editor/issues/1306)
```
angelosilvestre, line 388: Is this comment out of date? The linked issue is already closed.

super_editor/lib/src/super_reader/super_reader.dart
```diff
@@ -332,103 +365,296 @@ class SuperReaderState extends State<SuperReader> {
...
+      //   return IOSEditorControlsContext(
+      //     data: data,
+      //     child: child,
+      //   );
+      // case DocumentGestureMode.android:
+      //   // TODO: create context for Android
+      //   return IOSEditorControlsContext(
+      //     data: data,
+      //     child: child,
+      //   );
```
angelosilvestre, line 438: Please remove the commented code.

super_editor/lib/src/super_reader/read_only_document_ios_touch_interactor.dart
```diff
@@ -252,95 +305,44 @@ class _ReadOnlyIOSDocumentTouchInteractorState extends State<ReadOnlyIOSDocument
...
-      _removeEditingOverlayControls();
-    }
+  Offset _documentOffsetToViewportOffset(Offset documentOffset) {
+    final globalOffset = _docLayout.getGlobalOffsetFromDocumentOffset(documentOffset);
+    return viewportBox.globalToLocal(globalOffset);
   }
 
   void _onDocumentChange(_) {
-    _editingController.hideToolbar();
+    _controlsContext!.shouldShowToolbar.value = false;
```
angelosilvestre, line 345: Same comment here from previous review about adding a method to show/hide instead of directly mutating the value.

super_editor/lib/src/super_textfield/ios/ios_textfield.dart
```diff
@@ -596,7 +597,8 @@ class SuperIOSTextFieldState extends State<SuperIOSTextField>
 
 Widget _defaultPopoverToolbarBuilder(BuildContext context, IOSEditingOverlayController controller) {
   return IOSTextEditingFloatingToolbar(
-    focalPoint: controller.overlayController.toolbarTopAnchor!,
+    // TODO: bring back the focal point
+    // focalPoint: controller.overlayController.toolbarTopAnchor!,
```
angelosilvestre, line 601: Is this TODO meant to be addressed in this PR ?

super_editor/lib/src/test/super_editor_test/supereditor_inspector.dart
```diff
@@ -95,10 +96,13 @@ class SuperEditorInspector {
       return androidControls.editingController.caretTop!;
     }
 
-    final iOSControls =
-        find.byType(IosDocumentTouchEditingControls).evaluate().lastOrNull?.widget as IosDocumentTouchEditingControls?;
-    if (iOSControls != null) {
-      return iOSControls.editingController.caretTop!;
+    final iOSControls = (find.byType(IosHandlesDocumentLayer).evaluate().lastOrNull as StatefulElement?)?.state
+        as IosControlsDocumentLayerState?;
+    // find.byType(IosEditingToolbarOverlay).evaluate().lastOrNull?.widget as IosEditingToolbarOverlay?;
```
angelosilvestre, line 101: Please remove the commented code.

super_editor/test/super_editor/mobile/super_editor_ios_overlay_controls_test.dart
```diff
@@ -0,0 +1,111 @@
...
+void main() {
+  group("SuperEditor > iOS > overlay controls >", () {
+    group("on device and web > shows ", () {
+      testWidgetsOnIosDeviceAndWeb("caret", (tester) async {
+        await _pumpApp(tester);
+
+        // Create a collapsed selection.
+        await tester.tapInParagraph("1", 1);
+
+        // Ensure we have a collapsed.
```
angelosilvestre, line 17: "Ensure we have a collapsed selection."

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -378,6 +395,8 @@ class SuperEditorState extends State<SuperEditor> {
       _composer = widget.composer;
     }
 
+    // TODO: update the platform gesture controls controller
```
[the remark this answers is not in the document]
matthew-carroll: I'm not sure what I meant by that. Maybe it was from a transitory approach. That controller never changes within `SuperEditor`, though a different scope might be used. I removed the comment.

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -521,51 +545,99 @@ class SuperEditorState extends State<SuperEditor> {
...
+      //   return IOSEditorControlsContext(
+      //     data: data,
+      //     child: child,
+      //   );
+      // case DocumentGestureMode.android:
+      //   // TODO: create context for Android
+      //   return IOSEditorControlsContext(
+      //     data: data,
+      //     child: child,
+      //   );
```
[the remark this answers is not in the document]
matthew-carroll: I removed the commented returns, but left the cases and TODOs. Those need to be done ASAP after we merge this.

super_editor/lib/src/infrastructure/platforms/ios/toolbar.dart
matthew-carroll: @angelosilvestre I checked why I commented that out. For some reason the button text is white on light for `CupertinoPopoverToolbarMenuItem`. Rather than deal with that at the time, I recreated the button. Do you know why the theming is broken there?

super_editor/lib/src/super_reader/super_reader.dart
matthew-carroll: Hm, maybe that was some kind of merge conflict resolution issue. I'll remove the comment.

super_editor/lib/src/super_reader/read_only_document_ios_touch_interactor.dart
matthew-carroll: Done

super_editor/lib/src/infrastructure/platforms/ios/toolbar.dart
angelosilvestre: Looking at the code, `CupertinoPopoverToolbarMenuItem` has a hardcoded `Colors.white` as the text color.

matthew-carroll: Ok, can you file an issue for us to honor theming in the various overlay controls? Then I'll add a TODO to bring back the commented code when that issue is resolved.

angelosilvestre: Filed https://github.com/Flutter-Bounty-Hunters/overlord/issues/17

super_editor/lib/src/super_textfield/ios/ios_textfield.dart
matthew-carroll: I did some non-trivial work to bring back the focal point and migrate the arrow focus to Leader/Follower. Pushed the change.

super_editor/lib/src/infrastructure/platforms/ios/toolbar.dart
matthew-carroll: Thanks.

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -601,76 +663,152 @@ class SuperEditorState extends State<SuperEditor> {
...
+  }) {
+    switch (gestureMode) {
+      case DocumentGestureMode.iOS:
+        return SuperEditorIosToolbarOverlayManager(
+          defaultToolbarBuilder: (overlayContext, mobileToolbarKey, focalPoint) => defaultIosEditorToolbarBuilder(
+            overlayContext,
+            mobileToolbarKey,
+            focalPoint,
+            editContext.commonOps,
+            SuperEditorIosControlsScope.rootOf(context),
```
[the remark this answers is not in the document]
matthew-carroll: It looks like this is the only access, so I think I'll leave it as-is. If we notice a performance issue then we can cache it.

super_editor/lib/src/infrastructure/platforms/ios/ios_document_controls.dart
```diff
@@ -656,16 +266,601 @@ class IosDocumentGestureEditingController extends GestureEditingController {
...
+    // The current selection is expanded, or we went from expanded in the previous frame
+    // to non-expanded in this frame. Either way, we need to recalculate the toolbar focal
+    // point bounds.
+    if (mounted && SchedulerBinding.instance.schedulerPhase != SchedulerPhase.persistentCallbacks) {
+      // The Flutter pipeline isn't running. Schedule a re-build to re-position the toolbar focal point.
+      setState(() {
+        // The selection bounds, and Leader build, will take place in methods that
+        // run in response to setState().
+      });
+    }
```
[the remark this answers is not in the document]
matthew-carroll: I switched to our existing method `setStateAsSoonAsPossible()`

angelosilvestre (approved):
LGTM

matthew-carroll:
@angelosilvestre do you feel like you have a good idea about what this PR is trying to achieve and how it does it? I want to make sure that you really feel good about what's here before merging and moving forward with the other platforms.

matthew-carroll:
I'm gonna go ahead and merge this to move on to Android. @angelosilvestre please let me know if you have any architectural concerns about this approach.

PR #1471, opened 2023-09-26
[SuperEditor][iOS] Fix exception when moving floating cursor between paragraphs (Resolves #1449)

### LINKED ISSUE #1449: [SuperEditor][iOS] IME mapping error while dragging floating cursor between paragraphs
### Steps to reproduce

1. Run the example app on iOS
2. Double tap the word "more"
3. Press and hold the spacebar to show the floating cursor
4. Drag the floating cursor to the word "made" in the next paragraph

### Expected results

The selection moves without exceptions

### Actual results

The selection moves as expected, but the following exception is displayed in the console:

```console
Restarted application in 276ms.
flutter: Initializing logger: ExampleApp

════════ Exception caught by services library ══════════════════════════════════
Exception: Couldn't map an IME position to a document position. 
TextEditingValue: 'Ready-made solutions 📦'
IME position: TextPosition(offset: 98, affinity: TextAffinity.downstream)

#0      DocumentImeSerializer._imeToDocumentPosition (package:super_editor/src/default_editor/document_ime/document_serialization.dart:268:5)
#1      DocumentImeSerializer.imeToDocumentSelection (package:super_editor/src/default_editor/document_ime/document_serialization.dart:161:18)
#2      TextDeltasDocumentEditor._applyNonTextChange (package:super_editor/src/default_editor/document_ime/document_delta_editing.dart:232:41)
#3      TextDeltasDocumentEditor.applyDeltas (package:super_editor/src/default_editor/document_ime/document_delta_editing.dart:85:9)
#4      DocumentImeInputClient.updateEditingValueWithDeltas (package:super_editor/src/default_editor/document_ime/document_ime_communication.dart:203:30)
#5      DeltaTextInputClientDecorator.updateEditingValueWithDeltas (package:super_editor/src/default_editor/document_ime/ime_decoration.dart:129:14)
#6      TextInput._handleTextInputInvocation (package:flutter/src/services/text_input.dart:1873:63)
#7      TextInput._loudlyHandleTextInputInvocation (package:flutter/src/services/text_input.dart:1762:20)
#8      MethodChannel._handleAsMethodCall (package:flutter/src/services/platform_channel.dart:552:55)
#9      MethodChannel.setMethodCallHandler.<anonymous closure> (package:flutter/src/services/platform_channel.dart:545:34)
#10     _DefaultBinaryMessenger.setMessageHandler.<anonymous closure> (package:flutter/src/services/binding.dart:567:35)
#11     _invoke2 (dart:ui/hooks.dart:345:13)
#12     _ChannelCallbackRecord.invoke (dart:ui/channel_buffers.dart:45:5)
#13     _Channel.push (dart:ui/channel_buffers.dart:135:31)
#14     ChannelBuffers.push (dart:ui/channel_buffers.dart:343:17)
#15     PlatformDispatcher._dispatchPlatformMessage (dart:ui/platform_dispatcher.dart:689:22)
#16     _dispatchPlatformMessage (dart:ui/hooks.dart:257:31)

call: MethodCall(TextInputClient.updateEditingStateWithDeltas, [1, {deltas: [{selectionBase: 98, oldText: Super Editor is a toolkit to help you build document editors, document layouts, text fields, and more., selectionAffinity: TextAffinity.downstream, deltaEnd: -1, deltaText: , deltaStart: -1, composingExtent: -1, selectionIsDirectional: false, selectionExtent: 98, composingBase: -1}]}])

```

angelosilvestre:
[SuperEditor][iOS] Fix exception when moving floating cursor between paragraphs. Resolves #1449

Moving the floating cursor between paragraphs causes an exception when the selection sits at an offset bigger than the newly selected paragraph length.

The cause is that the IME generates non-text deltas while the floating cursor is moving. After the selection already changed to the other paragraph, we receive a non-text delta for the previously selected paragraph. This causes an IME mapping error.

This PR changes `DocumentImeInputClient` to ignore deltas while the floating cursor is being displayed.

Related flutter ticket: https://github.com/flutter/flutter/issues/134989

### DIFF
--- a/super_editor/test/super_editor/supereditor_floating_cursor_test.dart
+++ b/super_editor/test/super_editor/supereditor_floating_cursor_test.dart
@@ -1,10 +1,13 @@
 import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
 import 'package:flutter_test/flutter_test.dart';
+import 'package:flutter_test_robots/flutter_test_robots.dart';
 import 'package:flutter_test_runners/flutter_test_runners.dart';
 import 'package:super_editor/src/core/document.dart';
 import 'package:super_editor/src/core/document_selection.dart';
 import 'package:super_editor/src/default_editor/text.dart';
 import 'package:super_editor/src/infrastructure/blinking_caret.dart';
+import 'package:super_editor/src/test/ime.dart';
 import 'package:super_editor/src/test/super_editor_test/supereditor_inspector.dart';
 import 'package:super_editor/src/test/super_editor_test/supereditor_robot.dart';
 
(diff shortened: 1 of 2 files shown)


matthew-carroll:
> This PR changes DocumentImeInputClient to ignore deltas while the floating cursor is being displayed.

This sounds like a problem waiting to happen. What if content actually changes while the floating cursor is moving? For example, what would happen in a multi-user experience when another user is adding or deleting text while the primary user happens to be moving a floating cursor?

angelosilvestre:
But in a multi user experience, changes from other users won't generate IME deltas, right? 

We could also be cautious and only ignore if the delta list contains only non-text deltas

matthew-carroll:
Ok, let's try ignoring non-text deltas in that case. But, I think we still need to be careful about which ones we ignore. For example, it's possible that a non-text delta is used in the middle of editing deltas:

Set selection
Delete
Insert
Set selection
Insert

Even if we're moving the floating cursor, we can't ignore those non-text selection deltas, because those selection changes are being used to facilitate the editing.

I think we probably only want to ignore non-text deltas when the delta batch ONLY has non-text deltas.

What do you think?

angelosilvestre:
> I think we probably only want to ignore non-text deltas when the delta batch ONLY has non-text deltas.
What do you think?

I agree. In fact, I pushed a commit yesterday with that change. With this change we will only ignore the batch if it contains only non-text deltas.

super_editor/test/super_editor/supereditor_floating_cursor_test.dart
```diff
@@ -180,6 +183,50 @@ void main() {
...
+        await tester.ime.sendDeltas([
+          const TextEditingDeltaNonTextUpdate(
+            oldText: 'This is the first paragraph',
+            selection: TextSelection.collapsed(offset: 27),
+            composing: TextRange.empty,
+          )
+        ], getter: imeClientGetter);
+        await tester.pump();
+
+        // Ensure the selection changed to the end of the second paragraph.
```
matthew-carroll, line 219: Did you confirm that this test fails without the changes you made?

angelosilvestre: Yes, I was able to trigger a test failure removing the changes.

matthew-carroll (approved):
LGTM

PR #1500, opened 2023-10-12
[SuperEditor] Fix list items indentation (Resolves #1497)

### LINKED ISSUE #1497: [SuperEditor] Tab does not change the indent of a list item node
Steps to reproduce:
* Run demo app of `SuperEditor`
* Place the selection on a list item node
* Press Tab key

**Expected:** the indentation of the list item node changes
**Actual:** the editor loses focus

angelosilvestre:
[SuperEditor] Fix list items indentation. Resolves #1497

List items indentation is broken on mac when the widget tree has other focusable widgets. The reason is that pressing TAB dispatches a `NextFocusIntent`, which performs focus traversal and prevents the OS from receiving the key event to generate selectors.

This PR  adds `NextFocusIntent` to the list of blocked intents.

We already have tests for list indenting, but as `SuperEditor` is the only focusable widget in the tree  we weren't catching this situation. This PR updates the list indenting tests to add a `TextField` in the widget tree. With this change, I was able to trigger a test failure.

I also replaced `testWidgetsOnArbitraryDesktop` with `testWidgetsOnDesktop` because now we have two separate code paths to indenting list items: 
- Using selectors on mac
- Using key events on other platforms


matthew-carroll (approved):
LGTM

PR #1507, opened 2023-10-14
Created a SuperTextFieldContext with a BuildContext and a TextFieldScroller - passed it key handlers (Resolves #1487)

### LINKED ISSUE #1487: [SuperTextField] Introduce a scrolling api
**Description:**

Currently, `SuperTextField` doesn't have a proper scrolling api which exposes its scrolling artefacts. Due to the absence of such api, we don't have a single source that could alter the scrolling of the `SuperTextField` content and we've to rely on manually passing the scrolling artefacts through constructors at the places they're needed. This is resulting in boilerplate code that could be avoided by addition of an scrolling api that exposes this artefacts wherever needed. 

Compared to `SuperTextField`, `SuperEditor` has its own scrolling api that exposes the scrolling artefacts and scroll actions on `SuperEditor` through `SuperEditorContext`.  We could try to introduce something like this for `SuperTextField`. 

This scrolling api could prove helpful in case of `SuperDesktopField` where we've to scroll the textfield as well as the ancestor document when scroll actions are performed on certain keyboard shortcuts( #1463  ). Currently, we're passing the scrolling artefacts created within `SuperDesktopField` that manages its scrolling, directly to the scroll key handlers which has resulted in addition of boilerplate code that could be avoided by such api.

**Api requirements:**

I'm not sure abt the full specs of this api but some off the top of my mind are,

1. Access to `ScrollController` that controls textfield: This could be tricky. Currently there's no particular scroll controller within `SuperTextField` for this. But rather, `SuperDesktopField`, `SuperAndroidTextField` and `SuperIOSTextField` has their own independent ways to manage the scrolling of textfield within them which is not known to `SuperTextField`.

2. Access to ancestor scrollable if one's present.
3. Able to scroll content within the `SuperTextField`.
4. Able to scroll the ancestor scrollable if one's present.

matthew-carroll:
Created a SuperTextFieldContext with a BuildContext and a TextFieldScroller - passed it key handlers (Resolves #1487)

### DIFF
--- a/super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
+++ b/super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
@@ -16,8 +16,9 @@ import 'package:super_editor/src/infrastructure/keyboard.dart';
 import 'package:super_editor/src/infrastructure/multi_listenable_builder.dart';
 import 'package:super_editor/src/infrastructure/multi_tap_gesture.dart';
 import 'package:super_editor/src/infrastructure/platforms/mac/mac_ime.dart';
-import 'package:super_editor/src/super_textfield/super_textfield.dart';
 import 'package:super_editor/src/infrastructure/text_input.dart';
+import 'package:super_editor/src/super_textfield/infrastructure/text_field_scroller.dart';
+import 'package:super_editor/src/super_textfield/super_textfield.dart';
 import 'package:super_text_layout/super_text_layout.dart';
 
 import '../infrastructure/fill_width_if_constrained.dart';
@@ -150,8 +151,10 @@ class SuperDesktopTextFieldState extends State<SuperDesktopTextField> implements
   late FocusNode _focusNode;
   bool _hasFocus = false; // cache whether we have focus so we know when it changes
 
+  late SuperTextFieldContext _textFieldContext;
   late ImeAttributedTextEditingController _controller;
   late ScrollController _scrollController;
+  late TextFieldScroller _textFieldScroller;
 
   double? _viewportHeight;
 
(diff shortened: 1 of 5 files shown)


matthew-carroll:
@rutvik110 - please let me know if this API would work for your key handlers that scroll text fields

rutvik110:
This looks really great! I'll give it a try with `SuperDesktopTextField` scroll handlers.

rutvik110:
Just tried this updates on the `SuperDesktopTextField` scroll key handlers, works nicely! 

Beside this, I identified an issue when we are setting `TextInputSource` to `ime` for textfields, an extra line is being added into the textfield on enter. This behaviour is seen across both `SuperTextField` and `SuperDesktopField`. 

https://github.com/superlistapp/super_editor/assets/65209850/66e8c2c3-cabf-4c58-87f1-14e0f93d0eb2

angelosilvestre:
@rutvik110 We have an ongoing PR to fix that issue: https://github.com/superlistapp/super_editor/pull/1469

super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -171,6 +174,10 @@ class SuperDesktopTextFieldState extends State<SuperDesktopTextField> implements
     _controller.addListener(_onSelectionOrContentChange);
 
     _scrollController = ScrollController();
+    _textFieldScroller = TextFieldScroller() //
+      ..attach(_scrollController);
```
angelosilvestre, line 178: It seems we are not detaching from the scroll controller anywhere.

angelosilvestre (approved):
LGTM with a single comment.

PR #1511, opened 2023-10-14
[SuperEditor] Bumped http lower bound to v1.0.0 (Resolves #1464)

### LINKED ISSUE #1464: super_editor is incompatible with google_fonts due to http version
Because every version of super_editor depends on http ^0.13.1 and google_fonts >=4.0.5 depends on http ^1.0.0, super_editor is incompatible with google_fonts >=4.0.5.
So, because todoify depends on both google_fonts ^6.0.0 and super_editor ^0.2.6, version solving failed.
exit code 1

I checked out the super_editor in code and the version constraint is still 0.13.1. Could you please upgrade it?

rutvik110:
[SuperEditor] Bumped http lower bound to v1.0.0, example deps updated (Resolves #1464)

This PR updates the lower bound of http version from 0.13.1 to 1.0.0

Also, updated super editor example deps that got affected due to this update.

### DIFF
--- a/super_editor/pubspec.yaml
+++ b/super_editor/pubspec.yaml
@@ -25,7 +25,7 @@ dependencies:
   characters: ^1.2.0
   collection: ^1.15.0
   follow_the_leader: ^0.0.4+3
-  http: ">=0.13.1 <2.0.0"
+  http: ">=1.0.0 <2.0.0"
   linkify: ^5.0.0
   logging: ^1.0.1
   super_text_layout: ^0.1.7
--- a/super_editor/example/pubspec.yaml
+++ b/super_editor/example/pubspec.yaml
@@ -35,8 +35,8 @@ dependencies:
 
   charcode: ^1.1.3
   flutter_keyboard_visibility: ^5.0.3
-  google_fonts: ^2.0.0
-  http: ^0.13.1
+  google_fonts: ^6.0.0
+  http: ^1.0.0
   linkify: ^5.0.0
   logging: ^1.0.1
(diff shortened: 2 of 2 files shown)


super_editor/pubspec.yaml
```diff
@@ -25,7 +25,7 @@ dependencies:
   characters: ^1.2.0
   collection: ^1.15.0
   follow_the_leader: ^0.0.4+3
-  http: ">=0.13.1 <2.0.0"
+  http: ">=1.0.0 <2.0.0"
```
matthew-carroll, line 28: This is tightening the constraint, not loosening it. Does this actually solve the problem? If users need `http` v1.0.0, why doesn't the existing constraints allow for that? isn't `v1.0.0` >= `v0.13.1` and < `v2.0.0`?

rutvik110: Yes that's correct, it does tightens the constraints.

On the reported issue, I was able to reproduce it within super editor example when referring the `super_editor` through pub.dev but wasn't able to identify why the loose constraints didn't satisfy the dependencies. 
Referring the `super_editor` package that was present locally didn't seemed to introduced this issue when resolving deps which is unclear to me why.

This bump may be a wrong decision on my side but I'll have to give it another look to get proper answer to this and ensure if this would resolve the issue or if it's something  the users could fix on their end itself.

rutvik110:
Closing this PR as currently there doesn't seem to be any need for this update to resolve the reported issue #1464. The issue is likely related to something else which is not clear at this point.

PR #1524, opened 2023-10-16
Mobile mounted toolbar now appears above Android overlay controls, migrated mounted toolbar and Android controls to OverlayPortal (Resolves #893)(Resolves #1510)

### LINKED ISSUE #893: [SuperEditor] Carat, selection handles, and overlays should hide behind above-keyboard toolbars

### LINKED ISSUE #1510: [SuperEditor][SuperReader][SuperTextField] - Consider using OverlayPortal for overlay controls
The docs for `OverlayPortal` says:

> The [OverlayPortal](https://api.flutter.dev/flutter/widgets/OverlayPortal-class.html) uses [overlayChildBuilder](https://api.flutter.dev/flutter/widgets/OverlayPortal/overlayChildBuilder.html) to build its overlay child and renders it on the specified [Overlay](https://api.flutter.dev/flutter/widgets/Overlay-class.html) as if it was inserted using an [OverlayEntry](https://api.flutter.dev/flutter/widgets/OverlayEntry-class.html), while it can depend on the same set of [InheritedWidget](https://api.flutter.dev/flutter/widgets/InheritedWidget-class.html)s (such as [Theme](https://api.flutter.dev/flutter/material/Theme-class.html)) that this widget can depend on.

`SuperEditor`, `SuperReader`, and `SuperTextField` all place various controls in the app overlay, like drag handles, toolbars, magnifier. Toolbars and magnifiers can be built by external builders, from the app developer. Therefore, it's reasonable to assume that those builders might want access to `InheritedWidget`s within or above the editor/reader/text field. By default, an `OverlayEntry` is given the context for the `Overlay`, which, by definition, doesn't include the part of the tree that added the widgets to the `Overlay`.

Consider using `OverlayPortal` for all overlay controls, so that those controls can access ancestors within the editor/reader/text field ancestor path.

matthew-carroll:
Mobile mounted toolbar now appears above Android overlay controls, migrated mounted toolbar and Android controls to OverlayPortal (Resolves #893)(Resolves #1510)

I migrated the mobile mounted keyboard and the Android overlay controls from traditional `OverlayEntry`s in the app `Overlay` to use `OverlayPortal` widgets. I did this because we've had issues in the past where we wanted the toolbars to pull configuration from ancestor widgets, but the toolbars sat in a different subtree and therefore couldn't access the same ancestors as `SuperEditor` and `SuperReader`. By moving to `OverlayPortal`, the widgets still appear in the overlay, but they gain access to the same ancestor path as `SuperEditor` and `SuperReader`, as if the overlay widgets were built inside of `SuperEditor` and `SuperReader`.

After migrating to `OverlayPortal`, I created a special type of `OverlayPortalController` called `GroupedOverlayPortalController`, which manages a priority list of overlays. The priority list of overlays enforces priority-based z-index display, which allows us to ensure that the mounted toolbar remains in front of the caret, handles, toolbar, etc.


matthew-carroll:
CC @Jethro87 - This PR should help you with z-index ordering for the mounted toolbar on the software keyboard and the drag handles and toolbars.

super_editor/example/lib/demos/example_editor/example_editor.dart
```diff
@@ -297,50 +237,51 @@ class _ExampleEditorState extends State<ExampleEditor> {
...
+      child: OverlayPortal(
+        controller: _imageFormatBarOverlayController,
+        overlayChildBuilder: _buildImageToolbar,
+        child: ListenableBuilder(
+          listenable: _brightness,
+          builder: (context, _) {
+            return Theme(
+              data: ThemeData(brightness: _brightness.value),
+              child: Builder(
+                builder: (themedContext) {
```
angelosilvestre, line 260: Shouldn't the `Theme` be above the `OverlayPortal` in the tree?

super_editor/lib/src/infrastructure/flutter/flutter_scheduler.dart
```diff
@@ -0,0 +1,99 @@
...
+  /// Runs the given [stateChange] (which might call `setState()`) as early as
+  /// possible.
+  ///
+  /// Given that [stateChange] might call `setState()`, it can't be run during
+  /// Flutter's build phase. If Flutter is currently in the middle of the build
+  /// phase, another frame is scheduled, and [stateChange] is run after the
+  /// current build phase completes. Otherwise, [stateChange] is run immediately.
+  void runStateChangeAsSoonAsPossible(VoidCallback stateChange) {
+    WidgetsBinding.instance.runAsSoonAsPossible(stateChange);
+  }
```
angelosilvestre, line 51: What do you think about adding a `runSetStateAsSoonAsPossible` to automatically wrap the given callback with a `setState`?

With the current method we are doing this: 

```dart
runStateChangeAsSoonAsPossible(() => setState(() {
    _toolbarHeight = toolbarHeight;
  }));
```

With the other option we could write it like this: 


```dart
runSetStateAsSoonAsPossible(() => {
  _toolbarHeight = toolbarHeight;
});
```

angelosilvestre (approved):
LGTM with some comments.

angelosilvestre:
@matthew-carroll Looks like we have failing tests.

super_editor/lib/src/infrastructure/flutter/flutter_scheduler.dart
matthew-carroll: Done

super_editor/example/lib/demos/example_editor/example_editor.dart
matthew-carroll: I don't think it matters, but I switched around some of the widget ordering, anyway.

PR #1536, opened 2023-10-20
[SuperEditor][SuperReader][Android] Fix scrolling with an ancestor Scrollable (Resolves #1535)

### LINKED ISSUE #1535: [SuperEditor] Scrolling on Android in a CustomScrollView creates selection when it shouldn't
I'm unable to scroll a document on Android in a CustomScrollView without the keyboard coming up with selection. I've tested this in my own app (running super_editor stable / Flutter stable) as well as the `sliver_example_editor` in the example app (running super_editor stable / Flutter stable). Here's a demo:

https://github.com/superlistapp/super_editor/assets/6467808/6edf3555-3874-4ada-b876-5222eac398f9

I didn't catch this when pushing our app to production yesterday.

cc @matthew-carroll

angelosilvestre:
[SuperEditor][SuperReader][Android] Fix scrolling with an ancestor Scrollable. Resolves #1535.

When the editor or reader is inside a `CustomScrollview`, trying to drag is opening the software keyboard and changing the selection.

The root cause is the long press to select handling. When the user taps down we start a timer to track a long press. Dragging after the tap down starts a pan gesture and we cancel the timer. However,  if we have an ancestor `Scrollable`, dragging starts a pan gesture on the `Scrollable`, not in the editor/reader. Because of that, the timer still fires after the timeout.

This PR adds a guard clause to avoid doing a long press if we are scrolling.

### DIFF
--- a/super_editor/test/super_editor/supereditor_scrolling_test.dart
+++ b/super_editor/test/super_editor/supereditor_scrolling_test.dart
@@ -581,6 +581,60 @@ void main() {
         expect(caretOffset.dy, greaterThanOrEqualTo(screenSizeWithKeyboard.height - trailingBoundary));
       });
 
+      testWidgetsOnMobile('scrolls when dragging without attaching to IME', (tester) async {
+        final scrollController = ScrollController();
+
+        // Pump an editor inside a CustomScrollView without enough room to display
+        // the whole content.
+        await tester
+            .createDocument() //
+            .withLongTextContent()
+            .withCustomWidgetTreeBuilder(
+              (superEditor) => MaterialApp(
+                home: Scaffold(
+                  body: ConstrainedBox(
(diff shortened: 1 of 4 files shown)


super_editor/test/super_editor/supereditor_scrolling_test.dart
```diff
@@ -581,6 +581,60 @@ void main() {
         expect(caretOffset.dy, greaterThanOrEqualTo(screenSizeWithKeyboard.height - trailingBoundary));
       });
 
+      testWidgetsOnMobile('scrolls when dragging without attaching to IME', (tester) async {
```
matthew-carroll, line 584: This test name should probably read "scrolling doesn't cause the keyboard to open"

matthew-carroll (approved):
LGTM

Issue #1407, opened 2023-09-07
How to change caret position after changing text alignment?


The28AWG:
![image](https://github.com/superlistapp/super_editor/assets/1275371/5213086f-7366-4f03-a493-4308f2e870f5)
the screenshot shows the alignment result TextAlign.center -> TextAlign.right
It may be possible to force the caret position to be updated?

The28AWG:
I am using super_editor version 0.2.6
It looks like the algorithm for working with the caret has been slightly changed in the repository

The28AWG:
![image](https://github.com/superlistapp/super_editor/assets/1275371/6572f101-0d84-45d7-be1f-51d07cdeb51c)
I couldn't run the main branch, there are too many errors that I can’t fix, so the following code concerns the stable branch.
https://github.com/superlistapp/super_editor/blob/d2f50c09d276f607e90e241a6d40dfde3fc0249b/super_editor/example/lib/demos/example_editor/example_editor.dart#L410
I had to add the following parameters to SuperEditor because without them the toolbar loses focus:
```dart
            selectionPolicies: SuperEditorSelectionPolicies(
              clearSelectionWhenEditorLosesFocus: false,
              clearSelectionWhenImeConnectionCloses: false,
            ),
```
https://github.com/superlistapp/super_editor/blob/d2f50c09d276f607e90e241a6d40dfde3fc0249b/super_editor/example/lib/demos/example_editor/_toolbar.dart#L436
add
```dart
    widget.closeToolbar();
```
This small modification is enough to ensure that the toolbar works on desktop and the focus does not disappear when hovering over the text alignment.
This allows you to test the text alignment behavior. We see that caret position is incorrect.

matthew-carroll:
I think you still haven't described the actual behavior that's causing the problem. Please provide reproduction steps to see the issue in your screenshot.

The28AWG:
[text align.webm](https://github.com/superlistapp/super_editor/assets/1275371/2c06c852-29d9-44f0-953f-f55e61ec7dc7)

The recording shows that the caret did not change its position after the text was aligned.

rutvik110:
Fixed by #1714

PR #1538, opened 2023-10-20
[Infrastructure] Implement custom dropdown widget (Resolves #1527, Resolves #731, Resolves #689)

### LINKED ISSUE #689: [super_editor] - Example app toolbar disappears when tapping anything with focus
EDIT (Matt): I changed the title of this ticket to match what I found. I also added a comment in the thread that defines the problem.

---
When the app is running, DropdownButton doesnt contain any List of _TextType or TextAlign in the _toolbar.dart,but  the widget tree lists the DropdownButton items 
image
image

### LINKED ISSUE #731: [super_editor] - Example app toolbar dropdowns make the toolbar disappear

### LINKED ISSUE #1527: [SuperEditor][SuperReader] - Create custom dropdown button and list with shared focus
For a long time, our demo floating toolbar for desktop has been broken because Flutter's built-in drop down buttons and lists refuse to implement appropriate focus sharing so that we can share focus from `SuperEditor` in one `OverlayEntry` to a `DropdownButton` in another `OverlayEntry`.

Create a custom dropdown button + dropdown content that correctly implements focus sharing, so that our demos work as expected, and our users can similarly create their own toolbars with dropdowns.

angelosilvestre:
[Infrastructure] Implement custom dropdown widget. Resolves #1527, Resolves #731, Resolves #689

Tapping at any `DropdownButton` in the example app toolbar causes the toolbar to immediately close. The issue is that Flutter `DropdownButton` displays the dropdown in its own route, which steals the focus from the editor.

This PR introduces a custom dropdown button. I called it `SuperDropdownButton`, but we can choose a better name. This widget displays the dropdown using an `OverlayPortal`. By doing so, it can share focus with `SuperEditor`.

We need to define how we want the public API:

- What we want to make everything.
- What we should do by default.

For example: What should be its default height? Do we want to make the width as large as the largest item in the list or just enough to fit the selected item? 

This is how it looks so far:

https://github.com/superlistapp/super_editor/assets/7597082/d2af4fd7-f60a-4eef-9ed6-a71cfee6ed15

I noticed that changing the paragraph alignment doesn't immediately rebuilds the paragraph, but it isn't related to the dropdown. I will file a ticket after this PR is merged.

Another example, constraining the dropdown height:

https://github.com/superlistapp/super_editor/assets/7597082/ff45cd8c-5666-47a5-99c9-60e21f2c505e


matthew-carroll:
@angelosilvestre can we position the popover below the button like a more traditional popup menu? Also can we round the corners to fit better with the toolbar shape?

Generally the functionality looks good. Let's make it look the best we can for the Super Editor Example app, and then we can look at which aspects should be configurable.

super_editor/example/lib/demos/example_editor/_toolbar.dart
```diff
@@ -224,6 +224,9 @@ class _EditorToolbarState extends State<EditorToolbar> {
         ),
       ]);
     }
+
+    // Rebuild to display the selected item.
+    setState(() {});
```
matthew-carroll, line 229: Where is the selected item stored? Why do we want `setState()` to be empty?

super_editor/example/lib/demos/example_editor/_toolbar.dart
```diff
@@ -521,26 +527,30 @@ class _EditorToolbarState extends State<EditorToolbar> {
...
+                      padding: EdgeInsets.only(left: 8.0),
+                      child: Text(
+                        _getTextTypeName(item!),
+                        style: const TextStyle(
+                          color: Colors.black,
+                          fontSize: 12,
+                        ),
+                      ),
+                    ),
+                    parentFocusNode: widget.editorFocusNode,
```
matthew-carroll, line 551: Let's not hide properties with simple values down below builders. They're too easy to miss.

super_editor/example/lib/demos/example_editor/_toolbar.dart
```diff
@@ -584,26 +594,14 @@ class _EditorToolbarState extends State<EditorToolbar> {
...
-                      color: Colors.black,
-                      fontSize: 12,
-                    ),
-                    underline: const SizedBox(),
-                    elevation: 0,
-                    itemHeight: 48,
+                    items: [TextAlign.left, TextAlign.center, TextAlign.right, TextAlign.justify],
                     onChanged: _changeAlignment,
+                    boundaryKey: widget.editorViewportKey,
+                    parentFocusNode: widget.editorFocusNode,
```
matthew-carroll, line 602: Please move `parentFocusNode` to top of the property list and then `boundaryKey` right below that.

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,479 @@
...
+    required this.itemBuilder,
+    required this.buttonBuilder,
+    this.focusColor,
+    required this.boundaryKey,
+    required this.parentFocusNode,
+    this.dropdownContraints,
+    this.dropdownKey,
+  });
+
+  /// The items that will be displayed in the dropdown list.
```
matthew-carroll, line 35: Let's re-order these properties to match our typical ordering.

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,479 @@
...
+@visibleForTesting
+class SuperDropdownButtonState<T> extends State<SuperDropdownButton<T>> with SingleTickerProviderStateMixin {
+  final DropdownController _dropdownController = DropdownController();
+
+  @visibleForTesting
+  ScrollController get scrollController => _scrollController;
+  final ScrollController _scrollController = ScrollController();
+
+  @visibleForTesting
+  int? get focusedIndex => _focusedIndex;
```
matthew-carroll, line 80: Let's put focus up top above the controllers

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,479 @@
...
+/// is changed by calling [DropdownController.show] or [DropdownController.hide].
+/// The dropdown is automatically closed when the user taps outside of its bounds.
+///
+/// When the dropdown is displayed it requests focus to itself, so the user can
+/// interact with the content using the keyboard. The focus is shared
+/// with the [parentFocusNode]. Provide [onKeyEvent] to handle key presses
+/// when the dropdown is visible.
+///
+/// This widget doesn't enforce any style, dropdown position or decoration.
+class RawDropdown extends StatefulWidget {
```
matthew-carroll, line 324: What does "Raw" mean here?

super_editor/test/infrastructure/dropdown_test.dart
```diff
@@ -0,0 +1,470 @@
...
+import 'package:super_editor/src/core/editor.dart';
+import 'package:super_editor/src/default_editor/default_document_editor.dart';
+import 'package:super_editor/src/default_editor/super_editor.dart';
+import 'package:super_editor/src/default_editor/text.dart';
+import 'package:super_editor/src/infrastructure/dropdown.dart';
+import 'package:super_editor/src/infrastructure/text_input.dart';
+import 'package:super_editor/super_editor_test.dart';
+
+import '../super_editor/test_documents.dart';
+
```
matthew-carroll, line 18: Let's call this file `super_dropdown_test` to make it clear that we're testing our custom widget instead of some arbitrary dropdown.

super_editor/example/lib/demos/example_editor/_toolbar.dart
angelosilvestre: The dropdown itself doesn't store the selected value, it's always provided via constructor.

In the toolbar, this value also isn't stored, we query the node type from the selected node when we create the dropdown.

super_editor/lib/src/infrastructure/dropdown.dart
angelosilvestre: Re-ordered to match how we are using it on `_toolbar.dart`

super_editor/lib/src/infrastructure/dropdown.dart
angelosilvestre: It only mean that we don't make any decisions here about how the widget should look like. We could probably find a better name for it... `UnstyledDropdown`, `OverlayDropdown`,  or something else.

angelosilvestre:
> @angelosilvestre can we position the popover below the button like a more traditional popup menu? Also can we round the corners to fit better with the toolbar shape?

@matthew-carroll  Yes, initially I left the dropdown centered to match the material dropdown.

Updated:

https://github.com/superlistapp/super_editor/assets/7597082/3b7e37d8-e8f5-4a8a-a350-84d8e01413b8

With slow animations:

https://github.com/superlistapp/super_editor/assets/7597082/db9f3a57-8393-4dca-bdf4-9a364568c785

super_editor/example/lib/demos/example_editor/_toolbar.dart
matthew-carroll: I don't understand those explanations. If the selected item isn't stored, then why are calling `setState()`?

super_editor/lib/src/infrastructure/dropdown.dart
matthew-carroll: Can you elaborate on that explanation? You've said what this widget "isn't" doing. Can you describe what it "is" doing?

matthew-carroll:
@angelosilvestre I see in your latest video that the popover prevents itself from moving beyond the bottom of the screen, but that causes it to cover the toolbar. My first thought is that this is unexpected and we don't want it. Was there any particular reason that you went with a screen-constrained popup? 

The following heuristic comes to mind:
 * When there's enough room beneath the toolbar, show the popup below the toolbar, otherwise
 * When there's enough room above the toolbar, show the popup above the toolbar, otherwise
 * If there's a reasonable amount of space (enough to see a few items), make the popover shorter and re-apply above rules, otherwise
 * Pin the popup in the standard position and size below the button and let it go off-screen

super_editor/example/lib/demos/example_editor/_toolbar.dart
angelosilvestre: I'm calling `setState` to rebuild the toolbar. Then, it will query the value again.

super_editor/lib/src/infrastructure/dropdown.dart
angelosilvestre: I tried to summarize what this widget does in the previous lines:

- It listens for changes of its `OverlayPortalController` and shows/hides the `Overlay`.
- Setup focus sharing.
- Requests focus to the dropdown's `FocusNode` when the overlay is displayed.
- Delivers key events to `onKeyEvent`. 
- Closes the dropdown when tapping outside of it.

angelosilvestre:
@matthew-carroll As discussed, I implemented a positioning behavior similar to Notion's.

1. If there is enough room to display the dropdown list beneath the toolbar, position it below the toolbar.
2. If there is enough room to display the dropdown list above the toolbar, position it above the toolbar.
3. Pin the dropdown list to the bottom and let it cover the toolbar.

I also simplified the transition to be just a fade-in animation.

This is how it's looking right now:

https://github.com/superlistapp/super_editor/assets/7597082/06a9adaa-84ea-4346-b200-d0a43030724d

For reference, this is how Notion's dropdown looks:

https://github.com/superlistapp/super_editor/assets/7597082/134118e7-0163-43c0-ac03-666b5be1008a

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,506 @@
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+import 'package:follow_the_leader/follow_the_leader.dart';
+import 'package:super_editor/super_editor.dart';
+
+/// A button which displays a dropdown with a vertical list of items.
```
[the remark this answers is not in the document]
matthew-carroll: Let's replace all uses of "dropdown" with "popover list" or "dropdown popover" so that it remains clear the dropdown is a separate piece of UI which sits above the selected item.

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,506 @@
...
+///
+/// Unlike Flutter `DropdownButton`, which displays the dropdown in a separate route,
+/// this widget displays its dropdown in an `Overlay`. By using an `Overlay`, focus can shared
```
matthew-carroll, line 9: "focus can shared" -> "focus can be shared"

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,506 @@
...
+/// with the [parentFocusNode]. This means that when the dropdown requests focus, [parentFocusNode]
+/// still has non-primary focus.
+///
+/// The dropdown tries to fit all items on the available space. If there isn't enough room,
```
matthew-carroll, line 13: Let's simply enumerate the rules for clarity:

```
/// The popover list is positioned based on the following rules:
///
///    1. The popover is displayed below the selected item, if there's enough room, or
///    2. The popover is displayed above the selected item, if there's enough room, or
///    3. The popover is displayed with its bottom aligned with the bottom of
///         the given boundary, and it covers the selected item.
```

matthew-carroll: I didn't mention scrolling constraints in there. Please update accordingly. The goal is to make the rules quick and easy to understand.

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,506 @@
...
+/// the list of items becomes scrollable with an always visible toolbar.
+/// Provide [dropdownContraints] to enforce aditional constraints on the dropdown list.
```
matthew-carroll, line 15: Let's not allude to what `dropdownConstraints` is for. It's not clear what it means to "enforce additional constraints". Please provide the specific motivation and behavior for a developer to provide `dropdownConstraints`.

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,506 @@
...
+///
+/// The user can navigate between the options using the arrow keys, select an option with ENTER and
```
matthew-carroll, line 17: Rather than passively mention interactions, please specify precise impact.

```
/// The popover list includes keyboard selection behaviors:
/// 
///   * Pressing UP/DOWN moves the "active" item selection up/down.
///   * Pressing UP with the first item active, or DOWN with the last item active, does nothing.
///   * Pressing ENTER selects the currently active item and closes the popover list.
```

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,506 @@
...
+/// close the dropdown with ESC.
+class SuperDropdownButton<T> extends StatefulWidget {
```
matthew-carroll, line 19: Let's find a better name for this. What about `ItemSelector`? Something else?

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,506 @@
...
+    required this.value,
+    required this.onChanged,
+    required this.itemBuilder,
+    required this.buttonBuilder,
+    this.focusColor,
+    this.dropdownContraints,
+    this.dropdownKey,
+  });
+
+  /// [FocusNode] which will share focus with the dropdown list.
```
matthew-carroll, line 34: Please replace with:

```
/// The [FocusNode], to which the popover list's [FocusNode] will be added as a child.
///
/// In Flutter, [FocusNode]s have parents and children. This relationship allows an
/// entire ancestor path to "have focus", but only the lowest level descendant
/// in that path has "primary focus". This path is important because various
/// widgets alter their presentation or behavior based on whether or not they
/// currently have focus, even if they only have "non-primary focus".
///
/// When the popover list of items is visible, that list will have primary focus.
/// Moreover, because the popover list is built in an `Overlay`, none of your
/// widgets are in the natural focus path for that popover list. Therefore, if you
/// need your widget tree to retain focus while the popover list is visible, then
/// you need to provide the [FocusNode] that the popover list should use as its
/// parent, thereby retaining focus for your widgets.
```

matthew-carroll: Should the parent focus node be required? What if a developer wants to use this widget in a place where he doesn't care about sharing focus?

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,506 @@
...
+  final FocusNode parentFocusNode;
+
+  /// A [GlobalKey] to a widget that determines the bounds where the dropdown can be displayed.
```
matthew-carroll, line 37: Should this be required? Can we default to a screen boundary if its `null`?

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,506 @@
...
+  ///
+  /// Used to avoid the dropdown to be displayed off-screen.
```
matthew-carroll, line 39: Please try to be more specific. The person reading probably has no idea how this is "used to avoid the dropdown to be displayed off-screen".

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,506 @@
...
+  final GlobalKey boundaryKey;
+
+  /// The currently selected value or `null` if no item is selected.
```
matthew-carroll, line 42: Please make clear the connection between this value and the widget's display.

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,506 @@
...
+  final T? value;
+
+  /// The items that will be displayed in the dropdown list.
```
matthew-carroll, line 45: Same here. There's ambiguity between the concept of the value and the items vs the display of the value and the display of the items.

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,506 @@
...
+  final List<T> items;
+
+  /// Called when the user selects an item on the dropdown list.
```
matthew-carroll, line 48: Please enumerate the ways in which a user might "select an item in the popover list"

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,506 @@
...
+  final ValueChanged<T?> onChanged;
+
+  /// The background color of the focused list item.
```
matthew-carroll, line 51: There are few issues with this property.

First, this level of control is too narrow. What if the user wants to use a gradient? Or a border? Or a sparkle effect? We shouldn't be making this UI decision for the user.

Second, it's not obvious what it means for an item to "have focus". In fact, I don't think the item will have focus, will it? Isn't it the list that has focus?

I think what you're trying to describe here is usually called the "active" item. You can double check that. But I know the terms "selected" and "active" often complement one another.

Wherever we want to refer to the item that's selected in the list, but not the actual selected item, we should introduce an appropriate term and use is consistently through this documentation.

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,506 @@
...
+  final Color? focusColor;
+
+  /// A [GlobalKey] bound to the dropdown list.
```
matthew-carroll, line 54: Please describe why someone cares. Why is this provided? What does it mean if it's `null`? Etc.

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,506 @@
...
+  final GlobalKey? dropdownKey;
+
+  /// Constraints applied to the dropdown list.
```
matthew-carroll, line 57: Please describe HOW this constrains the popover list.

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,506 @@
...
+  final BoxConstraints? dropdownContraints;
+
+  /// Builds each item in the dropdown list.
```
matthew-carroll, line 60: Please describe the connection between this builder and the items property.

Also, is there not already a typedef for a builder like this?

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,506 @@
...
+    widget.onChanged(item);
+    _dropdownController.hide();
+  }
+
+  KeyEventResult _onKeyEvent(FocusNode node, KeyEvent event) {
+    if (event is! KeyDownEvent) {
+      return KeyEventResult.ignored;
+    }
+
+    if (![
```
matthew-carroll, line 130: Can you declare this list `const` so we're not re-instantiating it on every key event? Also, if order doesn't matter, then this should be a `Set` not a `List`.

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,506 @@
...
+      LogicalKeyboardKey.numpadEnter,
+      LogicalKeyboardKey.arrowDown,
+      LogicalKeyboardKey.arrowUp,
+      LogicalKeyboardKey.escape,
+    ].contains(event.logicalKey)) {
+      return KeyEventResult.ignored;
+    }
+
+    if (event.logicalKey == LogicalKeyboardKey.escape) {
+      _dropdownController.hide();
```
matthew-carroll, line 141: `hide()` sounds temporarily. Let's call this `close()` unless you have a reason not to.

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,506 @@
...
+      return KeyEventResult.handled;
+    }
+
+    if (event.logicalKey == LogicalKeyboardKey.enter || event.logicalKey == LogicalKeyboardKey.numpadEnter) {
+      if (_focusedIndex == null) {
```
matthew-carroll, line 146: Please describe what this condition means with a comment.

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,506 @@
...
+        _dropdownController.hide();
+        return KeyEventResult.handled;
+      }
+
+      _submitItem(widget.items[_focusedIndex!]);
```
matthew-carroll, line 151: Let's call this `_selectItem()` to remain consistent with all other places where we talk about this widget displaying a selected item.

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,506 @@
...
+
+      return KeyEventResult.handled;
+    }
+
+    int? newFocusedIndex;
+    if (event.logicalKey == LogicalKeyboardKey.arrowDown) {
+      if (_focusedIndex == null || _focusedIndex! >= widget.items.length - 1) {
+        // We don't have a focused item or we are at the end of the list. Focus the first item.
```
matthew-carroll, line 159: What does it mean to not have a focused item? Is that not an error state?

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,506 @@
...
+  @override
+  Widget build(BuildContext context) {
+    return RawDropdown(
+      controller: _dropdownController,
+      boundaryKey: widget.boundaryKey,
+      dropdownBuilder: _buildDropDown,
+      parentFocusNode: widget.parentFocusNode,
+      onKeyEvent: _onKeyEvent,
+      child: ConstrainedBox(
+        // TODO: what value should we use?
```
matthew-carroll, line 193: When I first read this, I thought this was the popover list constraint. But this is constraining the selected item display, right? If so, why are we constraining that at all?

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,506 @@
...
+          child: Center(
+            child: _buildSelectedItem(),
+          ),
+        ),
+      ),
+    );
+  }
+
+  Widget _buildSelectedItem() {
+    return Row(
```
matthew-carroll, line 206: Should this be a `Row` or a `Stack`? Imagine the developer wants to display a selected item with their own background color or gradient, how would they get a full-bleed display if we're pushing their widget to the side for a dropdown arrow?

super_editor/lib/src/infrastructure/dropdown.dart
matthew-carroll: I'm confused why the follower and aligner are configured elsewhere instead of here? If this widget does what you describe, why wouldn't it also include positioning?

Is this widget expected to be used publicy? If not, can we make it private?

If we make it private and move the follower and aligner in here, let's call this `_FollowerPopover`. The content of the popover can then be provided via `child` to decide what to display within the popover.

super_editor/lib/src/infrastructure/dropdown.dart
angelosilvestre: I'm not sure I understood what "scrolling constraints" means here.

super_editor/lib/src/infrastructure/dropdown.dart
angelosilvestre: Just a note, pressing UP with the first item active moves the active selection to the last item and  pressing DOWN with the last item active moves the active selection to the first item.

super_editor/lib/src/infrastructure/dropdown.dart
angelosilvestre: The dropdown internally uses a `SuperEditorPopover`. Should we pass a `FocusNode` created inside the dropdown if the app doesn't provide one?

super_editor/lib/src/infrastructure/dropdown.dart
angelosilvestre: Updated.

super_editor/lib/src/infrastructure/dropdown.dart
angelosilvestre: Updated.

super_editor/lib/src/infrastructure/dropdown.dart
angelosilvestre: We can't declared it as a `const Set` because of the following compilation error:

```console
An element in a constant set can't override the '==' operator, or 'hashCode', but the type 'LogicalKeyboardKey' does.
Try using a different value for the element, or removing the keyword 'const' from the set.
```

Should we use a non-const `Set` or a const `List`?

super_editor/lib/src/infrastructure/dropdown.dart
angelosilvestre: In the current implementation, the popover list opens without an "active item". The active item is set only after the user presses ARROW DOWN or ARROW UP.

This is compatible with other popover list implementations.

super_editor/lib/src/infrastructure/dropdown.dart
angelosilvestre: Updated.

super_editor/lib/src/infrastructure/dropdown.dart
angelosilvestre: I changed to use a `BoxDecoration`, if you want we should still offer more control, then we could use an widget builder.

About the naming, I updated to use the term "active" to represent the item selected in the list.

super_editor/lib/src/infrastructure/dropdown.dart
angelosilvestre: Currently, I'm only using this key to find the popover list during tests. Should I remove this and find the popover list with another approach?

super_editor/lib/src/infrastructure/dropdown.dart
angelosilvestre: I'm not sure how I should describe that... The only thing I'm doing with this `BoxConstraints` is passing it to a `ConstrainedBox` which wraps the popover list.

super_editor/lib/src/infrastructure/dropdown.dart
angelosilvestre: I didn't find any typedef for this.

super_editor/lib/src/infrastructure/dropdown.dart
angelosilvestre: I've made it that way so we (or others) could re-use the `RawDropDown` to implement other kinds of dropdown buttons.

For example, it would be relatively easy to implement something like a simple (and incomplete) color picker with the following code:

```dart
class ColorPicker extends StatefulWidget {
  const ColorPicker({super.key});

  @override
  State<ColorPicker> createState() => _ColorPickerState();
}

class _ColorPickerState extends State<ColorPicker> {
  final DropdownController _dropdownController = DropdownController();

  // This might be removed.
  final FocusNode _focusNode = FocusNode();

  final List<MaterialColor> colors = [
    Colors.red,
    Colors.blue,
    Colors.amber,
    Colors.brown,
    Colors.green,
    Colors.orange,
    Colors.pink,
    Colors.teal
  ];

  @override
  void dispose() {
    _dropdownController.dispose();
    _focusNode.dispose();
    super.dispose();
  }

  @override
    return RawDropdown(
      controller: _dropdownController,
      dropdownBuilder: _buildColorPopover,
      parentFocusNode: _focusNode,
      child: ElevatedButton(
        onPressed: () => _dropdownController.open(),
        child: Text("Pick a color"),
      ),
    );
  }

  Widget _buildColorPopover(BuildContext context, LeaderLink link, FollowerBoundary boundary) {
    return Follower.withOffset(
      link: link,
      leaderAnchor: Alignment.bottomCenter,
      followerAnchor: Alignment.topCenter,
      offset: Offset(0, 10),
      child: Material(
        elevation: 4.0,
        borderRadius: BorderRadius.circular(8.0),
        child: Container(
          height: 50,
          width: 200,
          padding: const EdgeInsets.all(8.0),
          child: ListView.separated(
            scrollDirection: Axis.horizontal,
            // TODO: implement selection
            itemBuilder: (context, index) => Container(
              decoration: BoxDecoration(
                shape: BoxShape.circle,
                color: colors[index],
              ),
              height: 20,
              width: 20,
            ),
            separatorBuilder: (context, index) => SizedBox(width: 10),
            itemCount: colors.length,
          ),
        ),
      ),
    );
  }
}
```

https://github.com/superlistapp/super_editor/assets/7597082/3cecd23f-8bbb-46c0-ae26-8f8503430363

super_editor/lib/src/infrastructure/dropdown.dart
angelosilvestre: I don't remember why I added that, but it seems it's unnecessary. Removed.

super_editor/example/lib/demos/example_editor/_toolbar.dart
angelosilvestre: After taking a better look at this, I think that the main problem here is that we aren't rebuilding the toolbar when the document changes.

After we change the selected node's metadata, we need to rebuild the toolbar so it can "see" the current state of the selected node.

I moved this `setState` call to the document change listener. Please let me know if this is an acceptable solution.

super_editor/lib/src/infrastructure/dropdown.dart
matthew-carroll: I'm not entirely sure what I meant, but I think I may have been referring to the content in the popover going from fully-visible to partially visible + scrolling. For example, let's say we pull up a list of options in a popover. There's likely some kind of decision whether to make the popover as tall as all those items in the list, or whether to make the popover shorter, and introduce scrolling to the list. That choice about adding scrolling, or not, and the resulting height, impacts all of the rules above. They're like two sides of the same coin.

super_editor/lib/src/infrastructure/dropdown.dart
matthew-carroll: Is that the most common behavior when at the top or bottom of a list with arrow keys?

super_editor/lib/src/infrastructure/dropdown.dart
matthew-carroll: Is `SuperEditorPopover` something we created in this PR or did it already exist? If we created it in this PR, then can we just not make it required there either? If that already existed and we don't want to change it, then yes, I think create a new `FocusNode()` locally and pass it through.

super_editor/lib/src/infrastructure/dropdown.dart
matthew-carroll: Can you let me know why we ask the user to specify anything at all? Whether a color or a decoration? Do we really need this property? Can the user not achieve the same result through widgets that the user is already building?

super_editor/lib/src/infrastructure/dropdown.dart
matthew-carroll: Let's typedef it ourselves then

super_editor/lib/src/infrastructure/dropdown.dart
matthew-carroll: Let's use a `const List()`

super_editor/lib/src/infrastructure/dropdown.dart
matthew-carroll: Ok, we can try that

super_editor/lib/src/infrastructure/dropdown.dart
matthew-carroll: Really? If we're looking at a typical address form on the web, and there's a country selector, and we have United States selected, when we open that selector and show the popover, "United States" isn't active? And the popover isn't centered on the "United States" item?

On a related note, if no item is active initially, which one becomes active when you press UP or DOWN?

super_editor/example/lib/demos/example_editor/_toolbar.dart
```diff
@@ -95,14 +97,32 @@ class _EditorToolbarState extends State<EditorToolbar> {
...
+
+    widget.document.removeListener(_onDocumentChanged);
     super.dispose();
   }
 
+  void _onDocumentChanged(DocumentChangeLog doc) {
+    // The document has changed.
+    // Some of the selected node's metadata, for example, the blockType, influences how the toolbar is displayed.
+    // Reflow the toolbar to acount for the new state of the selected node.
+    setState(() {});
```
matthew-carroll, line 123: Is there any particular reason that we don't listen for this change around, or within the pieces of the toolbar that care, rather than rebuilding the whole toolbar?

super_editor/example/lib/demos/example_editor/_toolbar.dart
```diff
@@ -521,26 +541,30 @@ class _EditorToolbarState extends State<EditorToolbar> {
               if (_isConvertibleNode()) ...[
                 Tooltip(
                   message: AppLocalizations.of(context)!.labelTextBlockType,
-                  child: DropdownButton<_TextType>(
+                  child: ItemSelector<_TextType>(
```
matthew-carroll, line 544: I'm wondering if there's a more useful way to breakdown responsibilities for these widgets. After you go through the other new comments, let's jump on a call and talk about how to break things up.

My gut feeling is that we want two different levels of widgets here. We want foundational pieces that make almost zero visual designs but provide some base level relationships. Then we want some widgets that give us default styling, and also take a few properties to configure obvious UI details. At the moment, we're trying to do both of those things in one group of widgets.

super_editor/lib/src/infrastructure/dropdown.dart
angelosilvestre: From what I've seen, it's pretty common.

super_editor/lib/src/infrastructure/dropdown.dart
angelosilvestre: It was introduced in https://github.com/superlistapp/super_editor/pull/1443. It was created specifically to share focus, so I think we don't want to change it.

I'm creating a local `FocusNode`.

super_editor/lib/src/infrastructure/dropdown.dart
angelosilvestre: To remove that property, we need to add another parameter in `itemBuilder` to indicate whether the item is active. Should we go in this direction?

super_editor/lib/src/infrastructure/dropdown.dart
angelosilvestre: > Really? If we're looking at a typical address form on the web, and there's a country selector, and we have United States selected, when we open that selector and show the popover, "United States" isn't active? And the popover isn't centered on the "United States" item?

It makes sense. Updated to make the selected value "active" in the list.

> On a related note, if no item is active initially, which one becomes active when you press UP or DOWN?

Pressing UP makes the last item active and pressing DOWN makes the first item active. We could also change the behavior to make the first item active if there isn't a selected item.

super_editor/lib/src/infrastructure/dropdown.dart
angelosilvestre: Updated.

super_editor/example/lib/demos/example_editor/_toolbar.dart
angelosilvestre: Updated.

super_editor/example/lib/demos/example_editor/_toolbar.dart
angelosilvestre: It makes sense, we can jump on a call to talk about that.

rutvik110:
> I noticed that changing the paragraph alignment doesn't immediately rebuilds the paragraph, but it isn't related to the dropdown. I will file a ticket after this PR is merged.

@angelosilvestre I'm taking look at it as part of the #1407

angelosilvestre:
@matthew-carroll As discussed, the API is changing. Some things are a bit different from what we've talked. Here's the summary:

We have now the following widgets:

### `PopoverScaffold` (still needs a better name), which takes: 

- A `PopoverController` that opens and closes the popover.
- A `buttonBuilder` to build the button.
- A `popoverBuilder` to build the popover.
- A `popoverGeometry` to control the size and position of the popover.
- An `onKeyEvent` to which key events will be forwarded. This widget uses a `SuperEditorPopover` to setup the focus sharing, and `SuperEditorPopover` is the widget which install the `Focus` widget on the tree. Because of that, instead of adding a `Focus` ourselves, we just forward the `onKeyEvent`.
- An optional `parentFocusNode` to share focus
- An optional `boundaryKey` to keep the popover within the boundary.

This widget doesn't take a list of items because it doesn't care about whats being displayed on the popover. We could use the popover to display a text, an image preview, etc.

### `ItemSelectionList`, which takes:
- The selected value.
- The list of items.
- A `buttonBuilder` to build the button.
- An `itemBuilder` to build the each item on the list.
- An  `onItemActivated` to notify when an item is activated.
- An `onItemSelected` to notify when an item is selected.
- A `popoverGeometry` to control the size and position of the popover.
- An optional `parentFocusNode` to share focus.
- An optional `boundaryKey` to keep the popover within the boundary.

This widget handles the keyboard navigation, auto-scrolling and builds the list of options. I ignored popovers with horizontal scrolling for now. I'm not sure if we need a widget in between  `PopoverScaffold` and `ItemSelectionList`, which instead of an `itemBuilder` would take a `popoverBuilder` to build the entire popover based on the list of `items`.

### `SuperEditorDemoTextItemSelector` and  `SuperEditorDemoIconItemSelector`:

Widgets with a predefined style to be used in the demos.

angelosilvestre:
Rebased to fix an import.

matthew-carroll:
@angelosilvestre can you please edit your comment up above to point out what you changed from our discussion and why? 

Based on what I'm seeing in that comment right now, here are a couple comments:

## PopoverScaffold
> An onKeyEvent to which key events will be forwarded. This widget uses a SuperEditorPopover to setup the focus sharing, and SuperEditorPopover is the widget which install the Focus widget on the tree. Because of that, instead of adding a Focus ourselves, we just forward the onKeyEvent.

What do you mean "This widget uses a SuperEditorPopover"? The scaffold is the lowest level widget here and `SuperEditorPopover` is the highest level. If the scaffold is truly using a `SuperEditorPopover`, then that needs to be reversed.

"instead of adding a Focus ourselves" - I'm not sure what that means either, but you said the scaffold takes a `parentFocusNode`, which indicates that the scaffold knows about and cares about focus. Unless I'm missing something, we don't want to split focus responsibilities across multiple widgets.

## ItemSelectionList
Why is this list taking a geometry? I believe the purpose of this widget, as discussed on our call, was to provide a list widget. That list widget might be placed within a popover, but its not, itself, a popover. It's a list that understands the concept of "selection" and "activation" and arrow key navigation, and ENTER to select the activated item, and auto-scrolling when pressing arrows keys near the top and bottom of the visible space. All of that is independent from use within a popover. Such a widget would be equally applicable for a selection list that's placed within a typical screen UI.

Please let me know if I'm missing something.

angelosilvestre:
> What do you mean "This widget uses a SuperEditorPopover"? The scaffold is the lowest level widget here and SuperEditorPopover is the highest level. If the scaffold is truly using a SuperEditorPopover, then that needs to be reversed.

Maybe we have some naming issue... `SuperEditorPopover` is a widget introduce in another PR. It sets up the focus sharing and also add the blocked intents needed for macOS. It's this widget:

https://github.com/superlistapp/super_editor/blob/6e2040f573e58fd6e88224fce66f3719fe30b0b7/super_editor/lib/src/infrastructure/popovers.dart#L5-L64

> Why is this list taking a geometry? I believe the purpose of this widget, as discussed on our call, was to provide a list widget. That list widget might be placed within a popover, but its not, itself, a popover. It's a list that understands the concept of "selection" and "activation" and arrow key navigation, and ENTER to select the activated item, and auto-scrolling when pressing arrows keys near the top and bottom of the visible space. All of that is independent from use within a popover. Such a widget would be equally applicable for a selection list that's placed within a typical screen UI.

I think I misunderstood it a little bit. I'll update it.

angelosilvestre:
@matthew-carroll  Updated the PR. I think now it's closer to what we've talked.

I temporarily commented the test until we are sure the API is right.

Can you do a first pass review?

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,1059 @@
...
+///    1. The popover is displayed as tall as all items in the list, if there's enough room, or
+///    2. The popover is displayed as tall as the available space and becomes scrollable.
+///
+/// The popover list includes keyboard selection behaviors:
+///
+///   * Pressing UP/DOWN moves the "active" item selection up/down.
+///   * Pressing UP with the first item active moves the active item selection to the last item.
+///   * Pressing DOWN with the last item active moves the active item selection to the first item.
+///   * Pressing ENTER selects the currently active item and closes the popover list.
+class SuperEditorDemoTextItemSelector extends StatefulWidget {
```
matthew-carroll, line 35: Given that this is a demo-oriented class, it should be defined within the example app project. Only the core popover building blocks should be defined within the package and exported.

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,1059 @@
...
+  const SuperEditorDemoTextItemSelector({
+    super.key,
+    this.value,
+    required this.items,
+    required this.onSelected,
+    this.parentFocusNode,
```
matthew-carroll, line 41: Please put focus nodes near the top of property lists. We typically place generic Flutter details first, e.g., `Key`, `BuildContext`, `FocusNode`, controllers.

Also, please move `boundaryKey` up as well, so that it's above `items`.

And, in general, ensure that callbacks tend to be the last thing in property lists, other than `child` or `children` parameters.

Make similar changes to the actual property declarations.

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,1059 @@
...
+      boundaryKey: widget.boundaryKey,
+      popoverBuilder: (context) => PopoverShape(
+        child: ItemSelectionList<SuperEditorDemoTextItem>(
+          value: widget.value,
+          items: widget.items,
+          itemBuilder: _buildPopoverListItem,
+          onItemSelected: _onItemSelected,
+          onCancel: () => _popoverController.close(),
+          focusNode: _popoverFocusNode,
+          parentFocusNode: widget.parentFocusNode,
```
matthew-carroll, line 118: In the comments, we discussed that `ItemSelectionList` should only include list-related behaviors. Does it really need the `parentFocusNode`? Could `PopoverScaffold` setup the focus relationship and have the `ItemSelectionList` be unaware of that parenting relationship?

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,1059 @@
...
+  Widget _buildPopoverListItem(BuildContext context, SuperEditorDemoTextItem item, bool isActive, VoidCallback onTap) {
+    return DecoratedBox(
+      decoration: BoxDecoration(
+        color: isActive ? Colors.grey.withOpacity(0.2) : Colors.transparent,
+      ),
+      child: InkWell(
+        onTap: onTap,
+        child: Container(
+          constraints: const BoxConstraints(minHeight: kMinInteractiveDimension),
+          alignment: AlignmentDirectional.centerStart,
```
matthew-carroll, line 133: What's this `AlignmentDirectional` about? I don't think I've seen that before.

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,1059 @@
...
+              fontSize: 12,
+            ),
+          ),
+        ),
+      ),
+    );
+  }
+
+  Widget _buildButton(BuildContext context) {
+    return PopoverArrowButton(
```
matthew-carroll, line 148: What's the relevance of saying "arrow" in this name? Did we have that term in there on our call? I thought I remembered two relevant variations: a text button and an icon button. Also, given that this is a styled button, it should probably be named as a Super Editor demo and moved into the Example app, right?

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,1059 @@
...
+              ),
+            ),
+          ),
+        ),
+      ),
+    );
+  }
+}
+
+/// A rounded rectangle shape with a fade-in transition.
```
matthew-carroll, line 692: I don't think we should combine the shape with the transition animation. Those don't seem related. Also, this popover shape doesn't allow any configuration for the shape...

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,1059 @@
...
+      clipBehavior: Clip.hardEdge,
+      child: FadeTransition(
+        opacity: _containerFadeInAnimation,
+        child: widget.child,
+      ),
+    );
+  }
+}
+
+/// A widget which displays a button built in [buttonBuilder] with a popover
```
matthew-carroll, line 745: Imagine reading this first sentence in docs - would you have any idea what this widget is meant to help with? Rather than describe this widget with a very generic concept of buttons and following, please describe this widget in terms of the UX that the user is looking for, i.e., a popover selection system.

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,1059 @@
...
+  void toggle() {
+    if (shouldShow) {
+      close();
+    } else {
+      open();
+    }
+  }
+}
+
+/// Controls the size and position of a popover.
```
matthew-carroll, line 955: This data structure doesn't control anything. This data structure *is* the offset and size of a popover in screen space. Please try to avoid the term "position" when you're not talking about text or document positions.

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,1059 @@
...
+  final PopoverAligner align;
+
+  /// [BoxConstraints] applied to the popover.
+  ///
+  /// If `null`, the popover can use all the available space.
+  final BoxConstraints? constraints;
+}
+
+/// A [FollowerAligner] which uses a [delegate] to align a [Follower].
+class _DelegateAligner implements FollowerAligner {
```
matthew-carroll, line 975: Does this mean we need a functional implementation of `FollowerAligner`? Or perhaps a default definition of `FollowerAligner` that takes a delegate? If so, let's make that change in follow_the_leader.

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,1059 @@
...
+      leaderAnchor: Alignment.bottomCenter,
+      followerAnchor: Alignment.topCenter,
+      followerOffset: Offset(0, 20),
+    );
+  }
+
+  return alignment;
+}
+
+/// A function to align a Widget following a leader Widget.
```
matthew-carroll, line 1044: Here again, similar to the description of the popover scaffold, you seem to be trying to avoid saying what this thing is actually meant to do. Why would we want to describe this as "A function to align a widget following a leader widget"? Also, doesn't every follower aligner fit that definition? Let's tell the reader what they're goin to want to know about when/where/why/how to use this thing.

matthew-carroll (commented):
I did an initial review. One big initial change should be to move all the demo stuff into the demo app, instead of in the package. That might clarify some of the relationships and also point out where some things are named poorly.

super_editor/example/lib/demos/example_editor/_toolbar.dart
```diff
@@ -521,26 +522,33 @@ class _EditorToolbarState extends State<EditorToolbar> {
...
                   message: AppLocalizations.of(context)!.labelTextBlockType,
-                  child: DropdownButton<_TextType>(
-                    value: _getCurrentTextType(),
-                    items: _TextType.values
-                        .map((textType) => DropdownMenuItem<_TextType>(
-                              value: textType,
-                              child: Padding(
-                                padding: const EdgeInsets.only(left: 16.0),
-                                child: Text(_getTextTypeName(textType)),
+                  child: _DocumentListenableBuilder(
```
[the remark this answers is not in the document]
angelosilvestre: With the recent changes we don't need it anymore.

super_editor/lib/src/infrastructure/dropdown.dart
angelosilvestre: Is just an `Alignment` that is directionality aware, so `centerStart` is `centerLeft` when left-to-right and `centerRight` when right-to-left, but I guess we could just use `Alignment`.

super_editor/lib/src/infrastructure/dropdown.dart
angelosilvestre: This widget is used by both the text button and the icon button. I moved this widget to the demo app.

super_editor/lib/src/infrastructure/dropdown.dart
angelosilvestre: I'm currently using a `Material` widget to build the popover shape, should I take a `ShapeBorder` to customize the popover shape?

super_editor/lib/src/infrastructure/dropdown.dart
angelosilvestre: If we don't pass the `parentFocusNode` to the list, we need a way to setup the key event handler in `PopoverScaffold` and find a way to forward this key events to `ItemSelectionList`.

super_editor/lib/src/infrastructure/dropdown.dart
angelosilvestre: Updated.

super_editor/lib/src/infrastructure/dropdown.dart
angelosilvestre: OK, should I open a PR adding a `FunctionalAligner` to follow_the_leader?

angelosilvestre: Opened https://github.com/Flutter-Bounty-Hunters/follow_the_leader/pull/42

super_editor/lib/src/infrastructure/dropdown.dart
matthew-carroll: I'm confused - how does the use of `Material` relate to the two issues I raised above?

super_editor/lib/src/infrastructure/dropdown.dart
matthew-carroll: Why exactly do we need the list to setup the parent focus relationship? Why can't the scaffold setup the parent relationship and let the list operate like any other list with focus?

angelosilvestre: We can setup the parent focus relationship in the scaffold with `FocusWithCustomParent`. However, in `ItemSelectionList` with have a `Focus` widget to setup the key handler. If we just declare it like the following:

```dart
// ItemSelectionListState

@override
  return Focus(
     focusNode: widget.focusNode,
     onKeyEvent: _onKeyEvent,
     ...
  );
}
```

then the focus sharing doesn't work anymore. When `PopoverScaffold` requests focus, the `widget.focusNode` loses focus.

Am I missing something here?

super_editor/lib/src/infrastructure/dropdown.dart
angelosilvestre: Now that https://github.com/Flutter-Bounty-Hunters/follow_the_leader/pull/42 was merged, this won't be used anymore.

super_editor/lib/src/infrastructure/dropdown.dart
angelosilvestre: Updated to use the `FunctionalAligner`. I temporarily added a dependency override to use `follow_the_leader` from git. We'll need to release a new version...

matthew-carroll: I just released a new version

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,635 @@
...
+/// [PopoverController.open] or [PopoverController.close]. The popover is automatically closed
+/// when the user taps outside of its bounds.
+///
+/// Provide a [popoverGeometry] to control the size and position of the popover. The popover
+/// is first sized given the [PopoverGeometry.constraints] and then positioned using the
+/// [PopoverGeometry.align].
+///
+/// When the popover is displayed it requests focus to itself, so the user can
+/// interact with the content using the keyboard.
+class PopoverScaffold extends StatefulWidget {
```
matthew-carroll, line 355: Let's move this scaffold to the top of the file because everything else is pretty much in service of this.

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,635 @@
...
+      clipBehavior: Clip.hardEdge,
+      child: FadeTransition(
+        opacity: _containerFadeInAnimation,
+        child: widget.child,
+      ),
+    );
+  }
+}
+
+/// A widget used to build UI controls that display a following popover.
```
matthew-carroll, line 343: Let's replace the top line with this: "A scaffold, which builds a popover selection system, comprised of a button and a popover that's positioned near the button."

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,635 @@
...
+  /// Controls the size and position of the popover.
+  ///
+  /// The popover is first sized, then positioned.
+  final PopoverGeometry popoverGeometry;
+
+  /// Called at each key press while the popover has focus.
+  final FocusOnKeyEventCallback? onKeyEvent;
+
+  /// [FocusNode] which will share focus with the popover.
+  final FocusNode? popoverFocusNode;
```
matthew-carroll, line 385: Let's move this up beneath `popoverGeometry` to keep `popover...` things together.

Also, can you remind me whey we need this? The dev builds the popover, right?

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,635 @@
...
+  /// Builds the content of the popover.
+  final WidgetBuilder popoverBuilder;
+
+  /// Controls the size and position of the popover.
+  ///
+  /// The popover is first sized, then positioned.
+  final PopoverGeometry popoverGeometry;
+
+  /// Called at each key press while the popover has focus.
+  final FocusOnKeyEventCallback? onKeyEvent;
```
matthew-carroll, line 382: Can you remind me why we need to let developers provide an `onKeyEvent`? If devs want to process keys within the button, they can build that into the button, and if devs want to process keys within the popover list, they can build that into the list, right?

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,635 @@
...
+  }
+
+  void _updateFollowerBoundary() {
+    if (widget.boundaryKey != null) {
+      _screenBoundary = WidgetFollowerBoundary(
+        boundaryKey: widget.boundaryKey,
+        devicePixelRatio: MediaQuery.devicePixelRatioOf(context),
+      );
+    } else {
+      _screenBoundary = ScreenFollowerBoundary(
```
matthew-carroll, line 463: Given that we're installing a screen boundary when the boundary key is null, please revisit all the Dart Docs that discuss the boundary key, because I think they say the popover might go off-screen.

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,635 @@
...
+class ItemSelectionListState<T> extends State<ItemSelectionList<T>> with SingleTickerProviderStateMixin {
+  int? _activeIndex;
+
+  @visibleForTesting
+  ScrollController get scrollController => _scrollController;
+  final ScrollController _scrollController = ScrollController();
+
+  final GlobalKey _scrollableKey = GlobalKey();
+
+  /// Holds keys to each item on the list.
```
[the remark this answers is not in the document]
angelosilvestre: Updated

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,635 @@
...
+      // We just opened the popover.
+      // Jump to the active item without animation.
+      _activateItem(selectedItemIndex, Duration.zero);
+    } else {
+      // A selected item was provided, but it isn't included in the list of items.
+      _activeIndex = null;
+    }
+  }
+
+  /// Called when the user taps an item or presses ENTER with an active item.
```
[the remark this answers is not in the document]
angelosilvestre: I ended up removing this method and calling `onItemActivated` directly.

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,635 @@
...
+  void _selectItem(T? item) {
+    widget.onItemSelected(item);
+  }
+
+  /// Activates the item at [itemIndex] and ensure it's visible on screen.
```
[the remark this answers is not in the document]
angelosilvestre: Updated.

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,635 @@
...
+  void _activateItem(int? itemIndex, Duration animationDuration) {
+    _activeIndex = itemIndex;
+    if (itemIndex != null) {
+      widget.onItemActivated?.call(widget.items[itemIndex]);
+    }
+
+    // Scrolls on the next frame to let the popover to be
```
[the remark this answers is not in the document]
angelosilvestre: Added a comment explaining.

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,635 @@
...
+        // The user pressed ENTER without an active item.
+        _selectItem(null);
+        return KeyEventResult.handled;
+      }
+
+      _selectItem(widget.items[_activeIndex!]);
+
+      return KeyEventResult.handled;
+    }
+
```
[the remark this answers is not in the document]
angelosilvestre: Updated.

super_editor/lib/src/infrastructure/dropdown.dart
angelosilvestre: It seems it isn't being used anymore. Removed.

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,635 @@
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+import 'package:follow_the_leader/follow_the_leader.dart';
+import 'package:super_editor/src/infrastructure/flutter/flutter_scheduler.dart';
+import 'package:super_editor/src/infrastructure/platforms/mac/mac_ime.dart';
+import 'package:super_editor/super_editor.dart';
+
+/// A selection control, which displays a selected item, and upon tap, displays a
```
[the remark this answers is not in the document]
angelosilvestre: Updated the doc. I could use some help to add an information in the doc about the distinction between "active" and "selected"...

super_editor/lib/src/infrastructure/dropdown.dart
```diff
@@ -0,0 +1,608 @@
...
+          ),
+        ),
+      ),
+    );
+  }
+}
+
+/// A list where the user can navigate between its items and select one of them.
+///
+/// This widget shares focus with its [parentFocusNode]. This means that when the list requests focus,
```
matthew-carroll, line 172: As discussed, let's get to the bottom of why this list needs the parent focus node.  It seems like the following should work:

Editor focus -> popover focus (using parent relationship) -> list focus (using regular focus relationship)

Let's find out why that isn't working.

angelosilvestre: I pushed a commit with this change. Now it's working as expected.

angelosilvestre:
@matthew-carroll Added tests

super_editor/lib/src/infrastructure/selectable_list.dart
```diff
@@ -0,0 +1,268 @@
...
+              ),
+            ),
+          ),
+        ),
+      ),
+    );
+  }
+}
+
+/// Builds a popover list item.
```
matthew-carroll, line 263: Should "popover" be mentioned here?

super_editor/lib/src/infrastructure/selectable_list.dart
```diff
@@ -0,0 +1,268 @@
...
+///
+/// [isActive] is `true` if [item] is the currently active item on the list, or `false` otherwise.
```
matthew-carroll, line 265: It's probably a good idea to say something about what it means to be active, e.g., the item in the list that currently has focus and can be moved up/down with arrow keys.

super_editor/test/infrastructure/popover_test.dart
```diff
@@ -0,0 +1,440 @@
...
+
+      // Taps outside of the popover.
+      await tester.tapAt(Offset.zero);
+      await tester.pumpAndSettle();
+
+      // Ensure the popover was closed.
+      expect(find.byType(RoundedRectanglePopoverAppearance), findsNothing);
+    });
+
+    testWidgetsOnAllPlatforms('enforces the given popover geometry', (tester) async {
```
matthew-carroll, line 98: If this is only testing height constraints, then please name the test accordingly. This name suggests that all aspects of geometry are being verified against all possible conditions.

Related to that, is testing just the height of the popover, in just one set of conditions, sufficient to lock down the geometry implementation?

super_editor/test/infrastructure/popover_test.dart
```diff
@@ -0,0 +1,440 @@
...
+
+      // Show the popover.
+      popoverController.open();
+      await tester.pumpAndSettle();
+
+      // Ensure the popover is displayed.
+      expect(find.byType(RoundedRectanglePopoverAppearance), findsOneWidget);
+
+      // Ensure the maxHeight was honored.
+      expect(tester.getRect(find.byType(RoundedRectanglePopoverAppearance)).height, 10);
```
matthew-carroll, line 131: Please ensure that we're testing with reasonable sizes and shapes so that we don't let little things sneak by. I doubt that any real popover will be 10px tall.

super_editor/test/infrastructure/popover_test.dart
```diff
@@ -0,0 +1,440 @@
...
+    });
+
+    testWidgetsOnAllPlatforms('shares focus with SuperEditor', (tester) async {
```
matthew-carroll, line 134: This test seems overly specific. Do we actually care about the popover sharing focus with SuperEditor? Or do we care about sharing focus with any arbitrary widget that wants to share?

super_editor/test/infrastructure/popover_test.dart
```diff
@@ -0,0 +1,440 @@
...
+          extent: DocumentPosition(
+            nodeId: documentNode.id,
+            nodePosition: const TextNodePosition(offset: 5),
+          ),
+        ),
+      );
+    });
+  });
+
+  group('ItemSelectionList', () {
```
matthew-carroll, line 208: Should this group be in this test file?

matthew-carroll:
@angelosilvestre wanted to make sure you didn't forget about this PR.

super_editor/lib/src/infrastructure/selectable_list.dart
angelosilvestre: Updated.

super_editor/test/infrastructure/popover_test.dart
angelosilvestre: Updated.

super_editor/test/infrastructure/popover_test.dart
angelosilvestre: Updated.

angelosilvestre:
@matthew-carroll Updated the PR

super_editor/test/infrastructure/popover_test.dart
matthew-carroll: I'm still confused about why a single test is sufficient for this. Don't we have rules about where the popover appears based on surrounding space, and how large the popover is allowed to be? If so, then it seems like we should have a thorough collection of tests to lock down each of those behaviors...

angelosilvestre: Ok, this test is to ensure we honor an app-configured geometry. We don't have tests for our default geometry. I'll add those.

angelosilvestre:
@matthew-carroll Added teste for the default popover aligner.

super_editor/test/infrastructure/popover_test.dart
```diff
@@ -0,0 +1,493 @@
...
+          expect(find.byType(RoundedRectanglePopoverAppearance), findsOneWidget);
+
+          final buttonRect = tester.getRect(find.byKey(buttonKey));
+          final popoverRect = tester.getRect(find.byType(RoundedRectanglePopoverAppearance));
+
+          // Ensure popover was displayed below the button.
+          expect(popoverRect.top, greaterThan(buttonRect.bottom));
+        });
+
+        testWidgetsOnAllPlatforms('positions the popover above button if there is room', (tester) async {
```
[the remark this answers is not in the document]
angelosilvestre: Updated.

super_editor/test/infrastructure/popover_test.dart
```diff
@@ -0,0 +1,493 @@
...
+          popoverController.open();
+          await tester.pumpAndSettle();
+
+          // Ensure the popover is displayed.
+          expect(find.byType(RoundedRectanglePopoverAppearance), findsOneWidget);
+
+          final popoverRect = tester.getRect(find.byType(RoundedRectanglePopoverAppearance));
+
+          // Ensure popover was pinned of the bottom to the boundary widget.
+          expect(popoverRect.bottom, 600);
```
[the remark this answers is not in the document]
angelosilvestre: Yeah, updated.

super_editor/test/infrastructure/popover_test.dart
```diff
@@ -0,0 +1,493 @@
...
+
+          final buttonRect = tester.getRect(find.byKey(buttonKey));
+          final popoverRect = tester.getRect(find.byType(RoundedRectanglePopoverAppearance));
+
+          // Ensure popover was displayed above the button.
+          expect(popoverRect.bottom, lessThan(buttonRect.top));
+        });
+
+        testWidgetsOnAllPlatforms(
+            'pins the popover to the bottom if there isn\'t room neither below or above the button', (tester) async {
```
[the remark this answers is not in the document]
angelosilvestre: I modified the tests to use a different screen size and different widget sizes with a comment explaining why.

super_editor/test/infrastructure/popover_test.dart
```diff
@@ -0,0 +1,493 @@
...
+
+          final popoverRect = tester.getRect(find.byType(RoundedRectanglePopoverAppearance));
+
+          // Ensure popover was displayed pinned to the bottom of the screen.
+          expect(popoverRect.bottom, 600);
+        });
+      });
+    });
+
+    testWidgetsOnAllPlatforms('shares focus with other widgets', (tester) async {
```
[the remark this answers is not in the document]
angelosilvestre: Updated.

super_editor/test/infrastructure/popover_test.dart
```diff
@@ -0,0 +1,530 @@
...
+          expect(find.byType(RoundedRectanglePopoverAppearance), findsOneWidget);
+
+          final buttonRect = tester.getRect(find.byKey(buttonKey));
+          final popoverRect = tester.getRect(find.byType(RoundedRectanglePopoverAppearance));
+
+          // Ensure popover was displayed above the button.
+          expect(popoverRect.bottom, lessThan(buttonRect.top));
+        });
+
+        testWidgetsOnAllPlatforms('pins the popover to the bottom if there is not room below or above the button',
```
matthew-carroll, line 258: Can you clarify "to the bottom" of what? screen? boundary? button?

matthew-carroll:
In addition to earlier comments, I checked out this branch so I could work on paragraph alignment. I discovered that we're displaying too many alignment options in the drop down. This is because we're looping through all `TextAlign` values which means we're showing both `left` and `start`, and `right` and `end`. So we're essentially repeating equivalent options.

```dart
items: TextAlign.values
          .map(
            (alignment) => SuperEditorDemoIconItem(
              icon: _buildTextAlignIcon(alignment),
              id: alignment.name,
            ),
          )
          .toList(),
```

matthew-carroll:
@angelosilvestre I'm gonna merge this so that I can fix paragraph alignment, which I based off of this branch so that I could select the paragarph alignment from the toolbar. Please put up a followup PR with the final adjustments from the review.

matthew-carroll:
Also, please cherry pick this to stable, per usual.

PR #1540, opened 2023-10-21
[SuperEditor][SuperReader][Android] Fix long-press firing after scrolling and releasing the pointer (Resolves #1535)

### LINKED ISSUE #1535: [SuperEditor] Scrolling on Android in a CustomScrollView creates selection when it shouldn't

angelosilvestre:
[SuperEditor][SuperReader][Android] Fix long-press firing after scrolling and releasing the pointer. Resolves #1535

When the editor or reader is inside a CustomScrollview, trying to drag is opening the software keyboard and changing the selection. This was partially solved in https://github.com/superlistapp/super_editor/pull/1536. However one issue remained:

Starting a scroll gestures and the releasing the pointer is still causing the issue. The cause is that, when the timer fires, we aren't scrolling anymore.

To fix that, this PR cancels the long-press timer as soon as the user starts the scrolling gesture.

I kept the existing test, which simulates the user dragging and holding down the pointer, and added another one simulating the user starting a drag gesture and then releasing the pointer.

### DIFF
--- a/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
@@ -324,6 +324,12 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
 
     if (isScrolling) {
       _isScrolling = true;
+
+      // The long-press timer is cancelled if a pan gesture is detected.
+      // However, if we have an ancestor scrollable, we won't receive a pan gesture in this widget.
+      // Cancel the timer as soon as the user started scrolling.
+      _tapDownLongPressTimer?.cancel();
+      _tapDownLongPressTimer = null;
     } else {
       onNextFrame((_) {
         // Set our scrolling flag to false on the next frame, so that our tap handlers
@@ -497,13 +503,6 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
 
   // Runs when a tap down has lasted long enough to signify a long-press.
   void _onLongPressDown() {
-    if (_isScrolling) {
-      // When the editor has an ancestor scrollable, dragging won't trigger a pan gesture
-      // is this widget. Because of that, the timer still fires after the timeout.
-      // Do nothing to let the user scroll.
-      return;
-    }
-
     _longPressStrategy = AndroidDocumentLongPressSelectionStrategy(
       document: widget.document,
       documentLayout: _docLayout,
--- a/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
@@ -325,9 +325,8 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
     if (isScrolling) {
       _isScrolling = true;
 
-      // The long-press timer is cancelled if a pan gesture is detected.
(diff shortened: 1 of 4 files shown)


angelosilvestre:
@Jethro87 could you please try this PR?

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
```diff
@@ -324,6 +324,12 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
 
     if (isScrolling) {
       _isScrolling = true;
+
+      // The long-press timer is cancelled if a pan gesture is detected.
```
matthew-carroll, line 328: This explanation is too difficult to understand. It's not clear which part is describing what's happening now, or what isn't happening now, or what might be happening now. Please try to describe the relevant details in a way that a reader would understand by opening this file and looking directly at this explanation, like I'm doing now. 

For example, reading this without all the pre-existing conversation and investigation, I don't know what it means to say "The long-press timer is cancelled if a pan gesture is detected". Does that mean we're cancelling it now? Does that mean it's already cancelled? Does that mean this if-statement "detected a pan gesture"? Does it mean that sometime earlier a pan gesture was detected? The explanation creates far more questions than it answers.

Jethro87:
@angelosilvestre This looks good. I am not experiencing the issue on Android anymore.

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
angelosilvestre: I simplified the comment. Do you think it's clear enough?

super_editor/lib/src/super_reader/read_only_document_android_touch_interactor.dart
```diff
@@ -304,6 +304,12 @@ class _ReadOnlyAndroidDocumentTouchInteractorState extends State<ReadOnlyAndroid
 
     if (isScrolling) {
       _isScrolling = true;
+
+      // The long-press timer is cancelled if a pan gesture is detected.
```
matthew-carroll, line 308: I'm gonna merge this because there's a critical long-press ticket from Superlist, but this comment is still ordered in a way that makes it unnecessarily difficult to understand what it's trying to say.

Generally, the first thing you should say in a comment is what the code is actually doing. Beginning with contextual qualifiers is confusing because you haven't event established what you're qualifying yet.

Here's a comment in reversed order:

```
// Cancel the long-press timer because we're scrolling, and scrolling is mutually
// independent from long-pressing.
//
// We also cancel the timer when a pan gesture starts. However, we won't receive
// a pan gesture event if we have an ancestor scrollable. Therefore, we cancel the
// timer here, too.
```

matthew-carroll (approved):
LGTM

matthew-carroll:
I cherry-picked this to unblock some other work: https://github.com/superlistapp/super_editor/pull/1553

PR #1541, opened 2023-10-21
[SuperEditor][SuperReader][Android] Fix scroll physics (Resolves #1539)

### LINKED ISSUE #1539: [SuperEditor] Overscrolling can pin the editor on Android
This was tested on Flutter stable / super_editor stable in the example app. The following is the basic editor:

https://github.com/superlistapp/super_editor/assets/6467808/1f16b98d-4c38-4c32-9d30-108f664220a3

If you remove the SliverAppBar from the sliver_example_editor, the same thing happens.

My app:

https://github.com/superlistapp/super_editor/assets/6467808/a5944c36-32dd-4dd0-8b5d-2fe4b0410eb9

cc @matthew-carroll

angelosilvestre:
[SuperEditor][SuperReader][Android] Fix scroll physics. Resolves #1539

On Android, we are noticing underscroll/overscroll with the editor being pinned at the position where the drag ended:

https://github.com/superlistapp/super_editor/assets/6467808/1f16b98d-4c38-4c32-9d30-108f664220a3

Android uses `ClampingScrollPhysics`, so it shouldn't even allow for underscroll/overscroll. It seems that the way we handle scrolling doesn't honor the scroll physics correctly. 

I replaced the call to `jumpTo` with a call to `pointerScroll` and the issue seems to be solved: 

[android.webm](https://github.com/superlistapp/super_editor/assets/7597082/57b339f5-a84c-466d-bcdd-bd013e58c94a)

The difference between `jumpTo` and `pointerScroll` is that `pointerScroll` calls `forcePixels` clamping the jump offset between `minScrollExtent` and `maxScrollExtent`, whereas `jumpTo` calls it with the given jump offset.  `jumpTo`  also updates the `userScrollDirection`.

I'm not sure that using `pointerScroll` is the right way for us to handle scrolling. I also tried using `ScrollPosition.drag` at `onPanStart` and then `Drag.update` at `onPanUpdate`, but doing so caused the following exception:

```console
The following assertion was thrown while handling a gesture:
'package:flutter/src/widgets/scroll_activity.dart': Failed assertion: line 370 pos 12: 'details.primaryDelta != null': is not true.
scroll_activity.dart:370

#2      ScrollDragController.update scroll_activity.dart:370
#3      _AndroidDocumentTouchInteractorState._onPanUpdate document_gestures_touch_android.dart:785
#4      DragGestureRecognizer._checkUpdate.<anonymous closure> monodrag.dart:548
```
 
I'm open to other ideas to fix this issue.

### DIFF
--- a/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
@@ -145,7 +145,7 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
   final _longPressMagnifierGlobalOffset = ValueNotifier<Offset?>(null);
 
   /// Holds the drag gesture that scrolls the document.
-  Drag? _drag;
+  Drag? _dragToScroll;
 
   @override
   void initState() {
@@ -743,7 +743,7 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
 
     if (!_isLongPressInProgress) {
       // We only care about starting a pan if we're long-press dragging.
-      _drag = scrollPosition.drag(details, () {
+      _dragToScroll = scrollPosition.drag(details, () {
         // Allows receiving touches while scrolling due to scroll momentum.
         // This is needed to allow the user to stop scrolling by tapping down.
         scrollPosition.context.setIgnorePointer(false);
@@ -792,9 +792,9 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
       return;
     }
 
-    if (_drag != null) {
+    if (_dragToScroll != null) {
       // The user is trying to scroll the document. Change the scroll offset.
-      _drag!.update(details);
+      _dragToScroll!.update(details);
     }
   }
 
@@ -826,14 +826,10 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
       return;
     }
 
-    if (_drag != null) {
-      _drag!.end(details);
-    }
-
-    final pos = scrollPosition;
-    if (pos is ScrollPositionWithSingleContext) {
-      pos.goBallistic(-details.velocity.pixelsPerSecond.dy);
-      pos.context.setIgnorePointer(false);
+    if (_dragToScroll != null) {
+      // The user was performing a drag gesture to scroll the document.
+      // End the drag gesture.
+      _dragToScroll!.end(details);
     }
   }
 
@@ -843,8 +839,10 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
       return;
     }
 
-    if (_drag != null) {
-      _drag!.cancel();
+    if (_dragToScroll != null) {
+      // The user was performing a drag gesture to scroll the document.
+      // End the drag gesture.
+      _dragToScroll!.cancel();
     }
   }
 
--- a/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
@@ -144,6 +144,9 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
   AndroidDocumentLongPressSelectionStrategy? _longPressStrategy;
   final _longPressMagnifierGlobalOffset = ValueNotifier<Offset?>(null);
 
+  /// Holds the drag gesture that scrolls the document.
+  Drag? _drag;
+
   @override
   void initState() {
     super.initState();
@@ -740,6 +743,11 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
 
     if (!_isLongPressInProgress) {
       // We only care about starting a pan if we're long-press dragging.
+      _drag = scrollPosition.drag(details, () {
+        // Allows receiving touches while scrolling due to scroll momentum.
+        // This is needed to allow the user to stop scrolling by tapping down.
+        scrollPosition.context.setIgnorePointer(false);
+      });
       return;
     }
 
@@ -784,8 +792,10 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
       return;
     }
 
-    // The user is trying to scroll the document. Change the scroll offset.
-    scrollPosition.pointerScroll(-details.delta.dy);
+    if (_drag != null) {
+      // The user is trying to scroll the document. Change the scroll offset.
+      _drag!.update(details);
+    }
   }
 
   void _updateLongPressSelection(DocumentSelection newSelection) {
@@ -816,6 +826,10 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
       return;
     }
 
+    if (_drag != null) {
+      _drag!.end(details);
+    }
+
     final pos = scrollPosition;
     if (pos is ScrollPositionWithSingleContext) {
       pos.goBallistic(-details.velocity.pixelsPerSecond.dy);
@@ -827,6 +841,10 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
     if (_isLongPressInProgress) {
       _onLongPressEnd();
       return;
+    }
+
+    if (_drag != null) {
+      _drag!.cancel();
     }
   }
 
@@ -1251,10 +1269,11 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
                 ..gestureSettings = gestureSettings;
             },
           ),
-          PanGestureRecognizer: GestureRecognizerFactoryWithHandlers<PanGestureRecognizer>(
-            () => PanGestureRecognizer(),
-            (PanGestureRecognizer recognizer) {
+          VerticalDragGestureRecognizer: GestureRecognizerFactoryWithHandlers<VerticalDragGestureRecognizer>(
+            () => VerticalDragGestureRecognizer(),
+            (VerticalDragGestureRecognizer recognizer) {
               recognizer
+                ..dragStartBehavior = DragStartBehavior.down
                 ..onStart = _onPanStart
                 ..onUpdate = _onPanUpdate
                 ..onEnd = _onPanEnd
--- a/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
@@ -136,6 +136,10 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
   double? _dragStartScrollOffset;
   Offset? _globalDragOffset;
   Offset? _dragEndInInteractor;
+
+  /// Holds the drag gesture that scrolls the document.
+  Drag? _scrollingDrag;
+
   SelectionHandleType? _selectionType;
 
   Timer? _tapDownLongPressTimer;
@@ -143,9 +147,6 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
   bool get _isLongPressInProgress => _longPressStrategy != null;
   AndroidDocumentLongPressSelectionStrategy? _longPressStrategy;
   final _longPressMagnifierGlobalOffset = ValueNotifier<Offset?>(null);
-
-  /// Holds the drag gesture that scrolls the document.
-  Drag? _dragToScroll;
 
   @override
   void initState() {
@@ -743,7 +744,7 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
 
     if (!_isLongPressInProgress) {
       // We only care about starting a pan if we're long-press dragging.
-      _dragToScroll = scrollPosition.drag(details, () {
+      _scrollingDrag = scrollPosition.drag(details, () {
         // Allows receiving touches while scrolling due to scroll momentum.
         // This is needed to allow the user to stop scrolling by tapping down.
         scrollPosition.context.setIgnorePointer(false);
@@ -792,9 +793,9 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
       return;
     }
 
-    if (_dragToScroll != null) {
+    if (_scrollingDrag != null) {
       // The user is trying to scroll the document. Change the scroll offset.
-      _dragToScroll!.update(details);
+      _scrollingDrag!.update(details);
     }
   }
 
@@ -826,10 +827,10 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
       return;
     }
 
-    if (_dragToScroll != null) {
+    if (_scrollingDrag != null) {
       // The user was performing a drag gesture to scroll the document.
-      // End the drag gesture.
-      _dragToScroll!.end(details);
+      // End the scroll activity and let the document scrolling with momentum.
+      _scrollingDrag!.end(details);
     }
   }
 
@@ -839,10 +840,10 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
       return;
     }
 
-    if (_dragToScroll != null) {
+    if (_scrollingDrag != null) {
       // The user was performing a drag gesture to scroll the document.
       // End the drag gesture.
-      _dragToScroll!.cancel();
+      _scrollingDrag!.cancel();
     }
   }
 
--- a/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
@@ -785,7 +785,7 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
     }
 
     // The user is trying to scroll the document. Change the scroll offset.
-    scrollPosition.jumpTo(scrollPosition.pixels - details.delta.dy);
+    scrollPosition.pointerScroll(-details.delta.dy);
   }
 
   void _updateLongPressSelection(DocumentSelection newSelection) {
--- a/super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
@@ -998,20 +998,25 @@ class _IosDocumentTouchInteractorState extends State<IosDocumentTouchInteractor>
       ..hideMagnifier()
       ..blinkCaret();
 
-    if (_dragMode == DragMode.scroll) {
-      // The user was performing a drag gesture to scroll the document.
-      // End the scroll activity and let the document scrolling with momentum.
-      _scrollingDrag!.end(details);
-      _scrollingDrag = null;
-      _dragMode = null;
-
-      return;
-    }
-
-    if (_dragMode != null) {
-      // The user was dragging a selection change in some way, either with handles
-      // or with a long-press. Finish that interaction.
-      _onDragSelectionEnd();
+    switch (_dragMode) {
+      case DragMode.scroll:
+        // The user was performing a drag gesture to scroll the document.
+        // End the scroll activity and let the document scrolling with momentum.
+        _scrollingDrag!.end(details);
+        _scrollingDrag = null;
+        _dragMode = null;
+        break;
+      case DragMode.collapsed:
+      case DragMode.base:
+      case DragMode.extent:
+      case DragMode.longPress:
+        // The user was dragging a selection change in some way, either with handles
+        // or with a long-press. Finish that interaction.
+        _onDragSelectionEnd();
+        break;
+      case null:
+        // The user wasn't dragging over a selection. Do nothing.
+        break;
     }
   }
 
--- a/super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
@@ -293,7 +293,7 @@ class _IosDocumentTouchInteractorState extends State<IosDocumentTouchInteractor>
   IosLongPressSelectionStrategy? _longPressStrategy;
 
   /// Holds the drag gesture that scrolls the document.
-  Drag? _drag;
+  Drag? _dragToScroll;
 
   @override
   void initState() {
@@ -827,7 +827,10 @@ class _IosDocumentTouchInteractorState extends State<IosDocumentTouchInteractor>
       _dragMode = DragMode.extent;
       _dragHandleType = HandleType.downstream;
     } else {
-      _drag = scrollPosition.drag(details, () {
+      _dragMode = DragMode.scroll;
+
+      // Start a drag gesture to scroll the document.
+      _dragToScroll = scrollPosition.drag(details, () {
         // Allows receiving touches while scrolling due to scroll momentum.
         // This is needed to allow the user to stop scrolling by tapping down.
         scrollPosition.context.setIgnorePointer(false);
@@ -837,10 +840,12 @@ class _IosDocumentTouchInteractorState extends State<IosDocumentTouchInteractor>
 
     _controlsController!.doNotBlinkCaret();
     _controlsController!.hideToolbar();
-    _controlsController!.showToolbar();
+
     if (selection == null) {
       return;
     }
+
+    _controlsController!.showToolbar();
 
     _globalStartDragOffset = details.globalPosition;
     final interactorBox = context.findRenderObject() as RenderBox;
@@ -917,10 +922,9 @@ class _IosDocumentTouchInteractorState extends State<IosDocumentTouchInteractor>
   }
 
   void _onPanUpdate(DragUpdateDetails details) {
-    // If the user isn't dragging a handle, then the user is trying to
-    // scroll the document. Scroll it, accordingly.
-    if (_dragMode == null && _drag != null) {
-      _drag!.update(details);
+    if (_dragMode == DragMode.scroll && _dragToScroll != null) {
+      // The user is trying to scroll the document. Scroll it, accordingly.
+      _dragToScroll!.update(details);
       return;
     }
 
@@ -998,25 +1002,17 @@ class _IosDocumentTouchInteractorState extends State<IosDocumentTouchInteractor>
       ..hideMagnifier()
       ..blinkCaret();
 
-    if (_drag != null) {
-      _drag!.end(details);
-      _drag = null;
+    if (_dragToScroll != null) {
+      // The user was performing a drag gesture to scroll the document.
+      // End the drag gesture.
+      _dragToScroll!.end(details);
+      _dragToScroll = null;
+      _dragMode = null;
 
       return;
     }
 
-    if (_dragMode == null) {
-      // User was dragging the scroll area. Go ballistic.
-      if (scrollPosition is ScrollPositionWithSingleContext) {
-        (scrollPosition as ScrollPositionWithSingleContext).goBallistic(-details.velocity.pixelsPerSecond.dy);
-
-        if (_activeScrollPosition != scrollPosition) {
-          // We add the scroll change listener again, because going ballistic
-          // seems to switch out the scroll position.
-          _activeScrollPosition = scrollPosition;
-        }
-      }
-    } else {
+    if (_dragMode != null) {
       // The user was dragging a selection change in some way, either with handles
       // or with a long-press. Finish that interaction.
       _onDragSelectionEnd();
@@ -1026,9 +1022,12 @@ class _IosDocumentTouchInteractorState extends State<IosDocumentTouchInteractor>
   void _onPanCancel() {
     _magnifierOffset.value = null;
 
-    if (_drag != null) {
-      _drag!.cancel();
-      _drag = null;
+    if (_dragToScroll != null) {
+      // The user was performing a drag gesture to scroll the document.
+      // Cancel the drag gesture.
+      _dragToScroll!.cancel();
+      _dragToScroll = null;
+      return;
     }
 
     if (_dragMode != null) {
@@ -1319,6 +1318,8 @@ enum DragMode {
   // Dragging after a long-press, which selects by the word
   // around the selected word.
   longPress,
+  // Dragging to scroll the document.
+  scroll
 }
 
 /// Adds and removes an iOS-style editor toolbar, as dictated by an ancestor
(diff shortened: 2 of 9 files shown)


angelosilvestre:
@Jethro87 Could you please try this PR to see if it solves the issue for your app?

matthew-carroll:
@angelosilvestre is there any reason you chose to switch the scrolling method rather than check the min/max pixels and do nothing if there's nothing more to scroll?

super_editor/test/super_editor/supereditor_scrolling_test.dart
```diff
@@ -500,6 +500,76 @@ void main() {
       expect(SuperEditorInspector.findDocumentSelection(), isNull);
     });
 
+    testWidgetsOnAndroid("doesn't underscroll when dragging down", (tester) async {
```
matthew-carroll, line 503: Is "underscroll" the correct word? I've only heard of it as "overscroll" regardless of which direction you're scrolling.

Also, are you sure we don't already have tests that are supposed to be verifying this? I'm surprised if we don't, and if we do, something is wrong with those tests.

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
```diff
@@ -787,7 +787,7 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
     }
 
     // The user is trying to scroll the document. Change the scroll offset.
-    scrollPosition.jumpTo(scrollPosition.pixels - details.delta.dy);
+    scrollPosition.pointerScroll(-details.delta.dy);
```
matthew-carroll, line 790: Did you happen to check how we're handling this on iOS?

Jethro87:
@angelosilvestre I am not having this issue in my Android app after this update.

FYI on iOS, there is a similar over-scroll issue, though it does not "pin" the screen as it did on Android. Just in case you want to include iOS changes in this branch as well:


https://github.com/superlistapp/super_editor/assets/6467808/19205e37-23e6-4053-bdf8-f80558dd86b3

angelosilvestre:
@matthew-carroll An update on this.

After Jeff's last comment I took a deeper look into the iOS scrolling. Here's how it is right now:

https://github.com/superlistapp/super_editor/assets/7597082/7a6ed263-2554-40e3-b916-b45b70ad841f

On iOS, we are also using `jumpTo`. We can see in the video that we seem to be "fighting" the scroll physics. I'm convinced we shouldn't be using `jumpTo`, as each call to `jumpTo` is like an entire drag gesture: tapping down/dragging/releasing the tap.

Trying to use `pointerScroll` on iOS prevents the bouncing effect from happening. 

I'm looking into using `scrollPosition.drag`, here's how it's looking:

https://github.com/superlistapp/super_editor/assets/7597082/fb143b5d-3b1d-4e32-abc1-5a32bec9333d

I figured out why `Drag.update` wasn't working. Now, using this method on Android, we can see the stretch animation on scroll:

[android_dragging.webm](https://github.com/superlistapp/super_editor/assets/7597082/980eb3b4-8bf1-4109-916b-1c6e60b6fa61)

I'm now trying to figure out why "tap to stop scroll momentum" on iOS broke after this change.

angelosilvestre:
@matthew-carroll After Jeff's last comment I took a deeper look into the iOS scrolling. Here's how it looks right now:

https://github.com/superlistapp/super_editor/assets/7597082/7a6ed263-2554-40e3-b916-b45b70ad841f

On iOS, we are also using `jumpTo`. We can see in the video that we seem to be "fighting" the scroll physics. I'm convinced we shouldn't be using `jumpTo`, as each call to `jumpTo` is like an entire drag gesture: tapping down/dragging/releasing the tap. Also, dragging a bit and then holding down for a moment causes the editor to scroll back to the start. When holding the pointer down, the scroll position shouldn't jump back to the start.

I tried using `pointerScroll` on iOS, but then we can't overscroll and trigger the bouncing effect.

I changed this PR to use `ScrollPosition.drag`, here's how it's looking:

https://github.com/superlistapp/super_editor/assets/7597082/fb143b5d-3b1d-4e32-abc1-5a32bec9333d

Using this method on Android, we can see the stretch animation on scroll:

[android_dragging.webm](https://github.com/superlistapp/super_editor/assets/7597082/980eb3b4-8bf1-4109-916b-1c6e60b6fa61)

I mentioned in the PR description that using this API caused an exception on Android. On iOS, it worked without issues, so I looked at what's different between Android on iOS. The difference is that on Android we are using `PanGestureRecognizer` and on iOS `VerticalDragGestureRecognizer`. Switching the Android implementation to user `VerticalDragGestureRecognizer` solved the issue.

super_editor/test/super_editor/supereditor_scrolling_test.dart
angelosilvestre: I saw "underscroll" in a comment in the `ClampingScrollPhysics` class while debugging, so I thought it was a thing. I changed to "overscroll".

I didn't find tests for this...

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
angelosilvestre: Updated in https://github.com/superlistapp/super_editor/pull/1541#issuecomment-1776465833

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
```diff
@@ -809,15 +812,12 @@ class _IosDocumentTouchInteractorState extends State<IosDocumentTouchInteractor>
     //       placement during onTapDown, and then pick that up here. I think the little
     //       bit of slop might be the problem.
     final selection = widget.selection.value;
-    if (selection == null) {
```
matthew-carroll, line 812: By removing this null check, you've allowed all following behaviors to execute with a null selection. Is that really what you intended?

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
```diff
@@ -827,12 +827,20 @@ class _IosDocumentTouchInteractorState extends State<IosDocumentTouchInteractor>
       _dragMode = DragMode.extent;
       _dragHandleType = HandleType.downstream;
     } else {
+      _drag = scrollPosition.drag(details, () {
```
matthew-carroll, line 830: Please add a comment about why we're creating a `Drag` here.

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
```diff
@@ -911,8 +919,8 @@ class _IosDocumentTouchInteractorState extends State<IosDocumentTouchInteractor>
   void _onPanUpdate(DragUpdateDetails details) {
     // If the user isn't dragging a handle, then the user is trying to
     // scroll the document. Scroll it, accordingly.
-    if (_dragMode == null) {
-      scrollPosition.jumpTo(scrollPosition.pixels - details.delta.dy);
+    if (_dragMode == null && _drag != null) {
```
matthew-carroll, line 922: Should the drag mode have a value for scrolling or something? It seems strange that we're checking for a null drag mode and then checking for a non-null drag. From a reader's standpoint, that looks we're checking for opposites at the same time.

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
```diff
@@ -988,6 +996,13 @@ class _IosDocumentTouchInteractorState extends State<IosDocumentTouchInteractor>
     _magnifierOffset.value = null;
     _controlsController!.hideMagnifier();
 
+    if (_drag != null) {
+      _drag!.end(details);
```
matthew-carroll, line 1000: In these if-statements, please add a comment describing the scenario. It's not immediately obvious what `_drag != null` represents.

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
```diff
@@ -144,6 +144,9 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
   AndroidDocumentLongPressSelectionStrategy? _longPressStrategy;
   final _longPressMagnifierGlobalOffset = ValueNotifier<Offset?>(null);
 
+  /// Holds the drag gesture that scrolls the document.
+  Drag? _drag;
```
matthew-carroll, line 148: Given all the types of dragging that we implement, should this variable be given a name that's specific to how we're actually using it?

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
angelosilvestre: Those behaviors won't be executed. `_isOverBaseHandle` and `_isOverExtentHandle` return `false` if there isn't a selection.

Also, I added a guard clause after all of the if statements to return early if there isn't a selection.

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
angelosilvestre: It makes sense. I included a `DragMode` for scrolling.

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
```diff
@@ -144,6 +144,9 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
   AndroidDocumentLongPressSelectionStrategy? _longPressStrategy;
   final _longPressMagnifierGlobalOffset = ValueNotifier<Offset?>(null);
 
+  /// Holds the drag gesture that scrolls the document.
+  Drag? _dragToScroll;
```
matthew-carroll, line 148: Can we move this up to a reasonable position within the existing drag properties?

Also let's call this `_scrollingDrag` which implies that the subject is "Drag"

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
```diff
@@ -816,10 +826,10 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
       return;
     }
 
-    final pos = scrollPosition;
-    if (pos is ScrollPositionWithSingleContext) {
-      pos.goBallistic(-details.velocity.pixelsPerSecond.dy);
-      pos.context.setIgnorePointer(false);
+    if (_dragToScroll != null) {
+      // The user was performing a drag gesture to scroll the document.
+      // End the drag gesture.
```
matthew-carroll, line 831: Rather than say "End the drag gesture", please describe the impact of this line of code. As your comment mentions above, the user has already ended the gesture, so it's unclear what it means for code to "End the drag gesture"

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
matthew-carroll: The change seems likely to create confusion when reading the code. We have if-statements that check if the user is pressing a handle, but moving this selection check creates the possibility that there is no handle, which probably isn't expected when you're reading code that checks for it.

Is there are more expectable order of code that won't lead to surprises?

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
```diff
@@ -909,10 +922,9 @@ class _IosDocumentTouchInteractorState extends State<IosDocumentTouchInteractor>
   }
 
   void _onPanUpdate(DragUpdateDetails details) {
-    // If the user isn't dragging a handle, then the user is trying to
-    // scroll the document. Scroll it, accordingly.
-    if (_dragMode == null) {
-      scrollPosition.jumpTo(scrollPosition.pixels - details.delta.dy);
+    if (_dragMode == DragMode.scroll && _dragToScroll != null) {
```
matthew-carroll, line 925: Should it ever be possible that our drag mode is `scroll` and `_dragToScroll` is `null`?

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
```diff
@@ -990,18 +1002,17 @@ class _IosDocumentTouchInteractorState extends State<IosDocumentTouchInteractor>
...
-          // We add the scroll change listener again, because going ballistic
-          // seems to switch out the scroll position.
-          _activeScrollPosition = scrollPosition;
-        }
-      }
-    } else {
+      return;
+    }
+
+    if (_dragMode != null) {
```
matthew-carroll, line 1015: Between this if-statement and the previous one, we're creating an easy opportunity for future developers to create bugs here. Though it's not immediately obvious, this 2nd if-statement isn't merely looking for a non-null `_dragMode`, it's looking for a `_dragMode` that isn't `scroll`. The average reader won't make that connection.

Please rework these two if-statements so that it's clear that there's two related conditions, not two independent conditions. In general, if `_dragToScroll` is always non-null when `_dragMode` is `scroll`, then we shouldn't check for null `_dragToScroll`, we should just look for a `_dragMode` of `scroll`.

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
angelosilvestre: Updated.

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
angelosilvestre: It shouldn't be possible. Updated.

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
angelosilvestre: Updated.

matthew-carroll: I don't think this latest version has resolved the concern. We still have two independent if-statements, and the 2nd one depends on the 1st one. This isn't obvious.

Please switch this to a switch statement and have it do nothing on `null`

super_editor/test/super_editor/supereditor_scrolling_test.dart
```diff
@@ -500,6 +500,155 @@ void main() {
...
+          .withSingleParagraph()
+          .withInputSource(TextInputSource.ime)
+          .withEditorSize(const Size(300, 300))
+          .withScrollController(scrollController)
+          .pump();
+
+      // Ensure the scrollview didn't start scrolled.
+      expect(scrollController.offset, 0);
+
+      final editorRect = tester.getRect(find.byType(SuperEditor));
```
matthew-carroll, line 517: I don't think we need to get the editor rectangle do we? Don't we just need to attempt to scroll by some significant amount?

super_editor/test/super_editor/supereditor_scrolling_test.dart
```diff
@@ -500,6 +500,155 @@ void main() {
...
+
+      const dragFrameCount = 10;
```
matthew-carroll, line 519: I assume this multi-frame drag gesture is repeated in these tests. I think I've done similar things elsewhere. Let's define an extension on `WidgetTester` that does this for an arbitrary `dragPerFrame` `Offset`.

super_editor/test/super_editor/supereditor_scrolling_test.dart
```diff
@@ -500,6 +500,155 @@ void main() {
...
+      final dragAmountPerFrame = editorRect.height / dragFrameCount;
+
+      // Drag from the top all the way up to the bottom of the editor.
+      final dragGesture = await tester.startGesture(editorRect.topCenter);
+      for (int i = 0; i < dragFrameCount; i += 1) {
+        await dragGesture.moveBy(Offset(0, dragAmountPerFrame));
+        await tester.pump();
+
+        // Ensure we don't scroll.
+        expect(scrollController.offset, 0);
```
matthew-carroll, line 529: Can't we check this one time after the drag is over?

super_editor/test/super_editor/supereditor_scrolling_test.dart
```diff
@@ -500,6 +500,155 @@ void main() {
...
+      }
+
+      // End the gesture.
+      await dragGesture.up();
+      await dragGesture.removePointer();
+      await tester.pumpAndSettle();
```
matthew-carroll, line 535: Do we need a full pump and settle here? Or just a pump? Or nothing at all?

super_editor/test/super_editor/supereditor_scrolling_test.dart
```diff
@@ -500,6 +500,155 @@ void main() {
       expect(SuperEditorInspector.findDocumentSelection(), isNull);
     });
 
+    testWidgetsOnAndroid("doesn't overscroll when dragging down", (tester) async {
+      final scrollController = ScrollController();
+
+      await tester
+          .createDocument()
+          .withSingleParagraph()
+          .withInputSource(TextInputSource.ime)
```
matthew-carroll, line 509: Do we care that this is IME? Why not use the default input source?

super_editor/test/super_editor/supereditor_scrolling_test.dart
```diff
@@ -500,6 +500,155 @@ void main() {
...
+          .withEditorSize(const Size(300, 300))
```
matthew-carroll, line 510: Do we care about the editor size in this test?

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
angelosilvestre: Updated.

super_editor/test/super_editor/supereditor_scrolling_test.dart
angelosilvestre: We need to pump and settle to let the timer resolve. Added a comment explaining.

super_editor/test/super_editor/supereditor_scrolling_test.dart
angelosilvestre: We don't. Removed.

super_editor/test/super_editor/supereditor_scrolling_test.dart

super_editor/test/super_editor/supereditor_scrolling_test.dart
angelosilvestre: I think that instead of passing a `dragPerFrame`, it makes more sense to pass the starting location and the total amount of dragging, letting the extension compute the `dragPerFrame`. What do you think?

super_editor/test/super_editor/supereditor_scrolling_test.dart

super_editor/test/super_editor/supereditor_scrolling_test.dart
angelosilvestre: Nope, removed.

super_editor/test/super_editor/supereditor_scrolling_test.dart
```diff
@@ -500,6 +501,133 @@ void main() {
...
+      await tester //
+          .createDocument()
+          .withSingleParagraph()
+          .withScrollController(scrollController)
+          .pump();
+
+      // Ensure the editor didn't start scrolled.
+      expect(scrollController.offset, 0);
+
+      // Drag an amount of pixels chosen experimentally from the top of the editor.
```
matthew-carroll, line 516: This says "pixels chosen experimentally" - what was the experiment? Isn't this amount arbitrary?

super_editor/test/super_editor/supereditor_scrolling_test.dart
```diff
@@ -500,6 +501,133 @@ void main() {
...
+      final dragGesture = await tester.dragInMultipleFrames(
+        startLocation: tester.getTopLeft(find.byType(SuperEditor)),
```
matthew-carroll, line 518: Let's not start a drag from the top left corner of the editor. That's a very unnatural place for a drag to ever begin with a real user, and it creates opportunities for things like safe areas, drawer drag areas, or other edge-based effects to interfere. A more natural place to test the drag would be the horizontal center, and at least some distance down from the top of the editor.

super_editor/test/super_editor/supereditor_scrolling_test.dart
matthew-carroll: Can you be specific with the comment? Which timer(s) are we waiting for?

super_editor/test/super_editor/supereditor_scrolling_test.dart
```diff
@@ -694,6 +822,204 @@ void main() {
         expect(tester.testTextInput.hasAnyClients, isFalse);
       });
 
+      testWidgetsOnAndroid("doesn't overscroll when dragging down", (tester) async {
+        final scrollController = ScrollController();
+
+        await tester
+            .createDocument()
+            .withSingleParagraph()
+            .withCustomWidgetTreeBuilder(
```
matthew-carroll, line 831: Do we need any of this custom tree complexity for this test?

super_editor/test/test_tools.dart
```diff
@@ -16,3 +16,27 @@ class TestUrlLauncher implements UrlLauncher {
     return true;
   }
 }
+
+/// Extension on [WidgetTester] to make it easier to perform drag gestures.
+extension DragExtensions on WidgetTester {
+  /// Drags from the [startLocation] by [dragAmount] in multiple frames.
```
matthew-carroll, line 22: Please replace with:

```
/// Simulates a user drag from [startLocation] to `startLocation + dragAmount`.
///
/// Starts a gesture at [startLocation] and repeatedly drags the gesture 
/// across [frameCount] frames, pumping a frame between each drag. 
/// The gesture moves a distance each frame that's calculated as 
/// `dragAmount / frameCount`.
///
/// This method does not call `pumpAndSettle()`, so that the client can inspect
/// the app state immediately after the drag completes.
///
/// The client must call [TestGesture.up] on the returned [TestGesture].
```

super_editor/test/test_tools.dart
```diff
@@ -16,3 +16,27 @@ class TestUrlLauncher implements UrlLauncher {
...
+  ///
+  /// The [dragAmount] is distributed evenly between [frameCount] frames.
+  ///
+  /// The caller must end the gesture by calling [TestGesture.up].
+  Future<TestGesture> dragInMultipleFrames({
+    required Offset startLocation,
+    required Offset dragAmount,
+    required int frameCount,
```
matthew-carroll, line 30: Let's reduce our verbosity a little bit by declaring a default frame count of `10`, or some other reasonable default. Then it can be omitted for most uses.

super_editor/test/test_tools.dart
```diff
@@ -16,3 +16,27 @@ class TestUrlLauncher implements UrlLauncher {
...
 }
+
+/// Extension on [WidgetTester] to make it easier to perform drag gestures.
+extension DragExtensions on WidgetTester {
+  /// Drags from the [startLocation] by [dragAmount] in multiple frames.
+  ///
+  /// The [dragAmount] is distributed evenly between [frameCount] frames.
+  ///
+  /// The caller must end the gesture by calling [TestGesture.up].
+  Future<TestGesture> dragInMultipleFrames({
```
matthew-carroll, line 27: Let's call this `dragByFrameCount()` - in the future we might introduce a `dragByDistance()` and/or `dragByVelocity()`.

super_editor/test/test_tools.dart
```diff
@@ -16,3 +16,27 @@ class TestUrlLauncher implements UrlLauncher {
...
+    required Offset startLocation,
+    required Offset dragAmount,
```
matthew-carroll, line 29: Let's call this `totalDragOffset` to make it crystal clear that this isn't the per-frame drag amount.

super_editor/test/super_editor/supereditor_scrolling_test.dart
angelosilvestre: Yeah, it's arbitrary.

super_editor/test/super_editor/supereditor_scrolling_test.dart
angelosilvestre: Updated.

super_editor/test/super_editor/supereditor_scrolling_test.dart
angelosilvestre: We need to setup an ancestor `CustomScrollView` for this test.

super_editor/test/super_editor/supereditor_scrolling_test.dart
angelosilvestre: We wait the long-press timer to resolve. Should I add that to the comment?

matthew-carroll: Yeah, if we're specifically waiting for a long-press timer, then we should specify that in the comment. Saying "any pending timers" means there might be any number of timers doing any number of things. That makes it difficult to come back to this a year from now and know which timers we thought we were waiting for

super_editor/test/super_editor/supereditor_scrolling_test.dart
matthew-carroll: Ok. Why do we need to constrain the height to 200px?

Also, if we need the same tree in multiple tests, is there any reason not to extract it to either a private widget or a private method? Or, perhaps what would make even more sense is to either break it into a widget shared within the test suite, or even add it as a configuration option to the builder, because it looks like there are tests across multiple files that need this same setup.

angelosilvestre: Added methods to the test builder to configure the widget tree with a `CustomScrollView`.

super_editor/test/super_editor/supereditor_scrolling_test.dart
```diff
@@ -589,23 +713,9 @@ void main() {
...
-                        ),
-                      ],
-                    ),
-                  ),
-                ),
-              ),
-            )
+            .withEditorSize(const Size(200, 200))
+            .insideCustomScrollView()
+            .withCustomScrollViewScrollController(scrollController)
```
matthew-carroll, line 718: Should this be an optional argument to `insideCustomScrollView()` instead? Would it ever make sense to call `withCustomScrollViewController()` but without having previously called `insideCustomScrollView()`?

super_editor/test/super_editor/supereditor_scrolling_test.dart
```diff
@@ -694,6 +790,142 @@ void main() {
...
+        // Drag an arbitrary amount of pixels from the bottom of the editor.
+        final dragGesture = await tester.dragByFrameCount(
+          startLocation: tester.getRect(find.byType(CustomScrollView)).bottomCenter - const Offset(0, 10),
+          totalDragOffset: const Offset(0, -400.0),
+        );
+
+        // Ensure we don't scroll.
+        expect(scrollController.offset, scrollController.position.maxScrollExtent);
+
+        // Ensure we don't scroll.
```
matthew-carroll, line 847: Is this a duplicated condition?

super_editor/test/super_editor/supereditor_test_tools.dart
```diff
@@ -397,6 +415,21 @@ class TestSuperEditorConfigurator {
     return superEditor;
   }
 
+  /// Places [child] inside a [CustomScrollView], based on configurations in this class.
+  Widget _buildAncestorScrollable({required Widget child}) {
+    if (_config.insideCustomScrollView) {
```
matthew-carroll, line 420: How about we reverse this guard statement to check not inside scroll view and return `child`? That seems to be the simpler and more exceptional condition for a method named `_buildAncestorScrollable()`.

super_editor/test/super_editor/supereditor_test_tools.dart
```diff
@@ -474,6 +507,9 @@ class SuperEditorTestConfiguration {
   final plugins = <SuperEditorPlugin>{};
 
   WidgetTreeBuilder? widgetTreeBuilder;
+
+  bool insideCustomScrollView = false;
```
matthew-carroll, line 511: These properties look out of place given the order and separation of everything else.

super_editor/test/super_reader/super_reader_scrolling_test.dart
```diff
@@ -220,6 +221,128 @@ void main() {
...
+      await tester.pumpAndSettle();
+    });
+
+    testWidgetsOnAndroid("doesn't overscroll when dragging up", (tester) async {
+      final scrollController = ScrollController();
+
+      await tester //
+          .createDocument()
+          .withSingleParagraph()
+          .withScrollController(scrollController)
```
matthew-carroll, line 258: I'm noticing that we have this existing scroll controller configuration. I assume there's one in the `SuperEditor` configuration, too? If so, is there a reason that we have separate configurations for the ancestor scrollable vs the built-in scrollable? Would it make sense to apply any given scroll controller to whichever scrollable ew happen to use?

super_editor/test/super_editor/supereditor_scrolling_test.dart
angelosilvestre: Yeah, I also thought about that. I guess it makes sense to add the `scrollController` as an optional argument.

super_editor/test/super_editor/supereditor_scrolling_test.dart
angelosilvestre: Yeah. Removed the duplicate.

super_editor/test/super_reader/super_reader_scrolling_test.dart
angelosilvestre: I added this separate configuration to make it clear that the `ScrollController` is attached to the ancestor scrollable. I'm also not sure if we'll ever need to write at test where we configure the `ScrollController`s to both `SuperEditor` and the ancestor scrollable.

Should I just use the existing configuration?

super_editor/test/super_editor/supereditor_test_tools.dart
angelosilvestre: Moved up right below the `scrollController`.

super_editor/test/super_reader/super_reader_scrolling_test.dart
matthew-carroll: What would it mean to provide a different scroll controller for each case? Can we actually have independent scroll controllers for SuperEditor vs the ancestor scrollable?

angelosilvestre: Currently we don't. But if in the future we need to place an editor with a fixed size inside a scrollable we might need.

matthew-carroll: doesnt that fixed size editor still take control of its ancestor scrollable?

angelosilvestre: Currently yes, but I mean if at some point we support a scenario where the editor with a fixed size first scrolls its content, then scrolls the ancestor scrollable

matthew-carroll: I agree that such a thing is likely, but I think at this moment that capability would require some unknown reworking of the scrolling system, so it's unclear if this configuration approach would work, even after making that change. So let's take a look at it from the perspective of what we have today. Do these two different controllers actually control different things? Or are they two different methods that effectively configure the same control?

angelosilvestre: With what we have today, these two configurations have the same effect.

matthew-carroll: Ok, then let's stick to a single configuration for the scroll controller. We will revisit that when we know how other forms of scrolling work.

matthew-carroll (approved):
LGTM

PR #1557, opened 2023-10-24
Fix: Caret not blinking on iOS after dragging to scroll (Resolves #1549)

### LINKED ISSUE #1549: [SuperEditor] iOS carat stops blinking on scroll
On the latest stable super_editor stable (https://github.com/superlistapp/super_editor/pull/1546), the iOS carat will stop blinking when the user scrolls.

https://github.com/superlistapp/super_editor/assets/6467808/da085258-6d65-4974-abc1-b5d333d77d41

matthew-carroll:
Fix: Caret not blinking on iOS after dragging to scroll (Resolves #1549)

The root cause was that we were disabling blinking on touch down, and on pan start, but we weren't re-enabling blinking on pan end in the case that the user was scrolling.

There wasn't any great way to add tests for this fix because we purposefully prevent caret blinking during tests so that `pumpAndSettle()` doesn't throw an exception.


matthew-carroll:
CC @Jethro87

angelosilvestre (commented):
LGTM

PR #1573, opened 2023-10-26
Fixed tap to show toolbar on iOS when tapping on caret a second time (Resolves #1567)

### LINKED ISSUE #1567: [SuperEditor] Single tap should bring up toolbar on iOS
Previously, single-tapping would bring up the Cut/Copy/Paste toolbar. Example of expected behavior (from Notes app):

https://github.com/superlistapp/super_editor/assets/6467808/c7837377-1a8c-4255-acfb-58e116fcafed

Now, it's very difficult to bring up the toolbar in super_editor without double-tapping an expanded selection. Very rarely am I able to get it to appear by single-tapping. Instead, I have to double-tap some text (which will bring up the toolbar), then I can go back and single-tap the original node to bring up the toolbar. The toolbar does appear but in the wrong position.

Here's a demo:

https://github.com/superlistapp/super_editor/assets/6467808/145ccb2a-b25a-438f-91b0-5c307a34fc85

matthew-carroll:
Fixed tap to show toolbar on iOS when tapping on caret a second time (Resolves #1567)

There were a few things that lead to this issue:

- The toolbar focal point layer was explicitly ignoring collapsed selections, now it places a leader when the selection is collapsed, too.
- The check for tapping on the same selection was giving false negatives due to differences in text affinity based on where the user taps. Now we use a different comparison method to avoid those mismatches.
- We didn't have a test for this, as far as I could tell, so I added one. However, this new test was passing even before I made the other changes in this PR. Therefore, this test is probably useless, but I can't figure out why.

### DIFF
--- a/super_editor/lib/src/infrastructure/platforms/ios/ios_document_controls.dart
+++ b/super_editor/lib/src/infrastructure/platforms/ios/ios_document_controls.dart
@@ -441,6 +441,7 @@ class _IosToolbarFocalPointDocumentLayerState extends DocumentLayoutLayerState<I
       // The selection didn't change from what it was the last time we calculated selection bounds.
       return;
     }
+    _selectionUsedForMostRecentLayout = selection;
 
     // The selection changed, which means the selection bounds changed, we need to recalculate the
     // toolbar focal point bounds.
--- a/super_editor/lib/src/infrastructure/platforms/ios/ios_document_controls.dart
+++ b/super_editor/lib/src/infrastructure/platforms/ios/ios_document_controls.dart
@@ -1,6 +1,5 @@
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
-import 'package:flutter/scheduler.dart';
 import 'package:follow_the_leader/follow_the_leader.dart';
 import 'package:overlord/follow_the_leader.dart';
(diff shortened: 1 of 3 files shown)


matthew-carroll:
CC @Jethro87

super_editor/lib/src/infrastructure/platforms/ios/ios_document_controls.dart
```diff
@@ -410,7 +409,7 @@ class IosToolbarFocalPointDocumentLayer extends DocumentLayoutLayerStatefulWidge
 
 class _IosToolbarFocalPointDocumentLayerState extends DocumentLayoutLayerState<IosToolbarFocalPointDocumentLayer, Rect>
     with SingleTickerProviderStateMixin {
-  bool _wasSelectionExpanded = false;
+  DocumentSelection? _selectionUsedForMostRecentLayout;
```
angelosilvestre, line 412: It seems we are not assigning a value to `_selectionUsedForMostRecentLayout` anywhere in this file.

angelosilvestre (approved):
LGTM

The only explanation that comes to my mind to why this test passes without the changes is that the taps are being recognized as a double tap, but I looked at the implementation and we already call `pump` with a duration to avoid that...

PR #1593, opened 2023-11-10
Migrate to KeyEvent from RawKeyEvent

gspencergoog:
## Description

This does a major migration for SuperEditor to use `KeyEvent` instead of using `RawKeyEvent`-based events.

The `RawKeyEvent` system will shortly be deprecated by the Flutter team, and this PR fixes SuperEditor to handle that deprecation.

Unfortunately, this requires changing the public interface of SuperEditor, since some of the keyboard event handlers need to now take `KeyEvent`s instead of `RawKeyEvent`s.

The `RawKeyEvent`-based system has had deprecation warnings ("this will be deprecated" in the comments, not actual `@Deprecated` annotations) for a couple of years, and Flutter is finally officially deprecating that system, since the replacement has proven to be robust.

@matthew-carroll I haven't bumped any version numbers or anything here, since I'm not quite sure how you'd like to proceed with that, or how you want to handle this breaking change.

## Related Issues
 - https://github.com/flutter/flutter/issues/136419

## Related PRs
- https://github.com/flutter/flutter/pull/136854
- https://github.com/flutter/flutter/pull/136677

## Tests
 - Modified tests to match.


matthew-carroll:
@gspencergoog is this only a breaking change in terms of API? Or are we losing behaviors that we depend upon? If it's just an API breaking change, can you please make the replacements and ensure all tests still pass?

@miguelcmedeiros @brian-superlist @Jethro87 @jmatth - Please be aware of this PR from the Flutter team

matthew-carroll:
PS - Goldens are expected to fail at the moment due to some minor adjustments for mobile controls. But I think all other tests should be made to pass.

gspencergoog:
As far as I can tell, I don't think you're losing any behaviors that you depend upon.

To be clear, removing the old system does remove some functionality (things having to do with specific platform implementations of key events in the `RawKeyEventData` subclasses), but I don't think you're using any of them.

Tests did pass locally for me.  I'll look into the ones that aren't passing on CI.

gspencergoog:
Oh, actually, I know why they're failing: they depend on a (breaking) change that hasn't landed yet in Flutter: https://github.com/flutter/flutter/pull/136854

matthew-carroll:
Ok, can you let us know in this PR when the other change is merged? We wouldn't want to merge a PR that breaks us against `master`.

gspencergoog:
> Ok, can you let us know in this PR when the other change is merged? We wouldn't want to merge a PR that breaks us against master.

Yes, of course.  How exactly do you want to do "the dance" here?  At some point, either when I merge the breaking change to the framework, or when we merge this change here before the framework is updated, your repo won't build against master for a short period of time. Once either one happens, presumably you wouldn't be publishing the package until they are in sync again and passing.

It sounds like you'd prefer me to push the framework change first, and then update this change to validate against the new framework master before you commit this change here. Does that sound right?

gspencergoog:
It looks like there are about 50 different APIs in SuperEditor that take a `RawKeyEvent` that I've converted to use `KeyEvent` here, but in order to land the change in the Framework, I'll either need to convert all of those to dynamic (and have duplicate code to handle either), commit them here, and then convert them all back again and remove the duplicate code once the framework PR lands, or we'll need to disable the SuperEditor customer tests temporarily until both the Framework change and this PR land.

Would you be OK with disabling the customer tests temporarily to save us some time and avoid the churn in your code?

matthew-carroll:
> Would you be OK with disabling the customer tests temporarily to save us some time and avoid the churn in your code?

Yes, as long as we re-enable them ASAP.

gspencergoog:
Yes, the plan would be to re-enable them ASAP (we want them enabled as much as you do!)

I wouldn't disable them until both of the PRs were ready and approved, and then I'd submit the framework PR, and then rebase this PR, submitting it (with your cooperation) as soon as the presubmits passed, and then re-enable the customer tests right after that.

gspencergoog:
@matthew-carroll Okay, the Flutter change is ready to be committed.  Once you approve https://github.com/flutter/tests/pull/312, I'll submit that, let the presubmit pass on https://github.com/flutter/flutter/pull/136854 and then submit it.  Once that's submitted, I'll push an empty change to this PR to re-run the presubmits. Then, once this PR is approved, you can submit it and I'll re-enable the customer_testing tests with an update ref. Sound good?

gspencergoog:
It looks like 1) there are the (expected, I think) golden file failures, and 2) the deploy tests are failing because it's trying to build this PR with the stable branch.  Is that expected?

super_editor/example/lib/demos/in_the_lab/popover_list.dart
```diff
@@ -102,7 +102,7 @@ class _PopoverListState extends State<PopoverList> {
       return KeyEventResult.ignored;
     }
 
-    if (event is KeyDownEvent) {
+    if (event is KeyDownEvent || event is KeyRepeatEvent) {
```
matthew-carroll, line 105: Why is `KeyRepeatEvent` relevant to this if-statement?

super_editor/lib/src/infrastructure/focus.dart
```diff
@@ -28,6 +29,9 @@ class NonReparentingFocus extends StatefulWidget {
   /// The callback invoked whenever [focusNode] receives key events.
   final FocusOnKeyCallback? onKey;
```
matthew-carroll, line 30: If Flutter is completely removing the raw key events then does `Focus` still declare `onKey`? This widget mirrors `Focus`.

super_editor/lib/src/infrastructure/keyboard.dart
```diff
@@ -3,6 +3,9 @@
 import 'package:flutter/foundation.dart';
 import 'package:flutter/widgets.dart';
 
+import 'package:super_editor/src/infrastructure/raw_key_event_extensions.dart';
```
matthew-carroll, line 6: Was this import added on purpose? The changes in this file don't seem like they should require any new imports.

super_editor/lib/src/infrastructure/raw_key_event_extensions.dart
```diff
@@ -1,9 +1,36 @@
...
+extension IsArrowKeyExtension on KeyEvent {
   bool get isArrowKeyPressed =>
       logicalKey == LogicalKeyboardKey.arrowUp ||
       logicalKey == LogicalKeyboardKey.arrowDown ||
       logicalKey == LogicalKeyboardKey.arrowLeft ||
       logicalKey == LogicalKeyboardKey.arrowRight;
 }
+
+
+extension IsKeyPressed on KeyEvent {
```
matthew-carroll, line 12: What's the story behind this new extension?

super_editor/lib/src/super_reader/read_only_document_keyboard_interactor.dart
```diff
@@ -5,6 +5,8 @@ import 'package:super_editor/src/core/document_layout.dart';
 import 'package:super_editor/src/document_operations/selection_operations.dart';
 import 'package:super_editor/src/infrastructure/_logging.dart';
 import 'package:super_editor/src/infrastructure/keyboard.dart';
+import 'package:super_editor/src/infrastructure/raw_key_event_extensions.dart';
```
matthew-carroll, line 8: What is this import adding?

super_editor/lib/src/infrastructure/raw_key_event_extensions.dart
```diff
@@ -1,9 +1,36 @@
 import 'package:flutter/services.dart';
 
-extension IsArrowKeyExtension on RawKeyEvent {
+extension IsArrowKeyExtension on KeyEvent {
```
matthew-carroll, line 3: If this file declares things on `KeyEvent` instead of `RawKeyEvent`, then should the file name be changed to say that?

super_editor/test/super_editor/text_entry/tagging/action_tags_test.dart
```diff
@@ -678,9 +679,9 @@ Future<TestDocumentContext> _pumpTestEditor(
 
 ExecutionInstruction _submitOnEnter({
   required SuperEditorContext editContext,
-  required RawKeyEvent keyEvent,
+  required KeyEvent keyEvent,
 }) {
-  if (keyEvent is KeyDownEvent) {
+  if (keyEvent is! KeyDownEvent && keyEvent is! KeyRepeatEvent) {
```
matthew-carroll, line 684: This appears to be an inverted condition as compared to the original code

super_editor/test/super_editor/text_entry/text_test.dart
```diff
@@ -50,19 +50,16 @@ void main() {
     });
 
     group('TextComposable text entry', () {
-      test('it does nothing when meta is pressed', () {
+      testWidgets('it does nothing when meta is pressed', (WidgetTester tester) async {
```
matthew-carroll, line 53: This was a Dart test on purpose. Please switch it back and update `FakeRawKeyEventData` to make that possible.

matthew-carroll:
@angelosilvestre @rutvik110 - Please do a review pass on this PR to see if you find anything that I missed.

@miguelcmedeiros @jmatth @Jethro87 - I've done an initial review of this PR from the Flutter team. Once that the review completes, I plan to merge this in. So if you have any input, now is the time.

rutvik110:
LGTM!

super_editor/example/lib/demos/in_the_lab/popover_list.dart
gspencergoog: Because `KeyRepeatEvent` is effectively a `KeyDownEvent` for repeated keys, and if you only want to handle key up events, you don't want to handle repeat events.

super_editor/test/super_editor/text_entry/tagging/action_tags_test.dart
gspencergoog: Actually, with this the way it used to be (happening only on Enter key up), then by the time the request is exectuted, the tag for the position is gone at this line: https://github.com/gspencergoog/super_editor/blob/migrate_key_event/super_editor/lib/src/default_editor/text_tokenizing/action_tags.dart#L141-L142 causing it to not execute the request, and failing the test.

Is it expected that it remove the tag? Is something else handling the Enter key down that removes the tag?

gspencergoog: All the tests pass when this is set to handle things on Enter key down (but maybe then it's not testing what it used to test?)

super_editor/example/lib/demos/in_the_lab/popover_list.dart
matthew-carroll: I think the purpose of our existing conditional is less about "up vs down" and it's more about not reacting to key press too early. In the case of a singular "down and up" key press, we want to respond to the "up". However, if the user holds down the key and generates repeated events, I imagine that most/all of these scenarios want to respond to repeated events too. Consider a key handler that inserts characters, the user will expect repeated characters. Consider an arrow key to move the caret, the user will expect repeated motion.

@angelosilvestre do you have any thoughts on this?

super_editor/lib/src/infrastructure/keyboard.dart
matthew-carroll: Can you provide some elaboration as to why that behavior moved on the Flutter side? Why were we able to ask that directly with `RawKeyEvent` but we can't ask those questions directly of a `KeyEvent`?

super_editor/lib/src/infrastructure/raw_key_event_extensions.dart
matthew-carroll: My initial thought is that if Flutter moved these inspections to `HardwareKeyboard` because they're really a query of the whole keyboard rather than a query of a key press, then I think it's preferable that we use `HardwareKeyboard` in the associated locations. That might change the code a bit more, but it puts us more inline with Flutter's view of the world, which is probably better for the future. Thoughts?

super_editor/test/super_editor/text_entry/tagging/action_tags_test.dart
matthew-carroll: Let me make sure I understand - you said that when keeping the existing "key up" behavior a test fails? If so, something weird is going on there, because we don't have any failing tests other than goldens when using `RawKeyEvent` with "key up". Why might that test pass with `RawKeyEvent` on "up" but fail with `KeyEvent` on "up"?

angelosilvestre (commented):
@matthew-carroll I don't have any other comments besides what you've already pointed out.

super_editor/example/lib/demos/in_the_lab/popover_list.dart
angelosilvestre: @matthew-carroll  If I understood it correctly, holding a key down will generate exactly one `KeyDownEvent`, possibly many  `KeyRepeatEvent`s and exactly one `KeyUpEvent`.

If that's the case, we probably need to change this code to handle `KeyDownEvent` and `KeyRepeatEvent` instead of the `KeyUpEvent`.

matthew-carroll: @angelosilvestre Change it how? Isn't the `KeyRepeatEvent` already running our handlers? We're only filtering out the "down" events, right?

angelosilvestre: In this specific handler, we are returning early for `KeyDownEvent` and `KeyRepeatEvent`. I imagine we should return early for `KeyUpEvent` and then handle the other events.

matthew-carroll: But we weren't previously returning early for `KeyRepeatEvent`, hence my original comment. It looks to me like we were previously responding to every `KeyRepeatEvent` and now with this PR change we're going to ignore every `KeyRepeatEvent`. Why not continue ignoring "down" events and then react to "up" and "repeat" events as we were before?

angelosilvestre: I didn't notice we were already using the `KeyDownEvent` instead of the `Raw` event.

> Why not continue ignoring "down" events and then react to "up" and "repeat" events as we were before?

I guess it makes to go that way.

gspencergoog: I only added the `KeyRepeatEvent` because the comment below said:
```
// Only handle up events, so we don't run our behavior twice
// for the same key press.
```

And if that's indeed true, then you don't want repeat events either: what @angelosilvestre said was correct: you get one down event, possibly many repeat events, and one up event.  If you don't reject repeat events, then you would run the behavior multiple times for the same keypress, as long as they hold the keypress down long enough that it repeats.

Anyhow, this doesn't have to do with the Raw conversion at all, so I'm happy to revert this line and let you decide what to do here in another PR.

super_editor/lib/src/infrastructure/raw_key_event_extensions.dart
gspencergoog: Those are good reasons to do that, but the reason I didn't end up doing it that way is that it was useful to have the extension methods to allow for injecting a fake into the process so that I didn't need to convert the tests to widget tests (Only in widget tests can we modify the test bindings to return the fake).

However, my method has some drawbacks:

1) if somewhere in your code you call `HardwareKeyboard.instance.isControlPressed` instead of the extension method on the event, then it will bypass the fake in a test. and
2) it requires an API on the extension that the test can call to set the fake, which is a public API that could be misused outside of a test.

The reason I didn't go the other way was because you wanted it to be a dart test, not a widget test.

super_editor/test/super_editor/text_entry/tagging/action_tags_test.dart
gspencergoog: It appears that the key down event is causing `enterToInsertBlockNewline` to be executed, which splits the paragraph and resets the caret, removing the position tag so that once the enter up is processed, there's no position tag.

In the previous code, it doesn't try to execute that behavior for some reason.  I'll keep looking for what the difference is.

matthew-carroll: I notice that `enterToInsertBlockNewline()` only responds to key down events with `if (keyEvent is! RawKeyDownEvent)`. Looking at other handlers, all of them seem to only respond to key down events. Maybe I misremembered and all of these operate on "down" instead of "up". I can't remember why.

I think it's a good idea to figure out why it went from working to not working, but after this PR I'll probably need to revisit the "down" vs "up" situation and make sure they're all uniform.

super_editor/lib/src/infrastructure/raw_key_event_extensions.dart
matthew-carroll: Ok. If switching them to widget tests allows us to fake the hardware keyboard state, and that's needed for the tests, then we can switch them to widget tests.

super_editor/example/lib/demos/in_the_lab/popover_list.dart
matthew-carroll: The commented code you referenced was only referring to "down" and "up". The comment was essentially saying "we're ignoring the down event so that we don't run this action on down and up". I don't think I even considered the repeat event when I wrote that. But let's try to keep this PR as close to the original behavior as possible for a clean migration and we can deal with solving existing problems in followup PRs.

super_editor/test/super_editor/text_entry/tagging/action_tags_test.dart
gspencergoog: Okay, I figured out why it was different.  In the [existing test code](https://github.com/gspencergoog/super_editor/blob/ef97038bfa809ebf1c1dfabbe2c025c92929101e/super_editor/test/super_editor/text_entry/tagging/action_tags_test.dart#L683), it is testing against `KeyDownEvent`, *not* `RawKeyDownEvent`, which means that the existing code fires on key down, even if the conditional looks like it shouldn't. This is because the existing test takes `RawKeyEvent` as an argument, so the `keyEvent` can never be a `KeyDownEvent`.

This is part of why we need to deprecate this: it's confusing to have two separate systems.

In any case, I think the test can now be modified to operate on key down, since although doesn't seem like that was the original intent, it is how the original worked. Sound good?

super_editor/example/lib/demos/in_the_lab/popover_list.dart
gspencergoog: Okay, I reverted my addition of `KeyRepeatEvent`, and you can decide at a later date what you want to do.

super_editor/lib/src/infrastructure/raw_key_event_extensions.dart
gspencergoog: I was able to do this without switching to widget tests (I just had to initialize a fake binding in `main`).

I also switched all the instances in the code to `HarwareKeyboard.instance.isMetaPressed` (etc) and eliminated the extension methods.

gspencergoog:
The tests didn't pass because of  `BlinkTimingMode` being missing (no idea why), and because the symbols in the Flutter master branch on `HardwareKeyboard` that I'm using are missing.  They are available in the master branch, so is there a ref or version or something that needs to be updated so that the tests run against that?

matthew-carroll:
> The tests didn't pass because of some golden test failures, and BlinkTimingMode being missing (no idea why), and because the symbols in the Flutter master branch on HardwareKeyboard that I'm using are missing. They are available in the master branch, so is there a ref or version or something that needs to be updated so that the tests run against that?

@gspencergoog I see all tests as green except the goldens. Which failures are you referring to?

gspencergoog:
> I see all tests as green except the goldens. Which failures are you referring to?

The suspicious-kepler ones, for example: https://app.netlify.com/sites/suspicious-kepler-28398e/deploys/6564fc545c05840008faeb79

gspencergoog:
Looks like all of the golden tests that are failing here are also failing on main against stable (with seemingly identical differences), so we can probably ignore those.

matthew-carroll:
@gspencergoog I think we're ignoring the marketing website CI checks. We should probably fix that at some point, but I don't think we need to worry about it here. All relevant tests appear to be passing. I'll do another PR review when I get a minute.

super_editor/lib/src/default_editor/document_hardware_keyboard/document_keyboard_actions.dart
```diff
@@ -69,9 +70,9 @@ ExecutionInstruction scrollOnPageDownKeyPress({
 /// CMD + HOME on Mac, or CTRL + HOME on all other platforms.
 ExecutionInstruction scrollOnCtrlOrCmdAndHomeKeyPress({
   required SuperEditorContext editContext,
-  required RawKeyEvent keyEvent,
+  required KeyEvent keyEvent,
 }) {
-  if (keyEvent is! RawKeyDownEvent) {
+  if (keyEvent is! KeyDownEvent && keyEvent is! KeyRepeatEvent) {
```
matthew-carroll, line 75: Looks like we still have some `KeyRepeatEvent` checks in this PR. Can you remove all of those, unless they already exist somewhere?

super_editor/lib/src/default_editor/text.dart
```diff
@@ -1490,15 +1490,15 @@ class InsertAttributedTextCommand implements EditCommand {
...
-  required RawKeyEvent keyEvent,
+  required KeyEvent keyEvent,
 }) {
-  if (keyEvent is! RawKeyDownEvent) {
+  if (keyEvent is! KeyDownEvent && keyEvent is! KeyRepeatEvent) {
     return ExecutionInstruction.continueExecution;
   }
 
-  // Do nothing if CMD or CTRL are pressed because this signifies an attempted
+  // Do nothing if META or CTRL are pressed because this signifies an attempted
```
matthew-carroll, line 1499: In the places where we say "CMD", we do that on purpose, because most people don't know what "meta" means. Unless there's more to that situation than I'm aware of, let's change any such alterations back to what they were.

super_editor/lib/src/infrastructure/focus.dart
```diff
@@ -197,7 +201,7 @@ class FocusWithCustomParent extends StatefulWidget {
   /// A handler for keys that are pressed when this object or one of its
   /// children has focus.
   ///
-  /// This is a legacy API based on [RawKeyEvent] and will be deprecated in the
+  /// This is a legacy API based on [KeyEvent] and will be deprecated in the
```
matthew-carroll, line 204: Should this reference to `RawKeyEvent` actually have been replaced? If this happened because you ran a generic "find and replace" in the source code then please be sure to check every place you changed.

super_editor/lib/src/infrastructure/keyboard.dart
```diff
@@ -2,6 +2,10 @@
 
 import 'package:flutter/foundation.dart';
 import 'package:flutter/widgets.dart';
+import 'package:flutter/services.dart';
+
+import 'package:super_editor/src/infrastructure/key_event_extensions.dart';
```
matthew-carroll, line 7: Do we need to import `key_event_extensions` here? If that's a superfluous import then there's probably a lint check warning that went unnoticed. Please check for any new lint warnings introduced in this PR.

super_editor/test/super_editor/text_entry/tagging/action_tags_test.dart
matthew-carroll: Sure

super_editor/test/super_editor/text_entry/text_test.dart
```diff
@@ -51,52 +53,49 @@ void main() {
 
     group('TextComposable text entry', () {
       test('it does nothing when meta is pressed', () {
+        // Make sure we're using our fake binding so we can override keyboard
+        // modifier state.
+        assert(ServicesBinding.instance == binding);
```
matthew-carroll, line 58: I'm confused by this assertion here with the binding being set up above. Is the assert here because this test just so happens to be the first test that runs, and all the other tests assume this test runs first? If so, that seems very fragile, and unlikely that other developers will comprehend that implicit order requirement.

If, on the other hand, this binding is only needed in this test, then let's initialize it here, use it, and reset it, all within the one test that cares.

super_editor/test/super_editor/text_entry/text_test.dart
```diff
@@ -2,10 +2,12 @@ import 'package:flutter/services.dart';
...
 import 'package:super_editor/super_editor.dart';
 
-import '../../test_tools_user_input.dart';
 import '../supereditor_test_tools.dart';
 
-void main() {
+Future<void> main() async {
+  // Replace the default test binding with our fake so we can override the
+  // keyboard modifier state.
+  final FakeServicesBinding binding = FakeServicesBinding();
```
matthew-carroll, line 10: Please add a blank line to visually separate setup from test definitions.

super_editor/test/super_editor/text_entry/text_test.dart
```diff
@@ -51,52 +53,49 @@ void main() {
...
-              isModifierKeyPressed: false,
-            ),
+          keyEvent: const KeyDownEvent(
+            logicalKey: LogicalKeyboardKey.keyA,
+            physicalKey: PhysicalKeyboardKey.keyA,
+            timeStamp: Duration.zero,
           ),
         );
-
+        binding.fakeKeyboard.isMetaPressed = false;
```
matthew-carroll, line 85: Please add a blank line to visually separate the behavior under test from the expectations.

super_editor/test/super_editor/text_entry/text_test.dart
```diff
@@ -51,52 +53,49 @@ void main() {
...
         // The handler should pass on handling the key.
         expect(result, ExecutionInstruction.continueExecution);
       });
-
-      test('it does nothing when nothing is selected', () {
+      testWidgets('it does nothing when nothing is selected', (WidgetTester tester) async {
```
matthew-carroll, line 89: Please include a blank line between every test definition.

super_editor/test/super_editor/text_entry/text_test.dart
```diff
@@ -173,19 +171,18 @@ void main() {
...
+            timeStamp: Duration.zero,
           ),
         );
 
         // The handler should pass on handling the key.
         expect(result, ExecutionInstruction.continueExecution);
       });
 
-      test('it does nothing when the key doesn\'t have a character', () {
+      testWidgets('it does nothing when the key doesn\'t have a character', (WidgetTester tester) async {
```
matthew-carroll, line 185: Why are some of these widget tests and others are Dart tests?

super_editor/lib/src/infrastructure/focus.dart
matthew-carroll: It's probably fine to remove that. I doubt that many of our clients were depending on it.

super_editor/lib/src/default_editor/document_hardware_keyboard/document_keyboard_actions.dart
matthew-carroll: Oh, ok. I misunderstood. I thought all the events were the same but with different names. Given that information, the changes make sense.

matthew-carroll (approved):
LGTM

matthew-carroll:
@angelosilvestre if you'd like to try logging Flutter `master` commits, here's one to track.

PR #1594, opened 2023-11-11
Adds paragraph alignment request and command (Resolves #1587)

### LINKED ISSUE #1587: Changing text alignment doesn't reflect the alignment change on the text node within the document immediately.
**The Issue:**

Changing text alignment of a paragraph node doesn't reflect the alignment change immediately in the document. This change is reflected only after I make any other edits to the document.

To reproduce: 

1. Select any text within `SuperEditor`.
2. Change text alignment through alignment dropdown within toolbar.
3. Make any arbitrary edits to the document to see the text alignment change.

Demo:

https://github.com/superlistapp/super_editor/assets/65209850/47c44e02-5988-4922-a3e5-17f3c1eeb780

**Expected behaviour:**

Change in text alignment should be reflected immediately in the document.

rutvik110:
Adds paragraph alignment request and command (Resolves #1587)

### **The Issue:** 

Changing text alignment of a paragraph node doesn't reflect the alignment change immediately in the document. This change is reflected only after I make any other edits to the document.  Reason seems to be due to the way we're updating the paragraph alignment. 

Sometime back ago, we introduced a request/command system to make changes to the document. The request/command to change the paragraph alignment was left out back then, and right now we seem to only update the node's metadata when trying to update the alignment of the paragraph through the toolbar alignment dropdown.

Demo:

https://github.com/superlistapp/super_editor/assets/65209850/47c44e02-5988-4922-a3e5-17f3c1eeb780

### **Solution:** 

This PR adds `ChangeParagraphAlignmentRequest` and `ChangeParagraphAlignmentCommand` to change the alignment of the paragraph with the request/command flow in `super_editor`. 

### **Remarks:** 

1. Adds `ChangeParagraphAlignmentRequest` and `ChangeParagraphAlignmentCommand` to handle paragraph alignment change request and execution.
2. Adds alignment request check for `ChangeParagraphAlignmentRequest` within existing `defaultRequestHandlers`.
3. `EditorToolbar` alignment change toggle function updated to make the alignment request with `ChangeParagraphAlignmentRequest` to the `Editor`.
4. Adds `selectionPolicies` to `ExampleEditor` super editor to persist the toolbar overlay when clicked.

### DIFF
--- a/super_editor/example/lib/demos/example_editor/example_editor.dart
+++ b/super_editor/example/lib/demos/example_editor/example_editor.dart
@@ -415,6 +415,10 @@ class _ExampleEditorState extends State<ExampleEditor> {
               keyboardActions: _inputSource == TextInputSource.ime ? defaultImeKeyboardActions : defaultKeyboardActions,
               androidToolbarBuilder: (_) => _buildAndroidFloatingToolbar(),
               overlayController: _overlayController,
+              selectionPolicies: SuperEditorSelectionPolicies(
+                clearSelectionWhenEditorLosesFocus: false,
+                clearSelectionWhenImeConnectionCloses: false,
+              ),
             ),
           ),
         ),
--- a/super_editor/lib/src/default_editor/paragraph.dart
+++ b/super_editor/lib/src/default_editor/paragraph.dart
@@ -1,6 +1,7 @@
 import 'package:attributed_text/attributed_text.dart';
 import 'package:flutter/painting.dart';
 import 'package:flutter/services.dart';
+import 'package:flutter/widgets.dart';
 import 'package:super_editor/src/core/document.dart';
 import 'package:super_editor/src/core/document_composer.dart';
 import 'package:super_editor/src/core/document_layout.dart';
@@ -227,6 +228,53 @@ class ChangeParagraphBlockTypeCommand implements EditCommand {
       DocumentEdit(
         NodeChangeEvent(nodeId),
       ),
+    ]);
+  }
+}
+
+/// [EditRequest] to change the alignment of the [ParagraphNode] to [textAlign].
+class ChangeParagraphAlignmentRequest implements EditRequest {
+  ChangeParagraphAlignmentRequest({
+    required this.nodeId,
+    required this.textAlign,
+  });
+
+  final String nodeId;
+  final String? textAlign;
+
+  @override
+  bool operator ==(Object other) =>
+      identical(this, other) ||
+      other is ChangeParagraphAlignmentRequest &&
+          runtimeType == other.runtimeType &&
+          nodeId == other.nodeId &&
+          textAlign == other.textAlign;
+
+  @override
+  int get hashCode => nodeId.hashCode ^ textAlign.hashCode;
+}
+
+/// Updates alignment of the [ParagraphNode] with the given [textAlign].
+class ChangeParagraphAlignmentCommand implements EditCommand {
+  const ChangeParagraphAlignmentCommand({
+    required this.nodeId,
+    required this.textAlign,
+  });
+
+  final String nodeId;
+  final String? textAlign;
+
+  @override
+  void execute(EditContext context, CommandExecutor executor) {
+    final document = context.find<MutableDocument>(Editor.documentKey);
+
+    final existingNode = document.getNodeById(nodeId)! as ParagraphNode;
+    existingNode.putMetadataValue('textAlign', textAlign);
+
+    executor.logChanges([
+      DocumentEdit(
+        NodeChangeEvent(nodeId),
+      )
     ]);
   }
 }
--- a/super_editor/lib/src/default_editor/paragraph.dart
+++ b/super_editor/lib/src/default_editor/paragraph.dart
@@ -1,7 +1,6 @@
 import 'package:attributed_text/attributed_text.dart';
 import 'package:flutter/painting.dart';
 import 'package:flutter/services.dart';
-import 'package:flutter/widgets.dart';
 import 'package:super_editor/src/core/document.dart';
 import 'package:super_editor/src/core/document_composer.dart';
 import 'package:super_editor/src/core/document_layout.dart';
--- a/super_editor/example/lib/demos/example_editor/_toolbar.dart
+++ b/super_editor/example/lib/demos/example_editor/_toolbar.dart
@@ -438,8 +438,14 @@ class _EditorToolbarState extends State<EditorToolbar> {
         break;
     }
 
-    final selectedNode = widget.document.getNodeById(widget.composer.selection!.extent.nodeId) as ParagraphNode;
-    selectedNode.putMetadataValue('textAlign', newAlignmentValue);
+    widget.editor!.execute([
+      ChangeParagraphAlignmentRequest(
+        nodeId: widget.composer.selection!.extent.nodeId,
+        textAlign: newAlignmentValue,
+      ),
+    ]);
+
+    widget.closeToolbar();
   }
 
   /// Returns the localized name for the given [_TextType], e.g.,
--- a/super_editor/lib/src/default_editor/default_document_editor.dart
+++ b/super_editor/lib/src/default_editor/default_document_editor.dart
@@ -195,6 +195,12 @@ final defaultRequestHandlers = List.unmodifiable(<EditRequestHandler>[
           composer: request.composer,
         )
       : null,
+  (request) => request is ChangeParagraphAlignmentRequest
+      ? ChangeParagraphAlignmentCommand(
+          nodeId: request.nodeId,
+          textAlign: request.textAlign,
+        )
+      : null,
 ]);
 
 final defaultEditorReactions = List.unmodifiable([
--- a/super_editor/lib/src/default_editor/text.dart
+++ b/super_editor/lib/src/default_editor/text.dart
@@ -533,6 +533,23 @@ class TextComponentState extends State<TextComponent> with DocumentComponent imp
   }
 
   @override
+  void didUpdateWidget(covariant TextComponent oldWidget) {
+    super.didUpdateWidget(oldWidget);
+
+    if (oldWidget.textAlign != widget.textAlign) {
+      // Text alignment has changed, force a re-layout.
+      invalidateLayout();
+    }
+  }
+
+  /// Invalidates layout for this widget to force a rebuild on the
+  /// parent widget.
+  void invalidateLayout() {
+    final renderObject = context.findRenderObject();
+    renderObject!.markNeedsLayout();
+  }
+
+  @override
   Offset getOffsetForPosition(dynamic nodePosition) {
     if (nodePosition is! TextPosition) {
       throw Exception('Expected nodePosition of type TextPosition but received: $nodePosition');
--- a/super_text_layout/lib/src/super_text.dart
+++ b/super_text_layout/lib/src/super_text.dart
@@ -392,6 +392,15 @@ class RenderLayoutAwareParagraph extends RenderParagraph {
   bool _needsLayout = true;
 
   @override
+  set textAlign(TextAlign value) {
+    if (super.textAlign == value) {
+      return;
+    }
+    super.textAlign = value;
+    markNeedsLayout();
+  }
+
+  @override
   void markNeedsLayout() {
     super.markNeedsLayout();
     _needsLayout = true;
--- a/super_text_layout/lib/src/super_text.dart
+++ b/super_text_layout/lib/src/super_text.dart
@@ -392,15 +392,6 @@ class RenderLayoutAwareParagraph extends RenderParagraph {
   bool _needsLayout = true;
 
   @override
-  set textAlign(TextAlign value) {
-    if (super.textAlign == value) {
-      return;
-    }
-    super.textAlign = value;
-    markNeedsLayout();
-  }
-
-  @override
   void markNeedsLayout() {
     super.markNeedsLayout();
     _needsLayout = true;


rutvik110:
@angelosilvestre This PR introduces the request/command to change the alignment of the paragraph to fix #1587. I'm currently also looking into incorporating some changes that we'll require to fix #1407 in this PR, as this two issues are closely related with each other.

rutvik110:
@angelosilvestre 
Is there a way for us to force the `CaretDocumentOverlay` to re-evaluate its caret position without having to make a selection change?

Currently, it listens for selection changes to update its caret position in the document. This selection change notification isn’t generated when we change the alignment of an paragraph with an active selection. In this case, only the alignment changes, and not the selection. So the caret position isn’t updated wrt to changed alignment for the paragraph, and is resulting in the #1407  issue.

I’m looking into if we can introduce something like following within the `composer`, that will allow us to force any listeners that listen to selection changes to re-evaulate their states,

```dart
  void reevauluateCaretPosition() {
    _selectionNotifier.notifyListeners();
  }
```

For e.g. `CaretDocumentOverlay` listens to selection changes with `_selectionNotifier` from `composer` . So whenever we change the alignment, we could schedule this callback after the current frame from within `ChangeParagraphAlignmentCommand` when the changed alignment comes into effect within the document, and force it to update the caret position,

```dart
WidgetsBinding.instance.addPostFrameCallback((timeStamp) {
      composer.reevauluateCaretPosition();
});
```

Other than this, I’ven’t found any other way to force `CaretDocumentOverlay`  to re-evaluate itself to update the caret position whenever we change alignment on the paragraph. Let me know if you've any other ideas on this.

angelosilvestre:
@rutvik110 Can `CaretDocumentOverlay` listen for document changes?

rutvik110:
It doesn't atm. It only listens for `composer` selection changes.

angelosilvestre:
So, can we change it to listen for document changes?  BTW, with the use of `ContentLayers` I would expect that the caret layer rebuilds itself when the document layout changes...

rutvik110:
It does rebuilds when the alg

> BTW, with the use of `ContentLayers` I would expect that the caret layer rebuilds itself when the document layout changes...

It does rebuilds itself, but it doesn't have the latest position of the paragraph within the document during the current frame when the alignment is changed. Due to this, the caret position doesn't update itself with the changed position of the paragraph component within the document. 

This information seems to only be available once the alignment changes have taken effect, i.e in the next frame.

angelosilvestre:
So can we make `CaretDocumentOverlay` listen for document changes and then recompute what's necessary in the next frame?

rutvik110:
> So, can we change it to listen for document changes?

Does it need to listen to document changes? As far as I've seen, we explicitly update the composer selection for cases where document layout is changed. For e.g. `SplitParagraphCommand` updates composer selection when splitting paragraphs.

rutvik110:
> So can we make `CaretDocumentOverlay` listen for document changes and then recompute what's necessary in the next frame?

I'll check.

angelosilvestre:
> Does it need to listen to document changes? As far as I've seen, we explicitly update the composer selection for cases where document layout is changed. For e.g. SplitParagraphCommand updates composer selection when splitting paragraphs.

When we split a paragraph, the selection do change. Here the selection itself doesn't change, what changes is the offset of the selected position on screen.

rutvik110:
> So can we make `CaretDocumentOverlay` listen for document changes and then recompute what's necessary in the next frame?

This doesn't help. It's likely due to the fact that any document changes are reported earlier before composer changes in which case we still don't have latest position of the paragraph within the document as the layout update hasn't taken place yet during that time.

rutvik110:
So I think a final solution for updating caret position on alignment changes is basically listening for document changes within `CaretDocumentOverlay`, and scheduling a post frame callback to update itself like follow, 

```dart
    widget.document.addListener((value) {
        WidgetsBinding.instance.addPostFrameCallback((timeStamp) {
          setState(() {});
        });
    });
```

@angelosilvestre Any final thoughts here? If it's good to go, then I'll get onto writing tests to lock this behaviour and push this PR forward?

angelosilvestre:
>     widget.document.addListener((value) {
        WidgetsBinding.instance.addPostFrameCallback((timeStamp) {
          setState(() {});
        });
    });

I think we have an extension method somewhere that does that. 

You can write tests, then I can do a proper review

super_editor/example/lib/demos/example_editor/example_editor.dart
```diff
@@ -409,6 +409,10 @@ class _ExampleEditorState extends State<ExampleEditor> {
               keyboardActions: _inputSource == TextInputSource.ime ? defaultImeKeyboardActions : defaultKeyboardActions,
               androidToolbarBuilder: (_) => _buildAndroidFloatingToolbar(),
               overlayController: _overlayController,
+              selectionPolicies: SuperEditorSelectionPolicies(
```
matthew-carroll, line 412: Why was this added?

super_editor/lib/src/default_editor/paragraph.dart
```diff
@@ -1,6 +1,7 @@
 import 'package:attributed_text/attributed_text.dart';
 import 'package:flutter/painting.dart';
 import 'package:flutter/services.dart';
+import 'package:flutter/widgets.dart';
```
matthew-carroll, line 4: Why do we need to import `widgets`? And if we need it, do we still need `painting` and `services`?

super_editor/lib/src/default_editor/paragraph.dart
```diff
@@ -231,6 +232,53 @@ class ChangeParagraphBlockTypeCommand implements EditCommand {
...
+  @override
+  void execute(EditContext context, CommandExecutor executor) {
+    final document = context.find<MutableDocument>(Editor.documentKey);
+
+    final existingNode = document.getNodeById(nodeId)! as ParagraphNode;
+    existingNode.putMetadataValue('textAlign', textAlign);
+
+    executor.logChanges([
+      DocumentEdit(
+        NodeChangeEvent(nodeId),
```
matthew-carroll, line 276: Do we not have any way to capture *what* changed?

matthew-carroll:
I read through the comments in this PR and it looks like the change log must be out of date because there are details mentioned in the comments that don't seem to be in the change log.

That said, we really don't want to be in the business of scheduling extra frames any more. All of `ContentLayers` is about moving away from that wherever possible. Rather than place a bandaid on this situation of paragraph alignment, we should step back and ask if we're missing infrastructure for a category of changes that can occur.

Paragraph alignment shows us that we can have changes which don't alter the primary content of a node, nor does it change the selection, and yet it impacts the layout of the node's component.

Can either of you think of any other situation where something like this happens?

Are we sure there isn't an existing execution path that's meant to handle this? For example, doesn't every document layout component rebuild itself whenever its node changes? If so, why isn't that rebuild causing enough of a change to get the caret in the right place?

angelosilvestre:
> Can either of you think of any other situation where something like this happens?

In theory, anything the relies on the node's metadata could run into this. But it isn't clear why this isn't the case with `blockType`. I know that changing the `blockType` might cause a totally different component to be rendered, but  changing from a regular paragraph to a header doesn't do that and it re-renders everything correctly. That suggests we might be missing something...

super_editor/lib/src/default_editor/paragraph.dart
rutvik110: That's an unwanted change. I've reverted it.

super_editor/lib/src/default_editor/paragraph.dart
rutvik110: I'll check.

rutvik110:
> Are we sure there isn't an existing execution path that's meant to handle this? For example, doesn't every document layout component rebuild itself whenever its node changes? If so, why isn't that rebuild causing enough of a change to get the caret in the right place?

Currently, the alignment change on node isn't triggering a rebuild on `ContentLayers` which is responsible for showing any document overlays. It works when changing the block type, but I'm not sure yet why the build isn't triggered for alignment change on it. This is the main reason behind the caret position not updating itself when alignment changes.  

I'll check once again into this, and see what we might be missing here.

super_editor/example/lib/demos/example_editor/example_editor.dart
rutvik110: Without this the toolbar disappears when tapping anything within it. Related to #689 

Won't need this once #1538 lands.

matthew-carroll:
> Currently, the alignment change on node isn't triggering a rebuild on ContentLayers which is responsible for showing any document overlays. It works when changing the block type, but I'm not sure yet why the build isn't triggered for alignment change on it. This is the main reason behind the caret position not updating itself when alignment changes.
> 
> I'll check once again into this, and see what we might be missing here.

Please do. This seems like a strong indication that we've completely missed the root problem.

rutvik110:
@matthew-carroll So @angelosilvestre and I had a catch up on this issue this week, and were able to diagnose the root cause of it. 

`RenderParagraph` that sits at the root of the text rendering in Flutter doesn't seem to schedule a re-layout when the text alignment related to it changes. Due to this parent widgets weren't updating themselves when alignment would change as it was happening in our case. 

The solution was to override the `textAlign` setter within `RenderLayoutAwareParagraph` to schedule a re-layout on alignment change that would force the parent widgets to rebuild and update the caret accordingly in our case.

```dart
 @override
  set textAlign(TextAlign value) {
    if (super.textAlign == value) {
      return;
    }
    super.textAlign = value;
    markNeedsLayout();
  }
```

matthew-carroll:
@rutvik110 I spoke with @angelosilvestre on a call about this. I see the solution you two proposed, and I can see why it technically solves the problem. My concern is that we're addressing a `SuperEditor` and `SuperReader` problem at the lower level of `SuperText`. That's probably not what we want to do.

Can we find an equally specific location within `super_editor` to achieve a similar effect? For example, we have `TextComponent`s that are displayed within the `SingleColumnDocumentLayout` - can a `TextComponent` recognize when this happens and then invalidate itself? Please see if you can find a reasonable solution in that area of the code.

rutvik110:
@matthew-carroll The solution that we discussed earlier, we could move that update logic to `TextComponent` level. Here's one of the ways we could do this. On alignment change, we can force a relayout through the `TextComponent`'s render object. 

```dart
  @override
  void didUpdateWidget(covariant TextComponent oldWidget) {
    super.didUpdateWidget(oldWidget);

    if (oldWidget.textAlign != widget.textAlign) {
      // Text alignment has changed, force a re-layout.
      invalidateLayout();
    }
  }

  /// Invalidates layout for this widget to force a rebuild on the
  /// parent widget.
  void invalidateLayout() {
    final renderObject = context.findRenderObject();
    renderObject!.markNeedsLayout();
  }
```

Might not be the prettiest solution but beside this, I'ven't found another way through which we can force the document to rebuild itself and let the overlay layers adjust to the new alignment change on text. Wdys?

matthew-carroll:
> we can force a relayout through the TextComponent's render object

This still sounds way too low level. Is there really not any reasonable place to recognize this change within our overall editor system? I'm talking about commands, the document layout, the component state objects, etc. I'm very skeptical that this solution requires such deep intervention.

rutvik110:
> > we can force a relayout through the TextComponent's render object
> 
> This still sounds way too low level. Is there really not any reasonable place to recognize this change within our overall editor system? I'm talking about commands, the document layout, the component state objects, etc. I'm very skeptical that this solution requires such deep intervention.

I'ven't checked into commands or the document layout system and whether or not it exposes any apis to achieve similar behaviour. Will check into that area. 
Also wdym by `component state objects`? Not sure if you're referring to any apis within the `TextComponent` here by that or anything else.

matthew-carroll:
> I'ven't checked into commands or the document layout system

In general, please always look through the existing behaviors and structure to find the most appropriate place for a fix or new feature. We don't want to stuff changes wherever we can, we want to make the appropriate changes given our architecture, so that we're able to make more changes in the future.

> Also wdym by component state objects?

Isn't a `TextComponent` a `StatefulWidget`? If so, it has a `State` object...

rutvik110:
> Isn't a `TextComponent` a `StatefulWidget`? If so, it has a `State` object...

yup, that was an irrelevant question from my end.

rutvik110:
@matthew-carroll I'm experimenting with the existing apis to see if we can use them to achieve the desired behaviour from [this](https://github.com/superlistapp/super_editor/pull/1594#issuecomment-1865887284) earlier solution within `TextComponent`. But, neither the document or `TextComponent` itself or any other api's within `super_editor` doesn't seem to have a way to invalidate the text component to force a re-layout on the document itself. 

And beside addressing this solution at either the `TextComponent` or the `RenderLayoutAwareParagraph` level as discussed [here](https://github.com/superlistapp/super_editor/pull/1594#issuecomment-1846754731), I'ven't found another place from where we can both detect this change and also resolve it without introducing any far-fetched solutions. 

On the other hand, I think we're still trying to address this issue at the wrong place. We initially discussed to introduce the [solution](https://github.com/superlistapp/super_editor/pull/1594#issuecomment-1846754731) at `RenderParagraphLevel`. But, abandoned it coze it didn't seemed the right place for it. I think, given the infrastructure and how things operate in `super_editor` where a text alignment change may affect layout of other parts like overlays/content layers, it only seems right that `RenderParagraphLevel` should invalidate it's layout on alignment change and let the respective parent elements rebuild themself when that happens.

If we still intend to address this at a higher level then atm I'ven't found any existing api's that can resolve this issue other than the solutions I've proposed earlier. We may need to introduce a new set of api's that will allow the components like `TextComponent` to invalidate itself and force a document rebuild along with any other parts like overlays. Not sure how this will look like.

matthew-carroll:
> But, neither the document or TextComponent itself or any other api's within super_editor doesn't seem to have a way to invalidate the text component to force a re-layout on the document itself.

Why do you need to relayout the entire document? Isn't this issue about paragraph alignment?

> it only seems right that RenderParagraphLevel should invalidate it's layout on alignment change and let the respective parent elements rebuild themself when that happens.

The point wasn't that `RenderParagraph` shouldn't invalidate itself. In fact, there's no way to avoid invalidating that render object's layout. The point was that it's inappropriate to have a high level `SuperEditor` command try to somehow reach down an mess with a render object. That's using one of the highest level artifacts to do one of the lowest level things. Think about how that approach would scale. Think about how many commands would be reaching into the depths of render objects all over the place. That's the problem.

> If we still intend to address this at a higher level then atm I'ven't found any existing api's that can resolve this issue other than the solutions I've proposed earlier. 

Ok, I'll take over this ticket.

PR #1596, opened 2023-11-13
Added SuperTextField entrypoint, includes Mac OS selectors and Actions example

matthew-carroll:
By default, `SuperTextField` on Mac now blocks all key events and sends them directly to the OS, in case they need to be turned into selectors. Apps can alter this behavior by placing higher priority key handlers above that one, or completely replacing the key handlers in the text field.

By forcibly bypassing all Flutter focus nodes, key events like ESC will go directly to the OS, generate a selector, like "cancel operation", which will be sent to the IME and received by the text field.

`SuperTextField` has a map of default selector behaviors. **If the defaults should be changed, please let me know.**

Apps can customize selector actions by providing their own entries in the selector action map. They can add to the default actions, or completely replace the default actions.

### DIFF
--- a/super_editor/example/lib/main_super_text_field.dart
+++ b/super_editor/example/lib/main_super_text_field.dart
@@ -1,9 +1,8 @@
 import 'package:flutter/material.dart';
-import 'package:flutter/services.dart';
-import 'package:flutter/src/foundation/diagnostics.dart';
 import 'package:super_editor/super_editor.dart';
 import 'package:super_editor/super_text_field.dart';
 
+/// An app that demos [SuperTextField].
 void main() {
   runApp(
     MaterialApp(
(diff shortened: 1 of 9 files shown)


super_editor/example/lib/main_super_text_field.dart
```diff
@@ -0,0 +1,196 @@
...
+
+  @override
+  State<SingleLineTextField> createState() => _SingleLineTextFieldState();
+}
+
+class _SingleLineTextFieldState extends State<SingleLineTextField> {
+  final _focusNode = FocusNode();
+  late final _textController = ImeAttributedTextEditingController(
+    controller: AttributedTextEditingController(),
+  );
```
angelosilvestre, line 71: It seems we don't need `late` here.

super_editor/example/lib/main_super_text_field.dart
```diff
@@ -0,0 +1,196 @@
...
+
+  @override
+  State<MultiLineTextField> createState() => _MultiLineTextFieldState();
+}
+
+class _MultiLineTextFieldState extends State<MultiLineTextField> {
+  final _focusNode = FocusNode();
+  late final _textController = ImeAttributedTextEditingController(
+    controller: AttributedTextEditingController(),
+  );
```
angelosilvestre, line 140: It seems we don't need `late` here.

super_editor/example/lib/main_super_text_field.dart
```diff
@@ -0,0 +1,196 @@
...
+            "Type some text...",
+            style: TextStyle(color: Colors.grey),
+          ),
+          minLines: 5,
+          maxLines: 5,
+          inputSource: TextInputSource.ime,
+          imeConfiguration: TextInputConfiguration(keyboardAppearance: Brightness.light),
+          selectorHandlers: {
+            ...defaultTextFieldSelectorHandlers,
+          },
```
angelosilvestre, line 185: Do we need to add the custom selector handlers here?

angelosilvestre (approved):
LGTM with some minor comments.

PR #1604, opened 2023-11-18
[SuperTextField][mobile] Migrate popover toolbars to OverlayPortal (Resolves #1602)

### LINKED ISSUE #1602: [SuperTextField] - Toolbar should respect app theme brightness
It was reported by Superlist that the `SuperTextField` popover toolbars don't respect the app theme, but the `SuperEditor` toolbars do. My guess is that we have yet to migrate `SuperTextField` to use `OverlayPortal`.

Investigate this difference and make changes to `SuperTextField` so that the popover toolbars respect the app theme in the same way as `SuperEditor`.

angelosilvestre:
[SuperTextField][mobile] Migrate popover toolbar to OverlayPortal (Resolves #1602)

Currently, the mobile toolbars are displayed in the app `Overlay`. Because of that, it doesn't have access to  `InheritedWidgets` placed below `MaterialApp`. 

As a result, if an app adds a `Theme` widget above `SuperTextField`, the `SuperTextField` uses this `Theme`, but the popover toolbar uses the `MaterialApp`'s theme.

This PR changes the popover toolbars to be displayed in an `OverlayPortal` instead of an `Overlay`. By doing so, the popovers "see" the same `InheritedWidgets` that the `SuperTextField` sees.

In the tests, I had to use the platform textfields directly, as `SuperTextField` doesn't expose the `popoverToolbarBuilder` in its public API.

I also noticed that the iOS textfield is always using the `_defaultPopoverToolbarBuilder` instead of the given `popoverToolbarBuilder`. This PR fixes that.

### DIFF
--- a/super_editor/lib/src/super_textfield/android/android_textfield.dart
+++ b/super_editor/lib/src/super_textfield/android/android_textfield.dart
@@ -491,57 +491,61 @@ class SuperAndroidTextFieldState extends State<SuperAndroidTextField>
     return OverlayPortal(
       controller: _popoverController,
       overlayChildBuilder: _buildPopoverToolbar,
-      child: TapRegion(
-        groupId: widget.tapRegionGroupId,
-        child: NonReparentingFocus(
-          key: _textFieldKey,
-          focusNode: _focusNode,
-          onKey: _onKeyPressed,
-          child: CompositedTransformTarget(
-            link: _textFieldLayerLink,
-            child: AndroidTextFieldTouchInteractor(
-              focusNode: _focusNode,
-              textKey: _textContentKey,
-              textFieldLayerLink: _textFieldLayerLink,
-              textController: _textEditingController,
-              editingOverlayController: _editingOverlayController,
-              textScrollController: _textScrollController,
-              isMultiline: _isMultiline,
-              handleColor: widget.handlesColor,
-              showDebugPaint: widget.showDebugPaint,
-              child: TextScrollView(
-                key: _scrollKey,
-                textScrollController: _textScrollController,
-                textKey: _textContentKey,
-                textEditingController: _textEditingController,
-                textAlign: widget.textAlign,
-                minLines: widget.minLines,
-                maxLines: widget.maxLines,
-                lineHeight: widget.lineHeight,
-                perLineAutoScrollDuration: const Duration(milliseconds: 100),
-                showDebugPaint: widget.showDebugPaint,
-                padding: widget.padding,
-                child: ListenableBuilder(
-                  listenable: _textEditingController,
-                  builder: (context, _) {
-                    final isTextEmpty = _textEditingController.text.text.isEmpty;
-                    final showHint = widget.hintBuilder != null &&
-                        ((isTextEmpty && widget.hintBehavior == HintBehavior.displayHintUntilTextEntered) ||
-                            (isTextEmpty &&
-                                !_focusNode.hasFocus &&
-                                widget.hintBehavior == HintBehavior.displayHintUntilFocus));
+      child: _buildTextField(),
+    );
+  }
 
-                    return CompositedTransformTarget(
-                      link: _textContentLayerLink,
-                      child: Stack(
-                        children: [
-                          if (showHint) widget.hintBuilder!(context),
-                          _buildSelectableText(),
-                        ],
-                      ),
-                    );
-                  },
-                ),
+  Widget _buildTextField() {
+    return TapRegion(
+      groupId: widget.tapRegionGroupId,
+      child: NonReparentingFocus(
+        key: _textFieldKey,
+        focusNode: _focusNode,
+        onKey: _onKeyPressed,
+        child: CompositedTransformTarget(
+          link: _textFieldLayerLink,
+          child: AndroidTextFieldTouchInteractor(
+            focusNode: _focusNode,
+            textKey: _textContentKey,
+            textFieldLayerLink: _textFieldLayerLink,
+            textController: _textEditingController,
+            editingOverlayController: _editingOverlayController,
+            textScrollController: _textScrollController,
+            isMultiline: _isMultiline,
+            handleColor: widget.handlesColor,
+            showDebugPaint: widget.showDebugPaint,
+            child: TextScrollView(
+              key: _scrollKey,
+              textScrollController: _textScrollController,
+              textKey: _textContentKey,
+              textEditingController: _textEditingController,
+              textAlign: widget.textAlign,
+              minLines: widget.minLines,
+              maxLines: widget.maxLines,
+              lineHeight: widget.lineHeight,
+              perLineAutoScrollDuration: const Duration(milliseconds: 100),
+              showDebugPaint: widget.showDebugPaint,
+              padding: widget.padding,
+              child: ListenableBuilder(
+                listenable: _textEditingController,
+                builder: (context, _) {
+                  final isTextEmpty = _textEditingController.text.text.isEmpty;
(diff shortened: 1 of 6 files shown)


angelosilvestre:
@matthew-carroll Tests are failing due to a compilation error:

```console
lib/src/default_editor/document_hardware_keyboard/document_physical_keyboard.dart:125:23: Error: The argument type 'RawKeyEvent' can't be assigned to the parameter type 'KeyEvent'.
 - 'RawKeyEvent' is from 'package:flutter/src/services/raw_keyboard.dart' ('/opt/hostedtoolcache/flutter/master-any-x64/packages/flutter/lib/src/services/raw_keyboard.dart').
 - 'KeyEvent' is from 'package:flutter/src/services/hardware_keyboard.dart' ('/opt/hostedtoolcache/flutter/master-any-x64/packages/flutter/lib/src/services/hardware_keyboard.dart').
      if (key.accepts(keyEvent, RawKeyboard.instance)) {
```

It seems to be related to this PR: https://github.com/superlistapp/super_editor/pull/1593

super_editor/lib/src/super_textfield/android/android_textfield.dart
```diff
@@ -168,10 +169,12 @@ class SuperAndroidTextFieldState extends State<SuperAndroidTextField>
...
-  // OverlayEntry that displays the toolbar and magnifier, and
-  // positions the invisible touch targets for base/extent
-  // dragging.
-  OverlayEntry? _controlsOverlayEntry;
+  /// Opens/closes the popover that displays the toolbar and magnifier, and
+  // positions the invisible touch targets for base/extent dragging.
+  final _popoverController = OverlayPortalController();
+
+  /// Notifies the popover toolbar to rebuild itself.
+  final _popoverRebuildSignal = SignalNotifier();
```
matthew-carroll, line 177: Is this rebuild signal here because we're scheduling extra frames to position the toolbar? If so, can we use a `Leader` and `Follower` instead?

super_editor/lib/src/super_textfield/android/android_textfield.dart
```diff
@@ -497,56 +488,60 @@ class SuperAndroidTextFieldState extends State<SuperAndroidTextField>
...
-            textController: _textEditingController,
-            editingOverlayController: _editingOverlayController,
-            textScrollController: _textScrollController,
-            isMultiline: _isMultiline,
-            handleColor: widget.handlesColor,
-            showDebugPaint: widget.showDebugPaint,
-            child: TextScrollView(
-              key: _scrollKey,
-              textScrollController: _textScrollController,
+    return OverlayPortal(
```
matthew-carroll, line 491: Now that we're bringing an overlay portal into this, let's breakdown this build method so it's clear which pieces are what.

super_editor/test/super_textfield/super_textfield_theme_test.dart
```diff
@@ -0,0 +1,130 @@
...
+import 'package:super_editor/src/super_textfield/infrastructure/attributed_text_editing_controller.dart';
+import 'package:super_editor/src/super_textfield/input_method_engine/_ime_text_editing_controller.dart';
+import 'package:super_editor/src/super_textfield/ios/ios_textfield.dart';
+import 'package:super_text_layout/super_text_layout.dart';
+
+import 'super_textfield_robot.dart';
+
+void main() {
+  group('SuperTextField', () {
+    testWidgetsOnIos('applies app theme to the popover toolbar (on iOS)', (tester) async {
```
matthew-carroll, line 16: You don't need to specify "(on iOS)" when using `testWidgetsOnIos()`

matthew-carroll:
This PR mentions that `SuperTextField` doesn't support any configuration of the overlay controls. We should probably provide that control. The question is, how should we do that?

In `SuperEditor`, I'm in the middle of moving overlay controls from widget properties over to ancestor composition with "scope" widgets. The rationale for this is that we were accumulating a lot of configurations: control color, handle builder, magnifier builder, toolbar builder, objects to show/hide each of those.

Should we repeat the approach from `SuperEditor` for `SuperTextField` and introduce scope widgets? Or should we introduce widget properties? Either way, we should probably do that in a separate PR, but we can decide the best approach while we're here.

super_editor/test/super_textfield/super_textfield_theme_test.dart
angelosilvestre: `testWidgetsOnIos` itself doesn't add the "(on iOS)" suffix. 

Maybe it's time to change that?

super_editor/lib/src/super_textfield/android/android_textfield.dart
angelosilvestre: Yes, this is used to position the toolbar. Currently, we are doing `_controlsOverlayEntry?.markNeedsBuild();` to do this.

We can use `Leader` and `Follower`, but it will require big change. I'm not sure we should do that on this PR.

super_editor/lib/src/super_textfield/android/android_textfield.dart
angelosilvestre: Updated.

angelosilvestre:
> Should we repeat the approach from SuperEditor for SuperTextField and introduce scope widgets? Or should we introduce widget properties? Either way, we should probably do that in a separate PR, but we can decide the best approach while we're here.

@matthew-carroll If we want to be consistent, we probably want to use the same approach from `SuperEditor`. The downsides is that it's a bit more difficult for app developers than just using widget properties and less discoverable (intelisense won't help them). We'll need good docs and probably mention that in the README.

super_editor/test/super_textfield/super_textfield_theme_test.dart
matthew-carroll: Really? I thought I saw "(on iOS)" appended to all sorts of tests that we run? If for some reason that test runner is the only one that isn't adding its name, then yes, we should add that, but I was almost certain that we already have iOS tests that add the suffix.

super_editor/lib/src/super_textfield/android/android_textfield.dart
matthew-carroll: Ok. We can do it in a follow up. Can you file a ticket for that?

matthew-carroll:
@angelosilvestre can you file an issue about compositional overlay controls to match `SuperEditor`?

matthew-carroll (approved):
LGTM - Pending the "(on iOS)" suffix for the test runner

super_editor/test/super_textfield/super_textfield_theme_test.dart
angelosilvestre: @matthew-carroll  It's the `testWidgetsOnMobile` and `testWidgetsOnAllPlatforms` that appends the suffixes. Can we make that change in a separate PR?

angelosilvestre:
@matthew-carroll Opened https://github.com/Flutter-Bounty-Hunters/testing/pull/1 to add the suffixes directly on the `testWidgetsOnXXXXX` and removed from the editor tests.

matthew-carroll:
@angelosilvestre I merged your PR in the testing repo

angelosilvestre:
@matthew-carroll Already removed the suffixes from here.

 I think https://github.com/superlistapp/super_editor/pull/1593 needs to get merged to fix the tests.

matthew-carroll:
@angelosilvestre did you check all the tests to make sure there aren't any new failures aside from those existing problems?

angelosilvestre:
@matthew-carroll All failing tests are failing due to the compilation error related to the RawKeyEvent -> KeyEvent migration.

PR #1609, opened 2023-11-21
Feat: added support for iOS in isPrimaryShortcutKeyPressed

### LINKED ISSUE #1625: [SuperEditor][SuperReader][iOS] - CMD key isn't reported correctly on iOS
When on iPad, when trying to use keybindings, the `isPrimaryShortcutKeyPressed` getter does not return true whatsoever, making it incompatible with iOS devices.

The fix is quite simple, just add the TargetPlatform.iOS check when meta is pressed inside the getter.

From this:
```dart
(defaultTargetPlatform == TargetPlatform.macOS && isMetaPressed)
```
To this:
```dart
([TargetPlatform.macOS, TargetPlatform.iOS].contains(defaultTargetPlatform) && isMetaPressed)
```

MSimone12:
This PR fixes some minor issues we're facing at CampaignComposer (@jmatth can say more), but, basically when on iPad, it can't detect the Meta key pressed for shortcuts.

This PR Closes #1625 

As explained in the issue, when adding the `TagetPlatform.iOS` to the clause, it will allow the keybindings, which was previously compatible with macOS, be used also on iOS.

### DIFF
--- a/super_editor/lib/src/infrastructure/keyboard.dart
+++ b/super_editor/lib/src/infrastructure/keyboard.dart
@@ -32,7 +32,11 @@ enum ExecutionInstruction {
 
 extension PrimaryShortcutKey on RawKeyEvent {
   bool get isPrimaryShortcutKeyPressed =>
-      (defaultTargetPlatform == TargetPlatform.macOS && isMetaPressed) ||
+      ([
+            TargetPlatform.macOS,
+            TargetPlatform.iOS,
+          ].contains(defaultTargetPlatform) &&
+          isMetaPressed) ||
       (defaultTargetPlatform != TargetPlatform.macOS && isControlPressed);
 }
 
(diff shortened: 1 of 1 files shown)


super_editor/lib/src/infrastructure/keyboard.dart
```diff
@@ -32,7 +32,11 @@ enum ExecutionInstruction {
 
 extension PrimaryShortcutKey on RawKeyEvent {
   bool get isPrimaryShortcutKeyPressed =>
-      (defaultTargetPlatform == TargetPlatform.macOS && isMetaPressed) ||
+      ([
```
matthew-carroll, line 35: Please return formatting throughout this file to what it was before.

matthew-carroll (commented):
I added one comment in-line. Please also add an appropriate test so that this support doesn't regress.

matthew-carroll:
@MSimone12 can you please update this PR based on the comments, and also resolve the merge conflicts?

PR #1277, opened 2023-08-08
Flutter migration: Remove rounding hack from TextPainter

LongCatIsLooong:
A long-standing rounding hack in `TextPainter` will soon be removed (https://github.com/flutter/flutter/issues/31707). As a result the horizontal width of some text reported by `TextPainter` will be slightly smaller as it's no longer rounded up.

PR #1615, opened 2023-11-22
Updates for Flutter 3.16

angelosilvestre:
Updates for Flutter 3.16

As of Flutter 3.16, we have some failing tests on stable branch and also some analyzer errors.

### Test "SuperEditor floating cursor shows grey caret when far from text"

This test is failing because, after Material3, the default primary color changed from blue to purple. We have a test for the floating cursor where we check the cursor color when it's far from the text and when it's close to the text. When it's close, the color defaults to the theme's primary color.

This PR changes the test to compare the value with the `ThemeData`'s primary color.

### Test "control + backspace: deletes a single upstream character":

This test is related to the macOS selectors. In the previous Flutter stable version, `deleteBackwardByDecomposingPreviousCharacter` wasn't generated during tests, so we have a temporary code to generate the selector manually. Now that Flutter already reports this selector, we can safely remove this code.

### Analyzer error due to deprecated `textScale` being deprecated in favor of `TextScaler`.

We already have this change in the main branch since https://github.com/superlistapp/super_editor/pull/1252. So I cherry-picked the commit.

### Test "SuperText RenderSuperTextLayout finds precise characters in first line"

We have this failure:

```console
The following TestFailure was thrown running a test:
  Expected: TextPosition:<TextPosition(offset: 25, affinity: TextAffinity.upstream)>
  Actual: TextPosition:<TextPosition(offset: 24, affinity: TextAffinity.downstream)>
```

The test fails on the following expectation:

```dart
  expect(
      textLayout.getPositionAtOffset(Offset(textBox.size.width / 2, firstLineEstimatedMiddle)),
      const TextPosition(offset: 25, affinity: TextAffinity.upstream),
   );
```

This PR changes the expected position. Is it safe to do that?

### Test "SuperText RenderSuperTextLayout finds nearest TextPosition on the top side"

This is the same cause as the previous test.

### DIFF
--- a/super_editor/lib/src/default_editor/common_editor_operations.dart
+++ b/super_editor/lib/src/default_editor/common_editor_operations.dart
@@ -12,9 +12,9 @@ import 'package:super_editor/src/core/editor.dart';
 import 'package:super_editor/src/default_editor/list_items.dart';
 import 'package:super_editor/src/default_editor/paragraph.dart';
 import 'package:super_editor/src/default_editor/selection_upstream_downstream.dart';
+import 'package:super_editor/src/default_editor/tasks.dart';
 import 'package:super_editor/src/default_editor/text.dart';
 import 'package:super_editor/src/infrastructure/_logging.dart';
-import 'package:super_editor/super_editor.dart';
 
 import 'attributions.dart';
 import 'horizontal_rule.dart';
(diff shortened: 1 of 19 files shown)


angelosilvestre:
It seems that tests are failing for super_text_layout on stable. I'll need to check.

angelosilvestre:
@matthew-carroll Updated the PR explaining the changes.

matthew-carroll:
@angelosilvestre I'm concerned about the text position change. Are you able to tell what changed and why?

angelosilvestre:
@matthew-carroll It seems this was already changed in master after https://github.com/superlistapp/super_editor/pull/1277

I cherry-picked the commit.

PR #1619, opened 2023-11-24
[SuperTextField] Fix text capitalization configuration (Resolves #1617)

### LINKED ISSUE #1617: [SuperEditor][SuperTextField][Mobile] Unable to change `TextCapitalization` of the text field
Changes to the `textCapitalization` in the `imeConfiguration` of a `SuperTextField` seem to have no effect.
The virtual keyboard always shows lower case characters and disregards any text capitalization setting.
Tested on the latest main. This happens both on Android and iOS.

I assume this is a bug because it works with the default Flutter TextField. Let me know if I've got something wrong.

### How to reproduce:
In the super_editor example, modify the `demo_superiostextfield.dart` and add any `textCapitalization` to the `imeConfiguration` of the `SuperIOSTextField`.
```dart
imeConfiguration: TextInputConfiguration(
  textCapitalization: TextCapitalization.characters,
),
```

angelosilvestre:
[SuperTextField] Fix text capitalization configuration. Resolves #1617

Passing a `TextInputConfiguration` with a `textCapitalization` different from `TextCapitalization.none` has  no effect.

Two issues are involved on this:

1. `ImeAttributedTextEditingController.updateTextInputConfiguration` doesn't take a `TextCapitalization`.

When we added support for passing a `TextInputConfiguration` to `SuperTextField`, we kept `updateTextInputConfiguration` taking individual arguments instead of a `TextInputConfiguration` to avoid a breaking change.

I suggest we make that breaking change and take a `TextInputConfiguration` as argument. Otherwise, we need to take all `TextInputConfiguration` properties in this method.

2. We are re-attaching to the IME on every call to `didUpdateWidget` if we have a `imeConfiguration`.

The reason is that we are comparing only `widget.imeConfiguration != oldWidget.imeConfiguration`. This compares  the references, not the `imeConfiguration` properties. So, if we don't use a `const` `imeConfiguration`, in each build we get a different instance, which causes us to re-attach to the IME. 

This PR changes the comparison to also compare all  `TextInputConfiguration` properties, by using an extension method.


super_editor/test/super_textfield/super_textfield_ime_test.dart
```diff
@@ -529,6 +528,148 @@ void main() {
...
+    expect(autocorrect, false);
+    expect(enableSuggestions, false);
+    expect(enableDeltaModel, true);
+    expect(textCapitalization, 'TextCapitalization.characters');
+    expect(keyboardAppearance, 'Brightness.dark');
+
+    // Change the brightness to rebuild the widget
+    // and re-attach to the IME.
+    brightnessNotifier.value = Brightness.light;
+    await tester.pumpAndSettle();
```
matthew-carroll, line 603: Do you need to settle here?

super_editor/test/super_textfield/super_textfield_ime_test.dart
```diff
@@ -529,6 +528,148 @@ void main() {
...
+    final rebuildSignal = SignalNotifier();
+
+    // Pump a SuperTextField with an IME configuration with values
+    // that differ from the defaults.
+    await tester.pumpWidget(
+      _buildScaffold(
+        child: ListenableBuilder(
+          listenable: rebuildSignal,
+          builder: (context, child) {
+            // Explicitly avoid using const to ensure that each build
```
matthew-carroll, line 629: How about instead of the careful non-const approach, you declare two different variables with the same IME config value and you switch from one to the other?

super_editor/test/super_textfield/super_textfield_ime_test.dart
```diff
@@ -529,6 +528,148 @@ void main() {
     });
   });
 
+  testWidgetsOnAllPlatforms('updates custom IME configuration when configuration changes', (tester) async {
```
matthew-carroll, line 531: I'm not sure the word "custom" is important here. Let's just say "updates IME configuration when it changes"

super_editor/lib/src/infrastructure/text_input.dart
```diff
@@ -35,3 +37,42 @@ enum WebPlatformOverride {
   /// Configuration to run the app as if we are on web.
   web,
 }
+
+extension TextInputConfigurationExtensions on TextInputConfiguration {
```
matthew-carroll, line 41: Does `TextInputConfiguration` really not implement a custom `==`? If not, can you file a Flutter issue to at least put this on their radar?

Let's name these extensions more specifically. We may never add another extension to them, so let's say what they really do, e.g., `TextInputConfigurationEquivalency`

Also, let's place these in a location that we don't export from the package. These exist for an internal detail of `super_editor`. I'd rather not be on the hook for maintaining these extensions, and the Flutter team might eventually implement this behavior directly.

super_editor/test/super_textfield/super_textfield_ime_test.dart
angelosilvestre: `pump` is enough. Updated.

super_editor/test/super_textfield/super_textfield_ime_test.dart
angelosilvestre: Updated.

super_editor/lib/src/infrastructure/text_input.dart
angelosilvestre: Filed https://github.com/flutter/flutter/issues/139033

angelosilvestre:
@matthew-carroll The test failures are the same as the previous PR.

super_editor/lib/src/infrastructure/flutter/text_input_configuration.dart
```diff
@@ -0,0 +1,41 @@
...
+  /// Whether this [TextInputConfiguration] is equivalent to [other].
+  ///
+  /// Two [TextInputConfiguration]s are considered to be equal
+  /// if all properties are equal.
+  bool isEquivalentTo(TextInputConfiguration other) {
+    return inputType == other.inputType &&
+        readOnly == other.readOnly &&
+        obscureText == other.obscureText &&
+        autocorrect == other.autocorrect &&
+        autofillConfiguration.isEquivalentTo(other.autofillConfiguration) &&
```
matthew-carroll, line 14: I wanted to point one thing out that might be relevant. I checked the API for `AutofillConfiguration` and it seems to include the current `TextEditingValue`. That suggests to me that every time the user types a character, deletes a character, or moves the caret, the overall `TextInputConfiguration` will be seen as unequal. Do you also see it that way? If so, do we really want that configuration getting sent every time those things happen?

angelosilvestre: As this is a configuration, I would expect that `AutofillConfiguration` stores only the initial `TextEditingValue` without updating it at every keystroke, but the docs aren't clear about that and I've never used the `AutofillConfiguration` before.

I don't think we want to consider the `TextInputConfiguration` unequal at every `TextEditingValue`. Should I remove it from the comparison?

angelosilvestre: Updated.

matthew-carroll (approved):
LGTM

PR #1626, opened 2023-11-28
[SuperReader][SuperTextField] Applies color attribution on textstyle created through default style builder (Resolves #1579)

### LINKED ISSUE #1579: [SuperEditor & SuperTextField] ColorAttribution support
Hello, 

Is text coloring for SuperEditor, SuperReader and SuperTextField under development? 

ColorAttribution doesn't seem to be working at the moment.

Thank you.

```
ParagraphNode(
  id: Editor.createNodeId(),
  text: AttributedText(
    'My text colored in orange',
    AttributedSpans(
      attributions: const [
        SpanMarker(
          attribution: boldAttribution,
          offset: 0,
        ),
        SpanMarker(
          attribution: ColorAttribution(
            Colors.orange,
          ),
          offset: 0,
        ),
        SpanMarker(
          attribution: ColorAttribution(
            Colors.orange,
          ),
          offset: 24,
        ),
        SpanMarker(
          attribution: boldAttribution,
          offset: 24,
        ),
      ],
    ),
  ),
),
```

Capture d’écran 2023-10-30 à 10 43 43

rutvik110:
[SuperReader][SuperTextField] Applies color attribution on textstyle created through default style builder (Resolves #1579)

### **The Issue:** 

Colored attributions set on text passed to `SuperReader` and `SuperTextField` aren't applied to the final rendered text within the document. 

With following attributed text passed to both `SuperReader` and `SuperTextField`, 

```
AttributedText(
    'My text colored in orange',
    AttributedSpans(
      attributions: const [
        SpanMarker(
          attribution: ColorAttribution(
            Colors.orange,
          ),
          offset: 0,
        ),
        SpanMarker(
          attribution: ColorAttribution(
            Colors.orange,
          ),
          offset: 24,
        ),
      ],
    ),
  ),
```
<br>

`SuperReader` output:

Screenshot 2023-11-29 at 2 44 24 AM
<br>
<br>

`SuperTextField` output(on Desktop):

Screenshot 2023-11-30 at 1 41 35 PM

<br>
<br>

### **Solution:**

Turns out that the `readOnlyDefaultStyleBuilder` used by `SuperReader` and `defaultTextFieldStyleBuilder` used by `SuperTextField` which are responsible for creating a text style based on set of text attributions weren't applying the passed color attribution to the created text style.

This PR updates `readOnlyDefaultStyleBuilder` and `defaultTextFieldStyleBuilder` to include a check to see if incoming `Attribution` is of type `ColorAttribution`, if it's then updates text style color with color set within `ColorAttribution`.

`SuperReader` output:

Screenshot 2023-11-29 at 2 44 10 AM

<br>

`SuperTextField` output(on Desktop):

Screenshot 2023-11-30 at 1 42 24 PM

### DIFF
--- a/super_editor/test/super_editor/supereditor_attributions_test.dart
+++ b/super_editor/test/super_editor/supereditor_attributions_test.dart
@@ -3,16 +3,11 @@ import 'package:flutter/services.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:flutter_test_robots/flutter_test_robots.dart';
 import 'package:flutter_test_runners/flutter_test_runners.dart';
-import 'package:super_editor/src/default_editor/attributions.dart';
-import 'package:super_editor/src/default_editor/text.dart';
-import 'package:super_editor/src/infrastructure/text_input.dart';
 import 'package:super_editor/src/test/ime.dart';
 import 'package:super_editor/src/test/super_editor_test/supereditor_inspector.dart';
 import 'package:super_editor/src/test/super_editor_test/supereditor_robot.dart';
 import 'package:super_editor/super_editor.dart';
-import 'package:super_text_layout/super_text_layout.dart';
 
-import '../super_textfield/super_textfield_inspector.dart';
 import 'supereditor_test_tools.dart';
 import 'test_documents.dart';
 
@@ -199,7 +194,7 @@ void main() {
 
         // Ensure the text is colored orange.
         expect(
-          SuperTextFieldInspector.findRichText().style?.color,
+          SuperEditorInspector.findRichTextInParagraph("1").style?.color,
           Colors.orange,
         );
       });
@@ -214,13 +209,19 @@ void main() {
 
         // Ensure the first span is colored black.
         expect(
-          SuperTextFieldInspector.findRichText().getSpanForPosition(const TextPosition(offset: 0))!.style!.color,
+          SuperEditorInspector.findRichTextInParagraph("1")
+              .getSpanForPosition(const TextPosition(offset: 0))!
+              .style!
+              .color,
           Colors.black,
         );
 
         // Ensure the second span is colored orange.
         expect(
-          SuperTextFieldInspector.findRichText().getSpanForPosition(const TextPosition(offset: 5))!.style!.color,
+          SuperEditorInspector.findRichTextInParagraph("1")
+              .getSpanForPosition(const TextPosition(offset: 5))!
+              .style!
+              .color,
           Colors.orange,
         );
       });
--- a/super_editor/test/super_editor/supereditor_attributions_test.dart
+++ b/super_editor/test/super_editor/supereditor_attributions_test.dart
@@ -12,6 +12,7 @@ import 'package:super_editor/src/test/super_editor_test/supereditor_robot.dart';
 import 'package:super_editor/super_editor.dart';
 import 'package:super_text_layout/super_text_layout.dart';
 
+import '../super_textfield/super_textfield_inspector.dart';
 import 'supereditor_test_tools.dart';
 import 'test_documents.dart';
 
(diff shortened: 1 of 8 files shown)


rutvik110:
@matthew-carroll Do you think we should add a common text styler for both `SuperEditor` and `SuperReader`? Currently both of them have independent text stylers, `readOnlyDefaultInlineTextStyler` for `SuperReader` and `defaultInlineTextStyler` for `SuperEditor`. They both achieve the same purpose at the end and their implementation isn't tied to either `SuperEditor` or `SuperReader` exclusively.

rutvik110:
@matthew-carroll Updated with changes from the recent `KeyEvent` migration, #1593

super_editor/test/super_editor/supereditor_attributions_test.dart
```diff
@@ -183,6 +186,95 @@ void main() {
...
+                            markerType: SpanMarkerType.end,
+                          ),
+                        ],
+                      ),
+                    ),
+                  )
+                ],
+              ),
+            )
+            .withInputSource(TextInputSource.ime)
```
matthew-carroll, line 219: Why do we need to specify IME?

super_editor/test/super_editor/supereditor_attributions_test.dart
```diff
@@ -183,6 +186,95 @@ void main() {
...
+            .pump();
+
+        final superText = tester.widget<SuperText>(find.byType(SuperText));
```
matthew-carroll, line 222: Is this how we're verifying text color in existing tests?

super_editor/test/super_editor/supereditor_attributions_test.dart
```diff
@@ -183,6 +186,95 @@ void main() {
...
+
+        // Ensure the text is colored orange.
+        expect(
+          superText.richText.style!.color,
+          Colors.orange,
+        );
+      });
```
matthew-carroll, line 229: Please add blank lines between tests.

super_editor/test/super_editor/supereditor_attributions_test.dart
```diff
@@ -183,6 +186,95 @@ void main() {
...
+                  )
+                ],
+              ),
+            )
+            .withInputSource(TextInputSource.ime)
+            .pump();
+
+        final superText = tester.widget<SuperText>(find.byType(SuperText));
+
+        // Ensure the first span is colored black.
```
matthew-carroll, line 264: You've checked the first character in each span, but you haven't checked any other characters. For example, if the 2nd character somehow became orange, this test would still pass.

super_editor/test/super_editor/supereditor_attributions_test.dart
```diff
@@ -183,6 +186,95 @@ void main() {
       });
     });
 
+    group("applies color attributions", () {
+      testWidgetsOnAllPlatforms("to full text", (tester) async {
+        await tester //
+            .createDocument()
+            .withCustomContent(
```
matthew-carroll, line 193: Instead of creating a verbose custom document in these tests, why not use the existing single paragraph document and then run an `Editor` request to apply the color attribution to the span you want? Then we test the full execution path to change text color.

Also, do we not have any existing tests for the text color attribution? I'm surprised that I merged that change without any tests...

rutvik110: > Also, do we not have any existing tests for the text color attribution? I'm surprised that I merged that change without any tests...

I'vent't found any traces of such tests anywhere. If we had them then I believe this issue would've been catched by those tests. That doesn't being the case means we don't seem to have tests for checking this behaviour.

super_editor/test/super_editor/supereditor_attributions_test.dart
rutvik110: We don't seem to have any tests rn that are verifying text color. Atm I didn't found any other good approach to test for text color reliably where we can be sure that the text style i.e constrcuted and used by the rich text object is the one that's being used by the final text that gets rendered.

If there's a better solution/suggestion for testing text color that you're aware of, I'm open to it.

matthew-carroll (approved):
LGTM

PR #1611, opened 2023-11-22
[SuperEditor][SuperReader][iOS] Fix horizontal text selection inside a CustomScrollView (Resolves #1610)

angelosilvestre:
[SuperEditor][SuperReader][iOS] Fix horizontal text selection inside a CustomScrollView. Resolves #1610

Performing a horizontal drag isn't changing the selection when the editor is inside a `CustomScrollView`.

The reason seems to be that the `VerticalDragGestureRecognizer` is losing in the gesture arena when inside an ancestor scrollable. Adding a `HorizontalDragGestureRecognizer` to the `RawGestureDetector` seems to fix the issue.

This PR doesn't add tests because so far I wasn't able to write a failing test with the current implementation. I'll file a ticket to write tests for this.

https://github.com/superlistapp/super_editor/assets/7597082/7bda2411-6374-4631-a5cf-81c23cde0879

Issue #1610, opened 2023-11-21
[SuperEditor] [iOS] Text selection breaks when SuperEditor is built inside a CustomScrollView


robertodoering:
When a `SuperEditor` is a child of a `CustomScrollView`, expanding the text selection via a horizontal drag in a single line is no longer possible.
This only happens on iOS.

Example:
https://github.com/superlistapp/super_editor/assets/20045287/2f07b1b8-9952-45d0-b605-10c04fdad180

How to reproduce:
Add a `CustomScrollView` with a single `SliverToBoxAdapter` sliver somewhere above the `SuperEditor`.

In the example app in `example_editor.dart`, I've modified to `_buildEditor` method to:
<details>

```dart
  Widget _buildEditor(BuildContext context) {
    final isLight = Theme.of(context).brightness == Brightness.light;

    return CustomScrollView(
      slivers: [
        SliverToBoxAdapter(
          child: ColoredBox(
            color: isLight ? _lightBackground : _darkBackground,
            child: SuperEditorDebugVisuals(
              config: _debugConfig ?? const SuperEditorDebugVisualsConfig(),
              child: KeyedSubtree(
                key: _viewportKey,
                child: SuperEditorIosControlsScope(
                  controller: _iosControlsController,
                  child: SuperEditor(
                    editor: _docEditor,
                    document: _doc,
                    composer: _composer,
                    focusNode: _editorFocusNode,
                    scrollController: _scrollController,
                    documentLayoutKey: _docLayoutKey,
                    documentOverlayBuilders: [
                      DefaultCaretOverlayBuilder(
                        caretStyle: const CaretStyle().copyWith(color: isLight ? Colors.black : Colors.redAccent),
                      ),
                      SuperEditorIosToolbarFocalPointDocumentLayerBuilder(),
                      SuperEditorIosHandlesDocumentLayerBuilder(),
                    ],
                    selectionLayerLinks: _selectionLayerLinks,
                    selectionStyle: isLight
                        ? defaultSelectionStyle
                        : SelectionStyles(
                            selectionColor: Colors.red.withOpacity(0.3),
                          ),
                    stylesheet: defaultStylesheet.copyWith(
                      addRulesAfter: [
                        if (!isLight) ..._darkModeStyles,
                        taskStyles,
                      ],
                    ),
                    componentBuilders: [
                      TaskComponentBuilder(_docEditor),
                      ...defaultComponentBuilders,
                    ],
                    gestureMode: _gestureMode,
[...]

PR #1627, opened 2023-11-29
[SuperEditor][iOS] Add tests for drag to select (Resolves #1613)

### LINKED ISSUE #1613: [SuperEditor][SuperReader][iOS] Add tests for selection gestures inside a CustomScrollView
It was reported in https://github.com/superlistapp/super_editor/issues/1610 that performing a horizontal drag isn't changing the selection:

https://github.com/superlistapp/super_editor/assets/20045287/2f07b1b8-9952-45d0-b605-10c04fdad180

https://github.com/superlistapp/super_editor/pull/1611 is an attempt to fix this issue, but it doesn't include tests.

Add tests simulating both horizontal and vertical drag when the editor is inside a `CustomScrollView` to avoid breaking this is the future.

angelosilvestre:
[SuperEditor][iOS] Add tests for drag to select. Resolves #1613

It was reported in https://github.com/superlistapp/super_editor/issues/1610 that performing a horizontal drag isn't changing the selection. https://github.com/superlistapp/super_editor/pull/1611 fixed the bug, but didn't include any tests.

This PR adds test for horizontal dragging inside a `CustomScrollView`. During the tests I got some failures because dragging can cause an `upstream` selection affinity.  I added a custom `Matcher` to compare the selections, ignoring the affinity. We have something similar to this in `NodePosition.isEquivalentTo`. 

Initially, these tests weren't failing, even without the fix. The reason is that `dragSelectDocumentFromPositionByOffset` always uses `PointerDeviceKind.mouse` as the device kind when starting the gesture. As this device kind isn't considered a drag device in mobile, the `CustomScrollView` didn't try to accept the gesture.

This PR adds a `pointerDeviceKind` parameter to `dragSelectDocumentFromPositionByOffset`, so we can override the device kind. 

We should consider changing all `dragXX` methods to default the device kind depending on the platform

### DIFF
--- a/super_editor/lib/src/test/super_editor_test/supereditor_robot.dart
+++ b/super_editor/lib/src/test/super_editor_test/supereditor_robot.dart
@@ -1,3 +1,4 @@
+import 'package:flutter/foundation.dart';
 import 'package:flutter/gestures.dart';
 import 'package:flutter/services.dart';
 import 'package:flutter/widgets.dart';
@@ -156,10 +157,14 @@ extension SuperEditorRobot on WidgetTester {
   /// to ensure that the drag rectangle never has a zero-width or a
   /// zero-height, because such a drag rectangle wouldn't be seen as
   /// intersecting any content.
+  ///
+  /// Provide a [pointerDeviceKind] to override the device kind used in the gesture.
+  /// If [pointerDeviceKind] is `null`, it defaults to [PointerDeviceKind.touch]
+  /// on mobile, and [PointerDeviceKind.mouse] on other platforms.
   Future<void> dragSelectDocumentFromPositionByOffset({
     required DocumentPosition from,
     required Offset delta,
-    PointerDeviceKind pointerDeviceKind = PointerDeviceKind.mouse,
+    PointerDeviceKind? pointerDeviceKind,
     Finder? superEditorFinder,
   }) async {
     final documentLayout = _findDocumentLayout(superEditorFinder);
@@ -197,8 +202,13 @@ extension SuperEditorRobot on WidgetTester {
       }
     }
 
+    final deviceKind = pointerDeviceKind ??
+        (defaultTargetPlatform == TargetPlatform.iOS || defaultTargetPlatform == TargetPlatform.android
+            ? PointerDeviceKind.touch
+            : PointerDeviceKind.mouse);
+
     // Simulate the drag.
-    final gesture = await startGesture(dragStartOffset, kind: pointerDeviceKind);
+    final gesture = await startGesture(dragStartOffset, kind: deviceKind);
 
     // Move slightly so that a "pan start" is reported.
     //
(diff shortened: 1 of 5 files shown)


super_editor/lib/src/test/super_editor_test/supereditor_robot.dart
```diff
@@ -159,6 +159,7 @@ extension SuperEditorRobot on WidgetTester {
   Future<void> dragSelectDocumentFromPositionByOffset({
     required DocumentPosition from,
     required Offset delta,
+    PointerDeviceKind pointerDeviceKind = PointerDeviceKind.mouse,
```
matthew-carroll, line 162: If this was added because all mobile tests need to drag with a finger instead of a mouse, then that means every mobile test will have to specify this. That seems fragile. Would it make more sense to make this nullable and then if it's `null`, the implementation chooses the most sensible device kind for the default platform?

super_editor/test/test_tools.dart
```diff
@@ -46,3 +47,74 @@ extension DragExtensions on WidgetTester {
     return dragGesture;
   }
 }
+
+/// Compares two selections, ignoring selection affinities.
```
matthew-carroll, line 51: Most of what's here looks like it should exist in the package, not in test code.  I assume that inspecting selection bounds is something that anyone might want to do.

Also, in this case, I don't think "selection equivalency" is appropriate, the way it was with node positions. Upstream vs downstream could truly result in different rendering, whereas the affinity for a single position shouldn't impact any visual behavior. I would approach this condition as one of equivalent bounds, or perhaps even equal bounds. But the point is that it's about the bounds, not the overall selection.

super_editor/lib/src/test/super_editor_test/supereditor_robot.dart
angelosilvestre: Updated.

super_editor/test/test_tools.dart
angelosilvestre: > Most of what's here looks like it should exist in the package, not in test code. I assume that inspecting selection bounds is something that anyone might want to do.

Are you suggesting that we add a method to the `DocumentSelection` class? I'm not sure we'll be using `Matcher`s in the package code.

> Also, in this case, I don't think "selection equivalency" is appropriate, the way it was with node positions. Upstream vs downstream could truly result in different rendering, whereas the affinity for a single position shouldn't impact any visual behavior.

My description wasn't accurate. I'm not ignoring the selection affinity itself, I'm ignoring the affinity of each `NodePosition` included in the selection.

matthew-carroll: Yeah, we wouldn't add a `Matcher` to the source code. I was looking mostly at the code related to inspecting the selection. If this matcher still respects affinity of the selection, but ignores the affinity of the node positions, then that's a very specific differentiation. I wonder if that nuance might get lost and misunderstood over time. 

Does this matcher need to care about selection affinity? Would it make sense to look only at the bounds, while ignoring all affinity?

angelosilvestre: > If this matcher still respects affinity of the selection, but ignores the affinity of the node positions, then that's a very specific differentiation.

This is what it does.

> Does this matcher need to care about selection affinity? Would it make sense to look only at the bounds, while ignoring all affinity?

I think it should care about selection affinity. As it was mentioned, the selection affinity will impact the selection rendering.

matthew-carroll (approved):
LGTM

PR #1630, opened 2023-11-30
Remove unused parameters on private classes

gspencergoog:
## Description

This removes unused parameters from some private classes in `content_layers_test.dart`, since there is currently a bug in `dart fix --apply` where it will remove the parameters that are unused, but not the associated members.

## Related Issues
 - https://github.com/dart-lang/sdk/issues/54190

## Related PRs
 - https://github.com/flutter/tests/pull/316

### DIFF
--- a/super_editor/test/infrastructure/content_layers_test.dart
+++ b/super_editor/test/infrastructure/content_layers_test.dart
@@ -542,9 +542,7 @@ class _RebuildableWidget extends StatefulWidget {
     Key? key,
     this.rebuildSignal,
     this.buildTracker,
-    this.elementTracker,
     this.onBuildScheduled,
-    this.onBuild,
     this.builder,
     this.child,
   })  : assert(child != null || builder != null, "Must provide either a child OR a builder."),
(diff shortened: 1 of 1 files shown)


super_editor/test/infrastructure/content_layers_test.dart
```diff
@@ -542,9 +542,7 @@ class _RebuildableWidget extends StatefulWidget {
     Key? key,
     this.rebuildSignal,
     this.buildTracker,
-    this.elementTracker,
```
gspencergoog, line 545: @matthew-carroll These parameters are not set anywhere in this file, and since they are in a private class, it should be safe to remove them.

matthew-carroll: I'd prefer that we annotate these to skip the linter. We may very well want to use those properties when something eventually goes wrong and we need to debug it again.

gspencergoog: Sure.  Done.

matthew-carroll (approved):
LGTM

PR #1635, opened 2023-12-01
[SuperText] - Added an underline layer (Resolves #1634)

### LINKED ISSUE #1634: [SuperText] - Add layer to draw underlines
We need to draw underlines across a `TextRange` so that we can show the composing region for #1603.

Add a `SuperText` layer that can draw underlines.

matthew-carroll:
[SuperText] - Added an underline layer (Resolves #1634)


matthew-carroll:
@angelosilvestre I'm gonna merge this so I can release a new version of super_text_layout and use it in super_editor.

super_text_layout/lib/src/text_underline_layer.dart
```diff
@@ -0,0 +1,159 @@
+import 'package:collection/collection.dart';
+import 'package:flutter/widgets.dart';
+
+import 'text_layout.dart';
+
+/// A [SuperText] layer that displays an underline beneath the text within a given
+/// selection.
```
angelosilvestre, line 7: Wouldn't it make more sense if we say "within a given range" instead of "within a given selection" ?

angelosilvestre (commented):
LGTM

PR #1639, opened 2023-12-02
[SuperEditor][SuperTextField] - Paint underline beneath text in composing region (Resolves 1603)

matthew-carroll:
[SuperEditor][SuperTextField] - Paint underline beneath text in composing region (Resolves 1603)

The underline is the same color as the base text style, as determined by the attributed style builder for the `SuperEditor` or the `SuperTextField`.

The painting of the underlines is built on top of a recent change that I merged in https://github.com/superlistapp/super_editor/pull/1635

https://github.com/superlistapp/super_editor/assets/7259036/d0de826a-2270-40c8-80d3-ea989ac9f7db

https://github.com/superlistapp/super_editor/assets/7259036/d70fc33a-c697-4254-9ed3-8a972f17afbc

## Split selection change from composing region change
Previously we were sending the composing region as part of selection change events. I separated these into different requests and events. I ran into places where selection changes didn't necessarily have an opinion about the composing region but were being forced to set it. To change a selection and composing region at the same time, two requests can be passed to a single `execute` call on the `Editor`, so that they're processed as a transaction.


matthew-carroll:
@miguelcmedeiros I recommend trying this PR with any/all situations where you expect a composing region. The only situation I remembered was the accent panel. Let me know if any other composing region use-case fails to show the expected underline.

super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
```diff
@@ -192,7 +200,9 @@ class TextDeltasDocumentEditor {
       return;
     }
 
-    replace(delta.replacedRange, delta.replacementText);
+    final replacementComposingRegion = _serializedDoc.imeToDocumentRange(delta.composing);
```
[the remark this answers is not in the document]
angelosilvestre: If I remember correctly, it should be the composing region after the replacement.

super_editor/lib/src/default_editor/layout_single_column/_styler_user_selection.dart
```diff
@@ -22,22 +22,29 @@ class SingleColumnLayoutSelectionStyler extends SingleColumnLayoutStylePhase {
...
+    _composingRegion?.addListener(() {
+      markDirty();
+    });
   }
 
   @override
   void dispose() {
     _selection.removeListener(markDirty);
+    _composingRegion?.removeListener(markDirty);
     super.dispose();
```
angelosilvestre, line 42: We are adding a listener of an anonymous function that calls `markDirty` and trying to remove `markDirty` from the listeners.

It seems we should add `markDirty` directly as a listener, like we do in the `_selection`.

super_editor/lib/src/default_editor/text.dart
```diff
@@ -905,20 +918,48 @@ class TextComponentState extends State<TextComponent> with DocumentComponent imp
...
+                  textLayout: textLayout,
+                  underlines: [
+                    TextLayoutUnderline(
+                      style: UnderlineStyle(
+                        color: widget.textStyleBuilder({}).color ?? //
+                            (Theme.of(context).brightness == Brightness.light ? Colors.black : Colors.white),
+                      ),
+                      range: widget.composingRegion!,
+                    ),
+                  ],
```
angelosilvestre, line 958: One thing to note is that Windows and Linux don't display an underline below the composing region.

miguelcmedeiros: True, other Windows shows something else when composing Japanese:

Screenshot 2023-12-05 at 00 04 43

Not sure about Linux...

miguelcmedeiros:
@matthew-carroll, from the recordings it looks great!!! 

I will be able to try it out the earliest tomorrow/Wednesday.

> The only situation I remembered was the accent panel.

In the meanwhile you can try it out with Japanese keyboard layout which also opens an IME panel.

super_editor/lib/src/default_editor/text.dart
miguelcmedeiros: To be honest, I wouldn't mind in a first instance to have the same underline on all platforms and then optimize per platform separately.

angelosilvestre:
With a portuguese keyboard, we also have an underline when entering compound characters like "ã", by typing ~ + a

super_editor/lib/src/default_editor/text.dart
matthew-carroll: I moved the composing region underline to its own editor style, and that's now configured to not show an underline on Windows or Linux by default. I made a similar change to SuperTextField.

matthew-carroll:
@angelosilvestre it looks like goldens are failing in CI but passing locally. I changed one of them use to our pixel diff allowance and that test says its got 100px of difference, which is significant. Let me know if you have any ideas for how we can debug that.

super_editor/lib/src/default_editor/layout_single_column/_styler_composing_region.dart
```diff
@@ -0,0 +1,280 @@
...
+  /// The ID of the node that's selected.
+  final String nodeId;
+
+  /// The selection within the given node.
+  final SelectionType? nodeSelection;
+
+  /// Whether this [_DocumentNodeSelection] forms the base position of a larger
+  /// document selection, `false` otherwise.
+  ///
+  /// [isBase] is `true` iff [nodeId] is the same as [DocumentSelection.base.nodeId].
```
angelosilvestre, line 247: "iff" -> "if"

super_editor/lib/src/default_editor/layout_single_column/_styler_composing_region.dart
```diff
@@ -0,0 +1,280 @@
...
+  final bool isBase;
+
+  /// Whether this [_DocumentNodeSelection] forms the extent position of a
+  /// larger document selection, `false` otherwise.
+  ///
+  /// [isExtent] is `true` iff [nodeId] is the same as [DocumentSelection.extent.nodeId].
```
angelosilvestre, line 253: "iff" -> "if"

super_editor/lib/src/default_editor/layout_single_column/_styler_composing_region.dart
```diff
@@ -0,0 +1,280 @@
...
+  @override
+  bool operator ==(Object other) =>
+      identical(this, other) ||
+      other is _DocumentNodeSelection &&
+          runtimeType == other.runtimeType &&
+          nodeId == other.nodeId &&
+          nodeSelection == other.nodeSelection;
+
+  @override
+  int get hashCode => nodeId.hashCode ^ nodeSelection.hashCode;
```
angelosilvestre, line 274: Don't we need the other properties in the `==` and `hashCode`?

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -504,13 +506,22 @@ class SuperEditorState extends State<SuperEditor> {
...
     );
 
+    _docLayoutComposingRegionStyler = SingleColumnLayoutComposingRegionStyler(
+      document: document,
+      composingRegion: editContext.composer.composingRegion,
+      showComposingUnderline: defaultTargetPlatform == TargetPlatform.macOS ||
+          defaultTargetPlatform == TargetPlatform.iOS ||
+          defaultTargetPlatform == TargetPlatform.android,
+    );
+
```
angelosilvestre, line 516: Wouldn't it be simpler if we don't add the `SingleColumnLayoutComposingRegionStyler` to the pipeline if we don't want the underline? That way, this phase won't need to run and copy all view models if we don't need.

super_editor/test_goldens/super_textfield/super_textfield_composing_region_underline_test.dart
```diff
@@ -0,0 +1,88 @@
...
+  await _pumpScaffold(tester, textController);
+
+  textController
+    ..selection = const TextSelection.collapsed(offset: 23)
+    ..composingRegion = const TextRange(start: 22, end: 23);
+  await tester.pumpAndSettle();
+
+  // Ensure the composing region is underlined.
+  // await screenMatchesGolden(
+  //     tester, "super-text-field_composing-region-shows-underline_${defaultTargetPlatform.name}_1");
```
angelosilvestre, line 35: A reminder to remove the commented code.

super_editor/lib/src/default_editor/layout_single_column/_styler_composing_region.dart
```diff
@@ -0,0 +1,280 @@
...
+        documentComposingRegion.start.nodeId == node.id) {
+      // There's a composing region and it's entirely within this text node.
+      // TODO: handle the possibility of a composing region extending across multiple nodes.
+      final startPosition = documentComposingRegion.start.nodePosition as TextNodePosition;
+      final endPosition = documentComposingRegion.end.nodePosition as TextNodePosition;
+      textComposingRegion = TextRange(start: startPosition.offset, end: endPosition.offset);
+    }
+
+    // TODO: instead of reporting the composing region to the view model, have the view model
+    //       take a collection of underlines with underline styles.
```
angelosilvestre, line 100: If we do that, we'll probably be able to reuse it to spell checking, showing an indicator below the misspelled words.

angelosilvestre:
> @angelosilvestre it looks like goldens are failing in CI but passing locally. I changed one of them use to our pixel diff allowance and that test says its got 100px of difference, which is significant. Let me know if you have any ideas for how we can debug that.

One possible thing is to prune all docker cache by running `docker system prune`.

I can try to run the golden tests on Linux via WSL to see if it also fails. One of the weird test failures I was able to reproduce running that way.

super_editor/lib/src/default_editor/layout_single_column/_styler_composing_region.dart
matthew-carroll: I deleted those other properties because they weren't needed. This stuff was copied from the selection styler.

super_editor/test_goldens/super_textfield/super_textfield_composing_region_underline_test.dart
matthew-carroll: I'm gonna leave that there until we figure out how to get these to pass in CI. I'm still getting passing tests locally and failing tests on GitHub.

super_editor/lib/src/default_editor/super_editor.dart
matthew-carroll: Yeah, makes sense.

super_editor/lib/src/default_editor/layout_single_column/_styler_composing_region.dart
matthew-carroll: I'll leave the TODO. We can make that change when we get there.

matthew-carroll:
The new goldens for the composing region pass locally, but fail in CI. I can't seem to figure out what's going on there. I'm gonna merge this because it's not clear what I can do to fix the CI issues or reproduce them locally.

PR #1648, opened 2023-12-06
[SuperEditor] Lazily compute nodes index in MutableDocument

jellynoone:
Currently, MutableDocument computes its nodes related indexes on every nodes modification. This seems inefficient when performing bulk edits known ahead of time. For instance, replacing all of the nodes with completely new ones, because, for instance, some remote contents changed.

My use case initially was replacing the whole document contents, but I realised this kind of lazy computation could provide benefits for other use-cases as well. For instance, removing nodes in ranges or similar.

### DIFF
--- a/super_editor/lib/src/core/editor.dart
+++ b/super_editor/lib/src/core/editor.dart
@@ -492,9 +492,7 @@ class MutableDocument implements Document, Editable {
   /// if provided, or empty content otherwise.
   MutableDocument({
     List<DocumentNode>? nodes,
-  }) : _nodes = nodes ?? [] {
-    _refreshNodeIdCaches();
-  }
+  }) : _nodes = nodes ?? [];
 
   void dispose() {
     _listeners.clear();
@@ -505,17 +503,16 @@ class MutableDocument implements Document, Editable {
   @override
   List<DocumentNode> get nodes => UnmodifiableListView(_nodes);
 
-  /// Maps a node id to its index in the node list.
-  final Map<String, int> _nodeIndicesById = {};
-
-  /// Maps a node id to its node.
-  final Map<String, DocumentNode> _nodesById = {};
-
   final _listeners = <DocumentChangeListener>[];
+
+  _Index? _index;
+  _Index _indexInit() {
+    return _index ??= _Index.fromNodes(nodes);
+  }
 
   @override
   DocumentNode? getNodeById(String nodeId) {
-    return _nodesById[nodeId];
+    return _indexInit().nodeFor(nodeId);
   }
 
   @override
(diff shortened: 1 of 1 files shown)


matthew-carroll:
I'd like to see direct evidence of performance problems before making a performance improvement. Do you have any evidence that the existing approach creates performance problems?

jellynoone:
Currently, I do not have direct evidence that this approach creates performance problems.
I didn't have a chance yet to stress test the editor.

I was simply looked at the feature set and am in the process of implementing a proof of concept editor built on top of super_editor. One of the requirements I have is being able to clear the document.
I saw that the only way to do this (without using a completely new document) is to remove all the nodes one by one.
I took a peek at the source code to see if there was any method available to do this in one go. There wasn't.

I then saw how indexes were being handled so I decided to make a contribution since the approach here, of essentially marking the document as needing an index rebuild, seemed more in-line with how everything else in the framework works i.e. marking a render object as needing layout or marking a widget as needing to be rebuilt etc.

My reasoning here was that if for instance the editor had 100 nodes and all were to be removed, the index would, with the current approach, need to be recomputed 100 times instead of just once.

If the change seems premature, don't hesitate to close the pull request.

matthew-carroll:
> Currently, I do not have direct evidence that this approach creates performance problems.
I didn't have a chance yet to stress test the editor.

In this case, we shouldn't move forward with this change. We should only act upon clear signals of performance issues.

> One of the requirements I have is being able to clear the document.

If this is your root requirement, then I would recommend filing a ticket for this. However, my first question on that ticket will be why you don't just replace your existing `MutableDocument` with a new `MutableDocument`? Maybe we've made that too difficult, or impossible, but that will be the first question.

> seemed more in-line with how everything else in the framework works i.e. marking a render object as needing layout or marking a widget as needing to be rebuilt etc

Marking things as dirty is a consequence of Flutter's internals moving through a pipeline. `super_editor`s core logical constructs, such as `MutableDocument`, have no relationship to Flutter's pipeline. However, all changes should be made through an `Editor` using `Request`s and `Command`s.

> My reasoning here was that if for instance the editor had 100 nodes and all were to be removed, the index would, with the current approach, need to be recomputed 100 times instead of just once.

I agree that if you want to clear a document, you shouldn't have to do it one node at a time. That seems to be the root issue here, not necessarily how we compute node IDs.

> If the change seems premature, don't hesitate to close the pull request.

I'm going to close this for the reasons mentioned above. I would encourage you to begin by filing issues rather than PRs so that we can hash out the fundamental features and causes of bugs before anyone spends time writing code.

jellynoone:
Thank you for a quick and detailed response. I really appreciate it.

> I'm going to close this for the reasons mentioned above. I would encourage you to begin by filing issues rather than PRs so that we can hash out the fundamental features and causes of bugs before anyone spends time writing code.

Will do.
For this particular issue I thought providing the code directly would make the conversation move faster.

PR #780, opened 2022-09-13
[SuperEditor] [Desktop] Fix trackpad scrolling (Resolves #771)

angelosilvestre:
[SuperEditor] [Desktop] Fix trackpad scrolling. Partially resolves #771 

After Flutter 3.3, dragging with two fingers on a trackpad triggers a pan gesture with a `kind` of `PointerDeviceKind.trackpad`, as documented [in this design doc](https://flutter.dev/go/trackpad-gestures). 

This change causes the editor to expand the selection instead of scrolling the document with this gesture.

This PR changes `DocumentMouseInteractor` to scroll the document instead of expanding selection when the drag device is a `PointerDeviceKind.trackpad`.

This PR is based on https://github.com/superlistapp/super_editor/pull/772.

PR #1649, opened 2023-12-06
[SuperEditor][SuperTextField][Desktop] Fix trackpad and scrollbar scrolling (Resolves #1628)

### LINKED ISSUE #1628: [SuperTextField] Content does not scroll
Steps to reproduce:
* Run the interactive demo for SuperTextField on macOS
* Add enough new lines to the content of the field so that it becomes scrollable
* Use trackpad or the scrollbar to scroll the content

**Expected:** content scrolls as expected
**Actual:** nothing happens

https://github.com/superlistapp/super_editor/assets/31278849/52c415a7-ee19-40dd-b544-fa057ea0ae6e

angelosilvestre:
[SuperEditor][SuperTextField][Desktop] Fix trackpad and scrollbar scrolling. Resolves #1628

There are two scrolling issues:

1. Dragging with two fingers doesn't cause the content to scroll.

We had a similar issue with `SuperEditor` in the past, which was resolved in https://github.com/superlistapp/super_editor/pull/780.  This PR applies the same solution applied to `SuperTextField`.

2. Dragging a scrollbar doesn't cause the content to scroll.

This is also an issue on `SuperEditor`. Flutter's `Scrollbar` uses the `ScrollPosition`'s scroll physics to determine whether or not the user gesture will scroll the content. As we are using `NeverScrollableScrollPhysics`, the user is prevented from dragging the scrollbar. We can't configure the scroll physics of a `Scrollbar`.

The only solution I found was to copy Flutter's `ScrollBar`, `MaterialScrollBar` and `CupertinoScrollbar` and a `ScrollPhysics` to them. I filed https://github.com/flutter/flutter/issues/139666 asking for this to be implemented in Flutter.

On `SuperTextField`, the scrollbar had to be placed above the gesture interactor. Otherwise, tapping the scrollbar thumb causes the selection to change.

### DIFF
--- a/super_editor/lib/src/default_editor/document_scrollable.dart
+++ b/super_editor/lib/src/default_editor/document_scrollable.dart
@@ -212,6 +212,14 @@ class _DocumentScrollableState extends State<DocumentScrollable> with SingleTick
     required Widget child,
   }) {
     final scrollBehavior = ScrollConfiguration.of(context);
+    // As we handle the scrolling gestures ourselves,
+    // we use NeverScrollableScrollPhysics to prevent SingleChildScrollView
+    // from scrolling. This also prevents the user from interacting
+    // with the scrollbar.
+    // We use a modified version of Flutter's Scrollbar that allows
+    // configuring it with a different scroll physics.
+    //
+    // See https://github.com/superlistapp/super_editor/issues/1628 for more details.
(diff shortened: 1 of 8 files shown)


matthew-carroll (approved):
LGTM

PR #1651, opened 2023-12-08
[SuperEditor] Change IME serializer to always prepend invisible characters (Resolves #1641, Resolves #1642)

### LINKED ISSUE #1641: [Web/macOS] [SuperEditor] First character in paragraph is not part of composing region when using Japanese-Romaji
Setup:
* macOS 13.4.1
* Browser: Chrome 119.0.6045.199
* Keyboard: Japanese-romaji

Steps to reproduce:
* Run the demo app on a browser
* In a `SuperEditor` demo, add a new paragraph
* Place the caret in the empty paragraph and type "kamera"

Expected: "カメラ"
Actual: ”kあめら”

Notes:
* Could not reproduce this issue when running on macOS native
* Could not reproduce this issue on Windows, but found a different one (filled a separate ticket)
* Could not reproduce this issue with `SuperTextField`

### LINKED ISSUE #1642: [Web/Windows] [SuperEditor] First character in paragraph is duplicated when using Japanese keyboard layout
Setup:
* Windows 11
Browser: Chrome 119.0.6045.200
Keyboard: Japanese

Steps to reproduce:
* Run the demo app on a browser
* In a SuperEditor demo, add a new paragraph
* Place the caret in the empty paragraph and type "k"

Expected: "の"
Actual: ”のの”

Notes:
* Could not reproduce this issue when running on macOS native
* Could not reproduce this issue on macOS, but found a different one (filled a separate ticket)
* Could not reproduce this issue with SuperTextField

angelosilvestre:
[SuperEditor] Change IME serializer to always prepend invisible characters. Resolves #1641, Resolves #1642

When the selection sits at the beginning of a paragraph, we prepend the editing value sent to the IME with invisible the invisible characters `". "`. When the selection changes to another offset, we remove the prepended characters.

This isn't a problem in most situations. However, on web, when the IME is composing a Japanese character and we change the editing value is response to a character being typed by the user, the IME seems to get confused. On macOS web, composition restarts and the previous character isn't included in the new composing region. On windows web, the IME generates another insertion delta of the same character.

This PR changes the `DocumentImeSerializer` to always prepend the invisible characters. The tests that manually generate deltas had to be changed to account for the two additional characters. Also, there is a test that expects us to remove the prepended characters, so this test had to be changed to expect that we don't remove the invisible characters.

I noticed an unrelated bug in the demo app. We are using the Android versions of the overlay controls instead of the iOS version. This results in the overlay controls not being displayed. This PR also fixes that.


matthew-carroll (approved):
LGTM

PR #1660, opened 2023-12-10
Fixed double-tap and drag by word upstream (Resolves #1480)

### LINKED ISSUE #1480: [super_editor] Double click and drag has incorrect behavior in the upstream direction
Super Editor implements double-click-and-drag behavior to select by word. This works fine in the downstream direction but produces incorrect results in the upstream direction.

### Steps to reproduce

- Enter a line of text
- Double click and hold on the last word of the line
- Drag the mouse to the first word of the line

### Expected results

The selection should contain the entire line

### Actual results

The selection runs from just after to first word of the line to just before the last.

### Video

Here you can see the mismatch between the upstream and downstream behavior:

https://github.com/superlistapp/super_editor/assets/1316184/ba5ec480-f1d9-411e-8f7d-4c428445f831

matthew-carroll:
Fixed double-tap and drag by word upstream (Resolves #1480)

The root issue is that our double tap selection behavior was always treating the start of a word as the base of a selection. That works when dragging downstream. However, when dragging upstream, the base of the selection should be the end of the word, not the beginning.

This PR also adds a test tool to simulate a double tap and drag. Creating this test tool uncovered some kind of issue within our multi-tap gesture recognizer. I made a change that avoids the crash, but I don't have any idea why it's needed. The problem is tied to Flutter's gesture arena protocol, which I still don't understand.

### DIFF
--- a/super_editor/lib/src/default_editor/document_gestures_mouse.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_mouse.dart
@@ -90,6 +90,10 @@ class _DocumentMouseInteractorState extends State<DocumentMouseInteractor> with 
   Offset? _dragStartGlobal;
   Offset? _dragEndGlobal;
   bool _expandSelectionDuringDrag = false;
+  // When selecting by word, this is the initial word's upstream position.
+  DocumentPosition? _wordSelectionUpstream;
+  // When selecting by word, this is the initial word's downstream position.
+  DocumentPosition? _wordSelectionDownstream;
 
   /// Holds which kind of device started a pan gesture, e.g., a mouse or a trackpad.
   PointerDeviceKind? _panGestureDevice;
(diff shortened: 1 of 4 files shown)


super_editor/lib/src/default_editor/document_gestures_mouse.dart
```diff
@@ -90,6 +90,10 @@ class _DocumentMouseInteractorState extends State<DocumentMouseInteractor> with
   Offset? _dragStartGlobal;
   Offset? _dragEndGlobal;
   bool _expandSelectionDuringDrag = false;
+  // When selecting by word, this is the initial word's upstream position.
+  DocumentPosition? _wordSelectionUpstream;
+  // When selecting by word, this is the initial word's downstream position.
+  DocumentPosition? _wordSelectionDownstream;
```
angelosilvestre, line 96: Is this intended to be a regular comment instead of a dart doc?

matthew-carroll: Yeah. Those are private properties.

PR #1678, opened 2023-12-12
[SuperEditor] Enable Mac shortcuts for iOS (Resolves #1609)

angelosilvestre:
[SuperEditor] Enable Mac shortcuts for iOS. Resolves #1609

Mac shortcuts, like cmd + c, don't execute for iOS. The reason is that the `isPrimaryShortcutKeyPressed` getter only checks if the current platform is mac.

This PR changes `isPrimaryShortcutKeyPressed` to consider iOS and also enable other shortcuts that are running for mac only.

### DIFF
--- a/super_editor/lib/src/default_editor/document_hardware_keyboard/document_keyboard_actions.dart
+++ b/super_editor/lib/src/default_editor/document_hardware_keyboard/document_keyboard_actions.dart
@@ -79,13 +79,11 @@ ExecutionInstruction scrollOnCtrlOrCmdAndHomeKeyPress({
     return ExecutionInstruction.continueExecution;
   }
 
-  final isMacOrIos = defaultTargetPlatform == TargetPlatform.macOS || defaultTargetPlatform == TargetPlatform.iOS;
-
-  if (isMacOrIos && !HardwareKeyboard.instance.isMetaPressed) {
+  if (isApple && !HardwareKeyboard.instance.isMetaPressed) {
     return ExecutionInstruction.continueExecution;
   }
 
-  if (!isMacOrIos && !HardwareKeyboard.instance.isControlPressed) {
+  if (!isApple && !HardwareKeyboard.instance.isControlPressed) {
     return ExecutionInstruction.continueExecution;
   }
 
@@ -114,13 +112,11 @@ ExecutionInstruction scrollOnCtrlOrCmdAndEndKeyPress({
     return ExecutionInstruction.continueExecution;
   }
 
-  final isMacOrIos = defaultTargetPlatform == TargetPlatform.macOS || defaultTargetPlatform == TargetPlatform.iOS;
-
-  if (isMacOrIos && !HardwareKeyboard.instance.isMetaPressed) {
+  if (isApple && !HardwareKeyboard.instance.isMetaPressed) {
     return ExecutionInstruction.continueExecution;
   }
 
-  if (!isMacOrIos && !HardwareKeyboard.instance.isControlPressed) {
+  if (!isApple && !HardwareKeyboard.instance.isControlPressed) {
     return ExecutionInstruction.continueExecution;
   }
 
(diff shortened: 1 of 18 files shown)


super_editor/lib/src/default_editor/document_hardware_keyboard/document_keyboard_actions.dart
```diff
@@ -221,7 +221,9 @@ ExecutionInstruction deleteDownstreamCharacterWithCtrlDeleteOnMac({
   required SuperEditorContext editContext,
   required KeyEvent keyEvent,
 }) {
-  if (defaultTargetPlatform != TargetPlatform.macOS) {
+  final isMacOrIos = defaultTargetPlatform == TargetPlatform.macOS || defaultTargetPlatform == TargetPlatform.iOS;
```
matthew-carroll, line 224: How about we create a synthesized property somewhere called `isApple`? Do we already have some kind of platform class that accumulates static things like that?

super_editor/test/super_editor/supereditor_copy_and_paste_test.dart
```diff
@@ -10,7 +9,7 @@ import 'supereditor_test_tools.dart';
 
 void main() {
   group('SuperEditor copy and paste > ', () {
-    testWidgetsOnMac('pastes within a paragraph', (tester) async {
+    testWidgetsOnMacAndIos('pastes within a paragraph', (tester) async {
```
matthew-carroll, line 12: We can call this `testWidgetsOnApple`

super_editor/lib/src/default_editor/document_hardware_keyboard/document_keyboard_actions.dart
angelosilvestre: Updated.

super_editor/test/super_editor/supereditor_copy_and_paste_test.dart
angelosilvestre: Updated.

super_editor/lib/src/default_editor/document_hardware_keyboard/document_keyboard_actions.dart
```diff
@@ -79,13 +79,11 @@ ExecutionInstruction scrollOnCtrlOrCmdAndHomeKeyPress({
     return ExecutionInstruction.continueExecution;
   }
 
-  final isMacOrIos = defaultTargetPlatform == TargetPlatform.macOS || defaultTargetPlatform == TargetPlatform.iOS;
-
-  if (isMacOrIos && !HardwareKeyboard.instance.isMetaPressed) {
+  if (isApple && !HardwareKeyboard.instance.isMetaPressed) {
```
matthew-carroll, line 82: I don't think we want this in the global namespace. In my previous comment I asked if we had an existing class for platform properties like this. Do we have a place for it or no?

super_editor/lib/src/infrastructure/keyboard.dart
```diff
@@ -34,8 +33,10 @@ enum ExecutionInstruction {
 
 extension PrimaryShortcutKey on KeyEvent {
   bool get isPrimaryShortcutKeyPressed =>
-      (defaultTargetPlatform == TargetPlatform.macOS && HardwareKeyboard.instance.isMetaPressed) ||
-      (defaultTargetPlatform != TargetPlatform.macOS && HardwareKeyboard.instance.isControlPressed);
+      (const [TargetPlatform.macOS, TargetPlatform.iOS].contains(defaultTargetPlatform) &&
```
matthew-carroll, line 36: These usages should be using the `isApple` check, too.

super_editor/lib/src/default_editor/document_hardware_keyboard/document_keyboard_actions.dart
angelosilvestre: I don't think we have a class for that.

angelosilvestre: Should we create a class for that?

matthew-carroll (approved):
LGTM

PR #1679, opened 2023-12-13
[SuperEditor] Don't execute keyboard handlers without primary focus (Resolves #183)

### LINKED ISSUE #183: [web] [bug] Pressing return when entering a hyperlink removes selected text
Using Chrome latest (90.0.4430.212).

Steps to reproduce:

- Start the sample editor (Editor Demo) for web
- Mark any word or sequence of characters
- Press the hyperlink icon
- Type in any link
- Hit return/enter

The marked word is removed.

Edit: this is working fine on MacOS.

angelosilvestre:
[SuperEditor] Don't execute keyboard handlers without primary focus. Resolves #183

In the demo app, pressing ENTER on the toolbar's URL field causes a new line to be inserted, instead of applying the link attribution.

The cause is that the editor keyboard handlers (`enterToInsertBlockNewline` in this case) are running even if the editor doesn't have primary focus. This prevents the key event to bubble up to the OS to generate an `onPerformAction` call.

This PR changes the shortcut system to avoid calling any keyboard handler if the editor doesn't have primary focus.


matthew-carroll (approved):
LGTM

PR #1682, opened 2023-12-13
[SuperEditor] Add constants for style metadata (Resolves #825)

### LINKED ISSUE #825: [SuperEditor] - Define constants for every stylesheet property
Currently, `StyleRule`s define style properties as key/value pairs, and the keys are hardcoded strings.

Those hardcoded strings should be moved to constants so that developers can easily determine what's supported, and avoid typos.

This will be a breaking change.

angelosilvestre:
[SuperEditor] Add constants for style metadata. Resolves #825

Currently, we use hardcoded strings for the stylesheet properties.

This PR adds constants to the available properties, so developers can use intelisense to discover them.

We could also add constants for node's metadata.

### DIFF
--- a/super_editor/README.md
+++ b/super_editor/README.md
@@ -125,7 +125,7 @@ class _MyAppState extends State<MyApp> {
                             }
 
                             return {
-                                "padding": const CascadingPadding.only(top: 24),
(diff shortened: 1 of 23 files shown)


matthew-carroll (approved):
LGTM

PR #1688, opened 2023-12-14
[SuperEditor] Make composing region underline continuous (Resolves #1684)

### LINKED ISSUE #1684: Underline for composing region is not continuous when it includes Japaneses and Latin characters
**Package Version**
Using https://github.com/superlistapp/super_editor/commit/0a58e704469ccaffe5ab14620a56afe75bee6574 in main.

**To Reproduce**
1. Go to SuperEditor demo
2. Use Japanese-Romanji keyboad layout
3. Place the selection in the text (not specific position needed)
4. In the physical English keyboard, type "as"

**Actual behavior**
The content will be "あs", but the line below those 2 characters is not aligned. See video below.

**Expected behavior**
Example from the same characters in slack:
Screenshot 2023-12-13 at 19 59 30

**Platform**
MacOS.

**Flutter version**

**Screenshots**

https://github.com/superlistapp/super_editor/assets/31278849/970c3745-667b-424f-8bfa-10c65d7cbf5b

angelosilvestre:
[SuperEditor] Make composing region underline continuous. Resolves #1684

Steps To Reproduce

1. Go to SuperEditor demo
2. Use Japanese-Romanji keyboad layout
3. Place the selection in the text (not specific position needed)
4. In the physical English keyboard, type "as"

With these steps the underlines aren't aligned:

tight

We use `getBoxesForSelection` to determine the position of the underline, which internally calls a `RenderParagraph` method of the same name. This method takes a `boxHeightStyle` parameter, which the default value (`BoxHeightStyle.tight`) has the following doc:

```dart
/// Provide tight bounding boxes that fit heights per run. This style may result
/// in uneven bounding boxes that do not nicely connect with adjacent boxes.
```

This is how it's rendered with the other styles:

`BoxHeightStyle.includeLineSpacingBottom`
includeLineSpacingBottom

`BoxHeightStyle.includeLineSpacingMiddle`
includeLineSpacingMiddle

`BoxHeightStyle.includeLineSpacingTop`
includeLineSpacingTop

`BoxHeightStyle.max`
max

`BoxHeightStyle.strut`
strut

Modifying `TextUnderlineLayer` to use `BoxHeightStyle.max` as the `boxHeightStyle` solves the alignment issue. However, it's noticeable that the gap between the text and the underline becomes much bigger, even if we set the `TextLayoutUnderline` gap to zero:

zero_gap

This PR exposes the `boxHeightStyle` and `boxWidthStyle` parameters in  the `TextLayout` interface and changes the `TextUnderlineLayer` to use `BoxHeightStyle.max`.

To make the gap smaller, we could manually expand the selection boxes, using the bottom of the first box of each line:

```dart
final selectionBoxes = widget.textLayout.getBoxesForSelection(
  TextSelection(baseOffset: underline.range.start, extentOffset: underline.range.end),
);

for (int i = 0; i < selectionBoxes.length; i++) {
  final box = selectionBoxes[i];

  if (paintBoxes.isEmpty) {
    // This is the first box, just add it.
    paintBoxes.add(box);
  } else if (box.top >= paintBoxes.last.bottom) {
    // This box belong to another line.
    // This is a naive check and we should be able to find a better way to do it.
    paintBoxes.add(box);
  } else {
    final lastBox = paintBoxes.last;
    paintBoxes[paintBoxes.length - 1] = TextBox.fromLTRBD(
      lastBox.left,
      lastBox.top,
      box.right, // extend here
      lastBox.bottom,
      lastBox.direction,
    );
  }
}
```

With this approach we get a much smaller gap:

Captura de Tela 2023-12-14 às 11 44 37

This PR doesn't include golden tests because I couldn't manage to render the Japanese correctly in the tests:
Captura de Tela 2023-12-14 às 11 36 37

### DIFF
--- a/super_editor/test/super_textfield/attributed_text_editing_controller_test.dart
+++ b/super_editor/test/super_textfield/attributed_text_editing_controller_test.dart
@@ -886,7 +886,11 @@ class _NoOpTextLayout implements ProseTextLayout {
   }
 
   @override
-  List<TextBox> getBoxesForSelection(TextSelection selection) {
+  List<TextBox> getBoxesForSelection(
+    TextSelection selection, {
+    BoxHeightStyle boxHeightStyle = BoxHeightStyle.tight,
+    BoxWidthStyle boxWidthStyle = BoxWidthStyle.tight,
+  }) {
     throw UnimplementedError();
   }
 
--- a/super_text_layout/lib/src/text_layout.dart
+++ b/super_text_layout/lib/src/text_layout.dart
@@ -1,3 +1,4 @@
+import 'dart:ui';
 import 'package:flutter/widgets.dart';
 
 import 'super_text.dart';
@@ -47,7 +48,11 @@ abstract class TextLayout {
   double? getHeightForCaret(TextPosition position);
 
   /// Returns a [List] of [TextBox]es that contain the given [selection].
-  List<TextBox> getBoxesForSelection(TextSelection selection);
+  List<TextBox> getBoxesForSelection(
+    TextSelection selection, {
+    BoxHeightStyle boxHeightStyle = BoxHeightStyle.tight,
+    BoxWidthStyle boxWidthStyle = BoxWidthStyle.tight,
+  });
 
   /// Returns a bounding [TextBox] for the character at the given [position] or `null`
   /// if a character box couldn't be found.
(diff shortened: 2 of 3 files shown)


angelosilvestre:
@matthew-carroll @miguelcmedeiros What's your thoughts on this?

matthew-carroll (approved):
LGTM

matthew-carroll:
> To make the gap smaller, we could manually expand the selection boxes, using the bottom of the first box of each line:

I don't think there's any guarantee that the first character in a line is the appropriate selection for the height. What if the first character is tiny? Then the underline will likely run into the following text.

I think the spacing in your screenshot images looks fine to me. I'll merge this and we can adjust later as needed.

PR #1693, opened 2023-12-14
[SuperEditor][SuperTextField] Clear composing region after selection changes (Resolves #1685)

### LINKED ISSUE #1685: Japanese IME panel does not close when the selection moves somewhere else
**Package Version**
Using https://github.com/superlistapp/super_editor/commit/0a58e704469ccaffe5ab14620a56afe75bee6574 in main.

**To Reproduce**
1. Go to SuperEditor demo
2. Use Japanese-Romanji (reproducible with Japanese-Kana) keyboad layout
3. Place the selection in the text (not specific position needed)
4. Type some text in the English keyboard like "hey" (IME panel is shown to the user)
5. Tap somewhere else in the document to move the selection

**Actual behavior**
Selection moves somewhere else, but the IME panel remains open and the composing region is still underlined.

**Expected behavior**
Selection moves somewhere else, IME panel closes and the composing region is not longer underlined.

**Platform**
MacOS.

**Flutter version**

**Screenshots**

https://github.com/superlistapp/super_editor/assets/31278849/4c11981d-bbda-4b50-a7b4-897f8f2a45e2

**Additional context**
Reproducible with `SuperEditor` and `SuperTextField`.

angelosilvestre:
[SuperEditor][SuperTextField] Clear composing region after selection changes. Resolves #1685

When an IME popover is visible, like the Japanese character picker, tapping somewhere else in the editor doesn't close the IME popover.

The reason is that we change the selection, but keep the composing region. To solve that, we need to clear the composing region after the selection changes.

This PR changes the places where we change the selection to also issue a `ChangeComposingRegionRequest`. `SuperTextField` was also modified to clear the composing region.

Initially I added the `ChangeComposingRegionRequest` in the same batch as the selection change request. This caused 
some tests to fail because `AdjustSelectionAroundTagReaction` expects a selection change as the only change in the changelist:

```dart
if (changeList.length > 1 || changeList.first is! SelectionChangeEvent) {
  // We only want to move the caret when we're confident about what changed. Therefore,
  // we only react to changes that are solely a selection change, i.e., we ignore
  // situations like text entry, text deletion, etc.
  editorStableTagsLog.info(" - change list isn't just a single SelectionChangeEvent: $changeList");
  return;
}
```

I changed the implementation to issue the `ChangeComposingRegionRequest` in a separate request, but I think it would probably be better if the change `AdjustSelectionAroundTagReaction` to handle the case where we have a selection change and a composing region change in the list.


matthew-carroll:
> I changed the implementation to issue the ChangeComposingRegionRequest in a separate request, but I think it would probably be better if the change AdjustSelectionAroundTagReaction to handle the case where we have a selection change and a composing region change in the list.

That's correct. Please try to make the change to the reaction.

angelosilvestre:
@matthew-carroll Updated.

angelosilvestre:
@matthew-carroll Updated a golden from super_text_layout that was failing due to the previous PR.

super_editor/lib/src/default_editor/document_gestures_mouse.dart
```diff
@@ -695,6 +699,8 @@ Updating drag selection:
     editorGesturesLog.fine("Clearing document selection");
     widget.editor.execute([
       const ClearSelectionRequest(),
+      // Clear the composing region to close any IME popovers.
+      ChangeComposingRegionRequest(null),
```
matthew-carroll, line 703: If we have a `ClearSelectionRequest` then should we also have a `ClearComposingRegionRequest`?

Also, I think repeating the comment about closing IME popovers is probably unnecessary. The reason we're clearing the composing region is because the previous region no longer applies, and should be cleared. It just so happens that underlining the composing region showed that we weren't maintaining the composing region correctly.

super_editor/test/super_editor/supereditor_input_ime_test.dart
```diff
@@ -1170,6 +1170,70 @@ Paragraph two
         expect(keyboardAppearance, 'Brightness.dark');
       });
     });
+
+    testWidgetsOnAllPlatforms('clears composing region after selection changes', (tester) async {
```
matthew-carroll, line 1174: Wasn't the composing region always clearing when selecting somewhere else in the same node? I thought we only observed the composing region staying around when selecting content in a different node?

super_editor/lib/src/default_editor/document_gestures_mouse.dart
angelosilvestre: Yeah, it makes sense to have a `ClearComposingRegionRequest`.

super_editor/test/super_editor/supereditor_input_ime_test.dart
angelosilvestre: Update the test to select another node.

matthew-carroll: Can you speak to my question though? Was this ever a problem for moving the selection within a single node? Or was this only a problem when moving between nodes?

angelosilvestre: When moving the selection in the same node, the composing region was changing to the selected word.

super_editor/lib/src/core/document_composer.dart
```diff
@@ -475,6 +475,10 @@ class ChangeComposingRegionRequest implements EditRequest {
   int get hashCode => composingRegion.hashCode;
 }
 
+class ClearComposingRegionRequest implements EditRequest {
```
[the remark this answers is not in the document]
angelosilvestre: Done.

matthew-carroll (approved):
LGTM

PR #1699, opened 2023-12-21
[SuperEditor][Android] Fix drag handle flashing on double tap (Resolves #1690)

### LINKED ISSUE #1690: [SuperEditor] [Android] Selection handle is painted on the top/left corner of editor for a few frames after double-tap on a word
**Package Version**
Using 0a58e704469ccaffe5ab14620a56afe75bee6574 in `main`.

**To Reproduce**
1. Go to SuperEditor demos
2. Double-tap on a word to select it

**Actual behavior**
The handles surrounding the selection are painted, but a handle is also painted at the top/left corner of the document for a few frames. 

**Expected behavior**
Only the handles surrounding the selection are painted.

**Platform**
Android.

**Flutter version**
Flutter `master` commit e278279a48f0479386c249cf18f26b29c6f9018f.

**Screenshots**

https://github.com/superlistapp/super_editor/assets/31278849/92af387f-a242-48e0-a1f0-0995851b280e

angelosilvestre:
[SuperEditor][Android] Fix drag handle flashing on double tap. Resolves #1690

Upon double tap, the drag handle quickly flashes at the top-left corner of the screen:

https://github.com/superlistapp/super_editor/assets/31278849/92af387f-a242-48e0-a1f0-0995851b280e

This happens because of the following:

- The `TapSequenceGestureRecognizer` reports both `onTapDown` and `onDoubleTapDown` (as expected), so we first show the collapsed drag handle, then we hide the collapsed drag handle and show the expanded handles.
- The `Follower` is configured to `showWhenUnlinked` (because it defaults to `true`), so when the `Leader` is unlinked, it's placed at the top-left corner. 
- The collapsed handle uses a `AnimatedOpacity` to fade in and out.

So, on double tap, the drag handle is positioned at the top-left corner and it's visible for a brief moment, while the animation is running.

This PR sets `showWhenUnlinked` to false, so when the `Leader` is unlinked, the drag handle is immediately hidden.

Another option is to keep the `Leader` for the collapsed drag handle on the widget tree even when we have an expanded selection.

### DIFF
--- a/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
@@ -1531,7 +1531,7 @@ class SuperEditorAndroidControlsOverlayManagerState extends State<SuperEditorAnd
       valueListenable: _controlsController!.shouldShowCollapsedHandle,
       builder: (context, shouldShow, child) {
         final selection = widget.selection.value;
-        if (selection == null || selection.isCollapsed) {
+        if (selection == null || !selection.isCollapsed) {
           // When the user double taps we first place a collapsed selection
           // and then an expanded selection.
           // Return a SizedBox to avoid flashing the collapsed drag handle.
--- a/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
@@ -1,4 +1,5 @@
 import 'dart:async';
+import 'dart:ffi';
 
 import 'package:flutter/foundation.dart';
 import 'package:flutter/gestures.dart';
@@ -1529,13 +1530,20 @@ class SuperEditorAndroidControlsOverlayManagerState extends State<SuperEditorAnd
     return ValueListenableBuilder(
       valueListenable: _controlsController!.shouldShowCollapsedHandle,
       builder: (context, shouldShow, child) {
+        final selection = widget.selection.value;
+        if (selection == null || selection.isCollapsed) {
+          // When the user double taps we first place a collapsed selection
+          // and then an expanded selection.
+          // Return a SizedBox to avoid flashing the collapsed drag handle.
+          return const SizedBox();
+        }
+
         // Note: If we pass this widget as the `child` property, it causes repeated starts and stops
         // of the pan gesture. By building it here, pan events work as expected.
         return Follower.withOffset(
           link: _controlsController!.collapsedHandleFocalPoint,
           leaderAnchor: Alignment.bottomCenter,
           followerAnchor: Alignment.topCenter,
-          showWhenUnlinked: false,
           child: AnimatedOpacity(
             // When the controller doesn't want the handle to be visible, hide it.
             opacity: shouldShow ? 1.0 : 0.0,
--- a/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
@@ -1535,6 +1535,7 @@ class SuperEditorAndroidControlsOverlayManagerState extends State<SuperEditorAnd
           link: _controlsController!.collapsedHandleFocalPoint,
           leaderAnchor: Alignment.bottomCenter,
           followerAnchor: Alignment.topCenter,
+          showWhenUnlinked: false,
           child: AnimatedOpacity(
             // When the controller doesn't want the handle to be visible, hide it.
             opacity: shouldShow ? 1.0 : 0.0,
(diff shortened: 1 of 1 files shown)


angelosilvestre:
@miguelcmedeiros Could you please try this PR?

matthew-carroll:
Based on your description of the problem, shouldn't we move the selection change into `onTapUp` and `onDoubleTapUp`? It sounds like we have a temporary incorrect selection state, but we're trying to hide it with rendering changes.

angelosilvestre:
@matthew-carroll Sorry, the collapsed selection is already being performed at `_onTapUp`, only the expanded selection is performed on `_onDoubleTapDown`. The flash still occurs if we move the expanded selection change to `onDoubleTapUp`

matthew-carroll:
> Sorry, the collapsed selection is already being performed at _onTapUp

Then why is this problem happening? Does `_onTapUp` run even when a double tap happens?

angelosilvestre:
> Then why is this problem happening? Does _onTapUp run even when a double tap happens?

Yes, onTapUp also runs for a double tap. I think that's expected based on the docs for the `TapSequenceGestureRecognizer`.

matthew-carroll:
> Yes, onTapUp also runs for a double tap.

Ok, understood.

Why do we find ourselves in a state where the `Leader` is unlinked, i.e., there's no `Leader` but there is a `Follower`? Where in the sequence of events does that happen?

angelosilvestre:
> > Yes, onTapUp also runs for a double tap.
> 
> Ok, understood.
> 
> Why do we find ourselves in a state where the `Leader` is unlinked, i.e., there's no `Leader` but there is a `Follower`? Where in the sequence of events does that happen?

@matthew-carroll  Because we always add a `Follower` to the tree, even when we have an expanded selection. We could add this "guard" clause:


```dart
if (widget.selection.value?.isCollapsed == false) {
  return const SizedBox();
}

// return Follower
```

It will also fix the issue.

matthew-carroll:
@angelosilvestre can you provide more of a snippet? I can't tell what part of the code you're trying to show me.

angelosilvestre:
@matthew-carroll Currently we have:

```dart
return ValueListenableBuilder(
  valueListenable: _controlsController!.shouldShowCollapsedHandle,
  builder: (context, shouldShow, child) {
    // Note: If we pass this widget as the `child` property, it causes repeated starts and stops
    // of the pan gesture. By building it here, pan events work as expected.
    return Follower.withOffset(
      link: _controlsController!.collapsedHandleFocalPoint,
      leaderAnchor: Alignment.bottomCenter,
      followerAnchor: Alignment.topCenter,
      child: AnimatedOpacity(
        // rest of the tree
```

The other change I suggested is to make it like this:

```dart
return ValueListenableBuilder(
  valueListenable: _controlsController!.shouldShowCollapsedHandle,
  builder: (context, shouldShow, child) {
    if (widget.selection.value?.isCollapsed == false) {
     return const SizedBox();
    }

    // Note: If we pass this widget as the `child` property, it causes repeated starts and stops
    // of the pan gesture. By building it here, pan events work as expected.
    return Follower.withOffset(
      link: _controlsController!.collapsedHandleFocalPoint,
      leaderAnchor: Alignment.bottomCenter,
      followerAnchor: Alignment.topCenter,
      child: AnimatedOpacity(
        // rest of the tree
```

matthew-carroll:
Instead of `if (widget.selection.value?.isCollapsed == false)` can you just say `if (!shouldShow)`?

angelosilvestre:
@matthew-carroll Then we would lose the fade-in animation when the drag handle disappears after the timeout, right?

matthew-carroll:
Yeah, that seems to be the case. In terms of checking for a collapsed selection, I noticed that we'd pass by your if-statement if the selection is `null`. Is that intended? Should the if-statement be inverted so that we also return a `SizedBox` when the selection is `null`, too?

angelosilvestre:
Yeah, probably.

matthew-carroll:
Ok, then let's go with that. Please also add a comment within the if-statement that explains why we care about returning a `SizedBox` under those conditions.

matthew-carroll:
Looks like tests are failing

angelosilvestre:
@matthew-carroll Updated

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
```diff
@@ -1,4 +1,5 @@
 import 'dart:async';
+import 'dart:ffi';
```
matthew-carroll, line 2: Should this import be here?

angelosilvestre: Removed.

matthew-carroll (approved):
LGTM

PR #1701, opened 2023-12-21
Move text entry test back to its original location after RawKeyEvent migration (Resolves #1700)

### LINKED ISSUE #1700: [SuperEditor] - Move text entry test back to where it came from
In a recent PR, Super Editor was migrated from `RawKeyEvent` to `KeyEvent`. This included the movement of a test to a new file because that test required the use of a fake services binding, and that fake services binding can't be reset within a given test file.

https://github.com/superlistapp/super_editor/pull/1593/files#diff-e0c877e428c0aa3cfc648b6272498fd8cc2dbc321d8e5ab1778b52afd67eacc3

We don't want our test files dictated by such a low level detail. Move the fake binding into our test tools, make it possible to turn it on and off, and then move the test back to where it came from.

matthew-carroll:
Move text entry test back to its original location after RawKeyEvent migration (Resolves #1700)


matthew-carroll:
@angelosilvestre this is just an FYI. Shouldn't need a real review. Just moving some things around. I'm gonna merge when the tests pass.

PR #1712, opened 2023-12-27
[SuperEditor] Add em-dash conversion reaction (Resolves #1177)

### LINKED ISSUE #1177: [SuperEditor] Support for em dash (—)
When the user types 2 dashes ("--") it should convert to a an em dash ("—").
This should work as well when the user types 3 dashes at the beginning of a paragraph. The first 2 dashes would first be converted to an em dash and then the third dash would convert the paragraph into an horizontal rule.

angelosilvestre:
[SuperEditor] Addem-dash conversion reaction. Resolves #1177 

This PR adds a reaction to convert 2 dashes ("--") to an em dash ("—"). 

It also allows typing three dashes at the beginning of the paragraph.  In this case, the first two dashes are converted to an em dash and after typing the third dash a horizontal rule is added before the current paragraph and the dashes are removed.

https://github.com/superlistapp/super_editor/assets/7597082/e69106be-9f26-423b-8290-7e799b7f0772

### DIFF
--- a/super_editor/lib/src/default_editor/default_document_editor_reactions.dart
+++ b/super_editor/lib/src/default_editor/default_document_editor_reactions.dart
@@ -67,7 +67,7 @@ class HeaderConversionReaction extends ParagraphPrefixConversionReaction {
     EditContext editContext,
     RequestDispatcher requestDispatcher,
     List<EditEvent> changeList,
-    ParagraphNode paragraph,
+    TextNode paragraph,
     String match,
   ) {
     final prefixLength = match.length - 1; // -1 for the space on the end
@@ -130,7 +130,7 @@ class UnorderedListItemConversionReaction extends ParagraphPrefixConversionReact
     EditContext editContext,
     RequestDispatcher requestDispatcher,
     List<EditEvent> changeList,
-    ParagraphNode paragraph,
+    TextNode paragraph,
     String match,
   ) {
     // The user started a paragraph with an unordered list item pattern.
@@ -172,7 +172,7 @@ class OrderedListItemConversionReaction extends ParagraphPrefixConversionReactio
     EditContext editContext,
     RequestDispatcher requestDispatcher,
     List<EditEvent> changeList,
-    ParagraphNode paragraph,
+    TextNode paragraph,
     String match,
   ) {
     // The user started a paragraph with an ordered list item pattern.
@@ -214,7 +214,7 @@ class BlockquoteConversionReaction extends ParagraphPrefixConversionReaction {
     EditContext editContext,
     RequestDispatcher requestDispatcher,
     List<EditEvent> changeList,
-    ParagraphNode paragraph,
+    TextNode paragraph,
     String match,
   ) {
     // The user started a paragraph with blockquote pattern.
@@ -321,9 +321,9 @@ class HorizontalRuleConversionReaction implements EditReaction {
 }
 
 /// Base class for [EditReaction]s that want to take action when the user types text at
-/// the beginning of a [TextNode], which matches a given [RegExp].
-abstract class TextNodePrefixConversionReaction implements EditReaction {
-  const TextNodePrefixConversionReaction({
+/// the beginning of a paragraph, which matches a given [RegExp].
+abstract class ParagraphPrefixConversionReaction implements EditReaction {
+  const ParagraphPrefixConversionReaction({
     bool requireSpaceInsertion = true,
   }) : _requireSpaceInsertion = requireSpaceInsertion;
 
@@ -334,7 +334,7 @@ abstract class TextNodePrefixConversionReaction implements EditReaction {
   /// user didn't insert a space into the paragraph.
   final bool _requireSpaceInsertion;
 
-  /// Pattern that is matched at the beginning of a node and then passed to
+  /// Pattern that is matched at the beginning of a paragraph and then passed to
   /// sub-classes for processing.
   RegExp get pattern;
 
@@ -348,61 +348,25 @@ abstract class TextNodePrefixConversionReaction implements EditReaction {
 
     final edit = changeList[changeList.length - 2] as DocumentEdit;
     final textInsertionEvent = edit.change as TextInsertionEvent;
-
-    final textNode = document.getNodeById(textInsertionEvent.nodeId) as TextNode;
-    if (!acceptsNode(textNode)) {
-      // The subclass doesn't accept this type of node. Fizzle.
-      return;
-    }
-
-    final match = pattern.firstMatch(textNode.text.text)?.group(0);
+    final paragraph = document.getNodeById(textInsertionEvent.nodeId) as TextNode;
+    final match = pattern.firstMatch(paragraph.text.text)?.group(0);
     if (match == null) {
       return;
     }
 
-    // The user started a node with the desired pattern. Delegate to the subclass
+    // The user started a paragraph with the desired pattern. Delegate to the subclass
     // to do whatever it wants.
-    onPrefixMatched(editContext, requestDispatcher, changeList, textNode, match);
+    onPrefixMatched(editContext, requestDispatcher, changeList, paragraph, match);
   }
-
-  /// Hook called by the superclass, before evaluating the [pattern], to check if a
-  /// given [node] is acceptable for this reaction.
-  ///
-  /// For example, a reaction that only runs for [ParagraphNode]s should return `true`
-  /// only if [node] is a [ParagraphNode].
-  @protected
-  bool acceptsNode(TextNode node);
-
-  /// Hook, called by the superclass, when the user starts the given [node] with
-  /// the given [match], which fits the desired [pattern].
-  @protected
-  void onPrefixMatched(
-    EditContext editContext,
-    RequestDispatcher requestDispatcher,
-    List<EditEvent> changeList,
-    TextNode node,
-    String match,
-  );
-}
-
-/// A [TextNodePrefixConversionReaction] that only runs for [ParagraphNode]s.
-///
-/// This reaction runs when the user types text at the beginning of a [ParagraphNode],
-/// which matches a given [RegExp].
-abstract class ParagraphPrefixConversionReaction extends TextNodePrefixConversionReaction {
-  const ParagraphPrefixConversionReaction();
-
-  @override
-  bool acceptsNode(TextNode node) => node is ParagraphNode;
 
   /// Hook, called by the superclass, when the user starts the given [paragraph] with
   /// the given [match], which fits the desired [pattern].
-  @override
+  @protected
   void onPrefixMatched(
     EditContext editContext,
     RequestDispatcher requestDispatcher,
     List<EditEvent> changeList,
-    covariant ParagraphNode paragraph,
+    TextNode paragraph,
     String match,
   );
 }
--- a/super_editor/lib/src/default_editor/default_document_editor_reactions.dart
+++ b/super_editor/lib/src/default_editor/default_document_editor_reactions.dart
@@ -67,7 +67,7 @@ class HeaderConversionReaction extends ParagraphPrefixConversionReaction {
     EditContext editContext,
     RequestDispatcher requestDispatcher,
     List<EditEvent> changeList,
-    TextNode paragraph,
+    ParagraphNode paragraph,
     String match,
   ) {
     final prefixLength = match.length - 1; // -1 for the space on the end
@@ -130,7 +130,7 @@ class UnorderedListItemConversionReaction extends ParagraphPrefixConversionReact
     EditContext editContext,
     RequestDispatcher requestDispatcher,
     List<EditEvent> changeList,
-    TextNode paragraph,
+    ParagraphNode paragraph,
     String match,
   ) {
     // The user started a paragraph with an unordered list item pattern.
@@ -172,7 +172,7 @@ class OrderedListItemConversionReaction extends ParagraphPrefixConversionReactio
     EditContext editContext,
     RequestDispatcher requestDispatcher,
     List<EditEvent> changeList,
-    TextNode paragraph,
+    ParagraphNode paragraph,
     String match,
   ) {
     // The user started a paragraph with an ordered list item pattern.
@@ -214,7 +214,7 @@ class BlockquoteConversionReaction extends ParagraphPrefixConversionReaction {
     EditContext editContext,
     RequestDispatcher requestDispatcher,
     List<EditEvent> changeList,
-    TextNode paragraph,
+    ParagraphNode paragraph,
     String match,
   ) {
     // The user started a paragraph with blockquote pattern.
@@ -321,9 +321,9 @@ class HorizontalRuleConversionReaction implements EditReaction {
 }
 
 /// Base class for [EditReaction]s that want to take action when the user types text at
-/// the beginning of a paragraph, which matches a given [RegExp].
-abstract class ParagraphPrefixConversionReaction implements EditReaction {
-  const ParagraphPrefixConversionReaction({
+/// the beginning of a [TextNode], which matches a given [RegExp].
+abstract class TextNodePrefixConversionReaction implements EditReaction {
+  const TextNodePrefixConversionReaction({
     bool requireSpaceInsertion = true,
   }) : _requireSpaceInsertion = requireSpaceInsertion;
 
@@ -334,7 +334,7 @@ abstract class ParagraphPrefixConversionReaction implements EditReaction {
   /// user didn't insert a space into the paragraph.
   final bool _requireSpaceInsertion;
 
-  /// Pattern that is matched at the beginning of a paragraph and then passed to
+  /// Pattern that is matched at the beginning of a node and then passed to
   /// sub-classes for processing.
   RegExp get pattern;
 
@@ -348,25 +348,61 @@ abstract class ParagraphPrefixConversionReaction implements EditReaction {
 
     final edit = changeList[changeList.length - 2] as DocumentEdit;
     final textInsertionEvent = edit.change as TextInsertionEvent;
-    final paragraph = document.getNodeById(textInsertionEvent.nodeId) as TextNode;
-    final match = pattern.firstMatch(paragraph.text.text)?.group(0);
+
+    final textNode = document.getNodeById(textInsertionEvent.nodeId) as TextNode;
+    if (!acceptsNode(textNode)) {
+      // The subclass doesn't accept this type of node. Fizzle.
+      return;
+    }
+
+    final match = pattern.firstMatch(textNode.text.text)?.group(0);
     if (match == null) {
       return;
     }
 
-    // The user started a paragraph with the desired pattern. Delegate to the subclass
+    // The user started a node with the desired pattern. Delegate to the subclass
     // to do whatever it wants.
-    onPrefixMatched(editContext, requestDispatcher, changeList, paragraph, match);
+    onPrefixMatched(editContext, requestDispatcher, changeList, textNode, match);
   }
 
-  /// Hook, called by the superclass, when the user starts the given [paragraph] with
+  /// Hook called by the superclass, before evaluating the [pattern], to check if a
+  /// given [node] is acceptable for this reaction.
+  ///
+  /// For example, a reaction that only runs for [ParagraphNode]s should return `true`
+  /// only if [node] is a [ParagraphNode].
+  @protected
+  bool acceptsNode(TextNode node);
+
+  /// Hook, called by the superclass, when the user starts the given [node] with
   /// the given [match], which fits the desired [pattern].
   @protected
   void onPrefixMatched(
     EditContext editContext,
     RequestDispatcher requestDispatcher,
     List<EditEvent> changeList,
-    TextNode paragraph,
+    TextNode node,
+    String match,
+  );
+}
+
+/// A [TextNodePrefixConversionReaction] that only runs for [ParagraphNode]s.
+///
+/// This reaction runs when the user types text at the beginning of a [ParagraphNode],
+/// which matches a given [RegExp].
+abstract class ParagraphPrefixConversionReaction extends TextNodePrefixConversionReaction {
+  const ParagraphPrefixConversionReaction();
+
+  @override
+  bool acceptsNode(TextNode node) => node is ParagraphNode;
+
+  /// Hook, called by the superclass, when the user starts the given [paragraph] with
+  /// the given [match], which fits the desired [pattern].
+  @override
+  void onPrefixMatched(
+    EditContext editContext,
+    RequestDispatcher requestDispatcher,
+    List<EditEvent> changeList,
+    covariant ParagraphNode paragraph,
     String match,
   );
 }
--- a/super_editor/lib/src/default_editor/default_document_editor_reactions.dart
+++ b/super_editor/lib/src/default_editor/default_document_editor_reactions.dart
@@ -67,7 +67,7 @@ class HeaderConversionReaction extends ParagraphPrefixConversionReaction {
     EditContext editContext,
     RequestDispatcher requestDispatcher,
     List<EditEvent> changeList,
-    TextNode paragraph,
+    ParagraphNode paragraph,
     String match,
   ) {
     final prefixLength = match.length - 1; // -1 for the space on the end
@@ -130,7 +130,7 @@ class UnorderedListItemConversionReaction extends ParagraphPrefixConversionReact
     EditContext editContext,
     RequestDispatcher requestDispatcher,
     List<EditEvent> changeList,
-    TextNode paragraph,
+    ParagraphNode paragraph,
     String match,
   ) {
     // The user started a paragraph with an unordered list item pattern.
@@ -172,7 +172,7 @@ class OrderedListItemConversionReaction extends ParagraphPrefixConversionReactio
     EditContext editContext,
     RequestDispatcher requestDispatcher,
     List<EditEvent> changeList,
-    TextNode paragraph,
+    ParagraphNode paragraph,
     String match,
   ) {
     // The user started a paragraph with an ordered list item pattern.
@@ -214,7 +214,7 @@ class BlockquoteConversionReaction extends ParagraphPrefixConversionReaction {
     EditContext editContext,
     RequestDispatcher requestDispatcher,
     List<EditEvent> changeList,
-    TextNode paragraph,
+    ParagraphNode paragraph,
     String match,
   ) {
     // The user started a paragraph with blockquote pattern.
@@ -250,7 +250,7 @@ class BlockquoteConversionReaction extends ParagraphPrefixConversionReaction {
 /// The horizontal rule is inserted before the current node and the remainder of
 /// the node's text is kept.
 ///
-/// Applied only to [ParagraphNode]s.
+/// Applied only to all [TextNode]s.
 class HorizontalRuleConversionReaction implements EditReaction {
   // Matches "---" or "—-" (an em-dash followed by a regular dash) at the beginning of a line,
   // followed by a space.
@@ -348,7 +348,10 @@ abstract class ParagraphPrefixConversionReaction implements EditReaction {
 
     final edit = changeList[changeList.length - 2] as DocumentEdit;
     final textInsertionEvent = edit.change as TextInsertionEvent;
-    final paragraph = document.getNodeById(textInsertionEvent.nodeId) as TextNode;
+    final paragraph = document.getNodeById(textInsertionEvent.nodeId);
+    if (paragraph is! ParagraphNode) {
+      return;
+    }
     final match = pattern.firstMatch(paragraph.text.text)?.group(0);
     if (match == null) {
       return;
@@ -366,7 +369,7 @@ abstract class ParagraphPrefixConversionReaction implements EditReaction {
     EditContext editContext,
     RequestDispatcher requestDispatcher,
     List<EditEvent> changeList,
-    TextNode paragraph,
+    ParagraphNode paragraph,
     String match,
   );
 }
--- a/super_editor/lib/src/default_editor/default_document_editor_reactions.dart
+++ b/super_editor/lib/src/default_editor/default_document_editor_reactions.dart
@@ -624,6 +624,186 @@ class LinkifyReaction implements EditReaction {
   }
 }
 
+/// An [EditReaction] which converts dashes to em-dashes and horizontal rules.
+///
+/// Performs the following conversions:
+///
+/// - Typing two dashes anywhere in a text node converts them to an em-dash.
+///
+/// - Typing three dashes at the beginning of a paragraph inserts a
+///   horizontal rule before the paragraph, removes the dashes, and places
+///   the caret at the beginning of the paragraph. The first two dashes
+///   are converted to an em-dash and the horizontal rule is inserted
+///   after the user types the third dash.
+class DashConversionReaction implements EditReaction {
+  const DashConversionReaction();
+
+  @override
+  void react(EditContext editorContext, RequestDispatcher requestDispatcher, List<EditEvent> changeList) {
+    final document = editorContext.find<MutableDocument>(Editor.documentKey);
+    final composer = editorContext.find<MutableDocumentComposer>(Editor.composerKey);
+
+    if (changeList.length != 2) {
+      // When the user types a characters, the changelist contains a text insertion
+      // and a selection change. The current changelist doesn't contain those changes.
+      // Ignore it.
+      return;
+    }
+
+    final edit =
+        changeList.firstWhereOrNull((e) => e is DocumentEdit && e.change is TextInsertionEvent) as DocumentEdit?;
+    if (edit == null) {
+      // The changelist doesn't contain a text insertion. Ignore it.
+      return;
+    }
+
+    final change = edit.change as TextInsertionEvent;
+
+    if (change.text.text != '-') {
+      // The only character that triggers the conversion is the dash.
+      // Ignore other characters.
+      return;
+    }
+
+    if (change.offset < 1) {
+      // The reaction needs at least two character and the inserted character is at
+      // the beginning of the paragraph.
+      return;
+    }
+
+    final node = document.getNodeById(change.nodeId) as TextNode;
+
+    final previousCharacter = node.text.text.substring(change.offset - 1, change.offset);
+    final textAfterInsertedCharacter = node.text.text.substring(change.offset);
+
+    if (!const [_emDash, '-'].contains(previousCharacter)) {
+      // The previous character isn't a dash or an em-dash.
+      // Skip any conversions.
+      return;
+    }
+
+    if (previousCharacter == _emDash) {
+      if (node is! ParagraphNode || change.offset != 1) {
+        // Conversions involving an em-dash followed by a regular dash only happen
+        // at the beginning of a paragraph. Other text nodes, e.g, list items, are ignored.
+        return;
+      }
+
+      // A dash was inserted after an em-dash at the beginning of a paragraph.
+      // The em-dash might be a result of the user typing two dashes.
+      // After typing the third dash we have "—-" (an em-dash followed by a regular dash).
+
+      if (textAfterInsertedCharacter.isNotEmpty) {
+        // A dash was inserted after an em-dash in a non-empty paragraph.
+        // - Remove the dashes.
+        // - Add a horizontal rule before the node.
+        // - Place the caret at the beginning of the paragraph.
+        // - Exit.
+        requestDispatcher.execute([
+          DeleteContentRequest(
+            documentRange: DocumentRange(
+              start: DocumentPosition(nodeId: node.id, nodePosition: const TextNodePosition(offset: 0)),
+              end: DocumentPosition(nodeId: node.id, nodePosition: const TextNodePosition(offset: 2)),
+            ),
+          ),
+          InsertNodeBeforeNodeRequest(
+            existingNodeId: node.id,
+            newNode: HorizontalRuleNode(
+              id: Editor.createNodeId(),
+            ),
+          ),
+          ChangeSelectionRequest(
+            DocumentSelection.collapsed(
+              position: DocumentPosition(
+                nodeId: node.id,
+                nodePosition: const TextNodePosition(offset: 0),
+              ),
+            ),
+            SelectionChangeType.placeCaret,
+            SelectionReason.contentChange,
+          ),
+        ]);
+
+        return;
+      }
+
+      // A dash was inserted after an em-dash in an empty paragraph.
+      // - Insert a horizontal rule before the paragraph.
+      // - Clear the paragraph's text.
+      // - Place the selection at the beginning of the paragraph.
+      // - Exit.
+      requestDispatcher.execute([
+        InsertNodeBeforeNodeRequest(
+          existingNodeId: node.id,
+          newNode: HorizontalRuleNode(
+            id: Editor.createNodeId(),
+          ),
+        ),
+        ReplaceNodeRequest(
+          existingNodeId: node.id,
+          newNode: ParagraphNode(
+            id: node.id,
+            text: AttributedText(),
+          ),
+        ),
+        ChangeSelectionRequest(
+          DocumentSelection.collapsed(
+            position: DocumentPosition(
+              nodeId: node.id,
+              nodePosition: const TextNodePosition(offset: 0),
+            ),
+          ),
+          SelectionChangeType.placeCaret,
+          SelectionReason.contentChange,
+        ),
+      ]);
+
+      return;
+    }
+
+    // A dash was inserted after another dash.
+    // Convert the two dashes to an em-dash.
+    requestDispatcher.execute([
+      DeleteContentRequest(
+        documentRange: DocumentRange(
+          start: DocumentPosition(nodeId: node.id, nodePosition: TextNodePosition(offset: change.offset - 1)),
+          end: DocumentPosition(nodeId: node.id, nodePosition: TextNodePosition(offset: change.offset + 1)),
+        ),
+      ),
+      ChangeSelectionRequest(
+        DocumentSelection.collapsed(
+          position: DocumentPosition(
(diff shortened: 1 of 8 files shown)


super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -624,6 +624,186 @@ class LinkifyReaction implements EditReaction {
   }
 }
 
+/// An [EditReaction] which converts dashes to em-dashes and horizontal rules.
```
matthew-carroll, line 627: Please include the literal examples in parentheses because not everyone knows what an "em dash" is, e.g., "which converts dash (--) to em-dashes (—)..."

Also, please include the scope of this reaction - does it apply to all `TextNode`s, or just `ParagraphNode`s?

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -624,6 +624,186 @@ class LinkifyReaction implements EditReaction {
...
+///
+/// Performs the following conversions:
+///
+/// - Typing two dashes anywhere in a text node converts them to an em-dash.
+///
+/// - Typing three dashes at the beginning of a paragraph inserts a
```
matthew-carroll, line 633: This description seems rather long for what it's describing. Is this paragraph simply trying to say that "---" still becomes a horizontal rule? If so, I think there's a more direct way to say it. However, this reaction also probably shouldn't know anything about horizontal rules, so I'll need to see what's happening below.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -624,6 +624,186 @@ class LinkifyReaction implements EditReaction {
...
+class DashConversionReaction implements EditReaction {
+  const DashConversionReaction();
+
+  @override
+  void react(EditContext editorContext, RequestDispatcher requestDispatcher, List<EditEvent> changeList) {
+    final document = editorContext.find<MutableDocument>(Editor.documentKey);
+    final composer = editorContext.find<MutableDocumentComposer>(Editor.composerKey);
+
+    if (changeList.length != 2) {
+      // When the user types a characters, the changelist contains a text insertion
```
matthew-carroll, line 647: This comment skips past its intention. This comment starts talking about a changelist containg a text insertion and a selection change, but it never mentions why we care about that.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -624,6 +624,186 @@ class LinkifyReaction implements EditReaction {
...
+    final change = edit.change as TextInsertionEvent;
+
+    if (change.text.text != '-') {
+      // The only character that triggers the conversion is the dash.
+      // Ignore other characters.
+      return;
+    }
+
+    if (change.offset < 1) {
+      // The reaction needs at least two character and the inserted character is at
```
matthew-carroll, line 669: This comment essentially says two unrelated things, rather than the details that it means to state.

I think what you intended to say was "The reaction needs at least two characters before the caret, but there's less than two. Fizzle."

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -624,6 +624,186 @@ class LinkifyReaction implements EditReaction {
...
+      // the beginning of the paragraph.
+      return;
+    }
+
+    final node = document.getNodeById(change.nodeId) as TextNode;
+
+    final previousCharacter = node.text.text.substring(change.offset - 1, change.offset);
+    final textAfterInsertedCharacter = node.text.text.substring(change.offset);
```
matthew-carroll, line 677: Is there a reason that you create this substring up here? It looks like there are two opportunities to exit early without creating this substring.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -624,6 +624,186 @@ class LinkifyReaction implements EditReaction {
...
+
+    if (!const [_emDash, '-'].contains(previousCharacter)) {
+      // The previous character isn't a dash or an em-dash.
+      // Skip any conversions.
+      return;
+    }
+
+    if (previousCharacter == _emDash) {
+      if (node is! ParagraphNode || change.offset != 1) {
+        // Conversions involving an em-dash followed by a regular dash only happen
```
matthew-carroll, line 687: This comment passively describes some general rule, but it doesn't clearly state the specific condition that this if-statement has identified, and why that specific situation isn't valid. Please try to phrase these comments about the situation that we're in, not about all the other situations that we're not in. For example, "Conversions involving an em-dash followed by a regular dash only happen at..." - this is a statement about all other situations, which is very confusing to read when we're trying to understand why this if-statement returns early.

Also, is there a reason that we're waiting until way down here to check if the node is a paragraph node? Up above it seems like we're ok with any TextNode but down here we seem to only want to deal with ParagraphNodes...

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -624,6 +624,186 @@ class LinkifyReaction implements EditReaction {
...
+        // at the beginning of a paragraph. Other text nodes, e.g, list items, are ignored.
+        return;
+      }
+
+      // A dash was inserted after an em-dash at the beginning of a paragraph.
```
matthew-carroll, line 692: Where is the dash-after-a-dash situation handled? I scrolled down - I guess it happens down at the very end. I have a few thoughts on what I'm seeing here. 

First, it's too difficult to visually track the code related to "—-" vs "--".  At a minimum, these two different scenarios should call out to functions to make it crystal clear that we're dealing with two possible recognitions and where each one happens.

That said, I don't think that this reaction should be responsible for creating horizontal rules. My initial thought is that the existing horizontal rule reaction should add "—-" to its existing "---" pattern.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -624,6 +624,186 @@ class LinkifyReaction implements EditReaction {
...
+    // A dash was inserted after another dash.
+    // Convert the two dashes to an em-dash.
+    requestDispatcher.execute([
+      DeleteContentRequest(
+        documentRange: DocumentRange(
+          start: DocumentPosition(nodeId: node.id, nodePosition: TextNodePosition(offset: change.offset - 1)),
+          end: DocumentPosition(nodeId: node.id, nodePosition: TextNodePosition(offset: change.offset + 1)),
+        ),
+      ),
+      ChangeSelectionRequest(
```
matthew-carroll, line 773: Do we need to two selection changes in a single transaction? What happens if you only include the final selection change?

super_editor/lib/src/test/super_editor_test/supereditor_robot.dart
```diff
@@ -316,6 +316,22 @@ extension SuperEditorRobot on WidgetTester {
     return gesture;
   }
 
+  /// Simulates typing [text], either as keyboard keys, or as insertion deltas of
+  /// a software keyboard.
+  ///
+  /// Provide an [imeOwnerFinder] if there are multiple [ImeOwner]s in the current
+  /// widget tree.
+  Future<void> typeTextAdaptive(String text, [Finder? imeOwnerFinder]) async {
```
matthew-carroll, line 324: Are you sure we don't have a method like this already? Including in the test robots package?

super_editor/test/super_editor/text_entry/dash_conversion_test.dart
```diff
@@ -0,0 +1,654 @@
...
+  group('SuperEditor dash conversion', () {
+    group('converts two dashes to an em dash', () {
+      testAllInputsOnAllPlatforms('at the beginning of an empty paragraph', (
+        tester, {
+        required TextInputSource inputSource,
+      }) async {
+        final context = await tester //
+            .createDocument()
+            .withSingleEmptyParagraph()
+            .withAddedReactions([const DashConversionReaction()]) //
```
matthew-carroll, line 29: Let's make this a standard reaction

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
angelosilvestre: I looked into the existing horizontal rule reaction and it requires a space to be inserted after the "---" to convert the paragraph. Should it also require a space when using the "—-" pattern?

matthew-carroll: Yeah, let's require a space. If we don't want that then we can change it later when someone complains.

angelosilvestre: There is another situation, the horizontal rule reaction (and the other prefix reactions) only perform the conversion in an empty paragraph, so it won't meet this requirement:

> Additionally, if the 3 dashes are typed at the start of a non-empty paragraph, then it could insert a horizontal rule above it, remove the dashes from the paragraph and place the selection at the start of the paragraph. This is not implemented on Superlist.

The `ParagraphPrefixConversionReaction` checks the following condition to perform the conversion:

```dart
final didTypeSpaceAtEnd = EditInspector.didTypeSpace(document, changeList);
if (_requireSpaceInsertion && !didTypeSpaceAtEnd) {
  return;
}
```

`EditInspector.didTypeSpace` has a guard clause that doesn't seem to belong to this method: 

```dart
final caretPosition = selectionEvent.newSelection!.extent.nodePosition as TextNodePosition;
final editedText = editedNode.text.text;
if (caretPosition.offset != editedText.length) {
  return false;
}
```

Although the method is called `didTypeSpace` it only returns `true` if the space was typed at the end of the paragraph.

We could move this condition from `didTypeSpace` to the `react` method, and add a `requireEmptyParagraph` constructor parameter to configure this behavior.

What do you think?

angelosilvestre: Or should we just change the horizontal rule reaction to avoid extending `ParagraphPrefixConversionReaction` ?

super_editor/lib/src/test/super_editor_test/supereditor_robot.dart
angelosilvestre: I didn't find any method like this. We only have `pressEnterAdaptive` and `pressNumpadEnterAdaptive`.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
angelosilvestre: It seems we don't need this intermediary selection change. Removed.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
matthew-carroll: @angelosilvestre I didn't follow your question. You started by talking about prefix reactions but then you switched to a description of some check for a space. I'm not sure what you were asking, but now that the horizontal rule reaction isn't just about prefixes, it probably shouldn't extend that prefix reaction base class.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
angelosilvestre: The conversion from "—-" to a horizontal rule was moved to the horizontal rule reaction, so this comment isn't applicable anymore.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
angelosilvestre: Removed because `textAfterInsertedCharacter` isn't needed anymore.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
angelosilvestre: The conversion from "—-" to a horizontal rule was moved to the horizontal rule reaction, so this comment isn't applicable anymore.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
angelosilvestre: Updated.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -243,41 +243,52 @@ class BlockquoteConversionReaction extends ParagraphPrefixConversionReaction {
   }
 }
 
-/// Converts full node content that looks like "--- " into a horizontal rule.
-class HorizontalRuleConversionReaction extends ParagraphPrefixConversionReaction {
-  static final _hrPattern = RegExp(r'^---\s$');
+/// Converts node content that looks like "--- " or "—- " into a horizontal rule.
```
matthew-carroll, line 246: At first glance I thought the "-- " was two dashes and got confused. But I assume that's an em-dash followed by a regular dash? If so, please add a parenthetical after that example which specifies that it's an em-dash followed by a regular dash.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -243,41 +243,52 @@ class BlockquoteConversionReaction extends ParagraphPrefixConversionReaction {
...
+///
+/// The horizontal rule is inserted before the current node and the remainder of
```
matthew-carroll, line 248: This description is missing a reference to what happens to the content that appears before the pattern.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -243,41 +243,52 @@ class BlockquoteConversionReaction extends ParagraphPrefixConversionReaction {
...
+/// the node's text is kept.
+///
+/// Applied only to [ParagraphNode]s.
+class HorizontalRuleConversionReaction implements EditReaction {
+  // Matches "---" or "—-" at the beginning of a line, followed by a space.
```
matthew-carroll, line 253: Here again, is that 2nd example an em-dash? If so, please note that in comment to avoid confusion.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -624,8 +635,92 @@ class LinkifyReaction implements EditReaction {
   }
 }
 
+/// An [EditReaction] which converts two dashes (--) to an em-dashes (—).
```
matthew-carroll, line 638: "em-dashes" -> "em-dash"

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -624,8 +635,92 @@ class LinkifyReaction implements EditReaction {
...
+    final document = editorContext.find<MutableDocument>(Editor.documentKey);
+    final composer = editorContext.find<MutableDocumentComposer>(Editor.composerKey);
+
+    if (changeList.length < 2) {
+      // If the user typed a dash, we should have at least an insertion and a
+      // selection change event.
+      return;
+    }
+
+    final selectionEvent = changeList[changeList.length - 1];
```
matthew-carroll, line 655: Can you say `changeList.last` instead of using the index?

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -624,8 +635,92 @@ class LinkifyReaction implements EditReaction {
...
+    if (selectionEvent is! SelectionChangeEvent) {
+      // If the user typed a dash, then the last event should be a selection change.
+      return;
+    }
+
+    final edit = changeList[changeList.length - 2];
```
matthew-carroll, line 661: I notice up above you say `selectionEvent` - shouldn't this be called `insertionEvent` to match?

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -624,8 +635,92 @@ class LinkifyReaction implements EditReaction {
...
+    if (edit is! DocumentEdit || edit.change is! TextInsertionEvent) {
+      // If the user typed a dash, then the second to last event should be a text
+      // insertion event with a dash "-".
+      return;
+    }
+
+    final change = edit.change as TextInsertionEvent;
+
+    if (change.text.text != '-') {
+      // The only character that triggers the conversion is the dash.
```
matthew-carroll, line 671: Please form comments in terms of what's happening at the point of the comment. For example, start with the fact that this if-statement was triggered because the inserted text wasn't a dash:

```
// The text that was inserted wasn't a dash. The only character that triggers a
// conversion is a dash. Fizzle.
```

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -624,8 +635,92 @@ class LinkifyReaction implements EditReaction {
...
+      // Ignore other characters.
+      return;
+    }
+
+    if (change.offset < 1) {
+      // The reaction needs at least two characters before the caret, but there's less than two.
+      // Fizzle.
```
matthew-carroll, line 678: I'm noticing here and in various other comments that there's a 2nd line that doesn't really need to be there. For example, in this case, " Fizzle." can be moved up to the line above at the end.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -624,8 +635,92 @@ class LinkifyReaction implements EditReaction {
...
+      return;
+    }
+
+    final node = document.getNodeById(change.nodeId) as TextNode;
+
+    final previousCharacter = node.text.text.substring(change.offset - 1, change.offset);
```
matthew-carroll, line 684: When describing direction, rather than time, please try to use "upstream" and "downstream" so that we don't get confused as to whether you might be talking about the character that existed before the edit.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -624,8 +635,92 @@ class LinkifyReaction implements EditReaction {
...
+        SelectionChangeType.placeCaret,
+        SelectionReason.contentChange,
+      ),
+    ]);
+  }
+}
+
 class EditInspector {
-  /// Whether [edits] ends with the user typing a space, i.e., typing a " ".
+  /// Whether [edits] ends with the user typing a space, i.e., typing a " " at the end of a node.
```
matthew-carroll, line 723: Is this truly specific to "at the end of a node"? Or does it find a space wherever the caret happens to be?

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -678,5 +773,51 @@ class EditInspector {
     return true;
   }
 
+  /// Whether [edits] ends with the user typing a space, i.e., typing a " ".
+  ///
+  /// Typing a space means that a space was inserted, and the caret moved from
+  /// just before the space, to just after the space.
+  static bool editEndsWithSpace(Document document, List<EditEvent> edits) {
```
matthew-carroll, line 780: Was this introduced because the other method only cares about the end of the node and this method cares about the middle of the node?

If so, we don't want methods that are so similar with such different names.

If that's the reason they're different, then let's rename the other one to `didTypeSpaceAtEndOfNode()` and let's name this one `didTypeSpace()`.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
angelosilvestre: The conversion only happens if the pattern is at the beginning of the paragraph. Updated the doc.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
angelosilvestre: > Was this introduced because the other method only cares about the end of the node and this method cares about the middle of the node?

Yes, updated.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -243,41 +243,54 @@ class BlockquoteConversionReaction extends ParagraphPrefixConversionReaction {
...
-  ) {
-    // The user started a paragraph with a horizontal rule pattern.
-    // Convert the paragraph to a horizontal rule.
-    final document = editContext.find<MutableDocument>(Editor.documentKey);
+    final didTypeSpace = EditInspector.didTypeSpace(document, changeList);
+    if (!didTypeSpace) {
+      return;
+    }
 
+    final edit = changeList[changeList.length - 2] as DocumentEdit;
```
matthew-carroll, line 269: How do we know there are at least 3 items in the list?

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -243,41 +243,54 @@ class BlockquoteConversionReaction extends ParagraphPrefixConversionReaction {
...
+    final textInsertionEvent = edit.change as TextInsertionEvent;
```
matthew-carroll, line 270: How do we know this is a `TextInsertionEvent`?

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -624,12 +637,95 @@ class LinkifyReaction implements EditReaction {
   }
 }
 
+/// An [EditReaction] which converts two dashes (--) to an em-dash (—).
+///
+/// This reaction applies to all [TextNode]s, anywhere in the text.
```
matthew-carroll, line 642: Let's replace "to all [TextNode]s, anywhere in the text." to "to all [TextNode]s in the document."

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -624,12 +637,95 @@ class LinkifyReaction implements EditReaction {
...
+class DashConversionReaction implements EditReaction {
+  const DashConversionReaction();
+
+  @override
+  void react(EditContext editorContext, RequestDispatcher requestDispatcher, List<EditEvent> changeList) {
+    final document = editorContext.find<MutableDocument>(Editor.documentKey);
+    final composer = editorContext.find<MutableDocumentComposer>(Editor.composerKey);
+
+    if (changeList.length < 2) {
+      // If the user typed a dash, we should have at least an insertion and a
```
matthew-carroll, line 652: Let's be explicit about why the if-statement means the given condition wasn't meant, e.g.,

```
// This reaction requires at least an insertion event and a selection change event.
// There are less than two events in the the change list, therefore this reaction
// shouldn't apply. Fizzle.
```

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -624,12 +637,95 @@ class LinkifyReaction implements EditReaction {
...
+      // selection change event.
+      return;
+    }
+
+    final selectionEvent = changeList.last;
+    if (selectionEvent is! SelectionChangeEvent) {
+      // If the user typed a dash, then the last event should be a selection change.
```
matthew-carroll, line 659: Here again, let's make it explicit how this if-statement contradicts the precondition.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -624,12 +637,95 @@ class LinkifyReaction implements EditReaction {
...
+      return;
+    }
+
+    final insertionEvent = changeList[changeList.length - 2];
+    if (insertionEvent is! DocumentEdit || insertionEvent.change is! TextInsertionEvent) {
+      // If the user typed a dash, then the second to last event should be a text
```
matthew-carroll, line 665: Please describe how the if-statement contradicts the precondition.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -624,12 +637,95 @@ class LinkifyReaction implements EditReaction {
...
+      // insertion event with a dash "-".
+      return;
+    }
+
+    final change = insertionEvent.change as TextInsertionEvent;
```
matthew-carroll, line 670: We know that there are at least two changes involved with this hcange list. Let's name this change more specifically.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -624,12 +637,95 @@ class LinkifyReaction implements EditReaction {
...
+      // conversion is a dash. Fizzle.
+      return;
+    }
+
+    if (change.offset < 1) {
+      // The reaction needs at least two characters before the caret, but there's less than two. Fizzle.
+      return;
+    }
+
+    final node = document.getNodeById(change.nodeId) as TextNode;
```
matthew-carroll, line 683: Let's name `node` more specifically.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -624,12 +637,95 @@ class LinkifyReaction implements EditReaction {
...
+
+    final upstreamCharacter = node.text.text.substring(change.offset - 1, change.offset);
```
matthew-carroll, line 685: If you want a single character, is there a reason that you don't use index-based access? `node.text.text[change.offset - 1]`

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -624,12 +637,95 @@ class LinkifyReaction implements EditReaction {
...
+        ),
+      ),
+      InsertTextRequest(
+        documentPosition: DocumentPosition(
+          nodeId: node.id,
+          nodePosition: TextNodePosition(
+            offset: change.offset - 1,
+          ),
+        ),
+        textToInsert: '—',
```
matthew-carroll, line 706: Do we have a constant for an em-dash? If not, we should probably introduce one so that we can clearly differentiate an em-dash from a regular dash.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -624,12 +637,95 @@ class LinkifyReaction implements EditReaction {
...
+        attributions: composer.preferences.currentAttributions,
+      ),
+      ChangeSelectionRequest(
+        DocumentSelection.collapsed(
+          position: DocumentPosition(
+            nodeId: node.id,
+            nodePosition: TextNodePosition(offset: change.offset),
+          ),
+        ),
+        SelectionChangeType.placeCaret,
```
matthew-carroll, line 716: I'm wondering if theres' a more appropriate selection change type description. Is there anything closer to fit this situation? The caret wasn't really "placed" in the sense that we typically mean that.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
matthew-carroll: The description of `didTypeSpaceAtEndOfNode()` and `didTypeSpace()` are still too similar and will create confusion. It looks like this first line should be:

```
/// Returns `true` if the given [edits] end with the user typing a space at the end of
/// a [TextNode], e.g., typing a " " at the end of a paragraph.
```

The first line for `didTypeSpace()` should probably be the following:

```
/// Returns `true` if the given [edits] end with the user typing a space anywhere
/// within a [TextNode], e.g., typing a " " between two words in a paragraph.
```

You can also remove the part that says "Typing a space means that a spacer..."

I would also move `didTypeSpaceAtEndOfNode()` below `didTypeSpace()`.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -678,5 +774,51 @@ class EditInspector {
     return true;
   }
 
+  /// Whether [edits] ends with the user typing a space, i.e., typing a " ".
+  ///
+  /// Typing a space means that a space was inserted, and the caret moved from
+  /// just before the space, to just after the space.
+  static bool didTypeSpace(Document document, List<EditEvent> edits) {
+    if (edits.length < 2) {
```
matthew-carroll, line 782: Please explain why.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -678,5 +774,51 @@ class EditInspector {
...
+    }
+    final textInsertionEvent = edit.change;
+    if (textInsertionEvent is! TextInsertionEvent) {
+      return false;
+    }
+    if (textInsertionEvent.text.text != " ") {
+      return false;
+    }
+
+    if (selectionEvent.oldSelection == null || selectionEvent.newSelection == null) {
```
matthew-carroll, line 806: Please explain why.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -678,5 +774,51 @@ class EditInspector {
...
+      return false;
+    }
+    if (selectionEvent.newSelection!.extent.nodeId != textInsertionEvent.nodeId) {
+      return false;
+    }
+
+    final editedNode = document.getNodeById(textInsertionEvent.nodeId)!;
+    // TODO: decide whether this inspection should be just for paragraphs or any text node
```
matthew-carroll, line 814: Let's go ahead and make this for all `TextNode`s and remove the TODO.

super_editor/test/super_editor/text_entry/dash_conversion_test.dart
```diff
@@ -0,0 +1,552 @@
...
+
+        // Ensure no conversion happened.
+        expect(SuperEditorInspector.findTextInParagraph('1').text, '-');
+
+        // Type the second dash.
+        await tester.typeTextAdaptive('-');
+
+        // Ensure the two dashes were converted to an em-dash.
+        expect(context.document.nodes.length, 1);
+        expect(SuperEditorInspector.findTextInParagraph('1').text, '—');
```
matthew-carroll, line 43: Let's use the new em-dash constant here, too.

super_editor/test/super_editor/text_entry/paragraph_conversions_test.dart
```diff
@@ -154,15 +155,18 @@ void main() {
     });
 
     group("paragraph to horizontal rule >", () {
-      testWidgetsOnAllPlatforms("with ---", (tester) async {
+      testAllInputsOnAllPlatforms("with --- in an empty paragraph", (
```
matthew-carroll, line 158: Given that it seems we're only converting prefixes, let's say "at the beginning of an empty paragraph"

Please make a similar change to other HR tests.

super_editor/test/super_editor/text_entry/paragraph_conversions_test.dart
```diff
@@ -349,7 +388,7 @@ final _orderedListVariant = ValueVariant({
 
 final _nonHrVariant = ValueVariant({
   // We ignore " - " because that is a conversion for unordered list items
-  "-- ",
-  "---- ",
-  " --- ",
+  "— ",
```
matthew-carroll, line 391: I'm curious why these changed? I'm not sure where these were being used, but I would assume that this PR only adds situations, rather than changing existing situations?

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
angelosilvestre: Updated

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
angelosilvestre: Same as before, we check in `EditInspector.didTypeSpace`.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
angelosilvestre: Updated.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
angelosilvestre: Updated.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
angelosilvestre: Updated.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
angelosilvestre: Updated.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
angelosilvestre: Nope, updated.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
angelosilvestre: Besides `placeCaret `, the only other option that seems to be related to this operation is `deleteContent`.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
angelosilvestre: I'm not entirely sure why this check exists in the existing method, but I assume that if the user typed a character, it's required that the selection moved from a position to another?

super_editor/test/super_editor/text_entry/dash_conversion_test.dart
angelosilvestre: I ended up removing the constant because it was only being used at two places. Should we re-introduce it? Should we create a class with static consts for characters?

angelosilvestre: I noticed I missed the previous comment about the constant.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
angelosilvestre: Added a `Characters` class, so we can add other constants if we need to.

angelosilvestre: `Characters` is already an extension name in dart. Do you have another suggestion? Or should we just use top level constants?

super_editor/test/super_editor/text_entry/paragraph_conversions_test.dart
angelosilvestre: They changed because now the two dashes are converted to em-dashes in all situations...

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
matthew-carroll: How about a `SpecialCharacter` class?

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
matthew-carroll: Yeah, let's check just in case. Eventually the right answer might be to have `didTypeSpace()` actually extract relevant information, but until then, let's make sure we don't access stuff that doesn't exist.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
matthew-carroll: Ok. Neither one is great. I guess we can leave it as-is.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
matthew-carroll: I can't see the surrounding code in this GitHub comment area, but it looks like the reaction wants to find a selection that moved, rather than a selection that just appeared. If you believe that it's unnecessary, given the already satisfied conditions, feel free to remove it.

super_editor/test/super_editor/text_entry/paragraph_conversions_test.dart
matthew-carroll: What if a user pastes "---" and then inserts a space "--- "?

angelosilvestre: It would still convert, these values are used in tests that type character by character and then check if the content wasn't converted.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
angelosilvestre: Yeah, it seems that the other conditions are sufficient. Removed.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
angelosilvestre: Looks good.

angelosilvestre:
@matthew-carroll Could you take a look again?

super_editor/test/super_editor/text_entry/paragraph_conversions_test.dart
matthew-carroll: Don't we still want to ensure that if the user were to type " ", "-", "-", "-", " ", that it results in an HR? In the meantime, it might also result in an em-dash, but from a user input perspective, the user typed regular dashes, right?

angelosilvestre: Yeah, but this strings are used to verify that the editor DON'T convert them to HRs. But I'm updating the test to keep the existing input.

angelosilvestre:
@matthew-carroll I noticed my last PR broke empty list items parsing (`* `) .

Added a fix for that in this PR.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -313,14 +341,14 @@ abstract class ParagraphPrefixConversionReaction implements EditReaction {
...
-    final didTypeSpaceAtEnd = EditInspector.didTypeSpace(document, changeList);
+    final didTypeSpaceAtEnd = EditInspector.didTypeSpaceAtEndOfNode(document, changeList);
     if (_requireSpaceInsertion && !didTypeSpaceAtEnd) {
       return;
     }
 
     final edit = changeList[changeList.length - 2] as DocumentEdit;
     final textInsertionEvent = edit.change as TextInsertionEvent;
-    final paragraph = document.getNodeById(textInsertionEvent.nodeId) as ParagraphNode;
+    final paragraph = document.getNodeById(textInsertionEvent.nodeId) as TextNode;
```
matthew-carroll, line 351: Here as well with the change from `ParagraphNode` to `TextNode` - should that change happen? The class name is `ParagraphPrefixConversionReaction`.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -624,13 +652,148 @@ class LinkifyReaction implements EditReaction {
   }
 }
 
+/// An [EditReaction] which converts two dashes (--) to an em-dash (—).
```
matthew-carroll, line 655: Can you be more specific about when it happens? Does this reaction check all text in every text node whenever any change happens to the text node? Does it only check after the user types a space? Let's make it clear when the reaction happens.

super_editor/lib/src/infrastructure/strings.dart
```diff
@@ -142,3 +142,8 @@ extension CharacterMovement on String {
     return range.current.length;
   }
 }
+
+/// Holds constants for special characters using within the package.
```
matthew-carroll, line 146: Let's say `Characters that are difficult to represent literally, and therefore are provided as constants.`

Also, please add a private constructor to class: `const SpecialCharacters._();`

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -66,7 +67,7 @@ class HeaderConversionReaction extends ParagraphPrefixConversionReaction {
     EditContext editContext,
     RequestDispatcher requestDispatcher,
     List<EditEvent> changeList,
-    ParagraphNode paragraph,
+    TextNode paragraph,
```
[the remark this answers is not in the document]
angelosilvestre: To be able to apply the horizontal rule conversion for all `TextNode`s, as requested in this comment https://github.com/superlistapp/super_editor/pull/1712#discussion_r1439954440, it requires changing the `ParagraphPrefixConversionReaction.onPrefixMatched` signature.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
angelosilvestre: Updated.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
angelosilvestre: Same answer from the previous comment.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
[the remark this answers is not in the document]
matthew-carroll: In that case, we should change the name to `TextNodePrefixConversionReaction`. And then, probably, we should create another `ParagraphPrefixConversionReaction` that extends `TextNodePrefixConversionReaction` and only accepts a `ParagraphNode`. Otherwise, this change is allowing things to happen in other reactions that shouldn't happen.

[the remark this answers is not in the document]
angelosilvestre: Updated.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -624,13 +655,152 @@ class LinkifyReaction implements EditReaction {
...
+
+    final insertionNode = document.getNodeById(insertionEvent.nodeId) as TextNode;
+
+    final upstreamCharacter = insertionNode.text.text[insertionEvent.offset - 1];
+    if (upstreamCharacter != '-') {
+      return;
+    }
+
+    // A dash was inserted after another dash.
+    // Convert the two dashes to an em-dash.
```
matthew-carroll, line 719: I'm not gonna hold up the merge on this, but this is another example of a two line comment that should be a one line comment.

matthew-carroll (approved):
LGTM

PR #1713, opened 2023-12-28
[SuperEditor][Desktop] Keep selection base while dragging (Resolves #1393)

### LINKED ISSUE #1393: [SuperEditor] - Selection base changes as user drags selection in animated component demo
Reproduction:

1. Launch Example app on Mac
2. Open the animated component demo, which shows a list of tasks
3. Start dragging a selection from a task near the bottom of the list
4. Drag the selection upward, causing a number of tasks to animate to become larger

Expected: The base of the selection remains at the same text offset in the initially selected task.

Actual: The base of the selection moves around based on the (x,y) offset of the initial selection, which results in changing document positions because the task components are changing size.

angelosilvestre:
[SuperEditor][Desktop] Keep selection base while dragging. Resolves #1393

In the animated component demo, dragging up from a task near the bottom causes the selection base to change while the user is dragging:

https://github.com/superlistapp/super_editor/assets/7597082/fbc8a6d1-23a7-4f0a-90a3-2121c860d6a5

Currently, on each pan update, we re-compute the whole selection by checking the selection in the drag region. This is a problem if a component resizes itself, because, at the start of the drag gestures, the offset points to a component and while the user is dragging and the components are resizing, the starting offset might point to another component.

To fix that, this PR caches the starting selection base and reuses-it at each pan update. This isn't a problem on mobile, because when the user drags the expanded drag handle, we are already expanding the current selection.

https://github.com/superlistapp/super_editor/assets/7597082/c4fc6ced-12d2-417f-aaa3-128583bf2b96

This change caused one test to fail, due to a different selection affinity.

### DIFF
--- a/super_editor/lib/src/default_editor/document_gestures_mouse.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_mouse.dart
@@ -88,14 +88,16 @@ class _DocumentMouseInteractorState extends State<DocumentMouseInteractor> with 
   // Tracks user drag gestures for selection purposes.
   SelectionType _selectionType = SelectionType.position;
   Offset? _dragStartGlobal;
+  // The selection's document position where the user started dragging an expanded selection.
+  // The selection base is cached instead of continuously re-computed because components
+  // can change size and position during selection.
+  DocumentPosition? _dragSelectionBase;
   Offset? _dragEndGlobal;
   bool _expandSelectionDuringDrag = false;
   // When selecting by word, this is the initial word's upstream position.
   DocumentPosition? _wordSelectionUpstream;
   // When selecting by word, this is the initial word's downstream position.
   DocumentPosition? _wordSelectionDownstream;
-  // Holds the first selection base of a drag gesture.
-  DocumentPosition? _startingSelectionBase;
 
   /// Holds which kind of device started a pan gesture, e.g., a mouse or a trackpad.
   PointerDeviceKind? _panGestureDevice;
@@ -551,7 +553,7 @@ class _DocumentMouseInteractorState extends State<DocumentMouseInteractor> with 
       _expandSelectionDuringDrag = false;
       _wordSelectionUpstream = null;
       _wordSelectionDownstream = null;
-      _startingSelectionBase = null;
+      _dragSelectionBase = null;
     });
 
     widget.autoScroller.disableAutoScrolling();
@@ -624,15 +626,9 @@ Updating drag selection:
       extentOffsetInDocument,
     );
 
-    // If a component resizes itself depending on whether or not it's selected,
-    // we might end up in a situation where the starting drag offset points
-    // to a different component than the component it was pointing at
-    // the start of the gesture.
-    // Cache and reuse the first selection base of the drag gesture,
-    // so the selection base never changes during a single drag gesture.
-    _startingSelectionBase ??= selection?.base;
+    _dragSelectionBase ??= selection?.base;
 
-    DocumentPosition? basePosition = _startingSelectionBase;
+    DocumentPosition? basePosition = _dragSelectionBase;
     DocumentPosition? extentPosition = selection?.extent;
     editorGesturesLog.fine(" - base: $basePosition, extent: $extentPosition");
 
--- a/super_editor/lib/src/default_editor/document_gestures_mouse.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_mouse.dart
@@ -94,6 +94,8 @@ class _DocumentMouseInteractorState extends State<DocumentMouseInteractor> with 
   DocumentPosition? _wordSelectionUpstream;
   // When selecting by word, this is the initial word's downstream position.
   DocumentPosition? _wordSelectionDownstream;
+  // Holds the first selection base of a drag gesture.
+  DocumentPosition? _startingSelectionBase;
 
   /// Holds which kind of device started a pan gesture, e.g., a mouse or a trackpad.
   PointerDeviceKind? _panGestureDevice;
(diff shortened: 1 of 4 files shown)


super_editor/lib/src/default_editor/document_gestures_mouse.dart
```diff
@@ -94,6 +94,8 @@ class _DocumentMouseInteractorState extends State<DocumentMouseInteractor> with
   DocumentPosition? _wordSelectionUpstream;
   // When selecting by word, this is the initial word's downstream position.
   DocumentPosition? _wordSelectionDownstream;
+  // Holds the first selection base of a drag gesture.
+  DocumentPosition? _startingSelectionBase;
```
matthew-carroll, line 98: It looks like the position of this property would make more sense just below `_dragStartGlobal`, given that both of these are related to starting a selection.

I think we should also probably called it `_dragSelectionBase`. And I'd change the comment to "The selection's document position where the user started dragging an expanded selection."

super_editor/lib/src/default_editor/document_gestures_mouse.dart
```diff
@@ -620,7 +623,16 @@ Updating drag selection:
       baseOffsetInDocument,
       extentOffsetInDocument,
     );
-    DocumentPosition? basePosition = selection?.base;
+
+    // If a component resizes itself depending on whether or not it's selected,
```
matthew-carroll, line 627: I don't think we need this comment because the purpose of this comment seems to be to explain an earlier bug, but no one is going to remember that bug moving forward. If you want to make sure that we remember why we cache the selection base, then we should document that in the comment that sits above the properties declaration. We can say "The selection base is cached instead of continuously re-computed because components can change size and position during selection."

super_editor/test/super_editor/supereditor_selection_test.dart
```diff
@@ -305,6 +306,84 @@ void main() {
       );
     });
 
+    testWidgetsOnArbitraryDesktop("keeps selection base while dragging an expandable component", (tester) async {
```
matthew-carroll, line 309: It looks like this test name should probably read "keeps selection base while dragging a selection across components that change size"

super_editor/test/super_editor/supereditor_selection_test.dart
```diff
@@ -305,6 +306,84 @@ void main() {
...
+      );
+
+      await tester //
+          .createDocument()
+          .withCustomContent(document)
+          .withAddedComponents([_ExpandingTaskComponentBuilder()]) //
+          .pump();
+
+      // Place the caret at "Tas|k 3" to make it expand.
+      await tester.placeCaretInParagraph('3', 3);
```
matthew-carroll, line 337: Does the place caret method have a pump and settle?

super_editor/test/super_editor/supereditor_selection_test.dart
```diff
@@ -305,6 +306,84 @@ void main() {
...
+          const DocumentSelection.collapsed(
+            position: DocumentPosition(
+              nodeId: '3',
+              nodePosition: TextNodePosition(offset: 3),
+            ),
+          ),
+        ),
+      );
+
+      // Start dragging from "Tas|k 3".
```
matthew-carroll, line 350: Let's also describe where we're trying to go with the drag.

super_editor/test/super_editor/supereditor_selection_test.dart
```diff
@@ -305,6 +306,84 @@ void main() {
...
+      final gesture = await tester.startDocumentDragFromPosition(
+        from: const DocumentPosition(
+          nodeId: '3',
+          nodePosition: TextNodePosition(offset: 3),
+        ),
+      );
+      addTearDown(() => gesture.removePointer());
+
+      // Gradually move up.
```
matthew-carroll, line 359: Move up to where?

super_editor/test/super_editor/supereditor_selection_test.dart
```diff
@@ -1142,6 +1221,83 @@ class _UnselectableHorizontalRuleComponent extends StatelessWidget {
   }
 }
 
+/// Builds [TaskComponentViewModel]s and [_ExpandingTaskComponent]s for every
+/// [TaskNode] in a document.
+class _ExpandingTaskComponentBuilder extends ComponentBuilder {
```
matthew-carroll, line 1226: This should probably be added to an existing or new test tool file. If we accumulate things of this size in this test suite for various purposes, we'll quickly be overrun with side details that make it difficult to find and work with the tests in this suite.

super_editor/test/super_editor/supereditor_selection_test.dart
angelosilvestre: Yes, it has a `pumpAndSettle` call.

super_editor/test/super_editor/supereditor_selection_test.dart
angelosilvestre: Moved into `supereditor_test_tools.dart`

matthew-carroll: It looks like you copied the code over, but left the original code

angelosilvestre: The component builder was left behind... Updated.

super_editor/lib/src/default_editor/document_gestures_mouse.dart
```diff
@@ -549,6 +553,7 @@ class _DocumentMouseInteractorState extends State<DocumentMouseInteractor> with
       _expandSelectionDuringDrag = false;
       _wordSelectionUpstream = null;
       _wordSelectionDownstream = null;
+      _dragSelectionBase = null;
```
matthew-carroll, line 556: Can you also move this up to match the same relative line position that we used when declaring the properties?

matthew-carroll (approved):
LGTM

PR #1714, opened 2023-12-28
Paragraph alignment now changes immediately - moved paragraph change from direct document manipulation to a request/command pair (Resolves #1587)

### LINKED ISSUE #1587: Changing text alignment doesn't reflect the alignment change on the text node within the document immediately.

matthew-carroll:
Paragraph alignment now changes immediately - moved paragraph change from direct document manipulation to a request/command pair (Resolves #1587)

https://github.com/superlistapp/super_editor/assets/7259036/fb46e49a-6c91-4ba4-ad9e-11cf8619ec60

### DIFF
--- a/super_editor/example/lib/demos/example_editor/_toolbar.dart
+++ b/super_editor/example/lib/demos/example_editor/_toolbar.dart
@@ -453,8 +453,14 @@ class _EditorToolbarState extends State<EditorToolbar> {
         break;
     }
 
-    final selectedNode = widget.document.getNodeById(widget.composer.selection!.extent.nodeId) as ParagraphNode;
-    selectedNode.putMetadataValue('textAlign', newAlignmentValue);
+    // final selectedNode = widget.document.getNodeById(widget.composer.selection!.extent.nodeId) as ParagraphNode;
+    // selectedNode.putMetadataValue('textAlign', newAlignmentValue);
+    widget.editor!.execute([
+      ChangeParagraphAlignmentRequest(
+        nodeId: widget.composer.selection!.extent.nodeId,
+        alignment: newAlignment,
+      ),
+    ]);
   }
 
   /// Returns the localized name for the given [_TextType], e.g.,
--- a/super_editor/example/lib/demos/example_editor/_toolbar.dart
+++ b/super_editor/example/lib/demos/example_editor/_toolbar.dart
@@ -435,26 +435,7 @@ class _EditorToolbarState extends State<EditorToolbar> {
     if (newAlignment == null) {
       return;
     }
-    String? newAlignmentValue;
-    switch (newAlignment) {
-      case TextAlign.left:
-      case TextAlign.start:
-        newAlignmentValue = 'left';
-        break;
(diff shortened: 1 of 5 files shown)


matthew-carroll:
@rutvik110 here's a tactical change that appears to resolve the alignment delay issue. This approach doesn't touch any render objects, or even touch any widgets. Everything happens within the standard `SuperEditor` pipeline. Here's how I approached this problem.

1. Locate the toolbar where paragraph alignment is controlled. I noticed that the toolbar was directly mutating the document, which is something that should never happen in `SuperEditor` any more. Everything should go through requests and commands.
2. I took the existing direct mutation approach from the toolbar and wrote a failing test that shows that after a pump the widget alignment doesn't change.
3. Opened the existing paragraph file and found that we already had a request/command pair that changes the block type. This pair was almost identical to what we needed for paragraph alignment. I copied this pair and adjusted for alignment.
4. I switched the toolbar to use the new request/command pair and the demo app worked fine.
5. I switched the test to use the request/command pair and the test passed.

By starting from the perspective of how alignment should be controlled in `SuperEditor`, rather than focusing on low level Flutter framework details, it was very quick and easy to solve this. When working on future issues, please try working in a similar direction, and by taking similar steps. Ask questions along the way if you can't find something.

rutvik110:
Followed the same approach in the closed PR and it does solves the issue of changing alignment immediately but the text cursor still doesn't get updated immediately wrt new text alignment. The change that I was proposing earlier was related to fixing this particular issue. 

https://github.com/superlistapp/super_editor/assets/65209850/9cf413f7-b2f5-4a3e-8a8c-30bacee7ae5c

matthew-carroll:
Ok, I'll follow up to investigate the caret placement. Thanks for letting me know.

matthew-carroll:
@rutvik110 I traced the various build and layout calls until I eventually discovered that Flutter's `RenderParagraph` setter for `textAlign` was only calling `markNeedsPaint()` instead of `markNeedsLayout()`. This means that when the text alignment changes, Flutter doesn't think any layout details changed, which prevents ancestors from running their own layout again.

I overrode the default setter and added a `markNeedsLayout()` call. I also filed https://github.com/flutter/flutter/issues/140756.

Was this the same path you were on in your PR?

rutvik110:
> Was this the same path you were on in your PR?

yah, the same one.

matthew-carroll:
@rutvik110 can you do a review?

rutvik110 (approved):
LGTM 🚀

PR #1717, opened 2023-12-28
Updates for #1538

angelosilvestre:
This PR addresses the following comments from #1538 

> Can you clarify "to the bottom" of what? screen? boundary? button?

https://github.com/superlistapp/super_editor/pull/1538#discussion_r1437425389

> In addition to earlier comments, I checked out this branch so I could work on paragraph alignment. I discovered that we're displaying too many alignment options in the drop down. This is because we're looping through all TextAlign values which means we're showing both left and start, and right and end. So we're essentially repeating equivalent options.

https://github.com/superlistapp/super_editor/pull/1538#issuecomment-1871473867

### DIFF
--- a/super_editor/example/lib/demos/example_editor/_toolbar.dart
+++ b/super_editor/example/lib/demos/example_editor/_toolbar.dart
@@ -632,7 +632,7 @@ class _EditorToolbarState extends State<EditorToolbar> {
         id: alignment.name,
         icon: _buildTextAlignIcon(alignment),
       ),
-      items: [TextAlign.left, TextAlign.right, TextAlign.center, TextAlign.justify]
+      items: const [TextAlign.left, TextAlign.center, TextAlign.right, TextAlign.justify]
           .map(
             (alignment) => SuperEditorDemoIconItem(
               icon: _buildTextAlignIcon(alignment),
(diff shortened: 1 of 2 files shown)


super_editor/example/lib/demos/example_editor/_toolbar.dart
```diff
@@ -633,6 +633,7 @@ class _EditorToolbarState extends State<EditorToolbar> {
         icon: _buildTextAlignIcon(alignment),
       ),
       items: TextAlign.values
+          .where((alignment) => !const [TextAlign.start, TextAlign.end].contains(alignment))
```
matthew-carroll, line 636: How about we say the ones we want instead of saying the one we don't want?

angelosilvestre: So we can probably use directly:

```dart
items: [TextAlign.left, TextAlign.right, TextAlign.center, TextAlign.justify]
.map(...)
```

instead of iterating over the enum values.

matthew-carroll: Yes, however, I think the typical order is left, center, right, justify - not left, right,...

angelosilvestre: I used this order because that's how it was before the change...

angelosilvestre: Iterating uses the enum order, which is left, right, center, justify

matthew-carroll: Then we should never have used the iteration order. We should explicitly use the order I mentioned above.

angelosilvestre: Updated.

Also, changed the list to be `const`.

matthew-carroll (approved):
LGTM

PR #1575, opened 2023-10-27
Fix: reporting IME geometry crashes when  is true in SuperEditor, caret doesn't blink when placed by 'autofocus' in SuperEditor (Resolves #1566)

matthew-carroll:
Fix: reporting IME geometry crashes when  is true in SuperEditor, caret doesn't blink when placed by 'autofocus' in SuperEditor (Resolves #1566)

One of the bugs is that with `autofocus` `true`, we apparently attempt to report IME geometry before the associated widget has run layout. I'm not sure why this is coming up now. Perhaps we never tried using `autofocus` since we implemented reporting IME geometry. I added a check to only report geometry if we have a layout size.

Also, when using `autofocus`, the caret appears immediately, but doesn't blink. To solve that, I changed the initial caret blinking `bool` in the controller from `false` to `true`. That might have unintended consequences, but I wasn't sure. It solved the blink problem without causing any obvious other problems. We should keep an eye out.

PR #1726, opened 2023-12-29
[SuperEditor][Android] Fix caret not blinking with autofocus (Resolves #1723)

### LINKED ISSUE #1723: [SuperEditor] Android carat doesn't blink when editor autoFocus set to true
**Package Version**
super_editor stable

**Platform**
Android

**Flutter version**
Flutter stable 3.16.5

@matthew-carroll This seems to be the same issue that was originally raised for iOS: https://github.com/superlistapp/super_editor/issues/1566

In that issue, two bugs were reported:
1. The carat doesn't blink when autoFocus in the editor is set to true
2. This issue (carat not blinking) also occurs when initializing an editor and attempting to set selection to the last selectable position

both of which are happening on Android. Here is the PR that fixed this on iOS: https://github.com/superlistapp/super_editor/pull/1575.

angelosilvestre:
[SuperEditor][Android] Fix caret not blinking with autofocus. Resolves #1723

When the editor has autofocus, the caret is display, but it doesn't blink. We had a similar issue on iOS, which was fixed in https://github.com/superlistapp/super_editor/pull/1575

This PR applies the same solution.

### DIFF
--- a/super_editor/lib/src/default_editor/document_caret_overlay.dart
+++ b/super_editor/lib/src/default_editor/document_caret_overlay.dart
@@ -57,6 +57,9 @@ class CaretDocumentOverlayState extends DocumentLayoutLayerState<CaretDocumentOv
   @visibleForTesting
   bool get isCaretVisible => _blinkController.opacity == 1.0;
 
+  @visibleForTesting
+  Duration get caretFlashPeriod => _blinkController.flashPeriod;
+
   @override
   void initState() {
     super.initState();
--- a/super_editor/lib/src/default_editor/document_caret_overlay.dart
+++ b/super_editor/lib/src/default_editor/document_caret_overlay.dart
@@ -52,8 +52,10 @@ class CaretDocumentOverlay extends DocumentLayoutLayerStatefulWidget {
 @visibleForTesting
 class CaretDocumentOverlayState extends DocumentLayoutLayerState<CaretDocumentOverlay, Rect?>
     with SingleTickerProviderStateMixin {
+  late final BlinkController _blinkController;
+
   @visibleForTesting
-  late final BlinkController blinkController;
+  bool get isCaretVisible => _blinkController.opacity == 1.0;
 
   @override
   void initState() {
@@ -61,9 +63,9 @@ class CaretDocumentOverlayState extends DocumentLayoutLayerState<CaretDocumentOv
 
     switch (widget.blinkTimingMode) {
       case BlinkTimingMode.ticker:
-        blinkController = BlinkController(tickerProvider: this);
+        _blinkController = BlinkController(tickerProvider: this);
       case BlinkTimingMode.timer:
-        blinkController = BlinkController.withTimer();
+        _blinkController = BlinkController.withTimer();
     }
 
     widget.composer.selectionNotifier.addListener(_onSelectionChange);
@@ -87,7 +89,7 @@ class CaretDocumentOverlayState extends DocumentLayoutLayerState<CaretDocumentOv
   void dispose() {
     widget.composer.selectionNotifier.removeListener(_onSelectionChange);
 
-    blinkController.dispose();
+    _blinkController.dispose();
 
     super.dispose();
   }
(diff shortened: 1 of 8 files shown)


super_editor/lib/src/default_editor/document_gestures_touch_android.dart
```diff
@@ -133,7 +133,7 @@ class SuperEditorAndroidControlsController {
 
   /// Whether the caret should blink right now.
   ValueListenable<bool> get shouldCaretBlink => _shouldCaretBlink;
-  final _shouldCaretBlink = ValueNotifier<bool>(false);
+  final _shouldCaretBlink = ValueNotifier<bool>(true);
```
matthew-carroll, line 136: Should we have a test for this?

angelosilvestre: Added tests

super_editor/lib/src/default_editor/document_caret_overlay.dart
```diff
@@ -46,22 +46,24 @@ class CaretDocumentOverlay extends DocumentLayoutLayerStatefulWidget {
...
+  DocumentLayoutLayerState<CaretDocumentOverlay, Rect?> createState() => CaretDocumentOverlayState();
 }
 
-class _CaretDocumentOverlayState extends DocumentLayoutLayerState<CaretDocumentOverlay, Rect?>
+@visibleForTesting
+class CaretDocumentOverlayState extends DocumentLayoutLayerState<CaretDocumentOverlay, Rect?>
     with SingleTickerProviderStateMixin {
-  late final BlinkController _blinkController;
+  @visibleForTesting
+  late final BlinkController blinkController;
```
matthew-carroll, line 56: Let's not make the controller public, let's just add a visible-for-testing method that reports whether the caret is currently visible.

super_editor/test/super_editor/supereditor_caret_test.dart
```diff
@@ -326,6 +328,81 @@ void main() {
         });
       });
     });
+
+    group('blinks caret at rest', () {
```
matthew-carroll, line 332: What does "rest" mean? Is this group about `SuperEditor` initialization?

I've been thinking that we should have a test suite specifically for `SuperEditor` initialization where we check all the ways it might initialize. Do you want to move these two tests into a new suite called `supereditor_initialization_test.dart`?

We could have a group called "SuperEditor > initialization >"

super_editor/test/super_editor/supereditor_caret_test.dart
```diff
@@ -409,3 +486,21 @@ Future<void> _resizeWindow({
     await tester.pumpAndSettle();
   }
 }
+
+bool _isCaretVisible(WidgetTester tester) {
```
matthew-carroll, line 490: This looks like it should be an inspector method.

super_editor/test/super_editor/supereditor_caret_test.dart
angelosilvestre: > What does "rest" mean? Is this group about SuperEditor initialization?

Here, rest just means that the user isn't typing. The first test is about `SuperEditor` initialization, but the second isn't.

I added a `supereditor_initialization_test.dart` and moved the first test.

super_editor/test/super_editor/supereditor_initialization_test.dart
```diff
@@ -0,0 +1,52 @@
...
+import 'package:super_editor/src/core/document_selection.dart';
+import 'package:super_editor/src/default_editor/text.dart';
+import 'package:super_editor/src/test/super_editor_test/supereditor_inspector.dart';
+import 'package:super_text_layout/super_text_layout.dart';
+
+import 'supereditor_test_tools.dart';
+
+void main() {
+  group('SuperEditor > initialization >', () {
+    testWidgetsOnAllPlatforms('blinks caret when autofocus is true', (tester) async {
```
matthew-carroll, line 13: Let's call this test 'immediately shows and blinks the caret when autofocus is true'

super_editor/test/super_editor/supereditor_caret_test.dart
```diff
@@ -326,6 +327,43 @@ void main() {
         });
       });
     });
+
+    testWidgetsOnAllPlatforms('blinks caret when focused by tap', (tester) async {
```
matthew-carroll, line 331: Let's call this test 'blinks the caret when the user places the caret with a single tap'

super_editor/test/super_editor/supereditor_caret_test.dart
```diff
@@ -326,6 +327,43 @@ void main() {
...
+      // Configure BlinkController to animate, otherwise it won't blink.
+      BlinkController.indeterminateAnimationsEnabled = true;
+      addTearDown(() => BlinkController.indeterminateAnimationsEnabled = false);
+
+      // Duration to switch between visible and invisible.
+      const flashPeriod = Duration(milliseconds: 500);
```
matthew-carroll, line 337: Can we expose this from the caret system in a way that allows us to create a `SuperEditorInspector` to query it?

super_editor/test/super_editor/supereditor_caret_test.dart
```diff
@@ -3,6 +3,7 @@ import 'package:flutter_test/flutter_test.dart';
 import 'package:flutter_test_runners/flutter_test_runners.dart';
 import 'package:super_editor/super_editor.dart';
 import 'package:super_editor/super_editor_test.dart';
+import 'package:super_text_layout/super_text_layout.dart';
```
matthew-carroll, line 6: Do we need this import?

angelosilvestre: The `BlinkController` seems to be exposed only by `super_text_layout.dart`. Should we expose it somewhere else?

super_editor/test/super_editor/supereditor_caret_test.dart
angelosilvestre: Added a `SuperEditorInspector.caretFlashPeriod` method.

super_editor/test/super_editor/supereditor_caret_test.dart
matthew-carroll: We can leave it as-is

matthew-carroll (approved):
LGTM

Issue #120, opened 2021-04-13
When hovering & scrolling, the editor should consume available scroll space before parent can scroll


roughike:
**Current behavior:**
When the mouse cursor is over the editor, the editor has scrollable content, and the page is being scrolled by the mouse wheel, the page and editor scroll at the same time. It's not possible to use the mouse scroll wheel to scroll to the bottom of the editor, because the editor moves at the same time while its inner content is scrolling.

![scrolling-actual](https://user-images.githubusercontent.com/13744304/114549421-a9d8f800-9c69-11eb-9add-ab90ea620677.gif)

The above example uses a `SingleChildScrollView` and a `Column`. I also tried to share a `ScrollController` between the `Editor` and `SingleChildScrollView`, but there was no difference in behavior.

<details><summary>Reproducible sample</summary>
<p>

```dart
import 'package:flutter/material.dart';
import 'package:super_editor/super_editor.dart';

void main() {
  runApp(MyApp());
}

class MyApp extends StatelessWidget {
  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      title: 'Flutter Demo',
      theme: ThemeData(
        primarySwatch: Colors.blue,
      ),
      home: MyHomePage(title: 'Flutter Demo Home Page'),
    );
  }
}

class MyHomePage extends StatefulWidget {
  MyHomePage({Key key, this.title}) : super(key: key);
  final String title;

  @override
  _MyHomePageState createState() => _MyHomePageState();
}

class _MyHomePageState extends State<MyHomePage> {
  ScrollController _scrollController;
  Document _doc;
  DocumentEditor _docEditor;

  @override
  void initState() {
    super.initState();
    _scrollController = ScrollController();
    _doc = _createInitialDocument();
    _docEditor = DocumentEditor(document: _doc);
  }

  @override
  void dispose() {
    _doc.dispose();
    super.dispose();
  }

  @override
  Widget build(BuildContext context) {
    return Scaffold(
      backgroundColor: Colors.blue,
      body: Scrollbar(
        child: SingleChildScrollView(
          controller: _scrollController,
          child: Column(
            children: [
              Center(
                child: ConstrainedBox(
                  constraints: BoxConstraints(maxWidth: 800),
                  child: Row(
                    children: [
                      Expanded(
                        child: Text(
                          'A supercharged rich text editor for Flutter',
                          style: TextStyle(fontSize: 40),
                        ),
                      ),
                      Expanded(
                        child: Container(
[...]

PR #1733, opened 2024-01-02
[SuperEditor][SuperReader] Adds tests editor's scrolling when present within parent scrollable

rutvik110:
[SuperEditor][SuperReader] Adds tests editor's scrolling when present within parent scrollable (Related #120)

This PR adds tests for locking down the behaviour for `SuperEditor` and `SuperReader` for the following scenario: 

When scrolling within a parent scrollable, makes sure parent scrollable is scrolled after all of the scrollable content is consumed within editor when the user's hovering over editor.

### DIFF
--- a/super_editor/example/lib/demos/example_editor/_example_document.dart
+++ b/super_editor/example/lib/demos/example_editor/_example_document.dart
@@ -4,14 +4,14 @@ import 'package:super_editor/super_editor.dart';
 MutableDocument createInitialDocument() {
   return MutableDocument(
     nodes: [
-      ImageNode(
-        id: "1",
-        imageUrl: 'https://i.ibb.co/5nvRdx1/flutter-horizon.png',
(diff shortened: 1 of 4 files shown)


matthew-carroll:
@rutvik110 this PR has been around for a while. Can you do something with it?

rutvik110:
> @rutvik110 this PR has been around for a while. Can you do something with it?

This is currently blocked by #1798. Will be working on that issue before I can get this moving.

rutvik110:
@matthew-carroll Looking deeper into the issues blocking this PR, it's clear that we'll need to get the #1798 and #1734 fixed before I can move on with this PR. More importantly, #1734 as I believe it may be part of the problem within #1798, still need to do some investigation here.

I think @angelosilvestre had some work done on the #1734 in the past, I'll try to take look into that and see what we can do here.

matthew-carroll:
Closing as stale

PR #1735, opened 2024-01-02
[SuperEditor][DemoApp] Fix image resizing (Resolves #1708)(Resolves #1529)

### LINKED ISSUE #1529: [SuperEditor][Example App] - Fix image size change with toolbar
When the user presses a button to change the size of an image in the example app, the size doesn't change immediately.

This is because, when changing the image size, the node's metadata is changed, but nothing triggers a re-run of the `SuperEditor` presenter.

We seem to lack an expected path to trigger a reflow in such a situation. Create such a path and then use it in the example app.

### LINKED ISSUE #1708: [SuperEditor][Example] - Header image doesn't change size immediately
In the `SuperEditor` example app, the header image can be selected and switched from full width to limited width. There are two issues with this behavior.

First, when pressing the button on the popover toolbar to change the image size, the image size doesn't change immediately. Instead, Flutter has to pump another frame. This can be accomplished by running a hot reload, or changing to dark mode, or some other action that forces a re-render, but it doesn't happen automatically, which makes it seem as if the image size controls don't work.

Second, the tooltips for the image size controls say "italics" and "bold". The tooltips should be updated to appropriate labels.

angelosilvestre:
[SuperEditor][DemoApp] Fix image resizing. Resolves #1708, Resolves #1529

We have an image toolbar on the demo app the lets users switch the image from full width to limited width. However, using this toolbar isn't updating the layout immediately.

The reason is that we are directly mutating the node, instead of using the request/command pipeline.

This PR adds  `ChangeComponentStylesRequest` and `ChangeComponentStylesCommand` to handle the style changes.

### DIFF
--- a/super_editor/example/lib/demos/example_editor/_toolbar.dart
+++ b/super_editor/example/lib/demos/example_editor/_toolbar.dart
@@ -838,7 +838,7 @@ class _ImageFormatToolbarState extends State<ImageFormatToolbar> {
                   onPressed: _makeImageConfined,
                   icon: const Icon(Icons.photo_size_select_large),
                   splashRadius: 16,
-                  tooltip: AppLocalizations.of(context)!.labelBold,
+                  tooltip: 'Limited width',
                 ),
               ),
               Center(
@@ -846,7 +846,7 @@ class _ImageFormatToolbarState extends State<ImageFormatToolbar> {
                   onPressed: _makeImageFullBleed,
                   icon: const Icon(Icons.photo_size_select_actual),
                   splashRadius: 16,
-                  tooltip: AppLocalizations.of(context)!.labelItalics,
+                  tooltip: 'Full width',
                 ),
               ),
             ],
--- a/super_editor/example/lib/demos/example_editor/_toolbar.dart
+++ b/super_editor/example/lib/demos/example_editor/_toolbar.dart
@@ -838,7 +838,7 @@ class _ImageFormatToolbarState extends State<ImageFormatToolbar> {
                   onPressed: _makeImageConfined,
                   icon: const Icon(Icons.photo_size_select_large),
                   splashRadius: 16,
-                  tooltip: 'Limited width',
+                  tooltip: AppLocalizations.of(context)!.labelLimitedWidth,
                 ),
               ),
               Center(
@@ -846,7 +846,7 @@ class _ImageFormatToolbarState extends State<ImageFormatToolbar> {
                   onPressed: _makeImageFullBleed,
                   icon: const Icon(Icons.photo_size_select_actual),
                   splashRadius: 16,
-                  tooltip: 'Full width',
+                  tooltip: AppLocalizations.of(context)!.labelFullWidth,
                 ),
               ),
             ],
--- a/super_editor/example/lib/demos/example_editor/example_editor.dart
+++ b/super_editor/example/lib/demos/example_editor/example_editor.dart
@@ -488,7 +488,7 @@ class _ExampleEditorState extends State<ExampleEditor> {
         );
 
         _docEditor.execute([
-          ChangeComponentStylesRequest(
+          ChangeSingleColumnLayoutComponentStylesRequest(
             nodeId: nodeId,
             styles: newStyles,
           )
--- a/super_editor/example/lib/demos/example_editor/example_editor.dart
+++ b/super_editor/example/lib/demos/example_editor/example_editor.dart
@@ -482,15 +482,14 @@ class _ExampleEditorState extends State<ExampleEditor> {
         print("Applying width $width to node $nodeId");
         final node = _doc.getNodeById(nodeId)!;
         final currentStyles = SingleColumnLayoutComponentStyles.fromMetadata(node);
-        final newStyles = SingleColumnLayoutComponentStyles(
-          width: width,
-          padding: currentStyles.padding,
-        );
 
         _docEditor.execute([
           ChangeSingleColumnLayoutComponentStylesRequest(
             nodeId: nodeId,
-            styles: newStyles,
+            styles: SingleColumnLayoutComponentStyles(
+              width: width,
+              padding: currentStyles.padding,
+            ),
           )
         ]);
       },
--- a/super_editor/example/lib/demos/example_editor/example_editor.dart
+++ b/super_editor/example/lib/demos/example_editor/example_editor.dart
@@ -482,14 +482,17 @@ class _ExampleEditorState extends State<ExampleEditor> {
         print("Applying width $width to node $nodeId");
         final node = _doc.getNodeById(nodeId)!;
         final currentStyles = SingleColumnLayoutComponentStyles.fromMetadata(node);
-        SingleColumnLayoutComponentStyles(
+        final newStyles = SingleColumnLayoutComponentStyles(
           width: width,
           padding: currentStyles.padding,
-        ).applyTo(node);
+        );
 
-        // TODO: schedule a presentation reflow so that the image changes size immediately (https://github.com/superlistapp/super_editor/issues/1529)
-        //       Right now, nothing happens when pressing the button, unless we force a
-        //       rebuild/reflow.
+        _docEditor.execute([
+          ChangeComponentStylesRequest(
+            nodeId: nodeId,
+            styles: newStyles,
+          )
+        ]);
       },
       closeToolbar: _hideImageToolbar,
     );
--- a/super_editor/lib/src/default_editor/text.dart
+++ b/super_editor/lib/src/default_editor/text.dart
@@ -1364,7 +1364,7 @@ class ToggleTextAttributionsCommand implements EditCommand {
   }
 }
 
-/// Changes layout styles, like padding and width, of a [DocumentNode].
+/// Changes layout styles, like padding and width, of a component within a [SingleColumnDocumentLayout].
 class ChangeSingleColumnLayoutComponentStylesRequest implements EditRequest {
   const ChangeSingleColumnLayoutComponentStylesRequest({
     required this.nodeId,
--- a/super_editor/lib/src/default_editor/text.dart
+++ b/super_editor/lib/src/default_editor/text.dart
@@ -1364,8 +1364,9 @@ class ToggleTextAttributionsCommand implements EditCommand {
   }
 }
 
-class ChangeComponentStylesRequest implements EditRequest {
-  const ChangeComponentStylesRequest({
+/// Changes layout styles, like padding and width, of a [DocumentNode].
+class ChangeSingleColumnLayoutComponentStylesRequest implements EditRequest {
+  const ChangeSingleColumnLayoutComponentStylesRequest({
     required this.nodeId,
     required this.styles,
   });
@@ -1374,8 +1375,8 @@ class ChangeComponentStylesRequest implements EditRequest {
   final SingleColumnLayoutComponentStyles styles;
 }
 
-class ChangeComponentStylesCommand implements EditCommand {
-  ChangeComponentStylesCommand({
+class ChangeSingleColumnLayoutComponentStylesCommand implements EditCommand {
+  ChangeSingleColumnLayoutComponentStylesCommand({
     required this.nodeId,
     required this.styles,
   });
--- a/super_editor/lib/src/default_editor/text.dart
+++ b/super_editor/lib/src/default_editor/text.dart
@@ -1364,6 +1364,40 @@ class ToggleTextAttributionsCommand implements EditCommand {
   }
 }
 
+class ChangeComponentStylesRequest implements EditRequest {
+  const ChangeComponentStylesRequest({
+    required this.nodeId,
+    required this.styles,
+  });
+
+  final String nodeId;
+  final SingleColumnLayoutComponentStyles styles;
+}
+
+class ChangeComponentStylesCommand implements EditCommand {
+  ChangeComponentStylesCommand({
+    required this.nodeId,
+    required this.styles,
+  });
+
+  final String nodeId;
+  final SingleColumnLayoutComponentStyles styles;
+
+  @override
+  void execute(EditContext context, CommandExecutor executor) {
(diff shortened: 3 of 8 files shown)


super_editor/example/lib/demos/example_editor/_toolbar.dart
```diff
@@ -851,15 +851,15 @@ class _ImageFormatToolbarState extends State<ImageFormatToolbar> {
                   onPressed: _makeImageConfined,
                   icon: const Icon(Icons.photo_size_select_large),
                   splashRadius: 16,
-                  tooltip: AppLocalizations.of(context)!.labelBold,
+                  tooltip: 'Limited width',
```
matthew-carroll, line 854: Please add these to app localizations.

super_editor/lib/src/default_editor/text.dart
```diff
@@ -1351,6 +1351,40 @@ class ToggleTextAttributionsCommand implements EditCommand {
   }
 }
 
+class ChangeComponentStylesRequest implements EditRequest {
```
matthew-carroll, line 1354: Let's change this name to `ChangeSingleColumnLayoutComponentStylesRequest`. Also, please add a Dart Doc because the name of this request might not be completely clear to most readers.

super_editor/example/lib/demos/example_editor/example_editor.dart
```diff
@@ -482,14 +482,17 @@ class _ExampleEditorState extends State<ExampleEditor> {
...
           width: width,
           padding: currentStyles.padding,
-        ).applyTo(node);
+        );
 
-        // TODO: schedule a presentation reflow so that the image changes size immediately (https://github.com/superlistapp/super_editor/issues/1529)
-        //       Right now, nothing happens when pressing the button, unless we force a
-        //       rebuild/reflow.
+        _docEditor.execute([
+          ChangeComponentStylesRequest(
```
matthew-carroll, line 491: Please add appropriate tests to check for style changes after a pump.

super_editor/lib/src/default_editor/text.dart
angelosilvestre: Updated.

super_editor/lib/src/default_editor/text.dart
```diff
@@ -1351,6 +1351,41 @@ class ToggleTextAttributionsCommand implements EditCommand {
   }
 }
 
+/// Changes layout styles, like padding and width, of a [DocumentNode].
```
matthew-carroll, line 1354: This description in't quite accurate. First, `DocumentNode`s are logical - they're never styled. This applies to a `SingleColumnLayoutComponent`, as you can see by the name of the request. Second, this request will only impact components within a `SingleColumnLayout`.

super_editor/test/super_editor/supereditor_single_column_layout_test.dart
```diff
@@ -0,0 +1,44 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:flutter_test_runners/flutter_test_runners.dart';
+import 'package:super_editor/super_editor.dart';
+import 'package:super_editor/super_editor_test.dart';
+
+import 'supereditor_test_tools.dart';
+
+void main() {
+  group('SuperEditor > single column layout >', () {
```
matthew-carroll, line 10: How about padding?

super_editor/lib/src/default_editor/text.dart
angelosilvestre: Updated.

super_editor/test/super_editor/supereditor_single_column_layout_test.dart
angelosilvestre: Updated.

super_editor/example/lib/demos/example_editor/example_editor.dart
```diff
@@ -482,14 +482,17 @@ class _ExampleEditorState extends State<ExampleEditor> {
         print("Applying width $width to node $nodeId");
         final node = _doc.getNodeById(nodeId)!;
         final currentStyles = SingleColumnLayoutComponentStyles.fromMetadata(node);
-        SingleColumnLayoutComponentStyles(
+        final newStyles = SingleColumnLayoutComponentStyles(
```
matthew-carroll, line 485: Do we need to declare this variable? Can we create the new styles directly in the request?

super_editor/test/super_editor/supereditor_single_column_layout_test.dart
matthew-carroll: Was the padding test missed in the latest push?

matthew-carroll: I see padding was added to the existing test. Please specify a test just for padding. Padding could always fail when width works, or vis-a-versa.

super_editor/test/super_editor/supereditor_single_column_layout_test.dart
```diff
@@ -0,0 +1,100 @@
...
+    testWidgetsOnAllPlatforms('updates component width after changing component styles', (tester) async {
```
matthew-carroll, line 11: I've read through this test a few times but I can't seem to connect what I'm reading in the test to the test name. I see some padding manipulation going on, but then I also see references to width. I can't tell what the goal is, or why padding is related to it. Can you please revisit the comments, and possibly the name of the test, to make it clearer?

super_editor/test/super_editor/supereditor_single_column_layout_test.dart
```diff
@@ -0,0 +1,100 @@
...
+                matching: find.byType(Padding),
+              )
+              .first)
+          .padding;
+      expect(paddingAfter, paddingBefore);
+      expect(SuperEditorInspector.findComponentSize('1').width, 1000.0 - paddingBefore.horizontal);
+    });
+  });
+
+  testWidgetsOnAllPlatforms('updates component padding after changing component styles', (tester) async {
```
matthew-carroll, line 63: I'm looking at this test name as well, finding it somewhat confusing, because it reads cyclically. Perhaps the following name would be more clear: `updates padding around each component in a document layout, when the overall document layout padding is changed by the editor`

That's a long name, which isn't great, but I think each of those details are relevant.

super_editor/test/super_editor/supereditor_single_column_layout_test.dart
angelosilvestre: Updated trying to make it more clear.

super_editor/test/super_editor/supereditor_single_column_layout_test.dart
```diff
@@ -0,0 +1,95 @@
...
+      // the maximum width a component can be.
+      final context = await tester
+          .createDocument() //
+          .withSingleEmptyParagraph()
+          .withEditorSize(const Size(1000.0, 5000.0))
+          .pump();
+
+      // Find the padding around the component to ensure the component took all
+      // the editor width, minus the padding.
+      final componentPadding = tester
```
matthew-carroll, line 23: I still don't understand why we care about the padding?

super_editor/test/super_editor/supereditor_single_column_layout_test.dart
```diff
@@ -0,0 +1,95 @@
...
+          .widget<Padding>(find
+              .ancestor(
+                of: find.byWidget(SuperEditorInspector.findWidgetForComponent('1')),
+                matching: find.byType(Padding),
+              )
+              .first)
+          .padding;
+
+      // Ensure the component doesn't start taking all the available width.
```
matthew-carroll, line 32: What do you mean when you say "doesn't start taking"? I'm confused by the tense of this comment...

super_editor/test/super_editor/supereditor_single_column_layout_test.dart
```diff
@@ -0,0 +1,95 @@
...
+      expect(
+        SuperEditorInspector.findComponentSize('1').width,
+        lessThan(1000.0 - componentPadding.horizontal),
+      );
+
+      // Changes the width.
```
matthew-carroll, line 38: Please use the active tense when describing behavior that follows the comment, also please be more specific to avoid confusion: "Change the width the of the first component in the document layout"

super_editor/test/super_editor/supereditor_single_column_layout_test.dart
```diff
@@ -0,0 +1,95 @@
...
+            nodeId: '1',
+            styles: SingleColumnLayoutComponentStyles(
+              width: double.infinity,
+            ),
+          ),
+        ],
+      );
+      await tester.pump();
+
+      // Ensure the component took all available width.
```
matthew-carroll, line 51: Wasn't the component already taking up all available width? It seems like you're using padding to test width, rather than using width to test width. I'm still confused by the goal of this test vs what's actually being done in the test.

super_editor/test/super_editor/supereditor_single_column_layout_test.dart
angelosilvestre: We use the padding because even if we are setting the width to `double.infinity`, the component still respects the padding. So, if the editor width is `1000` pixels and we have a padding of `100` pixels, setting the component width to `double.infinity` makes the component `900` pixels wide.

angelosilvestre: I'll change the test to use a stylesheet without padding.

super_editor/test/super_editor/supereditor_single_column_layout_test.dart
angelosilvestre: Updated.

super_editor/test/super_editor/supereditor_single_column_layout_test.dart
angelosilvestre: Updated.

super_editor/test/super_editor/supereditor_single_column_layout_test.dart
matthew-carroll: Can we avoid all of that concern by using two different non-infinite widths? If you start with a width of 200 and then go to width of 300 does that verify what matters? If so, I think that would eliminate any concern about padding or window size...

angelosilvestre: I modified to avoid needing to check the padding in this test. Should I still modify the test?

matthew-carroll: That depends, would doing what I mentioned in my previous comment still effectively verify the desired behavior?

angelosilvestre: I modified the test, please take a look if that's what you are suggesting.

super_editor/test/super_editor/supereditor_single_column_layout_test.dart
```diff
@@ -0,0 +1,88 @@
...
+      final context = await tester
+          .createDocument() //
+          .withSingleEmptyParagraph()
+          .withEditorSize(const Size(1000.0, 5000.0))
+          .useStylesheet(
+            defaultStylesheet.copyWith(addRulesAfter: [
+              StyleRule(
+                BlockSelector.all,
+                (doc, docNode) => {
+                  Styles.padding: const CascadingPadding.all(0.0),
```
matthew-carroll, line 24: This is still using padding to implement a width test. Am I missing something here? I still don't understand why we can't set one width here, and then another width down below, and verify the change. Wouldn't that be much more obvious for the purpose of the test?

angelosilvestre: We can do that, but the resulting size will be the size we set - 48px (the default padding). I zeroed the padding so the resulting size is exactly the size we configure in the request.

matthew-carroll: Does padding usually impact the width of something? If you create a `Container` with width 100px and padding 10px, does the container actually render at 80px, or is it 100px and its child is 80px wide?

angelosilvestre: If you use the `padding` property of the `Container` then yes. The following code:

```dart
Container(
    width: 100,
    height: 100,
    color: Colors.red,
    padding: const EdgeInsets.symmetric(horizontal: 10),
    child: Container(
      color: Colors.blue,
      width: double.infinity,
    ),
)
```

renders this:

Captura de Tela 2024-01-11 às 19 50 37

matthew-carroll: I can't tell what the dimensions are in that image. Is the blue area 80px wide or 100px wide?

angelosilvestre: The blue area is 80px

matthew-carroll: Ok, then please add a comment above the padding that explains that we're zero'ing out the padding so that the width of the component is exactly what we request. Then, please add a line that starts us with an explicit width.

angelosilvestre: Ok, updated.

super_editor/test/super_editor/supereditor_single_column_layout_test.dart
```diff
@@ -0,0 +1,100 @@
...
+                (doc, docNode) => {
+                  // Zeroes the padding so the component is exactly
+                  // the requested size.
+                  Styles.padding: const CascadingPadding.all(0.0),
+                },
+              )
+            ]),
+          )
+          .pump();
+
```
matthew-carroll, line 42: Shouldn't you make sure we're starting with the expected width?

angelosilvestre: Updated.

matthew-carroll (approved):
LGTM

PR #1738, opened 2024-01-02
[SuperEditor] Add image size awareness (Resolves #851)

### LINKED ISSUE #851: [SuperEditor][SuperReader] - add image size awareness
When loading images in a document, it's nice to know the size of the image being loaded, so that the image widget can take up the final amount of space before the image is loaded. That way, the content in the document doesn't "jump" when the image appears.

Add the relevant capabilities to do this. That includes relevant markdown parsing, node properties, and component behavior. The size should be optional, but it should be honored, when available.

angelosilvestre:
[SuperEditor] Add image size awareness. Resolves #851

Currently, we don't have a way to specify the image size. The images don't have an initial height and the component is resized when the image is loaded.

There isn't a standard in the markdown specification for serializing an image size. However, it seems some markdown renderers use one of the following syntaxes:

1.  `alt text` 
2. `alt text{width=800 height=600}`

This PR implements the first syntax, but I can change to the second if needed.

The markdown package doesn't have a way to plug into the image parsing, so I had to copy the whole parsing logic and change to be able to parse the size notation.

The main changes where:

- Added the `_tryParseImageSize` method.
- Added the `_tryParseNumber` method.
- Changed `_parseInlineBracketedLink` and `_parseInlineBareDestinationLink` to handle the size notation
- Added a `AsciiTable` class to hold the character codes instead of using the top level constants prefixed with a `$`.

One thing I noticed it that it doesn't seem to stretch the same way when setting the image height. Here's a comparison of the different rendering depending on the `BoxFit`:

Original, without setting the height:

https://github.com/superlistapp/super_editor/assets/7597082/f0fc169c-e205-4a1c-9c31-b2f843a18924

Setting height to 422px, with `BoxFit.contain`:

https://github.com/superlistapp/super_editor/assets/7597082/afc5521f-a568-4128-be25-212ae2f04ae1

Setting height to 422px, with `BoxFit.cover`:

https://github.com/superlistapp/super_editor/assets/7597082/46e2f9f9-92cd-45f4-813e-0ff3e0f6271b

Setting height to 422px, with `BoxFit.fill`:

https://github.com/superlistapp/super_editor/assets/7597082/9c9cff84-02dc-4d27-8c00-c55baa16a5ac

Setting height to 422px, with `BoxFit.fitHeight`:

https://github.com/superlistapp/super_editor/assets/7597082/08d24401-e705-47ee-b632-34fb76706a0c

Setting height to 422px, with `BoxFit.fitWidth`:

https://github.com/superlistapp/super_editor/assets/7597082/8d005464-5143-4b0d-9f3e-fb29621b11ef

Setting height to 422px, with `BoxFit.none`:

https://github.com/superlistapp/super_editor/assets/7597082/99060f14-03c0-40e1-99ab-0488bf225ebe

Setting height to 422px, with `BoxFit.scaleDown`:

https://github.com/superlistapp/super_editor/assets/7597082/96992852-127a-4b46-8026-ee40f57879d5

### DIFF
--- a/super_editor/example/lib/demos/example_editor/_example_document.dart
+++ b/super_editor/example/lib/demos/example_editor/_example_document.dart
@@ -8,7 +8,8 @@ MutableDocument createInitialDocument() {
         id: "1",
         imageUrl: 'https://i.ibb.co/5nvRdx1/flutter-horizon.png',
         expectedBitmapSize: ExpectedSize(
-          height: 422,
+          width: 1911,
+          height: 630,
         ),
         metadata: const SingleColumnLayoutComponentStyles(
           width: double.infinity,
--- a/super_editor/example/lib/demos/example_editor/_example_document.dart
+++ b/super_editor/example/lib/demos/example_editor/_example_document.dart
@@ -7,7 +7,9 @@ MutableDocument createInitialDocument() {
       ImageNode(
         id: "1",
         imageUrl: 'https://i.ibb.co/5nvRdx1/flutter-horizon.png',
-        height: 422,
+        expectedBitmapSize: ExpectedSize(
+          height: 422,
+        ),
         metadata: const SingleColumnLayoutComponentStyles(
           width: double.infinity,
           padding: EdgeInsets.zero,
--- a/super_editor/example/lib/demos/example_editor/_example_document.dart
+++ b/super_editor/example/lib/demos/example_editor/_example_document.dart
@@ -7,6 +7,7 @@ MutableDocument createInitialDocument() {
       ImageNode(
         id: "1",
         imageUrl: 'https://i.ibb.co/5nvRdx1/flutter-horizon.png',
+        height: 422,
         metadata: const SingleColumnLayoutComponentStyles(
           width: double.infinity,
           padding: EdgeInsets.zero,
--- a/super_editor/lib/src/default_editor/image.dart
+++ b/super_editor/lib/src/default_editor/image.dart
@@ -11,14 +11,12 @@ class ImageNode extends BlockNode with ChangeNotifier {
   ImageNode({
     required this.id,
     required String imageUrl,
-    double? width,
-    double? height,
+    ExpectedSize? expectedBitmapSize,
     String altText = '',
     Map<String, dynamic>? metadata,
   })  : _imageUrl = imageUrl,
-        _altText = altText,
-        _width = width,
-        _height = height {
+        _expectedBitmapSize = expectedBitmapSize,
+        _altText = altText {
     this.metadata = metadata;
 
     putMetadataValue("blockType", const NamedAttribution("image"));
@@ -36,6 +34,22 @@ class ImageNode extends BlockNode with ChangeNotifier {
     }
   }
 
+  /// The expected size of the image.
+  ///
+  /// Used to size the component while the image is still being loaded,
+  /// so the content don't shift after the image is loaded.
+  ExpectedSize? get expectedBitmapSize => _expectedBitmapSize;
+  ExpectedSize? _expectedBitmapSize;
+  set expectedBitmapSize(ExpectedSize? expectedSize) {
+    if (expectedSize == _expectedBitmapSize) {
+      return;
+    }
+
+    _expectedBitmapSize = expectedSize;
+
+    notifyListeners();
+  }
+
   String _altText;
   String get altText => _altText;
   set altText(String newAltText) {
@@ -43,28 +57,6 @@ class ImageNode extends BlockNode with ChangeNotifier {
       _altText = newAltText;
       notifyListeners();
     }
-  }
-
-  double? _width;
-  double? get width => _width;
-  set width(double? newWidth) {
-    if (newWidth == _width) {
-      return;
-    }
-
-    _width = newWidth;
-    notifyListeners();
-  }
-
-  double? _height;
-  double? get height => _height;
-  set height(double? newHeight) {
-    if (newHeight == _height) {
-      return;
-    }
-
-    _height = newHeight;
-    notifyListeners();
   }
 
   @override
@@ -80,8 +72,7 @@ class ImageNode extends BlockNode with ChangeNotifier {
   bool hasEquivalentContent(DocumentNode other) {
     return other is ImageNode &&
         imageUrl == other.imageUrl &&
-        width == other.width &&
-        height == other.height &&
+        expectedBitmapSize == other.expectedBitmapSize &&
         altText == other.altText;
   }
 
@@ -92,12 +83,11 @@ class ImageNode extends BlockNode with ChangeNotifier {
           runtimeType == other.runtimeType &&
           id == other.id &&
           _imageUrl == other._imageUrl &&
-          _width == other.width &&
-          _height == other.height &&
+          _expectedBitmapSize == other.expectedBitmapSize &&
           _altText == other._altText;
 
   @override
-  int get hashCode => id.hashCode ^ _imageUrl.hashCode ^ _altText.hashCode ^ _width.hashCode ^ _height.hashCode;
+  int get hashCode => id.hashCode ^ _imageUrl.hashCode ^ _altText.hashCode ^ _expectedBitmapSize.hashCode;
 }
 
 class ImageComponentBuilder implements ComponentBuilder {
@@ -112,8 +102,8 @@ class ImageComponentBuilder implements ComponentBuilder {
     return ImageComponentViewModel(
       nodeId: node.id,
       imageUrl: node.imageUrl,
-      width: node.width,
-      height: node.height,
+      width: node.expectedBitmapSize?.width?.toDouble(),
+      height: node.expectedBitmapSize?.height?.toDouble(),
       selectionColor: const Color(0x00000000),
     );
   }
@@ -245,4 +235,26 @@ class ImageComponent extends StatelessWidget {
       ),
     );
   }
+}
+
+/// A size with optional [width] and [height].
+class ExpectedSize {
+  ExpectedSize({
+    this.height,
+    this.width,
+  });
+
+  final int? width;
+  final int? height;
+
+  @override
+  bool operator ==(Object other) =>
+      identical(this, other) ||
+      other is ExpectedSize && //
+          runtimeType == other.runtimeType &&
+          width == other.width &&
+          height == other.height;
+
+  @override
+  int get hashCode => width.hashCode ^ height.hashCode;
 }
--- a/super_editor/lib/src/default_editor/image.dart
+++ b/super_editor/lib/src/default_editor/image.dart
@@ -11,10 +11,14 @@ class ImageNode extends BlockNode with ChangeNotifier {
   ImageNode({
     required this.id,
     required String imageUrl,
+    double? width,
+    double? height,
     String altText = '',
     Map<String, dynamic>? metadata,
   })  : _imageUrl = imageUrl,
-        _altText = altText {
+        _altText = altText,
+        _width = width,
+        _height = height {
     this.metadata = metadata;
 
     putMetadataValue("blockType", const NamedAttribution("image"));
@@ -41,6 +45,28 @@ class ImageNode extends BlockNode with ChangeNotifier {
     }
   }
 
+  double? _width;
+  double? get width => _width;
+  set width(double? newWidth) {
+    if (newWidth == _width) {
+      return;
+    }
+
+    _width = newWidth;
+    notifyListeners();
+  }
+
+  double? _height;
+  double? get height => _height;
+  set height(double? newHeight) {
+    if (newHeight == _height) {
+      return;
+    }
+
+    _height = newHeight;
+    notifyListeners();
+  }
+
   @override
   String? copyContent(dynamic selection) {
     if (selection is! UpstreamDownstreamNodeSelection) {
@@ -52,7 +78,11 @@ class ImageNode extends BlockNode with ChangeNotifier {
 
   @override
   bool hasEquivalentContent(DocumentNode other) {
-    return other is ImageNode && imageUrl == other.imageUrl && altText == other.altText;
+    return other is ImageNode &&
+        imageUrl == other.imageUrl &&
+        width == other.width &&
+        height == other.height &&
+        altText == other.altText;
   }
 
   @override
@@ -62,10 +92,12 @@ class ImageNode extends BlockNode with ChangeNotifier {
           runtimeType == other.runtimeType &&
           id == other.id &&
           _imageUrl == other._imageUrl &&
+          _width == other.width &&
+          _height == other.height &&
           _altText == other._altText;
 
   @override
-  int get hashCode => id.hashCode ^ _imageUrl.hashCode ^ _altText.hashCode;
+  int get hashCode => id.hashCode ^ _imageUrl.hashCode ^ _altText.hashCode ^ _width.hashCode ^ _height.hashCode;
 }
 
 class ImageComponentBuilder implements ComponentBuilder {
@@ -80,6 +112,8 @@ class ImageComponentBuilder implements ComponentBuilder {
     return ImageComponentViewModel(
       nodeId: node.id,
       imageUrl: node.imageUrl,
+      width: node.width,
+      height: node.height,
       selectionColor: const Color(0x00000000),
     );
   }
@@ -94,6 +128,8 @@ class ImageComponentBuilder implements ComponentBuilder {
     return ImageComponent(
       componentKey: componentContext.componentKey,
       imageUrl: componentViewModel.imageUrl,
+      width: componentViewModel.width,
+      height: componentViewModel.height,
       selection: componentViewModel.selection,
       selectionColor: componentViewModel.selectionColor,
     );
@@ -106,11 +142,15 @@ class ImageComponentViewModel extends SingleColumnLayoutComponentViewModel {
     double? maxWidth,
     EdgeInsetsGeometry padding = EdgeInsets.zero,
     required this.imageUrl,
+    this.width,
+    this.height,
     this.selection,
     required this.selectionColor,
   }) : super(nodeId: nodeId, maxWidth: maxWidth, padding: padding);
 
   String imageUrl;
+  double? width;
+  double? height;
   UpstreamDownstreamNodeSelection? selection;
   Color selectionColor;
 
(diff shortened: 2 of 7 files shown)


super_editor/example/lib/demos/example_editor/_example_document.dart
```diff
@@ -7,6 +7,7 @@ MutableDocument createInitialDocument() {
       ImageNode(
         id: "1",
         imageUrl: 'https://i.ibb.co/5nvRdx1/flutter-horizon.png',
+        height: 422,
```
matthew-carroll, line 10: While it should be allowable to provide only one dimension, in general, we should expect both dimensions to be provided.

super_editor/lib/src/default_editor/image.dart
```diff
@@ -11,10 +11,14 @@ class ImageNode extends BlockNode with ChangeNotifier {
   ImageNode({
     required this.id,
     required String imageUrl,
+    double? width,
```
matthew-carroll, line 14: Let's call these `expectedBitmapWidth` and `expectedBitmapHeight`.

super_editor/lib/src/default_editor/image.dart
```diff
@@ -11,10 +11,14 @@ class ImageNode extends BlockNode with ChangeNotifier {
...
+    double? height,
     String altText = '',
     Map<String, dynamic>? metadata,
   })  : _imageUrl = imageUrl,
-        _altText = altText {
+        _altText = altText,
```
matthew-carroll, line 19: Please try to keep assignments in the same order as the parameters.

super_editor/lib/src/default_editor/image.dart
```diff
@@ -41,6 +45,28 @@ class ImageNode extends BlockNode with ChangeNotifier {
     }
   }
 
+  double? _width;
```
matthew-carroll, line 48: Please put the property between the accessor methods, and please add Dart Docs

super_editor/lib/src/default_editor/image.dart
```diff
@@ -41,6 +45,28 @@ class ImageNode extends BlockNode with ChangeNotifier {
...
+  double? get height => _height;
+  set height(double? newHeight) {
+    if (newHeight == _height) {
+      return;
+    }
+
+    _height = newHeight;
+    notifyListeners();
+  }
+
```
matthew-carroll, line 69: We should probably have a method to set both width and height at the same time, given that listeners are notified when these methods are called.

super_editor/lib/src/default_editor/image.dart
```diff
@@ -52,7 +78,11 @@ class ImageNode extends BlockNode with ChangeNotifier {
 
   @override
   bool hasEquivalentContent(DocumentNode other) {
-    return other is ImageNode && imageUrl == other.imageUrl && altText == other.altText;
+    return other is ImageNode &&
+        imageUrl == other.imageUrl &&
+        width == other.width &&
```
matthew-carroll, line 83: I'm wondering if the expected width and height really impact content equality. Based on use-cases do you think they should be included?

super_editor/lib/src/default_editor/image.dart
```diff
@@ -182,6 +236,8 @@ class ImageComponent extends StatelessWidget {
                   : Image.network(
                       imageUrl,
                       fit: BoxFit.contain,
+                      width: width,
```
matthew-carroll, line 239: I'm guessing that the `width` and `height` property of `Image` probably isn't what we want here.

The purpose of the width and height (at least for this ticket) is to offer an expected or assumed bitmap size so that the layout can take up the appropriate space before the image actually loads and appears. This means that the given `width` and `height` shouldn't impact the display of the image, they should only impact the display of an empty box while the image is being loaded.

super_editor_markdown/lib/src/document_to_markdown_serializer.dart
```diff
@@ -80,11 +80,34 @@ abstract class NodeTypedDocumentNodeMarkdownSerializer<NodeType> implements Docu
...
+    StringBuffer sizeNotation = StringBuffer();
+    sizeNotation.write(' =');
+
+    if (node.width != null) {
+      sizeNotation.write(node.width!.toInt());
+    }
+
+    sizeNotation.write('x');
+
+    if (node.height != null) {
```
matthew-carroll, line 106: What if only one of these values is non-null? Will we actually produce a desirable encoding?

super_editor_markdown/lib/src/document_to_markdown_serializer.dart
```diff
@@ -80,11 +80,34 @@ abstract class NodeTypedDocumentNodeMarkdownSerializer<NodeType> implements Docu
...
+      // We don't want to use size notation or the image doesn't have
+      // size information. Use the regular syntax.
+      return '![${node.altText}](${node.imageUrl})';
+    }
+
+    StringBuffer sizeNotation = StringBuffer();
+    sizeNotation.write(' =');
+
+    if (node.width != null) {
+      sizeNotation.write(node.width!.toInt());
```
matthew-carroll, line 101: If width and height can only be integers, then should those properties on the node be integers, too?

super_editor_markdown/lib/src/image_syntax.dart
```diff
@@ -0,0 +1,744 @@
+import 'package:markdown/markdown.dart' as md;
+
+/// Matches images like `![alternate text](url optional_size "optional title")` and
```
matthew-carroll, line 3: Does this match all images? Or just images like the two examples listed here?

super_editor_markdown/lib/src/image_syntax.dart
```diff
@@ -0,0 +1,744 @@
...
+/// `![alternate text][label]`.
+///
+/// To define a size, use the notation `=widthxheight`. The size notation is optional and
+/// it can be providade partially. For example:
+///
+/// - ![alternate text](url =500x200)
+/// - ![alternate text](url =500x)
+/// - ![alternate text](url =x200)
+///
+/// Extracted and adapted from the markdown package.
```
matthew-carroll, line 13: Which parts were extracted and adapted from the markdown package?

super_editor_markdown/lib/src/image_syntax.dart
```diff
@@ -0,0 +1,744 @@
...
+
+    // The link text (inside `[...]`) was not followed with a opening `(` nor
+    // an opening `[`. Perhaps just a simple shortcut reference link (`[...]`).
+    return _tryCreateReferenceLink(parser, text, getChildren: getChildren);
+  }
+
+  /// Parses a size using the notation `=widthxheight`.
+  ///
+  /// Returns `null` if the size notation isn't provided.
+  PreferredSize? _tryParseImageSize(md.InlineParser parser) {
```
matthew-carroll, line 89: After seeing `PreferredSize` here, should we use that type for the `ImageNode` too?

super_editor_markdown/lib/src/image_syntax.dart
```diff
@@ -0,0 +1,744 @@
...
+
+  /// Character `}`.
+  static const int rightBrace = 0x7D;
+
+  /// Character `~`.
+  static const int tilde = 0x7E;
+}
+
+/// A size with optional [width] and [height].
+class PreferredSize {
```
matthew-carroll, line 725: Let's move this down below `MarkdownImage`

super_editor_markdown/test/super_editor_markdown_test.dart
```diff
@@ -453,6 +453,46 @@ Paragraph3""");
...
+          ImageNode(
+            id: '1',
+            imageUrl: 'https://someimage.com/the/image.png',
+            altText: 'some alt text',
+            height: 200.0,
+          ),
+        ]);
+
+        expect(serializeDocumentToMarkdown(doc), '![some alt text](https://someimage.com/the/image.png =x200)');
+      });
```
matthew-carroll, line 494: Do we have a test for an image without a size?

super_editor_markdown/lib/src/image_syntax.dart
angelosilvestre: It makes sense to use it on the `ImageNode`.

angelosilvestre: I notice there is already a  `PreferredSize` widget in flutter. Should we call this `DesiredSize`?

super_editor_markdown/test/super_editor_markdown_test.dart
angelosilvestre: Yes, we have one before 'image with size'.

super_editor/lib/src/default_editor/image.dart
angelosilvestre: If want that changing the size causes the image to rebuild, then we should include it.

I think it's possible that someone would want to be able to change the size during runtime, so I think we should include it.

super_editor_markdown/lib/src/document_to_markdown_serializer.dart
angelosilvestre: Yes, we should allow providing only one of these.

super_editor_markdown/lib/src/document_to_markdown_serializer.dart
angelosilvestre: Updated.

super_editor/lib/src/default_editor/image.dart
angelosilvestre: But the image can be displayed with a different height depending on the size of screen (or the size of its parent). So, even if we use a pre-configured `height`, it might be different from the actual height that the image will take.

super_editor/lib/src/default_editor/image.dart
angelosilvestre: Updated.

super_editor_markdown/lib/src/image_syntax.dart
angelosilvestre: I think this is all of the existing syntaxes for images.

super_editor_markdown/lib/src/image_syntax.dart
angelosilvestre: Basically, I extracted all of the `LinkSyntax` parsing logic and just changed to handle the ` =widthxheight` at the end.

super_editor/lib/src/default_editor/image.dart
matthew-carroll: Can you think of such a situation? This width and height don't impact the actual size of the final image. They only impact the expected size of the image while it's loading. Is the concern that this might change during loading?

super_editor/lib/src/default_editor/image.dart
matthew-carroll: I didn't understand that response. I'm not sure what you were trying to say, or how it relates to my comment above.

super_editor_markdown/lib/src/image_syntax.dart
matthew-carroll: Then let's make that clear. Please say `This class was modified from a copy of [LinkSyntax].`

super_editor_markdown/lib/src/image_syntax.dart
matthew-carroll: Let's go with `ExpectedSize`

super_editor/lib/src/default_editor/image.dart
angelosilvestre: The image resizes itself to try to maintain the aspect ratio, so it doesn't necessarily will be the same as the expected size.

We can also use an `AspectRatio` when a expected size is given, but then we need to require that both width and height are provided.

super_editor/lib/src/default_editor/image.dart
angelosilvestre: At the beginning I thought it would impact the actual size. If it's only while the image is loading then we don't need to include it.

super_editor/lib/src/default_editor/image.dart
matthew-carroll: I don't think this PR needs to impact the size or aspect ratio of the final image display.

angelosilvestre: I mean to use an `AspectRation` for the `SizedBox` which will be displayed while the image is loading.

matthew-carroll: Yeah, an `AspectRatio` makes sense when both width and height are offered. I'm not sure what the right answer is for when only one dimension is provided.

angelosilvestre: When only dimension is provided, I think the only option is to just use a `SizedBox` and possibly (or probably) still shift the content after the image finishes loading.

super_editor/lib/src/default_editor/image.dart
```diff
@@ -32,6 +34,27 @@ class ImageNode extends BlockNode with ChangeNotifier {
     }
   }
 
+  /// The expected size of the image.
+  ///
+  /// Used to size the component while the image is still being loaded,
+  /// so the content don't shift after the image is loaded.
+  ///
+  /// Although is allowed to specify only one dimension, it's preferable
```
matthew-carroll, line 42: Let's replace this paragraph with:

```
/// It's technically permissible to provide only a single expected dimension,
/// however providing only a single dimension won't provide enough information
/// to size an image component before the image is loaded. Providing only a
/// width in a vertical layout won't have any visual effect. Providing only a height
/// in a vertical layout will likely take up more space or less space than the final
/// image because the final image will probably be scaled. Therefore, to take
/// advantage of [ExpectedSize], you should try to provide both dimensions.
```

super_editor/lib/src/default_editor/image.dart
```diff
@@ -52,7 +75,10 @@ class ImageNode extends BlockNode with ChangeNotifier {
 
   @override
   bool hasEquivalentContent(DocumentNode other) {
-    return other is ImageNode && imageUrl == other.imageUrl && altText == other.altText;
+    return other is ImageNode &&
+        imageUrl == other.imageUrl &&
+        expectedBitmapSize == other.expectedBitmapSize &&
```
matthew-carroll, line 80: Is this size really an important aspect of content equivalency?

super_editor/lib/src/default_editor/image.dart
```diff
@@ -106,11 +137,15 @@ class ImageComponentViewModel extends SingleColumnLayoutComponentViewModel {
...
     EdgeInsetsGeometry padding = EdgeInsets.zero,
     required this.imageUrl,
+    this.width,
+    this.height,
     this.selection,
     required this.selectionColor,
   }) : super(nodeId: nodeId, maxWidth: maxWidth, padding: padding);
 
   String imageUrl;
+  double? width;
```
matthew-carroll, line 147: Is there a reason that we're breaking down the `ExpectedSize` into `width` and `height`? Doesn't this data still represent an expected size? Why not pass `ExpectedSize` through?

super_editor/lib/src/default_editor/image.dart
```diff
@@ -190,3 +253,25 @@ class ImageComponent extends StatelessWidget {
     );
   }
 }
+
+/// A size with optional [width] and [height].
+class ExpectedSize {
+  ExpectedSize({
+    this.height,
```
matthew-carroll, line 260: Let's replace the named parameters with unnamed parameters: `(int width, int height)`

Please make this a `const` constructor.

super_editor/lib/src/default_editor/image.dart
```diff
@@ -190,3 +253,25 @@ class ImageComponent extends StatelessWidget {
     );
   }
 }
+
+/// A size with optional [width] and [height].
```
matthew-carroll, line 257: This Dart Doc doesn't capture the motivation for why we're creating this class.

super_editor/lib/src/default_editor/image.dart
```diff
@@ -190,3 +253,25 @@ class ImageComponent extends StatelessWidget {
...
+class ExpectedSize {
+  ExpectedSize({
+    this.height,
+    this.width,
+  });
+
+  final int? width;
+  final int? height;
+
```
matthew-carroll, line 266: You might want to add an `aspectRatio` getter to avoid needing to recompute that data with the  width and height elsewhere.

super_editor_markdown/lib/src/image_syntax.dart
```diff
@@ -0,0 +1,735 @@
...
+  /// Character `}`.
+  static const int rightBrace = 0x7D;
+
+  /// Character `~`.
+  static const int tilde = 0x7E;
+}
+
+/// A parsed image notation.
+class MarkdownImage extends md.InlineLink {
+  MarkdownImage(
```
matthew-carroll, line 728: Please make sure to define all constructors as `const` where possible.

super_editor_markdown/lib/src/markdown_to_document_parsing.dart
```diff
@@ -123,6 +131,8 @@ class _MarkdownToDocument implements md.NodeVisitor {
             // TODO: handle null image URL
             imageUrl: inlineVisitor.imageUrl!,
             altText: inlineVisitor.imageAltText!,
+            width: inlineVisitor.width,
```
matthew-carroll, line 134: Is there any reason to call these `width` and `height` here instead of `expectedBitmapSize`?

super_editor/lib/src/default_editor/image.dart
angelosilvestre: Nope, removed.

super_editor/lib/src/default_editor/image.dart
angelosilvestre: Changed it to use `ExpectedSize`.

super_editor/lib/src/default_editor/image.dart
angelosilvestre: Should the `aspectRatio` getter also be nullable? Or should we assume that it will only be called after checking if both properties are non-null?

matthew-carroll: I think it can be non-null. Somtimes it might return `0.0` and sometimes it might return `double.infinity`, but I think we have a value no matter what. Or perhaps instead of infinity we want to throw some kind of NaN exception. I'm not sure what the typical dart approach is to division by zero.

super_editor/lib/src/default_editor/image.dart
angelosilvestre: Updated

angelosilvestre: Updated.

matthew-carroll (approved):
LGTM

PR #1744, opened 2024-01-03
Make the use and purpose of getRectForPosition() to be consistent across the codebase (Resolves #1614)

### LINKED ISSUE #1614: [SuperEditor][SuperReader] - Resolve differences between getRectForPosition and getRectForSelection
`TextComponent` implements `getRectForPosition()` and `getRectForSelection()`. Both of these methods include behavior related to getting a rect for a collapsed position, however each one includes different calculation behavior.

I think the history of these two methods is that for a long time we only ever used `getRectForPosition()` when we had a collapsed selection. However, there were situations where `getRectForPosition()` would return a `zero` `Rect` and we expected a non-zero result. That's when we added a special collapsed case to `getRectForSelection()`.

The problem now is that it's not clear why these two collapsed calculations are different. I just got done debugging a tricky test failure and it turns out the problem was that one area of code was using `getRectForPosition()` to calculate the caret height, and another execution path was using `getRectForSelection()` to calculate the caret height. This resulted in a small disparity in height calculation - a difference of about 1.8px - but that was enough to fail tests. (See #1607)

Revisit those two methods. Determine whether or not there should be a difference in collapsed `Rect` calculations. If there should be a difference, document that difference in the Dart Docs and check all uses to make sure each one is used in the right place. If there shouldn't be a difference, then move the `getRectForSelection()` behavior up into `getRectForPosition()`, then have `getRectForSelection()` defer to `getRectForPosition()` within its implementation.

matthew-carroll:
Make the use and purpose of getRectForPosition() to be consistent across the codebase (Resolves #1614)

It seemed that sometimes `getRectForPosition()` was being used to obtain a bounding box and other times to obtain a leading/trailing edge of content. This PR resolves this inconsistency by always returning a content bounding box with `getRectForPosition()`.

Most of the changes in this PR are changes of code that uses `getRectForPosition()` and `getRectForSelection()`, but one foundational change is that `BoxComponent`s now return their bounding box from `getRectForPosition()` regardless of whether the given position is `upstream` or `downstream`.

I discovered along the way that our `HoriziontalRuleComponent` and `UnselectableHorizontalRuleComponent` didn't include `IgnorePointer`s. I'm not sure why this didn't break any tests previously, but it did break some tests in the course of this PR, so I added `IgnorePointer`s.

I also added some test tools in this PR, as needed.


matthew-carroll:
@angelosilvestre can you show me again where I can find the CI golden failure files?

angelosilvestre:
@matthew-carroll You can go to the "Checks" tab and then click on the workflow. There will be a zip file at the end of the page.

super_editor/test/super_editor/supereditor_caret_test.dart
```diff
@@ -110,24 +111,21 @@ void main() {
...
-        final tapOffset = _getOffsetForPosition(docKey, tapPosition);
-        await tester.tapAt(tapOffset);
-        await tester.pumpAndSettle();
+        await tester.tapAtDocumentPosition(tapPosition);
+        await tester.pump();
 
         // Ensure that the caret is displayed at the correct (x,y) in the document before resizing the window
         final initialCaretOffset = SuperEditorInspector.findCaretOffsetInDocument();
-        final expectedInitialCaretOffset = _computeExpectedDesktopCaretOffset(tester, textPosition);
-        expect(initialCaretOffset, expectedInitialCaretOffset);
```
angelosilvestre, line 120: The `expect` was removed but the comment was kept.

super_editor/test/super_editor/supereditor_caret_test.dart
```diff
@@ -146,24 +144,21 @@ void main() {
...
```
angelosilvestre, line 156: Same here.

super_editor/test/super_editor/supereditor_caret_test.dart
```diff
@@ -188,26 +183,23 @@ void main() {
...
-          await tester.tapAt(tapOffset);
-          await tester.pumpAndSettle();
+          await tester.tapAtDocumentPosition(tapPosition);
+          await tester.pump();
 
           // Ensure that the caret is displayed at the correct (x,y) in the document before phone rotation
           final initialCaretOffset = SuperEditorInspector.findCaretOffsetInDocument();
-          final expectedInitialCaretOffset =
-              _computeExpectedMobileCaretOffsetInDocumentLayout(tester, docKey, tapPosition);
-          expect(initialCaretOffset, expectedInitialCaretOffset);
```
angelosilvestre, line 199: And here.

super_editor/test_goldens/editor/mobile/mobile_selection_test.dart
```diff
@@ -938,6 +719,19 @@ MutableDocument _createSingleParagraphDoc() {
   );
 }
 
+/// Calculates the delta between the center of the character at [startOffset] and and the
+/// center of the character at [endOffset] within the node with the given [nodeId].
+Offset _calculateDragDelta(GlobalKey docKey, String nodeId, int startOffset, int endOffset) {
```
angelosilvestre, line 724: Would it make sense to add a robot method for dragging between two offsets in a text node?

angelosilvestre (approved):
LGTM with a few comments.

super_editor/test_goldens/editor/mobile/mobile_selection_test.dart
matthew-carroll: I added a method to the `SuperEditorInspector` to calculate the delta, but I didn't add the drag method to the robot. I think that probably deserves a bit more thought. Some of these methods drag mobile handles and others drag the caret. We should think about what behaviors we should expose and how stable that logic would be that finds the handles or caret.

matthew-carroll:
@angelosilvestre I added a new document layout and component layout method called `getEdgeForPosition()` which returns the edge of the bounding box for the given position. That method returns the left side or right side of a box component depending on `upstream` or `downstream`.

super_editor/lib/src/default_editor/text.dart
```diff
@@ -569,6 +570,19 @@ class TextComponentState extends State<TextComponent> with DocumentComponent imp
...
+  Rect getEdgeForPosition(NodePosition nodePosition) {
+    if (nodePosition is! TextPosition) {
+      throw Exception('Expected nodePosition of type TextPosition but received: $nodePosition');
+    }
+
+    final characterBox = getRectForPosition(nodePosition);
+
+    // Always return the upstream edge because `getRectForPosition()` returns the box around
+    // the character immediately after the given position.
+    return characterBox.leftEdge;
```
angelosilvestre, line 583: Why `getRectForPosition` returns the box for the character after the given position? Shouldn't it return the box for the given position?

matthew-carroll: I suppose it depends whether you think of the position as a character or a caret. That statement was in terms of a caret. We return the leading edge of the character immediately after the caret. Do you think we should be doing something else?

angelosilvestre: Ok, understood that this statement is  in terms of a character. Initially I thought a bit confusing, because I thought the comment is saying that whatever position you give to `getRectForPosition()`, it would return the character box for the character after the given position.

matthew-carroll: I'll update the comment to be more clear

matthew-carroll: Actually I went ahead and respected the affinity. I didn't notice any breakages and no tests failed. But let's keep an eye out for weird caret placement moving forward.

matthew-carroll:
FYI @angelosilvestre @rutvik110 I was able to get goldens passing again in this PR. My final change was to use smaller golden file sizes that still fit all the content, but aren't nearly as large as the default size.

I don't know why that worked. I still don't know why CI thought goldens were failing. When I downloaded the goldens they were identical as far as I could tell. Nonetheless, they're passing again.

Let's try to keep the goldens passing moving forward.

PR #1489, opened 2023-10-07
[SuperEditor][SuperReader][iOS][Android] - Long-press content selection (Resolves #1203)

matthew-carroll:
[SuperEditor][SuperReader][iOS][Android] - Long-press content selection (Resolves #1203)

On iOS, implemented long-press drag selection word-by-word. Unlike Android, iOS always selects by the word while long-pressing.

There are some magnifier glitches in this demo, but I don't think it's related to this work.

https://github.com/superlistapp/super_editor/assets/7259036/f9128c21-b465-4730-8fe5-f541b3fe757e

On Android, implemented long-press drag selection, which goes word-by-word when moving away from the initial selection, and then selects character-by-character when moving back towards the initial selection.

https://github.com/superlistapp/super_editor/assets/7259036/9c24f010-a5b1-479d-b4f7-b524a3c435c7

PR #1745, opened 2024-01-03
[SuperEditor][SuperTextField][Android] Show toolbar upon tap on collapsed handle (Resolves #1737)

### LINKED ISSUE #1737: [Android] Tapping on drag handle while selection is collapsed does not show toolbar
**Package Version**
main branch with commit hash https://github.com/superlistapp/super_editor/commit/5304a3c2294dfa7ae2742539a5e9f2d3e05a3d16.

**To Reproduce**
1. Go to SuperEditor or Android SuperTextField demo
2. Tap one time on the content => collapsed selection is placed where the tap happened
3. Tap again on the drag handle

**Actual behavior**
Nothing happens.

**Expected behavior**
The toolbar is shown above with the action "Select all" and alternatively the action "Paste". 
The "Paste" action is the more relevant since it's a way for users to paste content on a specific offset.
The expected behavior here described is based on how "Google Keep" works.

**Platform**
Android.

**Flutter version**

**Screenshots**

https://github.com/superlistapp/super_editor/assets/31278849/a5d0f4ce-01ae-4d33-8c09-7ac509bb09ed

angelosilvestre:
[SuperEditor][SuperTextField][Android] Show toolbar upon tap on collapsed handle. Resolves #1737

On Android, tapping on the collapsed drag handle doesn't show the popover toolbar.

This PR changes the editor/textfield to show a toolbar containing only the options "Select All" and "Paste" after tapping on the collapsed handle:

[toolbar.webm](https://github.com/superlistapp/super_editor/assets/7597082/b9b66419-3575-41e9-b3eb-5882e620ca73)

Currently, the toolbar builders don't receive the document selection, so I'm using a workaround right now. We need to decide how we want to provide this information to the toolbar builders (for both the editor and the textfield) before merging this PR.

### DIFF
--- a/super_editor/lib/src/default_editor/super_editor.dart
+++ b/super_editor/lib/src/default_editor/super_editor.dart
@@ -1,5 +1,5 @@
 import 'package:attributed_text/attributed_text.dart';
-import 'package:flutter/foundation.dart' show defaultTargetPlatform;
+import 'package:flutter/foundation.dart' show ValueListenable, defaultTargetPlatform;
 import 'package:flutter/material.dart' hide SelectableText;
 import 'package:flutter/services.dart';
 import 'package:follow_the_leader/follow_the_leader.dart';
@@ -8,6 +8,7 @@ import 'package:super_editor/src/core/document_composer.dart';
 import 'package:super_editor/src/core/document_debug_paint.dart';
 import 'package:super_editor/src/core/document_interaction.dart';
 import 'package:super_editor/src/core/document_layout.dart';
+import 'package:super_editor/src/core/document_selection.dart';
 import 'package:super_editor/src/core/edit_context.dart';
 import 'package:super_editor/src/core/editor.dart';
 import 'package:super_editor/src/core/styles.dart';
@@ -756,6 +757,7 @@ class SuperEditorState extends State<SuperEditor> {
             mobileToolbarKey,
             editContext.commonOps,
             SuperEditorAndroidControlsScope.rootOf(context),
+            editContext.composer.selectionNotifier,
           ),
           child: child,
         );
@@ -875,11 +877,13 @@ Widget defaultAndroidEditorToolbarBuilder(
   Key floatingToolbarKey,
   CommonEditorOperations editorOps,
   SuperEditorAndroidControlsController editorControlsController,
+  ValueListenable<DocumentSelection?> selectionNotifier,
 ) {
   return DefaultAndroidEditorToolbar(
     floatingToolbarKey: floatingToolbarKey,
     editorOps: editorOps,
     editorControlsController: editorControlsController,
+    selectionNotifier: selectionNotifier,
   );
 }
 
@@ -890,20 +894,22 @@ class DefaultAndroidEditorToolbar extends StatelessWidget {
     this.floatingToolbarKey,
     required this.editorOps,
     required this.editorControlsController,
+    required this.selectionNotifier,
   });
 
   final Key? floatingToolbarKey;
   final CommonEditorOperations editorOps;
   final SuperEditorAndroidControlsController editorControlsController;
+  final ValueListenable<DocumentSelection?> selectionNotifier;
 
   @override
   Widget build(BuildContext context) {
     return AndroidTextEditingFloatingToolbar(
       floatingToolbarKey: floatingToolbarKey,
-      onCopyPressed: editorOps.composer.selection == null || !editorOps.composer.selection!.isCollapsed //
+      onCopyPressed: selectionNotifier.value == null || !selectionNotifier.value!.isCollapsed //
           ? _copy
           : null,
-      onCutPressed: editorOps.composer.selection == null || !editorOps.composer.selection!.isCollapsed //
+      onCutPressed: selectionNotifier.value == null || !selectionNotifier.value!.isCollapsed //
           ? _cut
           : null,
       onPastePressed: _paste,
--- a/super_editor/lib/src/default_editor/super_editor.dart
+++ b/super_editor/lib/src/default_editor/super_editor.dart
@@ -900,8 +900,12 @@ class DefaultAndroidEditorToolbar extends StatelessWidget {
   Widget build(BuildContext context) {
     return AndroidTextEditingFloatingToolbar(
       floatingToolbarKey: floatingToolbarKey,
-      onCopyPressed: _copy,
-      onCutPressed: _cut,
+      onCopyPressed: editorOps.composer.selection == null || !editorOps.composer.selection!.isCollapsed //
+          ? _copy
+          : null,
+      onCutPressed: editorOps.composer.selection == null || !editorOps.composer.selection!.isCollapsed //
+          ? _cut
+          : null,
       onPastePressed: _paste,
       onSelectAllPressed: _selectAll,
     );
--- a/super_editor/lib/src/default_editor/super_editor.dart
+++ b/super_editor/lib/src/default_editor/super_editor.dart
@@ -904,16 +904,21 @@ class DefaultAndroidEditorToolbar extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    return AndroidTextEditingFloatingToolbar(
-      floatingToolbarKey: floatingToolbarKey,
-      onCopyPressed: selectionNotifier.value == null || !selectionNotifier.value!.isCollapsed //
-          ? _copy
-          : null,
-      onCutPressed: selectionNotifier.value == null || !selectionNotifier.value!.isCollapsed //
-          ? _cut
-          : null,
-      onPastePressed: _paste,
-      onSelectAllPressed: _selectAll,
+    return ValueListenableBuilder(
+      valueListenable: selectionNotifier,
+      builder: (context, selection, child) {
+        return AndroidTextEditingFloatingToolbar(
+          floatingToolbarKey: floatingToolbarKey,
+          onCopyPressed: selection == null || !selection.isCollapsed //
+              ? _copy
+              : null,
+          onCutPressed: selection == null || !selection.isCollapsed //
+              ? _cut
+              : null,
+          onPastePressed: _paste,
+          onSelectAllPressed: _selectAll,
+        );
+      },
     );
   }
 
--- a/super_editor/lib/src/infrastructure/touch_controls.dart
+++ b/super_editor/lib/src/infrastructure/touch_controls.dart
@@ -16,14 +16,10 @@ enum HandleType {
 class ToolbarConfig {
   ToolbarConfig({
     required this.focalPoint,
-    this.collapsedSelection = false,
   });
 
   /// The desired point where a toolbar arrow should point to.
   ///
   /// Represented as global coordinates.
   final Offset focalPoint;
-
-  // TODO: change this.
-  final bool collapsedSelection;
 }
--- a/super_editor/lib/src/infrastructure/touch_controls.dart
+++ b/super_editor/lib/src/infrastructure/touch_controls.dart
@@ -16,10 +16,14 @@ enum HandleType {
 class ToolbarConfig {
   ToolbarConfig({
     required this.focalPoint,
+    this.collapsedSelection = false,
   });
 
   /// The desired point where a toolbar arrow should point to.
   ///
   /// Represented as global coordinates.
   final Offset focalPoint;
+
+  // TODO: change this.
+  final bool collapsedSelection;
 }
--- a/super_editor/lib/src/super_textfield/android/_user_interaction.dart
+++ b/super_editor/lib/src/super_textfield/android/_user_interaction.dart
@@ -176,6 +176,8 @@ class AndroidTextFieldTouchInteractorState extends State<AndroidTextFieldTouchIn
   void _onTapUp(TapUpDetails details) {
     _log.fine('User released a tap');
 
+    final previousSelection = widget.textController.selection;
+
     _selectAtOffset(details.localPosition);
 
     if (widget.focusNode.hasFocus && widget.textController.isAttachedToIme) {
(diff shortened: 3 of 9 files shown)


super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -900,8 +900,12 @@ class DefaultAndroidEditorToolbar extends StatelessWidget {
   Widget build(BuildContext context) {
     return AndroidTextEditingFloatingToolbar(
       floatingToolbarKey: floatingToolbarKey,
-      onCopyPressed: _copy,
-      onCutPressed: _cut,
+      onCopyPressed: editorOps.composer.selection == null || !editorOps.composer.selection!.isCollapsed //
```
matthew-carroll, line 903: I'm noticing that the toolbars know how to take editor actions. They use `editorOps`. Can we add a `ValueNotifier<DocumentSelection?>` to the property list for the toolbars to make the decision about which buttons to show?

angelosilvestre: This toolbar is also used for `SuperTextField`, so we can't take anything specific to documents.

matthew-carroll: If that's true, then how did `editorOps` get in here? Those are specific to `SuperEditor`...

angelosilvestre: `DefaultAndroidEditorToolbar` is only used by `SuperEditor`, but `AndroidTextEditingFloatingToolbar` is also used by `SuperTextField`.

matthew-carroll: Ok, help me out here. Without needing to go read all of that code, how are editor operations getting triggered by a toolbar that might be used for both `SuperEditor` and `SuperTextField`.

angelosilvestre: The `AndroidTextEditingFloatingToolbar` receives callbacks for `onCut`, `onCopy`, etc.

matthew-carroll: Does `DefaultAndroidEditorToolbar` use an `AndroidTextEditingFloatingToolbar` internally?

angelosilvestre: Yes

matthew-carroll: Ok....so if `DefaultAndroidEditorToolbar` is specific to `SuperEditor`, then we can provide `DefaultAndroidEditorToolbar` with a `ValueNotifier<Selection>`, right? And then `DefaultAndroidEditorToolbar` can decide whether or not to pass callbacks based on that selection. Am I missing something?

angelosilvestre: Ok, updated.

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -890,18 +894,24 @@ class DefaultAndroidEditorToolbar extends StatelessWidget {
     this.floatingToolbarKey,
     required this.editorOps,
     required this.editorControlsController,
+    required this.selectionNotifier,
   });
 
   final Key? floatingToolbarKey;
   final CommonEditorOperations editorOps;
   final SuperEditorAndroidControlsController editorControlsController;
+  final ValueListenable<DocumentSelection?> selectionNotifier;
```
matthew-carroll, line 903: Do we need to listen to the selection changes or no?

super_editor/lib/src/infrastructure/touch_controls.dart
```diff
@@ -16,10 +16,14 @@ enum HandleType {
...
     required this.focalPoint,
+    this.collapsedSelection = false,
   });
 
   /// The desired point where a toolbar arrow should point to.
   ///
   /// Represented as global coordinates.
   final Offset focalPoint;
+
+  // TODO: change this.
```
matthew-carroll, line 27: Please do something with this TODO

super_editor/lib/src/super_textfield/android/android_textfield.dart
```diff
@@ -658,27 +658,31 @@ Widget _defaultAndroidToolbarBuilder(
...
-      Clipboard.setData(ClipboardData(text: selectedText));
-    },
-    onCopyPressed: () {
-      final textController = controller.textController;
-      final selection = textController.selection;
-      final selectedText = selection.textInside(textController.text.text);
-
-      Clipboard.setData(ClipboardData(text: selectedText));
-    },
+    onCutPressed: config.collapsedSelection
```
matthew-carroll, line 661: Can we move the implementation of these behaviors to private methods so that this widget tree isn't so difficult to read?

super_editor/test/super_editor/mobile/super_editor_android_overlay_controls_test.dart
```diff
@@ -63,6 +63,30 @@ void main() {
       await tester.pump(kTapMinTime);
     });
 
+    testWidgetsOnAndroid("shows and hides toolbar upon tap on collapsed handle", (tester) async {
```
matthew-carroll, line 66: Can we also add a test to see what happens when the user taps to show the toolbar, and then taps to place the caret somewhere else?

super_editor/lib/src/super_textfield/android/android_textfield.dart
angelosilvestre: Updated.

super_editor/test/super_editor/mobile/super_editor_android_overlay_controls_test.dart
angelosilvestre: Test added.

super_editor/lib/src/super_textfield/android/android_textfield.dart
```diff
@@ -657,44 +657,58 @@ Widget _defaultAndroidToolbarBuilder(
   AndroidEditingOverlayController controller,
   ToolbarConfig config,
 ) {
+  final hasCollapsedSelection = controller.textController.selection != TextRange.empty;
```
matthew-carroll, line 660: Is this calculation correct? Why does a non-empty selection mean that the selection is expanded?

Also, looking at the ternaries down below, I think it would be easier to read the condition of the action is in the first branch and `null` is the second branch.

```
onCutPressed: isSelectionExpanded
  ? () => _onToolbarCutPressed(controller)
  : null,
```

super_editor/lib/src/super_textfield/android/android_textfield.dart
```diff
@@ -657,44 +657,58 @@ Widget _defaultAndroidToolbarBuilder(
...
+    onPastePressed: () => _onToolbarPastePressed(controller),
+    onSelectAllPressed: () => _onToolbarSelectAllPressed(controller),
+  );
+}
 
-      final selection = textController.selection;
-      if (selection.isCollapsed) {
-        return;
-      }
+void _onToolbarCutPressed(AndroidEditingOverlayController controller) {
```
matthew-carroll, line 674: Please move these methods up above the build methods.

super_editor/test/super_editor/mobile/super_editor_android_overlay_controls_test.dart
```diff
@@ -63,6 +63,53 @@ void main() {
...
+
+      // Tap the drag handle to hide the toolbar.
+      await tester.tap(SuperEditorInspector.findMobileCaretDragHandle());
+      await tester.pump();
+
+      // Ensure the toolbar isn't visible.
+      expect(SuperEditorInspector.isMobileToolbarVisible(), isFalse);
+    });
+
+    testWidgetsOnAndroid("hides toolbar when placing the caret", (tester) async {
```
matthew-carroll, line 90: This is more about moving the caret. Let's call it "hides toolbar when the user taps to move the caret"

super_editor/test/super_textfield/android/super_textfield_android_selection_test.dart
```diff
@@ -54,6 +55,33 @@ void main() {
       // Ensure that no toolbar is displayed.
       expect(find.byType(AndroidTextEditingFloatingToolbar), findsNothing);
     });
+
+    testWidgetsOnAndroid("tapping at collapsed handle shows/hides the toolbar", (tester) async {
```
matthew-carroll, line 59: Should there also be a test for moving the caret, like the editor tests?

super_editor/lib/src/super_textfield/android/android_textfield.dart
angelosilvestre: Updated.

super_editor/test/super_textfield/android/super_textfield_android_selection_test.dart
angelosilvestre: Test added

angelosilvestre:
@matthew-carroll While I was writing a test for this comment https://github.com/superlistapp/super_editor/pull/1745#discussion_r1445535380 I noticed a weird failure, the toolbar was being displayed after placing a collapsed selection.

After some debugging I noticed a problem in `_onTapUp`. We have this code:

```dart
final previousSelection = widget.textController.selection;
final didTapOnExistingSelection = previousSelection.isCollapsed
    ? tapTextPosition == previousSelection.extent
    : tapTextPosition.offset >= previousSelection.start && tapTextPosition.offset <= previousSelection.end;
```

Although we are trying to inspect the previous selection, at this point of the code we already changed the selection.

Also, looking at the code suggests that tapping on an existing selection should toggle the toolbar, but it doesn't seems that this is happening.

matthew-carroll:
@angelosilvestre for your last comment about the bug you found, can you please phrase that as a question for me to answer, or as a description of how you fixed it? Providing only a description of a problem doesn't really give us a path forward.

angelosilvestre:
@matthew-carroll The question is, should I fix this issue in this PR? Or should it be a separate ticket?

matthew-carroll:
Please fix it here

angelosilvestre:
@matthew-carroll Pushed changes to fix the other issue. Here's some changes made:

We have the following code

```dart
void _onTapDown(TapDownDetails details) {
  ...

  // When the user drags, the toolbar should not be visible.
  // A drag can begin with a tap down, so we hide the toolbar
  // preemptively.
  widget.editingOverlayController.hideToolbar();

  ...
}
```

We are always hiding the toolbar on tap down. I removed this code and tested drag selection and the toolbar correctly disappears. 

We also have some code the **should** be toggling the toolbar even when we tap on an expanded selection. However, I noticed that in `SuperEditor`, tapping on an expanded selection places the caret at the tap location instead of toggling the toolbar. I modified to have the same behavior. If this is really what we want, we could simplify the following code to look just for collapsed selections.

```dart
final didTapOnExistingSelection = previousSelection.isCollapsed
    ? tapTextPosition == previousSelection.extent
    : tapTextPosition.offset >= previousSelection.start && tapTextPosition.offset <= previousSelection.end;
```

matthew-carroll:
> We are always hiding the toolbar on tap down. I removed this code and tested drag selection and the toolbar correctly disappears.

Did you test long press dragging, too?

> However, I noticed that in SuperEditor, tapping on an expanded selection places the caret at the tap location instead of toggling the toolbar.

Please always check the actual standard platform behavior when you find inconsistent behaviors. What happens on standard Android and iOS in these cases?

angelosilvestre:
> Did you test long press dragging, too?

Do you mean drag the caret to move the selection? If that's the case, it seems to be working correctly. What I've found is that long press to select a word doesn't seem to be implemented in `SuperTextField`.

> Please always check the actual standard platform behavior when you find inconsistent behaviors. What happens on standard Android and iOS in these cases?

Tested some apps and the standard behavior is the same implemented in `SuperEditor`. Tapping at an expanded selection places the caret, instead of toggling the toolbar.

matthew-carroll:
> Do you mean drag the caret to move the selection?

No, the caret isn't a long-press drag. I'm talking about longpressing to initially select a word, and then dragging from there.

> What I've found is that long press to select a word doesn't seem to be implemented in SuperTextField.

Can you please double check that? Do we have open tickets for that? Did we only implement it for `SuperEditor`? Text fields should definitely support it.

angelosilvestre:
> Can you please double check that? Do we have open tickets for that? Did we only implement it for SuperEditor? Text fields should definitely support it.

I double checked and it isn't implemented. The original PR only adds this gesture to `SuperEditor` and `SuperReader`: https://github.com/superlistapp/super_editor/pull/1489

I didn't find any open tickets for the textfield long-press.

super_editor/lib/src/super_textfield/android/_user_interaction.dart
```diff
@@ -160,24 +160,11 @@ class AndroidTextFieldTouchInteractorState extends State<AndroidTextFieldTouchIn
...
-      _log.finer("Field isn't focused. Ignoring press.");
-      return;
-    }
-
-    // When the user drags, the toolbar should not be visible.
-    // A drag can begin with a tap down, so we hide the toolbar
-    // preemptively.
-    widget.editingOverlayController.hideToolbar();
-  }
+  void _onTapDown(TapDownDetails details) {}
```
matthew-carroll, line 163: Why do we have an empty function?

super_editor/lib/src/super_textfield/android/android_textfield.dart
matthew-carroll: No, nevermind

super_editor/lib/src/super_textfield/android/_user_interaction.dart
angelosilvestre: I forgot to removed. Updated.

matthew-carroll (approved):
LGTM

PR #1754, opened 2024-01-06
[SuperEditor][Android] Fix drag handles when selection collapses due to dragging (Resolves #1736)

### LINKED ISSUE #1736: [SuperEditor] [Android] Drag handle is painted at the top-left corner of the editor when selection is collapsed by dragging an handle
**Package Version**
`main` branch with commit hash 5304a3c2294dfa7ae2742539a5e9f2d3e05a3d16.

**To Reproduce**
1. Go to SuperEditor demo
2. Double tap on a word to get an expanded selection
3. Drag the left-handle till the offset of the right handle => selection is collapsed

**Actual behavior**
The Android selection handle is painted on the top-left corner of the editor.

**Expected behavior**
The Android selection handle is painted below the caret.

**Platform**
Android.

**Flutter version**

**Screenshots**

https://github.com/superlistapp/super_editor/assets/31278849/69bf3eab-b2ef-41e5-8b89-ccff68a141c9

angelosilvestre:
[SuperEditor][Android] Fix drag handles when selection collapses due to dragging. Resolves #1736

Dragging an expanded drag handle can cause the selection to collapse. When that happens, the expanded handles are painted in the top left corner:

https://github.com/superlistapp/super_editor/assets/31278849/69bf3eab-b2ef-41e5-8b89-ccff68a141c9

This happens because `AndroidControlsDocumentLayerState` is computing a collapsed handle layout data, while we want to display expanded handles.

This PR changes `AndroidControlsDocumentLayerState` to compute layout data for a expanded handles when the selection is collapsed but we are displaying expanded handles.

After the changes:

[expanded_handles.webm](https://github.com/superlistapp/super_editor/assets/7597082/391d9127-a98b-4382-87b9-adcc266d4f19)

### DIFF
--- a/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
@@ -1425,15 +1425,15 @@ class SuperEditorAndroidControlsOverlayManagerState extends State<SuperEditorAnd
     _moveSelectionAndMagnifierToDragHandleOffset(dragDx: details.delta.dx);
   }
 
-  void _onHandlePanEnd(DragEndDetails details) {
-    _onHandleDragEnd();
+  void _onHandlePanEnd(DragEndDetails details, HandleType handleType) {
+    _onHandleDragEnd(handleType);
   }
 
-  void _onHandlePanCancel() {
-    _onHandleDragEnd();
+  void _onHandlePanCancel(HandleType handleType) {
+    _onHandleDragEnd(handleType);
   }
 
-  void _onHandleDragEnd() {
+  void _onHandleDragEnd(HandleType handleType) {
     _dragHandleSelectionBound = null;
     _dragHandleType = null;
     _dragHandleSelectionGlobalFocalPoint.value = null;
@@ -1443,6 +1443,16 @@ class SuperEditorAndroidControlsOverlayManagerState extends State<SuperEditorAnd
     _controlsController!
       ..blinkCaret()
       ..hideMagnifier();
+
+    if (widget.selection.value?.isCollapsed == true &&
+        const [HandleType.upstream, HandleType.downstream].contains(handleType)) {
+      // The user dragged an expanded handle until the selection collapsed and then released the handle.
+      // While the user was dragging, the expanded handles were displayed.
+      // Show the collapsed.
+      _controlsController!
+        ..hideExpandedHandles()
+        ..showCollapsedHandle();
+    }
 
     // Stop auto-scrolling based on the drag-handle offset.
     widget.dragHandleAutoScroller.value?.stopAutoScrollHandleMonitoring();
@@ -1580,8 +1590,8 @@ class SuperEditorAndroidControlsOverlayManagerState extends State<SuperEditorAnd
                 },
                 onPanStart: (details) => _onHandlePanStart(details, HandleType.collapsed),
                 onPanUpdate: _onHandlePanUpdate,
-                onPanEnd: _onHandlePanEnd,
-                onPanCancel: _onHandlePanCancel,
+                onPanEnd: (details) => _onHandlePanEnd(details, HandleType.collapsed),
+                onPanCancel: () => _onHandlePanCancel(HandleType.collapsed),
                 dragStartBehavior: DragStartBehavior.down,
                 child: AndroidSelectionHandle(
                   key: DocumentKeys.androidCaretHandle,
@@ -1615,8 +1625,8 @@ class SuperEditorAndroidControlsOverlayManagerState extends State<SuperEditorAnd
               },
               onPanStart: (details) => _onHandlePanStart(details, HandleType.upstream),
               onPanUpdate: _onHandlePanUpdate,
-              onPanEnd: _onHandlePanEnd,
-              onPanCancel: _onHandlePanCancel,
+              onPanEnd: (details) => _onHandlePanEnd(details, HandleType.upstream),
+              onPanCancel: () => _onHandlePanCancel(HandleType.upstream),
               dragStartBehavior: DragStartBehavior.down,
               child: AndroidSelectionHandle(
                 key: DocumentKeys.upstreamHandle,
@@ -1644,8 +1654,8 @@ class SuperEditorAndroidControlsOverlayManagerState extends State<SuperEditorAnd
               },
               onPanStart: (details) => _onHandlePanStart(details, HandleType.downstream),
               onPanUpdate: _onHandlePanUpdate,
-              onPanEnd: _onHandlePanEnd,
-              onPanCancel: _onHandlePanCancel,
+              onPanEnd: (details) => _onHandlePanEnd(details, HandleType.downstream),
+              onPanCancel: () => _onHandlePanCancel(HandleType.downstream),
               dragStartBehavior: DragStartBehavior.down,
               child: AndroidSelectionHandle(
                 key: DocumentKeys.downstreamHandle,
--- a/super_editor/lib/src/infrastructure/platforms/android/android_document_controls.dart
+++ b/super_editor/lib/src/infrastructure/platforms/android/android_document_controls.dart
@@ -200,11 +200,14 @@ class AndroidControlsDocumentLayerState
     if (_controlsController != null) {
       _controlsController!.shouldCaretBlink.removeListener(_onBlinkModeChange);
       _controlsController!.caretJumpToOpaqueSignal.removeListener(_caretJumpToOpaque);
+      _controlsController!.shouldShowCollapsedHandle.removeListener(_onShouldShowCollapsedHandleChange);
     }
 
     _controlsController = SuperEditorAndroidControlsScope.rootOf(context);
     _controlsController!.shouldCaretBlink.addListener(_onBlinkModeChange);
     _controlsController!.caretJumpToOpaqueSignal.addListener(_caretJumpToOpaque);
+
+    _controlsController!.shouldShowCollapsedHandle.addListener(_onShouldShowCollapsedHandleChange);
     _onBlinkModeChange();
   }
 
@@ -222,6 +225,7 @@ class AndroidControlsDocumentLayerState
   void dispose() {
     widget.selection.removeListener(_onSelectionChange);
     _controlsController?.shouldCaretBlink.removeListener(_onBlinkModeChange);
+    _controlsController!.shouldShowCollapsedHandle.removeListener(_onShouldShowCollapsedHandleChange);
 
     _caretBlinkController.dispose();
     super.dispose();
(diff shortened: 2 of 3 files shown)


matthew-carroll:
@angelosilvestre is that video showing the problem or the solution? The user should be able to drag a handle across the boundary to move the other direction.

angelosilvestre:
@matthew-carroll The video is showing the problem. I forgot to upload the video of the solution. Updated the PR description to include the video.

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
```diff
@@ -1444,6 +1444,16 @@ class SuperEditorAndroidControlsOverlayManagerState extends State<SuperEditorAnd
       ..blinkCaret()
       ..hideMagnifier();
 
+    if (widget.selection.value?.isCollapsed == true &&
```
matthew-carroll, line 1447: Do we actually need to care about the handle type here? Regardless of what handle the user stopped dragging, we always want to show the collapsed handle for collapsed and expanded handles for expanded, right?

super_editor/lib/src/infrastructure/platforms/android/android_document_controls.dart
```diff
@@ -282,14 +286,24 @@ class AndroidControlsDocumentLayerState
     _caretBlinkController.jumpToOpaque();
   }
 
+  void _onShouldShowCollapsedHandleChange() {
```
matthew-carroll, line 289: The name of this method, and what it's listening to, don't reflect the comment and purpose. For example, whenever the user initially places the caret, this method is probably called, but the commented situation didn't occur.

I'm curious why this layer needs a special update at all for such a situation?

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
angelosilvestre: I added this condition because if the handle type is already `collapsed` we don't need to change anything, because we would already displaying the correct handle.

super_editor/lib/src/infrastructure/platforms/android/android_document_controls.dart
angelosilvestre: With the changes in this PR, even with a collapsed selection we might have a `DocumentSelectionLayout` which only contains the layout information for the expanded handles. Without this, when we `showCollapsedHandle()` we don't have the `Rect` for the caret and we don't show any drag handles.

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
matthew-carroll: Does that mean the same logic would apply equally well for both expanded and collapsed? If so, I'm wondering why we would want to introduce the extra tracking by passing around the handle type if what we're going to do at the end is the same either way.

super_editor/lib/src/infrastructure/platforms/android/android_document_controls.dart
matthew-carroll: > With the changes in this PR, even with a collapsed selection we might have a DocumentSelectionLayout which only contains the layout information for the expanded handles

Why is this the case? Wouldn't this mean that the overlay was already broken in a significant way if it doesn't reflect the latest selection?

angelosilvestre: I'm confused now. Shouldn't we display the expanded handles when the user drags an expanded handle, causing the selection to collapse? If so, we need `DocumentSelectionLayout` to hold layout information for expanded handles.

matthew-carroll: > Shouldn't we display the expanded handles when the user drags an expanded handle, causing the selection to collapse? 

Yes, but at the moment I don't know the relationship between that question and the code that I'm commenting on.

My question above was about building this widget when the selection changes. You're saying we need this method so that when the selection changes, we show the right handles. That implies that this widget isn't already rebuilding when the selection changes, right? If this widget isn't rebuilding when the selection changes, then how are the handles ever right? It seems like this UI would be completely broken if this widget wasn't already building when the selection changes. I assume I'm missing something here, but I'm not sure what.

angelosilvestre: Ok, when the selection changes the UI is already rebuilding. This code is to rebuild the user releases the drag handle, so we change from expanded handles to a collapsed handle.

matthew-carroll: If we want to react to the user ending a drag, then we should add a listener for when the user ends a drag. Baking that implication into a selection change is something that we'll immediately forget and we'll never be able to remove this method because we don't know why we're actually registering it for selection changes.

angelosilvestre: I didn't add `_onShouldShowCollapsedHandleChange` as a selection listener. It's listening for changes of `SuperEditorAndroidControlsController.shouldShowCollapsedHandle`.

matthew-carroll: Was there previously a method registered for selection changes in earlier versions of this PR? Or am I thinking of a different PR?

angelosilvestre: There are already a selection listener in the class that was introduced prior to this PR. That remains unchanged.

super_editor/lib/src/infrastructure/platforms/android/android_document_controls.dart
```diff
@@ -282,14 +286,24 @@ class AndroidControlsDocumentLayerState
...
+
   @override
   DocumentSelectionLayout? computeLayoutDataWithDocumentLayout(BuildContext context, DocumentLayout documentLayout) {
     final selection = widget.selection.value;
     if (selection == null) {
       return null;
     }
 
-    if (selection.isCollapsed) {
+    if (selection.isCollapsed && !_controlsController!.shouldShowExpandedHandles.value) {
```
matthew-carroll, line 306: With regard to the conversation about listening for expanded handles, I'm seeing that the reason you added that listener is because you added the appearance of expanded handles as a condition for computing the layout here.

I think what this issue shows us isn't that we should add another listener, but rather that our assumption that we can conditionally add/remove leaders was wrong. How about we always position all 3 handle types?

angelosilvestre: It seems that are some places that consider the caret to be "visible" if we have the caret rect in the layout data. For example:

```dart
@visibleForTesting
bool get isCaretDisplayed => layoutData?.caret != null;
```

Should we change this to the following?

```dart
@visibleForTesting
bool get isCaretDisplayed => layoutData?.caret != null && widget.selection.value?.isCollapsed == true;
```

We have a `shouldCaretBlink` in the `SuperEditorAndroidControlsController` class, but we don't have a `shouldShowCaret`.

matthew-carroll: I can't see the code around those snippets. What's the actual decision process for the code that paints the caret?

angelosilvestre: Currently, it just looks if the caret layout data is non-null. If it is, it paints the caret.

matthew-carroll: Ok, I guess we can keep it this way then. Can you please add appropriate comments so that it's clear why we're listening to the expanded handle signal. You can probably use some of my questions here to see where things aren't entirely clear.

angelosilvestre: Updated. Can you let me know if the comment is clear enough?

super_editor/lib/src/infrastructure/platforms/android/android_document_controls.dart
```diff
@@ -200,11 +200,18 @@ class AndroidControlsDocumentLayerState
...
       _controlsController!.shouldCaretBlink.removeListener(_onBlinkModeChange);
       _controlsController!.caretJumpToOpaqueSignal.removeListener(_caretJumpToOpaque);
+      _controlsController!.shouldShowCollapsedHandle.removeListener(_onShouldShowCollapsedHandleChange);
     }
 
     _controlsController = SuperEditorAndroidControlsScope.rootOf(context);
     _controlsController!.shouldCaretBlink.addListener(_onBlinkModeChange);
     _controlsController!.caretJumpToOpaqueSignal.addListener(_caretJumpToOpaque);
+
+    // The appearance of the collapsed handle is used to compute the selection layout data.
```
matthew-carroll, line 210: This comment falls into a couple of the recurring issues that we have.

First, as has happened before, the part of the comment that mentions what the code actually does is literally the last line of the comment. Comments should generally be written in a way that the reader might be able to stop reading at any time and move on. But this requires that the most critical information appear first, not last.

Second, while this comment includes a lot of words, and while it might make sense to someone who already knows all the details, it doesn't reveal much of anything to someone who doesn't already know what's going on, and therefore isn't particularly valuable.

Please compare the existing comment to this comment.

```dart
/// Listen for changes about whether we want to show the collapsed handle
/// or whether we want to show expanded handles for a selection. We listen to
/// this because there are some situations where the desired handle type is
/// ambiguous, such as when when the user drags an expanded handle such that
/// the selection collapses. In that case, the selection is collapsed but we want
/// to show the expanded handle. This signal clarifies which one we want.
```

super_editor/lib/src/infrastructure/platforms/android/android_document_controls.dart
```diff
@@ -282,14 +290,24 @@ class AndroidControlsDocumentLayerState
     _caretBlinkController.jumpToOpaque();
   }
 
+  void _onShouldShowCollapsedHandleChange() {
+    // When the user drags an expanded handle it might collapse the selection.
```
matthew-carroll, line 294: This comment is better than the comment above, but it's somewhat incorrect. This comment is written from the perspective that we've gone from expanded handles to collapsed handles "Schedule a new layout to recompute the data for the collapsed handle". We're re-running build because we don't know whether we want a collapsed handle or expanded handles. Please ensure that the comment accounts for both possibilities.

angelosilvestre: Updated. Could you take a look and see if it's better?

matthew-carroll: It still looks one sided. Doesn't this method run whenever "should show collapsed handle" goes from `true` to `false` and `false` to `true`?

angelosilvestre: Maybe:

```
// When the user drags an expanded handle it might collapse the selection.
// In this case, we have a collapsed selection, but layout data for the
// expanded handles. Schedule a rebuild whenever the information about
// which handles we want to show changes to recompute the layout data for the selection.
```
?

matthew-carroll: "In this case, we have a collapsed selection, but layout data for the expanded handles" - this statement sounds like you're saying "this method is running because we have a collapsed selection but we have layout data for the expanded handles". It's unnecessarily ambiguous and can be easily misinterpreted.

```
// The controller went from wanting a collapsed handle to wanting expanded handles,
// or vis-a-versa. This signal is relevant to us because of an ambiguous handle situation.
// The user might drag an expanded handle such  that the selection is collapsed, in which
// case we still want to show an expanded handle. Similarly, if the user then releases that
// expanded handle, we should switch to a collapsed handle for the same selection. This
// method tells us that the desired handle type has changed. Re-run layout and build to
// ensure that we're showing the correct handle.
```

angelosilvestre: OK, pushed.

matthew-carroll (approved):
LGTM

PR #1763, opened 2024-01-08
[SuperEditor][SuperTextField] Remove calls to cancel scroll momentum on tap (Resolves #824)

### LINKED ISSUE #824: [SuperEditor] - Stop scrolling on single finger touch trackpad
When a scrollable is scrolling with momentum, the user should be able to place a single finger on a trackpad to stop the scrolling momentum.

Apparently, an addition to Flutter 3.4 will provide an event to cancel scrolling when that happens, but that signal doesn't exist in Flutter 3.3.3: https://github.com/flutter/flutter/issues/112880#issuecomment-1276739090

When Flutter 3.4 reaches stable, integrate the new signal to stop scrolling momentum.

angelosilvestre:
[SuperEditor][SuperTextField] Remove calls to cancel scroll momentum on tap. Resolves #824

Flutter 3.3.3 had some changes related to scroll momentum that required us to cancel the momentum manually.  We created a `_cancelScrollMomentum` and called it in `onPointerDown`, `onPointerPanZoomStart` and `onPointerHover`.

However, after https://github.com/flutter/flutter/pull/116689 it seems these calls are not necessary anymore because the momentum is already being cancelled.

This PR removes the calls to cancel the scroll momentum `onPointerDown` and `onPointerPanZoomStart`. We also cancel the scroll momentum in `onPointerHover`, but I'm not sure why, so I didn't remove that call for now.

I modified the existing test because it was performing a full tap (down + up) before checking if the momentum was cancelled, but the test description suggests it should cancel after a tap down.

I added a slightly different test for desktop, because it seems that even with the existing code, the selection do change after the tap up.

This PR only affects desktop because we don't use this approach on mobile.

### DIFF
--- a/super_editor/test/super_editor/supereditor_scrolling_test.dart
+++ b/super_editor/test/super_editor/supereditor_scrolling_test.dart
@@ -507,7 +507,7 @@ void main() {
       expect(SuperEditorInspector.findDocumentSelection(), isNull);
     });
 
-    testWidgetsOnDesktop("stops momentum on tap down", (tester) async {
+    testWidgetsOnDesktop("stops momentum on tap down with trackpad and doesn't place the caret", (tester) async {
       final scrollController = ScrollController();
 
       await tester //
@@ -519,8 +519,13 @@ void main() {
       // Ensure the editor initially has no selection.
       expect(SuperEditorInspector.findDocumentSelection(), isNull);
 
-      // Fling scroll the editor.
-      await tester.fling(find.byType(SuperEditor), const Offset(0.0, -1000), 1000);
+      // Fling scroll the editor with the trackpad.
+      final scrollGesture = await tester.startGesture(
+        tester.getCenter(find.byType(SuperEditor)),
+        kind: PointerDeviceKind.trackpad,
+      );
+      await scrollGesture.moveBy(const Offset(0, -1000));
+      await scrollGesture.up();
 
       // Pump a few frames of momentum.
       for (int i = 0; i < 25; i += 1) {
@@ -528,17 +533,27 @@ void main() {
       }
       final scrollOffsetInMiddleOfMomentum = scrollController.offset;
 
+      // Ensure the editor scrolled.
+      expect(scrollOffsetInMiddleOfMomentum, greaterThan(0.0));
+
       // Tap down to stop the momentum.
-      final gesture = await tester.startGesture(tester.getCenter(find.byType(SuperEditor)));
+      final gesture = await tester.startGesture(
+        tester.getCenter(find.byType(SuperEditor)),
+        kind: PointerDeviceKind.trackpad,
+      );
 
       // Let any remaining momentum run (there shouldn't be any).
       await tester.pumpAndSettle();
 
       // Ensure that the momentum stopped exactly where we tapped.
-      expect(scrollOffsetInMiddleOfMomentum, scrollController.offset);
+      expect(scrollController.offset, scrollOffsetInMiddleOfMomentum);
 
       // Release the pointer.
       await gesture.up();
+      await tester.pump();
+
+      // Ensure that tapping on the editor didn't change the selection.
+      expect(SuperEditorInspector.findDocumentSelection(), isNull);
     });
 
     testWidgetsOnAndroid("doesn't overscroll when dragging down", (tester) async {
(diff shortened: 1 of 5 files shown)


super_editor/test/super_editor/supereditor_scrolling_test.dart
```diff
@@ -490,19 +490,57 @@ void main() {
...
 
+      // Release the pointer.
+      await gesture.up();
+      await tester.pump();
+
       // Ensure that tapping on the editor didn't place the caret.
       expect(SuperEditorInspector.findDocumentSelection(), isNull);
     });
 
+    testWidgetsOnDesktop("stops momentum on tap down", (tester) async {
```
matthew-carroll, line 510: How does this test differ from the other one? Please adjust the name to make it clear.

angelosilvestre: The only difference is that in this test we don't check if the selection remains unchanged, because on desktop it seems to change the selection.

What would be a good description?

matthew-carroll: Before we rename the test and move forward, can you check with a few editors and see if it's a standard practice to place the caret on tap down when stopping momentum?

angelosilvestre: In fact, I tested again and noticed it does NOT change the selection. What's happening in the test is that the `fling` method is changing the selection. `fling` doesn't support the trackpad.

Updated the test to avoid using the `fling` method.

super_editor/test/super_textfield/super_textfield_gesture_scrolling_test.dart
```diff
@@ -192,6 +192,53 @@ void main() {
       // Ensure the viewport is big enough so the text doesn't scroll vertically
       expect(viewportHeight, greaterThanOrEqualTo(totalHeight));
     });
+
+    testWidgetsOnAllPlatforms("stops momentum on tap down", (tester) async {
```
matthew-carroll, line 196: Should this name match the other test name?

super_editor/test/super_textfield/super_textfield_gesture_scrolling_test.dart
```diff
@@ -192,6 +192,53 @@ void main() {
...
+        minLines: 1,
+        maxLines: 2,
+        maxHeight: 20,
+      );
+
+      // Ensure the textfield initially has no selection.
+      expect(SuperTextFieldInspector.findSelection(), TextRange.empty);
+
+      // Fling scroll the textfield.
+      await tester.fling(find.byType(SuperTextField), const Offset(0.0, -1000), 1000);
```
matthew-carroll, line 221: Should this fling also be configured as a trackpad?

super_editor/test/super_textfield/super_textfield_gesture_scrolling_test.dart
angelosilvestre: Updated.

super_editor/test/super_textfield/super_textfield_gesture_scrolling_test.dart
angelosilvestre: Updated.

angelosilvestre:
@matthew-carroll I modified the test to run only for desktop, while testing I noticed that we cant drag to scroll the textfield on mobile. Should this be a separate ticket?

https://github.com/superlistapp/super_editor/assets/7597082/2f346eb1-1cc9-44ed-bc37-3ab660a9ae67

matthew-carroll:
@angelosilvestre I can't tell what you're describing or showing. Are you dragging with your finger? Are you dragging on a trackpad that's connected to the mobile device? What's the specific use-case and how might it relate to this ticket?

angelosilvestre:
@matthew-carroll I'm dragging with one finger. Dragging with one finger should scroll the text field, but it isn't doing that. This video was recorded using the latest main

matthew-carroll:
I think that might be because in my tests of single line fields they didn't seem to support drag scrolling. Are you sure that multi-line fields are supposed to support that? If so, please file a new issue with appropriate details.

matthew-carroll (approved):
LGTM

PR #1764, opened 2024-01-08
[SuperEditor] Implement link attribution update when text changes (Resolves #746)

### LINKED ISSUE #746: [super_editor] - URLs should linkify when typed
When users type URLs, they typically expect the URL to become a link automatically.

`SuperEditor` should recognize a URL, when typed. The URL should have a link attribution applied to it.

As the URL has characters added or removed, the attribution URL should be updated accordingly.

angelosilvestre:
[SuperEditor] Implement link attribution update when text changes. Resolves #746

Currently, we allow users to add or removes characters inside a link, but we don't update the link attribution to reflect the changes. For example, the user might type "www.google.com" and then change it to be "www.duckduckgo.com", but the link still points to "www.google.com".

I looked into some other editors. Some of them update the link, some keep the link unchanged and some remove the link entirely.

This PR adds a configuration to `LinkifyReaction` so apps can define how they want to handle changes in URL's. We could also introduce a separate reaction to this.

This PR also re-groups some existing link tests that are similar, to make the test file easier to navigate on.

### DIFF
--- a/super_editor/lib/src/default_editor/default_document_editor_reactions.dart
+++ b/super_editor/lib/src/default_editor/default_document_editor_reactions.dart
@@ -478,15 +478,33 @@ class ImageUrlConversionReaction implements EditReaction {
 
 /// An [EditReaction] which converts a URL into a styled link.
 ///
-/// This reaction only applies when the user enters a space " " after a token that
+/// When the URL has characters added or removed, the [updatePolicy] determines
+/// which action to take:
+///
+/// - [LinkUpdatePolicy.preserve] : the attribution remains unchanged.
+/// - [LinkUpdatePolicy.update] : the attribution is updated to reflect the new URL.
+///   The attribution is updated only if characters are inserted/removed at the middle of a link.
+/// - [LinkUpdatePolicy.remove] : the attribution is removed for the exinting link.
+///   The attribution is removed if characters are inserted at the middle of a link,
+///   or if characters are removed anywhere in a link.
+///
+/// The linkification only applies when the user enters a space " " after a token that
 /// looks like a URL. If the user doesn't enter a trailing space, or the preceding
-/// token doesn't look like a URL, then no reaction occurs.
+/// token doesn't look like a URL, then the linkification doesn't happen.
 class LinkifyReaction implements EditReaction {
-  const LinkifyReaction();
+  const LinkifyReaction({
+    this.updatePolicy = LinkUpdatePolicy.preserve,
+  });
+
+  /// Configures how a
+  final LinkUpdatePolicy updatePolicy;
 
   @override
   void react(EditContext editContext, RequestDispatcher requestDispatcher, List<EditEvent> edits) {
     final document = editContext.find<MutableDocument>(Editor.documentKey);
+    final composer = editContext.find<MutableDocumentComposer>(Editor.composerKey);
+    bool didInsertSpace = false;
+
     TextInsertionEvent? linkifyCandidate;
     for (int i = 0; i < edits.length; i++) {
       final edit = edits[i];
@@ -495,6 +513,7 @@ class LinkifyReaction implements EditReaction {
         if (change is TextInsertionEvent && change.text.text == " ") {
           // Every space insertion might appear after a URL.
           linkifyCandidate = change;
+          didInsertSpace = true;
         }
       } else if (edit is SelectionChangeEvent) {
         if (linkifyCandidate == null) {
@@ -553,6 +572,10 @@ class LinkifyReaction implements EditReaction {
         }
       }
     }
+
+    if (!didInsertSpace) {
+      _tryUpdateLinkAttribution(document, composer, edits);
+    }
   }
 
   /// Extracts a word ending at [endOffset] tries to linkify it.
@@ -582,9 +605,7 @@ class LinkifyReaction implements EditReaction {
     final int linkCount = extractedLinks.fold(0, (value, element) => element is UrlElement ? value + 1 : value);
     if (linkCount == 1) {
       // The word is a single URL. Linkify it.
-      final uri = word.startsWith("http://") || word.startsWith("https://") //
-          ? Uri.parse(word)
-          : Uri.parse("https://$word");
+      final uri = _parseLink(word);
 
       text.addAttribution(
         LinkAttribution(url: uri),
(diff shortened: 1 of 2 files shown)


super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -622,6 +644,145 @@ class LinkifyReaction implements EditReaction {
...
+      // followed a selection change event. There aren't any events in the the change list,
+      // therefore this reaction shouldn't apply. Fizzle.
+      return;
+    }
+
+    late NodeChangeEvent insertionOrDeletionEvent;
+    if (changeList.length == 1) {
+      final editEvent = changeList.last;
+      if (editEvent is! DocumentEdit || editEvent.change is! TextDeletedEvent) {
+        // This reaction can run with a single deletion event because
```
matthew-carroll, line 666: Let's replace the comment with this:

```
// There's only a single event in the change list, and it's not a deletion
// event. The only situation where a URL would change with a single
// event is a deletion event. Therefore, we don't need to change a URL.
// Fizzle.
```

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -622,6 +644,145 @@ class LinkifyReaction implements EditReaction {
...
+        // deleting the downstream character generates a single change.
+        // We have a single event, but it isn't a deletion. Fizzle.
+        return;
+      }
+
+      insertionOrDeletionEvent = editEvent.change as NodeChangeEvent;
+    } else {
+      final selectionEvent = changeList.last;
+      if (selectionEvent is! SelectionChangeEvent) {
+        // The change list contains more than one event.
```
matthew-carroll, line 676: Let's replace this comment with:

```
// The last event isn't a selection event. We expect a URL change
// to consist of an insertion or a deletion followed by a selection
// change. This event list doesn't fit the pattern. Fizzle.
```

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -622,6 +644,145 @@ class LinkifyReaction implements EditReaction {
...
+        // In this case, it's required that the last two events are an insertion/deletion event
+        // followed by a selection change event. The last event isn't a selection event,
+        // therefore this reaction shouldn't apply. Fizzle.
+        return;
+      }
+
+      final edit = changeList[changeList.length - 2];
+      if (edit is! DocumentEdit || //
+          (edit.change is! TextInsertionEvent && edit.change is! TextDeletedEvent)) {
+        // The change list contains more than one event.
```
matthew-carroll, line 686: Let's change this comment to:

```
// The second to last event isn't an insertion or deletion. We 
// expect a URL change to consist of an insertion or a deletion 
// followed by a selection change. This event list doesn't fit 
// the pattern. Fizzle.
```

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -622,6 +644,145 @@ class LinkifyReaction implements EditReaction {
...
+        // In this case, this reaction requires that the two last events are an insertion/deletion event
+        // followed by a selection change event.  The second to last event isn't a text insertion/deletion event,
+        // therefore this reaction shouldn't apply. Fizzle.
+        return;
+      }
+
+      insertionOrDeletionEvent = edit.change as NodeChangeEvent;
+    }
+
+    final changedNodeId = insertionOrDeletionEvent.nodeId;
```
matthew-carroll, line 696: Before this line would be a good place for a comment that brings us up to date with what's going on. I'm not entirely sure what the code below this point is doing, but would it make sense to say

```
// The change list includes an insertion or deletion followed by a selection
// change, therefore a URL may have changed. Look for a URL around the
// altered text.
```

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -622,6 +644,145 @@ class LinkifyReaction implements EditReaction {
...
+          )
+          .firstOrNull;
+    }
+
+    if (insertionOrDeletionOffset < changedNodeText.length) {
+      // Check if the downstream character has a link attribution.
+      downstreamLinkAttribution = changedNodeText
+          .getAttributionSpansInRange(
+            attributionFilter: (attribution) => attribution is LinkAttribution,
+            range: SpanRange(insertionOrDeletionOffset + 1, insertionOrDeletionOffset + 1),
```
matthew-carroll, line 720: A couple questions here. First, does `insertionOrDeletionOffset < changeNodeText.length` mean that you can access `insertionOrDeletionOffset + 1`?

Second, if a character was deleted, then shouldn't you be checking the character AT the effected index, not the character after it?

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -622,6 +644,145 @@ class LinkifyReaction implements EditReaction {
...
+          )
+          .firstOrNull;
+    }
+
+    if (upstreamLinkAttribution == null && downstreamLinkAttribution == null) {
+      // There isn't a link around the changed offset. Fizzle.
+      return;
+    }
+
+    // Whether or not the change happened at the middle of a link, i.e.,
```
matthew-carroll, line 730: If a condition needs explanation then please start with the motivation for the condition. I'm not sure, but I'm guessing that the following comment would apply here:

```
// We only want to update a URL if a change happened within an existing URL,
// not at the edges. Determine whether this change reflects an insertion or
// deletion within an existing URL by looking for identical link attributions
// both upstream and downstream from the edited text offset.
```

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -622,6 +644,145 @@ class LinkifyReaction implements EditReaction {
...
+        upstreamLinkAttribution.attribution == downstreamLinkAttribution.attribution;
+
+    if (!isAtMiddleOfLink && insertionOrDeletionEvent is TextInsertionEvent) {
+      // An insertion happened at an edge of the link.
+      // Insertion only updates the attribution when happening at the middle of a link. Fizzle.
+      return;
+    }
+
+    final rangeToUpdate = isAtMiddleOfLink //
+        ? SpanRange(upstreamLinkAttribution.start, downstreamLinkAttribution.end)
```
matthew-carroll, line 743: Looking at this, I'm curious if you need to do all this span end cap inspection and management. Even looking up above where you search for the link attribution, I'm wondering if we don't already have tools to accomplish such a thing.

Can you check `AttributedText` and see if there are tools that can remove some of this work?

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -622,6 +644,145 @@ class LinkifyReaction implements EditReaction {
...
+    final attributionsToRemove = changedNodeText.getAttributionSpansInRange(
+      attributionFilter: (attr) => attr is LinkAttribution,
+      range: rangeToUpdate,
+    );
+    for (final attributionSpan in attributionsToRemove) {
+      changedNodeText.removeAttribution(attributionSpan.attribution, attributionSpan.range);
+      composer.preferences.removeStyle(attributionSpan.attribution);
+    }
+
+    if (updatePolicy == LinkUpdatePolicy.update) {
```
matthew-carroll, line 756: This is definitely a point in the code where you should be very explicit in a comment about the branching behavior. There's a critical decision here that's easy to overlook.

Please add the following comment above the if-statement:

```
// A URL was changed and we have now removed the original link. Removing
// the original link was a necessary step for both `LinkUpdatePolicy.remove`
// and for `LinkUpdatePolicy.update`. 
//
// If the policy is `LinkUpdatePolicy.update` then we need to add a new
// link attribution that reflects the edited URL text. We do that below.
```

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -622,6 +644,145 @@ class LinkifyReaction implements EditReaction {
...
+      // We are configure to update the link. Add the new attribution.
```
matthew-carroll, line 757: With the comment I mentioned above, you can remove this comment here.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -622,6 +644,145 @@ class LinkifyReaction implements EditReaction {
...
+  Uri _parseLink(String text) {
+    final uri = text.startsWith("http://") || text.startsWith("https://") //
+        ? Uri.parse(text)
+        : Uri.parse("https://$text");
+    return uri;
+  }
+}
+
+/// Configuration for the action that should happen when a text containing
+/// a link attribution is modified.
```
matthew-carroll, line 776: Please include an example "e.g., "google.com" becomes "gogle.com"

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -622,6 +644,145 @@ class LinkifyReaction implements EditReaction {
...
+enum LinkUpdatePolicy {
+  /// The attribution must remain unchanged.
```
matthew-carroll, line 778: Saying "must" is a passive statement as if describing a contract that someone else needs to uphold. We want to describe what WILL happen when the user choose a given policy.

```
/// When a linkified URL has characters added or deleted, the link remains the same.
```

```
/// When a linkified URL has characters added or removed, the link is updated to reflect the new URL value.
```

```
/// When a linkified URL has characters added or removed, the link is completely removed.
```

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
```diff
@@ -478,15 +478,33 @@ class ImageUrlConversionReaction implements EditReaction {
...
+/// which action to take:
+///
+/// - [LinkUpdatePolicy.preserve] : the attribution remains unchanged.
+/// - [LinkUpdatePolicy.update] : the attribution is updated to reflect the new URL.
+///   The attribution is updated only if characters are inserted/removed at the middle of a link.
+/// - [LinkUpdatePolicy.remove] : the attribution is removed for the exinting link.
+///   The attribution is removed if characters are inserted at the middle of a link,
+///   or if characters are removed anywhere in a link.
+///
+/// The linkification only applies when the user enters a space " " after a token that
```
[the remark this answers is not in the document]
angelosilvestre: Updated.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
angelosilvestre: > Can you check AttributedText and see if there are tools that can remove some of this work?

I didn't find any methods that for that...

>  Even looking up above where you search for the link attribution, I'm wondering if we don't already have tools to accomplish such a thing.

The attribution query methods perform equality comparison, but here we want any `LinkAttribution`. We don't have a method that filters by type.

super_editor/lib/src/default_editor/default_document_editor_reactions.dart
angelosilvestre: > First, does insertionOrDeletionOffset < changeNodeText.length mean that you can access insertionOrDeletionOffset + 1?

It should be  `insertionOrDeletionOffset < changedNodeText.length - 1`.

> Second, if a character was deleted, then shouldn't you be checking the character AT the effected index, not the character after it?

That's right. Updated to only add 1 if it is an insertion.

matthew-carroll: > It should be insertionOrDeletionOffset < changedNodeText.length - 1.

This makes me worried that we're missing at least one test condition. Do you know there weren't any failing tests for the +1 instead of -1?

I'm also concerned that weren't any failing tests for using the insertion index for deletions.

Can you take another look at the tests and see if we should check some addition cases?

angelosilvestre: > Second, if a character was deleted, then shouldn't you be checking the character AT the effected index, not the character after it?

I was able to modify the deletion test to fail when using the wrong offset.

> It should be insertionOrDeletionOffset < changedNodeText.length - 1.

After debugging a bit more I modified to:

```dart
if ((insertionOrDeletionEvent is TextInsertionEvent && insertionOrDeletionOffset < changedNodeText.length - 1) ||
    (insertionOrDeletionEvent is TextDeletedEvent && insertionOrDeletionOffset < changedNodeText.length)) {
 //
}
```

For deletions, we are using only `insertionOrDeletionOffset` instead of `insertionOrDeletionOffset + 1`, so we should check for `insertionOrDeletionOffset < changedNodeText.length`.

For insertions, I couldn't write a failing test, if we query the attributions with a offset bigger then the last offset, it just returns an empty list of attributions, so the rest of the code still works.

matthew-carroll (approved):
LGTM

PR #1765, opened 2024-01-08
[attributed_text] Fix invalid end index with an marker that exceeds text length (Resolves #632)

### LINKED ISSUE #632: AttributedText: `collapsedSpans` returns overly large `MultiAttributionSpan` for unattributed string

angelosilvestre:
[attributed_text] Fix invalid end index with an marker that exceeds text length. Resolves #632

When an `AttributedText` has a `SpanMarker` with an end index equals to the text length (which is an invalid index, because the spans are inclusive), `collapseSpans` returns a span with an end index equal to the text length, instead of the text length -1, which is the expected index that would point to the last character.

This PR fixes the comparison that causes this behavior.

### DIFF
--- a/attributed_text/test/attributed_text_test.dart
+++ b/attributed_text/test/attributed_text_test.dart
@@ -781,15 +781,15 @@ void main() {
 
         final spans = text.spans.collapseSpans(contentLength: text.text.length);
 
-        // Ensure two spans were returned.
-        // The first containing no attributions and the second containing the attribution.
+        // Ensure two spans were returned. The first containing no attributions and
+        // the second containing the attribution.
         expect(spans.length, 2);
         expect(spans[0].attributions, isEmpty);
         expect(spans[0].start, 0);
         expect(spans[0].end, 5);
         expect(spans[1].attributions, isNotEmpty);
         expect(spans[1].start, 6);
-        expect(spans[1].end, text.text.length - 1);
+        expect(spans[1].end, 10);
       });
     });
   });
--- a/attributed_text/test/attributed_text_test.dart
+++ b/attributed_text/test/attributed_text_test.dart
@@ -681,6 +681,117 @@ void main() {
         expect(expectedVisits, isEmpty);
       });
     });
+
+    group('collapseSpans', () {
+      const boldAttribution = NamedAttribution('bold');
+
+      test('returns a single span for text without attributions', () {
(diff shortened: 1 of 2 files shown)


attributed_text/test/attributed_text_test.dart
```diff
@@ -681,6 +681,117 @@ void main() {
...
+
+        // Ensure two spans were returned.
+        // The first containing no attributions and the second containing the attribution.
+        expect(spans.length, 2);
+        expect(spans[0].attributions, isEmpty);
+        expect(spans[0].start, 0);
+        expect(spans[0].end, 5);
+        expect(spans[1].attributions, isNotEmpty);
+        expect(spans[1].start, 6);
+        expect(spans[1].end, text.text.length - 1);
```
matthew-carroll, line 792: Given that the original offset of `11` is passed in literally, it's very easy to miss the significance of this line. Let's pass a literal `10` as the expectation instead of calculation `length - 1`.

attributed_text/test/attributed_text_test.dart
```diff
@@ -681,6 +681,117 @@ void main() {
...
+                markerType: SpanMarkerType.end,
+                offset: 11,
+              ),
+            ],
+          ),
+        );
+
+        final spans = text.spans.collapseSpans(contentLength: text.text.length);
+
+        // Ensure two spans were returned.
```
matthew-carroll, line 784: When writing comments, please don't jump down to a second line after the first sentence. Comments should be formatted like a paragraph - the only thing that breaks lines is going too far to the right.

If you reach a situation where you feel that the first line should be separated from what follows because they're very different statements, then you should place a blank line after the first sentence. Otherwise it just looks like a strange formatting decision without much meaning.

matthew-carroll (approved):
LGTM

PR #1766, opened 2024-01-09
[SuperEditor] Fix selection shifting when moving between nodes (Resolves #1761)

### LINKED ISSUE #1761: [SuperEditor] Moving selection up/down between tasks shifts the offset
**Package Version**
`main` branch with commit hash https://github.com/superlistapp/super_editor/commit/fbddcaf3f7e6543514c8b05d0c710aaa7ca349a4.

**To Reproduce**
1. Go to "Animated task height" demo
2. Add more text to each task, like "Task 1 with some extra text", "Task 2 with some extra text", etc.
3. Place the selection at the end of the first task
4. Press down key 

**Actual behavior**
The caret moves vertically (to the node below), but moves 6 characters to the left.

**Expected behavior**
Since the length o the node below is the same, then the caret should just move vertically.

**Platform**
Desktop.

**Flutter version**

**Screenshots**

https://github.com/superlistapp/super_editor/assets/31278849/0b810924-ba05-4a8f-91c1-9fca023baad6

angelosilvestre:
[SuperEditor] Fix selection shifting when moving between nodes. Resolves #1761

In the "Animated task height" demo, moving the caret between nodes using UP or DOWN arrow is shifting the selection horizontally:

https://github.com/superlistapp/super_editor/assets/31278849/0b810924-ba05-4a8f-91c1-9fca023baad6

This would happen to any component that mixes in `ProxyDocumentComponent`. 

To move the selection vertically between nodes we first get the current offset for the selected position and then compute 
 which position in the upstream/downstream node is closer to this offset.

The issue is that for `ProxyDocumentComponent`, `getOffsetForPosition` only considers the text inside the component, not the whole component. In the example of the task component, it ignores the checkbox that appears before the text.

For example, consider this layout:

```
[ ]  This is the task 1
[ ]  This is the task 2
```

If the caret sits in `This is the task 1|`, pressing DOWN should move the caret to `This is the task 2|`.

However, we are computing the original offset as if the layout was like the following:

```
This is the task 1
[ ]  This is the task 2
```

So the offset at `This is the task 1|` would move down to `This is the t|ask 2`.

This PR changes `ProxyDocumentComponent` to consider the offset inside the whole component layout.

### DIFF
--- a/super_editor/lib/src/core/document_layout.dart
+++ b/super_editor/lib/src/core/document_layout.dart
@@ -347,7 +347,13 @@ mixin ProxyDocumentComponent<T extends StatefulWidget> implements DocumentCompon
 
   @override
   Offset getOffsetForPosition(NodePosition nodePosition) {
-    return _getOffsetFromChild(_childDocumentComponent.getOffsetForPosition(nodePosition));
+    // Get the offset from the child to return the offset in
+    // component space instead of text space. Without this,
+    // the offset for components that shift the text, like a task,
+    // won't be aligned with a text component.
+    return _getOffsetFromChild(
+      _childDocumentComponent.getOffsetForPosition(nodePosition),
+    );
   }
 
   @override
--- a/super_editor/lib/src/core/document_layout.dart
+++ b/super_editor/lib/src/core/document_layout.dart
@@ -347,10 +347,11 @@ mixin ProxyDocumentComponent<T extends StatefulWidget> implements DocumentCompon
 
   @override
   Offset getOffsetForPosition(NodePosition nodePosition) {
-    // Get the offset from the child to return the offset in
-    // component space instead of text space. Without this,
-    // the offset for components that shift the text, like a task,
-    // won't be aligned with a text component.
+    // In addition to the standard `getOffsetForPosition` of the child component, the proxy
+    // also calls `_getOffsetFromChild`, which returns the offset from the top-left of this
+    // proxy box, to the top-left of the child. Some proxy components, such as a task,
+    // add content that shifts the child component, like adding a checkbox. Any such
+    // shift of the child component must be accounted for when reporting a content offset.
     return _getOffsetFromChild(
       _childDocumentComponent.getOffsetForPosition(nodePosition),
     );
--- a/super_editor/lib/src/core/document_layout.dart
+++ b/super_editor/lib/src/core/document_layout.dart
@@ -451,35 +451,6 @@ mixin ProxyDocumentComponent<T extends StatefulWidget> implements DocumentCompon
   }
 }
 
-class ProxyTextComponent extends StatefulWidget {
-  const ProxyTextComponent({
-    super.key,
-    required this.childDocumentComponentKey,
-    required this.child,
-  });
-
-  final GlobalKey<State<StatefulWidget>> childDocumentComponentKey;
-
-  final Widget child;
-
-  @override
-  State<ProxyTextComponent> createState() => _ProxyTextComponentState();
-}
-
-class _ProxyTextComponentState extends State<ProxyTextComponent>
-    with ProxyDocumentComponent<ProxyTextComponent>, ProxyTextComposable {
-  @override
-  GlobalKey<State<StatefulWidget>> get childDocumentComponentKey => widget.childDocumentComponentKey;
-
-  @override
-  TextComposable get childTextComposable => childDocumentComponentKey.currentState as TextComposable;
-
-  @override
-  Widget build(BuildContext context) {
-    return widget.child;
-  }
-}
-
 /// Preferences for how the document selection should change, e.g.,
 /// move word-by-word instead of character-by-character.
 ///
--- a/super_editor/lib/src/core/document_layout.dart
+++ b/super_editor/lib/src/core/document_layout.dart
@@ -347,7 +347,7 @@ mixin ProxyDocumentComponent<T extends StatefulWidget> implements DocumentCompon
 
   @override
   Offset getOffsetForPosition(NodePosition nodePosition) {
-    return _childDocumentComponent.getOffsetForPosition(nodePosition);
+    return _getOffsetFromChild(_childDocumentComponent.getOffsetForPosition(nodePosition));
   }
 
   @override
--- a/super_editor/lib/src/core/document_layout.dart
+++ b/super_editor/lib/src/core/document_layout.dart
@@ -1,5 +1,6 @@
 import 'package:flutter/widgets.dart';
 import 'package:super_editor/src/core/editor.dart';
+import 'package:super_editor/src/infrastructure/composable_text.dart';
 
 import 'document_selection.dart';
 import 'document.dart';
@@ -447,6 +448,35 @@ mixin ProxyDocumentComponent<T extends StatefulWidget> implements DocumentCompon
   @override
   MouseCursor? getDesiredCursorAtOffset(Offset localOffset) {
     return _childDocumentComponent.getDesiredCursorAtOffset(_getChildOffset(localOffset));
+  }
+}
+
+class ProxyTextComponent extends StatefulWidget {
+  const ProxyTextComponent({
+    super.key,
+    required this.childDocumentComponentKey,
+    required this.child,
+  });
+
+  final GlobalKey<State<StatefulWidget>> childDocumentComponentKey;
+
+  final Widget child;
+
+  @override
+  State<ProxyTextComponent> createState() => _ProxyTextComponentState();
+}
+
+class _ProxyTextComponentState extends State<ProxyTextComponent>
+    with ProxyDocumentComponent<ProxyTextComponent>, ProxyTextComposable {
+  @override
+  GlobalKey<State<StatefulWidget>> get childDocumentComponentKey => widget.childDocumentComponentKey;
+
+  @override
+  TextComposable get childTextComposable => childDocumentComponentKey.currentState as TextComposable;
+
+  @override
+  Widget build(BuildContext context) {
+    return widget.child;
   }
 }
 
--- a/super_editor/lib/src/default_editor/list_items.dart
+++ b/super_editor/lib/src/default_editor/list_items.dart
@@ -316,7 +316,7 @@ class UnorderedListItemComponent extends StatelessWidget {
     final lineHeight = textScaler.scale(textStyle.fontSize! * (textStyle.height ?? 1.25));
     const manualVerticalAdjustment = 3.0;
 
-    return ProxyTextComponent(
+    return ProxyTextDocumentComponent(
       key: componentKey,
       childDocumentComponentKey: _textKey,
       child: Row(
@@ -418,7 +418,7 @@ class OrderedListItemComponent extends StatelessWidget {
     final textScaler = MediaQuery.textScalerOf(context);
     final lineHeight = textScaler.scale(textStyle.fontSize! * (textStyle.height ?? 1.0));
 
-    return ProxyTextComponent(
+    return ProxyTextDocumentComponent(
       key: componentKey,
       childDocumentComponentKey: _textKey,
       child: Row(
--- a/super_editor/lib/src/default_editor/list_items.dart
+++ b/super_editor/lib/src/default_editor/list_items.dart
@@ -1,7 +1,6 @@
 import 'package:attributed_text/attributed_text.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
-import 'package:super_editor/src/core/document_layout.dart';
 import 'package:super_editor/src/core/edit_context.dart';
 import 'package:super_editor/src/core/editor.dart';
 import 'package:super_editor/src/default_editor/attributions.dart';
@@ -306,7 +305,17 @@ class UnorderedListItemComponent extends StatelessWidget {
   final bool showComposingUnderline;
   final bool showDebugPaint;
 
-  final GlobalKey _textKey = GlobalKey();
+  /// A [GlobalKey] that connects a [ProxyTextDocumentComponent] to its
+  /// descendant [TextComponent].
+  ///
+  /// The [ProxyTextDocumentComponent] doesn't know where the [TextComponent] sits
+  /// in its subtree, but the proxy needs access to the [TextComponent] to provide
+  /// access to text layout details.
+  ///
+  /// This key doesn't need to be public because the given [componentKey]
+  /// provides clients with direct access to text layout queries, as well as
+  /// standard [DocumentComponent] queries.
+  final GlobalKey _innerTextComponentKey = GlobalKey();
 
   @override
   Widget build(BuildContext context) {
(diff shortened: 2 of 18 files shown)


angelosilvestre:
@matthew-carroll I noticed a similar issue when moving the selection from a list item to a paragraph, and from a paragraph to a list item.

However, it should be a different issue, because `UnorderedListItemComponent` doesn't uses `ProxyDocumentComponent`. Should I take a look at this?

super_editor/lib/src/core/document_layout.dart
```diff
@@ -347,7 +347,7 @@ mixin ProxyDocumentComponent<T extends StatefulWidget> implements DocumentCompon
 
   @override
   Offset getOffsetForPosition(NodePosition nodePosition) {
-    return _childDocumentComponent.getOffsetForPosition(nodePosition);
+    return _getOffsetFromChild(_childDocumentComponent.getOffsetForPosition(nodePosition));
```
matthew-carroll, line 350: Did you check all the uses of `getOffsetForPosition` to ensure that this change is inline with any related expectations?

Please add a comment that makes it clear why calling `_getOffsetFromChild()` here. Also, please put a comma and newline after `getOffsetForPosition()` so that we can clearly see that two calls are taking place.

super_editor/test/super_editor/supereditor_input_keyboard_actions_test.dart
```diff
@@ -823,6 +824,392 @@ void main() {
         },
         variant: inputSourceVariant,
       );
+
+      group("jumps to downstream node with DOWN ARROW", () {
```
matthew-carroll, line 828: Shouldn't some combination of the group and test names mention that we're preserving the approximate x offset? That's what the issue ticket is about, right?

super_editor/test/super_editor/supereditor_input_keyboard_actions_test.dart
```diff
@@ -823,6 +824,392 @@ void main() {
...
+          final document = MutableDocument(
+            nodes: [
+              TaskNode(id: '1', text: AttributedText('This is a task'), isComplete: false),
+              TaskNode(id: '2', text: AttributedText('This is another task'), isComplete: false),
+            ],
+          );
+          final composer = MutableDocumentComposer();
+          final editor = createDefaultDocumentEditor(document: document, composer: composer);
+
+          await tester.pumpWidget(
```
matthew-carroll, line 1169: I'm seeing a bunch of repetition in these tests, and the tests are pretty long. Can we simplify the unimportant details of these tests at all?

matthew-carroll:
> I noticed a similar issue when moving the selection from a list item to a paragraph, and from a paragraph to a list item.

Let's solve both issues in this PR. The UX problem remains the same, regardless of whether the implementation details are different.

super_editor/test/super_editor/supereditor_input_keyboard_actions_test.dart
angelosilvestre: Updated

super_editor/test/super_editor/supereditor_input_keyboard_actions_test.dart
angelosilvestre: Updated

super_editor/lib/src/core/document_layout.dart
angelosilvestre: > Did you check all the uses of getOffsetForPosition to ensure that this change is inline with any related expectations?

Currently, `getOffsetForPosition` seems to be used only for moving the caret.

> Please add a comment that makes it clear why calling `_getOffsetFromChild()` here

I added a comment, but I could use some help to make it more clear.

super_editor/lib/src/default_editor/list_items.dart
```diff
@@ -271,10 +273,9 @@ class ListItemComponentViewModel extends SingleColumnLayoutComponentViewModel wi
 /// Displays a un-ordered list item in a document.
 ///
 /// Supports various indentation levels, e.g., 1, 2, 3, ...
-class UnorderedListItemComponent extends StatelessWidget {
+class UnorderedListItemComponent extends StatefulWidget {
   const UnorderedListItemComponent({
     Key? key,
-    required this.textKey,
```
matthew-carroll, line 277: Can you explain why it's OK to remove this? Isn't the `textKey` passed from the component builder so that this widget shares the same key with other areas of the widget tree?

super_editor/lib/src/test/super_editor_test/supereditor_inspector.dart
```diff
@@ -146,7 +146,17 @@ class SuperEditorInspector {
   /// {@macro supereditor_finder}
   static AttributedText findTextInParagraph(String nodeId, [Finder? superEditorFinder]) {
     final documentLayout = findDocumentLayout(superEditorFinder);
-    return (documentLayout.getComponentByNodeId(nodeId) as TextComponentState).widget.text;
+    final component = documentLayout.getComponentByNodeId(nodeId);
```
matthew-carroll, line 149: This method no longer does what the name implies. We need to do something here. Are there current usages of this method (before this PR) that use it to find text in a non-paragraph `TextNode`?

super_editor/lib/src/test/super_editor_test/supereditor_inspector.dart
```diff
@@ -158,7 +168,7 @@ class SuperEditorInspector {
   static TextSpan findRichTextInParagraph(String nodeId, [Finder? superEditorFinder]) {
     final documentLayout = findDocumentLayout(superEditorFinder);
 
-    final textComponentState = documentLayout.getComponentByNodeId(nodeId) as TextComponentState;
+    final textComponentState = documentLayout.getComponentByNodeId(nodeId) as DocumentComponent;
```
matthew-carroll, line 171: Why did the type cast here change? The name of the variable suggests that its a `TextComponentState`.

super_editor/test/super_editor/supereditor_input_keyboard_actions_test.dart
```diff
@@ -823,6 +824,680 @@ void main() {
         },
         variant: inputSourceVariant,
       );
+
+      group("jumps to downstream node preserving approximate x position with DOWN ARROW", () {
```
matthew-carroll, line 828: "x position" -> "x-position"

super_editor/lib/src/default_editor/list_items.dart
angelosilvestre: I modified the component build to pass the `componentKey` directly as the key for the `UnorderedListItemComponent`.

To be able to get the `RenderBox` for the whole component, not just the text inside of it, we need to use the `componentKey` as the key for the "main" component.

super_editor/lib/src/test/super_editor_test/supereditor_inspector.dart
angelosilvestre: We already use it to find the text of list items. For example, the test "pastes within a list item" uses it.

super_editor/lib/src/test/super_editor_test/supereditor_inspector.dart
angelosilvestre: Because now the component might be a `ProxyDocumentComponent`.

super_editor/lib/src/test/super_editor_test/supereditor_inspector.dart
matthew-carroll: If list items already use the existing method (without your changes) then it suggests that there's more to this story. List items are proxies, right? If so, how was this method working for them before your changes?

angelosilvestre: Before this PR, the component key was applied to the `TextComponent` inside the list item component, so it wasn't working like the proxy component, it was as if the list item was just the `TextComponent` inside of it.

matthew-carroll: And which aspect of this PR requires us to change that? Where do you now need a key that's bound to the proxy component?

angelosilvestre: Because we need to get the character `Offset` in the whole component space, not just the `Offset` in the `TextComponent` space. We need the `Offset` in the whole component coordinate space to move to the correctly x-offset.

matthew-carroll: Sure, I appreciate that we need to be able to get the delta between the component bounds and the content bounds. But isn't it the component, itself, that calculates this? In other words, doesn't the proxy component widget run that calculation? If so, the proxy component widget already has direct access to its render object, i.e., its own origin and bounds. Am I missing something? Is some other object trying to lookup the proxy component origin?

super_editor/lib/src/default_editor/list_items.dart
```diff
@@ -305,10 +305,24 @@ class UnorderedListItemComponent extends StatelessWidget {
   final bool showComposingUnderline;
   final bool showDebugPaint;
 
+  @override
+  State<UnorderedListItemComponent> createState() => _UnorderedListItemComponentState();
+}
+
+class _UnorderedListItemComponentState extends State<UnorderedListItemComponent>
```
matthew-carroll, line 312: It seems unfortunate that we have to convert this to a stateful widget just to get the proxy component behavior. Is it possible to create a `StatelessProxyDocumentComponent` widget that would allow the list item component to do something like:

```dart
return Row(
  children: [
    // list item indicatory
    Expanded (
      child: StatelessProxyDocumentComponent(
        child: TextComponent(
          key: _textKey,
          // ...
        ),
      ),
    ],
  ),
);
```

super_editor/lib/src/test/super_editor_test/supereditor_inspector.dart
matthew-carroll: Please rename this one like the other one.

super_editor/lib/src/default_editor/list_items.dart
angelosilvestre: To implement a component we need to make it a stateful widget. The `DocumentComponent` is a mixin on `State`, not on the widget.

We could make a `ProxyComponentWidget extends StatefulWidget` and then use it like you described.

angelosilvestre: I implemented a `ProxyDocumentComponent`. Please take a look to see if you agree with the list component looks like now.

If you are happy with the implementation, I'll add the docs for the new component.

super_editor/lib/src/core/document_layout.dart
```diff
@@ -443,6 +451,35 @@ mixin ProxyDocumentComponent<T extends StatefulWidget> implements DocumentCompon
   }
 }
 
+class ProxyTextComponent extends StatefulWidget {
```
matthew-carroll, line 454: Do we need this to be specific to text? It looks like the only abstract behavior in `ProxyDocumentComponent` is `GlobalKey get childDocumentComponentKey` - can't we define a stateful widget that implements that one property and avoid making it specific to any particular content type?

If so, perhaps we should rename `ProxyDocumentComponent` to `ProxyDocumentComponentStateMixin` and then create a new widget called `ProxyDocumentComponent` that takes the desired key, and the subtree, like the `Row` for the list items.

angelosilvestre: We can't. This component need to mix in `ProxyTextComposable`, otherwise things like double tap to select by word doesn't work. This mixin need to be mixed by `ProxyTextComponent` because the `ProxyTextComponent` is bound to the component key.

super_editor/lib/src/default_editor/list_items.dart
```diff
@@ -305,6 +306,8 @@ class UnorderedListItemComponent extends StatelessWidget {
   final bool showComposingUnderline;
   final bool showDebugPaint;
 
+  final GlobalKey _textKey = GlobalKey();
```
matthew-carroll, line 309: Let's call this `_innerTextComponentKey`.

Let's add this Dart Doc:

```
/// A [GlobalKey] that connects a [ProxyTextDocumentComponent] to its
/// descendant [TextComponent].
///
/// The [ProxyTextDocumentComponent] doesn't know where the
/// [TextComponent] sits in its subtree, but the proxy needs access to the
/// [TextComponent] to provide access to text layout details.
///
/// This key doesn't need to be public because the given [componentKey]
/// provides clients with direct access to text layout queries, as well as
/// standard [DocumentComponent] queries.
```

super_editor/lib/src/default_editor/text.dart
```diff
@@ -1011,6 +1011,35 @@ class TextComponentState extends State<TextComponent> with DocumentComponent imp
   }
 }
 
+class ProxyTextDocumentComponent extends StatefulWidget {
```
matthew-carroll, line 1014: Dart Doc:

```
/// A [ProxyDocumentComponent] that adds [TextComposable] capabilities so
/// that simple text-based proxy components can meet their expected contract
/// without going through the work of defining a stateful widget that mixes in
/// the [ProxyDocumentComponent] methods.
///
/// Using a [ProxyTextDocumentComponent] is never technically necessary.
/// Custom [DocumentComponent]s can achieve a similar result by mixing in
/// [ProxyDocumentComponent] within a `State` object. This widget is provided
/// as a convenience so that some components can be defined as stateless
/// widgets while still providing access to component behaviors and text layout
/// queries.
```

matthew-carroll: BTW, the line length on all comments that I post on GitHub might be wrong. I space them to avoid new lines on GitHub - the actual lines should probably be longer.

super_editor/lib/src/default_editor/text.dart
```diff
@@ -1011,6 +1011,35 @@ class TextComponentState extends State<TextComponent> with DocumentComponent imp
...
+  const ProxyTextDocumentComponent({
+    super.key,
+    required this.childDocumentComponentKey,
+    required this.child,
+  });
+
+  final GlobalKey<State<StatefulWidget>> childDocumentComponentKey;
```
matthew-carroll, line 1021: The type definition seems redundant. Can a `GlobalKey` ever bind to something other than a `State<StatefulWidget>`?

Also, let's call this `textComponentKey`.

super_editor/lib/src/default_editor/text.dart
```diff
@@ -1011,6 +1011,35 @@ class TextComponentState extends State<TextComponent> with DocumentComponent imp
...
+
+  final Widget child;
```
matthew-carroll, line 1023:
```
/// The widget subtree, which must include a widget that implements 
/// `TextComposable`, and that `TextComposable` must be bound to
/// the given [textComponentKey].
```

super_editor/lib/src/test/super_editor_test/supereditor_inspector.dart
matthew-carroll: Let's rename this to `findTextInComponent()`

matthew-carroll:
@angelosilvestre there are merge conflicts

angelosilvestre:
@matthew-carroll I rebased and fixed the conflicts.

matthew-carroll (approved):
LGTM

PR #1767, opened 2024-01-09
Added a Google Docs example app to prove we can show a selection with out focus, also made expanded selection caret display configurable (Resolves #936)(Resolves #234)

### LINKED ISSUE #234: [SuperEditor] make caret display optional when selection is expanded
Most text editors apparently hide the caret when a selection is expanded. IDEs, however, tend to always display the caret.

Make the display of a caret optional when the selection is expanded.

### LINKED ISSUE #936: [SuperEditor] - Make it possible to display selection without focus
Currently, when `SuperEditor` loses focus, we remove the selection, and then we restore it when focus comes back. This isn't necessarily the desired behavior of an editor.

`SuperEditor` should have a configurable policy that allows it to retain a selection without focus, but have `SuperEditor` hide the caret, so that it doesn't appear as if user input will be sent to `SuperEditor`.

Notice in this video from Google Docs that I'm able to select some text and then open a menu without losing that selection. The editor clearly loses focus, because I'm able to navigate the menu system using keyboard arrow keys. Nonetheless, my selection is retained, albeit in a different color while I'm navigating the menu.

https://github.com/superlistapp/super_editor/assets/7259036/3e2d68aa-78fc-4e55-bba2-be871f4a157d

matthew-carroll:
Added a Google Docs example app to prove we can show a selection with out focus, also made expanded selection caret display configurable (Resolves #936)(Resolves #234)

The original goal of this ticket was to be able to show a selection highlight even when `SuperEditor` doesn't have focus. I discovered that we already implemented this ability. Therefore, I started a new example project that will approximate Google Docs, to ensure that we can achieve desktop editor use-cases.

As a result of this exercise, I ended up making the following enhancements and bug fixes:

- Added a property to configure the `CaretDocumentOverlay` to show or hide the caret when the selection is expanded.
- Fixed caret visibility query for Android and iOS to correctly report `false` when the selection is expanded. I didn't change the mobile caret behavior, but we were incorrectly reporting that the caret was visible when it wasn't.
- Introduced an `EmptyContentLayer` to easily return an empty content layer widget. Returning an `EmptyContentLayer` is like returning a `SizedBox` in a typical widget tree.
- Added a SuperEditor test tools configuration to show/hide the caret for expanded selections.

Here's a video of the new demo app, which shows the focus interaction with the editor selection:

https://github.com/superlistapp/super_editor/assets/7259036/710d6165-0ed0-4978-a0e3-64c666134ccc


super_editor/example_docs/lib/app_menu.dart
```diff
@@ -0,0 +1,314 @@
...
+import 'package:follow_the_leader/follow_the_leader.dart';
+import 'package:super_editor/super_editor.dart';
+
+import 'theme.dart';
+
+/// An application-level popup menu.
+///
+/// A menu button is displayed with the given [label].
+///
+/// When the user clicks the menu button, a popover appears, which displays a the given [items] as
```
angelosilvestre, line 11: "which displays a the given [items]" -> "which displays the given [items]"

super_editor/example_docs/lib/app_menu.dart
```diff
@@ -0,0 +1,314 @@
...
+            // followerOffset: Offset(0, -20),
+          );
+        } else {
+          // There isn't enough room to fully display the follower below or above the leader.
+          // Pin the popover list to the bottom, letting the follower cover the leader.
+          alignment = const FollowerAlignment(
+            leaderAnchor: Alignment.bottomLeft,
+            followerAnchor: Alignment.topLeft,
+            // followerOffset: Offset(0, 20),
+          );
```
angelosilvestre, line 186: Please remove the commented lines.

angelosilvestre (approved):
LGTM with a couple minor comments.

PR #1778, opened 2024-01-10
[SuperTextField][Android][iOS] Disables text field caret blink while dragging caret (Resolves #1705)

### LINKED ISSUE #1705: [SuperTextField][Android][iOS] - Caret blinks during a drag
As of the time that this ticket is filed, `SuperTextField` on Android and iOS continue to blink the caret even as the caret is in the middle of a drag.

The problem can be seen in the video in this other ticket: https://github.com/superlistapp/super_editor/issues/1704

We should check standard Android and iOS behavior, but I don't think the caret is supposed to blink while it's being dragged.

rutvik110:
[SuperTextField][Android][ios] Disables text field caret blink while dragging caret (Resolves #1705)

### **Issue:** 

Text field's caret keeps blinking while caret's being dragged within the text field. On both Android and iOS, in the native apps, the caret stops blinking as long as the user's dragging the caret. Same is expected from super text field on android and ios.

`SuperTextField` behaviour on Android and iOS:

https://github.com/superlistapp/super_editor/assets/65209850/6465ba53-134a-4d05-aeec-3107463c7bee

### Solution: 

This PR updates both `SuperAndroidTextField` and `SuperIOSTextField` and their necessary internal components to match the native behaviour for caret blinking when user's dragging the caret within text field.  As long as the drag's is happening, caret blinking is disabled, and once the drag ends, caret blinking is enabled for the text field.

With fix, `SuperTextField` behaviour on Android and iOS:

https://github.com/superlistapp/super_editor/assets/65209850/3b08287a-e71f-4791-b64d-03ca513ebcee

### Remarks: 

* `SuperAndroidTextField` and `SuperAndroidTextField` are updated to create an internal `BlinkController` that can be passed down to their respective `TextLayoutCaret` that displays and controlls the text field caret. 
* Fixes a bug in the `TextLayoutCaret`'s by updating the `createBlinkController` method to return the passed `BlinkController` if present instead of returning it's own internal `blinkController` variable.
* `AndroidTextFieldTouchInteractor` and `IOSEditingControls` are updated to receive the internal `BlinkController` from text field which is used to control blinking of caret on drag start/end within their drag interaction handlers.
* `AndroidTextFieldTouchInteractor` and `IOSTextFieldTouchInteractor` are updated to receive the internal `BlinkController` from text field which is used to control blinking of caret on drag start/end within their drag interaction handlers.

### DIFF
--- a/super_editor/lib/src/super_textfield/android/_editing_controls.dart
+++ b/super_editor/lib/src/super_textfield/android/_editing_controls.dart
@@ -824,7 +824,6 @@ class AndroidEditingOverlayController with ChangeNotifier {
     super.dispose();
   }
 
-  /// Text field caret blink controller.
   final BlinkController caretBlinkController;
 
   bool _isToolbarVisible = false;
--- a/super_editor/lib/src/super_textfield/android/_editing_controls.dart
+++ b/super_editor/lib/src/super_textfield/android/_editing_controls.dart
@@ -307,7 +307,9 @@ class _AndroidEditingOverlayControlsState extends State<AndroidEditingOverlayCon
       widget.editingController.showMagnifier(_localDragOffset!);
 
       if (widget.editingController.textController.selection.isCollapsed) {
-        widget.editingController.stopCaretBlinking();
+        if (widget.editingController.caretBlinkController.isBlinking) {
+          widget.editingController.stopCaretBlinking();
+        }
       }
 
       _log.fine(' - done updating all local state for drag update');
@@ -936,17 +938,13 @@ class AndroidEditingOverlayController with ChangeNotifier {
   ///
   /// If it's already blinking, does nothing.
   void startCaretBlinking() {
-    if (!caretBlinkController.isBlinking) {
-      caretBlinkController.startBlinking();
-    }
+    caretBlinkController.startBlinking();
   }
 
   /// Stops the text field caret blinking.
   ///
   /// If it's already stopped, does nothing.
   void stopCaretBlinking() {
-    if (caretBlinkController.isBlinking) {
-      caretBlinkController.stopBlinking();
-    }
+    caretBlinkController.stopBlinking();
   }
 }
--- a/super_editor/lib/src/super_textfield/android/_editing_controls.dart
+++ b/super_editor/lib/src/super_textfield/android/_editing_controls.dart
@@ -824,8 +824,6 @@ class AndroidEditingOverlayController with ChangeNotifier {
     super.dispose();
   }
 
-  final BlinkController caretBlinkController;
-
   bool _isToolbarVisible = false;
   bool get isToolbarVisible => _isToolbarVisible;
 
@@ -839,6 +837,8 @@ class AndroidEditingOverlayController with ChangeNotifier {
   /// selection. Those properties and behaviors are represented by
   /// this [textController].
   final AttributedTextEditingController textController;
+
+  final BlinkController caretBlinkController;
 
   void toggleToolbar() {
     if (isToolbarVisible) {
--- a/super_editor/lib/src/super_textfield/android/_editing_controls.dart
+++ b/super_editor/lib/src/super_textfield/android/_editing_controls.dart
@@ -192,9 +192,7 @@ class _AndroidEditingOverlayControlsState extends State<AndroidEditingOverlayCon
   void _onCollapsedPanStart(DragStartDetails details) {
     _log.fine('_onCollapsedPanStart');
 
-    widget.editingController
-      ..hideToolbar()
-      ..cancelCollapsedHandleAutoHideCountdown();
+    _onHandleDragStart();
 
     // TODO: de-dup the calculation of the mid-line focal point
     final globalOffsetInMiddleOfLine =
@@ -228,16 +226,13 @@ class _AndroidEditingOverlayControlsState extends State<AndroidEditingOverlayCon
   void _onBasePanStart(DragStartDetails details) {
     _log.fine('_onBasePanStart');
 
+    _onHandleDragStart();
+
     // TODO: de-dup the calculation of the mid-line focal point
     final globalOffsetInMiddleOfLine =
         _getGlobalOffsetOfMiddleOfLine(widget.editingController.textController.selection.base);
     _touchHandleOffsetFromLineOfText = globalOffsetInMiddleOfLine - details.globalPosition;
     _log.fine(' - global offset in middle of line: $globalOffsetInMiddleOfLine');
-
-    widget.editingController
-      ..hideToolbar()
-      ..cancelCollapsedHandleAutoHideCountdown();
-    _log.fine(' - hid the toolbar, cancelled countdown timer');
 
     // TODO: de-dup the repeated calculations of the effective focal point: globalPosition + _touchHandleOffsetFromLineOfText
     widget.textScrollController.updateAutoScrollingForTouchOffset(
@@ -262,14 +257,12 @@ class _AndroidEditingOverlayControlsState extends State<AndroidEditingOverlayCon
   void _onExtentPanStart(DragStartDetails details) {
     _log.fine('_onExtentPanStart');
 
+    _onHandleDragStart();
+
     // TODO: de-dup the calculation of the mid-line focal point
     final globalOffsetInMiddleOfLine =
         _getGlobalOffsetOfMiddleOfLine(widget.editingController.textController.selection.extent);
     _touchHandleOffsetFromLineOfText = globalOffsetInMiddleOfLine - details.globalPosition;
-
-    widget.editingController
-      ..hideToolbar()
-      ..cancelCollapsedHandleAutoHideCountdown();
 
     // TODO: de-dup the repeated calculations of the effective focal point: globalPosition + _touchHandleOffsetFromLineOfText
     widget.textScrollController.updateAutoScrollingForTouchOffset(
@@ -284,6 +277,19 @@ class _AndroidEditingOverlayControlsState extends State<AndroidEditingOverlayCon
       _isDraggingExtent = true;
       _localDragOffset = (context.findRenderObject() as RenderBox).globalToLocal(details.globalPosition);
     });
+  }
+
+  void _onHandleDragStart() {
+    _log.fine('_onHandleDragStart()');
+
+    widget.editingController
+      ..hideToolbar()
+      ..cancelCollapsedHandleAutoHideCountdown();
+    _log.fine(' - hid the toolbar, cancelled countdown timer');
+
+    if (widget.editingController.textController.selection.isCollapsed) {
+      widget.editingController.stopCaretBlinking();
+    }
   }
 
   void _onPanUpdate(DragUpdateDetails details) {
@@ -305,12 +311,6 @@ class _AndroidEditingOverlayControlsState extends State<AndroidEditingOverlayCon
     setState(() {
       _localDragOffset = _localDragOffset! + details.delta;
       widget.editingController.showMagnifier(_localDragOffset!);
-
-      if (widget.editingController.textController.selection.isCollapsed) {
-        if (widget.editingController.caretBlinkController.isBlinking) {
-          widget.editingController.stopCaretBlinking();
-        }
-      }
 
       _log.fine(' - done updating all local state for drag update');
     });
--- a/super_editor/lib/src/super_textfield/android/_editing_controls.dart
+++ b/super_editor/lib/src/super_textfield/android/_editing_controls.dart
@@ -379,15 +379,15 @@ class _AndroidEditingOverlayControlsState extends State<AndroidEditingOverlayCon
       _isDraggingExtent = false;
       widget.editingController.hideMagnifier();
 
-      if (widget.editingController.textController.selection.isCollapsed) {
-        widget.editingController.startCaretBlinking();
-      }
-
       if (!widget.editingController.textController.selection.isCollapsed) {
         // We hid the toolbar while dragging a handle. If the selection is
         // expanded, show it again.
         widget.editingController.showToolbar();
       } else {
+        // We stop the caret blink while dragging the collapsed handle. If the selection is
+        // collapsed, start the caret blink.
+        widget.editingController.startCaretBlinking();
+
         // The collapsed handle should disappear after some inactivity.
         widget.editingController
           ..unHideCollapsedHandle()
--- a/super_editor/lib/src/super_textfield/android/_editing_controls.dart
+++ b/super_editor/lib/src/super_textfield/android/_editing_controls.dart
@@ -814,7 +814,7 @@ class _AndroidEditingOverlayControlsState extends State<AndroidEditingOverlayCon
 class AndroidEditingOverlayController with ChangeNotifier {
   AndroidEditingOverlayController({
     required this.textController,
-    required this.blinkController,
+    required this.caretBlinkController,
     required LayerLink magnifierFocalPoint,
   }) : _magnifierFocalPoint = magnifierFocalPoint;
 
@@ -823,6 +823,9 @@ class AndroidEditingOverlayController with ChangeNotifier {
     _handleAutoHideTimer?.cancel();
     super.dispose();
   }
+
+  /// Text field caret blink controller.
+  final BlinkController caretBlinkController;
 
   bool _isToolbarVisible = false;
   bool get isToolbarVisible => _isToolbarVisible;
@@ -930,15 +933,12 @@ class AndroidEditingOverlayController with ChangeNotifier {
     }
   }
 
-  /// Text field caret blink controller.
-  final BlinkController blinkController;
-
   /// Starts the text field caret blinking.
   ///
   /// If it's already blinking, does nothing.
   void startCaretBlinking() {
-    if (!blinkController.isBlinking) {
-      blinkController.startBlinking();
+    if (!caretBlinkController.isBlinking) {
+      caretBlinkController.startBlinking();
     }
   }
 
@@ -946,8 +946,8 @@ class AndroidEditingOverlayController with ChangeNotifier {
   ///
   /// If it's already stopped, does nothing.
   void stopCaretBlinking() {
-    if (blinkController.isBlinking) {
-      blinkController.stopBlinking();
+    if (caretBlinkController.isBlinking) {
+      caretBlinkController.stopBlinking();
     }
   }
 }
--- a/super_editor/lib/src/super_textfield/android/_editing_controls.dart
+++ b/super_editor/lib/src/super_textfield/android/_editing_controls.dart
@@ -305,7 +305,7 @@ class _AndroidEditingOverlayControlsState extends State<AndroidEditingOverlayCon
     setState(() {
       _localDragOffset = _localDragOffset! + details.delta;
       widget.editingController.showMagnifier(_localDragOffset!);
-      widget.editingController.blinkController.stopBlinking();
+      widget.editingController.stopCaretBlinking();
       _log.fine(' - done updating all local state for drag update');
     });
   }
@@ -374,7 +374,7 @@ class _AndroidEditingOverlayControlsState extends State<AndroidEditingOverlayCon
       _isDraggingBase = false;
       _isDraggingExtent = false;
       widget.editingController.hideMagnifier();
-      widget.editingController.blinkController.startBlinking();
+      widget.editingController.startCaretBlinking();
 
       if (!widget.editingController.textController.selection.isCollapsed) {
         // We hid the toolbar while dragging a handle. If the selection is
@@ -831,9 +831,6 @@ class AndroidEditingOverlayController with ChangeNotifier {
   /// this [textController].
   final AttributedTextEditingController textController;
 
-  /// Text field caret blink controller.
-  final BlinkController blinkController;
-
   void toggleToolbar() {
     if (isToolbarVisible) {
       hideToolbar();
@@ -923,6 +920,27 @@ class AndroidEditingOverlayController with ChangeNotifier {
     if (!_isCollapsedHandleAutoHidden) {
       _isCollapsedHandleAutoHidden = true;
       notifyListeners();
+    }
+  }
+
+  /// Text field caret blink controller.
+  final BlinkController blinkController;
+
+  /// Starts the text field caret blinking.
+  ///
+  /// If it's already blinking, does nothing.
+  void startCaretBlinking() {
+    if (!blinkController.isBlinking) {
+      blinkController.startBlinking();
+    }
+  }
+
+  /// Stops the text field caret blinking.
+  ///
+  /// If it's already stopped, does nothing.
+  void stopCaretBlinking() {
+    if (blinkController.isBlinking) {
+      blinkController.stopBlinking();
     }
   }
 }
--- a/super_editor/lib/src/super_textfield/android/_editing_controls.dart
+++ b/super_editor/lib/src/super_textfield/android/_editing_controls.dart
@@ -32,7 +32,6 @@ class AndroidEditingOverlayControls extends StatefulWidget {
     Key? key,
     required this.editingController,
     required this.textScrollController,
-    required this.blinkController,
     required this.textFieldKey,
     required this.textContentKey,
     required this.textFieldLayerLink,
@@ -51,9 +50,6 @@ class AndroidEditingOverlayControls extends StatefulWidget {
   /// Controller that auto-scrolls text based on handle
   /// location.
   final TextScrollController textScrollController;
-
-  /// Text field caret blink controller.
-  final BlinkController blinkController;
 
   /// [LayerLink] that is anchored to the text field's boundary.
   final LayerLink textFieldLayerLink;
@@ -309,7 +305,7 @@ class _AndroidEditingOverlayControlsState extends State<AndroidEditingOverlayCon
     setState(() {
       _localDragOffset = _localDragOffset! + details.delta;
       widget.editingController.showMagnifier(_localDragOffset!);
-      widget.blinkController.stopBlinking();
+      widget.editingController.blinkController.stopBlinking();
       _log.fine(' - done updating all local state for drag update');
     });
   }
@@ -378,7 +374,7 @@ class _AndroidEditingOverlayControlsState extends State<AndroidEditingOverlayCon
       _isDraggingBase = false;
       _isDraggingExtent = false;
       widget.editingController.hideMagnifier();
-      widget.blinkController.startBlinking();
+      widget.editingController.blinkController.startBlinking();
 
       if (!widget.editingController.textController.selection.isCollapsed) {
         // We hid the toolbar while dragging a handle. If the selection is
@@ -811,6 +807,7 @@ class _AndroidEditingOverlayControlsState extends State<AndroidEditingOverlayCon
 class AndroidEditingOverlayController with ChangeNotifier {
   AndroidEditingOverlayController({
     required this.textController,
+    required this.blinkController,
     required LayerLink magnifierFocalPoint,
   }) : _magnifierFocalPoint = magnifierFocalPoint;
 
@@ -833,6 +830,9 @@ class AndroidEditingOverlayController with ChangeNotifier {
   /// selection. Those properties and behaviors are represented by
   /// this [textController].
   final AttributedTextEditingController textController;
+
+  /// Text field caret blink controller.
+  final BlinkController blinkController;
 
   void toggleToolbar() {
     if (isToolbarVisible) {
--- a/super_editor/lib/src/super_textfield/android/_editing_controls.dart
+++ b/super_editor/lib/src/super_textfield/android/_editing_controls.dart
@@ -32,6 +32,7 @@ class AndroidEditingOverlayControls extends StatefulWidget {
     Key? key,
     required this.editingController,
     required this.textScrollController,
+    required this.blinkController,
     required this.textFieldKey,
     required this.textContentKey,
     required this.textFieldLayerLink,
@@ -50,6 +51,9 @@ class AndroidEditingOverlayControls extends StatefulWidget {
   /// Controller that auto-scrolls text based on handle
   /// location.
   final TextScrollController textScrollController;
+
+  /// Text field caret blink controller.
+  final BlinkController blinkController;
 
   /// [LayerLink] that is anchored to the text field's boundary.
   final LayerLink textFieldLayerLink;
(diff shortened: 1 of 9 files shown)


super_editor/lib/src/super_textfield/android/_editing_controls.dart
```diff
@@ -298,6 +302,7 @@ class _AndroidEditingOverlayControlsState extends State<AndroidEditingOverlayCon
     setState(() {
       _localDragOffset = _localDragOffset! + details.delta;
       widget.editingController.showMagnifier(_localDragOffset!);
+      widget.blinkController.stopBlinking();
```
matthew-carroll, line 305: This is calling `stopBlinking()` 60 times per second while dragging, which seems unnecessary.

super_editor/lib/src/super_textfield/android/_editing_controls.dart
```diff
@@ -50,6 +51,9 @@ class AndroidEditingOverlayControls extends StatefulWidget {
   /// location.
   final TextScrollController textScrollController;
 
+  /// Text field caret blink controller.
+  final BlinkController blinkController;
```
matthew-carroll, line 55: Given that `AndroidEditingOverlayController` is the controller that links this overlay UI to the text field, wouldn't that be the correct place to introduce blinking start/stop controls?

super_editor/lib/src/super_textfield/android/_user_interaction.dart
```diff
@@ -77,6 +78,9 @@ class AndroidTextFieldTouchInteractor extends StatefulWidget {
 
   final TextScrollController textScrollController;
 
+  /// Text field caret blink controller.
+  final BlinkController blinkController;
```
matthew-carroll, line 82: Continuing with my comment from the overlay - we already have a controller that shares visual updates between the handles and the text field. It's `AndroidEditingOverlayController`. Shouldn't the blinking APIs be added there?

super_editor/lib/src/super_textfield/android/_user_interaction.dart
```diff
@@ -335,6 +339,7 @@ class AndroidTextFieldTouchInteractorState extends State<AndroidTextFieldTouchIn
       );
 
       widget.editingOverlayController.showMagnifier(_globalDragOffset!);
+      widget.blinkController.stopBlinking();
```
matthew-carroll, line 342: This one also gets called 60 times per second. That seems unnecessary.

super_editor/lib/src/super_textfield/android/_user_interaction.dart
```diff
@@ -359,6 +364,7 @@ class AndroidTextFieldTouchInteractorState extends State<AndroidTextFieldTouchIn
     setState(() {
       _isDraggingCaret = false;
       widget.editingOverlayController.hideMagnifier();
+      widget.blinkController.startBlinking();
```
matthew-carroll, line 367: Is it possible that the selection is expanded here? If so, what happens when you call start blinking with an expanded selection?

super_editor/lib/src/super_textfield/android/android_textfield.dart
```diff
@@ -628,6 +642,7 @@ class SuperAndroidTextFieldState extends State<SuperAndroidTextField>
                 ? _textEditingController.selection.extent
                 : null,
             blinkTimingMode: widget.blinkTimingMode,
+            blinkController: _blinkController,
```
matthew-carroll, line 645: What does it mean to pass a blink controller and pass a blink timing mode?

matthew-carroll (commented):
I left a number of comments on the Android approach. They probably apply similarly to the iOS approach. Also, you're missing tests.

super_editor/lib/src/super_textfield/android/_editing_controls.dart
rutvik110: yah, that's right.  Will update this to include the `blinkController` as part of `AndroidEditingOverlayController`.

super_editor/lib/src/super_textfield/android/_user_interaction.dart
rutvik110: Technically, it shouldn't affect anything. `start/stop` on `blinkController` is mainly responsible for starting/stoping it's ticker that's used to flash the caret, if one's not being shown as in case of expanded selection, then it shouldn't have any effect. Also, tested it out and didn't saw any side effects of calling those here.

super_editor/lib/src/super_textfield/android/android_textfield.dart
rutvik110: ah well, the `blinkTimingMode` is used by `TextLayoutCaret` to create a ticker/timer based `blinkController` but as we're creating the controller above, it shouldn't matter if we pass it or not. But, `TextLayoutCaret` may've other use of `blinkTimingMode`, so I'm gonna keep it as it's.

rutvik110:
@matthew-carroll Moved the caret start/stop controlls within the `editingController` for respective platforms. And also updated to signal start or stop blinking only if the caret has either stoped blinking or is blinking.  Tests are also added to ensure caret doesn't blink when dragging on android and ios, and when dragging the collapsed handle on android.

super_editor/lib/src/super_textfield/android/_user_interaction.dart
matthew-carroll: Please keep in mind the ability for future developers to make changes to code. When you say "Technically, it shouldn't affect anything", it sounds like what you're saying is "This behavior isn't relevant for expanded selections and shouldn't be called in that case, but it doesn't break anything if we do".

The problem is, how does everyone else know this? The default assumption when reading all code, is that the code you're reading exists for a reason. Therefore, other developers may be hesitant to take action that removes `startBlinking()` for expanded selections. Moreover, developers might avoid making convenient changes for expanded selections because they think that their changes need to continue working with the `startBlinking()` line.

It's generally harmful to let code exist where it shouldn't simply because that code doesn't currently break something.

super_editor/lib/src/super_textfield/android/android_textfield.dart
matthew-carroll: > But, TextLayoutCaret may've other use of blinkTimingMode,

This isn't something to be left to assumptions. Does it make other use of it, or not?

> ah well, the blinkTimingMode is used by TextLayoutCaret to create a ticker/timer based blinkController but as we're creating the controller above, it shouldn't matter if we pass it or not

For the same reason as my other comment on this PR, this perspective is generally unacceptable. We don't want to litter incorrect-but-not-breaking code just because we can. We want to maintain a codebase of intentional behaviors that do exactly what's intended.

> so I'm gonna keep it as it's.

Based on the rationale in your comment, this is probably the wrong conclusion.

super_editor/lib/src/super_textfield/android/_user_interaction.dart
rutvik110: That's right. Updated to call start/stop blinking only when the selection is collapsed.

super_editor/lib/src/super_textfield/android/android_textfield.dart
rutvik110: > This isn't something to be left to assumptions. Does it make other use of it, or not?

It doesn't atm. 

> For the same reason as my other comment on this PR, this perspective is generally unacceptable. We don't want to litter incorrect-but-not-breaking code just because we can. We want to maintain a codebase of intentional behaviors that do exactly what's intended.

That's true but `TextLayoutCaret` sets `blinkTimingMode` default value to be ticker based. If the user happens to change this to timer based, and we didn't passed the user defined mode to `blinkTimingMode` in this case, then we would eventually have a false value for `blinkTimingMode` within `TextLayoutCaret`. Even if it self doesn't use it for anything other than creating the `blinkController` atm but if something else in the tree has to access it's `blinkTimingMode`, then it would work with a false value eventually. Based on this, shouldn't keeping it is the right choice here?

rutvik110: @matthew-carroll Pinging in case you missed this.

matthew-carroll: I couldn't follow your explanation above. First, what specific API are you proposing? Second, what's a specific situation that requires the API you're proposing? Third, if you're proposing an API that takes a timing mode and a controller, how do those things not conflict with each other?

rutvik110: > First, what specific API are you proposing? Second, what's a specific situation that requires the API you're proposing? Third, if you're proposing an API that takes a timing mode and a controller, how do those things not conflict with each other?

I think this whole discussion is getting dragged by hypothetical what if questions from my side. For the moment, from our discussion and as I see it now, there's no need to pass the `BlinkTimingMode` here if we're already passing the controller. If there's any need in the future for passing `BlinkTimingMode` in presence of controller, we'll know it. But right now, this doesn't affect anything, so we're good.

Based on this, I'm removing the passed `blinkTimingMode` from here for now. Does that sound good?

matthew-carroll: Sure. I can take another look after you try that.

rutvik110: Pushed the updates. Let me know if we're good here.

super_editor/test/super_textfield/super_textfield_caret_test.dart
```diff
@@ -166,6 +167,60 @@ void main() {
...
+
+      await _pumpTestApp(tester, controller: controller);
+
+      await tester.placeCaretInSuperTextField(0);
+
+      // Drag caret by an arbitrary distance.
+      await tester.dragCaretByDistanceInSuperTextField(const Offset(100, 100));
+
+      // Ensure caret is visible.
+      expect(_isCaretVisible(tester), true);
```
matthew-carroll, line 185: I have two questions about this check. Did you make sure to re-enable caret animations for this test? If not, the caret will probably never blink.

Second, does `_isCaretVisible()` return `false` when the caret opacity is zero?

Third, did you check to make sure this test fails without your changes?

super_editor/test/super_textfield/super_textfield_caret_test.dart
```diff
@@ -166,6 +167,60 @@ void main() {
       expect(caret.style.width, caretStyle.width);
       expect(caret.style.borderRadius, caretStyle.borderRadius);
     });
+
+    testWidgetsOnMobile("does not blink while dragging", (tester) async {
```
matthew-carroll, line 171: To clarify how this test differs from the one below it, please add the ending "does not blink while dragging the caret"

super_editor/test/super_textfield/super_textfield_caret_test.dart
```diff
@@ -196,3 +251,29 @@ bool _isCaretVisible(WidgetTester tester) {
   final caretPainter = tester.widget<CustomPaint>(customPaint.last).painter as CaretPainter;
   return caretPainter.blinkController!.opacity == 1.0;
 }
+
+/// Pumps a [SuperTextField] experience.
+Future<void> _pumpTestApp(
```
matthew-carroll, line 256: A few things here. First, higher level methods should appear above lower level methods. For example, pumping a whole app is higher level than pumping just a scaffold.

Second, why doesn't this method re-use the existing `_buildScaffold()`? You seem to have used the same root widget tree.

Third, this test name introduces confusion. It begs the question "why isn't this method used for all tests in this suite if it's really a "test app". If this method pumps something more specific, then it should be named accordingly.

Fourth, why not just pump the `SuperTextField` that you want directly in the tests like all other tests in this suite? It's only two tests, there's no special configuration, and that would retain consistency with everything else in this file.

super_editor/lib/src/super_textfield/android/_editing_controls.dart
```diff
@@ -374,6 +379,10 @@ class _AndroidEditingOverlayControlsState extends State<AndroidEditingOverlayCon
       _isDraggingExtent = false;
       widget.editingController.hideMagnifier();
 
+      if (widget.editingController.textController.selection.isCollapsed) {
```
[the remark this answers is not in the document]
rutvik110: ah right, fixed it.

super_editor/test/super_textfield/super_textfield_caret_test.dart
rutvik110: > Did you make sure to re-enable caret animations for this test? If not, the caret will probably never blink.

If you're referring to this check within some of the above tests, then it's not present rn in this added tests. I assumed the caret blinking was on by default. I'll update the tests with this.
```dart
// Configure BlinkController to animate, otherwise it won't blink
BlinkController.indeterminateAnimationsEnabled = true;
```

> Second, does _isCaretVisible() return false when the caret opacity is zero?

yes, `_isCaretVisible` explicitly checks for the opacity to ensure caret's visible and based on existing tests, it does.

> Third, did you check to make sure this test fails without your changes?

yes

super_editor/lib/src/super_textfield/android/_editing_controls.dart
```diff
@@ -305,6 +305,11 @@ class _AndroidEditingOverlayControlsState extends State<AndroidEditingOverlayCon
     setState(() {
       _localDragOffset = _localDragOffset! + details.delta;
       widget.editingController.showMagnifier(_localDragOffset!);
+
+      if (widget.editingController.textController.selection.isCollapsed) {
```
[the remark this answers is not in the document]
rutvik110: I'll check on this.

super_editor/test/super_textfield/super_textfield_caret_test.dart
rutvik110: This was based on an earlier implementation of the tests but  is redundant now. Removed it.

super_editor/test/super_textfield/super_textfield_caret_test.dart
```diff
@@ -166,6 +167,60 @@ void main() {
...
+
+      await tester.placeCaretInSuperTextField(0);
+
+      // Drag caret by an arbitrary distance.
+      await tester.dragCaretByDistanceInSuperTextField(const Offset(100, 100));
+
+      // Ensure caret is visible.
+      expect(_isCaretVisible(tester), true);
+
+      await tester.pump(flashPeriod);
```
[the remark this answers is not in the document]
rutvik110: Updated.

super_editor/lib/src/super_textfield/android/_editing_controls.dart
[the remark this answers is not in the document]
rutvik110: Based on our discussion, I've moved this call to pan start event callbacks for both handle and user interaction handlers on text field.

I noticed that overlay controllers `IOSEditingControls` and `AndroidEditingOverlayControls` were missing a `onPanStart` method so I updated them include a `onHandleDragStart()` method that's called within the respective `onPanStart` methods passed down to created handles.  

`onHandleDragStart` is responsible for calling common drag start events like hiding the toolbar, etc.

`AndroidEditingOverlayControls` methods  `_onBasePanStart`, `_onExtentPanStart` and `_onCollapsedPanStart` are updated to use `onHandleDragStart`.

And similarly `IOSEditingControls` methods `_onBasePanStart` and `_onExtentPanStart` are updated as well.

Beside moving the blink stop call within `onHandleDragStart`, some of the common logic that was shared across this pan start methods is moved within it as well.

rutvik110:
@matthew-carroll Updated things based on prev review. Major updates are related to

* how we handle the blink stop call from interaction handles as I mentioned [here](https://github.com/superlistapp/super_editor/pull/1778#discussion_r1456402530) and other updates
* Tests are updated to check caret visibility after half caret flash period based on [this](https://github.com/superlistapp/super_editor/pull/1778#discussion_r1452829730) comment.

super_editor/test/super_textfield/super_textfield_caret_test.dart
```diff
@@ -166,6 +167,96 @@ void main() {
       expect(caret.style.width, caretStyle.width);
       expect(caret.style.borderRadius, caretStyle.borderRadius);
     });
+
+    testWidgetsOnMobile("does not blink while dragging the caret", (tester) async {
+      // Disable indeterminate animations.
+      BlinkController.indeterminateAnimationsEnabled = false;
```
matthew-carroll, line 173: This line should be unnecessary. Animations are already false at the beginning of every test.

However, the tear down is necessary, so please add a comment explaining why we're adding a teardown.

super_editor/test/super_textfield/super_textfield_caret_test.dart
```diff
@@ -166,6 +167,96 @@ void main() {
...
+
+      await tester.placeCaretInSuperTextField(0);
+
+      // Configure BlinkController to animate, otherwise it won't blink.
+      BlinkController.indeterminateAnimationsEnabled = true;
+
+      // Drag caret by an arbitrary distance.
+      await tester.dragCaretByDistanceInSuperTextField(const Offset(100, 100));
+
+      // Ensure caret is visible.
```
matthew-carroll, line 197: Add a comment here that explains why the following series of pumps and checks are important. Namely that we don't know exactly how much time has passed, and we might get unlucky and check the visibility when the caret is momentarily visible. Therefore, we test in half-period increments to ensure that we verify the caret isn't blinking at all.

super_editor/test/super_textfield/super_textfield_caret_test.dart
```diff
@@ -166,6 +167,96 @@ void main() {
...
+      expect(_isCaretVisible(tester), true);
+
+      await tester.pump(flashPeriod ~/ 2);
+
+      // Ensure caret is visible.
+      expect(_isCaretVisible(tester), true);
+    });
+
+    testWidgetsOnAndroid("does not blink while dragging collapsed handle", (tester) async {
+      // Disable indeterminate animations.
```
matthew-carroll, line 217: Please apply any relevant comments from the previous test in this one.

super_text_layout/lib/src/caret_layer.dart
```diff
@@ -77,7 +77,7 @@ class TextLayoutCaretState extends State<TextLayoutCaret> with TickerProviderSta
 
   BlinkController _createBlinkController() {
```
matthew-carroll, line 78: Let's rename this method from `_createBlinkController()` to `_obtainBlinkController()`

rutvik110:
@matthew-carroll Latest updates,

* Added tests to ensure caret doesn't blink when dragging expanded selection handles in text field.
* Added `dragHandleByDistanceInSuperTextFieldOnMobile` - A top level helper to drag a handle of specific type within text field on mobile.
* Added `_dragAndroidHandleByDistanceInSuperTextField` - Handles android text field handle drag.
* Added `_dragIOSHandleByDistanceInSuperTextField` - Handles iOS text field handle drag.
* Moved `caretBlinkController` dispose call out of schedule frame callback within `SuperAndroidTextField` and `SuperIOSTextField`.
* Other documentation and refactoring updates from last review.

super_editor/test/super_textfield/super_textfield_caret_test.dart
```diff
@@ -166,6 +167,182 @@ void main() {
       expect(caret.style.width, caretStyle.width);
       expect(caret.style.borderRadius, caretStyle.borderRadius);
     });
+
+    testWidgetsOnMobile("does not blink while dragging the caret", (tester) async {
+      // Disable indeterminate animations so that pumpAndSettle() doesn't time out.
```
matthew-carroll, line 172: This comment no longer applies here. Please verify that you don't have any other stale comments in this PR.

super_editor/test/super_textfield/super_textfield_caret_test.dart
```diff
@@ -166,6 +167,182 @@ void main() {
...
+          ),
+        ),
+      );
+
+      await tester.placeCaretInSuperTextField(0);
+
+      // Configure BlinkController to animate, otherwise it won't blink.
+      BlinkController.indeterminateAnimationsEnabled = true;
+
+      // Drag caret by an arbitrary distance.
```
matthew-carroll, line 194: We should clarify why we're dragging a distance. Readers might assume that this test is focused on what's happening during this drag, but that isn't really the case.

```
// Drag caret by a small distance so that we trigger a user drag event.
// This drag event is continued down below so that we can check for blinking
// during a user drag.
```

super_editor/test/super_textfield/super_textfield_caret_test.dart
```diff
@@ -166,6 +167,182 @@ void main() {
...
+      await tester.dragCaretByDistanceInSuperTextField(const Offset(100, 100));
```
matthew-carroll, line 195: We mentioned on the call that you need to get the gesture from this method and then end that gesture at the end of the test.

super_editor/test/super_textfield/super_textfield_robot.dart
```diff
@@ -122,6 +123,97 @@ extension SuperTextFieldRobot on WidgetTester {
     return gesture;
   }
 
+  /// Drags the [SuperTextField] handle by the given delta.
+  ///
+  /// {@macro supertextfield_finder}
+  Future<TestGesture> dragHandleByDistanceInSuperTextFieldOnMobile(
```
matthew-carroll, line 129: Can you remind me why you needed to create this? Weren't you already using a method for this when we had our call? Do we not already have dragging behavior in this robot?

super_editor/test/super_textfield/super_textfield_caret_test.dart
rutvik110: Updated.

super_editor/test/super_textfield/super_textfield_robot.dart
rutvik110: The existing `dragAndroidCollapsedHandleByDistanceInSuperTextField` could only be used for dragging the collapsed handle on `android`. And the earlier tests were testing only for collapsed handle but didn't consider the extended selection handle drag scenario.

After  I added tests for testing the handle drag during the extended text selection, I had to add these methods to support dragging the handles for both android and iOS in that scenario.

super_editor/test/super_textfield/super_textfield_caret_test.dart
```diff
@@ -166,6 +167,202 @@ void main() {
       expect(caret.style.width, caretStyle.width);
       expect(caret.style.borderRadius, caretStyle.borderRadius);
     });
+
+    testWidgetsOnMobile("does not blink while dragging the caret", (tester) async {
+      addTearDown(() => BlinkController.indeterminateAnimationsEnabled = false);
+
+      final controller = AttributedTextEditingController(
+        text: AttributedText(
+          '''SuperTextField with a content that spans multiple lines of text to test scrolling with  a scrollbar.''',
```
matthew-carroll, line 176: This string doesn't need to use triple quotes. This is a regular single-line string. Please update everywhere in this PR, accordingly.

super_editor/test/super_textfield/super_textfield_caret_test.dart
```diff
@@ -166,6 +167,202 @@ void main() {
...
+      await tester.pump(flashPeriod ~/ 2);
+
+      // Ensure caret is visible.
+      expect(_isCaretVisible(tester), true);
+
+      await tester.pump(flashPeriod ~/ 2);
+
+      // Ensure caret is visible.
+      expect(_isCaretVisible(tester), true);
+    });
```
matthew-carroll, line 221: You still aren't releasing the gesture.

super_editor/test/super_textfield/super_textfield_caret_test.dart
```diff
@@ -166,6 +167,202 @@ void main() {
...
+      // End the current drag gesture before we start the downstream handle drag.
+      // We don't want multiple gesture pointers active at the same time.
+      await gesture.up();
+      await tester.pump();
+
+      // Drag the downstream selection handle by a small distance so that we trigger a
+      // user drag event.
+      // This drag event is continued down below so that we can check for caret blinking
+      // during a user drag.
+      final TestGesture gesture2 =
```
matthew-carroll, line 285: Calling something `2` doesn't communicate anything. Please name the gestures in a way that describes what differentiates them, e.g., `upstreamHandleGesture` and `downstreamHandleGesture`

super_editor/test/super_textfield/super_textfield_caret_test.dart
```diff
@@ -166,6 +167,202 @@ void main() {
...
+      expect(_isCaretVisible(tester), true);
+
+      // End the current drag gesture before we start the downstream handle drag.
+      // We don't want multiple gesture pointers active at the same time.
+      await gesture.up();
+      await tester.pump();
+
+      // Drag the downstream selection handle by a small distance so that we trigger a
+      // user drag event.
+      // This drag event is continued down below so that we can check for caret blinking
```
matthew-carroll, line 283: Please be mindful of your comment formatting.

If you want multiple sentences together, then the second sentence should immediately follow the first.

If you want two independent sentences, then please add a blank line between them.

super_editor/test/super_textfield/super_textfield_robot.dart
matthew-carroll: If we already have a `dragAndroidCollapsedHandleByDistanceInSuperTextField` then rather than add a generic method for any handle type, which creates an ambiguous crossover with the existing method, let's offer explicit handle methods:

`dragUpstreamMobileHandleByDistanceInSuperTextField` and `dragDownstreamMobileHandleByDistanceInSuperTextField`

super_editor/test/super_textfield/super_textfield_robot.dart
```diff
@@ -122,6 +123,97 @@ extension SuperTextFieldRobot on WidgetTester {
...
+    );
+
+    expect(handleFinder, findsOne);
+
+    final handleCenter = getCenter(handleFinder);
+
+    final gesture = await startGesture(handleCenter);
+    await pump(kTapMinTime);
+
+    for (int i = 0; i < 50; i += 1) {
```
matthew-carroll, line 175: Does all of this really need to be repeated? Is there any meaningful difference between teh Android and iOS methods other than the finder for the handle widget?

super_editor/test/super_textfield/super_textfield_caret_test.dart
rutvik110: I added it as a `addTearDown` step after we create the gesture above. 

```
final TestGesture gesture = await tester.dragCaretByDistanceInSuperTextField(const Offset(100, 100));
addTearDown(() => gesture.removePointer());
```

super_editor/test/super_textfield/super_textfield_caret_test.dart
rutvik110: Updated

super_editor/test/super_textfield/super_textfield_robot.dart
rutvik110: No, there isn't. I think a separate `dragMobileHandle` could be created that'll take in the desired finder and use it, while keeping the other logic intact. That way we'll be able to de-dupe this drag steps across similar methods here.

rutvik110:
@matthew-carroll Latest updates, 

* Added explicit methods `dragUpstreamMobileHandleByDistanceInSuperTextField` and `dragDownstreamMobileHandleByDistanceInSuperTextField` to drag upstream or downstream handle on text field.
* Added explicit drag methods for android and ios, 
   * `_dragAndroidUpstreamHandleByDistanceInSuperTextField` 
   * `_dragAndroidDownstreamHandleByDistanceInSuperTextField`
   * `_dragIOSUpstreamHandleByDistanceInSuperTextField`
   * `_dragIOSDownstreamHandleByDistanceInSuperTextField`
* `_dragHandleByDistanceInSuperTextField` added to abstract away the drag logic on the handle.
* Docs updates based on last review.

super_editor/test/super_textfield/super_textfield_robot.dart
```diff
@@ -122,6 +123,168 @@ extension SuperTextFieldRobot on WidgetTester {
...
+    );
+
+    expect(handleFinder, findsOne);
+
+    final gesture = await _dragHandleByDistanceInSuperTextField(handleFinder, delta);
+
+    return gesture;
+  }
+
+  /// Drags the [SuperTextField] handle by the given delta.
```
matthew-carroll, line 270: The purpose of Dart Docs is to help people know when/how to use the method. This Dart Doc should include a reference to, and explanation of `superTextFieldHandleFinder`.

super_editor/test/super_textfield/super_textfield_robot.dart
```diff
@@ -122,6 +123,168 @@ extension SuperTextFieldRobot on WidgetTester {
...
+    );
+
+    expect(handleFinder, findsOne);
+
+    final gesture = await _dragHandleByDistanceInSuperTextField(handleFinder, delta);
+
+    return gesture;
+  }
+
+  /// Drags the [SuperIOSTextField] downstream handle by the given delta.
```
matthew-carroll, line 245: You should also mention that the method returns the live drag gesture. Please update all such occurrences of docs.

rutvik110:
@matthew-carroll Updated the docs on newly added methods based on last review.

super_editor/test/super_textfield/super_textfield_robot.dart
```diff
@@ -122,6 +123,185 @@ extension SuperTextFieldRobot on WidgetTester {
...
+
+    return gesture;
+  }
+
+  /// Drags the [SuperTextField] handle found by [superTextFieldHandleFinder] by
+  /// the given delta.
+  ///
+  /// Can be used to drag [SuperTextField] collapsed or expanded selection handles.
+  ///
+  /// Returns the [TestGesture] used to perform the drag.
```
matthew-carroll, line 287: The first sentence should mention what's returned. The first sentence should operate as a summary of the whole method.

It looks like we've had quite a few comments about Dart Docs in this PR. I would encourage you to read through a number of our existing Dart Docs to get a feel for how they should be written so that we can spend much fewer cycles going back and forth about how to write them effectively.

super_editor/test/super_textfield/super_textfield_robot.dart
```diff
@@ -122,6 +123,185 @@ extension SuperTextFieldRobot on WidgetTester {
...
+      matching: find.byWidgetPredicate(
+        (widget) =>
+            widget is IOSSelectionHandle && //
+            widget.handleType == HandleType.downstream,
+      ),
+    );
+
+    expect(handleFinder, findsOne);
+
+    final gesture = await _dragHandleByDistanceInSuperTextField(handleFinder, delta);
```
matthew-carroll, line 277: You can return this directly. There's no need to assign it to a variable.

super_editor/test/super_textfield/super_textfield_robot.dart
```diff
@@ -122,6 +123,185 @@ extension SuperTextFieldRobot on WidgetTester {
...
+
+    return gesture;
+  }
+
+  /// Drags the [SuperIOSTextField] downstream handle by the given delta.
+  ///
+  /// Returns the [TestGesture] used to perform the drag.
```
matthew-carroll, line 257: Same issue here about where the "return" description belongs.

matthew-carroll (approved):
LGTM

PR #1779, opened 2024-01-11
[SuperEditor][SuperReader] Adds empty document constructor to MutableDocument (Resolves #1758)

### LINKED ISSUE #1758: [SuperEditor][SuperReader] - Make it easier to create an empty document
Most developers will get started with `SuperEditor` by initially displaying an empty document. This first requires the creation of an empty document. Currently, it's overly verbose to create such a document:

```dart
@override
  void initState() {
    super.initState();

    _document = MutableDocument(nodes: [ParagraphNode(id: Editor.createNodeId(), text: AttributedText())]);
    _composer = MutableDocumentComposer();
  }
```

We should offer a `.empty()` factory method to create an equivalent document on the user's behalf:

```dart
_document = MutableDocument.empty();
```

After adding this factory method, we should update all places where we currently create an empty document using the overly verbose approach.

rutvik110:
[SuperEditor][SuperReader] Adds empty document constructor to MutableDocument (Resolves #1758)

### **Issue:** 

Currently, creating an empty document is somewhat verbose and makes it a bit difficult for users to get started with `super_editor`.

Current approach to creating an empty document: 

```dart
_document = MutableDocument(nodes: [ParagraphNode(id: Editor.createNodeId(), text: AttributedText())]);
```

### **Solution:** 

This PR adds an `.empty()` factory constructor to `MutableDocument` which will create the empty document using exactly what's mentioned above but hides away the complexity of creating one and allows users to get started quickly with using the package.

```dart
  factory MutableDocument.empty() {
    return MutableDocument(
      nodes: [
        ParagraphNode(
          id: Editor.createNodeId(),
          text: AttributedText(),
        ),
      ],
    );
  }
``` 

With this, users can create an empty document by simply using the new `.empty` factory constructor.
```dart
_document = MutableDocument.empty();
```

### DIFF
--- a/super_editor/lib/src/core/editor.dart
+++ b/super_editor/lib/src/core/editor.dart
@@ -498,12 +498,14 @@ class MutableDocument implements Document, Editable {
     _refreshNodeIdCaches();
   }
 
-  /// Creates an empty [Document].
-  factory MutableDocument.empty() {
+  /// Creates an [Document] with a single [ParagraphNode].
+  ///
+  /// Optionally, takes in a [nodeId] for the [ParagraphNode].
+  factory MutableDocument.empty([String? nodeId]) {
     return MutableDocument(
       nodes: [
         ParagraphNode(
-          id: Editor.createNodeId(),
+          id: nodeId ?? Editor.createNodeId(),
           text: AttributedText(),
         ),
       ],
--- a/super_editor/lib/src/core/editor.dart
+++ b/super_editor/lib/src/core/editor.dart
@@ -498,7 +498,7 @@ class MutableDocument implements Document, Editable {
     _refreshNodeIdCaches();
   }
 
-  /// Creates an empty [MutableDocument].
+  /// Creates an empty [Document].
   factory MutableDocument.empty() {
     return MutableDocument(
       nodes: [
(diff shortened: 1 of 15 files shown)


super_editor/lib/src/core/editor.dart
```diff
@@ -496,6 +498,18 @@ class MutableDocument implements Document, Editable {
     _refreshNodeIdCaches();
   }
 
+  /// Creates an empty [Document].
+  factory MutableDocument.empty() {
```
matthew-carroll, line 502: Let's add an optional parameter so that a user can specify a desired node ID for the first node:

```
factory MutableDocument.empty([String? emptyParagraphId])
```

In tests, it's often desirable to control the node ID for verification without looking up the node ID.

Also, can you confirm that you exhaustively looked for empty document usages? I was expecting more changes than what I'm seeing here. I'm especially surprised that I don't see any test changes. Did we really not have a single empty document in any of our tests?

rutvik110: > Also, can you confirm that you exhaustively looked for empty document usages?

Missed that, will check for those and update if there are any such instances.

matthew-carroll (approved):
LGTM

matthew-carroll:
@rutvik110 the auto cherry pick failed for this for some reason. Did you push from a fork instead of in this repo? Either way, please put up a cherry pick PR to stable.

rutvik110:
> @rutvik110 Did you push from a fork instead of in this repo? 

Nope. Will put up PR for stable.

PR #1787, opened 2024-01-12
[SuperEditor] Fix unnecessary rebuilds and add a performance demo app (Resolves #1780)

### LINKED ISSUE #1780: Updating single document node rebuilds entire document
Branch with example: https://github.com/superlistapp/super_editor/compare/main...knopp:super_editor:single_node_rebuilds_document

https://github.com/superlistapp/super_editor/assets/96958/49d8ebc0-c012-4b27-b1a4-534a38a92a57

cc @matthew-carroll

angelosilvestre:
[SuperEditor] Fix unnecessary rebuilds and add a performance demo app. Resolves #1780

Updating a single node is causing the whole document to rebuild:

https://github.com/superlistapp/super_editor/assets/96958/49d8ebc0-c012-4b27-b1a4-534a38a92a57

I was able to reproduce this issue with tasks. The reason is that the task view model includes the `setComplete` in the equality operator. We pass an anonymous function to this view model property, so it will never be equal to the previous `setComplete`, causing all the tasks to always rebuild.

This PR removes the `setComplete` from the equality operator and from the `hashCode` method. A test was included to ensure we aren't rebuild unmodified text nodes. We could expand this to other nodes by adding a build tracker to the other components.

This PR also introduces a new "example_perf" app for the performance demos. The UI was copied from the main example app. I also included a "Long Doc Demo", with the whole Frankenstein text split by nodes. This document was used a while ago to fix a performance issue described in https://github.com/superlistapp/super_editor/issues/751

Result:

https://github.com/superlistapp/super_editor/assets/7597082/8a03f074-673e-435e-99bd-5a249facc3a7


matthew-carroll:
@knopp can you please review the findings and the solution here and let us know if it covers the Superlist scenarios?

super_editor/test/super_editor/supereditor_text_layout_test.dart
```diff
@@ -1,6 +1,17 @@
...
 import 'package:flutter_test/flutter_test.dart';
 import 'package:flutter_test_runners/flutter_test_runners.dart';
+import 'package:super_editor/src/core/document.dart';
+import 'package:super_editor/src/core/document_composer.dart';
+import 'package:super_editor/src/core/editor.dart';
+import 'package:super_editor/src/default_editor/default_document_editor.dart';
+import 'package:super_editor/src/default_editor/list_items.dart';
+import 'package:super_editor/src/default_editor/paragraph.dart';
+import 'package:super_editor/src/default_editor/super_editor.dart';
+import 'package:super_editor/src/default_editor/tasks.dart';
```
matthew-carroll, line 12: Ideally, in a test, you import the whole library rather than individual pieces. Are some of these things hidden?

angelosilvestre: Nope, we can import the whole library. Updated.

matthew-carroll:
@angelosilvestre it looks like the new tests are failing

angelosilvestre:
@matthew-carroll The tests were failing due to the changes made in https://github.com/superlistapp/super_editor/pull/1766 where the list items components were changed to use proxy components.

In that PR, the list item components where kept stateless and a `GlobalKey` to the inner text component was created inside the widget. This doesn't make sense, because each time the widget is rebuild a new key instance is created. This was causing the whole state to be discarded, because the keys would never be equal. So, a new `SuperText` was being created at each widget rebuild.

Updated this PR to change the list items to be statefull.

matthew-carroll (approved):
LGTM

PR #1797, opened 2024-01-13
[SuperEditor] Remove the editor scrollbar when configured in the ScrollBehavior  (Resolves #1786)

### LINKED ISSUE #1786: ScrollBar usage
Hello,

I've integrated the super_editor package into my project using its latest stable version from GitHub. However, I'm encountering an issue with the default scrollbar. When the editor is used in landscape orientation mode, this scrollbar appears with its own padding. I've been trying to customize or even disable this default scrollbar, but without success.

Could you provide guidance on how to either customize or completely disable the default scrollbar in super_editor, especially in landscape orientation? What are your thoughts or suggestions on this matter?

Thank you.
IMG_0448

angelosilvestre:
[SuperEditor] Remove the editor scrollbar when configured in the ScrollBehavior.  Resolves #1786

When the editor isn't inside an ancestor scrollable we install our own scrollbar in the widget tree. However, some developers want to prevent this toolbar from being displayed.

This PR changes the editor to avoid installing the toolbar if the scroll behavior is configure to NOT display scrollbars. Developers will be able to hide the scrollbar with the following code:

```dart
ScrollConfiguration(
 behavior: ScrollConfiguration.of(context).copyWith(scrollbars: false),
 child: SuperEditor(...),
)
```

Flutter doesn't provide a way to query the `scrollbars` property, so I'm using a hack to determine if the app wants the scrollbar. I left a comment to change that as soon as we can.

### DIFF
--- a/super_editor/lib/src/default_editor/document_scrollable.dart
+++ b/super_editor/lib/src/default_editor/document_scrollable.dart
@@ -1,8 +1,8 @@
 import 'dart:math';
 import 'dart:ui';
 
+import 'package:flutter/material.dart';
 import 'package:flutter/scheduler.dart';
-import 'package:flutter/widgets.dart';
 import 'package:super_editor/src/infrastructure/_logging.dart';
 import 'package:super_editor/src/infrastructure/documents/document_scroller.dart';
 import 'package:super_editor/src/infrastructure/flutter/build_context.dart';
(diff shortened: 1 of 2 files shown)


matthew-carroll:
Please include a reference to the ticket that you filed with Flutter in the PR description.

super_editor/lib/src/default_editor/document_scrollable.dart
```diff
@@ -198,6 +198,41 @@ class _DocumentScrollableState extends State<DocumentScrollable> with SingleTick
...
+    // or not the buildScrollbar method returns a ScrollBar. If it doesn't,
+    // this means the app doesn't want us to add our own ScrollBar.
+    //
+    // Change this after https://github.com/flutter/flutter/issues/141508 is solved.
+    final maybeScrollBar = behavior.buildScrollbar(
+      context,
+      child,
+      ScrollableDetails.vertical(controller: _scrollController),
+    );
+    if (maybeScrollBar is! Scrollbar) {
```
matthew-carroll, line 231: Should this be a null check instead? Seems overly specific to check for a `Scrollbar` instance.

super_editor/lib/src/default_editor/document_scrollable.dart
```diff
@@ -198,6 +198,41 @@ class _DocumentScrollableState extends State<DocumentScrollable> with SingleTick
...
+        child: SingleChildScrollView(
+          controller: _scrollController,
+          physics: const NeverScrollableScrollPhysics(),
+          child: child,
+        ),
+      ),
+    );
+  }
+
+  Widget _buildScrollbar({
```
matthew-carroll, line 214: Should this be called `_maybeBuildScrollbar()` to borrow the existing semantics?

angelosilvestre:
@matthew-carroll Updated the PR description

super_editor/test/super_editor/supereditor_scrolling_test.dart
```diff
@@ -1180,6 +1181,60 @@ void main() {
...
+      expect(
+        find.descendant(
+          of: find.byType(SuperEditor),
+          matching: find.byType(ScrollbarWithCustomPhysics),
+        ),
+        findsOneWidget,
+      );
+    });
+
+    testWidgetsOnDesktop('does not show scrollbar when configured to', (tester) async {
```
matthew-carroll, line 1207: Please replace the name of the test with: "does not show scrollbar when ancestor ScrollConfiguration does not want one"

super_editor/lib/src/default_editor/document_scrollable.dart
angelosilvestre: We need to check for a `Scrollbar`. When configured to not display scrollbars, `buildScrollbar` returns the given `child`.

matthew-carroll: If that's the case, should we ask `if (maybeScrollBar == child)`? I'm concerned that the specific widget type used for the scrollbar could change and break this.

angelosilvestre: Ok. That will work. Updated.

matthew-carroll (approved):
LGTM

PR #1803, opened 2024-01-15
Fix hint padding and auto-scrolling for Android and iOS super text fields (Resolves #1776)

### LINKED ISSUE #1776: [SuperTextField][Mobile] - Bugs related to single-line scrolling and caret placement
The following video demonstrates a number of issues with a single-line mobile text field. This video is for iOS, but the issues probably exist on Android, too.

https://github.com/superlistapp/super_editor/assets/7259036/7419cf9e-3262-4400-8dcd-5441d51dde35

* The text field almost auto-scrolls correctly, but the caret is able to sit just beyond the text field boundary.
* Once the text field scrolls, single tapping doesn't place the caret.
* Double tapping selects a word, but dragging either handle glitches in an unusable manner.
* Padding isn't respected by the hint text

Screenshot 2024-01-09 at 3 03 54 PM

matthew-carroll:
Fix hint padding and auto-scrolling for Android and iOS super text fields (Resolves #1776)

This PR started as a fix for hint padding and some auto-scroll issues in mobile `SuperTextField`s, especially multiline fields which were impacted by earlier work.

What I've found in this process is that our implementation of mobile selection and scrolling is too complicated and intermingled. I spent a huge amount of time debugging this and trying to get to a workable place. Therefore, the goal of this PR is to get us to a workable place, even if it's not perfect.

After we get back to a mostly-working place, we need to come back and refactor `SuperTextField` to reduce the complexity and coupling. We should also re-evaluate some feature decisions, such as per-line scrolling, to see if we can get away with implementing a simpler feature instead, like pixel-based scrolling.


super_editor/test/super_textfield/super_textfield_inspector.dart
```diff
@@ -267,6 +311,40 @@ class SuperTextFieldInspector {
...
+    final caretDisplayElement = find
+        .descendant(of: androidTextField, matching: find.byType(TextLayoutCaret))
+        .evaluate()
+        .single as StatefulElement;
+    final caretDisplay = caretDisplayElement.state as TextLayoutCaretState;
+    final caretGlobalRect = caretDisplay.globalCaretGeometry!;
+
+    final viewportOffset = viewport.localToGlobal(Offset.zero);
+    return caretGlobalRect.translate(-viewportOffset.dx, -viewportOffset.dy);
+  }
```
angelosilvestre, line 346: These two methods look identical. Should we use a single method for Android and iOS?

angelosilvestre (approved):
LGTM with a minor comment.

PR #1819, opened 2024-01-24
[SuperEditor][Demo] Implement formatting toolbar for the Docs Demo (Resolves #1783)

### LINKED ISSUE #1783: [SuperEditor][Demo] - Add formatting toolbar to Docs example
Recently we introduced an example app that mimics Google Docs. When it was introduced, only the minimum behavior was added to reproduce a focus issue that needed to be solved.

For this ticket, fill out the formatting toolbar as much as possible. The formatting toolbar is the permanent toolbar that appears beneath the app menu and the editor, and includes controls for things like text alignment, font size, etc.

Make the bar look as close as possible to the real Google Docs app. Implement every behavior that should currently be supported by `SuperEditor`.

Make note of any missing features, bugs, or aspects of `SuperEditor` that make this effort more difficult than it should be.

angelosilvestre:
[SuperEditor][Demo] Implement formatting toolbar for the Docs Demo. Resolves #1783

Implements the following behaviors:

- Change zoom
- Change text types
- Change font family
- Change font size
- Toggle bold, italics and underline
- Change text color
- Change text background color
- Apply link
- Change alignment
- Convert to task
- Convert to unordered list
- Convert to ordered list
- Clicking on the "Menus" button shows a fixed list without search.
- The buttons that aren't implemented show an alert

When resizing, the toolbar hides groups of buttons and then displays a button to reveal the hidden items. Upon tapping this button, a popover toolbar is displayed with the remaining items.

Some remarks:

- It seems we can't query the fonts available in the system. We could do that with platform channels, but this is outside of the scope of this demo. I decided to show some arbitrary fonts from the Google fonts package.
- I used a combination of `AnimatedOpacity` and `AnimatedSize` to achieve a similar effect when the user resizes the screen and the groups are hidden. If the user quickly resizes this could show the overflow indicator for some frames.
- The link popover was extracted from our regular editor demo
- Some options of the popovers aren't implemented, like the "Custom Colors" option.

A problem I've found:

Placing the caret at the middle of a text with a different font family, size or color doesn't apply those attributions to the composer preferences. We can configure the composer preferences reaction with the attributions we want to extend, but it only works by equality, we can't tell the reaction we want to keep a `FontFamilyAttribution`, regardless of the configured font family.

Demo:

https://github.com/superlistapp/super_editor/assets/7597082/8b5e8273-c4d5-4ea0-9713-b80fc5cb99ad


super_editor/example_docs/lib/editor.dart
```diff
@@ -66,10 +81,38 @@ class _DocsEditorState extends State<DocsEditor> {
...
+          componentBuilders: [
+            ...defaultComponentBuilders,
+            TaskComponentBuilder(widget.editor),
+          ],
         );
       },
     );
   }
+
+  TextStyle _applyStyles(Set<Attribution> attributions, TextStyle existingStyle) {
```
matthew-carroll, line 93: Can we name this something that more clearly describes its purpose as an inline text styler? Or perhaps default document text styler? Or something that's a bit more specific?

super_editor/example_docs/lib/editor.dart
```diff
@@ -81,31 +124,101 @@ final _unfocusedEditorSelectionStyle = SelectionStyles(
...
-      ),
-    ],
-  );
+/// Attribution to be used within [AttributedText] to
+/// represent an inline span of a backgrounnd color change.
+///
+/// Every [BackgroundColorAttribution] is considered equivalent so
+/// that [AttributedText] prevents multiple [BackgroundColorAttribution]s
+/// from overlapping.
+class BackgroundColorAttribution implements Attribution {
```
matthew-carroll, line 133: Doesn't this already exist?

super_editor/example_docs/lib/editor.dart
```diff
@@ -81,31 +124,101 @@ final _unfocusedEditorSelectionStyle = SelectionStyles(
...
+  }
+}
+
+/// Attribution to be used within [AttributedText] to
+/// represent an inline span of a font family change.
+///
+/// Every [FontFamilyAttribution] is considered equivalent so
+/// that [AttributedText] prevents multiple [FontFamilyAttribution]s
+/// from overlapping.
+class FontFamilyAttribution implements Attribution {
```
matthew-carroll, line 166: We might as well add these attributions to `super_editor`, right?

super_editor/example_docs/lib/infrastructure/color_selector.dart
```diff
@@ -0,0 +1,371 @@
+import 'package:example_docs/infrastructure/selectable_grid.dart';
+import 'package:example_docs/theme.dart';
+import 'package:flutter/material.dart';
+import 'package:super_editor/super_editor.dart';
+
+/// A selection control, which displays a button with the selected color, and upon tap, displays a
+/// popover grid of available colors, from which the user can select a different color.
+///
+/// Unlike Flutter `DropdownButton`, which displays the popover list in a separate route,
```
matthew-carroll, line 9: Do all these docs belong here? Everything below the first sentence seems like it's written for some lower level scaffold or something...

super_editor/example_docs/lib/infrastructure/color_selector.dart
```diff
@@ -0,0 +1,371 @@
...
+    super.key,
+    this.parentFocusNode,
+    this.boundaryKey,
+    this.color,
+    this.items = defaultColors,
+    required this.onSelected,
+    required this.buttonBuilder,
+    this.columnCount = 10,
+    this.showClearButton = false,
+    this.tapRegionGroupId,
```
matthew-carroll, line 44: `tapRegionGroupId` probably makes more sense just below `parentFocusNode` because the two are related.

super_editor/example_docs/lib/infrastructure/color_selector.dart
```diff
@@ -0,0 +1,371 @@
...
+///   * Pressing DOWN with the last item active moves the active item selection to the first item.
+///   * Pressing ENTER selects the currently active item and closes the popover list.
+class ColorSelector extends StatefulWidget {
+  const ColorSelector({
+    super.key,
+    this.parentFocusNode,
+    this.boundaryKey,
+    this.color,
+    this.items = defaultColors,
+    required this.onSelected,
```
matthew-carroll, line 40: callbacks should typically be at the bottom of the property list.

super_editor/example_docs/lib/infrastructure/color_selector.dart
```diff
@@ -0,0 +1,371 @@
...
+///   * Pressing LEFT/RIGHT moves the "active" item selection left/right.
+///   * Pressing UP with the first item active moves the active item selection to the last item.
+///   * Pressing DOWN with the last item active moves the active item selection to the first item.
+///   * Pressing ENTER selects the currently active item and closes the popover list.
+class ColorSelector extends StatefulWidget {
+  const ColorSelector({
+    super.key,
+    this.parentFocusNode,
+    this.boundaryKey,
+    this.color,
```
matthew-carroll, line 38: Should this be called `selectedColor`?

super_editor/example_docs/lib/infrastructure/color_selector.dart
```diff
@@ -0,0 +1,371 @@
...
+    this.items = defaultColors,
+    required this.onSelected,
+    required this.buttonBuilder,
+    this.columnCount = 10,
+    this.showClearButton = false,
+    this.tapRegionGroupId,
+  });
+
+  /// The [FocusNode], to which the popover grid's [FocusNode] will be added as a child.
```
matthew-carroll, line 47: The term "popover grid" seems to appear in here a lot, but this widget is a "color selector"...

super_editor/example_docs/lib/infrastructure/color_selector.dart
```diff
@@ -0,0 +1,371 @@
...
+
+  /// The currently selected value or `null` if no item is selected.
+  ///
+  /// This value is used to build the button.
+  final Color? color;
+
+  /// The items that will be displayed in the popover grid.
+  ///
+  /// For each item, its color is displayed as a circle.
+  final List<Color> items;
```
matthew-carroll, line 82: Should this be called `colors`? `items` seems unnecessarily generic...

super_editor/example_docs/lib/infrastructure/color_selector.dart
```diff
@@ -0,0 +1,371 @@
...
+
+  /// Called when the user selects an item on the popover grid.
+  final void Function(Color? value) onSelected;
+
+  /// Builds the button of this [ColorSelector].
+  final Widget Function(BuildContext context, Color? selectedColor) buttonBuilder;
```
matthew-carroll, line 88: Should this be a `colorBuilder` or perhaps a `colorButtonBuilder`?

super_editor/example_docs/lib/infrastructure/color_selector.dart
```diff
@@ -0,0 +1,371 @@
...
+
+  void _onItemSelected(Color? value) {
+    _popoverController.close();
+    widget.onSelected(value);
+  }
+
+  /// Decides a font color for a [background] color based on the brightness of the [background].
+  ///
+  /// Returns [Colors.white] if [background] is a dark color and [Colors.black] otherwise.
+  Color _getFontColorForBackground(Color background) {
```
matthew-carroll, line 126: I'm curious about the intended purpose of this. It looks like it's selecting a font color based on the background - but doesn't all the text have an explicit color in a document editor? What's the situation where we'd want to dynamically alter the font color?

super_editor/example_docs/lib/infrastructure/color_selector.dart
```diff
@@ -0,0 +1,371 @@
...
+        clipBehavior: Clip.hardEdge,
+        color: Colors.white,
+        child: Padding(
+          padding: const EdgeInsets.all(8.0),
+          child: Column(
+            mainAxisSize: MainAxisSize.min,
+            crossAxisAlignment: CrossAxisAlignment.start,
+            children: [
+              if (widget.showClearButton) //
+                SizedBox(
```
matthew-carroll, line 153: This looks like a very long and complicated widget tree that could be decomposed into pieces that are easier to read.

super_editor/example_docs/lib/infrastructure/icon_selector.dart
```diff
@@ -0,0 +1,191 @@
+import 'package:example_docs/theme.dart';
+import 'package:flutter/material.dart';
+import 'package:super_editor/super_editor.dart';
+
+/// A selection control, which displays a button with the selected item, and upon tap, displays a
+/// popover list of available icons, from which the user can select a different option.
+///
+/// Unlike Flutter `DropdownButton`, which displays the popover list in a separate route,
```
matthew-carroll, line 8: I think some of the comments from the color selector apply here as well. I'd like to see Dart Docs that are more specific to the widget they're documenting, and I'd like to see terminology that's specific to the goal of the widget that refers to those terms in docs, property names, etc.

super_editor/example_docs/lib/infrastructure/increment_decrement_field.dart
```diff
@@ -0,0 +1,142 @@
+import 'dart:math';
+
+import 'package:example_docs/theme.dart';
+import 'package:flutter/material.dart';
+import 'package:super_editor/super_editor.dart';
+
+/// A textfield that contains buttons to increment and decrement a [value].
```
matthew-carroll, line 7: It's probably not true (or shouldn't be true) that "a textfield that contains buttons..."

This might be a widget that happens to contain a textfield AND buttons, but textfields don't contain buttons.

Moreover, when documenting a widget, the most important thing isn't what's inside the widget - the most important part of what the widget does for the user. If needed, you can provide additional information about what's in the widget below the first sentence.

super_editor/example_docs/lib/infrastructure/increment_decrement_field.dart
```diff
@@ -0,0 +1,142 @@
...
+  const IncrementDecrementField({
+    super.key,
+    required this.value,
+    required this.onChange,
+  });
+
+  /// The current value.
+  final int value;
+
+  /// Called whenever the value changes.
```
matthew-carroll, line 20: Let's say something that's more specific and useful. What does it mean for the value change? Did the user tap a button? Did the user type some text?

If the user directly edits the textfield, what happens when the user clears it out? Is that reported?

super_editor/example_docs/lib/infrastructure/increment_decrement_field.dart
```diff
@@ -0,0 +1,142 @@
...
+///
+/// Selects all the content on the textfield upon focus.
+class IncrementDecrementField extends StatefulWidget {
+  const IncrementDecrementField({
+    super.key,
+    required this.value,
+    required this.onChange,
+  });
+
+  /// The current value.
```
matthew-carroll, line 17: Is there a reason that we call this `value`? Seems very generic. Is this actually the `fontSize`? If this is meant to be a generic increment/decrement field, are we actually using it for multiple use-cases, or is it only being used for font size?

super_editor/example_docs/lib/infrastructure/selectable_grid.dart
```diff
@@ -0,0 +1,275 @@
...
+import 'package:flutter/services.dart';
+
+/// A grid where the user can navigate between its items and select one of them.
+///
+/// Includes the following keyboard selection behaviors:
+///
+///   * Pressing UP/DOWN moves the "active" item selection up/down.
+///   * Pressing LEFT/RIGHT moves the "active" item selection left/right.
+///   * Pressing ENTER selects the currently active item.
+class SelectableGrid<T> extends StatefulWidget {
```
matthew-carroll, line 11: Please name generics with full words so it's clear what they mean. For example, this one might be `GridItemType` or `ItemDataType`, etc.

super_editor/example_docs/lib/toolbar.dart
```diff
@@ -0,0 +1,1560 @@
...
+
+  final PopoverController _searchPopoverController = PopoverController();
+  final FocusNode _searchFocusNode = FocusNode();
+
+  TextItem _selectedZoom = const TextItem(id: '100', label: '100%');
+
+  /// Groups the aditional toolbar options popover, which is shown by tapping
+  /// the "more items" button with the popovers shown by the toolbar items,
+  /// like the color picker.
+  final _tapRegionGroupId = 'docs_toolbar';
```
matthew-carroll, line 56: Should this be a static const?

super_editor/example_docs/lib/toolbar.dart
```diff
@@ -0,0 +1,1560 @@
...
+    setState(() {});
+  }
+
+  void _onUrlFieldPerformAction(TextInputAction action) {
+    if (action == TextInputAction.done) {
+      _applyLink();
+    }
+  }
+
+  void _changeZoomLevel(TextItem? zoomLevel) {
```
matthew-carroll, line 99: A number of these methods read like commands, e.g., `_changeZoomLevel()` - I want to confirm that these are commands, and not user interaction reactions?

For example, if this is a reaction, then a more appropriate name might be `_onChangeZoomLevelPressed` or `_onChangeZoomLevelRequested` or something that indicates this is a reaction/response rather than a command.

super_editor/lib/src/default_editor/text.dart
```diff
@@ -238,6 +238,123 @@ extension DocumentSelectionWithText on Document {
 
     return false;
   }
+
+  Set<Attribution> getAllAttributions(DocumentSelection selection) {
```
matthew-carroll, line 242: Dart Doc? Tests?

super_editor/lib/src/infrastructure/popover_scaffold.dart
```diff
@@ -80,6 +81,12 @@ class PopoverScaffold extends StatefulWidget {
   /// If `null`, tapping outside closes the popover.
   final void Function(PopoverController) onTapOutside;
 
+  /// A group ID for a tap region that is shared with the popover.
+  ///
+  /// Tapping on a [TapRegion] with the same [tapRegionGroupId]
+  /// won't invoke [onTapOutside].
+  final String? tapRegionGroupId;
```
matthew-carroll, line 88: Same comment about property placement.

super_editor/lib/src/infrastructure/popover_scaffold.dart
```diff
@@ -30,6 +30,7 @@ class PopoverScaffold extends StatefulWidget {
     this.parentFocusNode,
     this.boundaryKey,
     this.onTapOutside = _PopoverScaffoldState.closePopoverOnTapOutside,
+    this.tapRegionGroupId,
```
matthew-carroll, line 33: Should we have tests for this?

super_editor/lib/src/infrastructure/selectable_list.dart
```diff
@@ -58,6 +60,10 @@ class ItemSelectionList<T> extends StatefulWidget {
   /// The [FocusNode] of the list.
   final FocusNode? focusNode;
 
+  final Axis axis;
+
+  final IndexedWidgetBuilder? separatorBuilder;
```
matthew-carroll, line 65: Dart Docs? Also, please check your property order.

super_editor/lib/src/infrastructure/selectable_list.dart
```diff
@@ -258,6 +266,19 @@ class ItemSelectionListState<T> extends State<ItemSelectionList<T>> with SingleT
       ),
     );
   }
+
+  Widget _buildContainer({required List<Widget> children}) {
```
matthew-carroll, line 270: Calling this a `Container` has two issues. First, isn't not actually a `Container` widget, which is what some might expect. Second, it doesn't tell us anything about what it actually builds, forcing the user to read all the source code within the method to try to figure it out.

matthew-carroll (commented):
The UI is looking good in the video. I did an initial pass. Obviously there's a lot of code in here. The changes to `super_editor`, at a minimum, need corresponding tests. More broadly, we can probably add tests for the Docs app in a followup PR, but we definitely should add tests for it so that we can also confirm that app developers are able to test their use of `SuperEditor`, too.

super_editor/example_docs/lib/editor.dart
angelosilvestre: We have an attribution for the font color. I didn't find one for the background color.

super_editor/example_docs/lib/editor.dart
angelosilvestre: The font family attribution has some gotchas. If the user is using the google_fonts package, it needs to create a `TextStyle` using `GoogleFonts.getFont('my font', textStyle: myOtherStyle)` instead of just setting the font family.

super_editor/example_docs/lib/infrastructure/color_selector.dart
angelosilvestre: This isn't related to the editor colors.

The color picker displays a "check" icon over the selected color. I'm using this method to choose whether the icon color should be white or black. If the selected color is a dark color, the icon is white, otherwise it is black.

Maybe calling it _getColorForCheckIcon would make more sense.

super_editor/example_docs/lib/infrastructure/color_selector.dart
angelosilvestre: Yeah, we should be fine getting rid of most of it.

super_editor/example_docs/lib/infrastructure/color_selector.dart
angelosilvestre: Updated.

super_editor/example_docs/lib/infrastructure/icon_selector.dart
angelosilvestre: Should I include the keyboard interactions in this dart doc? Also, should we include the same docs for `parentFocusNode` and `boundaryKey` that we have in other widgets?

super_editor/example_docs/lib/infrastructure/increment_decrement_field.dart
angelosilvestre: Currently we are using only for the font size. But we could use it for any integer value that we want. Should we change it to be specific to font sizes?

super_editor/example_docs/lib/toolbar.dart
angelosilvestre: Most of then are user interaction reactions. Updated the names.

super_editor/lib/src/infrastructure/selectable_list.dart
angelosilvestre: Updated.

super_editor/lib/src/infrastructure/popover_scaffold.dart
angelosilvestre: Test added.

super_editor/example_docs/lib/editor.dart
matthew-carroll: Can you frame your response related to the question? Are you implying that we should keep this specific to the demo? Are you suggesting additional work to figure out how to make it work in general? Should there be multiple versions of such an attribution? What's the suggested action?

super_editor/example_docs/lib/infrastructure/icon_selector.dart
matthew-carroll: Documenting keyboard actions is probably a good idea, but make sure they're described in terms of the widget they're documenting. If you're documenting navigation around a grid of colors, then describe the selection and movement in terms of colors, rather than arbitrary child widgets, etc.

For the focus node and boundary key, I think we can include per-property docs, but I doubt that we need to write a bunch about them in the class doc. The property docs might also want to link to the inner scaffold properties as a breadcrumb.

super_editor/example_docs/lib/editor.dart
matthew-carroll: Spoke offline - we'll leave these just in the demo

super_editor/lib/src/default_editor/text.dart
angelosilvestre: Updated.

super_editor/example_docs/lib/infrastructure/icon_selector.dart
angelosilvestre: Updated.

angelosilvestre:
@matthew-carroll When I was writing the test for `getAttributionsByType` I found a bug in `AttributedText.getAllAttributionsThroughout`. When we check each index for the attributions, we are not passing the current attribution as argument, so it was looking for any attributions. I included a fix and tests for that.

angelosilvestre:
@matthew-carroll It seems the goldens need update. I re-generated them.

attributed_text/test/attributed_text_test.dart
```diff
@@ -428,6 +428,42 @@ void main() {
             .getAttributedRange({ExpectedSpans.bold, ExpectedSpans.italics, ExpectedSpans.strikethrough}, 2);
         expect(range, const SpanRange(1, 3));
       });
+
+      group('getAllAttributionsThroughout', () {
```
matthew-carroll, line 432: Should this still be here? We merged this fix in separately, didn't we?

angelosilvestre: Removed from this PR.

angelosilvestre:
Hi @alterhuman. This PR is waiting for a review from @matthew-carroll. He will review it when he finds some time.

super_editor/lib/src/default_editor/text.dart
```diff
@@ -238,6 +238,125 @@ extension DocumentSelectionWithText on Document {
 
     return false;
   }
+
+  /// Returns all attributions that appear throughout the entirety of the selected range.
```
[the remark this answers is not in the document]
rutvik110: yes, it could certainly replace some code I've for getting attributions that are applied across entire user selection.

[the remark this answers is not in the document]
rutvik110: @angelosilvestre I find the naming for this method a bit confusing. Shouldn't this be named something like `getAllAttributionsPresentThroughoutSelection/`getAllAttributionsAppliedThroughoutSelection``? based on its purpose.

super_editor/example_docs/lib/editor.dart
```diff
@@ -66,10 +81,61 @@ class _DocsEditorState extends State<DocsEditor> {
               displayCaretWithExpandedSelection: false,
             ),
           ],
+          componentBuilders: [
+            ...defaultComponentBuilders,
+            TaskComponentBuilder(widget.editor),
```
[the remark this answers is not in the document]
angelosilvestre: Yeah, it does support.

super_editor/example_docs/lib/infrastructure/text_item_selector.dart
```diff
@@ -0,0 +1,267 @@
+import 'package:example_docs/theme.dart';
+import 'package:flutter/material.dart';
+import 'package:super_editor/super_editor.dart';
+
+/// A selection control, which displays a button with the selected item, and upon tap, displays a
+/// popover list of available text options, from which the user can select a different
+/// option.
+///
+/// Unlike Flutter `DropdownButton`, which displays the popover list in a separate route,
```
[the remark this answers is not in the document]
angelosilvestre: Updated.

super_editor/example_docs/lib/toolbar.dart
```diff
@@ -0,0 +1,1560 @@
...
+          .preferences
+          .currentAttributions;
+    }
+
+    return widget.document.getAllAttributions(selection);
+  }
+
+  TextStyle _getDefaultTextStyleForBlockType(TextItem item) {
+    return switch (item.id) {
+      'header1' => const TextStyle(
```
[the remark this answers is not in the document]
angelosilvestre: Updated.

super_editor/lib/src/default_editor/text.dart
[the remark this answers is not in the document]
angelosilvestre: @rutvik110 This method only returns the attributions that are applied throughout the entire selection.

super_editor/lib/src/default_editor/attributions.dart
```diff
@@ -73,6 +73,72 @@ class ColorAttribution implements Attribution {
   }
 }
 
+/// Attribution to be used within [AttributedText] to
+/// represent an inline span of a backgrounnd color change.
+///
+/// Every [BackgroundColorAttribution] is considered equivalent so
```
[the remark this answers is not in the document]
angelosilvestre: I based this attribution on the `ColorAttribution`, but it seems that the implementation doesn't match what the docs describe. The `id` should be fixed for us to avoid applying two different colors to the same range.

I updated the implementation and added tests.

super_editor/test/super_editor/supereditor_attributions_test.dart
```diff
@@ -289,5 +289,110 @@ void main() {
...
+        () => editor.execute([
+          AddTextAttributionsRequest(
+            documentRange: const DocumentRange(
+              start: DocumentPosition(nodeId: '1', nodePosition: TextNodePosition(offset: 1)),
+              end: DocumentPosition(nodeId: '1', nodePosition: TextNodePosition(offset: 3)),
+            ),
+            attributions: {const ColorAttribution(Colors.red)},
+          ),
+        ]),
+        throwsException,
```
matthew-carroll, line 324: Throwing an exception seems like a very strong reaction to trying to apply a red color over a yellow color. What prevents this exception from blowing up the editor if the user happens to partially highlight some yellow text and change the color to red?

matthew-carroll (approved):
LGTM - I left one question about an exception that's being verified in tests. We can continue that conversation after this PR.

PR #1829, opened 2024-02-02
Remove the remaining reference to the text painter rounding migration flag.

LongCatIsLooong:
The default value of `disableRoundingHack` is now true, the migration flag is no longer needed and will be removed soon.

### DIFF
--- a/super_text_layout/test/super_text_test.dart
+++ b/super_text_layout/test/super_text_test.dart
@@ -1,5 +1,3 @@
-import 'dart:ui' show ParagraphBuilder;
-
 import 'package:flutter/material.dart';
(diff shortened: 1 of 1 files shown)


matthew-carroll (approved):
LGTM

matthew-carroll:
@LongCatIsLooong is this change on `stable` too? Or only on `master`?

LongCatIsLooong:
It was removed in [3.15.0-1.0.pre](https://github.com/flutter/flutter/releases/tag/3.15.0-1.0.pre) so yeah the default should be the same (no rounding) for both branches.

matthew-carroll:

PR #1837, opened 2024-02-08
Cherry-Pick: [SuperTextField][Android][iOS] Disables text field caret blink while dragging caret (Resolves #1705) (#1778)

### LINKED ISSUE #1705: [SuperTextField][Android][iOS] - Caret blinks during a drag

rutvik110:
[SuperTextField][Android][iOS] Disables text field caret blink while dragging caret (Resolves #1705)


matthew-carroll:
This was already cherry picked: https://github.com/superlistapp/super_editor/pull/1835

PR #1859, opened 2024-02-23
Changed FillWidthIfConstrained to use ancestor width constraints instead of ancestor size (Resolves #1857)

### LINKED ISSUE #1857: [SuperTextField] - Single line field has phantom scrollbar when the width changes during mouse hover
Consider a single-line text field. When the user hovers the mouse over that text field, a small "x" appears at the right side of the text field to clear the text field. This "x" doesn't sit atop the text field. Instead, the "x" pushes the width of the actual `SuperTextField` towards the left, making the text field about 20px more narrow.

Expected Result: When the "x" appears, the text field continues NOT to display a scrollbar, or have any scrollable content, because the text still doesn't overflow the available width.

Actual Result: When the "x" appears, the `SuperTextField` text remains as wide as it was before the "x" appeared, resulting in 20px of horizontal scrollable distance, along with the appearance of a horizontal scrollbar.

Likely culprit: The width of the text in a `SuperTextField` is carefully calculated based on multiple possible text field layouts. For single-line text fields, this calculation depends upon the viewport width of the ancestor scrollable. My guess is, when the ancestor scrollable becomes more narrow, it doesn't invalidate its content, because the content within a scrollable isn't based on the viewport dimensions. But in our case, the scrollable content IS sized based on the viewport dimensions, hence the bug.

matthew-carroll:
Changed FillWidthIfConstrained to use ancestor width constraints instead of ancestor size (Resolves #1857)

The reported use-cases problem is as follows: A single-line text field has some text that doesn't exceed the available space. The user hovers over the text field, and the app adds an "x" icon to the right side, slightly shrinking with the width of the text field. The textfield shows a horizontal scrollbar even though the text still doesn't exceed the available space.

We have a special widget called `FillWidthIfConstrained` that's used to set the width of the text within a `SuperTextField` based, in part, on the width of an ancestor viewport. The existing widget behavior worked correctly for an arbitrary frame. However, when the ancestor viewport changes width, the existing approach didn't work because the scrollable viewport doesn't update its width until after its children are done with layout. And `FillWidthIfConstrained` IS one of those children.

The most reasonable workaround I could come up with is to switch from checking `scrollable.size.width` to `scrollable.constraints.maxWidth`. This is because constraints are set before layout, whereas size is set after layout. So now `FillWidthIfConstrained` correctly identifies the ancestor scrollable width during its own layout phase.


angelosilvestre (approved):
LGTM

PR #1860, opened 2024-02-24
[SuperEditor][Web] Fix cmd + left to move selection (Resolves #1852)

### LINKED ISSUE #1852: [SuperEditor] [Web] CMD+Left-arrow does not move
**Package Version**
`main` branch with commit hash 94d8fec5fabbc8ae3a8023769fb98d80a4a2606e

**To Reproduce**
1. On web, go to SuperEditor demo app
2. Place the selection in the middle of a text node
3. Press CMD-left-arrow

**Actual behavior**
Nothing happens, i.e. the selection remains the same.

**Expected behavior**
Selection moves to the start of the node.

**Platform**
Web

**Flutter version**

**Screenshots**

https://github.com/superlistapp/super_editor/assets/31278849/adca2187-f45e-4a7a-ab2b-e4ea23006571

angelosilvestre:
[SuperEditor][Web] Fix cmd + left to move selection (Resolves #1852)

On web, pressing CMD + LEFT ARROW isn't changing the selection. The selection movement on web is done using the non-text deltas that the IME generates instead of handling key presses. This causes an issue when the user tries to move the selection to the first character, because the first character the IME sees sits in the invisible region. For example, for the paragraph with the content "This is the first line", the IME content looks like this:

```
. This is the first line
```

So, when the IME moves the selection to the beginning of the line, the selection sits before the `.`. Because of that, we can't map the IME position to a document position and the selection remains unchanged.

This PR changes how we handle non-text deltas on web when the a collapsed selection or composing region sits in the invisible area. Now, in those cases we move the selection/composing region to the first visible character.

### DIFF
--- a/super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
+++ b/super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
@@ -14,6 +14,7 @@ import 'package:super_editor/src/default_editor/selection_upstream_downstream.da
 import 'package:super_editor/src/default_editor/tasks.dart';
 import 'package:super_editor/src/default_editor/text.dart';
 import 'package:super_editor/src/infrastructure/_logging.dart';
+import 'package:super_editor/src/infrastructure/platforms/platform.dart';
 
 import 'document_serialization.dart';
 
@@ -97,9 +98,29 @@ class TextDeltasDocumentEditor {
     // the IME composing region.
     editorImeLog.fine("After applying all deltas, converting the final composing region to a document range.");
     editorImeLog.fine("Raw IME delta composing region: ${textEditingDeltas.last.composing}");
+
+    final lastDelta = textEditingDeltas.last;
+
+    DocumentRange? docComposingRegion;
+    if (CurrentPlatform.isWeb &&
+        lastDelta.composing.isCollapsed &&
+        _serializedDoc.isPositionInsidePlaceholder(TextPosition(offset: lastDelta.composing.end))) {
+      // On web, pressing CMD + LEFT ARROW generates a non-text delta moving
+      // the selection, and possibly the composing region to the first character. However, the first character
+      // is in a region invisible to the user. Adjust the document composing region to be the first visible character.
+      // Expanded regions are already adjusted by the serializer.
+      docComposingRegion = _serializedDoc.imeToDocumentRange(
+        TextRange.collapsed(
+          _serializedDoc.firstVisiblePosition().offset,
+        ),
+      );
+    } else {
+      docComposingRegion = _serializedDoc.imeToDocumentRange(lastDelta.composing);
+    }
+
     editor.execute([
       ChangeComposingRegionRequest(
-        _serializedDoc.imeToDocumentRange(textEditingDeltas.last.composing),
+        docComposingRegion,
       ),
     ]);
     editorImeLog.fine("Document composing region: ${composingRegion.value}");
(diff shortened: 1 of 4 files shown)


angelosilvestre:
@matthew-carroll  We should merge https://github.com/superlistapp/super_editor/pull/1859 first because of the updated goldens.

matthew-carroll:
@angelosilvestre are you saying that our keyboard handler doesn't run on web for CMD+LEFT? If not, how about any other such command, like CMD+RIGHT, SHIFT+ALT+RIGHT? Is Flutter not reporting those key presses on web?

angelosilvestre:
@matthew-carroll 

> are you saying that our keyboard handler doesn't run on web for CMD+LEFT? 

Yes, most of the caret movement key handlers don't run on web. Flutter reports them, but we ignore the key presses and handle the non-text deltas. This change was made in a previous PR, because caret movement doesn't work reliably on web if we handle those key presses.

The only affected handlers are those that move the caret to the beginning of the text. As all of them will generate the same deltas, they are all covered by this PR.

matthew-carroll:
@angelosilvestre is this ready for review?

angelosilvestre:
@matthew-carroll Yeah, it's ready for review.

super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
```diff
@@ -97,9 +98,29 @@ class TextDeltasDocumentEditor {
     // the IME composing region.
     editorImeLog.fine("After applying all deltas, converting the final composing region to a document range.");
     editorImeLog.fine("Raw IME delta composing region: ${textEditingDeltas.last.composing}");
+
+    final lastDelta = textEditingDeltas.last;
```
matthew-carroll, line 102: There are two maintenance and readability concerns here. First, we're dumping a bunch of nuanced lines of a code in the middle of the high level processing function. Notice how the existing code around this is all higher level and more abstract. When reading a loop that applies deltas, it's difficult to come down here and have any context for what this bundle of code means.

Second, it seems that this exact code, or very similar code, is replicated down below. That creates an opportunity to change one version and not the other, leading to a bug.

Please extract this behavior into a method called:

```
DocumentRange? _calculateNewComposingRegion(List<TextEditingDelta> deltas) {}
```

super_editor/lib/src/default_editor/document_ime/document_serialization.dart
```diff
@@ -243,15 +243,26 @@ class DocumentImeSerializer {
     return true;
   }
 
+  /// Returns the first visible position in the IME content.
+  ///
+  /// If a placeholder is prepended, returns the first position after the placeholder,
+  /// otherwise, returns the first position.
+  TextPosition firstVisiblePosition() {
```
matthew-carroll, line 250: Should this be a getter?

super_editor/lib/src/default_editor/document_ime/document_serialization.dart
```diff
@@ -243,15 +243,26 @@ class DocumentImeSerializer {
...
+    return didPrependPlaceholder //
+        ? TextPosition(offset: _prependedPlaceholder.length)
+        : const TextPosition(offset: 0);
+  }
+
   DocumentPosition _imeToDocumentPosition(TextPosition imePosition, {required bool isUpstream}) {
+    final affinity = isUpstream ? TextAffinity.upstream : TextAffinity.downstream;
```
matthew-carroll, line 257: Why was this needed? If it's needed, why wasn't the absence of this breaking any tests?

super_editor/test/super_editor/supereditor_keyboard_test.dart
```diff
@@ -284,6 +284,42 @@ void main() {
       });
     });
 
+    group('on web', () {
+      group('moves caret', () {
```
matthew-carroll, line 288: Is there a reason this group is here?

super_editor/lib/src/default_editor/document_ime/document_serialization.dart
angelosilvestre: It can be.  Updated.

super_editor/test/super_editor/supereditor_keyboard_test.dart
angelosilvestre: Ungrouped. If we need other related tests in the future we can re-group.

super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
angelosilvestre: Updated. Also extracted the `DocumentSelection` calculation to be consistent with the composing region calculation.

super_editor/lib/src/default_editor/document_ime/document_serialization.dart
angelosilvestre: I can't remember if I needed this for something. Removed and nothing seems to break...

matthew-carroll (approved):
LGTM

PR #1868, opened 2024-03-01
Fix getAllAttributionsThroughout

angelosilvestre:
Fix getAllAttributionsThroughout

In https://github.com/superlistapp/super_editor/pull/1819 I found a bug on the getAllAttributionsThroughout method:

When we check each index for the attributions, we are not passing the current attribution as argument, so it was looking for any attributions.

This PR fixes the issue.


matthew-carroll (approved):
LGTM

Issue #1863, opened 2024-02-28
Editor crashes if first character typed is an emoji, when plugged with pattern, stable or action tags plugin


BazinC:
**Package Version**
GitHub, stable branch

**To Reproduce**
1. Run the example and open "In the Lab" Hash Tags, User Tags or Action Tags page
2. Type an emoji as your first character, e.g: 💙
4. Editor doesn't display the character and no longer accept any other user input
<details>
  <summary>See error:</summary>
  
  ```
════════ Exception caught by services library ══════════════════════════════════
The following _TypeError was thrown during method call TextInputClient.updateEditingStateWithDeltas:
Null check operator used on a null value

When the exception was thrown, this was the stack:
#0      TagFinder.findTagAroundPosition (package:super_editor/src/default_editor/text_tokenizing/tags.dart:40:95)
tags.dart:40
#1      PatternTagReaction._findTagAtCaret (package:super_editor/src/default_editor/text_tokenizing/pattern_tags.dart:292:22)
pattern_tags.dart:292
#2      PatternTagReaction._adjustTagAttributionsAroundAlteredTags (package:super_editor/src/default_editor/text_tokenizing/pattern_tags.dart:244:17)
pattern_tags.dart:244
#3      PatternTagReaction.react (package:super_editor/src/default_editor/text_tokenizing/pattern_tags.dart:218:5)
pattern_tags.dart:218
#4      Editor._reactToChanges (package:super_editor/src/core/editor.dart:218:16)
editor.dart:218
#5      Editor.execute (package:super_editor/src/core/editor.dart:161:9)
editor.dart:161
#6      TextDeltasDocumentEditor._insertPlainText (package:super_editor/src/default_editor/document_ime/document_delta_editing.dart:342:12)
document_delta_editing.dart:342
#7      TextDeltasDocumentEditor.insert (package:super_editor/src/default_editor/document_ime/document_delta_editing.dart:294:23)
document_delta_editing.dart:294
#8      TextDeltasDocumentEditor._applyInsertion (package:super_editor/src/default_editor/document_ime/document_delta_editing.dart:171:5)
document_delta_editing.dart:171
#9      TextDeltasDocumentEditor.applyDeltas (package:super_editor/src/default_editor/document_ime/document_delta_editing.dart:80:9)
document_delta_editing.dart:80
#10     DocumentImeInputClient.updateEditingValueWithDeltas (package:super_editor/src/default_editor/document_ime/document_ime_communication.dart:219:30)
[...]

PR #1879, opened 2024-03-04
Fix TagFinder.findTagAroundPosition to support emojis

BazinC:
Fixes https://github.com/superlistapp/super_editor/issues/1863 and https://github.com/superlistapp/super_editor/issues/1882

+ fix findTagAroundPosition to support of emojis
+ add widgetTest "support emojis" and "composing tag with consecutive trigger characters"

There was edge case in findTagAroundPosition looping through the parameter text code point one by one.
Fixed it with characters iterators.


BazinC:
@matthew-carroll let me know if your more detail. I'm happy to discuss it

matthew-carroll:
I didn't see this PR, but I just put up my own PR for this: https://github.com/superlistapp/super_editor/pull/1887

If you think I missed any conditions in that PR then please post the reproduction steps over in that PR for any cases that still fail.

If you're satisfied with my PR, please close this one.

BazinC:
I just uploaded a video and submitted it in your PR, requesting a review, without realizing you were already checking it out. I believe my addition includes more tests. I also had reservations regarding the use of the String extensions `moveOffsetUpstreamByCharacter` and '`moveOffsetDownstreamByCharacter`.

For example, '`moveOffsetUpstreamByCharacter`' iterates through the string from the beginning to the specified offset, and `TagFinder` invokes `moveOffsetUpstreamByCharacter` within a loop. With lengthy text, I can think of potential performance issues in the future. In my opinion, it would be more efficient to use two CharacterRange iterators directly, achieving the same result as TagFinder but with fewer iterations. I may be wrong, I would need to run some perf benchmark to prove it.

super_editor/test/super_editor/text_entry/tagging/stable_tags_test.dart
```diff
@@ -100,7 +100,9 @@ void main() {
               ),
             ],
           ),
-          const TagRule(trigger: "@"),
+          plugin: StableTagPlugin(
```
matthew-carroll, line 103: Why did this need to change?

super_editor/test/super_editor/text_entry/tagging/stable_tags_test.dart
```diff
@@ -391,6 +393,91 @@ void main() {
         // Ensure that we received a notification when the tag was cancelled.
         expect(tagNotificationCount, 3);
       });
+
+      testWidgetsOnAllPlatforms("support emojis", (tester) async {
```
matthew-carroll, line 397: Saying "support emojis" doesn't tell us anything about what it means to support emojis. Please give each test a clear and focused name/purpose. If you want to test multiple emoji behaviors, you can put them in a `group()` with an appropriate name.

super_editor/test/super_editor/text_entry/tagging/stable_tags_test.dart
```diff
@@ -391,6 +393,91 @@ void main() {
...
+          SuperEditorInspector.findDocumentSelection(),
+          const DocumentSelection.collapsed(
+            position: DocumentPosition(
+              nodeId: "1",
+              nodePosition: TextNodePosition(offset: 2),
+            ),
+          ),
+        );
+
+        //Place the caret before the emoji
```
matthew-carroll, line 428: This doesn't "place the caret" it "moves the caret". Please make sure behaviors are described accurately because small differences can result in very different execution paths.

super_editor/test/super_editor/text_entry/tagging/stable_tags_test.dart
```diff
@@ -391,6 +393,91 @@ void main() {
...
+        await tester.pressLeftArrow();
+
+        // Type @, the TagRule trigger
```
matthew-carroll, line 431: This comment doesn't really help the reader unless they already know all the implementation details. Can you rephrase this in a way that a general reader can understand what you're trying to do?

super_editor/test/super_editor/text_entry/tagging/stable_tags_test.dart
```diff
@@ -391,6 +393,91 @@ void main() {
...
+        // Ensure TagFinder.findTagArounPosition doesn't lead to a cropped emoji
+        // This would make Flutter ParagraphBuilder raise
+        // "Invalid argument(s): string is not well-formed UTF-16" on Flutter
+        await tester.typeImeText("@");
+
+        final text = SuperEditorInspector.findTextInComponent("1");
+        expect(text.text, "@💙");
+      });
```
matthew-carroll, line 439: What about an emoji that appears before a trigger? And an emoji that appears in the middle of a tag? Seems like there are some other variations that should probably be tested, too.

super_editor/test/super_editor/text_entry/tagging/stable_tags_test.dart
```diff
@@ -391,6 +393,91 @@ void main() {
...
+
+      testWidgetsOnAllPlatforms("composing tag with consecutive trigger characters", (tester) async {
```
matthew-carroll, line 441: Don't we already have tests for consecutive trigger characters?

super_editor/lib/src/default_editor/text_tokenizing/tags.dart
```diff
@@ -23,6 +23,91 @@ class TagFinder {
...
+    }
+
+    // Create 2 splits of characters to navigate upstream and downstream the caret position.
+    final charactersBefore = rawText.substring(0, splitIndex).characters;
+    final iteratorUpstream = charactersBefore.iteratorAtEnd;
+
+    final charactersAfter = rawText.substring(splitIndex).characters;
+    final iteratorDownstream = charactersAfter.iterator;
+
+    var movedBack = iteratorUpstream.moveBack();
```
matthew-carroll, line 43: Please don't use `var` - instead, declare the type. Also, if this is a boolean condition, please name is `didMoveBack`. In fact, given the specific usage, it should probably be named `didMoveUpstream`, because "back" doesn't have a clear meaning in this context.

super_editor/lib/src/default_editor/text_tokenizing/tags.dart
```diff
@@ -23,6 +23,91 @@ class TagFinder {
...
+
+    if (iteratorUpstream.current != tagRule.trigger) {
+      while (movedBack) {
+        final current = iteratorUpstream.current;
+        if (tagRule.excludedCharacters.contains(current)) {
+          // The upstream character isn't allowed to appear in a tag. Break before moving
```
matthew-carroll, line 49: Please add a description as to what this means in terms of our goal in this method. For example, when I read this, I'm wondering why we don't return immediately. If we haven't found a trigger character, and we ran into an excluded character, then doesn't that mean there's no tag here? If that's not what it means, then please make it clear in the comment what situation we're dealing with.

super_editor/lib/src/default_editor/text_tokenizing/tags.dart
```diff
@@ -23,6 +23,91 @@ class TagFinder {
...
+          // the starting character index any further upstream.
+          break;
+        }
+
+        if (current == tagRule.trigger) {
+          // The character we just added to the token bounds is the trigger.
```
matthew-carroll, line 55: Please update this comment to explicitly describe the index behavior to avoid confusion. It's not immediately obvious why we run another `moveBack()` call when we determined that the `current` character is the trigger.

super_editor/lib/src/default_editor/text_tokenizing/tags.dart
```diff
@@ -23,6 +23,91 @@ class TagFinder {
...
+          // We include it and stop looking any further upstream
+          iteratorUpstream.moveBack();
+          break;
+        }
+        movedBack = iteratorUpstream.moveBack();
+      }
+    } else if (movedBack) {
+      iteratorUpstream.moveNext();
```
matthew-carroll, line 63: What is this doing?

super_editor/lib/src/default_editor/text_tokenizing/tags.dart
```diff
@@ -23,6 +23,91 @@ class TagFinder {
...
+    final charactersBefore = rawText.substring(0, splitIndex).characters;
+    final iteratorUpstream = charactersBefore.iteratorAtEnd;
+
+    final charactersAfter = rawText.substring(splitIndex).characters;
+    final iteratorDownstream = charactersAfter.iterator;
+
+    var movedBack = iteratorUpstream.moveBack();
+
+    if (iteratorUpstream.current != tagRule.trigger) {
+      while (movedBack) {
```
matthew-carroll, line 46: Are you able to simplify some of the conditional logic down below by moving `iteratorUpstream.current != tagRule.trigger` into the `while()` loop?

super_editor/lib/src/default_editor/text_tokenizing/tags.dart
```diff
@@ -23,6 +23,91 @@ class TagFinder {
...
+          iteratorUpstream.moveBack();
+          break;
+        }
+        movedBack = iteratorUpstream.moveBack();
+      }
+    } else if (movedBack) {
+      iteratorUpstream.moveNext();
+    }
+
+    while (iteratorDownstream.moveNext()) {
```
matthew-carroll, line 66: Above this `while()` look, and also before the section above, please add a comment describing that you're searching upstream and downstream to find the end-cap indices of a tag.

super_editor/lib/src/default_editor/text_tokenizing/tags.dart
```diff
@@ -23,6 +23,91 @@ class TagFinder {
...
+      final current = iteratorDownstream.current;
+      if (current != tagRule.trigger && tagRule.excludedCharacters.contains(current)) {
+        break;
+      }
+    }
+
+    final tokenRange =
+        SpanRange(splitIndex - iteratorUpstream.stringAfterLength, splitIndex + iteratorDownstream.stringBeforeLength);
+    final tokenStartOffset = splitIndex - iteratorUpstream.stringAfterLength;
```
matthew-carroll, line 75: This `tokenStartOffset` appears to be a duplicate calculation from what was just done in `tokenRange`. Can you use the `tokenRange` instead of defining `tokenStartOffset`?

super_editor/lib/src/default_editor/text_tokenizing/tags.dart
```diff
@@ -23,6 +23,91 @@ class TagFinder {
...
+
+    final tagText = text.substringInRange(tokenRange);
+    if (!tagText.startsWith(tagRule.trigger)) {
```
matthew-carroll, line 78: Did you already check this condition higher up in this method where you chose the starting text offset? If we never found the trigger, then we don't need to search downstream, and we never need to make it this far down in the method.

super_editor/lib/src/default_editor/text_tokenizing/tags.dart
```diff
@@ -23,6 +23,91 @@ class TagFinder {
...
+      return null;
+    }
+
+    final tokenAttributions = text.getAttributionSpansInRange(attributionFilter: (a) => true, range: tokenRange);
```
matthew-carroll, line 82: Is there a way to get this information without passing `attributionFilter: (a) => true`?

super_editor/lib/src/default_editor/text_tokenizing/tags.dart
```diff
@@ -23,6 +23,91 @@ class TagFinder {
...
+    if (!isTokenCandidate(tokenAttributions.map((span) => span.attribution).toSet())) {
```
matthew-carroll, line 83: There's a lot going on in this if-statement. At a minimum, please add a comment inside the if-statement which describes the condition you just found and why we're returning `null`.

super_editor/lib/src/default_editor/text_tokenizing/tags.dart
```diff
@@ -23,6 +23,91 @@ class TagFinder {
...
+    final tagAroundPosition = TagAroundPosition(
+      indexedTag: IndexedTag(
+        Tag(tagRule.trigger, tagText.substring(1)),
+        nodeId,
+        tokenStartOffset,
+      ),
+      searchOffset: expansionPosition.offset,
+    );
+
+    return tagAroundPosition;
```
matthew-carroll, line 96: Can you return this directly? Is there a reason to create a new variable for this?

super_editor/lib/src/default_editor/text_tokenizing/tags.dart
```diff
@@ -23,6 +23,91 @@ class TagFinder {
...
+  }
+
+  static TagAroundPosition? _previousVersionOffindTagAroundPosition({
```
matthew-carroll, line 99: Please delete this if it's not needed any more.

matthew-carroll:
@angelosilvestre feel free to review this PR as well, if you find anything that I missed.

matthew-carroll:
Also @BazinC this PR is targeting the wrong branch - it should be targeting `main`, and it looks like tests are failing.

If you'd like to get this merged, please update all relevant details as soon as possible. We need this particular fix merged quickly.

super_editor/test/super_editor/text_entry/tagging/stable_tags_test.dart
BazinC: Because I changed `_pumpTestEditor` signature to take an optional SuperEditorPlugin. It was hard to keep troubleshoot emojis bugs and this way in 'support emoji' test I have access to the plugin and make useful `expect` checks on the committed tags and composing tag.

super_editor/lib/src/default_editor/text_tokenizing/tags.dart
BazinC: here no because the result won't be the same

BazinC:
@matthew-carroll thanks for your feedback! I'm working on it

super_editor/lib/src/default_editor/text_tokenizing/tags.dart
BazinC: This was also used in old version. I don't see any other method returning the expected result. I tried text.getAllAttributionsThroughout(tokenRange) but it doesn't return the same result (always empty)

super_editor/test/super_editor/text_entry/tagging/stable_tags_test.dart
BazinC: I needed this test because I broke the StableTagIndex behavior while I thought I fixed the main issue. But no test indicaticated something was broken. I added this test to ensure this behavior. I moved it after your comment into 'commit' group test.

BazinC: I don't see any consecutive trigger characters

BazinC:
@matthew-carroll I think I handled all your comments. Let me know if I missed anything.

BazinC:
Let me do another PR for main..

BazinC:
PR for the fix targeting `main`: https://github.com/superlistapp/super_editor/pull/1891

matthew-carroll:
Closing this PR in favor of `main` - there should only be one PR for this fix.

PR #1880, opened 2024-03-05
[SuperEditor][SuperReader] Fix setting attributions across multiple nodes (Resolves #1865)

### LINKED ISSUE #1865: [SuperEditor] Attributions are not toggled properly across multiple nodes
**Package Version**
`main` branch with commit hash 8a2a32fc2611f91de8834101f00d571845554b9b

**To Reproduce**
1. Go to SuperEditor demo
2. Add a paragraph node with bold text and another one below without bold text
3. Select the text of both nodes
4. Press CMD+b

**Actual behavior**
The attribution is toggled separately for both nodes, 1st node loses the bold attribution and the 2nd becomes bold.

**Expected behavior**
Both nodes get the same attribution, which with the current implementation, would both become bold.

**Platform**
All platforms.

**Flutter version**

**Screenshots**

https://github.com/superlistapp/super_editor/assets/31278849/dcf43a36-9ad1-498c-82b0-787df49d9641

rutvik110:
[SuperEditor][SuperReader] Fix setting attributions across multiple nodes (Resolves #1865)

### **Issue:** 

Toggling attributions across multiple nodes selection where one node has the attribution applied already and other doesn't results in a behaviour where the attribution would be toggled off for the node that already had it and toggled on for node that didn't had that attribution. 

https://private-user-images.githubusercontent.com/31278849/308606392-dcf43a36-9ad1-498c-82b0-787df49d9641.mov?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MDk3NDMyMDMsIm5iZiI6MTcwOTc0MjkwMywicGF0aCI6Ii8zMTI3ODg0OS8zMDg2MDYzOTItZGNmNDNhMzYtOWFkMS00OThjLTgyYjAtNzg3ZGY0OWQ5NjQxLm1vdj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDAzMDYlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwMzA2VDE2MzUwM1omWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWMxY2M3NjJiNTI0ZGVhZGVlODM5ZDY2ODg4YzhkOTUyZGRjZTdmMjc1YjA5YTg2MzI4YmM0ZDkzMTIzNjBhZTMmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.dxpZ5L3oGi-qL4cXVqxCGvYyKCbj85b2CBEAV6G0vNw

### **Solution:**

`ToggleTextAttributionsCommand` attribution toggling mechanism doesn't seem to be considering the scenario presented in the issue when we want to apply an attribution across multiple nodes and some of those nodes already having that attribution turned on. 

This PR updates `ToggleTextAttributionsCommand` toggling mechanism to check for two things when deciding to toggle an incoming attribution on a node. 

1. Is the attribution already applied throughout the entire node.
2. Are the number of nodes having all of the incoming attributions applied to at least one of their characters, match the number of selected nodes.

Based on this checks, we validate when to toggle an attribution for the node or not. 

Toggles attribution for an node if: 

1. If a node has the attribution applied partially.
2. If all nodes have all of the incoming attributions applied already to any of their characters.

Avoid toggling attribution for an node if: 

1. If a node already has the attribution applied throughout it and all of the nodes have all of the incoming attributions applied to any of their characters.

**With this change:** 

You can see when toggling attributions across two nodes, with one of the nodes either having bold attribution partially or fully, toggles bold attribution across both nodes as expected.

https://github.com/superlistapp/super_editor/assets/65209850/b1f10d59-264e-4888-bf20-f5cef7a412ef

### DIFF
--- a/super_editor/lib/src/default_editor/text.dart
+++ b/super_editor/lib/src/default_editor/text.dart
@@ -1444,9 +1444,7 @@ class ToggleTextAttributionsCommand implements EditCommand {
     // ignore: prefer_collection_literals
     final nodesAndSelections = LinkedHashMap<TextNode, SpanRange>();
 
-    /// Number of nodes having all of the given attributions applied to any of
-    /// their characters.
-    int nodesWithExistingAttributions = 0;
+    bool alreadyHasAttributions = true;
 
     for (final textNode in nodes) {
       if (textNode is! TextNode) {
@@ -1487,16 +1485,11 @@ class ToggleTextAttributionsCommand implements EditCommand {
 
       final selectionRange = SpanRange(startOffset, endOffset);
 
-      final bool alreadyHasAttributions = textNode.text.hasAttributionsWithin(
-        attributions: attributions,
-        range: selectionRange,
-      );
-
-      if (alreadyHasAttributions) {
-        // Each of the given attributions exists for atleast a single character within
-        // the current text node.
-        nodesWithExistingAttributions++;
-      }
+      alreadyHasAttributions = alreadyHasAttributions &&
+          textNode.text.hasAttributionsWithin(
+            attributions: attributions,
+            range: selectionRange,
+          );
 
       nodesAndSelections.putIfAbsent(textNode, () => selectionRange);
     }
@@ -1507,8 +1500,7 @@ class ToggleTextAttributionsCommand implements EditCommand {
         final node = entry.key;
         final range = entry.value;
 
-        if (node.text.hasAttributionsThroughout(attributions: {attribution}, range: range) &&
-            nodesWithExistingAttributions != nodesAndSelections.entries.length) {
+        if (!alreadyHasAttributions && node.text.hasAttributionsThroughout(attributions: {attribution}, range: range)) {
           // Attribution is applied throughout this entire node and there is alteast one node
           // that doesn't have this attribution. In that case, avoid toggling current node attribution
           // as this indicates the current attribution should be applied across all the nodes.
--- a/super_editor/lib/src/default_editor/text.dart
+++ b/super_editor/lib/src/default_editor/text.dart
@@ -1443,7 +1443,10 @@ class ToggleTextAttributionsCommand implements EditCommand {
 
     // ignore: prefer_collection_literals
     final nodesAndSelections = LinkedHashMap<TextNode, SpanRange>();
-    int nodesAndSelectionsWithExistingSimilarAttributions = 0;
+
+    /// Number of nodes having all of the given attributions applied to any of
+    /// their characters.
+    int nodesWithExistingAttributions = 0;
 
     for (final textNode in nodes) {
       if (textNode is! TextNode) {
@@ -1490,7 +1493,9 @@ class ToggleTextAttributionsCommand implements EditCommand {
       );
 
       if (alreadyHasAttributions) {
-        nodesAndSelectionsWithExistingSimilarAttributions++;
+        // Each of the given attributions exists for atleast a single character within
+        // the current text node.
+        nodesWithExistingAttributions++;
       }
 
       nodesAndSelections.putIfAbsent(textNode, () => selectionRange);
@@ -1502,8 +1507,13 @@ class ToggleTextAttributionsCommand implements EditCommand {
         final node = entry.key;
         final range = entry.value;
 
-        if (nodesAndSelectionsWithExistingSimilarAttributions != nodesAndSelections.entries.length &&
-            node.text.hasAttributionsThroughout(attributions: attributions, range: range)) {
+        if (node.text.hasAttributionsThroughout(attributions: {attribution}, range: range) &&
+            nodesWithExistingAttributions != nodesAndSelections.entries.length) {
+          // Attribution is applied throughout this entire node and there is alteast one node
+          // that doesn't have this attribution. In that case, avoid toggling current node attribution
+          // as this indicates the current attribution should be applied across all the nodes.
+
+          editorDocLog.info(' - attribution not toggled: $attribution. Range: $range');
           continue;
         }
 
--- a/super_editor/lib/src/default_editor/text.dart
+++ b/super_editor/lib/src/default_editor/text.dart
@@ -1443,7 +1443,7 @@ class ToggleTextAttributionsCommand implements EditCommand {
 
     // ignore: prefer_collection_literals
     final nodesAndSelections = LinkedHashMap<TextNode, SpanRange>();
-    bool alreadyHasAttributions = false;
+    int nodesAndSelectionsWithExistingSimilarAttributions = 0;
 
     for (final textNode in nodes) {
       if (textNode is! TextNode) {
@@ -1484,11 +1484,14 @@ class ToggleTextAttributionsCommand implements EditCommand {
 
       final selectionRange = SpanRange(startOffset, endOffset);
 
-      alreadyHasAttributions = alreadyHasAttributions ||
-          textNode.text.hasAttributionsWithin(
-            attributions: attributions,
-            range: selectionRange,
-          );
+      final bool alreadyHasAttributions = textNode.text.hasAttributionsWithin(
+        attributions: attributions,
+        range: selectionRange,
+      );
+
+      if (alreadyHasAttributions) {
+        nodesAndSelectionsWithExistingSimilarAttributions++;
+      }
 
       nodesAndSelections.putIfAbsent(textNode, () => selectionRange);
     }
@@ -1498,6 +1501,12 @@ class ToggleTextAttributionsCommand implements EditCommand {
       for (Attribution attribution in attributions) {
         final node = entry.key;
         final range = entry.value;
+
+        if (nodesAndSelectionsWithExistingSimilarAttributions != nodesAndSelections.entries.length &&
+            node.text.hasAttributionsThroughout(attributions: attributions, range: range)) {
+          continue;
+        }
+
         editorDocLog.info(' - toggling attribution: $attribution. Range: $range');
 
         // Create a new AttributedText with updated attribution spans, so that the presentation system can
--- a/super_editor/lib/src/default_editor/text.dart
+++ b/super_editor/lib/src/default_editor/text.dart
@@ -1444,8 +1444,6 @@ class ToggleTextAttributionsCommand implements EditCommand {
     // ignore: prefer_collection_literals
     final nodesAndSelections = LinkedHashMap<TextNode, SpanRange>();
 
-    bool alreadyHasAttributions = true;
-
     for (final textNode in nodes) {
       if (textNode is! TextNode) {
         continue;
@@ -1485,28 +1483,29 @@ class ToggleTextAttributionsCommand implements EditCommand {
 
       final selectionRange = SpanRange(startOffset, endOffset);
 
-      alreadyHasAttributions = alreadyHasAttributions &&
-          textNode.text.hasAttributionsWithin(
-            attributions: attributions,
-            range: selectionRange,
-          );
-
       nodesAndSelections.putIfAbsent(textNode, () => selectionRange);
     }
 
-    // Toggle attributions.
-    for (final entry in nodesAndSelections.entries) {
-      for (Attribution attribution in attributions) {
+    for (Attribution attribution in attributions) {
+      final allNodesHaveThisAttributionAppliedThroughout = nodesAndSelections.entries.every((entry) {
+        final node2 = entry.key;
+        final range2 = entry.value;
+        return node2.text.hasAttributionsThroughout(attributions: {attribution}, range: range2);
+      });
+
+      for (final entry in nodesAndSelections.entries) {
         final node = entry.key;
         final range = entry.value;
 
-        if (!alreadyHasAttributions && node.text.hasAttributionsThroughout(attributions: {attribution}, range: range)) {
-          // Attribution is applied throughout this entire node and there is alteast one node
-          // that doesn't have this attribution. In that case, avoid toggling current node attribution
-          // as this indicates the current attribution should be applied across all the nodes.
+        if (nodesAndSelections.length > 1) {
+          final bool hasAttributionAppliedThroughout =
+              node.text.hasAttributionsThroughout(attributions: {attribution}, range: range);
 
-          editorDocLog.info(' - attribution not toggled: $attribution. Range: $range');
-          continue;
+          if (hasAttributionAppliedThroughout) {
+            if (!allNodesHaveThisAttributionAppliedThroughout) {
+              continue;
+            }
+          }
         }
 
         editorDocLog.info(' - toggling attribution: $attribution. Range: $range');
--- a/super_editor/lib/src/default_editor/text.dart
+++ b/super_editor/lib/src/default_editor/text.dart
@@ -1444,6 +1444,8 @@ class ToggleTextAttributionsCommand implements EditCommand {
     // ignore: prefer_collection_literals
     final nodesAndSelections = LinkedHashMap<TextNode, SpanRange>();
 
+    bool alreadyHasAttributions = true;
+
     for (final textNode in nodes) {
       if (textNode is! TextNode) {
         continue;
@@ -1483,23 +1485,13 @@ class ToggleTextAttributionsCommand implements EditCommand {
 
       final selectionRange = SpanRange(startOffset, endOffset);
 
+      alreadyHasAttributions = alreadyHasAttributions &&
+          textNode.text.hasAttributionsThroughout(
+            attributions: attributions,
+            range: selectionRange,
+          );
+
       nodesAndSelections.putIfAbsent(textNode, () => selectionRange);
-    }
-
-    // Attributions that exists throughout the entire selection.
-    List<Attribution> attributionsPresentThroughoutEntireSelection = [];
-
-    for (Attribution attribution in attributions) {
-      final presentThroughoutEntireSelection = nodesAndSelections.entries.every((entry) {
-        final node2 = entry.key;
-        final range2 = entry.value;
-        return node2.text.hasAttributionsThroughout(attributions: {attribution}, range: range2);
-      });
-
-      if (presentThroughoutEntireSelection) {
-        /// Attribution exists across entire selection.
-        attributionsPresentThroughoutEntireSelection.add(attribution);
-      }
     }
 
     for (final entry in nodesAndSelections.entries) {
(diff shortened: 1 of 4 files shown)


rutvik110:
@matthew-carroll Latest updates, 

I had to try a different approach to support few edge cases I found, this resulted in following changes, 

1. Keeping track of which attributions are already applied throughout the entire selection. Needed when we want to decide whether to toggle an attribution for a node selection depending on whether it exists across entire user selection. Consider the scenario presented in the description.

2. When applying an attribution to a selection within node, if a node already has attribution applied throughout it then the decision to toggle the attribution depends on the presence of that attribution across entire user selection. If all nodes have it already applied, then we wanna toggle the attribution, if there exists even a single node selection that doesn't have the attribution applied throughout it then we want to keep the attribution on for nodes that already have it applied. 

3. Also added tests for different selection scenarios across single node, multiple nodes and multiple attributions.


Let me know if that makes sense.

matthew-carroll:
@rutvik110 on our call we identified a single boolean change that should have solved this issue. Why exactly did that not work? What situations did that fail to account for? Please be specific.

rutvik110:
> @rutvik110 on our call we identified a single boolean change that should have solved this issue. Why exactly did that not work? What situations did that fail to account for? Please be specific.

That solution didn't properly took into account whether the attributions that were present across all nodes selections were either applied partially or throughout all nodes selections.
This caused failures in toggling attributions correctly in situations where an attribution was present across all nodes selections which would meant either all nodes selections had that attribution applied throughout them or some nodes had it applied partially. Depending on which case it's, each node needs to decide whether it needs to keep the attribution or toggle it.

The current solution accounts for that as well as scenarios when multiple attributions are being applied.

matthew-carroll:
> That solution didn't properly took into account whether the attributions that were present across all nodes selections were either applied partially or throughout all nodes selections.

Please be specific so that I can confirm the claim. In what way was it not accounting for that? Please show the code in the recommended approach that was making the wrong decision.

rutvik110:
> > That solution didn't properly took into account whether the attributions that were present across all nodes selections were either applied partially or throughout all nodes selections.
> 
> Please be specific so that I can confirm the claim. In what way was it not accounting for that? Please show the code in the recommended approach that was making the wrong decision.

Earlier approach used this `alreadyHasAttributions` variable to check whether all of the incoming attributions already existed across every node. One thing this approach didn't took into account was whether the presence of those attributions was either partially or throughout those nodes, it only checked whether the attribution was present or not. 

```dart
    bool alreadyHasAttributions = true;

    for (final textNode in nodes) {
      alreadyHasAttributions = alreadyHasAttributions &&
          textNode.text.hasAttributionsWithin(
            attributions: attributions,
            range: selectionRange,
          );
     }

    for (final entry in nodesAndSelections.entries) {
      for (Attribution attribution in attributions) {
        if (!alreadyHasAttributions && node.text.hasAttributionsThroughout(attributions: {attribution}, range: range)) {

          // attribution not toggled
          continue;
        }
        
        // toggle attribution
    }
 }
```

The current approach very clearly tells us about such attributions by maintaining this list, `attributionsPresentThroughoutEntireSelection`. This makes it easier down the line to toggle attribution for a node relative to how the attribution is present throughout entire selection.

rutvik110:
@matthew-carroll This is ready for review.

super_editor/lib/src/default_editor/text.dart
```diff
@@ -1484,33 +1485,54 @@ class ToggleTextAttributionsCommand implements EditCommand {
...
+          // Attribution is present throughout the user selection. Remove attribution.
+
+          editorDocLog.info(' - Removing attribution: $attribution. Range: $range');
+
+          // Create a new AttributedText with updated attribution spans, so that the presentation system can
+          // see that we made a change, and re-renders the text in the document.
+          node.text = AttributedText(
+              node.text.text,
+              node.text.spans.copy()
+                ..removeAttribution(
```
matthew-carroll, line 1514: Do we have a method that can remove multiple attributions at the same time?

super_editor/lib/src/default_editor/text.dart
```diff
@@ -1484,33 +1485,54 @@ class ToggleTextAttributionsCommand implements EditCommand {
...
-            ),
-        );
+        if (alreadyHasAttributions) {
+          // Attribution is present throughout the user selection. Remove attribution.
+
+          editorDocLog.info(' - Removing attribution: $attribution. Range: $range');
+
+          // Create a new AttributedText with updated attribution spans, so that the presentation system can
+          // see that we made a change, and re-renders the text in the document.
+          node.text = AttributedText(
```
matthew-carroll, line 1511: Do we have a method directly on `AttributedText` that can remove an attribution? I'm surprised to see what we're directly manipulating the spans.

super_editor/lib/src/default_editor/text.dart
```diff
@@ -1484,33 +1485,54 @@ class ToggleTextAttributionsCommand implements EditCommand {
...
+                  end: range.end,
+                ));
+        } else {
+          // Attribution isn't present throughout the user selection. Apply attribution.
+
+          editorDocLog.info(' - Adding attribution: $attribution. Range: $range');
+
+          // Create a new AttributedText with updated attribution spans, so that the presentation system can
+          // see that we made a change, and re-renders the text in the document.
+          node.text = AttributedText(
```
matthew-carroll, line 1526: Same questions here as above.

super_editor/lib/src/test/super_editor_test/supereditor_inspector.dart
```diff
@@ -291,6 +301,24 @@ class SuperEditorInspector {
     return documentLayoutElement.state as DocumentLayout;
   }
 
+  /// Toggles given [attributions] for the [documentSelection].
+  ///
+  /// {@macro supereditor_finder}
+  static void toggleAttributionsForDocumentSelection(
```
matthew-carroll, line 307: I'm confused why this is here? This appears to be defined in the `SuperEditorInspector`, but this method isn't inspecting anything...

super_editor/test/super_editor/supereditor_attributions_test.dart
```diff
@@ -396,3 +1108,220 @@ void main() {
     });
   });
 }
+
+MutableDocument _paragraphDoc() => MutableDocument(
```
matthew-carroll, line 1112: Did you check for existing test documents before creating these?

super_editor/test/super_editor/supereditor_attributions_test.dart
```diff
@@ -181,112 +182,823 @@ void main() {
...
+              .createDocument()
+              .withCustomContent(
+                _paragraphDoc(),
+              )
+              .pump();
+
+          final doc = SuperEditorInspector.findDocument()!;
+
+          // Ensure no attributions are present.
+          expect(
```
matthew-carroll, line 198: To check attributions, all you need to do is access the node's AttributedText spans and ensure the markers are empty.

When serializing to Markdown saves a lot of verbosity then it's an acceptable thing to do. But in this case, and some others in this test suite, I see 3 issues:

- You're barely saving any code length
- You're adding the processing time to serialize Markdown
- Your comment that says "Ensure no attributions are present" is technically wrong because not all attributions have Markdown equivalents (like colored text)

super_editor/test/super_editor/supereditor_attributions_test.dart
```diff
@@ -181,112 +182,823 @@ void main() {
...
+              base: firstNode.beginningDocumentPosition,
+              extent: firstNode.endDocumentPosition,
+            ),
+            {boldAttribution},
+          );
+
+          // Ensure attribution was applied throughout the selection.
+          expect(
+            doc,
+            equalsMarkdown(
```
matthew-carroll, line 218: This use of Markdown serialization is reasonable because it saves you from a bit of excess code and it avoids fragile boundary indices.

super_editor/test/super_editor/supereditor_attributions_test.dart
```diff
@@ -181,112 +182,823 @@ void main() {
...
+          // Ensure attribution was removed from the selection.
+          expect(
+            doc,
+            equalsMarkdown(
+              "This is the first node in a document.",
+            ),
+          );
+        });
+
+        testWidgetsOnAllPlatforms("toggles attribution for a partially selected fully attributed node", (tester) async {
```
matthew-carroll, line 293: I don't think "fully attributed node" is really what you care about here, is it? Aren't you really testing what happens when toggling an attribution within a subrange of an existing attribution?

super_editor/test/super_editor/supereditor_attributions_test.dart
```diff
@@ -181,112 +182,823 @@ void main() {
...
+
+          // Ensure bold attribution is applied throughout the node.
+          expect(
+            doc,
+            equalsMarkdown(
+              "**This is the first**** node in a document.**",
+            ),
+          );
+        });
+        testWidgetsOnAllPlatforms("toggles a different attribution for fully attributed node", (tester) async {
```
matthew-carroll, line 345: Make sure there's a blank line between every test.

matthew-carroll, line 345: Here again I don't think "fully attributed node" is the important part of the test.

super_editor/test/super_editor/supereditor_attributions_test.dart
```diff
@@ -181,112 +182,823 @@ void main() {
...
+          // Ensure bold attribution is applied throughout the node.
+          expect(
+            doc,
+            equalsMarkdown(
+              "**This is the first node in a document.**",
+            ),
+          );
+        });
+
+        testWidgetsOnAllPlatforms("toggles multiple attributions throughout a node", (tester) async {
```
matthew-carroll, line 398: This test name doesn't communicate very much. Based on what I'm seeing in this test, it looks like you're testing that multiple attributions can be added and removed at the same time. 

The "throughout a node" part seems to be incidental. Does it need to be "throughout a node"? Is there some difference when this same situation happens and its not "throughout a node"?

super_editor/test/super_editor/supereditor_attributions_test.dart
```diff
@@ -181,112 +182,823 @@ void main() {
...
+            doc,
+            equalsMarkdown(
+              "This is the first node in a document.",
+            ),
+          );
+        });
+      });
+
+      group("when multiple nodes are selected", () {
+        testWidgetsOnAllPlatforms("toggles attribution across multiple nodes", (tester) async {
```
matthew-carroll, line 453: This test isn't merely "across multiple nodes", this test is toggling the attribution throughout the entirety of multiple nodes.

super_editor/test/super_editor/supereditor_attributions_test.dart
```diff
@@ -181,112 +182,823 @@ void main() {
...
+          // Ensure bold attribution was removed from both nodes.
+          expect(
+            doc,
+            equalsMarkdown(
+              "This is the first node in a document.\n\nThis is the second node in a document.",
+            ),
+          );
+        });
+
+        testWidgetsOnAllPlatforms("toggles an attribution across a fully attributed node and a plain node",
```
matthew-carroll, line 507: "plain node" doesn't mean anything. Also, for the "fully attributed node" I think you missed the most important part. You're toggling the same attribution that's applied to the first node.

super_editor/test/super_editor/supereditor_attributions_test.dart
```diff
@@ -181,112 +182,823 @@ void main() {
...
+
+          SuperEditorInspector.toggleAttributionsForDocumentSelection(
+            DocumentSelection(
+              base: firstNode.beginningDocumentPosition,
+              extent: secondNode.endDocumentPosition,
+            ),
+            {boldAttribution},
+          );
+
+          // Ensure bold attribution is applied throughout both nodes.
```
matthew-carroll, line 537: You should WHY its now applied throughout both nodes.

super_editor/test/super_editor/supereditor_attributions_test.dart
```diff
@@ -181,112 +182,823 @@ void main() {
...
+          // Ensure bold attribution was removed from both nodes.
+          expect(
+            doc,
+            equalsMarkdown(
+              "This is the first node in a document.\n\nThis is the second node in a document.",
+            ),
+          );
+        });
+
+        testWidgetsOnAllPlatforms("toggles an attribution across a partially attributed node and a plain node",
```
matthew-carroll, line 562: Same comment on this name as one up above. You should revisit all test names and ensure that they are as accurate and precise as possible. We're testing a lot of slight variations in here so the names are really important to do well.

super_editor/test/super_editor/supereditor_attributions_test.dart
```diff
@@ -181,112 +182,823 @@ void main() {
...
+          );
+
+          // Ensure both bold and italic attributions were removed from the selection.
+          expect(
+            doc,
+            equalsMarkdown(
+              "This is the first node in a document.\n\nThis is the second node in a document.",
+            ),
+          );
+        });
```
matthew-carroll, line 894: I don't think I noticed any variation in selection bounds. They seemed to all go from the beginning of paragraph 1 to end of paragraph 2. What about something like the halfway point of paragraph 1 to the halfway point of paragraph 2? And how about a three paragraph situation where you go from halfway in paragraph 1, all the way through paragraph 2, and half way into paragraph 3?

Seems like we should at least have a few tests that vary the selection bounds.

super_editor/lib/src/default_editor/text.dart
rutvik110: We don't seem to have such method.

super_editor/lib/src/default_editor/text.dart
rutvik110: yah, we do have add/remove methods on `AttributedText`. I've updated this part to use those methods on newly created attributed text.

super_editor/lib/src/test/super_editor_test/supereditor_inspector.dart
rutvik110: ah yah, it doesn't seem the right place for this. Do you think we should create a `SuperEditorUtils` for tests where we can have such methods that help reduce boilerplate or simplify larger commands like this one and don't fall within `SuperEditorRobot` and `SuperEditorInspector`?

super_editor/test/super_editor/supereditor_attributions_test.dart
rutvik110: yes, we didn't have any existing document creators that I found suitable for tests here.

super_editor/test/super_editor/supereditor_attributions_test.dart
rutvik110: > Is there some difference when this same situation happens and its not "throughout a node"?

No, there's no difference between these two cases. 

> Does it need to be "throughout a node"?
As for the test, it's the entire node for which we're toggling the attribution, so I think it helps here clarify that.

super_editor/test/super_editor/supereditor_attributions_test.dart
rutvik110: I'll add some tests for those cases.

super_editor/test/super_editor/supereditor_attributions_test.dart
rutvik110: yes, will update the naming for tests.

rutvik110:
@matthew-carroll Latest updates, 

* Using add/remove attribution methods from `AttributedText` instead of directly performing them on text spans. 
* Using text markers to check for absence of attributions across text instead of document serialization approach. 
* Test names updated to reflect test behaviour more clearly.
* Tests for partial selections across multiple nodes added.
* Moved some of the newly added document creators to the `test_documents`.
* Using existing `AttributedText` `positionAt` method to get document position at a specific index.

Waiting for update on [this](https://github.com/superlistapp/super_editor/pull/1880#discussion_r1536748466) comment.

super_editor/lib/src/test/super_editor_test/supereditor_inspector.dart
matthew-carroll: We should never create anything called "utils" - that's just a junk drawer where people will throw random code. If you call something "utils" it's because you haven't take the time to figure out where it really belongs or what role it plays.

My first question is why this behavior needs to be shared at all. Do we have other test files that use this?

rutvik110: > My first question is why this behavior needs to be shared at all. Do we have other test files that use this?

Nope. I thought it might aid in toggling attributions in tests without writing much of the boilerplate for the command. Though right now, I think it's best to put it in the test file.

super_editor/lib/src/default_editor/text.dart
```diff
@@ -1484,33 +1485,51 @@ class ToggleTextAttributionsCommand implements EditCommand {
...
+          // Attribution is present throughout the user selection. Remove attribution.
+
+          editorDocLog.info(' - Removing attribution: $attribution. Range: $range');
+
+          // Create a new AttributedText with updated attribution spans, so that the presentation system can
+          // see that we made a change, and re-renders the text in the document.
+          node.text = AttributedText(
+            node.text.text,
+            node.text.spans.copy(),
+          )..removeAttribution(
```
matthew-carroll, line 1514: Can you just say `node.text.removeAttribution(...)`?

I see that the comment above says we're "Creating a new AttributedText...", though it's not clear why we're creating a new one just to assign it right back. 

I suppose it's possible we did that in case other areas of the code hold a reference to that text. If you mutate it directly, do any tests fail? Do any real editor behaviors fail when you try them?

super_editor/lib/src/default_editor/text.dart
```diff
@@ -1484,33 +1485,51 @@ class ToggleTextAttributionsCommand implements EditCommand {
...
+        } else {
+          // Attribution isn't present throughout the user selection. Apply attribution.
+
+          editorDocLog.info(' - Adding attribution: $attribution. Range: $range');
+
+          // Create a new AttributedText with updated attribution spans, so that the presentation system can
+          // see that we made a change, and re-renders the text in the document.
+          node.text = AttributedText(
+            node.text.text,
+            node.text.spans.copy(),
```
matthew-carroll, line 1527: Same question as above for copying vs mutating.

super_editor/test/super_editor/test_documents.dart
```diff
@@ -277,3 +277,118 @@ MutableDocument singleParagraphFullColor() => MutableDocument(
         )
       ],
     );
+
+MutableDocument paragraphDoc() => MutableDocument(
```
matthew-carroll, line 281: Don't we already have a one-paragraph and two-paragraph document in here? Why do we need more?

super_editor/test/super_editor/test_documents.dart
```diff
@@ -277,3 +277,118 @@ MutableDocument singleParagraphFullColor() => MutableDocument(
...
+        ParagraphNode(
+          id: "2",
+          text: AttributedText(
+            "This is the second node in a document.",
+          ),
+        ),
+      ],
+    );
+
+MutableDocument paragraphThenParagraphThenParagraphDoc() => MutableDocument(
```
matthew-carroll, line 309: Notice that this name doesn't follow the way documents are named at the top of this file. This document should probably be called `threeParagraphDoc`. Moreover, if we really need this document, it should probably be placed right below the existing two paragraph document(s).

super_editor/test/super_editor/test_documents.dart
```diff
@@ -277,3 +277,118 @@ MutableDocument singleParagraphFullColor() => MutableDocument(
...
+        ParagraphNode(
+          id: "3",
+          text: AttributedText(
+            "This is the third node in a document.",
+          ),
+        ),
+      ],
+    );
+
+MutableDocument fullyBoldParagraphDoc() => MutableDocument(
```
matthew-carroll, line 332: To match existing naming styles, this should probably be called `singleParagraphDocAllBold`

super_editor/test/super_editor/test_documents.dart
```diff
@@ -277,3 +277,118 @@ MutableDocument singleParagraphFullColor() => MutableDocument(
...
+                offset: 36,
+                markerType: SpanMarkerType.end,
+              ),
+            ]),
+          ),
+        ),
+      ],
+    );
+
+MutableDocument paragraphFullBoldThenParagraphFullyBold() => MutableDocument(
```
matthew-carroll, line 355: This should probably be called `twoParagraphDocAllBold`

super_editor/lib/src/test/super_editor_test/supereditor_inspector.dart
```diff
@@ -49,6 +49,16 @@ class SuperEditorInspector {
     return imeInteractor.isAttachedToIme;
   }
 
+  /// Returns the [Editor] within the [SuperEditor] matched by [finder],
+  /// or the singular [SuperEditor] in the widget tree, if [finder] is `null`.
+  ///
+  /// {@macro supereditor_finder}
+  static Editor findEditor([Finder? finder]) {
```
matthew-carroll, line 56: Why do we need this? Did you check how many other tests currently obtain the `Editor`?

super_editor/lib/src/default_editor/text.dart
rutvik110: > If you mutate it directly, do any tests fail? Do any real editor behaviors fail when you try them?

A single test seems to be failing with this change. The test is trying to validate whether a fontstyle on a span changed after applying the attribution, but fails.

And when tested in running application, the editor doesn't seem to be receiving the update that an attribution has been toggled on a text and doesn't update itself,

https://github.com/superlistapp/super_editor/assets/65209850/3c2e7810-8851-4043-aba6-d1e270b6fd42

super_editor/test/super_editor/test_documents.dart
rutvik110: The existing ones we have are either with an empty paragraph or a combination of a paragraph and a horizontal node in the document. There's `singleParagraphDoc` but the paragraph content much longer than what we need for these tests.

The newly added `paragraphDoc` and `paragraphThenParagraphDoc` only consists of paragraphs and have shorter content for testing.

super_editor/lib/src/test/super_editor_test/supereditor_inspector.dart
rutvik110: Fixed it. Removed this method and updated my tests to follow what other tests are doing to obtain editor.

rutvik110:
@matthew-carroll Latest updates, 

* Removed `findEditor` method I added to `SuperEditorInspector` and following what other tests do to obtain the editor.
* Updated naming for newly added test documents creators.

matthew-carroll (approved):
LGTM

PR #1896, opened 2024-03-17
[super_editor_markdown] - Parse and apply inline Markdown styles in an EditorReaction (Resolves #1866)

### LINKED ISSUE #1866: [SuperEditor] markdown support for bold and italics
Support edit reactions to detect markdown annotation for bold and italics:
* when the user types `something *italic*` it should convert the content to "something *italic*"
* when the user types `something **bold**` it should convert the content to "something **bold**"

matthew-carroll:
Parse and apply inline Markdown styles in an EditorReaction.

Applying inline Markdown means recognizing inline Markdown syntax, removing the syntax, and adding appropriate attributions where the syntax was. Example: `"Hello *italics*"` with "Hello *italics*".

Inline Markdown could appear as a result of multiple causes, e.g., the user typing a character, the user pasting text, or a programmer edit operation. Therefore, it's not always clear what we should consider when attempting to apply inline Markdown parsing.

This PR inspects all text content in every `TextNode` that's edited during a given transaction. All inline Markdown that's found in an edited node is applied.

### Ambiguous Changes
Markdown syntax for bold and italics makes it ambiguous to identify the user's intention when the user enters only one character at a time.

Imagine that the user wants to achieve '"Hello **bold**"'. Consider the following steps to type that:

```
Hello **bold
Hello **bold*
Hello **bold**
```

If the inline Markdown is blindly applied to the text, the user would end up with the following results:

Hello **bold

Hello **bold*

Hello *bold*

The user would end up with italic text, not bold text, because each "*" was applied on its own. The user never got a chance to enter two "**" back to back.

To solve this problem, the inline Markdown reaction first converts the existing `TextNode`s `AttributedText` to a Markdown string. Then, that Markdown string is parsed back to `AttributedText`. That process looks like the following:

Start with: `Hello **bold`

User types "*", creating text: `Hello **bold*`

The reaction serializes the `AttributedText` and gets: `Hello **bold*` (no change)

The reaction deserializes the Markdown and gets: "Hello *bold" (with italics)

User types "*", creating text: `Hello *bold*` (with italics)

The reaction serializes the `AttributedText` and gets: `Hello **bold**` (notice that we've recovered the extra "*"s)

The reaction deserializes the Markdown and gets: "Hello **bold**" (with bold)

By going from `AttributedText` -> Markdown -> `AttributedText`, we successfully honor ambiguous Markdown syntax.

### Surgical changes
The easiest way to apply a Markdown change is to throw away the previous paragraph and replace it with the newly parsed and styled paragraph. However, when considering undo/redo systems, which remember all deleted text, repeatedly deleting and recreating entire paragraphs quickly becomes costly.

This PR brings in a text diff'ing package so that text and styles are only inserted/deleted where necessary. This adds complexity to the reaction, but reduces the memory footprint for the history ledger.

### Selection adjustments
The selection base/extent might sit within the `TextNode` where inline Markdown is applied. Therefore, the reaction in this PR adjusts the selection bounds based on the amount of content that's either inserted or deleted.


matthew-carroll:
CC @dhikshith12

dhikshith12:
Curious to know how `***Bold And Italic***` is handled while typing the closing `***` stars.

In Notion, It will start with bold and then become `*BOLD` and as soon as another star is added at the end it then becomes bold and italic.

dhikshith12:
https://github.com/superlistapp/super_editor/assets/59935432/5d2f1765-b50a-4df5-91ce-4015c4877e66

matthew-carroll:
@miguelcmedeiros I fixed the situation you identified over in the issue ticket where starting a block of text with Markdown syntax causes that syntax to disappear.

matthew-carroll:
For reference, I filed this issue in the `markdown` package: https://github.com/dart-lang/markdown/issues/599

miguelcmedeiros (approved):
LGTM

angelosilvestre:
@matthew-carroll Yeah, I'll take a look in a few minutes

super_editor_markdown/lib/src/markdown_inline_upstream_plugin.dart
```diff
@@ -0,0 +1,574 @@
...
+      final nodeId = (change.change as NodeDocumentChange).nodeId;
+      if (editedTextNodes.containsKey(nodeId)) {
+        continue;
+      }
+
+      if (document.getNodeById(nodeId) is! TextNode) {
+        continue;
+      }
+
+      editedTextNodes[nodeId] = document.getNodeById(nodeId)!.id;
```
angelosilvestre, line 132: Isn't this the same as `editedTextNodes[nodeId] = nodeId;` ? Also, couldn't `editedTextNodes` be a set instead of a map?

super_editor_markdown/lib/src/markdown_inline_upstream_plugin.dart
```diff
@@ -0,0 +1,574 @@
...
+
+    int offset = caretOffset - 1;
+
+    // Start visiting upstream characters by visiting the first character
+    // and checking for possible syntaxes.
+    // final styleSyntax =
+    //     StyleUpstreamMarkdownSyntaxParser._maybeCreateForCharacter(attributedText.text[offset], offset);
+    // if (styleSyntax != null) {
+    //   _possibleSyntaxes.add(styleSyntax);
+    // }
```
angelosilvestre, line 248: Please remove the commented code.

super_editor_markdown/pubspec.yaml
```diff
@@ -13,7 +13,9 @@ dependencies:
 
   super_editor: ^0.2.5
   logging: ^1.0.1
-  markdown: ^4.0.0
+  markdown: ^5.0.0
+  diff_match_patch: ^0.4.1
+  collection: ^1.18.0
```
angelosilvestre, line 18: Are these dependencies being used? I didn't notice any new imports related to these packages in the modified files.

super_editor_markdown/test/markdown_inline_upstream_plugin_test.dart
```diff
@@ -0,0 +1,612 @@
...
+import 'package:super_editor/super_editor.dart';
+import 'package:super_editor/super_editor_test.dart';
+import 'package:super_editor_markdown/src/markdown_inline_upstream_plugin.dart';
+import 'package:super_editor_markdown/src/markdown_to_document_parsing.dart';
+
+void main() {
+  group("Super Editor upstream Markdown reaction >", () {
+    group("at beginning of paragraph >", () {
+      testWidgets("bold", (tester) async {
+        final document = deserializeMarkdownToDocument("");
```
angelosilvestre, line 12: Is there any reason why are you deserializing an empty markdown instead of manually creating the document?

super_editor_markdown/test/markdown_inline_upstream_plugin_test.dart
```diff
@@ -0,0 +1,612 @@
...
+                editor: editor,
+                document: document,
+                composer: composer,
+                plugins: {
+                  MarkdownInlineUpstreamSyntaxPlugin(),
+                },
+              ),
+            ),
+          ),
+        );
```
angelosilvestre, line 60: Also, this initialization code looks repeated across many tests. How about extracting it to a method?

angelosilvestre (commented):
I left some comments.

PR #1907, opened 2024-03-25
Upgrade markdown package to v7.2.0

Maksimka101:
Additionally, I found a bug in the `unordered list followd by empty list item` test.

In previous versions of the markdown package, parsing this markdown `- list item 1\n   - ` resulted in the following structure:
A list (li) consisting of one item (ul). And in this item, there was a header (h2) containing the text `list item 1`

The header appeared in the first list item because during internal transformations, the text initially turned into `list item 1\n  - `, which was then interpreted as header syntax. In the new version of the package, this behavior has been corrected, and now this markdown is parsed as a list with one item containing the text `list item 1\n  -`

I modified the test to accurately reflect its description. That is, now it indeed contains a list of two items, the last of which is empty

### DIFF
--- a/super_editor_markdown/lib/src/image_syntax.dart
+++ b/super_editor_markdown/lib/src/image_syntax.dart
@@ -24,18 +24,18 @@ class SuperEditorImageSyntax extends md.LinkSyntax {
         );
 
   @override
-  md.Node? close(
+  Iterable<md.Node>? close(
     md.InlineParser parser,
     covariant md.SimpleDelimiter opener,
     md.Delimiter? closer, {
     String? tag,
     required List<md.Node> Function() getChildren,
   }) {
-    var text = parser.source!.substring(opener.endPos, parser.pos);
+    var text = parser.source.substring(opener.endPos, parser.pos);
     // The current character is the `]` that closed the link text. Examine the
     // next character, to determine what type of link we might have (a '('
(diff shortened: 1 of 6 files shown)


matthew-carroll:
Thanks for the PR. At the moment, I'm working on typing inline Markdown and automatically converting it to attributions. We'll need to get that merged first and then see how it works with newer version of the markdown package. If things continue to work, then we can merge the upgraded dependency.

Maksimka101:
You are working on the markdown functionality, so you will update the markdown package to the latest version yourself, and my pull request is not needed. Did I understand correctly?

matthew-carroll:
No, I'm not going to upgrade to the latest version, because my work didn't need the latest version.

Once I merge my Markdown work, you'll need to rebase this PR against `main` and check again to make sure everything still compiles and tests pass with the latest version.

This is the PR we're waiting for: https://github.com/superlistapp/super_editor/pull/1896

matthew-carroll:
We just merged in the other PR. Can you bring the latest from `main` and adjust as needed?

Maksimka101:
Thanks for notifying me. I'll update it. But probably not very soon

matthew-carroll:
Gonna close this PR in favor of this one: https://github.com/superlistapp/super_editor/pull/2364

PR #1908, opened 2024-03-26
Added change event for adding/removing attributions, also switched from generic event to a specific TextInsertionEvent for InsertAttributedTextCommand (Resolves #1884)

### LINKED ISSUE #1884: Need more information what changed in a TextNode when adding/removing/modifying attributions
Blocker for #1883.

When a user adds/removes/modifies attributions such as bold/italic/strikethrough in a paragraph in a document, I get `NodeChangeEvent (<node id>)` and that's it.

I'd like to get more context what changed:
* When a new attribution is added, I need to know the attribution itself + the range of text in a given text node that the attribution applies to.
* When an existing attribution is removed, I need to know the attribution itself + the range of text that the attribution was removed from.

The reason is that I need to construct a Quill Delta for adding/removing the attribution here: https://github.com/superlistapp/super_editor/blob/22ca94897427070a7eba8c62017af1523e52924a/super_editor_quill/lib/src/delta_document_change_listener.dart#L42

Sample on how the Quill Deltas I need to construct would look like:
* abc -> a**b**c = `Delta()..retain(1)..retain(1, {'bold': true})`
* a**b**c -> abc = `Delta()..retain(1)..retain(1, {'bold': null})`

matthew-carroll:
Added change event for adding/removing attributions, also switched from generic event to a specific TextInsertionEvent for InsertAttributedTextCommand (Resolves #1884)

### DIFF
--- a/super_editor/lib/src/core/editor.dart
+++ b/super_editor/lib/src/core/editor.dart
@@ -441,6 +441,13 @@ class DocumentEdit implements EditEvent {
 
   @override
   String toString() => "DocumentEdit -> $change";
+
+  @override
+  bool operator ==(Object other) =>
+      identical(this, other) || other is DocumentEdit && runtimeType == other.runtimeType && change == other.change;
+
+  @override
+  int get hashCode => change.hashCode;
 }
 
 /// An object that's notified with a change list from one or more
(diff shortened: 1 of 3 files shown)


matthew-carroll:
@roughike Please let me know if the new event I added solves your needs. Also, please let me know if I missed any relevant change that needs a more surgical event.

roughike:
@matthew-carroll Yes - this indeed solves #1884, thanks!

I do need some information (=other events) for other cases, but for the specific issue I filed in #1884, this PR fixes it.

roughike (approved):

angelosilvestre (approved):
LGTM

PR #1916, opened 2024-04-02
Push SliverToBoxAdapter down to SingleColumnDocumentLayout

knopp:
**Background:** The way SuperEditor currently handles document widgets where all the widgets in the document are instantiated at all times is presenting a significant bottleneck to us. We aim to implement an alternative document layout based on [super_sliver_list](https://github.com/superlistapp/super_sliver_list), but in order to accommodate a sliver inside layout some adjustments needs to be made to `SuperEditor` infrastructure.

This PR ensures that when `SuperEditor` is placed into a `CustomScrollView`, it propagates the sliver geometry all the way down to the layout, instead of relying to be wrapped inside a `SliverToBoxAdapter`.

Breaking use-facing changes introduced in this PR:

- To use `SuperEditor` inside a custom scroll view, it must no longer be wrapped in a `SliverToBoxAdapter`. That's because the `SuperEditor` itself acts like a sliver. When SuperEditor is used as "standalone" widget, no changes are necessary.
- It is no longer possible to use `IntrinsicHeight` around SuperEditor, as this does not work with `CustomScrollView`, which is now used instead of `SingleChildScrollView` in `DocumentScrollable`. Instead of that, `shrinkWrap` argument has been added to `SuperEditor` and `SuperReader` to accomplish similar behavior.
- Getting the document layout state and casting it to `RenderBox` will fail, because the layout is a `RenderSliver`.

Some implementation details:

`SliverHybridStack` widget was introduced to allow combining slivers and render boxes. Exactly on of the children in the stack must produce a `RenderSliver`, while other children produce `RenderObject`s. The render objects are laid-out and painted in a way where they cover the entire sliver scrollExtent.

`ContentLayers` has been modified to work similar to `SliverHybridStack`, where the content is a sliver while the overlays and underlays are render boxes. I think this is acceptable solution as the underlays and overlays are expected to be much sparsely populated then the content itself.

`ViewportBoundsReporter` and `RenderViewportBoundsReplicator` are gone. The viewport bounds are part of sliver constraints. `SliverHybridStack` has an optional mode which ensures that the renderboxes fill up entire viewport, which is enabled for gesture system RenderBox in case the entire works in standalone mode (not in user provided scrollable).


super_editor/test/super_reader/super_reader_scrolling_test.dart
```diff
@@ -175,7 +176,7 @@ void main() {
         DocumentSelection(
           base: DocumentPosition(
             nodeId: lastParagraph.id,
-            nodePosition: lastParagraph.endPosition.copyWith(affinity: TextAffinity.upstream),
```
knopp, line 178: Note that the `affinity: TextAffinity.upstream` was actually incorrect, and was just an artifact of `AutoScroller.deltaWhileAutoScrolling` being one frame late, giving slightly wrong base position after scrolling.

matthew-carroll:
I haven't looked at the source code yet, but I have some thoughts based on the write-up. Mostly, I want to make sure that we don't over correct in the direction of slivers, to the detriment of those who would like to operate on boxes.

> Getting the document layout state and casting it to RenderBox will fail, because the layout is a RenderSliver.

Unless I'm misunderstanding, this seems like a problem for those who want to work with boxes and not slivers. Boxes make it trivial to calculate offsets and assemble visual rectangle regions, so my first thought is that we don't want to eliminate the ability to treat a document as a box, if the user is actually using it as a box.

> ContentLayers has been modified to work similar to SliverHybridStack, where the content is a sliver while the overlays and underlays are render boxes. I think this is acceptable solution as the underlays and overlays are expected to be much sparsely populated then the content itself.

I've long planned to move `ContentLayers` into its own package because I think it's generally useful. In so doing, I'm sure people will want to use boxes as the content. Should we instead look into the options to have two different content layer implementations - one for boxes and one for slivers? We'd need to chat about the optimal API for such a thing, but I thought I'd introduce this point at the start of the review.

knopp:
> I haven't looked at the source code yet, but I have some thoughts based on the write-up. Mostly, I want to make sure that we don't over correct in the direction of slivers, to the detriment of those who would like to operate on boxes.

Slivers are a superset of boxes. Whatever you want to do with boxes you can do with slivers. 

> 
> > Getting the document layout state and casting it to RenderBox will fail, because the layout is a RenderSliver.
> 
> Unless I'm misunderstanding, this seems like a problem for those who want to work with boxes and not slivers. Boxes make it trivial to calculate offsets and assemble visual rectangle regions, so my first thought is that we don't want to eliminate the ability to treat a document as a box, if the user is actually using it as a box.

We can expose the underlying box from the document if needed. But you can calculate those from slivers too if needed.

> 
> > ContentLayers has been modified to work similar to SliverHybridStack, where the content is a sliver while the overlays and underlays are render boxes. I think this is acceptable solution as the underlays and overlays are expected to be much sparsely populated then the content itself.
> 
> I've long planned to move `ContentLayers` into its own package because I think it's generally useful. In so doing, I'm sure people will want to use boxes as the content. Should we instead look into the options to have two different content layer implementations - one for boxes and one for slivers? We'd need to chat about the optimal API for such a thing, but I thought I'd introduce this point at the start of the review.

This would make things much more complicated. Basically to be able to use sliver as layout, every render object produced by super editor all the way down to the layout must be a RenderSliver. If you want to use both boxes and slivers throughout to super editor you'll need to duplicate all the render objects and then pass along a flag to know which one to produce.

matthew-carroll:
> Slivers are a superset of boxes. Whatever you want to do with boxes you can do with slivers.

Let's dig into this claim a bit. Certainly, from an API standpoint, this isn't the case. `RenderSliver` extends `RenderObject` and `RenderBox` extends `RenderObject`. It's not the case that `RenderSliver` extends `RenderBox`. For easy reader reference, here are some API links:

https://api.flutter.dev/flutter/rendering/RenderObject-class.html

https://api.flutter.dev/flutter/rendering/RenderBox-class.html

https://api.flutter.dev/flutter/rendering/RenderSliver-class.html

---

Looking beyond the inheritance differences, let's look at a few methods and properties. 

It's very common to want to call `localToGlobal` and `globalToLocal`. These methods don't exist on `RenderSliver`. Perhaps that's an equally simple way of achieving the same result, but I'm not sure what it is. Can you point that out?

`RenderBox` exposes its `size`, which, again, is very useful. Given the `localToGlobal` and `size`, it's trivial to find the global bounding rect for a given box. However, `RenderSliver` doesn't have a `size` property. It has a `getAbsoluteSize()` method, but that's marked "protected". `RenderSliver` has a `geometry`, but that geometry doesn't have a `size` property. Instead, the geometry reports things like `paintExtent`, `layoutExtent`, `cacheExtent`, etc. Is there a combination of these properties that would recover the expected return value of `size` on a `RenderBox`?

We've already talked a little bit about intrinsic width and height. While it may be acceptable for a specific chosen `SuperEditor` layout to use slivers, and therefore not support intrinsic measurement, I think it's reasonable that a developer might want a `SuperEditor` layout that does support intrinsic measurement. Is there a workaround for this with slivers? If the choice between boxes and slivers is provided, then this point doesn't matter. But if we're essentially migrating exclusively to slivers, then this does matter.

---

I have some concerns in other areas as well, but let's start by aligning on what the relationship really looks like between slivers and boxes. I want to make sure we have a crystal clear picture for everyone involved as to what it means to eliminate the guarantee of boxes and force people to either work with slivers, or force people to work only with generic `RenderObject`s.

matthew-carroll:
Let me also loop in some other stakeholders here for visibility. CC @Jethro87 @jmatth

knopp:
@matthew-carroll, please look at [render_object_ext](https://github.com/superlistapp/super_editor/blob/126a2131c4009d8ab4f9870eac723c257b692785/super_editor/lib/src/infrastructure/render_object_ext.dart) inside the PR. It implements `size`, `hasSize`, `globalToLocal` and `localToGlobal` on `RenderObject`, which can be either a `RenderSliver` or `RenderBox`.

knopp:
The alternative for intrinsic height is `shrinkWrap`. As is common for other scrollable widgets.

matthew-carroll:
> The alternative for intrinsic height is shrinkWrap. As is common for other scrollable widgets.

But `shrinkWrap` is a request, not a query, right? Whereas a `RenderBox` `getMaxIntrinsicWidth` is a query, rather than a request?

knopp:
> > The alternative for intrinsic height is shrinkWrap. As is common for other scrollable widgets.
> 
> But `shrinkWrap` is a request, not a query, right? Whereas a `RenderBox` `getMaxIntrinsicWidth` is a query, rather than a request?

Yes. ShrinkWrap will cause the scrollview to size itself to content, but does not allow querying intrinsic size. The question is, whether this difference is significant enough in practice to warrant supporting both slivers and boxes through-out super editor, which would increase the complexity significantly and add pretty big surface for runtime errors in cases where sliver and boxes would be mixed by mistake.

matthew-carroll:
> The question is, whether this difference is significant enough in practice to warrant supporting both slivers and boxes through-out super editor, which would increase the complexity significantly and add pretty big surface for runtime errors in cases where sliver and boxes would be mixed by mistake.

I agree that's the final question. I'm trying to start by putting all the tradeoffs on the table.

I'll take a look at your implementation of box stuff in the sliver extensions.

super_editor/lib/src/infrastructure/render_object_ext.dart
```diff
@@ -0,0 +1,55 @@
+import 'package:flutter/rendering.dart';
+
+/// Extension on [RenderObject] that that brings some of the [RenderBox] functionality down
+/// to [RenderObject] while supporting [RenderSliver].
+extension RenderObjectExt on RenderObject {
```
matthew-carroll, line 5: I'm concerned about this extension. This extension is essentially defining a union type between `RenderBox` and `RenderSliver`, but providing that as if it applies to all `RenderObject`s.

On the one hand, `RenderBox` and `RenderSliver` are the only mainstream `RenderObject` subtypes. Expecting one or the other is very likely to succeed for the vast majority of cases.

On the other hand, nothing prevents Flutter from introducing more types of `RenderObject`s, nor are individual apps precluded from doing the same. In fact, I've been thinking about creating a new type of `RenderObject` for a long time now, which would know how to layout children in lines, essentially bringing text layout into app level code, and bringing some aspects of box layout into app level code. The use of such a `RenderObject` would blow up with code that uses this extension.

What can we do to truly respect the type system here? Are these methods specifically required when talking to the `DocumentLayout`? If so, would it make  sense to declare each of these as interface properties and methods on the `DocumentLayout`, thereby allowing `RenderBox`, `RenderSliver`, and `RenderLines` implementations to each fulfill the necessary contract?

super_editor/lib/src/infrastructure/sliver_hybrid_stack.dart
```diff
@@ -0,0 +1,196 @@
+import "package:flutter/rendering.dart";
+import "package:flutter/widgets.dart";
+
+class SliverHybridStack extends MultiChildRenderObjectWidget {
```
matthew-carroll, line 4: Please add a Dart Doc that captures the purpose of this widget, and any relevant details for how to use it.

super_editor/lib/src/infrastructure/sliver_hybrid_stack.dart
```diff
@@ -0,0 +1,196 @@
...
+  const SliverHybridStack({
+    super.key,
+    super.children,
```
matthew-carroll, line 7: Please keep `child` and `children` at the bottom of the parameter list for consistency.

super_editor/lib/src/infrastructure/sliver_hybrid_stack.dart
```diff
@@ -0,0 +1,196 @@
...
+        sliver = child;
+        break;
+      }
+      child = childAfter(child);
+    }
+    if (sliver == null) {
+      geometry = SliverGeometry.zero;
+      return;
+    }
+    (sliver.parentData! as SliverLogicalParentData).layoutOffset = 0.0;
```
matthew-carroll, line 48: Can you please add some blank lines between different concerns to make this easier to read? For example, between the `sliver` guard clause and this `layoutOffset` assignment.

super_editor/lib/src/infrastructure/sliver_hybrid_stack.dart
```diff
@@ -0,0 +1,196 @@
...
+
+  bool fillViewport;
+
+  @override
+  void performLayout() {
+    RenderSliver? sliver;
+    var child = firstChild;
+    while (child != null) {
+      if (child is RenderSliver) {
+        assert(sliver == null, "There can only be one sliver in a SliverHybridStack");
```
matthew-carroll, line 38: This would be a good thing to explain in the Dart Doc for the widget.

super_editor/lib/src/super_reader/read_only_document_ios_touch_interactor.dart
```diff
@@ -1061,10 +1060,15 @@ class SuperReaderIosToolbarOverlayManagerState extends State<SuperReaderIosToolb
 
   @override
   Widget build(BuildContext context) {
-    return OverlayPortal(
-      controller: _overlayPortalController,
-      overlayChildBuilder: _buildToolbar,
-      child: widget.child ?? const SizedBox(),
+    return SliverHybridStack(
```
matthew-carroll, line 1063: Surprise sliver reference.

super_editor/lib/src/super_reader/read_only_document_ios_touch_interactor.dart
```diff
@@ -1115,10 +1119,15 @@ class SuperReaderIosMagnifierOverlayManagerState extends State<SuperReaderIosMag
 
   @override
   Widget build(BuildContext context) {
-    return OverlayPortal(
-      controller: _overlayPortalController,
-      overlayChildBuilder: _buildMagnifier,
-      child: widget.child ?? const SizedBox(),
+    return SliverHybridStack(
```
matthew-carroll, line 1122: Surprise sliver reference.

super_editor/lib/src/super_reader/super_reader.dart
```diff
@@ -204,6 +205,10 @@ class SuperReader extends StatefulWidget {
   /// debugging.
   final DebugPaintConfig debugPaint;
 
+  /// Whether the scroll view used by the reader should shrink-wrap its contents.
+  /// Only used when reader is not inside an scrollable.
+  final bool shrinkWrap;
```
matthew-carroll, line 210: Same comments here as the related ones in `SuperEditor`.

super_editor/test/infrastructure/content_layers_test.dart
```diff
@@ -33,7 +37,9 @@ void main() {
       await _pumpScaffold(
         tester,
         child: ContentLayers(
-          content: (_) => const SizedBox.expand(),
+          content: (_) => SliverToBoxAdapter(
```
matthew-carroll, line 40: I'd like to note that slivers have clearly impacted the natural usage of `ContentLayers` from an API standpoint. The fact that an adapter must be added to every single use of `ContentLayers` in this test indicates a mismatch between the generic nature of `ContentLayers` and the changes to it in this PR.

I don't know what the solution is yet. I'm still in the middle of reviewing the PR. But from an API use perspective this is definitely a regrettable change.

super_editor/test/infrastructure/content_layers_test.dart
```diff
@@ -13,15 +14,18 @@ void main() {
...
-              return const SizedBox.expand();
-            },
+          content: (_) => SliverToBoxAdapter(
+            child: LayoutBuilder(
+              builder: (context, constraints) {
+                // The content should be able to take up whatever width it wants, within the available space.
+                // The height is infinite because `ContentLayers` is a sliver.
+                expect(constraints.isTight, isFalse);
+                expect(constraints.maxWidth, _windowSize.width);
+                expect(constraints.maxHeight, double.infinity);
```
matthew-carroll, line 24: This seems to remove a critical, intended feature of `ContentLayers`. Getting the width of the document is generally trivial - so retaining width lookup but not height lookup isn't particularly valuable. 

If a user wants to provide a layer that positions a widget at the very bottom of the document, that use-case should be supported by this, just as easily as the user can position a widget at the very top of the document, or the left side, or the right side.

super_editor/test/infrastructure/content_layers_test.dart
```diff
@@ -320,18 +340,23 @@ void main() {
...
-                    expect(contentBox, isNotNull);
-                    expect(contentBox!.hasSize, isTrue);
-                    expect(contentBox.localToGlobal(Offset.zero), isNotNull);
+                    final contentSliver = contentKey.currentContext!.findRenderObject() as RenderSliver?;
+                    expect(contentSliver, isNotNull);
+                    expect(contentSliver!.geometry, isNotNull);
+                    final viewport = context.findAncestorRenderObjectOfType<RenderViewport>();
+                    // Build happens during viewport layout, which is not finished at this point. So transform to viewport
+                    // coordinate space is as far as we can go.
+                    expect(contentSliver.localToGlobal(Offset.zero, ancestor: viewport), isNotNull);
```
matthew-carroll, line 359: Here again is another change to the features of `ContentLayers`. Previously, it looks like we could actually map to global coordinates. Now we can't.

This is at least the 2nd feature to be restricted by this PR. So we'll need to look very carefully at what's going on here. If the behaviors at this point between boxes and slivers truly aren't interchangeable, then the case for different implementations of `ContentLayers` is looking pretty strong.

super_editor/test/super_editor/supereditor_scrolling_test.dart
```diff
@@ -651,7 +652,7 @@ void main() {
 
       await tester //
           .createDocument()
-          .withSingleParagraph()
+          .withLongDoc()
```
matthew-carroll, line 655: Why did this need to be changed?

super_editor/test/super_editor/supereditor_scrolling_test.dart
```diff
@@ -865,7 +866,7 @@ void main() {
 
         // Drag an arbitrary amount of pixels from the top of the editor.
         final dragGesture = await tester.dragByFrameCount(
-          startLocation: tester.getRect(find.byType(SuperEditor)).topCenter + const Offset(0, 5),
+          startLocation: tester.getRect(find.byType(CustomScrollView)).topCenter + const Offset(0, 5),
```
matthew-carroll, line 869: Does this mean we can no longer refer to the `SuperEditor` widget when basing coordinates on what the developer thinks of as Super Editor? I don't want developers to have to know that every a `CustomScrollView` reference appears in a test suite it's actually a proxy statement for "the bounds of SuperEditor". That's the kind of implementation knowledge that will bread confusion over time.

super_editor/test/super_editor/supereditor_selection_test.dart
```diff
@@ -57,7 +57,7 @@ void main() {
       // directions: right-to-left is upstream for a single line, and up-to-down is
       // downstream for multi-node. This test ensures that the single-line direction is
       // honored by the document layout, rather than the more common multi-node calculation.
-      final selection = layout.getDocumentSelectionInRegion(const Offset(200, 35), const Offset(150, 45));
+      final selection = layout.getDocumentSelectionInRegion(const Offset(1100, 35), const Offset(1050, 45));
```
matthew-carroll, line 60: Why were changes needed in this file?

super_editor/test/super_reader/super_reader_scrolling_test.dart
```diff
@@ -60,6 +60,7 @@ void main() {
 
       // Jump to the end of the document
       scrollController.jumpTo(scrollController.position.maxScrollExtent);
+      await tester.pump();
```
matthew-carroll, line 63: Please document why this is needed.

super_editor/test/super_reader/super_reader_scrolling_test.dart
```diff
@@ -535,7 +536,7 @@ void main() {
 
         // Drag an arbitrary amount of pixels from the top of the reader.
         final dragGesture = await tester.dragByFrameCount(
-          startLocation: tester.getRect(find.byType(SuperReader)).topCenter + const Offset(0, 5),
+          startLocation: tester.getRect(find.byType(Viewport)).topCenter + const Offset(0, 5),
```
matthew-carroll, line 539: Same comment here as in another test suite - the goal of the line is obscured by requiring knowledge of internal widget structure. When reading this, the reader will likely wonder "what Viewport?". We want to avoid those kinds of confusions wherever possible.

super_editor/test/super_reader/super_reader_selection_test.dart
```diff
@@ -25,7 +25,7 @@ void main() {
       // directions: right-to-left is upstream for a single line, and up-to-down is
       // downstream for multi-node. This test ensures that the single-line direction is
       // honored by the document layout, rather than the more common multi-node calculation.
-      final selection = layout.getDocumentSelectionInRegion(const Offset(200, 35), const Offset(150, 45));
+      final selection = layout.getDocumentSelectionInRegion(const Offset(1100, 35), const Offset(1050, 45));
```
matthew-carroll, line 28: Why were these changes needed?

super_editor/lib/src/infrastructure/documents/document_scaffold.dart
```diff
@@ -91,66 +104,54 @@ class _DocumentScaffoldState extends State<DocumentScaffold> {
...
-        scroller: widget.scroller,
-        showDebugPaint: widget.debugPaint.scrolling,
-        child: child,
-      ),
+    return DocumentScrollable(
+      autoScroller: widget.autoScrollController,
+      scrollController: widget.scrollController,
+      scrollingMinimapId: widget.debugPaint.scrollingMinimapId,
+      scroller: widget.scroller,
+      showDebugPaint: widget.debugPaint.scrolling,
```
matthew-carroll, line 112: Please keep debug paint properties as the final property above `child` or `children` because it's the least relevant when reading the declaration.

super_editor/lib/src/infrastructure/documents/document_scaffold.dart
```diff
@@ -91,66 +104,54 @@ class _DocumentScaffoldState extends State<DocumentScaffold> {
...
-            child: widget.gestureBuilder(context),
-          ),
-          child,
-        ],
-      ),
+    final ancestorScrollable = context.findAncestorScrollableWithVerticalScroll;
+    return SliverHybridStack(
+      // Ensure that gesture object fill entire viewport when not being
+      // in user specified scrollable.
+      fillViewport: ancestorScrollable == null,
```
matthew-carroll, line 128: Is the gesture system now inside the viewport? If so, what does that mean for offsets reported by the gesture system? Were they reported in viewport space before, and they're reported in document space now? Or does this PR retain the same coordinate space for gestures that are reported from `widget.gestureBuilder()`?

super_editor/lib/src/infrastructure/documents/document_scaffold.dart
```diff
@@ -91,66 +104,54 @@ class _DocumentScaffoldState extends State<DocumentScaffold> {
...
-            key: widget.documentLayoutKey,
-            presenter: widget.presenter,
-            componentBuilders: widget.componentBuilders,
-            onBuildScheduled: onBuildScheduled,
-            showDebugPaint: widget.debugPaint.layout,
-          ),
-          underlays: widget.underlays,
-          overlays: widget.overlays,
-        ),
+    return ContentLayers(
```
matthew-carroll, line 145: Why was the `documentLayoutLink` removed?

super_editor/lib/src/infrastructure/documents/document_scaffold.dart
```diff
@@ -38,6 +41,9 @@ class DocumentScaffold<ContextType> extends StatefulWidget {
   /// beneath the document, at the same size as the viewport.
   final WidgetBuilder gestureBuilder;
 
+  /// Builds platform specific viewport decoration (such as toolbar overlay manager or magnifier overlay manager).
+  final Widget Function(BuildContext context, {required Widget child}) viewportDecorationBuilder;
```
matthew-carroll, line 45: Why was this added? WRT to a toolbar, for example, that toolbar is likely sitting in the application `Overlay`. Even if some apps prefer the toolbar to sit only in the document, we still need to support use-cases where the toolbar can sit in the `Overlay`, thereby not getting cutoff when the toolbar gets near the edge of the document layout.

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -324,6 +326,10 @@ class SuperEditor extends StatefulWidget {
   /// debugging.
   final DebugPaintConfig debugPaint;
 
+  /// Whether the scroll view used by the editor should shrink-wrap its contents.
+  /// Only used when editor is not inside an scrollable.
+  final bool shrinkWrap;
```
matthew-carroll, line 331: Same comment here as elsewhere with regard to the naming of `shrinkWrape`, and the Dart Doc intro sentence style.

super_editor/lib/src/default_editor/layout_single_column/_layout.dart
```diff
@@ -73,6 +73,10 @@ class _SingleColumnDocumentLayoutState extends State<SingleColumnDocumentLayout>
 
   late SingleColumnLayoutPresenterChangeListener _presenterListener;
 
+  // The key for the renderBox that contains the actual document layout.
```
matthew-carroll, line 76: Please specify what "actual document layout" means, and please specify why we need to attach a key.

super_editor/lib/src/default_editor/layout_single_column/_layout.dart
```diff
@@ -73,6 +73,10 @@ class _SingleColumnDocumentLayoutState extends State<SingleColumnDocumentLayout>
...
+  final GlobalKey _boxKey = GlobalKey();
+  BuildContext get boxContext => _boxKey.currentContext!;
```
matthew-carroll, line 78: Please flip the order, so the public accessor is above the private property.

Also, is there a better name than `box`? That won't mean much to people who are reading the use of this key and/or the context throughout the remainder of this code. It begs the question, which box?

super_editor/lib/src/default_editor/layout_single_column/_layout.dart
```diff
@@ -706,12 +710,15 @@ class _SingleColumnDocumentLayoutState extends State<SingleColumnDocumentLayout>
   @override
   Widget build(BuildContext context) {
     editorLayoutLog.fine("Building document layout");
-    final result = Padding(
-      padding: widget.presenter.viewModel.padding,
-      child: Column(
-        mainAxisSize: MainAxisSize.min,
-        crossAxisAlignment: CrossAxisAlignment.center,
-        children: _buildDocComponents(),
+    final result = SliverToBoxAdapter(
```
matthew-carroll, line 713: The `SliverToBoxAdapter` prevents this widget from ever being used outside a `CustomScrollView`, right? If so, I think that's a restriction that's too narrow. 

This layout is one of the pieces of `SuperEditor` that may very well be used in the future by apps to essentially compose their own version of the `SuperEditor` widget. That custom editor may not use or want a `CustomScrollView`. Nor is this widget doing anything that's inherently related to scrolling.

super_editor/lib/src/infrastructure/content_layers.dart
```diff
@@ -328,26 +328,23 @@ class ContentLayersElement extends RenderObjectElement {
...
     assert(oldSlot != null);
     assert(newSlot != null);
     assert(_isContentLayersSlot(oldSlot!), "Invalid ContentLayers slot: $oldSlot");
     assert(_isContentLayersSlot(newSlot!), "Invalid ContentLayers slot: $newSlot");
 
-    if (oldSlot == _contentSlot) {
-      renderObject.moveChildContentToLayer(child as RenderBox, newSlot!);
-    } else if (newSlot == _contentSlot) {
-      renderObject.moveChildFromLayerToContent(child as RenderBox, oldSlot!);
+    // Can't move renderBox children to and from content slot (which is a sliver)
```
matthew-carroll, line 345: What are the implications of this change as compared to the behaviors of the original implementation?

super_editor/lib/src/infrastructure/content_layers.dart
```diff
@@ -632,13 +603,23 @@ class RenderContentLayers extends RenderBox {
 
     contentLayersLog.fine("Laying out layers (${_underlays.length} underlays, ${_overlays.length} overlays)");
     // Layout the layers below and above the content.
-    final layerConstraints = BoxConstraints.tight(size);
+    final layerConstraints = ScrollingBoxConstraints(
+      minWidth: constraints.crossAxisExtent,
+      maxWidth: constraints.crossAxisExtent,
+      minHeight: sliverLayoutGeometry.scrollExtent,
+      maxHeight: sliverLayoutGeometry.scrollExtent,
```
matthew-carroll, line 610: Are there any implications with this definition of constraints in the case that `SuperEditor` is used with an ancestor scrollable, in which there's additional scrollable content above and/or below the editor?

super_editor/lib/src/infrastructure/content_layers.dart
```diff
@@ -687,19 +676,44 @@ class RenderContentLayers extends RenderBox {
       return;
     }
 
+    void paintChild(RenderObject child) {
```
matthew-carroll, line 679: Can you do this without defining a method inside of a method?

super_editor/lib/src/infrastructure/content_layers.dart
```diff
@@ -687,19 +676,44 @@ class RenderContentLayers extends RenderBox {
...
 
     // Third, paint the overlays.
     for (final overlay in _overlays) {
-      context.paintChild(overlay, offset);
+      paintChild(overlay);
     }
   }
+
+  @override
+  void applyPaintTransform(covariant RenderObject child, Matrix4 transform) {
```
matthew-carroll, line 702: Given how convoluted render objects are, can you please document each of the standard behaviors that you've added and adjusted in this PR? I'd like to make sure that every Super Editor contributor is able to understand the intended behavior of these overrides.

super_editor/lib/src/infrastructure/content_layers.dart
```diff
@@ -924,7 +938,7 @@ abstract class ContentLayerState<WidgetType extends ContentLayerStatefulWidget,
   @override
   Widget build(BuildContext context) {
     final contentLayers = (context as Element).findAncestorContentLayers();
-    final contentElement = contentLayers?._content;
+    var contentElement = contentLayers?._content;
```
matthew-carroll, line 941: Why is this a `var`?

super_editor/example/lib/demos/interaction_spot_checks/toolbar_following_content_in_layer.dart
```diff
@@ -40,49 +40,52 @@ class _ToolbarFollowingContentInLayerState extends State<ToolbarFollowingContent
...
-                  ),
-                  const SizedBox(height: 96),
-                  TextButton(
-                    onPressed: () {
-                      _expansionExtent.value = Random().nextDouble() * 200;
-                    },
-                    child: Text("Change Size"),
-                  ),
-                  const Spacer(),
+          child: CustomScrollView(
```
matthew-carroll, line 43: Why was a `CustomScrollView` added here?

matthew-carroll (commented):
I did an initial pass of the PR. Overall I'd say it's a good sign that relatively little code was needed to migrate `SuperEditor` and `SuperReader` behavior to a sliver world.

I do have concerns about some places where this migration has narrowed the composability and/or features of existing artifacts, including `ContentLayers` and the `SingleColumnDocumentLayout`. I'm also concerned about the `RenderObject` extension that bridges between `RenderSliver` and `RenderBox`.

super_editor/example/lib/demos/interaction_spot_checks/toolbar_following_content_in_layer.dart
knopp: Becuase ContentLayers is a Sliver and as such it needs to be put in a CustomScrollView.

super_editor/lib/src/default_editor/document_hardware_keyboard/document_physical_keyboard.dart
```diff
@@ -103,6 +103,7 @@ class _SuperEditorHardwareKeyHandlerState extends State<SuperEditorHardwareKeyHa
       focusNode: _focusNode,
       onKeyEvent: widget.keyboardActions.isEmpty ? null : _onKeyPressed,
       autofocus: widget.autofocus,
+      includeSemantics: false,
```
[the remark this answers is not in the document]
knopp: Sure.

super_editor/lib/src/infrastructure/flutter/build_context.dart
```diff
@@ -18,4 +19,20 @@ extension ScrollableFinder on BuildContext {
 
     return ancestorScrollable;
   }
+
+  /// Returns the RenderBox of the nearest ancestor [RenderAbstractViewport].
+  RenderBox findViewportBox() {
+    // findAncestorRenderObjectOfType traverses the element tree, which is
```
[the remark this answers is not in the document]
knopp: Not really. The code in `findAncestorRenderObjectOfType` seems very old. I don't know what the trade-offs were back then.

super_editor/lib/src/infrastructure/platforms/ios/ios_document_controls.dart
```diff
@@ -136,17 +136,7 @@ class IosDocumentGestureEditingController extends GestureEditingController {
...
-  /// Layer link that's aligned to the top-left corner of the document layout.
-  ///
-  /// Some of the offsets reported by this controller are based on the
-  /// document layout coordinate space. Therefore, to honor those offsets on
-  /// the screen, this `LayerLink` should be used to align the controls with
-  /// the document layout before applying the offset that sits within the
-  /// document layout.
-  LayerLink get documentLayoutLink => _documentLayoutLink;
-  final LayerLink _documentLayoutLink;
+  });
```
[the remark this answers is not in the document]
knopp: Was it used for anything? I couldn't find any usage of it, either in tests or examples. There have significant changes to document layout and overlays in the past and this not being used anywhere anymore it seemed vestigial.

super_editor/lib/src/infrastructure/render_object_ext.dart
knopp: I think I agree that this smells a bit. I think at any point we should be able to cast to ConcreteInstance and this category should only be on `RenderSliver`. I will rework this.

super_editor/test/infrastructure/content_layers_test.dart
knopp: You always have access to DocumentLayout if you need to query the document height. These constraints are provided to the document layout itself, not to individual layers.

super_editor/test/infrastructure/content_layers_test.dart
knopp: The change here is because in the original test, the ContentLayers widget was not being put into a scrollable (it was a box). 

If you have a real world use case of ContentLayers widget that is not in scrollable then I'd suggest making a separate content layer implementation that is a box. 

But in actual super editor usage, unlike this test, even before this PR, the ContentLayer widget is always placed into scrollable so you'd need to address exact same situation (global offset not available during build) even prior this PR.

super_editor/test/super_editor/supereditor_scrolling_test.dart
knopp: My bad. Checked that in by mistake. Will revert it.

super_editor/test/super_editor/supereditor_scrolling_test.dart
knopp: This is limitation of  widget tester - it cannot find a rect of `Sliver` (which SuperEditor is now). There's nothing preventing WidgetTester from doing it, it's just that nobody bothered implementing it. I can try to get this fixed in Flutter.

super_editor/test/super_editor/supereditor_selection_test.dart
knopp: The change is needed because the layout now covers the width of entire editor. The horizontal centering is now done in within the layout itself, as opposed to above layout. It is not possible to do horizontal centring of entire document above layout, since the layout itself is a sliver and it always fills the full cross axis extent.

super_editor/test/super_reader/super_reader_scrolling_test.dart
knopp: I don't quite understand why this is necessary. Every time the scroll position changes `pump()` is necessary to recalculate layout and possibly rebuild content layers. In every other test there is `pump()` or `pumpAndSettle()` after `scrollPosition.jump()`, the fact the the test passed without pump was an accident.

super_editor/test/super_reader/super_reader_selection_test.dart
knopp: Explained here: https://github.com/superlistapp/super_editor/pull/1916#discussion_r1565426702

super_editor/lib/src/infrastructure/documents/document_scaffold.dart
knopp: The gesture system always reported in document space and that behavior was preserved in the PR. What this line does is to make sure that if the document is too short and does not fill entire viewport, the gesture system gets stretched to fill the entire viewport, preserving the original behavior enforced through ViewportBoundsReporter and ViewportBoundsReplicator.

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
```diff
@@ -1529,10 +1528,15 @@ class SuperEditorAndroidControlsOverlayManagerState extends State<SuperEditorAnd
 
   @override
   Widget build(BuildContext context) {
-    return OverlayPortal(
-      controller: _overlayController,
-      overlayChildBuilder: _buildOverlay,
-      child: widget.child ?? const SizedBox(),
+    return SliverHybridStack(
```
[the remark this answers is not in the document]
knopp: The fact that SuperEditor insist on bringing it own scrolling infrastructure is the surprising part to me. This seems like a rather peculiar decision and results in a rather magical behavior where SuperEditor builds differently when placed into a scrollable or not.

If SuperEditor did not bring it's own scrollable, then conceptually every layer of it would need to be sliver, which would basically mean replacing all Stack instances with SliverHybridStack and making ContentLayers a sliver.

However because super editor does bring it's own scrollable sometimes, the PR made some adjustment to ensure that the scrollable is at the top level, and all layers are within the scrollable. There are no layers of SuperEditor anymore on top of scrollable. 

So the only difference is that when not put into a scrollable, SuperEditor wraps itself in a CustomScrollView. All the other layers remain same regardless of whether put in a scrollable or not. This also means that each layer of SuperEditor is a sliver now. I'm hoping that explains all the "Surprise" sliver references. Yes, this PR turns SuperEditor into a sliver, which means every layer of it down to the layout is a Sliver.

Turning SuperEditor into a sliver does make things slightly more complicated, but I don't see what other way there is to achieve acceptable performance for large documents.

super_editor/lib/src/default_editor/document_scrollable.dart
```diff
@@ -378,7 +387,12 @@ class AutoScrollController with ChangeNotifier {
     // The scroll position changed. Probably because the position scrolled
     // up or down. Notify our listeners so that they can adjust the document
     // selection bounds or other related properties.
-    notifyListeners();
+    // The scroll position may trigger layout changes.
```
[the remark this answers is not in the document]
knopp: I updated the comment slightly. Scroll position change causes the layout to change, calling the listeners too soon causes issues when converting from / to global coordinates. This for example caused `deltaWhileAutoScrolling` being one frame too late. This manifested in the "auto-scrolls up" test, but instead of fixing it the test was adjusted instead to account for the wrong behavior.

super_editor/test/infrastructure/content_layers_test.dart
knopp: I mentioned this elsewhere. If somebody out there is using ContentLayers inside a box environment directly (outside of editor) then it might be worth providing a box version of it. Otherwise the ContentLayers is only created within the scope of the editor and then there is no adapter needed because the editor itself is a sliver.

super_editor/lib/src/infrastructure/content_layers.dart
knopp: The changes are that you can't turn the `content` element into and `overlay` or `underlay`. I don't know if there was use-case for this before, but now the `content` element is a sliver and overlays and underlays are render boxes so it doesn't really makes sense to do.

super_editor/lib/src/infrastructure/content_layers.dart
knopp: None I can think of.

knopp:
I tried running Superlist with this branch, and apart from making sure there are no render boxes outside of editor (which is a given) the only change I needed was to not make assumption about the layout object render type:

That meant changing code that positions overlay from
```dart
    final docBox =
        _documentLayoutKey.currentContext?.findRenderObject() as RenderBox?;
    if (docBox == null) return;

    final position = Rect.fromPoints(
      docBox.localToGlobal(selectionBoundingBox.topLeft),
      docBox.localToGlobal(selectionBoundingBox.bottomRight),
    );
```

Into

```dart
    final layout = _documentLayoutKey.currentState as DocumentLayout;
    final position = Rect.fromPoints(
      layout.getGlobalOffsetFromDocumentOffset(selectionBoundingBox.topLeft),
      layout.getGlobalOffsetFromDocumentOffset(selectionBoundingBox.bottomRight),
    );
```

super_editor/lib/src/infrastructure/render_object_ext.dart
knopp: I refactored the code a little bit. The category now only extends `RenderSliver`.

super_editor/lib/src/infrastructure/documents/document_scaffold.dart
knopp: This has been added because the Viewport decoration is a Sliver. However in some cases, `DocumentScaffold` creates the scroll view that hosts the sliver, so it  needs to be in charge of building the decoration so that it can put it into the scroll view.

If the overlay doesn't want to get clipped by viewport then it should probably use OverlayPortal rather than depending on being placed right above the document scrollable (which would only work when actually using the document scrollable).

matthew-carroll:
@knopp I think the most reasonable path forward is to just land this and see how it shakes out. If we need to bring more box stuff back in, we can.

Can you address the few minor comments that were originally made?

matthew-carroll:
@knopp ping on this

knopp:
Apologies for the delay. Right now trying to dogfooding this in superlist to figure out if there are some loose ends that need to be addressed in this PR.

super_editor/lib/src/infrastructure/flutter/build_context.dart
```diff
@@ -18,4 +19,20 @@ extension ScrollableFinder on BuildContext {
...
+    // render object tree.
+    var renderObject = findRenderObject();
+    while (renderObject != null) {
+      if (renderObject is RenderAbstractViewport) {
+        return renderObject as RenderBox;
+      }
+      renderObject = renderObject.parent;
+    }
+
+    throw StateError('No RenderAbstractViewport ancestor found');
```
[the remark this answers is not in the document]
knopp: I don't think there is any calliste right now that can under normal circumstances not be in a viewport. Unless we have one I'd prefer having the exception here instead of duplicating it several times.

knopp:
The failing test is because of https://github.com/superlistapp/super_editor/issues/2056. By a coincidence the existing tests didn't catch the regression in main, but did catch it with changes in this PR.

matthew-carroll:
@knopp should I go ahead and merge this? I think the best we can do on my side is to see what happens - I'm not sure there's much I can provide that's useful on the review front.

matthew-carroll:
I think we said on our call that the sliver padding should be reverted because it's now available on stable. After that I think we can merge in.

matthew-carroll (approved):

matthew-carroll:

KevinBrendel:
For us (in QuikFlow) this change is very problematic, because we rely on being able to compute the dry layout for sizing our nodes correctly. We have a custom layout for this using boxy. We do not use any traditional scrolling (we transform the whole canvas including all widgets).

As far as I can tell, there is no way to compute the dry layout now?

matthew-carroll:
@knopp can you address @KevinBrendel's use-case? We'll want to find a way to re-enable their use-case.

knopp:
@KevinBrendel, can you be more specific about the use-case? Do you calculate dry layout multiple times with different constraints or do you just need super_editor to take required height for given width?

Ultimately the layout here is still a `RenderBox`, so it should be possible to skip all the slivers get the dry layout from the `RenderBox`.

KevinBrendel:
@knopp Yes, we first calculate the dry layout of SuperEditor and some other widgets with one set of constraints, then, based on the results, we define new constraints and then perform a layout again.

Exactly, in our case, we would like to just skip past the slivers and get to the `RenderBox` directly. We are fine with losing all of the optimization that comes from slivers in that case. Is this already possible somehow?

knopp:
@KevinBrendel, I think something like this should work for your use case:

```dart
class FakeViewport extends SingleChildRenderObjectWidget {
  FakeViewport({required super.child});

  @override
  RenderObject createRenderObject(BuildContext context) {
    return _RenderFakeViewport();
  }
}

class _RenderFakeViewport extends RenderBox
    with RenderObjectWithChildMixin<RenderSliver>
    implements RenderAbstractViewport {
  @override
  void debugAssertDoesMeetConstraints() {}

  @override
  RevealedOffset getOffsetToReveal(RenderObject target, double alignment, {Rect? rect, Axis? axis}) {
    return RevealedOffset(offset: 0, rect: Rect.zero);
  }

  @override
  void setupParentData(RenderObject child) {}

  @override
  Rect get paintBounds => Rect.zero;

  @override
  void performLayout() {
    final childConstraints = SliverConstraints(
      axisDirection: AxisDirection.down,
      growthDirection: GrowthDirection.forward,
      userScrollDirection: ScrollDirection.forward,
      scrollOffset: 0,
      precedingScrollExtent: 0,
      overlap: 0,
      remainingPaintExtent: constraints.maxHeight,
      crossAxisExtent: constraints.maxWidth,
      crossAxisDirection: AxisDirection.right,
      viewportMainAxisExtent: constraints.maxHeight,
      remainingCacheExtent: double.infinity,
      cacheOrigin: 0,
    );
    child!.layout(childConstraints, parentUsesSize: true);
    final geometry = child!.geometry;
    size = Size(constraints.maxWidth, geometry!.scrollExtent);
  }

  RenderBox _getBox(RenderSliver sliver) {
    RenderSliver? firstSliver;
    RenderBox? firstBox;
    sliver.visitChildren((child) {
      if (child is RenderSliver && firstSliver == null) {
        firstSliver = child;
      }
      if (child is RenderBox && firstBox == null) {
        firstBox = child;
      }
    });
    return firstSliver != null ? _getBox(firstSliver!) : firstBox!;
  }

  @override
  Size computeDryLayout(covariant BoxConstraints constraints) {
    final layoutBox = _getBox(child!);
    return layoutBox.computeDryLayout(constraints);
  }

  @override
  void applyPaintTransform(RenderObject child, Matrix4 transform) {}

  @override
  bool hitTestChildren(BoxHitTestResult result, {required Offset position}) {
    return child!.hitTest(
      SliverHitTestResult.wrap(result),
      mainAxisPosition: position.dy,
      crossAxisPosition: position.dx,
    );
  }

  @override
  void paint(PaintingContext context, Offset offset) {
    context.paintChild(child!, offset);
  }

  @override
  void performResize() {}

  @override
  Rect get semanticBounds => Offset.zero & size;
}
```

Wrap `SuperEditor` inside `FakeViewport` and the render object should forward dry layout request to the actual layout render box. This assumes that there is an actual scrollable somewhere above so that super editor doesn't try to create it's own scrollable. I'd assume that you already have a scrollable in hierarchy in your case, right?

KevinBrendel:
@knopp Thanks! This definitely goes into the right direction.

We actually do not have a scrollable in our hierarchy. Our app is canvas-based and we have it all in an InteractiveViewer with caching of the widgets to keep it performant. Some of our users have hundreds of nodes, each with their own SuperEditor instance.

I had to wrap it in a SingleChildScrollView to get it to compile. This kind of works, but we lose the intrinsic width of the editor, which is necessary to size the nodes. Essentially, only fixed-width nodes are possible.

Here is a small video to better demonstrate the use-case:
https://github.com/user-attachments/assets/bcb4148e-3a82-4643-ab51-4abd5c96f987

But with this workaround, auto-width nodes seem no longer possible:
https://github.com/user-attachments/assets/1e75ae05-37e9-4958-8d75-7131dac5f50e

We just released an initial version of QuikFlow with SuperEditor integration. If you want to play around with it yourself, just tell me your OS and I can send you a free key for the Pro version.

knopp:
If you don't have scrollable in hierarchy, SuperEditor will create a single child scroll view. I think maybe you'd need a way to disable this behavior completely since you don't want to have any kind of scrolling there.

Intrinsic width is doable same way dry layout is. You just implement `computeMax/MinIntrinsicWidth`  and forward the call to the box obtained by `_getBox` (same way dryLayout does it).

KevinBrendel:
Ahh, thanks!

Indeed the behavior seems correct now that I have implemented the four intrinsic functions. Probably it would make sense to have an example somewhere with this widget for other people with similar issues.

Will have to test for a noticeable performance impact, but indeed it would be great to be able to disable the scrolling to avoid having to use the ScrollView workarounds.

PR #1917, opened 2024-04-02
Upgrade uuid package to v4.0.0 (Resolves #1899)

### LINKED ISSUE #1899: [SuperEditor] Upgrade uuid dependency
**User Behavior**
Currently SuperEditor depends on uuid 3.0.3, which was released 3 years ago. Many packages today require uuid with major version 4, resulting in dependency conflicts. Upgrading this dependency would solve this problem and make SuperEditor more compatible.

**Feature Value**
Without this, SuperEditor is only usable with dependency_overrides in many projects, making it more likely to scare users off and making it possible that dependency_overrides cause unforeseen issues, in case SuperEditor is broken by breaking changes of uuid 4. Upgrading the dependency would increase adoption of this package and reduce risk.

KevinBrendel:
Upgrade uuid dependency as per #1899. 

Should not be very risky and should be unaffected by breaking changes as the usage of UUID is very simple:

Screenshot 2024-04-02 at 18 02 20


angelosilvestre (approved):
LGTM

matthew-carroll:
@KevinBrendel we merged something that altered the pubspec so now we have conflicts here. Can you resolve those conflicts and then we'll merge this in?

KevinBrendel:
Done.

matthew-carroll:
@angelosilvestre can you cherry pick this when you get a chance?

PR #1918, opened 2024-04-02
[SuperEditor] Remove empty task on pressing enter/return

### LINKED ISSUE #1898: [SuperEditor] Empty task doesn't get removed on pressing enter/return
Empty task should be removed if enter/return key is pressed when the caret is on it. Similar to ordered and unordered lists.

#### Expected and actual behavior

https://github.com/superlistapp/super_editor/assets/92245694/4bc70193-3d9a-46ff-ae63-6d9738f89572

toseefkhan403:
Fixes #1898 
Converting task to paragraph if enter/return is pressed on an empty task (similar to ordered and unordered lists). 
If you have any suggestions, do let me know.

https://github.com/superlistapp/super_editor/assets/43390808/1140e30f-0a26-4719-8041-a12c8842f729

### DIFF
--- a/super_editor/lib/src/default_editor/tasks.dart
+++ b/super_editor/lib/src/default_editor/tasks.dart
@@ -312,12 +312,17 @@ ExecutionInstruction enterToInsertNewTask({
 
   final splitOffset = (selection.extent.nodePosition as TextNodePosition).offset;
 
-  editContext.editor.execute([
-    SplitExistingTaskRequest(
-      existingNodeId: node.id,
-      splitOffset: splitOffset,
-    ),
-  ]);
+  // If the task text is empty, convert it to paragraph.
+  if (node.text.text.isEmpty) {
+    editContext.commonOps.convertToParagraph();
+  } else {
+    editContext.editor.execute([
+      SplitExistingTaskRequest(
+        existingNodeId: node.id,
+        splitOffset: splitOffset,
+      ),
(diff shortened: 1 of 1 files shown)


super_editor/lib/src/default_editor/tasks.dart
```diff
@@ -312,12 +312,17 @@ ExecutionInstruction enterToInsertNewTask({
...
 
-  editContext.editor.execute([
-    SplitExistingTaskRequest(
-      existingNodeId: node.id,
-      splitOffset: splitOffset,
-    ),
-  ]);
+  // If the task text is empty, convert it to paragraph.
+  if (node.text.text.isEmpty) {
+    editContext.commonOps.convertToParagraph();
```
matthew-carroll, line 317: This decision should probably be made in the `SplitExistingTaskRequest`, not the keyboard handler. For example, what happens on mobile when the user presses "enter" on the software keyboard? That path goes through the IME, not the physical keyboard handler. We'll need to support both input options.

matthew-carroll (commented):
Thanks for the PR - I added a comment. Also, please be sure to add appropriate tests for both desktop and mobile.

toseefkhan403:
> @toseefkhan403 do you plan to continue with this PR?

Hey, I won't be able to work on this for at least two weeks. I'll close this PR for now.

PR #1920, opened 2024-04-03
[SuperEditor] Fix list item vertical alignment (Resolves #1834)

### LINKED ISSUE #1834: [SuperTextEditor] vertical alignment of list items incorrect when settings line height
**Package Version**
super_editor: git main branch (at the time at 091b17b20f8c80dcc5e387b166382597ef96131f)

**To Reproduce**
Display a list of items, use the stylesheet to set all nodes to a custom line height. Check the display output, the item indicators will not align with the content of the list item.

**Minimal Reproduction Code**

<details>
  <summary>Stylesheet</summary>

  ```dart
  Stylesheet(
    rules: <StyleRule>[
      StyleRule(
        BlockSelector.all,
        (doc, docNode) {
          return {
            "textStyle": TextStyle(
              height: 3.0, // <- extreme case to show the issue, but even setting it to 1.25 is already enough
              //leadingDistribution: TextLeadingDistribution.proportional, // <- setting this fixes it for the ordered list item but not for the unordered item
            ),
          };
        },
      ),
  ])
  ```
</details>

**Actual behavior**
List numbers and bullets do not align with list item text

**Expected behavior**
List indicators should align with the text in the list item

**Platform**
any

**Flutter version**
Flutter 3.16.9 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 41456452f2 (11 days ago) • 2024-01-25 10:06:23 -0800
Engine • revision f40e976bed
Tools • Dart 3.2.6 • DevTools 2.28.5

**Screenshots**
image

**Additional context**
While the alignment is incorrect for both ordered and unordered items, the actual cause seems different. I looked into it a bit and found the following causes:

*The problem for the unordered list item seems to be,* that the Text() Widget in _defaultOrderedListItemNumeralBuilder
https://github.com/superlistapp/super_editor/blob/091b17b20f8c80dcc5e387b166382597ef96131f/super_editor/lib/src/default_editor/list_items.dart#L504
defaults to ```TestStyle(leadingDistribution: TextLeadingDistribution.even)``` while the TextComponent at
https://github.com/superlistapp/super_editor/blob/091b17b20f8c80dcc5e387b166382597ef96131f/super_editor/lib/src/default_editor/list_items.dart#L468
defaults to ```TestStyle(leadingDistribution: TextLeadingDistribution.even)```. Explicitly setting the ```leadingDistribution``` in the stylesheet bypasses the issue, as this value is copied in both places.

*The problem for the unordered list item seems to be, that simply centering the dot widget is just not enough. There should be some consideration of the baseline. I guess something with a ```Baseline``` or ````RichText / WidgetSpan``` would do the trick.
https://github.com/superlistapp/super_editor/blob/091b17b20f8c80dcc5e387b166382597ef96131f/super_editor/lib/src/default_editor/list_items.dart#L374

And here is a fix for the unordered list item (my version, maybe ugly but effective): https://github.com/quaaantumdev/super_editor/commit/e55111d97432aa4cbf91e4072dd0a8ef04a1aa82

angelosilvestre:
[SuperEditor] Fix list item vertical alignment. Resolves #1834

The dot in unordered list items is a bit misaligned when using a line height multiplier greater than `1.0`.  Here's one example with a `3.0` multiplier:

SCR-20240402-slvd

The misalignment is worse when using different font sizes (this also affects ordered list items)

SCR-20240402-smfh

This PR changes the list component to place the dot inside a `WidgetSpan` to properly align it. Also, the list components were changed use the attributions of the first character. This change was made because the list item might contain a `FontSizeAttribution`, which will affect the line height. 

This was the result:

SCR-20240402-smzu

The alignment change caused some golden tests to fail, because now the dot is slightly above from where it was before.

We are still rendering differently from Google Docs.  It scales the dot and doesn't seem to indent based on the font size:

SCR-20240402-spbt

Avoiding scaling the dot was intentional, due to previous conversations. We could change that if needed.


matthew-carroll:
@angelosilvestre is there a reason why we'd want different indentation based on font size? That sounds like a bug to me.

super_editor/lib/src/default_editor/list_items.dart
```diff
@@ -368,16 +371,38 @@ class _UnorderedListItemComponentState extends State<UnorderedListItemComponent>
...
     alignment: Alignment.centerRight,
-    child: Container(
-      width: 4,
-      height: 4,
-      margin: const EdgeInsets.only(right: 10),
-      decoration: BoxDecoration(
-        shape: BoxShape.circle,
-        color: component.styleBuilder({}).color,
+    child: Text.rich(
+      // Place a zero-width joiner before the bullet point to make it properly aligned. Without this,
```
matthew-carroll, line 383: Can you point out what specifically happens if we don't include this zero-width character? Also, should we move this comment immediately above the `text:` line? Because right now this comment appears to apply to the whole span, which includes the dot.

angelosilvestre: Without the zero-width character the dot appears misaligned for small font sizes:

Without the zero-width character:

SCR-20240409-quir

With the zero-width character:

SCR-20240409-qumu

angelosilvestre:
> @angelosilvestre is there a reason why we'd want different indentation based on font size? That sounds like a bug to me.

@matthew-carroll  This seems to be done on purpose. This is the method that computes then indentation:

```dart
double _defaultIndentCalculator(TextStyle textStyle, int indent) {
  return (textStyle.fontSize! * 0.60) * 4 * (indent + 1);
}
```

matthew-carroll:
@angelosilvestre it seems likely that that behavior was designed while thinking about a list where all list items are the same size. If all list items have really big text, then it indents more, if all list items have tiny text, it indents less.

I doubt this work grappled with varying text sizes. It's obvious from your screenshots that the result of displaying list items at varying text sizes looks horrible and is unexpected. So we need to fix that.

Seems like we can either make it constant, or we'll need to search up and down the list to figure out the largest font size and base all list item indentations on that font size.

angelosilvestre:
@matthew-carroll Based on the google docs example that I posted before I would say that we should be fine using a constant indentation.

SCR-20240402-spbt

matthew-carroll:
Ok, then let's do that.

angelosilvestre:
@matthew-carroll For ordered lists I'm not sure how we should compute the indentation. Here's an example from Google Docs:

SCR-20240410-szwm

We can see that the numeral is left-aligned, but the distance between the numeral and the content isn't fixed.

matthew-carroll:
Hm. That's pretty weird. Can you post a few screenshots from a few other apps so we can compare?

angelosilvestre:
## Apple Notes

SCR-20240411-rmpg

The numerals look right-aligned, but it's also weird with bigger font sizes.  Also, with two digits numerals it looks even more weird:

SCR-20240411-rnhm

## Microsoft Word online

SCR-20240411-rpxm

Numerals are left-aligned and the indentation increases on bigger font sizes

## Evernote

SCR-20240411-rrka

Doesn't scale the numeral.

## Canva Docs

SCR-20240411-rxec

Numerals are left-aligned and the indentation increases on bigger font sizes

## Other apps

Notion, Obsidian and ClickUp don't look like they support setting font sizes.

matthew-carroll:
Looking at those examples, I think right aligning the numerals looks best. The problem with the Apple version is that they allow different widths for the numerals. What if we make all the numerals in the same list take up the same width? 

To accomplish that, we could have each list item take its font, font size, font style, and then immediately measure every numeral in the font: 0-9. That list item takes the widest value and multiplies it by the number of numerals used by that list item. Then all list items respect the widest value. We can cache the measurement so that the same font, at the same size, with the same style never needs to re-measure. That would work fine most lists.

What do you think?

angelosilvestre:
@matthew-carroll Can we a separate PR's to avoid the fix for the vertical alignment with the horizontal alignment change? 

Or do you prefer to keep all on the same PR?

matthew-carroll:
@angelosilvestre we can do them separately. Please file the appropriate issue and let me know when you think this PR has what it should.

angelosilvestre:
@matthew-carroll I filed https://github.com/superlistapp/super_editor/issues/1950

Could you please re-review this PR?

super_editor/lib/src/default_editor/list_items.dart
```diff
@@ -368,16 +371,38 @@ class _UnorderedListItemComponentState extends State<UnorderedListItemComponent>
...
-    child: Container(
-      width: 4,
-      height: 4,
-      margin: const EdgeInsets.only(right: 10),
-      decoration: BoxDecoration(
-        shape: BoxShape.circle,
-        color: component.styleBuilder({}).color,
+    child: Text.rich(
+      TextSpan(
+        // Place a zero-width joiner before the bullet point to make it properly aligned. Without this,
```
matthew-carroll, line 384: I won't hold up on the PR on this, but can you file an issue with Flutter including repro steps to show mis-alignment in the absence of the spacer? No need to include attributed text in that - just use direct Flutter stuff that demonstrates both options.

matthew-carroll (approved):
LGTM

matthew-carroll:
@angelosilvestre it looks like the cherry pick failed here. Can you do it manually?

PR #1931, opened 2024-04-12
[SuperEditor] Avoid throwing UnimplementedError in currentAutofillScope (Resolves #1923)

### LINKED ISSUE #1923: `DocumentImeInputClient` occasionally throws `UnsupportedError`
**Description**
Occasionally, I hit an `UnsupportedError` from `DocumentImeInputClient`. The error is thrown here: https://github.com/superlistapp/super_editor/blob/98762748b1a666d81c2a48f0a8a5ee6927a2f963/super_editor/lib/src/default_editor/document_ime/document_ime_communication.dart#L163

I'm not sure if this throw is intentional or if the method should return null.

**Package Version**
`github stable`

**To Reproduce**
Cannot reliable reproduce as I don't know exactly what causes Flutter to call this particular code-path.

**Platform**
iOS simulator

**Flutter version**
`Flutter (Channel stable, 3.19.5, on macOS 14.4 23E214 darwin-arm64)`

angelosilvestre:
[SuperEditor] Avoid throwing UnimplementedError in currentAutofillScope. Resolves #1923

In `DocumentImeInputClient` we have this override throwing an `UnimplementedError`:

https://github.com/superlistapp/super_editor/blob/98762748b1a666d81c2a48f0a8a5ee6927a2f963/super_editor/lib/src/default_editor/document_ime/document_ime_communication.dart#L163

This doesn't look like it's intentional. It's most likely that this is an auto-generated code. The docs for `currentAutofillScope` describe that we should return `null` if we don't support auto-fill. In `SuperTextField` we already do this.

This PR changes this method to return `null`.

### DIFF
--- a/super_editor/lib/src/default_editor/document_ime/document_ime_communication.dart
+++ b/super_editor/lib/src/default_editor/document_ime/document_ime_communication.dart
@@ -160,7 +160,7 @@ class DocumentImeInputClient extends TextInputConnectionDecorator with TextInput
   }
 
   @override
-  AutofillScope? get currentAutofillScope => throw UnimplementedError();
+  AutofillScope? get currentAutofillScope => null;
 
   @override
   TextEditingValue get currentTextEditingValue => _currentTextEditingValue;


matthew-carroll (approved):
LGTM

PR #1939, opened 2024-04-14
[SuperEditor][Web] Fix entering spaces inside a Scrollable (Resolves #1784)

### LINKED ISSUE #1784: [SuperEditor] Can't type a space character on web when SuperEditor is the child of a SingleChildScrollView
**Package Version**
super_editor, git main branch, `304e778`

**To Reproduce**
Wrap the SuperEditor in a SingleChildScrollView in demo_empty_document (see below), run in chrome and attempt to type spaces into the editor.

**Minimal Reproduction Code**
  ```dart
--- a/super_editor/example/lib/demos/demo_empty_document.dart
+++ b/super_editor/example/lib/demos/demo_empty_document.dart
@@ -35,13 +35,13 @@ class _EmptyDocumentDemoState extends State<EmptyDocumentDemo> {
   @override
     return SafeArea(
-      child: SuperEditor(
+      child: SingleChildScrollView(child: SuperEditor(
         editor: _docEditor,
         document: _doc,
         composer: _composer,
         gestureMode: DocumentGestureMode.mouse,
       ),
-    );
+    ));
   }
 }
  ```

**Actual behavior**
Normal characters appear fine as you type, but hitting spacebar results in nothing happening. Itlookslikethis.

**Platform**
Chrome and Firefox both have this issue. Works fine in Android and as a desktop Linux app. Haven't tested other platforms.

**Flutter version**
Flutter 3.19.0-4.0.pre.36 • channel master

**Additional context**
This looks similar to [this issue](https://github.com/superlistapp/super_editor/issues/1446) and [this corresponding PR](https://github.com/superlistapp/super_editor/pull/1455). Could be an edge case that wasn't handled, or just a regression.

It seems like other odd things are happening around this, but I haven't really nailed them down. I can't seem to brink up the developer tools with ctl-shift-i when the editor is focused, for example.

angelosilvestre:
[SuperEditor][Web] Fix entering spaces inside a Scrollable. Resolves #1784

On Windows web, users can't type a space character if the editor is inside a `Scrollable`.

The cause is that Flutter installs some shortcuts by default, including using the space key to scroll the page. On macOS, we already stop this `Intent` together with others that prevent the OS from generating selectors.

This PR adds another map of disabled intents that should be used in non-apple platforms. Currently, only the space shortcut is included.

To be able to test this and to avoid regressions, I extracted the Flutter-installed shortcuts to a file within `super_editor`. That way, we can force the tests to use the web shortcuts when needed. The downside of this is that we'll need to update our list whenever Flutter adds new default shortcuts, although this should happen very often.


super_editor/lib/src/default_editor/document_ime/supereditor_ime_interactor.dart
```diff
@@ -478,7 +478,7 @@ class SuperEditorImeInteractorState extends State<SuperEditorImeInteractor> impl
     return SuperEditorImeDebugVisuals(
       imeConnection: _imeConnection,
       child: Actions(
-        actions: defaultTargetPlatform == TargetPlatform.macOS ? disabledMacIntents : {},
+        actions: CurrentPlatform.isApple ? disabledMacIntents : disabledWindowsIntents,
```
matthew-carroll, line 481: What about Linux? Web?

super_editor/lib/src/infrastructure/actions.dart
```diff
@@ -0,0 +1,172 @@
+import 'package:flutter/foundation.dart';
+import 'package:flutter/services.dart';
+import 'package:flutter/widgets.dart';
+import 'package:super_editor/src/infrastructure/platforms/platform.dart';
+
+/// Prevents a [PrioritizedIntents] from bubbling up if [intentFilter] returns
```
matthew-carroll, line 6: Please start by describing the motivation, rather than the technical details. The reader probably doesn't know what a `PrioritizedIntents` is, or why this might be done, and therefore won't be able to figure out what to provide in the `intentFilter`.

super_editor/lib/src/infrastructure/actions.dart
```diff
@@ -0,0 +1,172 @@
...
+
+  bool _isActionEnabled(Action action, Intent intent, BuildContext? context) {
+    if (action is ContextAction<Intent>) {
+      return action.isEnabled(intent, context);
+    }
+    return action.isEnabled(intent);
+  }
+}
+
+/// Default shortcuts for Windows, Linux, Android and Fuchsia.
```
matthew-carroll, line 60: Please mention in the Dart Doc why this was copied, and whether we should keep it up to date with the framework.

super_editor/lib/src/infrastructure/actions.dart
```diff
@@ -0,0 +1,172 @@
...
+  SingleActivator(LogicalKeyboardKey.arrowUp, control: true): ScrollIntent(direction: AxisDirection.up),
+  SingleActivator(LogicalKeyboardKey.arrowDown, control: true): ScrollIntent(direction: AxisDirection.down),
+  SingleActivator(LogicalKeyboardKey.arrowLeft, control: true): ScrollIntent(direction: AxisDirection.left),
+  SingleActivator(LogicalKeyboardKey.arrowRight, control: true): ScrollIntent(direction: AxisDirection.right),
+  SingleActivator(LogicalKeyboardKey.pageUp): ScrollIntent(direction: AxisDirection.up, type: ScrollIncrementType.page),
+  SingleActivator(LogicalKeyboardKey.pageDown):
+      ScrollIntent(direction: AxisDirection.down, type: ScrollIncrementType.page),
+};
+
+/// Default shortcuts for the Apple platforms.
```
matthew-carroll, line 92: Same comment here.

super_editor/lib/src/infrastructure/actions.dart
```diff
@@ -0,0 +1,172 @@
...
+      return action.isEnabled(intent, context);
+    }
+    return action.isEnabled(intent);
+  }
+}
+
+/// Default shortcuts for Windows, Linux, Android and Fuchsia.
+///
+/// Copied from [WidgetsApp._defaultShortcuts].
+const Map<ShortcutActivator, Intent> defaultShortcuts = <ShortcutActivator, Intent>{
```
matthew-carroll, line 63: Seems like these aren't "default shortcuts" - they're "default Windows, Linux, Android, Fuchsia shortcuts"

super_editor/lib/src/infrastructure/actions.dart
```diff
@@ -0,0 +1,172 @@
...
+  SingleActivator(LogicalKeyboardKey.arrowUp, meta: true): ScrollIntent(direction: AxisDirection.up),
+  SingleActivator(LogicalKeyboardKey.arrowDown, meta: true): ScrollIntent(direction: AxisDirection.down),
+  SingleActivator(LogicalKeyboardKey.arrowLeft, meta: true): ScrollIntent(direction: AxisDirection.left),
+  SingleActivator(LogicalKeyboardKey.arrowRight, meta: true): ScrollIntent(direction: AxisDirection.right),
+  SingleActivator(LogicalKeyboardKey.pageUp): ScrollIntent(direction: AxisDirection.up, type: ScrollIncrementType.page),
+  SingleActivator(LogicalKeyboardKey.pageDown):
+      ScrollIntent(direction: AxisDirection.down, type: ScrollIncrementType.page),
+};
+
+/// Default shortcuts for web.
```
matthew-carroll, line 122: Same comment here.

super_editor/lib/src/infrastructure/actions.dart
```diff
@@ -0,0 +1,172 @@
...
+  SingleActivator(LogicalKeyboardKey.arrowRight): ScrollIntent(direction: AxisDirection.right),
+  SingleActivator(LogicalKeyboardKey.pageUp): ScrollIntent(direction: AxisDirection.up, type: ScrollIncrementType.page),
+  SingleActivator(LogicalKeyboardKey.pageDown):
+      ScrollIntent(direction: AxisDirection.down, type: ScrollIncrementType.page),
+};
+
+/// Generates the default shortcut key bindings based on the [defaultTargetPlatform].
+///
+/// Copied from [WidgetsApp.defaultShortcuts] to make it possible to force to use web shortcuts.
+Map<ShortcutActivator, Intent> get defaultFlutterShortcuts {
```
matthew-carroll, line 157: Please follow high level to low level code ordering.

super_editor/lib/src/infrastructure/platforms/windows/windows_ime.dart
```diff
@@ -0,0 +1,21 @@
...
+/// These intents have default handlers installed by Flutter
+/// that prevents the key events from bubbling up to the IME.
+///
+/// Use this object in an [Actions] widget.
+///
+/// See [WidgetsApp.defaultShortcuts] for the list of keybindings that Flutter
+/// adds by default.
+final Map<Type, Action<Intent>> disabledWindowsIntents = {
+  // Generated by pressing SPACE.
+  // PrioritizedIntents might contain intents that we don't want to prevent
```
matthew-carroll, line 15: Please insert blank line above this.

Also, how do we know what we want to block? I only see a type check for `ActivateIntent`, which makes it appear that there's a very specific thing we're trying to block...

super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart
```diff
@@ -461,7 +461,7 @@ class SuperDesktopTextFieldState extends State<SuperDesktopTextField> implements
     required Widget child,
   }) {
     return Actions(
-      actions: defaultTargetPlatform == TargetPlatform.macOS ? disabledMacIntents : {},
+      actions: CurrentPlatform.isApple ? disabledMacIntents : disabledWindowsIntents,
```
matthew-carroll, line 464: Same comment about other platforms.

super_editor/test/super_editor/supereditor_input_ime_test.dart
```diff
@@ -959,6 +959,26 @@ Paragraph two
       }, variant: inputSourceVariant);
     });
 
+    testWidgetsOnWebDesktop('inside a CustomScrollView > enters a space character with SPACE', (tester) async {
```
matthew-carroll, line 962: Can you name this test in terms of the problem? Seems very likely that a future reader would ask "why does this test need to exist?". If this test ensures that a bug doesn't happen, then we should probably name it in terms of the bug, right?

Maybe something like "SuperEditor inserts space instead of scrolling"?

super_editor/test/super_editor/supereditor_input_ime_test.dart
```diff
@@ -1494,3 +1514,20 @@ class _TestImeOverrides extends DeltaTextInputClientDecorator {
     performActionCallback(action);
   }
 }
+
+/// Simulates pressing the SPACE key.
+///
+/// First, this method simulates pressing the SPACE key on a physical keyboard. If that key event goes unhandled
+/// then this method generates an insertion delta of " ".
+///
+// TODO: extract this to the flutter_test_robots package.
```
matthew-carroll, line 1523: Let's go ahead and do this.

super_editor/test/super_editor/supereditor_test_tools.dart
```diff
@@ -409,6 +409,9 @@ class TestSuperEditorConfigurator {
     }
     return MaterialApp(
       theme: _config.appTheme,
+      // Add our own version of the shortcuts so we can use
```
matthew-carroll, line 412: This comment isn't very clear. I don't understand the implications of it. Please reword and elaborate on the significance of the `shortcuts` line.

super_editor/test/super_editor/supereditor_input_ime_test.dart
angelosilvestre: We need to release a new `flutter_test_robots` version with the new method.

super_editor/lib/src/infrastructure/actions.dart
angelosilvestre: Updated.

super_editor/lib/src/infrastructure/platforms/windows/windows_ime.dart
angelosilvestre: Updated.

angelosilvestre:
@matthew-carroll I moved our copy of the shortcuts into the test folder, because we don't need them in the editor code.

angelosilvestre:
@matthew-carroll As discussed on Discord, I made several changes on how we prevent the desired intents from bubbling up.

matthew-carroll (approved):
LGTM

PR #1942, opened 2024-04-16
[SuperEditor] Fix ordered list items sequence computation (Resolves #1938)

### LINKED ISSUE #1938: Computed ordinalValue is unexpected
**Package Version**
`main`

**To Reproduce**
1. Run the example application (test on web)
2. Scroll down to `Explore the toolkit 🔎`
3. Select last unordered list item under that section
4. Change the paragraph to "ordered list item"
5. Observe the `ordinalValue` being `5.`

**Actual behavior**
When an ordered list item is directly preceded by multiple unordered list items, its count is as if those unordered list items were ordered.

**Expected behavior**
I'd expect that the ordinalValue is computed based on previous list items that are also of "ordered" type not just all preceding list items with the same indent.
In other words a list such as:
```md
* item
* item
1. item
```
should translate to the above, rather than:
```md
* item
* item
3. item
```

**Platform**
Any

**Additional**
The problem seems to be here: https://github.com/superlistapp/super_editor/blob/661ecb84b4f79bd3f81c1bfd148b36ab7be6993b/super_editor/lib/src/default_editor/list_items.dart#L113-L123
Here computing the ordinal values doesn't take into consideration the type of the preceded ListItem.

angelosilvestre:
[SuperEditor] Fix ordered list items sequence computation. Resolves #1938

When an ordered list item is placed after an unordered list item, the unordered items are also being counted when we compute the ordered list item sequence. For example, consider the following markdown:

```
- First unordered item
- Second unordered item
1. First ordered item
2. Second ordered item
```

Instead of 1 and 2, the sequence is being computed as 2 and 3.

This PR fixes this issue by checking the list item type before incrementing the ordinal value.

### DIFF
--- a/super_editor/lib/src/default_editor/list_items.dart
+++ b/super_editor/lib/src/default_editor/list_items.dart
@@ -112,10 +112,16 @@ class ListItemComponentBuilder implements ComponentBuilder {
 
     int? ordinalValue;
     if (node.type == ListItemType.ordered) {
+      // Counts the number of ordered list items above the current node with the same indentation level.
       ordinalValue = 1;
       DocumentNode? nodeAbove = document.getNodeBefore(node);
       while (nodeAbove != null && nodeAbove is ListItemNode && nodeAbove.indent >= node.indent) {
-        if (nodeAbove.type == ListItemType.ordered && nodeAbove.indent == node.indent) {
+        if (nodeAbove.indent == node.indent) {
+          if (nodeAbove.type != ListItemType.ordered) {
+            // We found an unordered list item with the same indentation level as the ordered list item.
+            // Other ordered list items aboce this one do not belong to the same list.
+            break;
+          }
           ordinalValue = ordinalValue! + 1;
         }
         nodeAbove = document.getNodeBefore(nodeAbove);
--- a/super_editor/lib/src/default_editor/list_items.dart
+++ b/super_editor/lib/src/default_editor/list_items.dart
@@ -112,7 +112,9 @@ class ListItemComponentBuilder implements ComponentBuilder {
 
     int? ordinalValue;
     if (node.type == ListItemType.ordered) {
-      // Counts the number of ordered list items above the current node with the same indentation level.
+      // Counts the number of ordered list items above the current node with the same indentation level. Ordered
+      // list items with the same indentation level might be separated by ordered or unordered list items with
+      // different indentation levels.
       ordinalValue = 1;
       DocumentNode? nodeAbove = document.getNodeBefore(node);
       while (nodeAbove != null && nodeAbove is ListItemNode && nodeAbove.indent >= node.indent) {
--- a/super_editor/lib/src/default_editor/list_items.dart
+++ b/super_editor/lib/src/default_editor/list_items.dart
@@ -115,7 +115,7 @@ class ListItemComponentBuilder implements ComponentBuilder {
       ordinalValue = 1;
       DocumentNode? nodeAbove = document.getNodeBefore(node);
       while (nodeAbove != null && nodeAbove is ListItemNode && nodeAbove.indent >= node.indent) {
-        if (nodeAbove.indent == node.indent) {
+        if (nodeAbove.type == ListItemType.ordered && nodeAbove.indent == node.indent) {
           ordinalValue = ordinalValue! + 1;
         }
         nodeAbove = document.getNodeBefore(nodeAbove);
--- a/super_editor/lib/src/test/super_editor_test/supereditor_inspector.dart
+++ b/super_editor/lib/src/test/super_editor_test/supereditor_inspector.dart
@@ -217,6 +217,24 @@ class SuperEditorInspector {
     return findRichTextInParagraph(nodeId, superEditorFinder).style;
   }
 
+  /// Finds the ordered list item with the given [nodeId] and returns its index on the list.
+  ///
+  /// List items are indexed starting from 1.
+  ///
+  /// {@macro supereditor_finder}
+  static int findListItemOrdinal(String nodeId, [Finder? superEditorFinder]) {
+    final listItem = find
+        .ancestor(
+          of: find.byWidget(SuperEditorInspector.findWidgetForComponent(nodeId)),
+          matching: find.byType(OrderedListItemComponent),
+        )
+        .evaluate()
+        .single
+        .widget as OrderedListItemComponent;
+
+    return listItem.listIndex;
+  }
+
   /// Calculates the delta between the center of the character at [textOffset1] and and the
   /// center of the character at [textOffset2] within the node with the given [nodeId].
   ///
--- a/super_editor/lib/src/test/super_editor_test/supereditor_inspector.dart
+++ b/super_editor/lib/src/test/super_editor_test/supereditor_inspector.dart
@@ -217,9 +217,9 @@ class SuperEditorInspector {
     return findRichTextInParagraph(nodeId, superEditorFinder).style;
   }
 
-  /// Finds the ordered list item with the given [nodeId] and returns its index on the list.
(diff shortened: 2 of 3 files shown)


super_editor/lib/src/default_editor/list_items.dart
```diff
@@ -115,7 +115,7 @@ class ListItemComponentBuilder implements ComponentBuilder {
       ordinalValue = 1;
       DocumentNode? nodeAbove = document.getNodeBefore(node);
       while (nodeAbove != null && nodeAbove is ListItemNode && nodeAbove.indent >= node.indent) {
-        if (nodeAbove.indent == node.indent) {
+        if (nodeAbove.type == ListItemType.ordered && nodeAbove.indent == node.indent) {
```
jellynoone, line 118: How should markdown like this be translated?
```md
1. First ordered item
- First unordered item
- Second unordered item
1. First ordered item
2. Second ordered item
```

Wouldn't the current implementation translate it to:
```md
1. First ordered item
- First unordered item
- Second unordered item
2. First ordered item
3. Second ordered item
```

angelosilvestre: @jellynoone 

```
1. First ordered item
- First unordered item
- Second unordered item
1. First ordered item
2. Second ordered item
```

This should be the correct sequence. I updated the PR.

super_editor/lib/src/default_editor/list_items.dart
```diff
@@ -112,10 +112,16 @@ class ListItemComponentBuilder implements ComponentBuilder {
 
     int? ordinalValue;
     if (node.type == ListItemType.ordered) {
+      // Counts the number of ordered list items above the current node with the same indentation level.
       ordinalValue = 1;
       DocumentNode? nodeAbove = document.getNodeBefore(node);
       while (nodeAbove != null && nodeAbove is ListItemNode && nodeAbove.indent >= node.indent) {
         if (nodeAbove.indent == node.indent) {
+          if (nodeAbove.type != ListItemType.ordered) {
```
matthew-carroll, line 120: Any reason that the `ordered` check isn't added to the existing `while` condition, which already includes `is ListItemNode`?

super_editor/lib/src/test/super_editor_test/supereditor_inspector.dart
```diff
@@ -217,6 +217,24 @@ class SuperEditorInspector {
     return findRichTextInParagraph(nodeId, superEditorFinder).style;
   }
 
+  /// Finds the ordered list item with the given [nodeId] and returns its index on the list.
```
matthew-carroll, line 220: I think "on the list" should be "in the list". But also, this concept of a "list", do you intend for it to be partitioned by switching from ordered to unordered? Does it include different indentation levels? Please make sure the definition for what constitutes a "list" is documented here.

Also please define the term "index" in this case because it's not immediately obvious, given that a `Document` is a series of indexed nodes. Which "index" is being returend?

super_editor/test/super_editor/components/list_items_test.dart
```diff
@@ -512,6 +512,77 @@ void main() {
         expect(secondOrderedItem.listIndex, 2);
       });
 
+      testWidgetsOnArbitraryDesktop('starts sequence in the first ordered item', (tester) async {
```
matthew-carroll, line 515: I was going to offer a better name for this, but I'm not entirely sure what the current name is meant to capture, so I couldn't come up with anything.

super_editor/test/super_editor/components/list_items_test.dart
```diff
@@ -512,6 +512,77 @@ void main() {
...
+        expect(context.document.nodes[3], isA<ListItemNode>());
+        expect((context.document.nodes[3] as ListItemNode).type, ListItemType.ordered);
+
+        // Ensure the sequence started only after the first ordered item, i.e., the unordered items
+        // didn't affect the sequence.
+        expect(SuperEditorInspector.findListItemOrdinal(context.document.nodes[2].id), 1);
+        expect(SuperEditorInspector.findListItemOrdinal(context.document.nodes[3].id), 2);
+      });
+
+      testWidgetsOnArbitraryDesktop('does not keep sequence for items split by unordered items', (tester) async {
```
matthew-carroll, line 546: Let's call this "restarts item order when separated by an unordered item"

super_editor/lib/src/default_editor/list_items.dart
angelosilvestre: The check should be inside of the `while` because if we find an unordered list item with a higher indentation level, we still need to continue searching upstream for an ordered list item. For example:

```
1. First ordered item
   - First unordered item
   - Second unordered item
2. First ordered item
3. Second ordered item
```

We always walk backwards. When we are computing the ordinal value for "First ordered item" we need to ignore the two upstream unordered items to find the first ordered item.

If we check in the while condition it will result in the following sequence:

```
1. First ordered item
   - First unordered item
   - Second unordered item
1. First ordered item
2. Second ordered item
```

Which is something we don't want.

super_editor/lib/src/test/super_editor_test/supereditor_inspector.dart
angelosilvestre: I think that "Finds the ordered list item with the given [nodeId] and returns its ordinal value." might be more appropriate, so developer don't get confused thinking this is the index of the node in the `Document`

super_editor/test/super_editor/components/list_items_test.dart
angelosilvestre: In fact, I think the other test I added already captures everything we want to check here. We can have just one test.

super_editor/lib/src/default_editor/list_items.dart
```diff
@@ -112,10 +112,16 @@ class ListItemComponentBuilder implements ComponentBuilder {
...
+            // We found an unordered list item with the same indentation level as the ordered list item.
```
matthew-carroll, line 121: Reading this comment - is the situation really just unordered list item at the same indentation level? What about an unordered list item at a higher indentation level?

```
 1. ordered at level 1
 - unordered at level 1
    1. ordered at level 2 
```

Is this situation already handled somewhere else in here?

super_editor/test/super_editor/components/list_items_test.dart
```diff
@@ -512,6 +512,46 @@ void main() {
         expect(secondOrderedItem.listIndex, 2);
       });
 
+      testWidgetsOnArbitraryDesktop('restarts item order when separated by an unordered item', (tester) async {
+        final context = await tester //
+            .createDocument()
+            .fromMarkdown("""
+1. First ordered item
```
matthew-carroll, line 519: This test covers the case of a single-level list split between ordered and unordered. But do we have any tests to verify hierarchy combinations of ordered and unordered? Like the example list I posted in another comment?

super_editor/lib/src/default_editor/list_items.dart
angelosilvestre: It's already handled here. When we find an ordered list item with higher indentation, we ignore it and keep looking for a list item with the same indentation.

super_editor/test/super_editor/components/list_items_test.dart
angelosilvestre: We have a test for this example:


```

1. First ordered item
   - First unordered item
   - Second unoredered item

2. Second ordered item
   - First unordered item
   - Second unoredered item
```

And I'm adding another test for:

```
 1. list item 1
 2. list item 2
    1. list item 2.1
    2. list item 2.2
 3. list item 3
    1. list item 3.1
```

Do you think is there another input we need to test?

matthew-carroll (approved):
LGTM

PR #1943, opened 2024-04-18
Changed LinkAttribution to allow invalid (unchecked) URLs (Resolves #1927)

### LINKED ISSUE #1927: Can't create LinkAttributions with invalid URLs
**Package Version**
super_editor, main branch

**To Reproduce**
Try creating a `LinkAttribution` with an invalid url. Since `Uri` only allows valid urls, a `LinkAttribution` can't be created.

**Minimal Reproduction Code**
`LinkAttribution(url: Uri.parse('https://tel:+12%20345%20678%209'));`

**Actual behavior**
Can't create a `LinkAttribution` with invalid URL.
```
FormatException: Invalid port (at character 13)
https://tel:+12%20345%20678%209
            ^
```

**Expected behavior**
Changing `url` parameter in `LinkAttribution` to `String` instead, which would allow for invalid URLs:

`LinkAttribution(url: 'https://tel:+12%20345%20678%209');`

**Additional context**
We're taking an email in HTML format and converting that to a SuperEditor document.

Because an email is user generated content, we want to avoid modifying anything and just display it as-is. The problem is that the email contains an invalid link. SuperEditor does not let us render the document at all, because the `Uri` required by `LinkAttribution` does not allow us pass invalid URIs to it.

Apple Mail allows displaying the invalid link, but clicking on the link shows an error dialog:
image

matthew-carroll:
Changed LinkAttribution to allow invalid (unchecked) URLs (Resolves #1927)

The `LinkAttribution`'s `url` property is now a `String` instead of a `Uri`, so that `LinkAttribution`s can hold invalid URLs.

For easy `Uri` access, I added a `uri` property, which parses the `url` when it's accessed. This is a convenience so that users don't have to explicitly parse a `Uri` all over the place.

FYI: The motivation for this change is that an existing document, possibly from another source, is parsed and loaded into `SuperEditor`. This process fails for a document that happens to contain links for invalid URLs.

### DIFF
--- a/super_editor/example/lib/demos/demo_markdown_serialization.dart
+++ b/super_editor/example/lib/demos/demo_markdown_serialization.dart
@@ -137,11 +137,11 @@ MutableDocument _createInitialDocument() {
             AttributedSpans(
               attributions: [
                 SpanMarker(
-                    attribution: LinkAttribution(url: Uri.https('example.org', '')),
+                    attribution: LinkAttribution.fromUri(Uri.https('example.org', '')),
                     offset: 30,
                     markerType: SpanMarkerType.start),
                 SpanMarker(
-                    attribution: LinkAttribution(url: Uri.https('example.org', '')),
+                    attribution: LinkAttribution.fromUri(Uri.https('example.org', '')),
                     offset: 35,
                     markerType: SpanMarkerType.end),
               ],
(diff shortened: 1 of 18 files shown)


matthew-carroll:
@roughike please let me know if this covers what you need.

angelosilvestre (approved):
LGTM

matthew-carroll:
@roughike - checking again on this. Can you let me know if this change provides what you need?

matthew-carroll:
@miguelcmedeiros @brian-superlist - Can one of you follow up with @roughike on this?

roughike (approved):
Yup, this is exactly what is needed. Thanks!

PR #1945, opened 2024-04-18
Center caret based on width instead of expanding to the right

KevinBrendel:
Currently, if the caret width is increased, it increases towards the right, making the caret appear visually in the wrong position, as it is positioned based on its left edge. This pull request fixes this, bringing the behavior also more in line with the Flutter TextFields.


KevinBrendel:
Should be ready now.

matthew-carroll (approved):
LGTM

matthew-carroll:

PR #1949, opened 2024-04-19
[SuperEditor] Fix toggling attributions crash (Resolves #1948)

### LINKED ISSUE #1948: [SuperEditor] Toggling attribution throws
**Package Version**
Using `main` branch with commit e083e6bb550892785e6bccd5a7a33a51ed462850.

**To Reproduce**
1. Go to SuperEditor demo
2. Place the selection at the end of a text node and expand the selection to the content of the next text node
3. Press CMD+B twice

**Actual behavior**
Text is made bold and then turned to normal.

**Expected behavior**
Text is made bold, but an exception is thrown and the editor is left in a broken state:
```
======== Exception caught by services library ======================================================
The following _Exception was thrown while processing the key message handler:
Exception: removeAttribution() did not satisfy start < 0 and start > end, start: 70, end: 69

#0      AttributedSpans.removeAttribution (package:attributed_text/src/attributed_spans.dart:422:7)
#1      AttributedText.removeAttribution (package:attributed_text/src/attributed_text.dart:189:11)
#2      ToggleTextAttributionsCommand.execute (package:super_editor/src/default_editor/text.dart:1526:14)
#3      _DocumentEditorCommandExecutor.executeCommand (package:super_editor/src/core/editor.dart:252:15)
#4      Editor._executeCommand (package:super_editor/src/core/editor.dart:195:22)
#5      Editor.execute (package:super_editor/src/core/editor.dart:148:30)
#6      CommonEditorOperations.toggleAttributionsOnSelection (package:super_editor/src/default_editor/common_editor_operations.dart:1466:12)
#7      cmdBToToggleBold (package:super_editor/src/default_editor/document_hardware_keyboard/document_keyboard_actions.dart:337:27)
#8      _SuperEditorHardwareKeyHandlerState._onKeyPressed (package:super_editor/src/default_editor/document_hardware_keyboard/document_physical_keyboard.dart:84:50)
#9      _HighlightModeManager.handleKeyMessage (package:flutter/src/widgets/focus_manager.dart:2069:29)
#10     KeyEventManager._dispatchKeyMessage (package:flutter/src/services/hardware_keyboard.dart:1099:34)
#11     KeyEventManager.handleRawKeyMessage (package:flutter/src/services/hardware_keyboard.dart:1171:17)
#12     BasicMessageChannel.setMessageHandler.<anonymous closure> (package:flutter/src/services/platform_channel.dart:235:49)
#13     _DefaultBinaryMessenger.setMessageHandler.<anonymous closure> (package:flutter/src/services/binding.dart:576:35)
#14     _invoke2 (dart:ui/hooks.dart:344:13)
#15     _ChannelCallbackRecord.invoke (dart:ui/channel_buffers.dart:45:5)
#16     _Channel.push (dart:ui/channel_buffers.dart:135:31)
#17     ChannelBuffers.push (dart:ui/channel_buffers.dart:343:17)
#18     PlatformDispatcher._dispatchPlatformMessage (dart:ui/platform_dispatcher.dart:750:22)
#19     _dispatchPlatformMessage (dart:ui/hooks.dart:257:31)
KeyMessage: KeyMessage([KeyDownEvent#ae18e(physicalKey: PhysicalKeyboardKey#7e294(usbHidUsage: "0x00070005", debugName: "Key B"), logicalKey: LogicalKeyboardKey#15902(keyId: "0x00000062", keyLabel: "B", debugName: "Key B"), character: "b", timeStamp: 1474:27:27.470860)])
====================================================================================================
```

**Platform**
MacOS

**Flutter version**
Master channel with commit b9e9435b25c45c41314979c06115373cc811ce1d.

**Screenshots**

https://github.com/superlistapp/super_editor/assets/31278849/f08bc48a-50ab-46ba-af45-7004c57ace8d

angelosilvestre:
[SuperEditor] Fix toggling attributions crash. Resolves #1948

When selecting across nodes, starting from the end of the upstream node, and toggling any attribution the following exception happens:

```console
======== Exception caught by services library ======================================================
The following _Exception was thrown while processing the key message handler:
Exception: removeAttribution() did not satisfy start < 0 and start > end, start: 70, end: 69
```

https://github.com/superlistapp/super_editor/assets/31278849/f08bc48a-50ab-46ba-af45-7004c57ace8d

The cause is that we are trying to remove the attribution from an invalid range (start > end). Since the selection starts at the end of the upstream node, there isn't any selected text there to apply attributions.

This PR changes `ToggleTextAttributionsCommand` to ignore a nodes if the selection starts after the last character.

### DIFF
--- a/super_editor/lib/src/default_editor/text.dart
+++ b/super_editor/lib/src/default_editor/text.dart
@@ -1495,6 +1495,12 @@ class ToggleTextAttributionsCommand implements EditCommand {
         endOffset = max(textNode.text.length - 1, 0);
       }
 
+      if (startOffset >= textNode.text.length) {
+        // The selection begins after the last character in the node. There's no text
+        // to toggle any attributions. Skip this node.
+        continue;
+      }
+
       final selectionRange = SpanRange(startOffset, endOffset);
 
       alreadyHasAttributions = alreadyHasAttributions &&
--- a/super_editor/lib/src/default_editor/text.dart
+++ b/super_editor/lib/src/default_editor/text.dart
@@ -1480,6 +1480,14 @@ class ToggleTextAttributionsCommand implements EditCommand {
         editorDocLog.info(' - selecting part of the first node: ${textNode.id}');
         startOffset = (normalizedRange.start.nodePosition as TextPosition).offset;
         endOffset = max(textNode.text.length - 1, 0);
+
+        if (startOffset >= textNode.text.length) {
+          // The range spans multiple nodes, starting at the end of the first node of the
+          // range. From the first node's perspective, this is equivalent to a collapsed
+          // selection at the end of the node. There's no text to toggle any attributions.
+          // Skip this node.
+          continue;
+        }
       } else if (textNode == nodes.last) {
         // Handle partial node selection in last node.
         editorDocLog.info(' - toggling part of the last node: ${textNode.id}');
@@ -1488,17 +1496,19 @@ class ToggleTextAttributionsCommand implements EditCommand {
         // -1 because TextPosition's offset indexes the character after the
         // selection, not the final character in the selection.
         endOffset = (normalizedRange.end.nodePosition as TextPosition).offset - 1;
+
+        if (endOffset == 0) {
+          // The range spans multiple nodes, ending at the beginning of the last node of the
+          // range. From the last node's perspective, this is equivalent to a collapsed
+          // selection at the beginning of the node. There's no text to toggle any attributions.
+          // Skip this node.
+          continue;
+        }
       } else {
         // Handle full node selection.
         editorDocLog.info(' - toggling full node: ${textNode.id}');
         startOffset = 0;
         endOffset = max(textNode.text.length - 1, 0);
-      }
-
-      if (startOffset >= textNode.text.length) {
-        // The selection begins after the last character in the node. There's no text
-        // to toggle any attributions. Skip this node.
-        continue;
       }
 
       final selectionRange = SpanRange(startOffset, endOffset);
--- a/super_editor/lib/src/default_editor/text.dart
+++ b/super_editor/lib/src/default_editor/text.dart
@@ -1497,7 +1497,7 @@ class ToggleTextAttributionsCommand implements EditCommand {
         // selection, not the final character in the selection.
         endOffset = (normalizedRange.end.nodePosition as TextPosition).offset - 1;
 
-        if (endOffset == 0) {
+        if (endOffset <= 0) {
           // The range spans multiple nodes, ending at the beginning of the last node of the
           // range. From the last node's perspective, this is equivalent to a collapsed
           // selection at the beginning of the node. There's no text to toggle any attributions.
--- a/super_editor/test/super_editor/supereditor_attributions_test.dart
+++ b/super_editor/test/super_editor/supereditor_attributions_test.dart
@@ -584,7 +584,8 @@ void main() {
           );
         });
 
-        testWidgetsOnAllPlatforms("toggles an attribution when selection starts at the end of the upstream node",
+        testWidgetsOnAllPlatforms(
+            "toggles an attribution when selection spans multiple nodes and starts at the end of the first selected node",
             (tester) async {
           final context = await tester //
               .createDocument()
--- a/super_editor/test/super_editor/supereditor_attributions_test.dart
+++ b/super_editor/test/super_editor/supereditor_attributions_test.dart
@@ -584,6 +584,50 @@ void main() {
           );
         });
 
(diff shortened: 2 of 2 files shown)


matthew-carroll:
> This PR changes ToggleTextAttributionsCommand to ignore a nodes if the selection starts after the last character.

That sounds like a band-aid. Why is the command being told to toggle attributions outside of the valid range. We don't want to ignore such things, we want to solve them.

angelosilvestre:
@matthew-carroll We are issuing a `ToggleTextAttributionsRequest` with the current selection, without looking where the selection is. Since the selection can start at the end of the node, we can issue a `ToggleTextAttributionsRequest` starting at the end of a node.  

Here's the method that issues the request:

```dart
void toggleAttributionsForDocumentSelection(
    DocumentSelection documentSelection,
    Set<Attribution> attributions,
  ) {
    return execute([
      ToggleTextAttributionsRequest(
        documentRange: documentSelection,
        attributions: attributions,
      )
    ]);
  }
```

Are you suggesting that we should check here if the selection starts at the end of a node, and then adjust the `DocumentRange` that we pass to `ToggleTextAttributionsRequest`? Or are you suggesting something else?

matthew-carroll:
What's a specific paragraph of text, and a specific selection range that exemplifies this issue? I want to make sure I understand the issue with real numbers and text.

angelosilvestre:
Here's an example, where `|` represents the selection boundaries:

```
First Node.|

Last| Node
```

Range:  Node 1, offset 11 -> Node 2, offset 4

The offset of the second node doesn't matter though.

matthew-carroll:
You used a single "|" to represent an expanded selection. Please include both sides of the selection.

angelosilvestre:
I didn't get it, the selection starts at the first node and ends at the last node, each "|" represents one side of the selection.  Another example if `>` and `<` as delimiters:

```
First Node.>

Last< Node
```

matthew-carroll:
Ok. I'll think about it. In general, the code that requests the change should make any needed adjustments for the change to be valid. I'm not sure how tedious that would be here.

matthew-carroll:
I guess in this case, the request and command are both based on a document selection. Therefore, any valid document selection should accepted, including this one.

The confusion with how the original explanation was written is that it sounded like the selection was invalid. But that's not the case. The selection is valid, but once we break the selection down by node, we might end up with an empty selection at the beginning or end of a node.

I'll add specific comments to your code changes.

super_editor/lib/src/default_editor/text.dart
```diff
@@ -1495,6 +1495,12 @@ class ToggleTextAttributionsCommand implements EditCommand {
         endOffset = max(textNode.text.length - 1, 0);
       }
 
+      if (startOffset >= textNode.text.length) {
```
matthew-carroll, line 1498: By placing this check in a way that always runs, it makes it appear as if this possibility could happen at any time. But that's not true, is it? Based on your example of the problem, this issue seems to be tied to a very specific situation. If that's the case, we don't want to over-generalize and confuse future readers as to why the start offset might be beyond the length of the text.

Additionally, while the error might only trigger when the selection collapses at the end of the text, it's equally pointless to try to toggle attributions if the same exact thing happens at the start of the text, too, right?

If my understanding is correct, then what you've identified is a possible degenerate situation that only comes up when the selection spans multiple nodes. We already have an if-statement up above that deals with partial node selections. Can you adjust that part of the code, specifically, to resolve this?

super_editor/test/super_editor/supereditor_attributions_test.dart
```diff
@@ -584,6 +584,50 @@ void main() {
           );
         });
 
+        testWidgetsOnAllPlatforms("toggles an attribution when selection starts at the end of the upstream node",
```
matthew-carroll, line 587: If my comment up above is accurate, then the name of this method should be changed to reflect the true meaning of the situation. It's not as if the selection randomly ended up at an illegal offset.

angelosilvestre: I updated the test name, but I'm not sure if it was improved enough...

super_editor/test/super_editor/supereditor_attributions_test.dart
```diff
@@ -584,6 +584,51 @@ void main() {
           );
         });
 
+        testWidgetsOnAllPlatforms(
+            "toggles an attribution when selection spans multiple nodes and starts at the end of the first selected node",
```
matthew-carroll, line 588: Please add another test for the version where the selection ends at the beginning of the last node. Also, please add the issue ticket number in a comment on the first line of each test so we know where these specific conditions came from.

angelosilvestre: Updated.

matthew-carroll (approved):
LGTM

PR #1951, opened 2024-04-20
[SuperEditor][Android] Hide expanded drag handles after deleting the selected text (Resolves #1936)

### LINKED ISSUE #1936: Exception thrown when dragging Android handles after `cut` or regular delete
**Package Version**
`main`

**To Reproduce**
The behavior can be reproduced in the example application.

1. Run example application on an Android device
2. Select text
3. Select cut from editing toobar
4. See non-collapsed text selection handles
5. Attempt to drag one
7. Observe `Exception: Tried to drag an expanded Android handle but the selection is collapsed.` followed by `Exception: Tried to pan an Android drag handle but the focal point is null. The focal point is set when the drag begins. This shouldn't be possible.`

The same behavior can be observed if the selected text is removed via backspace not just cut.

Additionally, when text is cut or deleted and you type something, the text selection handlers continue to indicate an uncollapsed selection even though it is actually collapsed.

**Actual behavior**
The selection handlers indicate a non-collapsed document selection and it seems other parts of super_editor don't expect this.

**Expected behavior**
After cut or deletion, the selection handlers should probably be removed or collapsed.

**Platform**
Android

**Flutter version**
```
Flutter 3.19.5 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 300451adae (3 weeks ago) • 2024-03-27 21:54:07 -0500
Engine • revision e76c956498
Tools • Dart 3.3.3 • DevTools 2.31.1
```

angelosilvestre:
[SuperEditor][Android] Hide expanded drag handles after deleting the selected text. Resolves #1936

On Android with an expanded selection, performing any action that causes text to be deleted, such as cutting, deleting, pasting or typing to replace the text, keeps the expanded handles visible. Trying to drag those handles throws an exception:

```console
The following _Exception was thrown while handling a gesture:
Exception: Tried to drag an expanded Android handle but the selection is collapsed.
```

The expanded handles are kept visible because of https://github.com/superlistapp/super_editor/pull/1754, where we changed the expanded handles to be allowed to be visible even with a collapsed selection. However, this should happen when the user is dragging the handles, not when the selection collapses due to a deletion.

Looking at other Android apps, it seems that the expected behavior is that no handle should be visible after deleting the selected text, not even the collapsed drag handle.

This PR changes the Android touch interactor to hide the expanded handles and any other controls when text is deleted, causing the selection to collapse. 

We could also implement this by adding a listener directly to the editor and then listening for selection change events. If a better solution would be to find a way to make `shouldShowExpandedHandles` to be `false` if the user is not dragging then I can dig into that.

On iOS this issue doesn't happen, because the expanded handles are never visible when the selection is collapsed. I added the same test I added for android.

### DIFF
--- a/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
@@ -13,7 +13,6 @@ import 'package:super_editor/src/core/document_selection.dart';
 import 'package:super_editor/src/core/edit_context.dart';
 import 'package:super_editor/src/core/editor.dart';
 import 'package:super_editor/src/default_editor/super_editor.dart';
-import 'package:super_editor/src/default_editor/text.dart';
 import 'package:super_editor/src/default_editor/text_tools.dart';
 import 'package:super_editor/src/document_operations/selection_operations.dart';
 import 'package:super_editor/src/infrastructure/_logging.dart';
@@ -656,38 +655,7 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
     widget.dragHandleAutoScroller.value!.ensureOffsetIsVisible(extentOffsetInViewport);
   }
 
-  void _onDocumentChange(DocumentChangeLog changeLog) {
-    final selectionAfterChange = widget.selection.value;
-    final areExpandedHandlesVisible = _controlsController!.shouldShowExpandedHandles.value == true;
-
-    // Checks whether a deletion ocurred and caused the selection to collapse.
-    //
-    // Even though we are checking against the new selection, if the user deleted text while the selection
-    // was expanded, the new selection will have the same offset. We use the selection after the document
-    // changes because the changelog doesn't contain selection change events.
-    //
-    // The reason of expanded handles being allowed to be visible even when the selection is collapsed is
-    // that dragging an expanded handle can cause the selection to collapse. The drag handle remains visible
-    // so the user can continue dragging, possibly causing the selection to expand again.
-    final deletionCausedSelectionToCollapse = areExpandedHandlesVisible &&
-        selectionAfterChange != null &&
-        selectionAfterChange.isCollapsed &&
-        changeLog.changes
-            .whereType<TextDeletedEvent>()
-            .where((deletion) =>
-                deletion.nodeId == selectionAfterChange.start.nodeId &&
-                deletion.offset == (selectionAfterChange.start.nodePosition as TextNodePosition).offset)
-            .isNotEmpty;
-
-    if (deletionCausedSelectionToCollapse) {
-      _controlsController!
-        ..hideCollapsedHandle()
-        ..hideExpandedHandles()
-        ..hideMagnifier()
-        ..hideToolbar()
-        ..blinkCaret();
-    }
-
+  void _onDocumentChange(_) {
     onNextFrame((_) {
       _ensureSelectionExtentIsVisible();
     });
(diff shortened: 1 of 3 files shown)


super_editor/lib/src/default_editor/document_gestures_touch_android.dart
```diff
@@ -655,7 +656,38 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
     widget.dragHandleAutoScroller.value!.ensureOffsetIsVisible(extentOffsetInViewport);
   }
 
-  void _onDocumentChange(_) {
+  void _onDocumentChange(DocumentChangeLog changeLog) {
```
matthew-carroll, line 659: My initial thought is that this a backwards place to recognize such a situation. We have various Android-specific widgets and configurations for the UI. The UI is where the user's interaction originates. Is there not a much simpler location in the Android-specific UI where we can recognize this situation, rather than try to reconstruct our knowledge of it after the fact?

angelosilvestre: I don't think so, even though we can hide the controls in the `cut` handler, for the cases where the user presses the backspace button, or type when the selection is expanded, the IME will generate deltas that will modify the document, without any user interaction with the app. The user interaction in this case is only with the software keyboard.

matthew-carroll: When the handles rebuild for the new collapsed selection, doesn't the Android interactor know that the user isn't currently dragging a handle, and therefore knows that we should collapse the handles?

matthew-carroll:
@angelosilvestre are you waiting for me on this? It looks like I had an open question.

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
angelosilvestre: I'm not sure if I understood where you are suggesting to write this code...

matthew-carroll: @angelosilvestre can you please frame that as a question? Where are you getting confused?

angelosilvestre: @matthew-carroll I'll take a look at the existing code to frame a question.

angelosilvestre:
@matthew-carroll I modified the implementation as you suggested.

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
```diff
@@ -1363,6 +1370,23 @@ class SuperEditorAndroidControlsOverlayManagerState extends State<SuperEditorAnd
   @visibleForTesting
   bool get wantsToDisplayMagnifier => _controlsController!.shouldShowMagnifier.value;
 
+  void _onDocumentChange(_) {
```
matthew-carroll, line 1373: It looks like this only moved the problem I identified before. The problem is that you're trying to identify a user interaction through a document change. My original question, which is still my question, is why can't you identify the user interaction when the user interaction happens, instead of after the fact?

This issue says its about "deleting the selected text". I assume that involves an expanded selection, then some kind of keyboard interaction, then a collapsed selection.

Doesn't this widget already know when the selection is expanded, and when the selection is collapsed? Why can't the widget recognize that the selection has collapsed and then hide the expanded handles, without worrying about document changes at all?

angelosilvestre: Updated.

matthew-carroll (approved):
LGTM

PR #1961, opened 2024-04-22
Reparent focus node in deactivate

knopp:
Fixes the following exception that we're getting
```
SEVERE: 'package:flutter/src/widgets/focus_manager.dart': Failed assertion: line 1299 pos 12: '_focusedChildren.isEmpty || _focusedChildren.last.enclosingScope == this': Focused child does not have the same idea of its enclosing scope as the scope does.
#0      _AssertionError._doThrowNew (dart:core-patch/errors_patch.dart:50:61)
#1      _AssertionError._throwNew (dart:core-patch/errors_patch.dart:40:5)
#2      FocusScopeNode.focusedChild (package:flutter/src/widgets/focus_manager.dart:1299:12)
#3      FocusNodeDirectionalExtension.isSelected (package:superlist/ui/focus/focus.dart:25:40)
#4      _TaskListTileScaffoldState.build.<anonymous closure> (package:superlist/ui/task_list/private/task_list_tile_scaffold.dart:444:48)
#5      _HoverBuilderState.build (package:super_utils/src/widgets/hover_builder.dart:78:32)
#6      StatefulElement.build (package:flutter/src/widgets/framework.dart:5593:27)
```

It seems to be caused by NonReparentingNode moving to another part of widget hierarchy using global key, which causes the focus node to be stuck with wrong parent. Alternatively we should reconsider if NonReparentingNode is necessary at all, given that the overlays should now be handled by overlay portal.

### DIFF
--- a/super_editor/lib/src/infrastructure/focus.dart
+++ b/super_editor/lib/src/infrastructure/focus.dart
@@ -52,6 +52,13 @@ class _NonReparentingFocusState extends State<NonReparentingFocus> {
   }
 
   @override
+  void deactivate() {
+    super.deactivate();
+    // See _FocusState.deactivate.
+    _keyboardFocusAttachment.reparent();
+  }
+
+  @override
   void didUpdateWidget(NonReparentingFocus oldWidget) {
     super.didUpdateWidget(oldWidget);
 


knopp:
@matthew-carroll, this is currently P1 issue for us. Latest Flutter introduced caching of enclosing scope, which changes subtle change of behavior and results in an exception now.

knopp:
The `NonReparentingFocus` widget is a constant source of issues, and if it's not necessary anymore (i.e. due to usage of overlay portal) it would be best if removed.

matthew-carroll (approved):
LGTM

matthew-carroll:
@angelosilvestre can you check if this should be cherry-picked? I don't know if this change might cause an issue on stable or not.

knopp:
> Can you please elaborate? We certainly haven't seen a constant stream of issue tickets about it. What other problems are you running into?

Last thing I remember debugging was problem with shortcuts not working correctly after reparenting. It's been a while. We have a workaround where there is another focus node right above supertextfield just for this, but ideally I'd want to get rid of it.

matthew-carroll:
> but ideally I'd want to get rid of it.

This is true of many things in Super Editor, because we've had to do many things to deal with holes in Flutter. However, I'd like to stay away from hyperbole. If it's been so long that you don't recall the most previous issue, then please don't say it's a constant source of issues. That doesn't seem to be accurate, and it could lead us to make a hasty change that hurts others.

And, of course, we must always consider others. Which is why I've asked @angelosilvestre to look into it. If Flutter has filled all relevant holes, then sure, we can get rid of it. But if Flutter still has a number of the original issues that forced us to create the widget, then it would probably not be advisable to remove it. Though, perhaps we could look into minimizing its use, if that would truly reduce problems for Superlist.

knopp:
I might be very wrong here, but I think the hole in this case was filled with the introduction of Overlay portal, which decouples the render object tree from element tree.

angelosilvestre:
@matthew-carroll If `OverlayPortal` is used, it seems we don't need `NonReparentingFocus` anymore. Also, it seems we don't need `FocusWithCustomParent` either, because Flutter added a `parentNode` property to `Focus`.

The `parentNode` docs say:

```dart
/// If [parentNode] is null, then [Focus.maybeOf] is used to find the parent
/// in the widget tree, which is typically what is desired, since it is easier
/// to reason about the focus tree if it mirrors the shape of the widget tree.
```

When placed inside an `OverlayPortal`, `Focus.maybeOf` should return the expected parent focus node.

We can't force people to use `OverlayPortal`, but we still should be able to share focus by wrapping the subtree with a `Focus` and passing the desired `parentNode`. Maybe we need a test to confirm that it works.

angelosilvestre:
> @angelosilvestre can you check if this should be cherry-picked? I don't know if this change might cause an issue on stable or not.

@matthew-carroll  The scope caching isn't in Flutter stable yet,  but it shouldn't cause any issues if we cherry-pick this PR to stable.

matthew-carroll:
I filed https://github.com/superlistapp/super_editor/issues/1963 for @angelosilvestre to do some deeper analysis and attempt to remove the custom widgets.

Also @angelosilvestre if we can cherry pick this to stable then please go ahead and do so.

PR #1964, opened 2024-04-23
Optimize ContentLayers._isSubtreeDirty

knopp:
I was investigating excessive garbage collection when changing selection in documents with large number of tasks (large widget hierarchy):

isSubtreeDirty - before

Much of the garbage seems to be generated from `isSubtreeDirty`, which is used to traverse, in this instance, a very large widget tree.

```dart
  bool _isSubtreeDirty(Element element) {
    contentLayersLog.finest("Finding dirty children for: $element");
    if (element.dirty) {
      contentLayersLog.finest("Found a dirty child: $element");
      return true;
    }

    bool isDirty = false;
    element.visitChildren((childElement) {
      isDirty = isDirty || _isSubtreeDirty(childElement);
    });
    return isDirty;
  }
```

There are two issues here. Interpolating the log string (and serializing the diagnosticable nodes) and the closure allocated for each recursive call. The string interpolation can be put behind a log level check and the closure allocation can be avoided by doing a static closure. 

With this PR the garbage collection pretty much disappears from the profile:

Screenshot 2024-04-23 at 16 01 48

### DIFF
--- a/super_editor/lib/src/infrastructure/content_layers.dart
+++ b/super_editor/lib/src/infrastructure/content_layers.dart
@@ -2,6 +2,7 @@ import 'package:flutter/foundation.dart';
 import 'package:flutter/rendering.dart';
 import 'package:flutter/scheduler.dart';
 import 'package:flutter/widgets.dart';
+import 'package:logging/logging.dart';
 import 'package:super_editor/src/infrastructure/_logging.dart';
 
 /// Widget that displays [content] above a number of [underlays], and beneath a number of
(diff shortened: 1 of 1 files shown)


knopp:
cc @matthew-carroll

super_editor/lib/src/infrastructure/content_layers.dart
```diff
@@ -208,31 +209,48 @@ class ContentLayersElement extends RenderObjectElement {
...
     }
+    element.visitChildren(_isSubtreeDirtyVisitor);
+  }
 
-    bool isDirty = false;
-    element.visitChildren((childElement) {
-      isDirty = isDirty || _isSubtreeDirty(childElement);
-    });
-    return isDirty;
+  bool _isSubtreeDirty(Element element) {
```
matthew-carroll, line 250: Please order method declarations in the order that they're used. In this case `_isSubtreeDirty()` before `_isSubtreeDirtyVisitor()`.

PS - Typically we place all static members above all instance members. However, in this very unusual situation, we'll treat the static property and method as instance members because we're working around a language issue.

super_editor/lib/src/infrastructure/content_layers.dart
```diff
@@ -208,31 +209,48 @@ class ContentLayersElement extends RenderObjectElement {
...
+      assert(() {
+        if (contentLayersLog.isLoggable(Level.FINEST)) {
+          contentLayersLog.finest("Found a dirty child: $element");
+        }
+        return true;
+      }());
+      _isDirty = true;
+      return;
     }
+    element.visitChildren(_isSubtreeDirtyVisitor);
```
matthew-carroll, line 247: I'm surprised the compiler wasn't already in-lining this.

super_editor/lib/src/infrastructure/content_layers.dart
```diff
@@ -208,31 +209,48 @@ class ContentLayersElement extends RenderObjectElement {
...
-  bool _isSubtreeDirty(Element element) {
-    contentLayersLog.finest("Finding dirty children for: $element");
-    if (element.dirty) {
-      contentLayersLog.finest("Found a dirty child: $element");
+  static bool _isDirty = false;
+
+  // This is intentionally static to prevent closure allocation during
+  // the traversal of the element tree.
+  static void _isSubtreeDirtyVisitor(Element element) {
+    // Can't use the () => message syntax because it allocates a closure.
```
matthew-carroll, line 230: If `() => ` allocates a closure, and that's harmful to performance, then why did you use that syntax up above?

matthew-carroll:
@knopp - Can you check the failing analysis? I'm guessing it's some new lint check that's only on master? I'm not sure.

super_editor/lib/src/infrastructure/content_layers.dart
knopp: It's an indirect call to a closure or function pointer passed to a virtual method. I don't think it's really possible to inline this. The call overhead is much less of a problem than the need to allocate closure that references local variable (which on top of it is modified from closure so itself needs to be boxed).

super_editor/lib/src/infrastructure/content_layers.dart
knopp: Because there's probably several orders of magnitude less widgets in content layers than in document. In this case I think we have over 100000 elements in document all being traversed. If you want me to I can do the same thing above (for consistency).

knopp:
> @knopp - Can you check the failing analysis? I'm guessing it's some new lint check that's only on master? I'm not sure.

Will do.

super_editor/lib/src/infrastructure/content_layers.dart
knopp: Btw. I introduced the `() =>` closure to prevent diagnosticable tree creates in debug mode during every call (closures in log are only evaluated when log lovel is reached), which is quite slow especially in debug mode.

super_editor/lib/src/infrastructure/content_layers.dart
knopp: Where should the static variable go?

knopp:
@matthew-carroll, I moved the visitor method below and the PR to fix the analyser complaining is [here](https://github.com/superlistapp/super_editor/pull/1968).

knopp:
Friendly ping, @matthew-carroll, does this need anything else?

matthew-carroll (approved):
LGTM

matthew-carroll:

PR #1966, opened 2024-04-23
[SuperEditor][SuperReader] Do not stop scroll momentum on mouse move (Resolves #1946)

### LINKED ISSUE #1946: Scrolling with magic mouse doesn't work properly
`DocumentMouseInteractor` calls `cancelScrollMomentum` on mouse move, which stops scrolling, but with magic mouse it is possible (and normal) to move mouse. I don't really understand the intent behind reimplementing scrolling this way, but IIRC there is `PointerScrollInertiaCancelEvent` which would normally signal when the scrolling need to stop.

angelosilvestre:
[SuperEditor][SuperReader] Do not stop scroll momentum on mouse move. Resolves #1946

On desktop, moving the mouse while the content is scrolling causes the scrolling to stop.  We introduced this change sometime ago, when the Flutter team made changes related to scroll momentum.

Now, Flutter has a dedicated signal for cancelling the scroll momentum: `PointerScrollInertiaCancelEvent`, which is already handled by `Scrollable`. Therefore, we don't need to cancel momentum ourselves.

This PR changes the mouse interactor to avoid cancelling momentum on mouse move.

### DIFF
--- a/super_editor/lib/src/default_editor/document_gestures_mouse.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_mouse.dart
@@ -577,13 +577,6 @@ class _DocumentMouseInteractorState extends State<DocumentMouseInteractor> with 
     }
   }
 
-  /// Beginning with Flutter 3.3.3, we are responsible for starting and
-  /// stopping scroll momentum. This method cancels any scroll momentum
-  /// in our scroll controller.
-  void _cancelScrollMomentum() {
-    widget.autoScroller.goIdle();
-  }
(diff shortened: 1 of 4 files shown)


matthew-carroll:
@angelosilvestre can you check on the failing test for super text layout? I'm guessing its something that `master` introduced, but not sure.

angelosilvestre:
@matthew-carroll This seems to be the analyzer complaining. I think that a PR was already merged to main to fix that.

PR #1967, opened 2024-04-24
[SuperEditor][Mobile] Fix downstream image caret position (Resolves #1959)

### LINKED ISSUE #1959: Mobile: Caret position is wrong when dealing with ImageNode
**Package Version**
Stable or main branch

**To Reproduce**
On SuperEditor Editor Demo:

- Tap on the left side of an image.
- The cursor is placed on the left side of the image.
- Tap backspace ➝ nothing happens, as expected, since you are at the beginning of the document.
- Tap on the right side of the image.
- The cursor remains on the left side.
- Tap backspace ➝ the image block is deleted, indicating that the selection was actually on the right.

[cursor_does_not_represent_real_selection.webm](https://github.com/superlistapp/super_editor/assets/6229343/4cb74b91-5d13-48a6-ba5f-4ad2bd6bae8e)

**Actual behavior**
The caret position on screen does not represent the real selection position.

**Expected behavior**
When taping on the right side of the image, the caret should be visible along the right side of the image

**Platform**
iOS and Android

**Flutter version**
[✓] Flutter (Channel stable, 3.19.6, on macOS 14.3 23D56 darwin-arm64, locale en-FR)

**Screenshots**

**Additional context**

angelosilvestre:
[SuperEditor][Mobile] Fix downstream image caret position. Resolves #1959

Tapping close to the trailing edge of an image places the selection correctly, but the caret isn't displayed. We had the same issue in the past on desktop. It was fixed by https://github.com/superlistapp/super_editor/pull/1820, but mobile doesn't use the same caret layer that desktop uses, so the issue is still present on mobile.

The root cause is the same of the desktop issue:

The downstream caret is positioned at (right, 0), so when the image takes all the available screen width (like in the demo app), the caret is positioned off-screen. Block level elements that have padding don't have this issue, because there is room to render the caret.

The same solution was applied:

This PR changes the caret overlay to check whether the caret rect is placed outside of its bounds at the right side. If it is, the caret position is adjusted. 

The golden  tests for Android are still failing due to this exception:

```
The following assertion was thrown running a test:
'package:flutter_test/src/_matchers_io.dart': Failed assertion: line 29 pos 10:
'!renderObject.debugNeedsPaint': is not true.

#2      captureImage (package:flutter_test/src/_matchers_io.dart:29:10)
#3      MatchesGoldenFile.matchAsync (package:flutter_test/src/_matchers_io.dart:96:21)
#4      _expect (package:matcher/src/expect/expect.dart:109:26)
#5      expectLater (package:matcher/src/expect/expect.dart:73:5)
#6      expectLater (package:flutter_test/src/widget_tester.dart:511:25)
#7      compareWithGolden (package:golden_toolkit/src/testing_tools.dart:296:9)
```

It seems to be caused by the `Follower` used by the Android collapsed handle.


angelosilvestre:
@matthew-carroll Can you do a first pass review? 

I still don't know what to do with this assertion failure, but it should impact the fix.

angelosilvestre:
@matthew-carroll We had the golden test issue on the past on linux and worked around that with the following code:

```dart
 @override
  void markNeedsPaint() {
    super.markNeedsPaint();

    if (kDebugMode &&
        !kIsWeb &&
        Platform.isLinux &&
        Platform.environment.containsKey('FLUTTER_TEST') &&
        !WidgetsBinding.instance.hasScheduledFrame) {
      // We are running on a linux test and we don't have a scheduled frame.
      //
      // We ran into an issue in some golden tests, on linux only, where `debugNeedsPaint`
      // is `true` after the build/layout/phase pipeline. This causes the tests to fail,
      // because trying to capture the image of a `RenderObject` throws an assertion failure
      // if `debugNeedsPaint` is `true`. To avoid that, immediately schedule a new frame,
      // so Flutter pipeline runs again before we try to capture the image.
      WidgetsBinding.instance.scheduleFrame();
    }
  }
```

Should we also check for Android?

matthew-carroll:
@angelosilvestre I think we mentioned it on our call, but you can go ahead and let that particular check run for all platforms.

matthew-carroll:
CC @BazinC

angelosilvestre:
@matthew-carroll I'm getting a failure on the new goldens on CI only that seems related to some antialiasing on the drag handle.

Test image:

super-editor-image-caret-upstream-android_testImage

Master image:

super-editor-image-caret-upstream-android_masterImage

Isolate diff:

super-editor-image-caret-upstream-android_isolatedDiff

Any ideas?

matthew-carroll:
@angelosilvestre can you just regenerate that golden?

angelosilvestre:
@matthew-carroll Locally, the regenerated golden is identical to the previous one...

matthew-carroll:
@angelosilvestre if your local Flutter version is identical to the CI Flutter version, then I'm not sure why we're seeing different results. If that's truly the case, then consider using the tooling that we created, which allows a certain number of pixels to be different before failing the test.

angelosilvestre:
After using our tooling that tolerates small differences, I got back to the infamous `debugNeedsPaint` assertion failure.

matthew-carroll:
@angelosilvestre are the goldens the only remaining problem? If so, let's file an issue to look into it and get this fix merged.

angelosilvestre:
@matthew-carroll Yes, the only problem is the failing goldens on Android. I modified these tests to be skipped so we can try to find out the reason.

matthew-carroll (approved):
LGTM

matthew-carroll:
@angelosilvestre please be sure to file an issue, if you haven't already

super_editor/lib/src/infrastructure/platforms/android/android_document_controls.dart
```diff
@@ -313,8 +313,26 @@ class AndroidControlsDocumentLayerState
...
+
+      // Default caret width used by the Android caret.
+      const caretWidth = 2;
+
+      final layerBox = context.findRenderObject() as RenderBox?;
+      if (layerBox != null && layerBox.hasSize && caretRect.left + caretWidth >= layerBox.size.width) {
+        // Ajust the caret position to make it entirely visible because it's currently placed
+        // partially or entirely outside of the layers' bounds. This can happen for downstream selections
+        // of block components that take all the available width.
+        caretRect = Rect.fromLTWH(
```
knopp, line 326: @angelosilvestre, I don't think this is correct thing to do. `computeLayoutDataWithDocumentLayout` is called during build, but `ContentLayers` first calls build on layers and only then layouts them. So when you're accessing the layerBox, the box itself has not been laid out yet for the frame.

PR #1971, opened 2024-04-25
[SuperEditor][SuperTextField] Remove NonReparentingFocus and FocusWithCustomParent (Resolves #1963)

### LINKED ISSUE #1963: [SuperEditor][SuperReader][SuperTextField] - Attempt to get rid of `NonParentingFocus` and `FocusWithCustomParent`
We created `NonParentingFocus` and `FocusWithCustomParent` to work around focus parenting limitations in Flutter. This was especially (or specifically) a concern with UI elements displayed in the application overlay, which needed to share focus with `SuperEditor` or `SuperReader` or `SuperTextField` focus. For example, a popover toolbar needs to hold focus so that the user can tab between buttons and/or type a URL for a link - but at the same time, `SuperEditor` must continue to also hold focus because `SuperEditor` has the caret, waiting for more input.

Flutter has made some changes to how focus is parented. Flutter has also embraced `OverlayPortal`, which internally seems to share focus across disjointed pieces of the widget tree.

For this ticket, attempt to remove our custom focus widgets, instead relying only on Flutter's built-in widgets. When doing this, make note of any loss in general capability. For example, while Super Editor can force itself to always use `OverlayPortal`, we can't necessarily force outside users to use `OverlayPortal`. Are there any places where we depend upon shared focus where an outside developer could still setup a widget tree that breaks Super Editor?

angelosilvestre:
In the past, we had some issues when trying to share the editor focus with popovers displayed in overlays, because they are not in the same subtree. Because of the different subtrees, the popover focus wasn’t naturally a child of the editor focus in the focus tree.

To work around that, we introduced `NonParentingFocus`, which is used solely by `SuperTextfield` to avoid natural focus reparenting, and `FocusWithCustomParent` to make it possible to force the popover focus node to be a child of the editor focus node.

Since them, Flutter introduced `OverlayPortal`, which makes it possible to share focus with widgets displayed in the overlay, and added a `parentNode` property to `Focus`, with the same purpose as our `FocusWithCustomParent` .

With these Flutter changes, our widgets aren’t necessary anymore. Even if developers use `SuperEditor` and show popovers using `OverlayEntry`, they can setup focus sharing by wrapping the popover with a `Focus` and setting the new `parentNode` property.

We already have a number of tests related to focus sharing, including tests that use `OverlayEntry` and they are still pass after removing `NonParentingFocus` from the `SuperTextField` subtree.

The usage of `OverlayEntry` is the only thing I can think that could cause problems with focus sharing. But since we can setup focus sharing by using a regular `Focus`, I believe it’s safe for us to just remove `NonParentingFocus` and `FocusWithCustomParent`.

However, while I was testing I got the following crash when showing the URL field in the demo app popover toolbar:

```console
00:08 +502 -2: /Users/angelosilvestre/dev/super_editor/super_editor/test/super_editor/supereditor_selection_test.dart: SuperEditor selection by default keeps IME connection open when it loses primary focus (on Linux)
The following assertion was thrown building SuperTextFieldKeyboardInteractor(state:
_SuperTextFieldKeyboardInteractorState#dd121):
Tried to make a child into a parent of itself.
'package:flutter/src/widgets/focus_manager.dart':
Failed assertion: line 1013 pos 12: 'child != this'

  SuperTextFieldKeyboardInteractor
  SuperTextFieldKeyboardInteractor:file:///Users/angelosilvestre/dev/super_editor/super_editor/lib/src/super_textfield/desktop/desktop_textfield.dart:465:14

#2      FocusNode._reparent (package:flutter/src/widgets/focus_manager.dart:1013:12)
#3      FocusAttachment.reparent (package:flutter/src/widgets/focus_manager.dart:255:14)
#4      _FocusState.didChangeDependencies (package:flutter/src/widgets/focus_scope.dart:575:23)
#5      StatefulElement._firstBuild (package:flutter/src/widgets/framework.dart:5630:11)
#6      ComponentElement.mount (package:flutter/src/widgets/framework.dart:5457:5)
...    
```

This happens because there are places where we are using the same FocusNode that we use to setup focus sharing as the SuperTextField focus node. For example:

```dart
Focus(
  focusNode: textFieldFocusNode,
  parentFocusNode: editorFocusNode,
  child: SuperTextField(
    focusNode: textFieldFocusNode,
  )
)
```

`SuperTextField` also places a `Focus` widget in the tree, so it seems we are violating a Flutter rule by using the same `FocusNode` in two different `Focus` widgets. It makes sense, since there is a focus tree a node can’t be child of itself.

To solve this, we should use a different `FocusNode` to setup focus sharing. For example:

```dart
Focus(
  focusNode: popoverFocusNode,
  parentFocusNode: editorFocusNode,
  child: SuperTextField(
    focusNode: textFieldFocusNode,
  )
)
```

It’s possible that apps have this same pattern, so this change can cause apps to crash.

Some existing tests related to focus sharing:

https://github.com/superlistapp/super_editor/blob/76c6562fd5e1665cb978886cfb9df2d5bad5cbb0/super_editor/test/super_editor/supereditor_popover_focus_test.dart#L12-L13

This test shows a popover in an `OverlayEntry`: 

https://github.com/superlistapp/super_editor/blob/76c6562fd5e1665cb978886cfb9df2d5bad5cbb0/super_editor/test/super_editor/supereditor_popover_focus_test.dart#L107-L122

Another test:

https://github.com/superlistapp/super_editor/blob/76c6562fd5e1665cb978886cfb9df2d5bad5cbb0/super_editor/test/super_editor/supereditor_selection_test.dart#L616C38-L654

This test sets up focus sharing with a different subtree without an `OverlayEntry`. This is one of the tests that failed because we are passing `textFieldFocus` to two `Focus`  widgets.


angelosilvestre:
cc @knopp

knopp:
This looks great. It's probably a bad idea to have multiple focus attachments for a single node so it's good to fix that too.

matthew-carroll (approved):
LGTM

Issue #1957, opened 2024-04-22
magnifier magnifies the wrong area


BazinC:
**Package Version**
stable or main branch

**To Reproduce**
On ios, long press on text to display magnifier. 

**Actual behavior**
See video below. simulator (iPhone 15 iOS 17.0) running super editor example. magnified area is off. It looks like it magnifies the content below the cursor.

[supereditor ios magnifier.webm](https://github.com/superlistapp/super_editor/assets/6229343/de218bc7-7fb4-41d9-b648-140ccb1897b4)

**Expected behavior**

[native ios magnifier.webm](https://github.com/superlistapp/super_editor/assets/6229343/895f63b4-525e-4f92-91a8-b40243ea2377)

Same with Android simulator (Pixel 7 Pro API 32):
![Screenshot_1713780723](https://github.com/superlistapp/super_editor/assets/6229343/77effa40-f1a6-49b2-8c56-3a0a03e7bdd8)

**Platform**
iOS and Android

**Flutter version**
3.19.1

**Additional context**
I suggest the super_editor default magnifier to use or take inspiration from Flutter CupertinoTextMagnifier/TextMagnifier

angelosilvestre:
@BazinC Are you using some specific settings to trigger this issue on Android? I was able to reproduce it on iOS only.

BazinC:
@angelosilvestre For Android I'm just running the demo app, from stable branch, on an Android Simulator (Pixel 7 Pro API 32).
I don't have a physical Android device with me right now. I can try later when I have one.

<img width="1721" alt="image" src="https://github.com/superlistapp/super_editor/assets/6229343/99008ca6-0539-4b97-9a6f-40e06e8fe699">

matthew-carroll:
@angelosilvestre @BazinC - It looks like we're seeing unpredictable API level problems for Android:

 - Pixel 7 Pro emulator, API 32: The magnification region is too low
 - Pixel 7 Pro emulator, API 33: The magnification region is correct
 - Pixel 7 Pro emulator, API 34: The magnification region is too low

angelosilvestre:
For the Pixel 7 Pro emulator, the `devicePixelRatio` is 3.5, and `Offset(0, -22 * MediaQuery.of(context).devicePixelRatio)` doesn't produce correct results:

![image](https://github.com/superlistapp/super_editor/assets/7597082/20d9259f-a687-483f-b3b6-c804470b9f07)

[...]

PR #1973, opened 2024-04-28
[SuperEditor][SuperReader][SuperTextField] Update iOS magnifier (Resolves #1958)

### LINKED ISSUE #1958: iOS: magnifier entrance and exit animation, and shape
**User Behavior**
Match the iOS 17 native "Loupe" (magnifying glass). For now the magnifier entrance and exit animation, and shape doesn't match native component.

See below: iOS native behavior. Message app

[native ios magnifier.webm](https://github.com/superlistapp/super_editor/assets/6229343/0825347e-4264-4819-a513-df6c2702e542)

- Bigger loupe
- Border to better sepateration between overlay and  content behind
- Border color matches the cursor color
- Entrance/Exit animation

See below: iOS SuperEditor behavior. Example app. Use IOSFollowingMagnifier.roundedRectangle

[supereditor ios magnifier.webm](https://github.com/superlistapp/super_editor/assets/6229343/62d2f692-571f-4884-bd1b-2b15a3e8c13b)

See below: Flutter Textfield on iOS. Even flutter doesn't match native.

[flutter textfield magnifier.webm](https://github.com/superlistapp/super_editor/assets/6229343/db84d118-48c3-458d-af14-2b52b7792874)

**Feature Value**
iOS users love to have the same "native" UI/UX feel. They can complain when the experience is different and tell you're app doesn't feel polish.

**Additional context**
https://www.macobserver.com/tips/how-to/text-magnification-ios-17/
https://developer.apple.com/videos/play/wwdc2023/10058/

angelosilvestre:
[SuperEditor][SuperReader][SuperTextField] Update iOS magnifier. Resolves #1958

The current appearance and behavior of our iOS magnifier doesn't match the native appearance and behavior.

Native iOS:

[native ios magnifier.webm](https://github.com/superlistapp/super_editor/assets/6229343/0825347e-4264-4819-a513-df6c2702e542)

SuperEditor:

[supereditor ios magnifier.webm](https://github.com/superlistapp/super_editor/assets/6229343/62d2f692-571f-4884-bd1b-2b15a3e8c13b)

This PR adds entrance/exit animations and updates the size and shape of the magnifier:

https://github.com/superlistapp/super_editor/assets/7597082/f0de708a-2efa-4f58-b74c-05515f371e6c

Here's the appearance with slow animations:

https://github.com/superlistapp/super_editor/assets/7597082/a83a351c-5349-4a89-bab2-bbff88391a75

This PR also includes the fix for the issue iOS described in https://github.com/superlistapp/super_editor/issues/1957

### DIFF
--- a/super_editor/example/lib/demos/editor_configs/demo_mobile_editing_ios.dart
+++ b/super_editor/example/lib/demos/editor_configs/demo_mobile_editing_ios.dart
@@ -14,7 +14,7 @@ class MobileEditingIOSDemo extends StatefulWidget {
   State<MobileEditingIOSDemo> createState() => _MobileEditingIOSDemoState();
 }
 
-class _MobileEditingIOSDemoState extends State<MobileEditingIOSDemo> {
+class _MobileEditingIOSDemoState extends State<MobileEditingIOSDemo> with SingleTickerProviderStateMixin {
   final GlobalKey _docLayoutKey = GlobalKey();
 
   late MutableDocument _doc;
@@ -133,12 +133,14 @@ class _MobileEditingIOSDemoState extends State<MobileEditingIOSDemo> {
     );
   }
 
-  Widget _buildIosMagnifier(BuildContext context, Key magnifierKey, LeaderLink focalPoint) {
+  Widget _buildIosMagnifier(
+      BuildContext context, Key magnifierKey, LeaderLink focalPoint, AnimationController animationController) {
     return Center(
       child: IOSFollowingMagnifier.roundedRectangle(
         magnifierKey: magnifierKey,
         leaderLink: focalPoint,
-        offsetFromFocalPoint: const Offset(0, -72),
+        offsetFromFocalPoint: const Offset(0, -230),
+        animationController: animationController,
       ),
     );
   }
--- a/super_editor/example/lib/demos/editor_configs/demo_mobile_editing_ios.dart
+++ b/super_editor/example/lib/demos/editor_configs/demo_mobile_editing_ios.dart
@@ -133,14 +133,13 @@ class _MobileEditingIOSDemoState extends State<MobileEditingIOSDemo> with Single
     );
   }
 
-  Widget _buildIosMagnifier(
-      BuildContext context, Key magnifierKey, LeaderLink focalPoint, AnimationController animationController) {
+  Widget _buildIosMagnifier(BuildContext context, Key magnifierKey, LeaderLink focalPoint, bool visible) {
     return Center(
       child: IOSFollowingMagnifier.roundedRectangle(
         magnifierKey: magnifierKey,
         leaderLink: focalPoint,
         offsetFromFocalPoint: const Offset(0, -230),
-        animationController: animationController,
+        show: visible,
       ),
     );
   }
--- a/super_editor/example/lib/demos/editor_configs/demo_mobile_editing_ios.dart
+++ b/super_editor/example/lib/demos/editor_configs/demo_mobile_editing_ios.dart
@@ -133,13 +133,16 @@ class _MobileEditingIOSDemoState extends State<MobileEditingIOSDemo> with Single
     );
   }
 
-  Widget _buildIosMagnifier(BuildContext context, Key magnifierKey, LeaderLink focalPoint, bool visible) {
+  Widget _buildIosMagnifier(BuildContext context, Key magnifierKey, LeaderLink focalPoint, bool isVisible) {
     return Center(
       child: IOSFollowingMagnifier.roundedRectangle(
         magnifierKey: magnifierKey,
         leaderLink: focalPoint,
-        offsetFromFocalPoint: const Offset(0, -230),
-        show: visible,
+        // The bottom of the magnifier sits above the focal point.
+        // Leave a few pixels between the bottom of the magnifier and the focal point. This
+        // value was chosen empirically.
+        offsetFromFocalPoint: const Offset(0, -20),
+        show: isVisible,
       ),
     );
   }
--- a/super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
@@ -1893,5 +1893,5 @@ class SuperEditorIosHandlesDocumentLayerBuilder implements SuperEditorLayerBuild
   }
 }
 
-const defaultIosMagnifierAnimationDuration = Duration(milliseconds: 150);
-const defaultIosMagnifierExitAnimationDuration = Duration(milliseconds: 100);
+const defaultIosMagnifierAnimationDuration = Duration(milliseconds: 180);
+const defaultIosMagnifierExitAnimationDuration = Duration(milliseconds: 150);
--- a/super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
@@ -1507,7 +1507,9 @@ class SuperEditorIosMagnifierOverlayManagerState extends State<SuperEditorIosMag
       magnifierKey: magnifierKey,
       leaderLink: magnifierFocalPoint,
       show: visible,
-      offsetFromFocalPoint: const Offset(0, -230),
+      // The bottom of the magnifier sits above the focal point.
+      // Leave a few pixels between the bottom of the magnifier and the focal point.
+      offsetFromFocalPoint: const Offset(0, -20),
     );
   }
 }
@@ -1853,3 +1855,4 @@ class SuperEditorIosHandlesDocumentLayerBuilder implements SuperEditorLayerBuild
 const defaultIosMagnifierEnterAnimationDuration = Duration(milliseconds: 180);
 const defaultIosMagnifierExitAnimationDuration = Duration(milliseconds: 150);
 const defaultIosMagnifierAnimationCurve = Curves.easeInOut;
+const defaultIosMagnifierSize = Size(133, 96);
--- a/super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
@@ -283,7 +283,9 @@ class _IosDocumentTouchInteractorState extends State<IosDocumentTouchInteractor>
   Offset? _startDragPositionOffset;
   double? _dragStartScrollOffset;
   Offset? _globalDragOffset;
-  final _magnifierFocalPoint = ValueNotifier<Offset?>(null);
+
+  /// The [Offset] of the magnifier's focal point in the [DocumentLayout] coordinate space.
+  final _magnifierFocalPointInContentSpace = ValueNotifier<Offset?>(null);
   Offset? _dragEndInInteractor;
   DragMode? _dragMode;
   // TODO: HandleType is the wrong type here, we need collapsed/base/extent,
@@ -577,7 +579,7 @@ class _IosDocumentTouchInteractorState extends State<IosDocumentTouchInteractor>
       return;
     }
 
-    _updateMagnifierFocalPoint();
+    _placeFocalPointNearTouchOffset();
     _controlsController!
       ..hideToolbar()
       ..showMagnifier();
@@ -957,7 +959,7 @@ class _IosDocumentTouchInteractorState extends State<IosDocumentTouchInteractor>
       dragEndInViewport: dragEndInViewport,
     );
 
-    _updateMagnifierFocalPoint();
+    _placeFocalPointNearTouchOffset();
   }
 
   void _updateSelectionForNewDragHandleLocation() {
@@ -1084,8 +1086,8 @@ class _IosDocumentTouchInteractorState extends State<IosDocumentTouchInteractor>
   }
 
   void _updateMagnifierFocalPointOnAutoScrollFrame() {
-    if (_magnifierFocalPoint.value != null) {
-      _updateMagnifierFocalPoint();
+    if (_magnifierFocalPointInContentSpace.value != null) {
+      _placeFocalPointNearTouchOffset();
     }
   }
 
@@ -1237,7 +1239,7 @@ class _IosDocumentTouchInteractorState extends State<IosDocumentTouchInteractor>
   }
 
   /// Updates the magnifier focal point in relation to the current drag position.
-  void _updateMagnifierFocalPoint() {
+  void _placeFocalPointNearTouchOffset() {
     late DocumentPosition? docPositionToMagnify;
 
     if (_globalTapDownOffset != null) {
@@ -1255,7 +1257,7 @@ class _IosDocumentTouchInteractorState extends State<IosDocumentTouchInteractor>
       _docLayout.getRectForPosition(docPositionToMagnify!)!.center,
     );
 
-    _magnifierFocalPoint.value = centerOfContentAtOffset;
+    _magnifierFocalPointInContentSpace.value = centerOfContentAtOffset;
   }
 
   @override
@@ -1333,7 +1335,7 @@ class _IosDocumentTouchInteractorState extends State<IosDocumentTouchInteractor>
 
   Widget _buildMagnifierFocalPoint() {
     return ValueListenableBuilder(
-      valueListenable: _magnifierFocalPoint,
+      valueListenable: _magnifierFocalPointInContentSpace,
       builder: (context, magnifierFocalPoint, child) {
         if (magnifierFocalPoint == null) {
           return const SizedBox();
@@ -1456,7 +1458,7 @@ class SuperEditorIosMagnifierOverlayManagerState extends State<SuperEditorIosMag
   ///
   /// This is different from [SuperEditorIosControlsController.shouldShowMagnifier] because
   /// [_showMagnifier] becomes false only when the exit animation finishes.
-  late ValueNotifier<bool> _showMagnifier;
+  final ValueNotifier<bool> _showMagnifier = ValueNotifier(false);
 
   @visibleForTesting
   bool get wantsToDisplayMagnifier => _controlsController!.shouldShowMagnifier.value;
@@ -1466,7 +1468,7 @@ class SuperEditorIosMagnifierOverlayManagerState extends State<SuperEditorIosMag
     super.initState();
     _animationController = AnimationController(
       vsync: this,
-      duration: defaultIosMagnifierAnimationDuration,
+      duration: defaultIosMagnifierEnterAnimationDuration,
       reverseDuration: defaultIosMagnifierExitAnimationDuration,
     );
     _animationController.addStatusListener(_hideMagnifierOnAnimationEnd);
@@ -1478,7 +1480,7 @@ class SuperEditorIosMagnifierOverlayManagerState extends State<SuperEditorIosMag
 
     _controlsController = SuperEditorIosControlsScope.rootOf(context);
     _controlsController!.shouldShowMagnifier.addListener(_onShouldShowMagnifierChanged);
-    _showMagnifier = ValueNotifier(_controlsController!.shouldShowMagnifier.value);
+    _showMagnifier.value = _controlsController!.shouldShowMagnifier.value;
 
     _overlayPortalController.show();
   }
@@ -1893,5 +1895,6 @@ class SuperEditorIosHandlesDocumentLayerBuilder implements SuperEditorLayerBuild
   }
 }
 
-const defaultIosMagnifierAnimationDuration = Duration(milliseconds: 180);
+const defaultIosMagnifierEnterAnimationDuration = Duration(milliseconds: 180);
 const defaultIosMagnifierExitAnimationDuration = Duration(milliseconds: 150);
+const defaultIosMagnifierAnimationCurve = Curves.easeInOut;
--- a/super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
@@ -160,7 +160,7 @@ class SuperEditorIosControlsController {
   /// (Optional) Builder to create the visual representation of the magnifier.
   ///
   /// If [magnifierBuilder] is `null`, a default iOS magnifier is displayed.
-  final DocumentAnimatedMagnifierBuilder? magnifierBuilder;
+  final DocumentMagnifierBuilder? magnifierBuilder;
 
   /// Whether the iOS floating toolbar should be displayed right now.
   ValueListenable<bool> get shouldShowToolbar => _shouldShowToolbar;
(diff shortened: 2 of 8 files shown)


matthew-carroll:
@angelosilvestre can you show a zoomed out video of iOS with slow animations that clearly shows the appearance and disappearance of the native iOS magnifier? It's difficult to compare the zoomed out Super Editor version against that initial iOS video.

I'm curious if the Super Editor magnifier border should really be that wide. Also, given that we can't render a true squircle shape, I'm wondering if our intro and exit animation should go from a small circle to a rounded rectangle. Or, similarly, keep the same corner radius the whole time, which will naturally start as a circle and then become a rounded rectangle.

matthew-carroll:
CC @BazinC

angelosilvestre:
@matthew-carroll I updated the animation. At the beginning of the animation the shape looks like more with a pill

https://github.com/superlistapp/super_editor/assets/7597082/8a0cf6ec-0049-4135-a738-8ddfec628b90

With slow animations:

https://github.com/superlistapp/super_editor/assets/7597082/f1c207c8-042d-45f7-88d9-f5d9214aca88

super_editor/example/lib/demos/editor_configs/demo_mobile_editing_ios.dart
```diff
@@ -133,12 +133,14 @@ class _MobileEditingIOSDemoState extends State<MobileEditingIOSDemo> {
...
 
-  Widget _buildIosMagnifier(BuildContext context, Key magnifierKey, LeaderLink focalPoint) {
+  Widget _buildIosMagnifier(
+      BuildContext context, Key magnifierKey, LeaderLink focalPoint, AnimationController animationController) {
     return Center(
       child: IOSFollowingMagnifier.roundedRectangle(
         magnifierKey: magnifierKey,
         leaderLink: focalPoint,
-        offsetFromFocalPoint: const Offset(0, -72),
+        offsetFromFocalPoint: const Offset(0, -230),
```
matthew-carroll, line 142: Why did the magic number change? There probably shouldn't have been a magic number in the first place, so let's at least comment where that number came from and how someone can get a new one, if needed.

On a related note, did you check what happens on desktop when using the magnifier? I'm wondering if the reason you had to change this number was due to pixel density. If so, it shouldn't be a constant.

super_editor/example/lib/demos/editor_configs/demo_mobile_editing_ios.dart
```diff
@@ -133,12 +133,14 @@ class _MobileEditingIOSDemoState extends State<MobileEditingIOSDemo> {
...
+        animationController: animationController,
```
matthew-carroll, line 143: Is there a reason we pass the `AnimationController` from the outside? Isn't the animation always the same? If so, it seems excessive to make the client responsible for configuring an animation and running it. Is there any reason that this widget shouldn't operate like an implicitly animated widget?

If you want optional outside control for verification purposes, then the controller should be optional and documented as such.

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
```diff
@@ -283,7 +283,7 @@ class _IosDocumentTouchInteractorState extends State<IosDocumentTouchInteractor>
   Offset? _startDragPositionOffset;
   double? _dragStartScrollOffset;
   Offset? _globalDragOffset;
-  final _magnifierOffset = ValueNotifier<Offset?>(null);
+  final _magnifierFocalPoint = ValueNotifier<Offset?>(null);
```
matthew-carroll, line 286: The new name for this property doesn't make it clear whether this is an absolute position, or whether this is a delta. Please double check the name, and also add a comment for clarity.

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
```diff
@@ -1240,6 +1236,28 @@ class _IosDocumentTouchInteractorState extends State<IosDocumentTouchInteractor>
     });
   }
 
+  /// Updates the magnifier focal point in relation to the current drag position.
+  void _updateMagnifierFocalPoint() {
```
matthew-carroll, line 1240: Notice that the Dart Doc for this method includes relevant details that aren't available in the name of the method. In general, naming something like "updateThing" is a pain because nobody can know what it means until they read the implementation or Dart Doc.

Instead, you can name this `_placeFocalPointNearTouchOffset` and now it gives everyone a pretty good idea about what it actually does.

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
```diff
@@ -1240,6 +1236,28 @@ class _IosDocumentTouchInteractorState extends State<IosDocumentTouchInteractor>
...
+    late DocumentPosition? docPositionToMagnify;
+
```
matthew-carroll, line 1242: What ended up being the issue that caused this change in code? What were we doing wrong before?

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
```diff
@@ -1427,20 +1445,69 @@ class SuperEditorIosMagnifierOverlayManager extends StatefulWidget {
...
+    _animationController.addStatusListener(_hideMagnifierOnAnimationEnd);
+  }
+
   @override
   void didChangeDependencies() {
     super.didChangeDependencies();
 
     _controlsController = SuperEditorIosControlsScope.rootOf(context);
+    _controlsController!.shouldShowMagnifier.addListener(_onShouldShowMagnifierChanged);
+    _showMagnifier = ValueNotifier(_controlsController!.shouldShowMagnifier.value);
```
matthew-carroll, line 1481: If you re-instantiate this notifier on every call of htis method then you're likely to end up with things that registered as listeners on older instances that don't have any reference to the new instance.

Is there any reason that you don't instantiate this notifier at its declaration point and give it a value of `false`, and then update that value here?

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
```diff
@@ -1824,3 +1892,6 @@ class SuperEditorIosHandlesDocumentLayerBuilder implements SuperEditorLayerBuild
     );
   }
 }
+
+const defaultIosMagnifierAnimationDuration = Duration(milliseconds: 180);
```
matthew-carroll, line 1896: What about the default curve? I would guess that iOS isn't using a linear curve.

BTW, didn't we determine that iOS disappears faster than it appears?

super_editor/lib/src/infrastructure/platforms/ios/magnifier.dart
```diff
@@ -31,35 +36,81 @@ class IOSFollowingMagnifier extends StatelessWidget {
...
+        return Follower.withOffset(
+          link: widget.leaderLink,
+          leaderAnchor: Alignment.center,
+          followerAnchor: Alignment.topLeft,
+          offset: Offset(
+            widget.offsetFromFocalPoint.dx,
+            // Animate the magnfier up on entrance and down on exit.
+            widget.offsetFromFocalPoint.dy * percentage,
+          ),
+          // Theoretically, we should be able to use a leaderAnchor and followerAnchor of "center"
```
matthew-carroll, line 70: Then let's try to fix this in follow_the_leader instead of hacking it here.

super_editor/lib/src/infrastructure/platforms/ios/magnifier.dart
```diff
@@ -155,3 +234,16 @@ class IOSCircleMagnifyingGlass extends StatelessWidget {
...
+/// Parameters used to render the magnifier.
+class MagnifierInfo {
+  MagnifierInfo({
+    required this.offsetFromFocalPoint,
+    this.animationPercentage = 100.0,
+    required this.borderColor,
+  });
+
+  final Offset offsetFromFocalPoint;
+  final double animationPercentage;
```
matthew-carroll, line 247: It doesn't make a lot of sense to have a concept of "animation percentage". It begs the question "what animation"? Moreover, as mentioned above, we might actually need different intro vs exit animations.

Instead, you should capture the individual properties that you care about and then let the animation code handle the responsibility of translating animation progress into magnifier properties.

super_editor/lib/src/infrastructure/platforms/ios/magnifier.dart
```diff
@@ -155,3 +234,16 @@ class IOSCircleMagnifyingGlass extends StatelessWidget {
     );
   }
 }
+
+/// Parameters used to render the magnifier.
+class MagnifierInfo {
```
matthew-carroll, line 239: Should this be public? Let's call this `IosMagnifierViewModel`.

super_editor/lib/src/infrastructure/platforms/mobile_documents.dart
```diff
@@ -132,6 +132,10 @@ typedef DocumentFloatingToolbarBuilder = Widget Function(
 /// ```
 typedef DocumentMagnifierBuilder = Widget Function(BuildContext, Key magnifierKey, LeaderLink focalPoint);
 
+/// A [DocumentMagnifierBuilder] that animates the magnifier's appearance and disappearance.
```
matthew-carroll, line 135: Does Android have intro and exit animations, too? If so, we should probably migrate both magnifiers to the new signature, and keep the existing builder name, instead of introducing a second one.

Moreover, this builder should probably take two optional values, perhaps `introPercent` and `exitPercent`. Only one is provided at a given time.

```
typedef DocumentMagnifierBuilder = Widget Function(buildContext, Key magnifierKey, LeaderLink focalPoint, {double? introPercent, double? exitPercent});
```

Would that work?

super_editor/example/lib/demos/editor_configs/demo_mobile_editing_ios.dart
angelosilvestre: > Why did the magic number change? 

Since the magnifier size changed, this value needs to be adjusted to prevent the magnifier from covering the caret. This value was chosen empirically...

> On a related note, did you check what happens on desktop when using the magnifier? I'm wondering if the reason you had to change this number was due to pixel density. If so, it shouldn't be a constant.

Yeah, it doesn't look right when running on desktop:

https://github.com/superlistapp/super_editor/assets/7597082/cc9eeb3c-230a-4044-aa61-aded5c640716

super_editor/example/lib/demos/editor_configs/demo_mobile_editing_ios.dart
angelosilvestre: Should we pass the animation value as `ValueListenable` here? Just the percentage?

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
angelosilvestre: Updated

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
angelosilvestre: This code is here to fix the magnification area issue described in https://github.com/superlistapp/super_editor/issues/1957

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
angelosilvestre: Yeah, We have another constant for the exit animation.

super_editor/lib/src/infrastructure/platforms/ios/magnifier.dart
angelosilvestre: Would it make more sense to have the animation progress and the animation direction (forward/reverse) ?

super_editor/lib/src/infrastructure/platforms/mobile_documents.dart
angelosilvestre: > Moreover, this builder should probably take two optional values, perhaps introPercent and exitPercent. Only one is provided at a given time.

Wouldn't it be better if we provide only one percent and the direction of the animation?

super_editor/example/lib/demos/editor_configs/demo_mobile_editing_ios.dart
matthew-carroll: The question is "why?". What kind of control is needed? Would it suffice to simply pass a `bool` to a property called `show`? Could the magnifier then show/hide itself?

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
matthew-carroll: Ok, but I asked what ended up being the issue? What was the root problem that you found and fixed with that code?

super_editor/lib/src/infrastructure/platforms/mobile_documents.dart
matthew-carroll: That should be fine

super_editor/lib/src/infrastructure/platforms/ios/magnifier.dart
matthew-carroll: Ignore this comment if you copied the commented code from Android or elsewhere.

super_editor/example/lib/demos/editor_configs/demo_mobile_editing_ios.dart
angelosilvestre: > Would it suffice to simply pass a bool to a property called show? Could the magnifier then show/hide itself?

Yeah, it can. Updated to do that.

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
angelosilvestre: The root cause is that we were using the finger position as the focal point.

super_editor/lib/src/infrastructure/platforms/mobile_documents.dart
angelosilvestre: > Does Android have intro and exit animations, too?

Android doesn't seem to have intro and exit animations.

angelosilvestre:
@matthew-carroll I tried to simplify the usage of magic numbers and adapted the code to take the pixel density into consideration:

https://github.com/superlistapp/super_editor/assets/7597082/0ad0f4a9-0cfc-4938-b9cb-87c99d056ddb

With slow animations:

https://github.com/superlistapp/super_editor/assets/7597082/afec3b0c-9ba3-401f-a839-d6dae78667a3

super_editor/lib/src/super_reader/read_only_document_ios_touch_interactor.dart
```diff
@@ -1131,28 +1152,34 @@ class SuperReaderIosMagnifierOverlayManagerState extends State<SuperReaderIosMag
...
       // Defer to the browser to display overlay controls on mobile.
       return const SizedBox();
     }
 
     return IOSFollowingMagnifier.roundedRectangle(
       magnifierKey: magnifierKey,
+      show: visible,
       leaderLink: magnifierFocalPoint,
-      offsetFromFocalPoint: const Offset(0, -72),
+      offsetFromFocalPoint: const Offset(0, -230),
```
matthew-carroll, line 1182: Should this be based on device pixel ratio?

super_editor/lib/src/super_textfield/ios/_editing_controls.dart
```diff
@@ -543,11 +528,15 @@ class _IOSEditingControlsState extends State<IOSEditingControls> with WidgetsBin
...
-        leaderLink: widget.editingController.magnifierFocalPoint,
-        offsetFromFocalPoint: const Offset(0, -72),
-      ),
+    return ValueListenableBuilder(
+      valueListenable: widget.editingController.shouldShowMagnifier,
+      builder: (context, showMagnifier, child) {
+        return IOSFollowingMagnifier.roundedRectangle(
+          leaderLink: widget.editingController.magnifierFocalPoint,
+          show: showMagnifier,
+          offsetFromFocalPoint: const Offset(0, -230),
```
matthew-carroll, line 537: Should this use device pixel ratio?

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
```diff
@@ -1006,7 +1008,6 @@ class _IosDocumentTouchInteractorState extends State<IosDocumentTouchInteractor>
   }
 
   void _onPanEnd(DragEndDetails details) {
-    _magnifierOffset.value = null;
```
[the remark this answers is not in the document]
angelosilvestre: I don't think it's necessary. We already call `_controlsController.hideMagnifier()`.

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
```diff
@@ -283,7 +283,9 @@ class _IosDocumentTouchInteractorState extends State<IosDocumentTouchInteractor>
   Offset? _startDragPositionOffset;
   double? _dragStartScrollOffset;
   Offset? _globalDragOffset;
-  final _magnifierOffset = ValueNotifier<Offset?>(null);
+
+  /// The [Offset] of the magnifier's focal point in the [DocumentLayout] coordinate space.
+  final _magnifierFocalPointInContentSpace = ValueNotifier<Offset?>(null);
```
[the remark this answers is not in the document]
angelosilvestre: Yes, updated.

super_editor/lib/src/infrastructure/platforms/ios/magnifier.dart
```diff
@@ -155,3 +292,18 @@ class IOSCircleMagnifyingGlass extends StatelessWidget {
...
+  IosMagnifierViewModel({
+    required this.offsetFromFocalPoint,
+    this.animationValue = 1.0,
+    this.animationDirection = AnimationStatus.forward,
+    required this.borderColor,
+  });
+
+  final Offset offsetFromFocalPoint;
+  final double animationValue;
+  final AnimationStatus animationDirection;
```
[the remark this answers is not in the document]
angelosilvestre: Updated

super_editor/lib/src/super_textfield/ios/_editing_controls.dart
angelosilvestre: This value is already adjusted to the pixel ratio in `IOSFollowingMagnifier`.

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
```diff
@@ -1482,7 +1506,10 @@ class SuperEditorIosMagnifierOverlayManagerState extends State<SuperEditorIosMag
     return IOSFollowingMagnifier.roundedRectangle(
       magnifierKey: magnifierKey,
       leaderLink: magnifierFocalPoint,
-      offsetFromFocalPoint: const Offset(0, -72),
+      show: visible,
+      // The bottom of the magnifier sits above the focal point.
+      // Leave a few pixels between the bottom of the magnifier and the focal point.
+      offsetFromFocalPoint: const Offset(0, -20),
```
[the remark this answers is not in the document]
angelosilvestre: This changed because the follower alignment was changed, so this number needed to be adjusted. It was chosen empirically.

super_editor/lib/src/infrastructure/platforms/ios/magnifier.dart
```diff
@@ -31,35 +36,139 @@ class IOSFollowingMagnifier extends StatelessWidget {
...
+            child: FractionalTranslation(
+              translation: Offset(0.0, -0.5 * percentage),
+              child: widget.magnifierBuilder(
+                context,
+                IosMagnifierViewModel(
+                  // In theory, the offsetFromFocalPoint should either be `widget.offsetFromFocalPoint.dy` to match
+                  // the actual offset, or it should be `widget.offsetFromFocalPoint.dy / magnificationLevel`. Neither
+                  // of those align the focal point correctly. The following offset was found empirically to give the
+                  // desired results. These values seem to work even with different pixel densities.
+                  offsetFromFocalPoint: Offset(
```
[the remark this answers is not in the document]
angelosilvestre: While I was testing this, the value didn't seem to be affected by the pixel ratio.

super_editor/example/lib/demos/editor_configs/demo_mobile_editing_ios.dart
angelosilvestre: This value is already adjusted to the pixel ratio in the magnifier widget.

super_editor/lib/src/super_reader/read_only_document_ios_touch_interactor.dart
angelosilvestre: This value is already adjusted to the pixel ratio in `IOSFollowingMagnifier`.

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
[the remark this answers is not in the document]
matthew-carroll: Ok, but what about pixel density?

super_editor/lib/src/infrastructure/platforms/ios/magnifier.dart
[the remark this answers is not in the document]
matthew-carroll: Do you know why it wasn't effected? That seems concerning.

super_editor/lib/src/super_reader/read_only_document_ios_touch_interactor.dart
matthew-carroll: What do you mean by adjusted? This is a hard-coded offset...

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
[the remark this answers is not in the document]
angelosilvestre: In the `_IOSFollowingMagnifierState` build method, we multiply the `offsetFromFocalPoint` by the device pixel ratio.

super_editor/lib/src/super_reader/read_only_document_ios_touch_interactor.dart
angelosilvestre: In the `_IOSFollowingMagnifierState` build method, we multiply the `offsetFromFocalPoint` by the device pixel ratio.

super_editor/lib/src/infrastructure/platforms/ios/magnifier.dart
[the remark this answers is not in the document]
angelosilvestre: I'm not sure why this is the case...

matthew-carroll:
@angelosilvestre It looks like there are multiple `Offset`s that I assumed were DIP values, which were actually pixel values. Can you please carefully document each such `Offset` so that the reader knows whether he's reading a DIP value or a pixel value. Otherwise, users won't know whether they have the responsibility to divide by the pixel density, or whether it's done for them.

matthew-carroll (approved):
LGTM - I'll go ahead and merge this for forward progress, but please come back and document the DPI vs pixel values when you get a chance.

PR #1978, opened 2024-04-30
[SuperEditor] Show software keyboard upon tap (Resolves #1816)

### LINKED ISSUE #1816: [Super Editor] [Android] Unable to open the keyboard once I hide it with the system button.
**Package Version**
Github stable version.

**To Reproduce**
1. Open the example app on Android device with option to hide the keybaord.
2. From the drawer menu select "Text with hint".
3. Click on the first "header goes here...".
4. After the keyboard is open press the system button to hide/close the keybaord.
5. After the keybaord is hidden, click again on the same (first one) "header goes here...".

**Actual behavior**
The keyboard stays hidden and I'm not able to enter the text. 

**Expected behavior**
I expect the keybaord to re-appear so I can add text.

**Platform**
Samsung A50 with Android 11

**Flutter version**
Flutter 3.16.4 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 2e9cb0aa71 (6 weeks ago) • 2023-12-11 14:35:13 -0700
Engine • revision 54a7145303
Tools • Dart 3.2.3 • DevTools 2.28.4

angelosilvestre:
[SuperEditor] Show software keyboard upon tap. Resolves #1816

Steps to reproduce:

1. Place the selection at an arbitrary position
2. Close the software keyboard using the system button
3. Tap again on the same selected position

After that, the software keyboard isn't shown.

The reason is that we show the software keyboard when selection changes. Because the selection didn't change, the software keyboard isn't show.

This PR forwards `SoftwareKeyboardController` to the Android touch interactor so it can show the keyboard upon tap.

I couldn't reproduce this on iOS because I didn't find any system button to close the keyboard. We have an above-the-keyboard panel button to close the keyboard, but in reality we are closing the IME connection when tapping on it. I noticed that Flutter doesn't expose a method to close the software keyboard, so it's probably a good idea to file an issue for that.

Open questions:

- Should we remove the code that opens the keyboard when selection changes? 
- Should we replicate the same implementation to iOS?

### DIFF
--- a/super_editor/test/super_editor/supereditor_input_ime_test.dart
+++ b/super_editor/test/super_editor/supereditor_input_ime_test.dart
@@ -1315,7 +1315,7 @@ Paragraph two
       });
     });
 
-    testWidgetsOnAndroid('shows software keyboard when tapping at the selected position', (tester) async {
+    testWidgetsOnMobile('shows software keyboard when tapping at the selected position', (tester) async {
       await tester
           .createDocument() //
           .withSingleParagraph()
--- a/super_editor/test/super_editor/supereditor_input_ime_test.dart
+++ b/super_editor/test/super_editor/supereditor_input_ime_test.dart
@@ -1315,6 +1315,41 @@ Paragraph two
       });
     });
 
+    testWidgetsOnAndroid('shows software keyboard when tapping at the selected position', (tester) async {
+      await tester
+          .createDocument() //
+          .withSingleParagraph()
+          .withInputSource(TextInputSource.ime)
+          .pump();
+
+      // Place the caret at "Lorem| ipsum".
+      await tester.placeCaretInParagraph('1', 5);
+
+      // Hide the software keyboard using the system button.
+      tester.testTextInput.hide();
+
+      bool wasKeyboardShown = false;
+
+      // Intercept the messages sent to the platform to check if
+      // we showed the software keyboard.
+      tester
+          .interceptChannel(SystemChannels.textInput.name) //
+          .interceptMethod(
+        'TextInput.show',
+        (methodCall) {
+          wasKeyboardShown = true;
+
+          return null;
+        },
+      );
+
+      // Tap again on the same selected position.
+      await tester.placeCaretInParagraph('1', 5);
+
+      // Ensure the keyboard was shown.
+      expect(wasKeyboardShown, isTrue);
+    });
+
     testWidgetsOnAllPlatforms('clears composing region after losing focus', (tester) async {
       final focusNode = FocusNode();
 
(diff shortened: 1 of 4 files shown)


matthew-carroll:
My initial thought is that if the Android gesture interactor only needs to be able to open the keyboard, then we probably shouldn't pass in a whole controller. In general, any API that you expose to an object will eventually be used by that object, even if not strictly necessary. So we should be careful when passing things around like that.

Is there any reason that we need to pass a whole controller instead of just a callback method?

> Should we remove the code that opens the keyboard when selection changes?

I didn't follow the rationale there. Why do we want to remove it?

> Should we replicate the same implementation to iOS?

Which implementation? Is there a behavior that's expected on iOS that we currently aren't honoring on iOS?

angelosilvestre:
@matthew-carroll 

> Is there any reason that we need to pass a whole controller instead of just a callback method?

No particular reason. We can pass just a callback if needed.

> > Should we remove the code that opens the keyboard when selection changes?
> 
> I didn't follow the rationale there. Why do we want to remove it?

If we will always show the keyboard upon tap, showing the keyboard also when selection changes might be unnecessary.

> > Should we replicate the same implementation to iOS?
> 
> Which implementation? Is there a behavior that's expected on iOS that we currently aren't honoring on iOS?

I mean to show the keyboard upon tap

matthew-carroll:
> No particular reason. We can pass just a callback if needed.

Ok, let's do that.

> If we will always show the keyboard upon tap, showing the keyboard also when selection changes might be unnecessary.

Do we have existing policies related to focus change, or programmatic placing of the caret, which would be impacted by that?

> I mean to show the keyboard upon tap

Yeah, I suppose we should duplicate that change.

angelosilvestre:
@matthew-carroll Updated the implementation.

> If we will always show the keyboard upon tap, showing the keyboard also when selection changes might be unnecessary.

After checking again I think we shouldn't remove that, since it's configurable via policies.

super_editor/test/super_editor/supereditor_input_ime_test.dart
```diff
@@ -1315,6 +1315,41 @@ Paragraph two
       });
     });
 
+    testWidgetsOnMobile('shows software keyboard when tapping at the selected position', (tester) async {
```
matthew-carroll, line 1318: Please change this to "opens software keyboard when tapping on caret"

Also, should there be a test for an expanded selection?

angelosilvestre: Updated.

matthew-carroll (approved):
LGTM

PR #1982, opened 2024-05-05
[SuperEditor] Center caret based on width instead of expanding to the right

KevinBrendel:
Same as #1945, but for SuperEditor itself.


matthew-carroll (approved):
LGTM

matthew-carroll:

PR #2012, opened 2024-05-17
[SuperTextField][iOS] Ignore auto-corrections after handling input actions (Resolves #2004)

### LINKED ISSUE #2004: [SuperTextField] [iOS] Overriding imeController.performAction() does not behave as expected
**Package Version**
Using `main` on e20eb1b917bf552c1991d79c21eafd0afb89a8cd.

**To Reproduce**
1. Run the iOS SuperTextField with the code below
2. Write "Run tom"
3. Press the return button

**Minimal Reproduction Code**
<details>

```dart
import 'package:logging/logging.dart';

import '../_mobile_textfield_demo.dart';

/// Demo of [SuperIOSTextField].
class SuperIOSTextFieldDemo extends StatefulWidget {
  @override
  State<SuperIOSTextFieldDemo> createState() => _SuperIOSTextFieldDemoState();
}

class _SuperIOSTextFieldDemoState extends State<SuperIOSTextFieldDemo> {
  final String _tapRegionGroupId = "iOS";
  final _controller = _TextController();

  late final FocusNode _focusNode;

  @override
  void initState() {
    super.initState();
    initLoggers(Level.FINE, {
      // iosTextFieldLog,
      // imeTextFieldLog,
    });

    _focusNode = FocusNode();
  }

  @override
  void dispose() {
    _focusNode.dispose();

    deactivateLoggers({iosTextFieldLog, imeTextFieldLog});
    super.dispose();
  }

  @override
    return MobileSuperTextFieldDemo(
      initialText: AttributedText(
        'This is a custom textfield implementation called SuperIOSTextfield. It is super long so that we can mess with scrolling. This drags it out even further so that we can get multiline scrolling, too. Lorem ipsum dolor sit amet, consectetur adipiscing elit. Proin tempor sapien est, in eleifend purus rhoncus fringilla. Class aptent taciti sociosqu ad litora torquent per conubia nostra, per inceptos himenaeos. Nulla varius libero lorem, eget tincidunt ante porta accumsan. Morbi quis ante at nunc molestie ullamcorper.',
      ),
      textFieldFocusNode: _focusNode,
      textFieldTapRegionGroupId: _tapRegionGroupId,
      createTextField: _buildTextField,
    );
  }

  Widget _buildTextField(MobileTextFieldDemoConfig config) {
    final genericTextStyle = config.styleBuilder({});
    final lineHeight =
        genericTextStyle.fontSize! * (genericTextStyle.height ?? 1.0);

    return DecoratedBox(
      decoration: BoxDecoration(
        border: Border.all(color: Colors.black),
      ),
      child: SuperIOSTextField(
        textController: _controller,
        focusNode: _focusNode,
        tapRegionGroupId: _tapRegionGroupId,
        textStyleBuilder: config.styleBuilder,
        padding: const EdgeInsets.symmetric(horizontal: 48, vertical: 24),
        hintBuilder: StyledHintBuilder(
          hintText: AttributedText("Enter text"),
          hintTextStyleBuilder: (attributions) {
            return config
                .styleBuilder(attributions)
                .copyWith(color: Colors.grey);
          },
        ).build,
        selectionColor: Colors.blue.withOpacity(0.4),
        caretStyle: const CaretStyle(color: Colors.blue),
        blinkTimingMode: BlinkTimingMode.timer,
        handlesColor: Colors.blue,
        minLines: config.minLines,
        maxLines: config.maxLines,
        lineHeight: lineHeight,
        textInputAction: TextInputAction.done,
        showDebugPaint: config.showDebugPaint,
      ),
    );
  }
}

class _TextController extends ImeAttributedTextEditingController {
  @override
    // super.performAction(action);
  }
}
```
</details>

**Actual behavior**
With the text mentioned in the reproduction steps ("Run tom"), a suggestion ("Tom") is applied.

**Expected behavior**
Nothing should happen since `_TextController.performAction()` does not call `super.performAction(action)`.

**Platform**
Only on iOS

**Flutter version**
Master channel with commit `d5bcd587aa72570faf12875bdfb326ab51eba961`.

**Screenshots**

https://github.com/superlistapp/super_editor/assets/31278849/e83e63b7-a983-4940-97ec-3aa543f89847

**Additional context**
With `SuperEditor` we can extend `DeltaTextInputClientDecorator` to override the behaviour of `performAction()` and there we don't have the problem where the IME suggestion is applied when `super.performAction()` is not called, as expected.

angelosilvestre:
[SuperTextField][iOS] Ignore auto-corrections after handling input actions. Resolves #2004

On iOS, auto-corrections can happen after pressing the action button. For example, the user might edit some content, press "done", and after the "done" action is handled the IME reports the text replacement.

This PR prevents that by ignoring text replacements in the same frame that input actions are reported, except for `TextInputAction.newLine`.  `TextInputAction.newLine` can't prevent replacements from being applied, otherwise the user won't be able to replace the selected content with a new line.

### DIFF
--- a/super_editor/lib/src/super_textfield/input_method_engine/_ime_text_editing_controller.dart
+++ b/super_editor/lib/src/super_textfield/input_method_engine/_ime_text_editing_controller.dart
@@ -393,6 +393,7 @@ class ImeAttributedTextEditingController extends AttributedTextEditingController
   AutofillScope? get currentAutofillScope => null;
 
   @override
+  @mustCallSuper
   void performAction(TextInputAction action) {
     _onPerformActionPressed?.call(action);
 
--- a/super_editor/lib/src/super_textfield/input_method_engine/_ime_text_editing_controller.dart
+++ b/super_editor/lib/src/super_textfield/input_method_engine/_ime_text_editing_controller.dart
@@ -394,7 +394,7 @@ class ImeAttributedTextEditingController extends AttributedTextEditingController
 
   @override
   void performAction(TextInputAction action) {
-    handleTextInputAction(action);
+    _onPerformActionPressed?.call(action);
 
     // Keep track that we have performed a text input action on this frame so we can ignore auto-corrections
     // reported after we handled the text input action.
@@ -700,10 +700,6 @@ class ImeAttributedTextEditingController extends AttributedTextEditingController
   @override
   void selectAll() {
     _realController.selectAll();
-  }
-
-  void handleTextInputAction(TextInputAction action) {
-    _onPerformActionPressed?.call(action);
   }
 }
 
--- a/super_editor/lib/src/super_textfield/input_method_engine/_ime_text_editing_controller.dart
+++ b/super_editor/lib/src/super_textfield/input_method_engine/_ime_text_editing_controller.dart
@@ -112,6 +112,9 @@ class ImeAttributedTextEditingController extends AttributedTextEditingController
   /// Used to determine whether or not we need to send our editing value to the IME.
   TextEditingValue _osCurrentTextEditingValue = const TextEditingValue();
 
+  /// Whether or not a `TextInputAction` differente from `TextInputAction.newLine` was performed on the current frame.
+  bool _hasPerformedNonNewLineTextInputActionThisFrame = false;
+
   void attachToIme({
     bool autocorrect = true,
     bool enableSuggestions = true,
@@ -306,6 +309,16 @@ class ImeAttributedTextEditingController extends AttributedTextEditingController
       return;
     }
 
+    if (_hasPerformedNonNewLineTextInputActionThisFrame &&
+        defaultTargetPlatform == TargetPlatform.iOS &&
+        deltas.every((e) => e is TextEditingDeltaReplacement)) {
+      // On iOS, pressing the action button can trigger the IME to try to apply auto-corrections
+      // after we have already processed the input action. Ignore replacement deltas on the same frame
+      // and forcefully update the IME with our current state.
+      _sendEditingValueToPlatform();
+      return;
+    }
+
     // Update our view from the OS editing value.
     for (final delta in deltas) {
       _osCurrentTextEditingValue = delta.apply(_osCurrentTextEditingValue);
@@ -382,6 +395,19 @@ class ImeAttributedTextEditingController extends AttributedTextEditingController
   @override
   void performAction(TextInputAction action) {
     _onPerformActionPressed?.call(action);
+
+    // Keep track that we have performed a text input action on this frame so we can ignore auto-corrections
+    // reported after we handled the text input action.
+    //
+    // We don't ignore TextInputAction.newline because the insertion of the new line happens after the action
+    // is reported, and we need to handle the new line insertion to let users replace the selected content
+    // with a new line.
+    //
+    // See https://github.com/superlistapp/super_editor/issues/2004 for more information.
+    _hasPerformedNonNewLineTextInputActionThisFrame = action != TextInputAction.newline;
+    WidgetsBinding.instance.addPostFrameCallback((timeStamp) {
+      _hasPerformedNonNewLineTextInputActionThisFrame = false;
+    });
   }
 
   @override
--- a/super_editor/lib/src/super_textfield/input_method_engine/_ime_text_editing_controller.dart
+++ b/super_editor/lib/src/super_textfield/input_method_engine/_ime_text_editing_controller.dart
@@ -394,7 +394,7 @@ class ImeAttributedTextEditingController extends AttributedTextEditingController
 
   @override
   void performAction(TextInputAction action) {
-    _onPerformActionPressed?.call(action);
+    handleTextInputAction(action);
 
     // Keep track that we have performed a text input action on this frame so we can ignore auto-corrections
     // reported after we handled the text input action.
@@ -700,6 +700,10 @@ class ImeAttributedTextEditingController extends AttributedTextEditingController
   @override
   void selectAll() {
     _realController.selectAll();
+  }
+
+  void handleTextInputAction(TextInputAction action) {
+    _onPerformActionPressed?.call(action);
   }
 }
 
--- a/super_editor/test/super_textfield/super_textfield_ime_test.dart
+++ b/super_editor/test/super_textfield/super_textfield_ime_test.dart
@@ -627,6 +627,36 @@ void main() {
       expect(enableDeltaModel, true);
       expect(keyboardAppearance, 'Brightness.dark');
     });
+
+    group('on iPhone 15 (iOS 17.5)', () {
+      testWidgetsOnIos('ignores keyboard autocorrections when pressing the action button', (tester) async {
+        await _pumpEmptySuperTextField(tester);
+
+        // Place the caret at the start of the text field.
+        await tester.placeCaretInSuperTextField(0);
+
+        // Type some text.
+        await tester.typeImeText('run tom');
+
+        // Press the "Done" button.
+        await tester.testTextInput.receiveAction(TextInputAction.done);
+
+        // Simulate the IME sending a delta replacing "tom" with "Tom".
+        await tester.ime.sendDeltas([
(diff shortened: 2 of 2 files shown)


angelosilvestre:
@miguelcmedeiros Can you please try this PR and see if it work for you? It will be required that `super.performAction(action)` is called.

miguelcmedeiros:
> It will be required that super.performAction(action) is called.

@angelosilvestre if the user presses "done" and then do not call `super.performAction()` then I would expect that the replacement is not applied and the user can keep typing. 

Is that possible?

angelosilvestre:
@miguelcmedeiros If you do not intend to call `super.performAction()` then we would need to have some method like `preventReplacementsUntilNextFrame()` for subclasses to call and signal us that we should do that.

Would that work for you?

miguelcmedeiros:
@angelosilvestre I guess that would work. 

The goal is to be able to handle the "done" key press and only apply a custom behaviour from our side.

matthew-carroll:
> The goal is to be able to handle the "done" key press and only apply a custom behaviour from our side.

@miguelcmedeiros can you please elaborate on what that means in practice? What custom behavior would you want to add, and what default behavior are you trying to avoid?

We can't make any guarantees about completely stopping all default behavior, because it seems that some default behavior isn't in our control.

Also, do you feel that you understand the nature of the problem that we've discovered, and what we're doing to solve it?

I want to make sure we're on the same page because I'm not a fan of the option to add a very obscure public method that no implementer will understand.

miguelcmedeiros:
@matthew-carroll similar to keyboard handlers, it would be great that as a developer I can intercept the pressing of the return button in the soft keyboard so that I can define what happens, e.g. perform a custom action defined on the app side, and then block or allow it to execute the default action.

matthew-carroll:
I understand the abstract desire. Can you please provide some examples of what's needed, concretely? As I mentioned, we do not have the level of control to provide the abstract goal. Instead, we'll need to look at general areas of control that you definitely want, and then see what we need to do to make those possible.

miguelcmedeiros:
A concrete can be what is shown in the screen recording in the ticket description. The user typed some text and then presses return. In our case, we are presenting the user with a due date suggestion, which the user can accept by tapping on the suggestion overlay or press return. 

Currently, if IME suggests a change to the text, then it will be applied, which is not desired for our use case.

matthew-carroll:
So does Angelo's change fix that problem (excluding the idea of a new public method)? If so, what new problems does Angelo's approach create for you? Or what other needs are not being met by that change?

matthew-carroll:
@angelosilvestre let's automatically apply this delta-ignoring behavior outside of the `performAction()` call. No configuration, or opting in or opting out, we'll just always ignore deltas that come in the frame after the action.

Then, the purpose of the `performAction()` override will go back to what it was before. If you want the default behavior, don't override it. If you want totally custom responses to actions, then override it.

Make sense?

angelosilvestre:
@matthew-carroll This is the current implementation of this PR.

miguelcmedeiros:
@angelosilvestre I tested out the changes with the example code and it still performs the text replacement when pressing return.

angelosilvestre:
@miguelcmedeiros Did you call super.performAction?

miguelcmedeiros:
Nope. My understanding was that if super.performAction is not called, then nothing else happens to the text field.

angelosilvestre:
@miguelcmedeiros The original issue isn't caused by the default performAction implementation. The issue is that the iOS IME is reporting a replacement due to autocorrection after reporting the input action.

The fix in this PR consists in discarding replacement deltas reported in the same frame that an input action is performed. For that, the default performAction implementation needs to be called, so we can keep track that an action was performed.

matthew-carroll:
> For that, the default performAction implementation needs to be called, so we can keep track that an action was performed.

@angelosilvestre my point in my earlier comment is that we need this statement to not be true. We do not want to require any particular use, or non-use, of `performAction` to ignore the auto-correction.

angelosilvestre:
@matthew-carroll But if the IME controller is extended to avoid calling the default performAction implementation we cannot know that an action was performed.

matthew-carroll:
Why can't we install our own private callback with the IME, implement ignoring deltas after actions, and then forward from our private callback to the regular callback in the controller?

angelosilvestre:
@matthew-carroll I think that, for SuperTextField, the IME controller IS the delta input client, isn't it? The performAction is called directly on the IME controller.

matthew-carroll:
So what are our options to control that callback?

angelosilvestre:
@matthew-carroll Maybe introduce a `handleAction` and then tell people to override that instead of `performAction`? 

Then, our performAction implementation will be just setting the flag to ignore the deltas and calling the new handleAction method.

matthew-carroll:
Ok, let's see how that looks in code.

angelosilvestre:
@matthew-carroll Updated. With this change, the controller subclass will look like the following:

```dart
class _TextController extends ImeAttributedTextEditingController {
  @override
  void handleTextInputAction(TextInputAction action) {
    // Do not perform actions, while still letting the base controller to keep track of internal flags.
  }
}
```

super_editor/lib/src/super_textfield/input_method_engine/_ime_text_editing_controller.dart
```diff
@@ -381,7 +394,20 @@ class ImeAttributedTextEditingController extends AttributedTextEditingController
 
   @override
   void performAction(TextInputAction action) {
-    _onPerformActionPressed?.call(action);
+    handleTextInputAction(action);
```
matthew-carroll, line 397: This method call is directly forwarding to another method call. Why do we need the `handleTextInputAction()`?

I know what you mentioned in the PR description, but it seems to me like the situation here isn't exactly what was described.

As you can see, we already have a `_onPerformActionPressed` callback. That callback is publicly settable. It's called every time an action is pressed.

Why was there ever a need for anyone to try to override `performAction()`? If a client wants to customize the action behavior, why wouldn't that client just set their own value on `onPerformActionPressed`?

miguelcmedeiros: @angelosilvestre I tried again (by setting `onPerformActionPressed`) and it works as expected.

I'm guessing that yesterday when I tested, I picked the wrong commit and so, was hitting the same issue continuously since your fix was not included. Sorry about the confusion.

angelosilvestre: @matthew-carroll So I guess we can just revert the last commit, right?

matthew-carroll: @angelosilvestre I don't know. I can't tell exactly what the situation is. If you want to revert the commit then I can take a look at what that looks like. I still think we want the deltas to automatically be ignored in the next frame, no matter what the app chooses to override. If that was already accomplished in the previous commit, then maybe that commit works.

angelosilvestre: @matthew-carroll Yeah, I believe the previous commit will work.

super_editor/lib/src/super_textfield/input_method_engine/_ime_text_editing_controller.dart
```diff
@@ -382,6 +395,19 @@ class ImeAttributedTextEditingController extends AttributedTextEditingController
   @override
   void performAction(TextInputAction action) {
     _onPerformActionPressed?.call(action);
+
+    // Keep track that we have performed a text input action on this frame so we can ignore auto-corrections
```
matthew-carroll, line 399: Looks like we're back to the situation where if this method is overridden by a subclass without calling super(), we're not going to get the expected behavior to ignore the deltas.

Either this method should be annotated with "must call super", or this behavior needs to be moved to a place where its guaranteed to run.

angelosilvestre: @matthew-carroll Added `@mustCallSuper`.

matthew-carroll (approved):
LGTM

PR #2014, opened 2024-05-18
[SuperEditor][Android] Show expanded handles when expanding the selection (Resolves #1937)

### LINKED ISSUE #1937: Under certain conditions, text selection handlers appear in the wrong place on Android
**Package Version**
`main`

**To Reproduce**
The bug can be reproduced in the Example app on Android.

1. Run the example application on Android device
2. Click anywhere and see the collapsed text selection handler
3. Click on handler
4. See toolbar appear
5. Choose `Select All`
6. See text selection handler appear at `Offset.zero` instead of the start and ends of the document

This behavior can only be observed if the text selection is collapsed if it isn't, the behavior is as expected.

**Actual behavior**
The text selection handler appears at `Offset.zero` when "Select All" is initiated with a collapsed selection.

**Expected behavior**
The text selection handlers should be displayed where the text selection actually starts and ends in the document.

**Platform**
Android

angelosilvestre:
[SuperEditor][Android] Show expanded handles when expanding the selection. Resolves #1937

When the selection changes from collapsed to expanded, the collapsed handle is visible at the beginning of the selection, and the expanded handles aren't visible:

[collapsed.webm](https://github.com/superlistapp/super_editor/assets/7597082/c2eb5476-5887-4c2c-a33d-a30e693aa849)

This PR fixes the issue by changing the visibility of the handles when needed.

### DIFF
--- a/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
@@ -1345,6 +1345,7 @@ class SuperEditorAndroidControlsOverlayManagerState extends State<SuperEditorAnd
     _controlsController = SuperEditorAndroidControlsScope.rootOf(context);
     // TODO: Replace Cupertino aligner with a generic aligner because this code runs on Android.
     _toolbarAligner = CupertinoPopoverToolbarAligner();
+    widget.selection.addListener(_onSelectionChange);
   }
 
   @override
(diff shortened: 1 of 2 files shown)


matthew-carroll:
@angelosilvestre please post a video of the solution

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
```diff
@@ -1363,6 +1370,21 @@ class SuperEditorAndroidControlsOverlayManagerState extends State<SuperEditorAnd
   @visibleForTesting
   bool get wantsToDisplayMagnifier => _controlsController!.shouldShowMagnifier.value;
 
+  void _onSelectionChange() {
+    if (widget.selection.value != null &&
+        widget.selection.value?.isCollapsed == false &&
+        _controlsController!.shouldShowCollapsedHandle.value == true) {
+      // The selection is expanded, but the collapsed handle is visible. This can happen when the
```
matthew-carroll, line 1377: This reads like a bandaid - it says "here's a state that shouldn't happen, but does happen, so we'll correct it here".

Why is the wrong thing happening to begin with? Why is this the right place to correct it?

angelosilvestre:
[expanded.webm](https://github.com/superlistapp/super_editor/assets/7597082/8c7b0346-6c43-409a-888f-9485d6820f2e)

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
angelosilvestre: Well, usually we show/hide the handles in the touch handlers of the interactor. A double tap shows the expanded handle, a single tap shows the collapsed handle, etc...

In this case the selection is changed elsewhere. So I think we should be changing the handle visibility when the selection changes.

matthew-carroll: > In this case the selection is changed elsewhere

Where is it being changed?

matthew-carroll:
@angelosilvestre this PR now has some conflicts after another PR was merged.

Also, can you take a look at the failing website deployment actions? They seem to be failing in a number of PRs but I'm not sure why.

angelosilvestre:
@matthew-carroll I resolved the conflicts.

The website build failure seems to be due to the `LinkAttribution` constructor change, where now we expect a String. I already solved this in my website PR, so for new PR's this should fail again.

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
matthew-carroll: @angelosilvestre did you see this question?

angelosilvestre: @matthew-carroll I haven't noticed this, sorry.

The "Select All" button issues a `ChangeSelectionRequest`, this isn't initiated by one of our widgets.

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
```diff
@@ -1383,7 +1384,12 @@ class SuperEditorAndroidControlsOverlayManagerState extends State<SuperEditorAnd
   bool get wantsToDisplayMagnifier => _controlsController!.shouldShowMagnifier.value;
 
   void _onSelectionChange() {
-    if (widget.selection.value?.isCollapsed == true &&
+    final selection = widget.selection.value;
+    if (selection == null) {
+      return;
+    }
+
+    if (selection.isCollapsed == true &&
```
matthew-carroll, line 1392: Is there a reason that you're doing `== true` and `== false` for a boolean value?

angelosilvestre: Previously we were handling a nullable value and I forgot to update it. I updated it now.

PR #2017, opened 2024-05-19
[SuperEditor] Add list item customization via stylesheet (Resolves #1912)

### LINKED ISSUE #1912: [SuperEditor] Expose dot size and color for unordered lists
It would be great to have the ability to customize the size and color of the dots used for unordered lists. This would allow users to have more control over the appearance of their lists and match them with their app's design without creating a custom component just for this.

Requesting this as this styling is pretty common for editors like other options available in the stylesheet.

### Proposed changes
* Add a way to set the dot size (e.g., a property or method)
* Add a way to change the dot color (e.g., a property or method)

### Benefits
* More customization options for users
* Improved visual consistency with the app's theme

angelosilvestre:
[SuperEditor] Add list item customization via stylesheet. Resolves #1912

This PR adds the ability to customize list items using stylesheets. 

Customization for unordered list items:
- Dot size
- Dot shape
- Dot color

Customization for ordered list items:
- Numeral style, which can be:
  - Arabic (the default)
  - Lowercase roman numbers
  - Uppercase roman numbers
  - Lowercase alpha
  - Uppercase alpha

For roman numbers, we support until 3999 and them switch to arabic. I tried the w3 playground in https://www.w3schools.com/css/tryit.asp?filename=trycss_list-style-type_ex and it seems that html lists also work like that. 

CSS list style references: https://www.w3schools.com/css/css_list.asp


super_editor/lib/src/default_editor/list_items.dart
```diff
@@ -154,6 +155,9 @@ class ListItemComponentBuilder implements ComponentBuilder {
         text: componentViewModel.text,
         styleBuilder: componentViewModel.textStyleBuilder,
         indent: componentViewModel.indent,
+        dotColor: componentViewModel.dotColor,
```
matthew-carroll, line 158: Would it be more convenient to bundle these dot properties into a data structure?

super_editor/lib/src/default_editor/list_items.dart
```diff
@@ -189,6 +194,10 @@ class ListItemComponentViewModel extends SingleColumnLayoutComponentViewModel wi
     required this.type,
     this.ordinalValue,
     required this.indent,
+    this.dotColor,
```
matthew-carroll, line 197: I'm noticing that these properties are the first situation where this view model doesn't apply equally to both types of list items. Looks like we should split this view model into two - one for ordered list items and one for unordered list items.

It might make sense to create a shared superclass that has everything but those dot and numeral styles.

angelosilvestre: Updated.

super_editor/lib/src/default_editor/list_items.dart
```diff
@@ -130,9 +131,18 @@ class ListItemComponentBuilder implements ComponentBuilder {
       }
     }
 
-    return ListItemComponentViewModel(
+    if (node.type == ListItemType.unordered) {
```
matthew-carroll, line 134: It probably makes more sense to do a `switch` on `node.type` to ensure compiler level exhaustiveness.

super_editor/lib/src/default_editor/list_items.dart
```diff
@@ -130,9 +131,18 @@ class ListItemComponentBuilder implements ComponentBuilder {
...
+      return UnorderedListItemComponentViewModel(
+        nodeId: node.id,
+        indent: node.indent,
```
matthew-carroll, line 137: Should we be initializing these with default dot and numeral choices?

super_editor/lib/src/default_editor/list_items.dart
```diff
@@ -385,6 +534,10 @@ Widget _defaultUnorderedListItemDotBuilder(BuildContext context, UnorderedListIt
   final attributions = component.text.getAllAttributionsAt(0).toSet();
   final textStyle = component.styleBuilder(attributions);
 
+  final dotColor = component.dotStyle?.color ?? textStyle.color;
```
matthew-carroll, line 537: Is there any reason that these values aren't set inline within the widget tree? Do we need variables for them?

super_editor/lib/src/default_editor/list_items.dart
```diff
@@ -533,6 +688,8 @@ Widget _defaultOrderedListItemNumeralBuilder(BuildContext context, OrderedListIt
   final attributions = component.text.getAllAttributionsAt(0).toSet();
   final textStyle = component.styleBuilder(attributions);
 
+  final numeral = _numeralForIndex(component.listIndex, component.numeralStyle);
```
matthew-carroll, line 691: Do we need a variable for this?

super_editor/lib/src/default_editor/list_items.dart
angelosilvestre: The constructor already have default values for that, so I think we could omit them.

super_editor/lib/src/default_editor/list_items.dart
angelosilvestre: For the dot size I think it makes sense because we need access both width and height. I removed the others.

matthew-carroll (approved):
LGTM

PR #2018, opened 2024-05-19
Document usage of density independent pixels

angelosilvestre:
Document usage of density independent pixels

After the work done in https://github.com/superlistapp/super_editor/pull/1973, this PR documents that the `Offset`s used in the magnifier are expected to be in density independent pixels.

It seems that the only place where real pixel values are used is in follow_the_leader.


matthew-carroll (approved):
LGTM

PR #2019, opened 2024-05-19
[SuperTextField][iOS] Don't show the mobile toolbar upon first tap (Resolves #2005)

### LINKED ISSUE #2005: [SuperTextField][Mobile] - Popover toolbar appears when it shouldn't
**Package Version**
main

**To Reproduce**
1. Tap on an empty SuperTextField on mobile

**Minimal Reproduction Code**
Run the SuperTextField demo

**Actual behavior**
A popover toolbar appears.

**Expected behavior**
No popover toolbar should appear because there isn't any content to interact with.

If iOS and Android show a similar behavior to what we have, then feel free to correct me, but this seems wrong. 

**Platform**
Repro'd on iOS, but probably on Android, too.

**Flutter version**
v.3.22

**Screenshots**

https://github.com/superlistapp/super_editor/assets/7259036/fa51726c-6c3d-4a89-a653-7e52e0eb3f21

angelosilvestre:
[SuperTextField][iOS] Don't show the mobile toolbar upon first tap. Resolves #2005 

On iOS, tapping to focus the textfield is showing the toolbar. Looking at some iOS app, it seems the toolbar should only be displayed on tap if the textfield already has focus.

This PR changes the iOS textfield to only show the toolbar on tap if it already has focus.

For Android, it seems that doesn't show the toolbar even if the textfield already has focus.

### DIFF
--- a/super_editor/test/super_textfield/super_textfield_inspector.dart
+++ b/super_editor/test/super_textfield/super_textfield_inspector.dart
@@ -1,7 +1,6 @@
 import 'package:flutter/cupertino.dart';
 import 'package:flutter/widgets.dart';
 import 'package:flutter_test/flutter_test.dart';
-import 'package:super_editor/src/super_textfield/android/_editing_controls.dart';
 import 'package:super_editor/super_editor.dart';
 import 'package:super_text_layout/super_text_layout.dart';
 
--- a/super_editor/test/super_textfield/super_textfield_inspector.dart
+++ b/super_editor/test/super_textfield/super_textfield_inspector.dart
@@ -1,6 +1,7 @@
 import 'package:flutter/cupertino.dart';
 import 'package:flutter/widgets.dart';
 import 'package:flutter_test/flutter_test.dart';
+import 'package:super_editor/src/super_textfield/android/_editing_controls.dart';
 import 'package:super_editor/super_editor.dart';
 import 'package:super_text_layout/super_text_layout.dart';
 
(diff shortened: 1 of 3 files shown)


super_editor/test/super_textfield/super_textfield_inspector.dart
```diff
@@ -293,6 +294,45 @@ class SuperTextFieldInspector {
         "Couldn't find the caret rectangle because we couldn't find a SuperTextField. Finder: $superTextFieldFinder");
   }
 
+  /// Returns `true` if the mobile text field is configured to display a toolbar.
```
matthew-carroll, line 297: Both the name of this method, and the documentation confuses me. I don't know what "is configured to display a toolbar" means, nor do I know what "wantsMobileToolbarToBeVisible" means. Can you please revisit those to help the reader understand what they're querying here?

angelosilvestre: They mean that the overlay controller has `shouldDisplayToolbar` set to `true`. I didn't documented as if this guarantees that the toolbar is actually displayed because we are not checking for any widgets in the tree...

matthew-carroll: Then should we check for widgets, instead? Is there a reason that you chose to lookup a property variable rather than verify the actual toolbar?

angelosilvestre: In `SuperEditor`, the toolbar builder provides a `Key` for the user to attach to the toolbar. With that key we can find the toolbar widget.

In `SuperTextField`, there is no provided key that apps can attach to the toolbar. Since the toolbar is fully in control of the app we don't have a easy way to query the toolbar widget.

A possible solution would be to wrap the toolbar with another widget that we could easily find...

matthew-carroll: > Since the toolbar is fully in control of the app we don't have a easy way to query the toolbar widget.

Can you clarify what that means?

We're talking about tests, right? Our tests decide what to build for the toolbar, and our tests should be able to find whatever toolbar they create, right?

angelosilvestre: I changed to override the toolbar builders, so we can check if the toolbar widget is present. Unfortunately, we cannot configure this at `SuperTextField` level, so I had to create `SuperAndroidTextField` and `SuperIosTextField` directly.

super_editor/test/super_textfield/super_textfield_inspector.dart
```diff
@@ -1,6 +1,7 @@
 import 'package:flutter/cupertino.dart';
 import 'package:flutter/widgets.dart';
 import 'package:flutter_test/flutter_test.dart';
+import 'package:super_editor/src/super_textfield/android/_editing_controls.dart';
```
matthew-carroll, line 4: Should this be here?

matthew-carroll (approved):
LGTM

PR #2020, opened 2024-05-19
[SuperEditor][Mobile] Add caret customization (Resolves #1935)

### LINKED ISSUE #1935: Provide a way to customize caret width for platforms other than desktop
**Feature description**
Currently, caret width is not customizable for Android or iOS. Neither:
- `SuperEditorIosHandlesDocumentLayerBuilder` or
- `SuperEditorAndroidHandlesDocumentLayerBuilder`

expose these configurations.

However, `DefaultCaretOverlayBuilder` does: https://github.com/superlistapp/super_editor/blob/661ecb84b4f79bd3f81c1bfd148b36ab7be6993b/super_editor/lib/src/default_editor/super_editor.dart#L1093-L1103

Would it be possible to offer the same customization options to iOS and Android counterparts?

**User Behavior**
The request doesn't ask for any additional features the user could perform. Simply, additional styling capabilities for the developer to use.
(The developer could extend these capabilities to then offer users a way to personalize the appearance of the caret if desired. Which is currently only possible if the developer chooses to rewrite the existing layers themselves.)

**Feature value**
Ideally, the client of the package would be able to customize small details like caret width, since these details can end up, making a big difference in feeling of the application.

angelosilvestre:
[SuperEditor][Mobile] Add caret customization. Resolves #1935

This PR adds the following customizations:
- Android: Caret width
- iOS:
  - Caret width
  - Handle ball diameter


super_editor/test/super_editor/supereditor_test_tools.dart
```diff
@@ -177,6 +177,13 @@ class TestSuperEditorConfigurator {
     return this;
   }
 
+  /// Configures the [SuperEditor] to ignore the default layers and to use the given only
```
matthew-carroll, line 180: Looks like "the given" is repeated twice.

Also, this dart doc doesn't help me understand why this method exists, or when/why I might use it.

angelosilvestre: Updated.

super_editor/test/super_editor/supereditor_test_tools.dart
```diff
@@ -177,6 +177,16 @@ class TestSuperEditorConfigurator {
     return this;
   }
 
+  /// Configures the [SuperEditor] to use the given [documentLayers].
+  ///
+  /// This can be used, for example, to customize the caret style for a specific platform.
```
matthew-carroll, line 182: This seems like a lot of responsibility to just change the caret style. Do we have similar examples of minor customizations that require this level of intervention?

I know we have keyboard handlers, but that situation is mostly about adding a few new handlers on top of existing ones. 

In this case, the user is altering an existing UI property in a small way....

angelosilvestre: Updated.

super_editor/test/super_editor/supereditor_test_tools.dart
```diff
@@ -578,13 +603,20 @@ class _TestSuperEditorState extends State<_TestSuperEditor> {
       // iOS floating toolbar.
       const SuperEditorIosToolbarFocalPointDocumentLayerBuilder(),
       // Displays caret and drag handles, specifically for iOS.
-      const SuperEditorIosHandlesDocumentLayerBuilder(),
+      SuperEditorIosHandlesDocumentLayerBuilder(
```
matthew-carroll, line 606: What I was talking about in my layers comment was meant to go behind the specific place where custom layers were being created. I was also talking about how we expect app developers to customize these properties.

Am I reading this correctly that app developers will need to provide a custom list of layers to `SuperEditor` if they want a 3px wide caret instead of a 2px wide caret?

angelosilvestre: > Am I reading this correctly that app developers will need to provide a custom list of layers to SuperEditor if they want a 3px wide caret instead of a 2px wide caret?

Currently yes. We already have caret customization for desktop, which involves creating the whole list of layers...

matthew-carroll (approved):
LGTM

PR #2034, opened 2024-05-24
[SuperEditor] Convert empty task to paragraph on ENTER (Resolves #1898)

### LINKED ISSUE #1898: [SuperEditor] Empty task doesn't get removed on pressing enter/return

angelosilvestre:
[SuperEditor] Convert empty task to paragraph on ENTER. Resolves #1898

Pressing enter on an empty task isn't  converting it to a paragraph. This is the default behavior for list items.

This PR change the new line handling, both with keyboard and deltas to convert empty list items to paragraphs.

### DIFF
--- a/super_editor/lib/src/default_editor/common_editor_operations.dart
+++ b/super_editor/lib/src/default_editor/common_editor_operations.dart
@@ -1792,6 +1792,11 @@ class CommonEditorOperations {
         ]);
       }
     } else if (extentNode is TaskNode) {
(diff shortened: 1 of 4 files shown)


matthew-carroll (approved):
LGTM

PR #2035, opened 2024-05-25
[SuperEditor] Split incompatible attributions (Resolves #1983)

### LINKED ISSUE #1983: [SuperEditor] Add Valued Attribution support in SuperEditor (ColorAttribution, SizeAttribution, etc...)
Hello, I write this issue as Feature, tell me if It's supported or not because I'm not able to find this behavior in Demo.

**User Behavior**
As a User, I use DemoEditor
I would like to change some text color from Selection or just from Cursor.

I use `ToggleTextAttributionsRequest`.
In some case, the Command cannot be applied du to 'canMerge' of ColorAttribute

Ex:
`[color A]All my text is colored into A color[/color A]`
If I decide to color any word between beacons: ERROR

**Feature Value**
When an attribution is associated with value Editor cannot split attribution or manage them to apply the new one if already the same Attribution is applied.

With this feature, editor will be able to use Attribution with value by applying attribution on some TextNode portions an not only paragraph.

Ex 1:
`[color A]All my text is colored into A color[/color A]`
If I decide to color any word between beacons: `[color A]All my text is [/color A][color B]colored[/color B] [color A]into A color[/color A]`

Ex 2:
`Only the word [color A]color[/color A] is colored in A color`
If I decide to color all the sentence with B color: `[color B]Only the word color is colored in A color[/color B]`

The behavior targeted is the same as a classic editor

This behavior can be the same with SizeAttribution for text size, Hightlight, FontFamily etc...

**Additional context**
If you need more, code example or else don't hesitate to ask for.

angelosilvestre:
[SuperEditor] Split incompatible attributions. Resolves #1983

Valued attributions, like `ColorAttribution` can't be applied in a range where another `ColorAttribution` exists.

Although it doesn't make sense to have two `ColorAttribution`s in the same range, users should be able to change the color of an already attributed range.

For example, consider the following range, fully attributed with blue:

```
   blue green blue
b [xxxxxxxxxxxxxxx]
```

Users should be able to apply green color to the word "green", causing the blue attributed range to be split:

```
   blue green blue
b [xxxxx-----xxxxx]
g [-----xxxxx-----]
```

This PR changes `AddTextAttributionsCommand` and `ToggleTextAttributionsCommand` to find and remove any conflicting attributions in the range before trying to apply the desired attribution.

The following changes were made:

- Introduce `AttributedSpans.findConflictingAttributions` to return a list of conflicting attributions.
- Change `AddTextAttributionsCommand` and `ToggleTextAttributionsCommand` to remove conflicting attributions
- Update the tests that previously ensured we couldn't apply conflicting attributions

Also, I found a bug in `AttributedSpans.getMatchingAttributionsWithin`, where we always tried to get the attributions from the same index. This PR fixes this bug.

We'll need to release a new version of `attributed_text` because of this new method.

### DIFF
--- a/attributed_text/lib/src/attributed_spans.dart
+++ b/attributed_text/lib/src/attributed_spans.dart
@@ -280,25 +280,48 @@ class AttributedSpans {
 
   /// Applies the [newAttribution] from [start] to [end], inclusive.
   ///
-  /// When [autoMerge] is `true`, the new attribution is merged with any
-  /// preceding or following attribution whose [Attribution.canMergeWith] returns
-  /// `true`.
+  /// If [start] is less than `0`, nothing happens.
   ///
-  /// It [newAttribution] overlaps a conflicting span, or if [newAttribution]
-  /// overlaps a merge-able span but [autoMerge] is `false`, one of the following
-  /// outcomes happen:
+  /// [AttributedSpans] doesn't have any knowledge about content length, so [end] can
+  /// take any value that's desired. However, users of [AttributedSpans] should take
+  /// care to avoid values for [end] that exceed the content length.
   ///
-  /// - If [splitConflictingAttributions] is `true`, the conflicting span is removed from
-  /// the overlapping region, and the new attribution is added in its place.
+  /// The effect of adding an attribution is straight forward when the text doesn't
+  /// contain any other attributions with the same ID. However, there are various
+  /// situations where [newAttribution] can't necessarily co-exist with other
+  /// attribution spans that already exist in the text.
   ///
-  /// - If [splitConflictingAttributions] is `false`, an [IncompatibleOverlappingAttributionsException]
-  /// is thrown.
+  /// Attribution overlaps can take one of two forms: mergeable or conflicting.
+  ///
+  /// ## Mergeable Attribution Spans
+  /// An example of a mergeable overlap is where two bold spans overlap each
+  /// other. All bold attributions are interchangeable, so when two bold spans
+  /// overlap, those spans can be merged together into a single span.
+  ///
+  /// However, mergeable overlapping spans are not automatically merged. Instead,
+  /// this decision is left to the user of this class. If you want [AttributedSpans] to
+  /// merge overlapping mergeable spans, pass `true` for [ignoreMergeableOverlaps]. Otherwise,
+  /// if [ignoreMergeableOverlaps] is `false`, an exception is thrown when two mergeable spans
+  /// overlap each other.
+  ///
+  ///
+  /// ## Conflicting Attribution Spans
+  /// An example of a conflicting overlap is where a black text color overlaps a red
+  /// text color. Text is either black, OR red, but never both. Therefore, the black
+  /// attribution cannot co-exist with the red attribution. Something must be done
+  /// to resolve this.
+  ///
+  /// There are two possible ways to handle conflicting overlaps. The new attribution
+  /// can overwrite the existing attribution where they overlap. Or, an exception can be
+  /// thrown. To overwrite the existing attribution with the new attribution, pass `true`
+  /// for [overwriteConflictingSpans]. Otherwise, if [overwriteConflictingSpans]
+  /// is `false`, an exception is thrown.
   void addAttribution({
     required Attribution newAttribution,
     required int start,
     required int end,
-    bool autoMerge = true,
-    bool splitConflictingAttributions = true,
+    bool ignoreMergeableOverlaps = true,
+    bool overwriteConflictingSpans = true,
   }) {
     if (start < 0 || start > end) {
       _log.warning("Tried to add an attribution ($newAttribution) at an invalid start/end: $start -> $end");
@@ -308,19 +331,64 @@ class AttributedSpans {
     _log.info("Adding attribution ($newAttribution) from $start to $end");
     _log.finer("Has ${_markers.length} markers before addition");
 
-    final conflicts = findConflictingAttributions(
-      attribution: newAttribution,
-      start: start,
-      end: end,
-      allowMerging: autoMerge,
-    );
+    final conflicts = <AttributionConflict>[];
 
-    if (conflicts.isNotEmpty && !splitConflictingAttributions) {
-      throw IncompatibleOverlappingAttributionsException(
-        existingAttribution: conflicts.first.existingAttribution,
-        newAttribution: newAttribution,
-        conflictStart: conflicts.first.conflictStart,
-      );
+    // Check if conflicting attributions overlap the new attribution.
+    final matchingAttributions = getMatchingAttributionsWithin(attributions: {newAttribution}, start: start, end: end);
+    if (matchingAttributions.isNotEmpty) {
+      for (final matchingAttribution in matchingAttributions) {
+        bool areAttributionsMergeable = newAttribution.canMergeWith(matchingAttribution);
+        if (!areAttributionsMergeable || !ignoreMergeableOverlaps) {
+          int? conflictStart;
+          int? conflictEnd;
+
+          for (int i = start; i <= end; ++i) {
+            if (hasAttributionAt(i, attribution: matchingAttribution)) {
+              conflictStart ??= i;
+              conflictEnd = i;
+
+              if (areAttributionsMergeable) {
+                // Both attributions are mergeable, but the caller doesn't want to merge them.
+                throw IncompatibleOverlappingAttributionsException(
+                  existingAttribution: matchingAttribution,
+                  newAttribution: newAttribution,
+                  conflictStart: conflictStart,
+                );
+              }
+            } else if (conflictStart != null) {
+              // We found the end of the conflict.
+              conflicts.add(AttributionConflict(
+                newAttribution: newAttribution,
+                existingAttribution: matchingAttribution,
+                conflictStart: conflictStart,
+                conflictEnd: conflictEnd!,
+              ));
+
+              // Reset so we can find the next conflict.
+              conflictStart = null;
+              conflictEnd = null;
+            }
+          }
+
+          if (conflictStart != null && conflictEnd != null) {
+            // We found a conflict that extends to the end of the range.
+            conflicts.add(AttributionConflict(
+              newAttribution: newAttribution,
+              existingAttribution: matchingAttribution,
+              conflictStart: conflictStart,
+              conflictEnd: conflictEnd,
+            ));
+          }
+        }
+      }
+
+      if (conflicts.isNotEmpty && !overwriteConflictingSpans) {
+        throw IncompatibleOverlappingAttributionsException(
+          existingAttribution: conflicts.first.existingAttribution,
+          newAttribution: newAttribution,
+          conflictStart: conflicts.first.conflictStart,
+        );
+      }
     }
 
     // Removes any conflicting attributions. For example, consider the following text,
@@ -346,9 +414,8 @@ class AttributedSpans {
       );
     }
 
-    if (!autoMerge) {
-      // There are not conflicting attributions in the desired range, and we don't
-      // want to merge this new attribution with any other nearby attribution.
+    if (!ignoreMergeableOverlaps) {
+      // We don't want to merge this new attribution with any other nearby attribution.
       // Therefore, we can blindly create the new attribution range without any
       // further adjustments, and then be done.
       _insertMarker(SpanMarker(
@@ -537,63 +604,6 @@ class AttributedSpans {
     } else {
       addAttribution(newAttribution: attribution, start: start, end: end);
     }
-  }
-
-  /// Finds all attributions that appear between [start] and [end] (inclusive) with
-  /// the same id as the given [attribution].
-  ///
-  /// If [allowMerging] is `false`, attributions with the same id will always be reported
-  /// as conflicting, even if [Attribution.canMergeWith] returns `true`.
-  List<AttributionConflict> findConflictingAttributions({
-    required Attribution attribution,
-    required int start,
-    required int end,
-    bool allowMerging = true,
-  }) {
-    final conflicts = <AttributionConflict>[];
-
-    // Check if conflicting attributions overlap the new attribution.
-    final matchingAttributions = getMatchingAttributionsWithin(attributions: {attribution}, start: start, end: end);
-
-    if (matchingAttributions.isNotEmpty) {
-      for (final matchingAttribution in matchingAttributions) {
-        if (!attribution.canMergeWith(matchingAttribution) || !allowMerging) {
-          int? conflictStart;
-          int? conflictEnd;
-
-          for (int i = start; i <= end; ++i) {
-            if (hasAttributionAt(i, attribution: matchingAttribution)) {
-              conflictStart ??= i;
-              conflictEnd = i;
-            } else if (conflictStart != null) {
-              // We found the end of the conflict.
-              conflicts.add(AttributionConflict(
-                newAttribution: attribution,
-                existingAttribution: matchingAttribution,
-                conflictStart: conflictStart,
-                conflictEnd: conflictEnd!,
-              ));
-
-              // Reset so we can find the next conflict.
-              conflictStart = null;
-              conflictEnd = null;
-            }
-          }
-
-          if (conflictStart != null && conflictEnd != null) {
-            // We found a conflict that extends to the end of the range.
-            conflicts.add(AttributionConflict(
-              newAttribution: attribution,
-              existingAttribution: matchingAttribution,
-              conflictStart: conflictStart,
-              conflictEnd: conflictEnd,
-            ));
-          }
-        }
-      }
-    }
-
-    return conflicts;
   }
 
   /// Returns [true] if the given [attribution] exists from [start] to
--- a/attributed_text/lib/src/attributed_spans.dart
+++ b/attributed_text/lib/src/attributed_spans.dart
@@ -96,7 +96,7 @@ class AttributedSpans {
     final matchingAttributions = <Attribution>{};
     for (int i = start; i <= end; ++i) {
       for (final attribution in attributions) {
-        final otherAttributions = getAllAttributionsAt(start);
+        final otherAttributions = getAllAttributionsAt(i);
         for (final otherAttribution in otherAttributions) {
           if (otherAttribution.id == attribution.id) {
             matchingAttributions.add(otherAttribution);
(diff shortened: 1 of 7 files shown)


angelosilvestre:
@KevinBrendel Could you please try this PR?

matthew-carroll:
@angelosilvestre I'm wondering if an `AttributedText` user should be able to insert a span of green within blue and have that issue automatically handled by `AttributedText`. What do you think?

Requiring explicit intervention on this point might proliferate these same behaviors through different commands. Also every app that uses `AttributedText` would have to do the same thing for a similar situation.

Being able to find conflicting attribution spans is probably fine either way, but I'm thinking it might make more sense to do something like splitting blue, and adding green, within AttributedText, itself.

angelosilvestre:
@matthew-carroll I guess it should be fine to handle that automatically.

Can you think of an use case where an `AttributedText` should prevent an attribution addition to cause a conflicting attribution to be split?

KevinBrendel:
@angelosilvestre Thanks for the PR!
I tried it and it generally does exactly what is needed.

For my purposes, the editor-level commands are what is needed, but I can see the argument for adding this feature directly into AttributedText.

I also found one issue, which is likely related to this:
If the SuperEditor has a selectedTextColorStrategy which modifies the color, there can be exceptions triggered by addAttribution being called on AttributedSpans.

matthew-carroll:
> Can you think of an use case where an AttributedText should prevent an attribution addition to cause a conflicting attribution to be split?

Not off the top of my head. But you could add a boolean to whichever methods are relevant. Maybe by default we automatically split, but if you set the flag we'll throw an exception when you try to do that.

matthew-carroll:
> If the SuperEditor has a selectedTextColorStrategy which modifies the color, there can be exceptions triggered by addAttribution being called on AttributedSpans.

@KevinBrendel - Please be sure to provide reproduction steps and a code sample if you run into issues so that @angelosilvestre can reproduce it and fix it.

KevinBrendel:
> @KevinBrendel - Please be sure to provide reproduction steps and a code sample if you run into issues so that @angelosilvestre can reproduce it and fix it.

Will try to create a minimal sample when I have some more time tomorrow. Doesn't seem to be a very special case that triggers it.

KevinBrendel:
Here is the reproduction. Just try to select the red text ("Hello"):
```dart

void main() {
  runApp(const MainApp());
}

  const MainApp({super.key});

  @override
}

  static const _colorAttribution = ColorAttribution(Colors.red);
    ParagraphNode(
      id: Editor.createNodeId(),
      text: AttributedText(
          'Hello, world!',
          AttributedSpans(attributions: [
            const SpanMarker(attribution: _colorAttribution, offset: 0, markerType: SpanMarkerType.start),
            const SpanMarker(attribution: _colorAttribution, offset: 4, markerType: SpanMarkerType.end),
          ])),
    ),
  ]);
  late final Editor _editor;

  @override
  void initState() {
    super.initState();
  }

  @override
    return MaterialApp(
      home: Scaffold(
        body: Center(
          child: SuperEditor(
            editor: _editor,
            document: _document,
            composer: _composer,
              selectedTextColorStrategy: ({required originalTextColor, required selectionHighlightColor}) =>
                  Colors.grey,
            ),
          ),
        ),
      ),
    );
  }
}
```

angelosilvestre:
@matthew-carroll Modified the implementation to handle attribution splitting directly in `AttributedSpans`.

angelosilvestre:
@matthew-carroll This flag might be a bit unintuitive, since we already have a `allowMerge` parameter.

With `allowMerge == false`, even mergeable attributions cause the add method to throw.

attributed_text/lib/src/attributed_spans.dart
```diff
@@ -285,13 +285,20 @@ class AttributedSpans {
...
-  /// overlaps a merge-able span but [autoMerge] is `false`, a
-  /// [IncompatibleOverlappingAttributionsException] is thrown.
+  /// overlaps a merge-able span but [autoMerge] is `false`, one of the following
+  /// outcomes happen:
+  ///
+  /// - If [splitConflictingAttributions] is `true`, the conflicting span is removed from
+  /// the overlapping region, and the new attribution is added in its place.
+  ///
+  /// - If [splitConflictingAttributions] is `false`, an [IncompatibleOverlappingAttributionsException]
+  /// is thrown.
```
matthew-carroll, line 295: Here's a new Dart Doc to completely replace the existing one:

```
/// Applies the [newAttribution] from [start] to [end], inclusive.
///
/// If [start] is less than `0`, nothing happens.
///
/// [AttributedSpans] doesn't have any knowledge about content length, so [end] can
/// take any value that's desired. However, users of [AttributedSpans] should take
/// care to avoid values for [end] that exceed the content length.
///
/// The effect of adding an attribution is straight forward when the text doesn't 
/// contain any other attributions with the same ID. However, there are various
/// situations where [newAttribution] can't necessarily co-exist with other
/// attribution spans that already exist in the text.
///
/// Attribution overlaps can take one of two forms: mergeable or conflicting.
///
/// ## Mergeable Attribution Spans
/// An example of a mergeable overlap is where two bold spans overlap each
/// other. All bold attributions are interchangeable, so when two bold spans
/// overlap, those spans can be merged together into a single span.
///
/// However, mergeable overlapping spans are not automatically merged. Instead,
/// this decision is left to the user of this class. If you want [AttributedSpans] to
/// merge overlapping mergeable spans, pass `true` for `autoMerge`. Otherwise,
/// if `autoMerge` is `false`, an exception is thrown when two mergeable spans
/// overlap each other.
///
///
/// ## Conflicting Attribution Spans
/// An example of a conflicting overlap is where a black text color overlaps a red
/// text color. Text is either black, OR red, but never both. Therefore, the black
/// attribution cannot co-exist with the red attribution. Something must be done
/// to resolve this.
///
/// There are two possible ways to handle conflicting overlaps. The new attribution
/// can overwrite the existing attribution where they overlap. Or, an exception can be
/// thrown. To overwrite the existing attribution with the new attribution, pass `true`
/// for `overwriteConflictingSpans`. Otherwise, if `overwriteConflictingSpans` 
/// is `false`, an exception is thrown.
```

attributed_text/lib/src/attributed_spans.dart
```diff
@@ -285,13 +285,20 @@ class AttributedSpans {
...
   void addAttribution({
     required Attribution newAttribution,
     required int start,
     required int end,
     bool autoMerge = true,
+    bool splitConflictingAttributions = true,
```
matthew-carroll, line 301: Let's call this `overwriteConflictingSpans`.

Also, as discussed offline, let's ensure the implementation is such that `autoMerge` only impacts cases of mergeable spans, and `overwriteConflictingSpans` only impacts non-mergeable spans. Let's not have any crossover control.

attributed_text/lib/src/attributed_spans.dart
```diff
@@ -301,27 +308,42 @@ class AttributedSpans {
...
+    //   |xxxxx-----xxxxx|
+    //
+    // After that, we can apply the desired attribution, because there isn't a conflicting attribution
+    // in this range anymore.
+    for (final conflict in conflicts) {
+      removeAttribution(
+        attributionToRemove: conflict.existingAttribution,
+        start: conflict.conflictStart,
+        end: conflict.conflictEnd,
+      );
```
matthew-carroll, line 346: Down below this is an if-statement with `if (!autoMerge)...` - we discussed offline that we should be able to delete that block because mergeable overlaps should have been handled at the top.

attributed_text/lib/src/attributed_spans.dart
```diff
@@ -517,6 +539,63 @@ class AttributedSpans {
...
+  /// Finds all attributions that appear between [start] and [end] (inclusive) with
+  /// the same id as the given [attribution].
+  ///
+  /// If [allowMerging] is `false`, attributions with the same id will always be reported
+  /// as conflicting, even if [Attribution.canMergeWith] returns `true`.
+  List<AttributionConflict> findConflictingAttributions({
+    required Attribution attribution,
+    required int start,
+    required int end,
+    bool allowMerging = true,
```
matthew-carroll, line 551: Let's call this `ignoreMergeableOverlaps`

attributed_text/lib/src/attributed_spans.dart
```diff
@@ -517,6 +539,63 @@ class AttributedSpans {
     }
   }
 
+  /// Finds all attributions that appear between [start] and [end] (inclusive) with
+  /// the same id as the given [attribution].
+  ///
+  /// If [allowMerging] is `false`, attributions with the same id will always be reported
+  /// as conflicting, even if [Attribution.canMergeWith] returns `true`.
+  List<AttributionConflict> findConflictingAttributions({
```
matthew-carroll, line 547: To verify, do you want this to be a public API? Do you think it should be?

attributed_text/lib/src/attributed_spans.dart
```diff
@@ -1197,3 +1276,45 @@ class IncompatibleOverlappingAttributionsException implements Exception {
     return 'Tried to insert attribution ($newAttribution) over a conflicting existing attribution ($existingAttribution). The overlap began at index $conflictStart';
   }
 }
+
+/// A conflict between the [newAttribution] and [existingAttribution] between [conflictStart] and [conflictEnd] (inclusive).
+///
+/// This means [newAttribution] and [existingAttribution] have the same id, but they can't be merged.
+class AttributionConflict {
```
matthew-carroll, line 1283: Verifying that you think this should be public.

attributed_text/test/attributed_spans_test.dart
```diff
@@ -601,10 +601,126 @@ void main() {
             newAttribution: _LinkAttribution(url: 'https://pub.dev'),
             start: 4,
             end: 12,
+            splitConflictingAttributions: false,
           );
         }, throwsA(isA<IncompatibleOverlappingAttributionsException>()));
       });
 
+      test('splits incompatible attributions at the beginning', () {
```
matthew-carroll, line 609: Let's revisit the test names to ensure we're using the most accurate names given that we've identified the concept is better known as "overwriting" in some cases than "splitting".

attributed_text/test/attributed_spans_test.dart
```diff
@@ -601,10 +601,126 @@ void main() {
...
+              ),
+              ..._createSpanMarkersForAttribution(
+                attribution: _LinkAttribution(url: 'https://pub.dev'),
+                startOffset: 4,
+                endOffset: 12,
+              ),
+            ],
+          ),
+        );
+      });
```
matthew-carroll, line 722: How about a test where a new attribution covers multiple conflicting attributions:

```
XXXXXXX---XXXXXX---XXXXXX  <- existing
------XXXXXXXXXXXXXXX-----  <- new
```

super_editor/test/super_editor/supereditor_attributions_test.dart
```diff
@@ -1276,108 +1276,201 @@ void main() {
       });
     });
 
-    testWidgetsOnArbitraryDesktop('does not merge different colors', (tester) async {
-      final context = await tester //
+    testWidgetsOnArbitraryDesktop('splits spans with different colors', (tester) async {
```
matthew-carroll, line 1279: Let's also revisit the use of the term "split" in this test suite to make sure it's always the best name for the given situation.

attributed_text/lib/src/attributed_spans.dart
KevinBrendel: It might be a good idea not to hide too much functionality from the user. Some use-cases require some more low level knowledge/control and I have had to copy SuperEditor code into my app code on multiple occasions because it was private.

KevinBrendel: I am not 100% sure yet wether we will need this specific function, as implementation of the related features will only commence after this PR is merged.

It is just the kind of helper function that could be useful for some probably unforeseen use-cases and is currently often private. E.g. `_textInSelection` from `CommonEditorOperations` was very useful to build a search feature that already pre-populates the search field with the currently selected text upon pressing CMD+F.

I can create issues/pull requests for such things, but I also understand that it increases the maintenance burden. And I expect that dealing with all kinds of minor issues/pull requests would also serve as a distraction from the more important issues you guys are working on.

Currently I am trying to only create issues/pull requests when it seems something is more generally applicable, as our use-case is probably a bit more exotic (a hybrid flowchart/mind mapping app where there can be hundreds of nodes on a canvas, each containing a SuperEditor which has to interact with our node and graph-level systems for styling, undo/redo, etc. and which has to be integrated into often complex interactions requiring fine-grained control).

matthew-carroll: Filing issues is pretty much free. It only takes the time needed to explain the use-case/motivation. No need to submit PRs for those things. We can consider the motivation and decide whether we think it warrants a public API.

attributed_text/lib/src/attributed_spans.dart
```diff
@@ -301,32 +331,91 @@ class AttributedSpans {
...
+    // in this range anymore.
+    for (final conflict in conflicts) {
+      removeAttribution(
+        attributionToRemove: conflict.existingAttribution,
+        start: conflict.conflictStart,
+        end: conflict.conflictEnd,
+      );
+    }
+
+    if (!ignoreMergeableOverlaps) {
```
matthew-carroll, line 417: At this point in the code, why does `ignoreMergeableOverlaps` matter? Either there aren't any overlaps, in which case this value doesn't matter, or there are overlaps and you've already thrown an exception, right?

attributed_text/lib/src/attributed_spans.dart
matthew-carroll: You resolved this but there's no comment here on why this should be public, nor do I think I saw any comment on the PR about this...

attributed_text/test/attributed_spans_test.dart
```diff
@@ -601,10 +601,178 @@ void main() {
             newAttribution: _LinkAttribution(url: 'https://pub.dev'),
             start: 4,
             end: 12,
+            overwriteConflictingSpans: false,
           );
         }, throwsA(isA<IncompatibleOverlappingAttributionsException>()));
       });
 
+      test('overwrites incompatible attributions at the beginning', () {
```
matthew-carroll, line 609: Please be more specific about what "beginning" you're talking about. There's the beginning of text, there's the beginning of an existing span, there's the beginning of a new span, etc.

Please adjust all test descriptions accordingly.

attributed_text/test/attributed_spans_test.dart
```diff
@@ -601,10 +601,178 @@ void main() {
...
+                startOffset: 5,
+                endOffset: 6,
+              ),
+            ],
+          ),
+        );
+      });
+
+      test('splits incompatible attributions at the middle', () {
+        final spans = AttributedSpans(attributions: [
```
matthew-carroll, line 645: In each test, please try to add a little comment with attribution swimlanes for the starting value and the final value so that the reader doesn't need to carefully read and process each start and end offset to figure out what's happening in these tests.

attributed_text/test/attributed_spans_test.dart
```diff
@@ -813,3 +981,47 @@ class _LinkAttribution implements Attribution {
...
 }
+
+/// Creates an [AttributedSpans] for the [attribution] starting at [startOffset]
+/// and ending at [endOffset].
+AttributedSpans _createAttributedSpansForAttribution({
+  required Attribution attribution,
+  required int startOffset,
+  required int endOffset,
+}) {
+  return AttributedSpans(
```
matthew-carroll, line 992: Should we add a constructor that's more concise? For example: `AttributedSpans.single(attribution, start, end)`?

super_editor/lib/src/default_editor/text.dart
```diff
@@ -1274,7 +1274,7 @@ class AddTextAttributionsCommand implements EditCommand {
               newAttribution: attribution,
               start: range.start,
               end: range.end,
-              autoMerge: autoMerge,
+              ignoreMergeableOverlaps: autoMerge,
```
matthew-carroll, line 1277: Now that I'm noticing that `AttributedText` exposes the same API as `AttributedSpans`, maybe we should leave the name `autoMerge` in `AttributedSpans` so that we don't drift away from `AttributedText`.

super_editor/lib/src/default_editor/text.dart
```diff
@@ -1552,11 +1552,14 @@ class ToggleTextAttributionsCommand implements EditCommand {
...
           node.text = AttributedText(
             node.text.text,
-            node.text.spans.copy(),
-          )..addAttribution(
-              attribution,
-              range,
-            );
+            node.text.spans.copy()
+              ..addAttribution(
+                newAttribution: attribution,
```
matthew-carroll, line 1557: Let's also leave this parameter called `attribution` to avoid unnecessary breaking changes.

attributed_text/lib/src/attributed_spans.dart
angelosilvestre: This code is needed for the cases where we try to apply a mergeable attribution right after another one. For example:

```
|one two|
|bbb----|
```

When `ignoreMergeableOverlaps` is `true`, applying a bold attribution to the span " two" produces a single span covering the entire text:

```
|one two|
|bbbbbbb|
```

When `ignoreMergeableOverlaps` is `false`, applying a bold attribution to the span " two " produces two spans:

```
|one two|
|bbb----|
|---bbbb|
```

attributed_text/lib/src/attributed_spans.dart
angelosilvestre: It seems I forgot to make it private.

attributed_text/test/attributed_spans_test.dart
angelosilvestre: It seems that after this isn't being used anymore... I removed it

super_editor/lib/src/default_editor/text.dart
angelosilvestre: The `addAttribution` used before this PR is from `AttributedText`. Now we are using the method from `AttributedSpans`. We are not changing the parameter names.

matthew-carroll (approved):
LGTM

KevinBrendel:
> Here is the reproduction. Just try to select the red text ("Hello"):
> 
> ```dart
> import 'package:flutter/material.dart';
> import 'package:super_editor/super_editor.dart';
> 
> void main() {
>   runApp(const MainApp());
> }
> 
> class MainApp extends StatefulWidget {
>   const MainApp({super.key});
> 
>   @override
>   State<MainApp> createState() => _MainAppState();
> }
> 
> class _MainAppState extends State<MainApp> {
>   static const _colorAttribution = ColorAttribution(Colors.red);
>   final _document = MutableDocument(nodes: [
>     ParagraphNode(
>       id: Editor.createNodeId(),
>       text: AttributedText(
>           'Hello, world!',
>           AttributedSpans(attributions: const [
>             SpanMarker(attribution: _colorAttribution, offset: 0, markerType: SpanMarkerType.start),
>             SpanMarker(attribution: _colorAttribution, offset: 4, markerType: SpanMarkerType.end),
>           ])),
>     ),
>   ]);
>   final _composer = MutableDocumentComposer();
>   late final Editor _editor;
> 
>   @override
>   void initState() {
>     super.initState();
>     _editor = createDefaultDocumentEditor(document: _document, composer: _composer);
>   }
> 
>   @override
>   Widget build(BuildContext context) {
>     return MaterialApp(
>       home: Scaffold(
>         body: Center(
>           child: SuperEditor(
>             editor: _editor,
>             document: _document,
>             composer: _composer,
>             stylesheet: defaultStylesheet.copyWith(
>               selectedTextColorStrategy: ({required originalTextColor, required selectionHighlightColor}) =>
>                   Colors.grey,
>             ),
>           ),
>         ),
>       ),
>     );
>   }
> }
> ```

@matthew-carroll @angelosilvestre This was not fixed and there is not way around it as the corresponding implementation hardcodes `overwriteConflictingSpans` to false.

Is this intentional, and if so, is it possible to work around this?

Otherwise, `selectedTextColorStrategy` will stay unusable in combination with `ColorAttribution`.

matthew-carroll:
@KevinBrendel can you please file that reproduction and the problem as a new issue? I'll have @angelosilvestre investigate it when you do.

PR #2036, opened 2024-05-25
[SuperEditor][Android] Fix auto-scrolling start region (Resolves #2000)

### LINKED ISSUE #2000: Android interactor assumes global positions which delays auto scrolling
**Package Version**
`stable`

**To Reproduce**
1. Run the example on Android (the bug affects Android handlers / interactor)
2. Scroll down to the end of the document
3. Long press on a word
4. Drag one of the extended handlers towards the top of the screen
5. Observe how the document only begins to auto-scroll after dragging over the app header

**Minimal Reproduction Code**
<details>

  ```dart

void main() {
  runApp(const MainApp());
}

  const MainApp({super.key});

  @override
}

    for (var i = 0; i < 40; i++)
      ParagraphNode(
        id: Editor.createNodeId(),
        text: AttributedText('Some random text. ' * (1 + i)),
      ),
  ]);
  late final _editor = createDefaultDocumentEditor(
    composer: _composer,
    document: _document,
  );

  @override
  void dispose() {
    _editor.dispose();
    _document.dispose();
    _composer.dispose();
    super.dispose();
  }

  @override
    return MaterialApp(
      home: Scaffold(
        appBar: const PreferredSize(
          preferredSize: Size(double.infinity, 90),
          child: SafeArea(
            child: SizedBox(
              height: 90,
              child: ColoredBox(color: Colors.black),
            ),
          ),
        ),
        body: SuperEditor(
          editor: _editor,
          document: _document,
          composer: _composer,
        ),
      ),
    );
  }
}
  ```
</details>

**Actual behavior**
Currently, on Android, autoscroll is only started after dragging the over the scaffold's header. On iOS, the autoscroll happens immediately at the boundary of the document.

**Expected behavior**
The auto-scroll should begin before the start of the document and not above it.
(It would also be nice to be able to control where the scroll start boundary is. It seems the bits are already there just not exposed through `SuperEditor`)

**Platform**
Android

**Flutter version**
```
Flutter 3.22.0 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 5dcb86f68f (5 days ago) • 2024-05-09 07:39:20 -0500
Engine • revision f6344b75dc
Tools • Dart 3.4.0 • DevTools 2.34.3
```

**Additional context**
Originally, I wanted to adapt the example above to always use the Android gesture mode, however, it seems that setting `gestureMode` only effects selection styling. Popover toolbar and selection handlers are not displayed. It seems this is because `SuperEditorAndroidHandlesDocumentLayerBuilder` and other builders test for the `defaultTargetPlatform` to decide whether to build or not thus ignorring the `gestureMode`.

angelosilvestre:
[SuperEditor][Android] Fix auto-scrolling start region. Resolves #2000

On Android, the auto-scrolling only starts when the user drags near the top of the screen. This causes problems when there are widgets above the editor, like an `AppBar`:

[before.webm](https://github.com/superlistapp/super_editor/assets/7597082/2961af48-10c5-4937-982f-ca3bede79daf)

The reason is that we are not converting the `Offset` to the offset inside the viewport.

This PR fixes the issue by converting the `Offset`:

[after.webm](https://github.com/superlistapp/super_editor/assets/7597082/1fc6e8fb-cbec-4af0-ad86-df28279a678c)

### DIFF
--- a/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
@@ -1335,18 +1335,6 @@ class SuperEditorAndroidControlsOverlayManagerState extends State<SuperEditorAnd
   final _dragHandleSelectionGlobalFocalPoint = ValueNotifier<Offset?>(null);
   final _magnifierFocalPoint = ValueNotifier<Offset?>(null);
 
-  /// Returns the `RenderBox` for the scrolling viewport.
-  ///
-  /// If this widget has an ancestor `Scrollable`, then the returned
-  /// `RenderBox` belongs to that ancestor `Scrollable`.
-  ///
-  /// If this widget doesn't have an ancestor `Scrollable`, then this
-  /// widget includes a `ScrollView` and this `State`'s render object
-  /// is the viewport `RenderBox`.
-  RenderBox get viewportBox =>
-      (context.findAncestorScrollableWithVerticalScroll?.context.findRenderObject() ?? context.findRenderObject())
-          as RenderBox;
-
   @override
   void initState() {
     super.initState();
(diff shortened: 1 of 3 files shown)


super_editor/lib/src/default_editor/document_gestures_touch_android.dart
```diff
@@ -1335,6 +1335,18 @@ class SuperEditorAndroidControlsOverlayManagerState extends State<SuperEditorAnd
...
 
+  /// Returns the `RenderBox` for the scrolling viewport.
+  ///
+  /// If this widget has an ancestor `Scrollable`, then the returned
+  /// `RenderBox` belongs to that ancestor `Scrollable`.
+  ///
+  /// If this widget doesn't have an ancestor `Scrollable`, then this
+  /// widget includes a `ScrollView` and this `State`'s render object
+  /// is the viewport `RenderBox`.
+  RenderBox get viewportBox =>
```
matthew-carroll, line 1346: We have similar mapping methods in other widgets. Is it usually declared above `initState()`?

angelosilvestre: I checked and we declare below `dispose`. Updated.

matthew-carroll (approved):
LGTM

PR #2038, opened 2024-05-25
Slack-style tag plugin

matthew-carroll:
This PR adds a new plugin which brings Slack-style tagging to Super Editor. Slack-style tagging is very similar to our existing "stable tag" plugin. However, Slack tagging appears to consider everything between the trigger character and the caret as the search term. Additionally, Slack tokenizes the search term and then matches names token by token.

The following videos demonstrate the plugin in action. The colors in these demos are optional and controllable. We use colors in the video to show exactly where the composing and stable attributions are applied.

https://github.com/superlistapp/super_editor/assets/7259036/12fdcb29-4910-484e-bb38-8a0cd61cef97

https://github.com/superlistapp/super_editor/assets/7259036/16c073cb-f013-462e-bb61-50358e537d2a

### DIFF
--- a/super_editor/lib/src/clones/slack/slack_tags.dart
+++ b/super_editor/lib/src/clones/slack/slack_tags.dart
@@ -1,0 +1,1803 @@
+import 'dart:math';
+
+import 'package:attributed_text/attributed_text.dart';
+import 'package:collection/collection.dart';
+import 'package:flutter/cupertino.dart';
+import 'package:flutter/foundation.dart';
+import 'package:flutter/services.dart';
+import 'package:super_editor/src/core/document.dart';
+import 'package:super_editor/src/core/document_composer.dart';
+import 'package:super_editor/src/core/document_selection.dart';
+import 'package:super_editor/src/core/edit_context.dart';
+import 'package:super_editor/src/core/editor.dart';
+import 'package:super_editor/src/default_editor/document_hardware_keyboard/document_input_keyboard.dart';
+import 'package:super_editor/src/default_editor/multi_node_editing.dart';
+import 'package:super_editor/src/default_editor/super_editor.dart';
+import 'package:super_editor/src/default_editor/text.dart';
+import 'package:super_editor/src/default_editor/text_tokenizing/tags.dart';
+import 'package:super_editor/src/infrastructure/_logging.dart';
+import 'package:super_editor/src/infrastructure/keyboard.dart';
+import 'package:super_editor/src/infrastructure/strings.dart';
+
+/// A [SuperEditor] plugin that adds the ability to create Slack-style tags, such as
+/// persistent user references, e.g., "@dash".
+///
+/// Slack tagging includes three modes:
+///  * Composing: a tag is being assembled, i.e., typed.
+///  * Committed: a tag is done being assembled - it's now uneditable.
+///  * Unbound: a tag is doing being assembled, but it references an unknown entity - it's still editable.
+///  * Cancelled: a tag was being composed, but the composition was cancelled.
+class SlackTagPlugin extends SuperEditorPlugin {
+  /// The key used to access the [SlackTagIndex] in an attached [Editor].
+  static const slackTagIndexKey = "slackTagIndex";
+
+  static const _trigger = "@";
+
+  SlackTagPlugin() : tagIndex = SlackTagIndex() {
+    _requestHandlers = <EditRequestHandler>[
+      (request) =>
+          request is FillInComposingSlackTagRequest ? FillInComposingSlackTagCommand(_trigger, request.tag) : null,
+      (request) => request is CancelComposingSlackTagRequest //
+          ? const CancelComposingSlackTagCommand(_trigger)
+          : null,
+    ];
+
+    _reactions = [
+      SlackTagReaction(
+        trigger: _trigger,
+        onUpdateComposingTag: tagIndex._onComposingTagFound,
+      ),
+      const AdjustSelectionAroundSlackTagReaction(_trigger),
+    ];
+  }
+
+  /// Index of all slack tags in the document, which changes as the user adds and removes tags.
+  final SlackTagIndex tagIndex;
+
+  @override
+  void attach(Editor editor) {
+    editor
+      ..context.put(SlackTagPlugin.slackTagIndexKey, tagIndex)
+      ..requestHandlers.insertAll(0, _requestHandlers)
+      ..reactionPipeline.insertAll(0, _reactions);
+  }
+
+  @override
+  void detach(Editor editor) {
+    editor
+      ..context.remove(SlackTagPlugin.slackTagIndexKey)
+      ..requestHandlers.removeWhere((item) => _requestHandlers.contains(item))
+      ..reactionPipeline.removeWhere((item) => _reactions.contains(item));
+  }
+
+  late final List<EditRequestHandler> _requestHandlers;
+
+  late final List<EditReaction> _reactions;
+
+  @override
+  List<DocumentKeyboardAction> get keyboardActions => [_cancelOnEscape];
+  ExecutionInstruction _cancelOnEscape({
+    required SuperEditorContext editContext,
+    required KeyEvent keyEvent,
+  }) {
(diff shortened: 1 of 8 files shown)


matthew-carroll:
CC @BazinC @josephrcox

matthew-carroll:
A couple change requests have been made:
 * Don't show the popover when there aren't any results
 * Cancel the composition if a space immediately follows the trigger

matthew-carroll:
Video of latest two modifications


https://github.com/superlistapp/super_editor/assets/7259036/7629ace4-87b5-43c5-bb84-67c7c9c3cc69

BazinC:
Thank you @matthew-carroll ! I'll try it in a bit :)

super_editor/lib/src/clones/slack/slack_tags.dart
```diff
@@ -0,0 +1,1517 @@
...
+        SelectionReason.contentChange,
+      ),
+    );
+
+    // Remove the composing region after we apply the stable attribution to avoid a
+    // reaction automatically re-applying the composing tag.
+    // FIXME: Use a transaction to bundle these changes so order doesn't matter.
+    if (removeComposingAttributionCommand != null) {
+      executor.executeCommand(removeComposingAttributionCommand);
+      print("Attributions immediately after removing composing attribution:");
```
BazinC, line 210: There are a few print statements that need to be changed to editorSlackTagsLoga

super_editor/lib/src/clones/slack/slack_tags.dart
```diff
@@ -0,0 +1,1517 @@
...
+            attributionFilter: (attribution) => attribution is CommittedSlackTagAttribution,
+            range: SpanRange(0, textNode.text.length - 1),
+          )
+          .sorted((tag1, tag2) => tag2.start - tag1.start);
+
+      // Track the impact of deletions on selection bounds, then update the selection
+      // to reflect the deletions.
+      final composer = editContext.find<MutableDocumentComposer>(Editor.composerKey);
+
+      final baseBeforeDeletions = composer.selection!.base;
```
BazinC, line 639: Selection should be check if null before reading base

super_editor/lib/src/clones/slack/slack_tags.dart
```diff
@@ -0,0 +1,1517 @@
...
+        // Move the caret to the side of the tag in the direction of push motion.
+        _pushCaretToOppositeTagEdge(editContext, requestDispatcher, selectionChangeEvent, textNode.id, tagAroundCaret);
+        break;
+      case SelectionChangeType.placeExtent:
+      case SelectionChangeType.pushExtent:
+      case SelectionChangeType.expandSelection:
+        throw AssertionError(
+            "A collapsed selection reported a SelectionChangeType for an expanded selection: ${selectionChangeEvent.changeType}\n${selectionChangeEvent.newSelection}");
+      case SelectionChangeType.clearSelection:
+        throw AssertionError("Expected a collapsed selection but there was no selection.");
```
BazinC, line 1023: I would change these thrown errors with logs, at least for the beginning. For instance if you try the plugin on android with drag handles, moving the handle will throw here (SelectionChangeType.pushCaret). Editor is then stuck.

super_editor/lib/src/clones/slack/slack_tags.dart
```diff
@@ -0,0 +1,1517 @@
...
+  TagAroundPosition? _findTagAtCaret(
+    EditContext editContext,
+    bool Function(Set<Attribution> attributions) tagSelector,
+  ) {
+    final composer = editContext.find<MutableDocumentComposer>(Editor.composerKey);
+    if (composer.selection == null || !composer.selection!.isCollapsed) {
+      // We only tag when the selection is collapsed. Our selection is null or expanded. Return.
+      return null;
+    }
+    final selectionPosition = composer.selection!.extent;
```
BazinC, line 756: same as above
```suggestion
    final selection = composer.selection;
    if (selection == null) {
      return null;
    }
    if (!selection.isCollapsed) {
      // We only tag when the selection is collapsed or expanded. Return.
      return null;
    }

    final selectionPosition = selection.extent;
```

super_editor/lib/src/clones/slack/slack_tags.dart
```diff
@@ -0,0 +1,1517 @@
...
+/// Slack tagging includes three modes:
+///  * Composing: a tag is being assembled, i.e., typed.
+///  * Committed: a tag is done being assembled - it's now uneditable.
+///  * Unbound: a tag is doing being assembled, but it references an unknown entity - it's still editable.
+///  * Cancelled: a tag was being composed, but the composition was cancelled.
+class SlackTagPlugin extends SuperEditorPlugin {
+  /// The key used to access the [SlackTagIndex] in an attached [Editor].
+  static const slackTagIndexKey = "slackTagIndex";
+
+  static const _trigger = "@";
```
BazinC, line 34: Please add a way to provide another trigger if the user wants to

super_editor/lib/src/clones/slack/slack_tags.dart
```diff
@@ -0,0 +1,1517 @@
...
+
+  SlackTagPlugin() : tagIndex = SlackTagIndex() {
+    _requestHandlers = <EditRequestHandler>[
+      (request) =>
+          request is FillInComposingSlackTagRequest ? FillInComposingSlackTagCommand(_trigger, request.tag) : null,
+      (request) => request is CancelComposingSlackTagRequest //
+          ? const CancelComposingSlackTagCommand()
+          : null,
+    ];
```
BazinC, line 43: Can we think of letting package user provide custom handlers? Let's say in the example the popover enable the end user to pick not only a username, but also different kind of entity. The package user would use the plugin to get the composing value from the tagIndex, and send custom request to commit not only user tags.

super_editor/lib/src/clones/slack/slack_tags.dart
```diff
@@ -0,0 +1,1517 @@
...
+
+  const SlackTagFinder._();
+}
+
+/// An attribution for a slack tag that's currently being composed.
+const slackTagComposingAttribution = NamedAttribution("slack-tag-composing");
+
+/// An attribution for a slack tag that was previously being composed, but was then
+/// abandoned (not cancelled) without committing it.
+const slackTagUnboundAttribution = NamedAttribution("slack-tag-unbound");
```
BazinC, line 1482: unused

matthew-carroll:
@BazinC - As mentioned on our recent call, what I'm looking to get from you is a list of UX requirements to consider this feature done. Doing a per-line code review isn't relevant until this code actually does what's needed.

Can you please provide the list of UX things this plugin needs to do, which it currently doesn't do?

BazinC:
Hello @matthew-carroll, I understand this PR is not finished but I tried it and noticed bugs not related to any specific UX requirements so I found relevant to report them now, with repro steps and point the involved lines of code.

Furthermore, I think the plugin could be more open to let Super Editor package users do more than it is possible in the current state.

Requirements:
- Having the possibility to use a different trigger, not only @ (not possible in the current state. see https://github.com/superlistapp/super_editor/pull/2038#discussion_r1686586012)
- Search should not happen in the middle of a word (see https://github.com/superlistapp/super_editor/pull/2038#issuecomment-2156058849)
- Fix Android bug (currently crashing. see https://github.com/superlistapp/super_editor/pull/2038#discussion_r1686578195)
- Null bug fixed (currently crashing if editor executes clear selection. see https://github.com/superlistapp/super_editor/pull/2038#discussion_r1686261441)

- Finally, the plugin in its current state can only handle request like "commit a username string" (FillInComposingSlackTagRequest/Command, which commits text with slackTagAttribution.) What if, like in slack, we would like to commit a "team" (a group of users). What if we want to give more info to the text attribution (like user id, avatar url..)?
Do you see what I mean? There is a lot of possibilities. I cannot provide specific UX requirement for that.
I think the plugin is great to find, manage and expose composing tag, but it should not be exclusively responsible of what can be committed. That is why I proposed to make the plugin to expose it's tag index and accept custom request handlers.

matthew-carroll:
Note: We want to support triggers of both arbitrary content and arbitrary character count.

We'll further discuss what team commitment means.

PR #2047, opened 2024-05-27
Fix lint warnings

matthew-carroll:
Fix lint warnings - mostly switching from material state to widget state (essentially a rename).


angelosilvestre (approved):
LGTM

PR #2049, opened 2024-05-27
[super_editor_markdown] Prevent empty syntax from being converted (Resolves #2033)

### LINKED ISSUE #2033: [SuperEditor] Typing `~~` removes the text
**Package Version**
Latest main

**To Reproduce**
1. Go to SuperEditor demo
2. Type ~~

**Actual behavior**
The text is removed

**Expected behavior**
The text is preserved

**Platform**
Any platform

**Flutter version**
Fluter 3.22 or master channel

**Additional context**
This happens because the markdown plugin is attempting to use `~~` as a strikethrough syntax, even though there is no text between the characters.

angelosilvestre:
[super_editor_markdown] Prevent empty syntax from being converted. Resolves #2033

Typing "~~" or "``" is causing the text to be removed.

This happens because the markdown plugin looks for opening and closing syntaxes, but does not check if there is any text between then. As a result, we try to apply strikethrough/code attributions to an empty text.

This PR changes the markdown plugin to check the parsed text length. If it's exactly the length of the opening + closing characters, we don't perform any conversion.

### DIFF
--- a/super_editor_markdown/lib/src/markdown_inline_upstream_plugin.dart
+++ b/super_editor_markdown/lib/src/markdown_inline_upstream_plugin.dart
@@ -485,10 +485,17 @@ class StyleUpstreamMarkdownToken implements UpstreamMarkdownToken {
         }
 
         if (_openingSyntax == _closingSyntax) {
-          // We just found an opening syntax that matches our closing syntax.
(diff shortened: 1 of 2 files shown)


matthew-carroll (approved):
LGTM

PR #2057, opened 2024-05-29
[SuperEditor][Mobile] Fix caret position after rotation (Resolves #2056)

### LINKED ISSUE #2056: Cursor at wrong position after portrait to landscape rotation
https://github.com/superlistapp/super_editor/assets/96958/c957e1f8-a627-49b6-b9e1-1fdc1349d7e4

Seems like a regression caused by https://github.com/superlistapp/super_editor/pull/1967

The problem is `IosControlsDocumentLayerState` accessing the RenderBox to get size inside `computeLayoutDataWithDocumentLayout`, but the way `ContentLayers` currently work, the `computeLayoutDataWithDocumentLayout` method is called during build, before the render box is laid out. 

cc @angelosilvestre

angelosilvestre:
[SuperEditor][Mobile] Fix caret position after rotation. Resolves #2056

On mobile, rotating the phone is causing the caret to be displayed at the wrong position:

https://github.com/superlistapp/super_editor/assets/96958/c957e1f8-a627-49b6-b9e1-1fdc1349d7e4

This was caused by https://github.com/superlistapp/super_editor/pull/1967

We are accessing the layer's `RenderBox` to get the size, to be able to adjust the caret for situations where the caret is pushed off-screen.

The issue is that, at the point in the pipeline, the layer hasn't been laid out yet for the current frame.

To fix the issue, I changed it to get the width from the content's `RenderBox` instead of the layer's  `RenderBox`. Since both should always be the same size, this shouldn't be an issue.

We have tests for caret position after phone rotation, but I'm not sure yet why these tests didn't catch this issue. Maybe it's related to the text used in those tests...

Also, it seems that when rotating from landscape to portrait, this issue doesn't happen. It only seems to happen when rotating from portrait to landscape.


knopp:
I tested it and it fixes the issue.

super_editor/lib/src/infrastructure/platforms/android/android_document_controls.dart
```diff
@@ -353,6 +361,12 @@ class AndroidControlsDocumentLayerState
     }
   }
 
+  @override
+  DocumentSelectionLayout? computeLayoutDataWithDocumentLayout(BuildContext context, DocumentLayout documentLayout) {
```
matthew-carroll, line 365: Is there a reason that this method now needs to be implemented when it wasn't before?

super_editor/lib/src/infrastructure/platforms/android/android_document_controls.dart
```diff
@@ -321,13 +327,15 @@ class AndroidControlsDocumentLayerState
       // Default caret width used by the Android caret.
       const caretWidth = 2;
 
-      final layerBox = context.findRenderObject() as RenderBox?;
-      if (layerBox != null && layerBox.hasSize && caretRect.left + caretWidth >= layerBox.size.width) {
+      // Use the content's RenderBox instead of the layer's RenderBox, because at this point, the layer
```
matthew-carroll, line 330: Can you explain this? I'm not sure I understand what this is saying, or why it's the case...

super_editor/test_goldens/editor/supereditor_caret_test.dart
```diff
@@ -192,3 +292,22 @@ Future<void> _pumpCaretTestApp(WidgetTester tester) async {
     ],
   ).pump();
 }
+
+Future<TestDocumentContext> _pumpPhoneRotationTestApp(WidgetTester tester) async {
```
matthew-carroll, line 296: Can you add a comment for when/how this is meant to be used? Typically pumping is a thing that happens initially, so the idea of pumping for rotation isn't immediately obvious without further instructions.

super_editor/lib/src/infrastructure/platforms/android/android_document_controls.dart
angelosilvestre: Updated the comment.

super_editor/lib/src/infrastructure/platforms/android/android_document_controls.dart
angelosilvestre: Previously, all the implementation was already in the `computeLayoutDataWithDocumentLayout` method. But now that we need to access the content's `RenderBox` I switched to use `computeLayoutData` directly.

Now we don't need `computeLayoutDataWithDocumentLayout`,  but since it doesn't have a default implementation in the superclass, we need to implement it...

super_editor/test_goldens/editor/supereditor_caret_test.dart
angelosilvestre: I changed the name to make it simpler. This method has just the initial pump, I added the `phoneRotation` just to clarify in which kind of test this is used.

super_editor/lib/src/infrastructure/platforms/android/android_document_controls.dart
```diff
@@ -309,7 +309,13 @@ class AndroidControlsDocumentLayerState
   }
 
   @override
-  DocumentSelectionLayout? computeLayoutDataWithDocumentLayout(BuildContext context, DocumentLayout documentLayout) {
+  DocumentSelectionLayout? computeLayoutData(Element? contentElement, RenderObject? contentLayout) {
```
matthew-carroll, line 312: I took a look back the declaration of `DocumentLayoutLayer` and it seems like we're missing something more centrally. The whole point of `DocumentLayoutLayer` is that subclasses don't need to implement `computeLayoutData()`. Otherwise, the subclass could simply implement `ContentLayer`. 

So is the root problem here that we can't get the `RenderObject` associated with a `DocumentLayout`?

Historically, I've done my best to keep render object knowledge out of `DocumentLayout`, but I'm wondering if this means that `DocumentLayout` should have a `RenderObject? get renderObject`?

Or, if it's still desirable to keep render object stuff out of `DocumentLayout`, then maybe we should pass the content's `Element` to `computeLayoutDataWithDocumentLayout()`? Maybe something like:

```
LayoutDataType? computeLayoutDataWithDocumentLayout(BuildContext contentLayersContext, BuildContext documentContext, DocumentLayout documentLayout);
```

What do you think?

angelosilvestre: I think I prefer to provide the `contentLayersContext` directly to `computeLayoutDataWithDocumentLayout` instead of a `RenderObject` getter.

Another option is to add a `Size? get size` in `DocumentLayout`, but it might not work for sliver-based document layouts.

matthew-carroll: Render objects can be used for more than just size, so we probably shouldn't overconstrain the API to size. Go ahead with passing both contexts if that makes sense.

super_editor/test_goldens/editor/supereditor_caret_test.dart
```diff
@@ -1,13 +1,16 @@
...
 import 'package:golden_toolkit/golden_toolkit.dart';
 import 'package:super_editor/super_editor.dart';
 import 'package:super_editor/super_editor_test.dart';
 
 import '../../test/super_editor/supereditor_test_tools.dart';
 import '../test_tools_goldens.dart';
 
-void main() {
+Future<void> main() async {
+  await loadAppFonts();
```
matthew-carroll, line 12: Is this `loadAppFonts()` going to impact tests that are run in other files after this one?

angelosilvestre: It shouldn't. If we don't specify a style rule with the font family we will still use Ahem.

matthew-carroll: Can you verify that? I was just doing some golden work in swift_ui and if I call `loadAppFonts()` from the test config, I get regular fonts in the goldens. The only difference between calling `loadAppFonts()` from the test config vs calling it from a specific test file is that some tests might run before the test file and won't be impacted. But I'm worried about any tests that run after the given test file.

angelosilvestre: It can impact if other files specify the font family. Once we load the fonts, it seems we cannot unload them.

matthew-carroll: Then should we move this into the global test config to avoid burying it in a single file?

angelosilvestre: It seems our global test config already has that, so we don't need to add it in this file. Removed.

matthew-carroll (approved):
LGTM

PR #2061, opened 2024-05-29
[SuperEditor] Allow continuing list item sequence by typing (Resolves #1894)

### LINKED ISSUE #1894: [SuperEditor] Numbered list doesn't work on manual addition of items and when number is not 1
If you type
```dart
1. List Item
```
and press enter/return, the result is as follows:
```dart
1. List Item
2. 
```

This is correct behaviour. But if you type:
```dart
1. List Item

```
and then manually enter `2.` below it, the attribution doesn't change.
Ideally it should change it to the second list item, resulting in:
```dart
1. List Item
2. 
```

> In some text editors, even typing any other number other than one greater than the previous list item, changes the list item correctly. For example, 
 if you type:
```dart
1. List Item

```
and then manually enter `4.` below it, it changes it to:
```dart
1. List Item
2. 
```

#### Video Example

https://github.com/superlistapp/super_editor/assets/92245694/b2d3afb5-81af-4926-a15f-c52a9eae8ef4

#### Related Issue

If you type `1. `, it changes to list attribution, but typing any other number `2. ` or `12. ` etc. doesn't change the attribution to numbered list attribution.

Ideally it should start a list from that very number. For example, `2. ` should start a list from 2.

https://github.com/superlistapp/super_editor/assets/92245694/a8c0c774-f693-4e42-bbca-2dc0e2b3c553

angelosilvestre:
[SuperEditor] Allow continuing list item sequence by typing. Resolves #1894 

### Current behavior

Type: 

```
1. List item
```

Press enter. This will result in the following:

```
1. List Item
2.
```
Delete `2.` and type `2.` again.

The editor doesn't convert `2.` to a list item.

### Other apps

Most apps do convert the `2.` to a list item, allowing the user to continue the sequence. Some of them allow numbers out of sequence, like:

```
1. List Item
2. List Item
5.
```

And some of them allows starting a list with a number different from `1`.

### Changes in this PR

This PR adds the ability to continue the list, by typing the next number in the sequence. For example, given the list

```
1. List Item
2. List Item
|
```

Typing `3.` will convert the paragraph to a list item, but typing any other number will not.

### Limitations

We cannot allow starting lists with numbers different from one without major changes. The list item ordinal value is only computed and stored at the view model level. To allow starting a list with arbitrary numbers we would need to store the ordinal value directly on the node.

### DIFF
--- a/super_editor/lib/src/default_editor/list_items.dart
+++ b/super_editor/lib/src/default_editor/list_items.dart
@@ -113,22 +113,7 @@ class ListItemComponentBuilder implements ComponentBuilder {
 
     int? ordinalValue;
     if (node.type == ListItemType.ordered) {
-      // Counts the number of ordered list items above the current node with the same indentation level. Ordered
-      // list items with the same indentation level might be separated by ordered or unordered list items with
-      // different indentation levels.
-      ordinalValue = 1;
-      DocumentNode? nodeAbove = document.getNodeBefore(node);
-      while (nodeAbove != null && nodeAbove is ListItemNode && nodeAbove.indent >= node.indent) {
-        if (nodeAbove.indent == node.indent) {
-          if (nodeAbove.type != ListItemType.ordered) {
-            // We found an unordered list item with the same indentation level as the ordered list item.
-            // Other ordered list items aboce this one do not belong to the same list.
-            break;
-          }
-          ordinalValue = ordinalValue! + 1;
-        }
-        nodeAbove = document.getNodeBefore(nodeAbove);
-      }
+      ordinalValue = computeListItemOrdinalValue(node, document);
     }
 
     return switch (node.type) {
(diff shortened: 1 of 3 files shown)


super_editor/lib/src/default_editor/list_items.dart
```diff
@@ -1198,3 +1183,31 @@ ExecutionInstruction splitListItemWhenEnterPressed({
   final didSplitListItem = editContext.commonOps.insertBlockLevelNewline();
   return didSplitListItem ? ExecutionInstruction.haltExecution : ExecutionInstruction.continueExecution;
 }
+
+/// Computes the ordinal value of an ordered list item.
+///
+/// Walks backwards counting the number of ordered list items above the [listItem] with the same indentation level.
+///
+/// The ordinal value starts at 1.
+int computeListItemOrdinalValue(ListItemNode listItem, Document document) {
```
matthew-carroll, line 1192: Why did this need to be extracted? Why does it need to be public?

angelosilvestre: I extracted it because I need the same computation in the reaction, which lives in another file.  I extracted to avoid having the same code in two places, but I can just copy the code to there if needed.

PR #2063, opened 2024-05-30
[SuperEditor] Allow custom view models to be aware of node selection (Resolves #1944)

### LINKED ISSUE #1944: Selection colour of components with custom view models
Fist of all super editor is an amazing tool, thanks for creating it 🚀

**User Behavior**
I would like to be able to provide a selection colour for my custom components. For the view models provided by package the selection colour seems to be assigned inside a private function _applySelection in SingleColumnLayoutSelectionStyler, but other components do not change colour when selected.  (I also noticed horizontal ruler selection colour differs from the rest. Would be great to be able to override it)

**Feature Value**
It provides proper feedback to user that the element is selected. 

**Additional context**
If this is achievable already in a different way please let me know 🙏

angelosilvestre:
[SuperEditor] Allow custom view models to be aware of node selection. Resolves #1944

Currently, during the styling pass, we only apply the selection and selection color to our own viewmodels: `TextComponentViewModel`, `ImageComponentViewModel` and `HorizontalRuleComponentViewModel`.

Because of that, app-level view models, don't receive this information

https://github.com/superlistapp/super_editor/assets/7597082/e0e89088-f030-4013-bd64-0db68ae2f5e7

This PR introduces a `SelectionAwareViewModelMixin`. Any view model that mixes it in will receive the `DocumentNodeSelection` and selection color during the style pass.

Using this mixin, app-level components can easily render their selection:

https://github.com/superlistapp/super_editor/assets/7597082/c9c01546-b601-49f7-bd77-e1d145c88d9d

### DIFF
--- a/super_editor/lib/src/default_editor/horizontal_rule.dart
+++ b/super_editor/lib/src/default_editor/horizontal_rule.dart
@@ -76,14 +76,14 @@ class HorizontalRuleComponentBuilder implements ComponentBuilder {
 
 class HorizontalRuleComponentViewModel extends SingleColumnLayoutComponentViewModel with SelectionAwareViewModelMixin {
   HorizontalRuleComponentViewModel({
-    required String nodeId,
-    double? maxWidth,
-    EdgeInsetsGeometry padding = EdgeInsets.zero,
+    required super.nodeId,
+    super.maxWidth,
+    super.padding = EdgeInsets.zero,
     DocumentNodeSelection? selection,
     Color selectionColor = Colors.transparent,
     this.caret,
     required this.caretColor,
-  }) : super(nodeId: nodeId, maxWidth: maxWidth, padding: padding) {
+  }) {
     super.selection = selection;
     super.selectionColor = selectionColor;
   }
--- a/super_editor/lib/src/default_editor/horizontal_rule.dart
+++ b/super_editor/lib/src/default_editor/horizontal_rule.dart
@@ -1,5 +1,6 @@
 import 'package:attributed_text/attributed_text.dart';
 import 'package:flutter/material.dart';
+import 'package:super_editor/src/default_editor/layout_single_column/selection_aware_viewmodel.dart';
 import 'package:super_editor/src/default_editor/selection_upstream_downstream.dart';
 
 import '../core/document.dart';
@@ -65,7 +66,7 @@ class HorizontalRuleComponentBuilder implements ComponentBuilder {
 
     return HorizontalRuleComponent(
       componentKey: componentContext.componentKey,
-      selection: componentViewModel.selection,
+      selection: componentViewModel.selection?.nodeSelection as UpstreamDownstreamNodeSelection?,
       selectionColor: componentViewModel.selectionColor,
       showCaret: componentViewModel.caret != null,
       caretColor: componentViewModel.caretColor,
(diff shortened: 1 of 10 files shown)


angelosilvestre:
@kasyrm Could you please try this PR to see if it fits your needs?

super_editor/lib/src/default_editor/layout_single_column/_styler_user_selection.dart
```diff
@@ -168,6 +169,11 @@ class SingleColumnLayoutSelectionStyler extends SingleColumnLayoutStylePhase {
         ..selection = selection
         ..selectionColor = _selectionStyles.selectionColor;
     }
+    if (viewModel is SelectionAwareViewModelMixin) {
```
matthew-carroll, line 172: To stress test this approach, can you refactor the other view models to use this mixin, too? If a solution is versatile enough for all 3rd party components then it should be versatile enough for our existing components.

super_editor/lib/src/default_editor/selection_aware_viewmodel.dart
```diff
@@ -0,0 +1,35 @@
+import 'package:flutter/material.dart';
```
matthew-carroll, line 1: I'm thinking this file should probably live under the `layout_single_column` directory

super_editor/lib/src/default_editor/layout_single_column/_styler_user_selection.dart
angelosilvestre: Updated.

super_editor/lib/src/default_editor/horizontal_rule.dart
```diff
@@ -65,27 +66,28 @@ class HorizontalRuleComponentBuilder implements ComponentBuilder {
...
     EdgeInsetsGeometry padding = EdgeInsets.zero,
-    this.selection,
-    required this.selectionColor,
+    DocumentNodeSelection? selection,
+    Color selectionColor = Colors.transparent,
     this.caret,
     required this.caretColor,
-  }) : super(nodeId: nodeId, maxWidth: maxWidth, padding: padding);
+  }) : super(nodeId: nodeId, maxWidth: maxWidth, padding: padding) {
+    super.selection = selection;
```
matthew-carroll, line 87: Can you use `super` in the parameter list instead of down here?

super_editor/lib/src/default_editor/image.dart
```diff
@@ -122,27 +123,28 @@ class ImageComponentBuilder implements ComponentBuilder {
...
     EdgeInsetsGeometry padding = EdgeInsets.zero,
     required this.imageUrl,
     this.expectedSize,
-    this.selection,
-    required this.selectionColor,
-  }) : super(nodeId: nodeId, maxWidth: maxWidth, padding: padding);
+    DocumentNodeSelection? selection,
+    Color selectionColor = Colors.transparent,
+  }) : super(nodeId: nodeId, maxWidth: maxWidth, padding: padding) {
+    this.selection = selection;
```
matthew-carroll, line 142: Same question here about `super`.

kasyrm:
> @kasyrm Could you please try this PR to see if it fits your needs?

Yes! Thank you :)
Is it correct to assume nodeSelection of selection will be UpstreamDownstreamNodeSelection, so it's possible to check whether component is fully selected or selection is collapsed?

matthew-carroll:
> Is it correct to assume nodeSelection of selection will be UpstreamDownstreamNodeSelection

@angelosilvestre if that is assumed, it probably shouldn't get. Imagine something like a table component, or anything else more complicated than a singe block, that component would need to use its own definition of selection.

angelosilvestre:
@kasyrm @matthew-carroll It depends on the node type. For `BlockNode` subclasses, it will be an 
`UpstreamDownstreamNodeSelection` by default. For `TextNode` subclasses, it will be a `TextNodeSelection` by default.

Any node can override `DocumentNode.computeSelection` to return its own selection representation.

super_editor/lib/src/default_editor/horizontal_rule.dart
angelosilvestre: We can for nodeId, maxWidth and padding, but not for selection and selectionColor. Since the mixin can't have a constructor we cannot use the `super` keyword here.

matthew-carroll (approved):
LGTM

PR #2064, opened 2024-05-30
[SuperEditor] Prevent long-press to resolve when swiping to pop (Resolves #1947)

### LINKED ISSUE #1947: When swiping to pop, Magnifier and Toolbar appears until released.
**Package Version**
super_editor:
    git:
      url: https://github.com/superlistapp/super_editor
      path: super_editor
      ref: stable # or "main"
  super_editor_markdown:
    git:
      url: https://github.com/superlistapp/super_editor
      path: super_editor_markdown
      ref: stable
  super_text_layout:
    git:
      url: https://github.com/superlistapp/super_editor
      path: super_text_layout
      ref: stable
  attributed_text:
    git:
      url: https://github.com/superlistapp/super_editor
      path: attributed_text
      ref: stable

**To Reproduce**
1. Open Editor Screen
2. Swipe to the right to pop

**Actual behavior**
Magnifier and Toolbar show up until finger is released

**Expected behavior**
Magnifier and Toolbar should stay hidden

**Platform**
iPhone 12 Pro Max iOS 17.4.1

**Flutter version**
Flutter Channel stable, 3.19.3
Dart 3.3.1 • DevTools 2.31.1

**Screenshots**
Bildschirmfoto 2024-04-19 um 12 01 08

**Additional context**
The FocusNode gets focused when swiping to pop the screen. This should somehow be avoided.

angelosilvestre:
[SuperEditor] Prevent long-press to resolve when swiping to pop. Resolves #1947

On mobile, dragging from left to right starts a swipe-to-pop gesture. However, when doing so we are changing the selection on both mobile platforms, showing the magnifier on iOS and showing the mobile toolbar on Android:

[android_swipe_before.webm](https://github.com/superlistapp/super_editor/assets/7597082/dba2c11b-bc28-4b80-9758-ac26da7a9274)

https://github.com/superlistapp/super_editor/assets/7597082/ea9484bc-1c88-4897-b5f5-b2ca9f61c07b

The issue is that, even though the horizontal drag recognizer is winning in the gesture arena, the `onTapDown` is still called, causing the long-press timer to be started.

To resolve that, we need to cancel the timer when we lose on the arena. 

I found another issue, where the `onCancel` isn't being called after the recognizer was defeated in the arena, even if the `onTapDown` was already fired.

This PR changes the `TapSequenceGestureRecognizer` to call `onTapCancel` after we lose on the arena and change the mobile interactors to cancel the long-press timers on that callback.

I wasn't able to write a failing test for Android, I guess this back gesture is handled different for Android, since I'm not seeing any other recognizers on the arena after starting the swipe.

With this PR's implementation:

https://github.com/superlistapp/super_editor/assets/7597082/b5826302-65e3-4dc2-8ada-f3602f1eede4

[android_swipe_after.webm](https://github.com/superlistapp/super_editor/assets/7597082/4a976c98-da24-4448-bd11-a0dd16a31a4e)

### DIFF
--- a/super_editor/lib/src/infrastructure/multi_tap_gesture.dart
+++ b/super_editor/lib/src/infrastructure/multi_tap_gesture.dart
@@ -199,22 +199,38 @@ class TapSequenceGestureRecognizer extends GestureRecognizer {
     tracker.entry.resolve(GestureDisposition.rejected);
     _freezeTracker(tracker);
 
-    // Both _firstTap and _secondTap are only set when a pointer up event is received,
-    // but we also need to handle cases when we already fired the onTapDown event, and
-    // we were defeated on the gesture arena before the onTapUp event. For example, if
-    // an app has both a TapSequenceGestureRecognizer and a HorizontalDragGestureRecognizer,
-    // when the user taps down and then drag horizontally, the onTapDown event is fired, and after that
-    // the HorizontalDragGestureRecognizer declares itself as the winner. In this case, we need to
-    // fire the onTapCancel event. Check if _firstTapDownDetails is not null to confirm if the tap
-    // down already happened.
-    if (_firstTap != null || _secondTap != null || _firstTapDownDetails != null) {
-      if (tracker == _firstTap || tracker == _secondTap) {
+    if (_firstTap == null && _firstTapDownDetails != null) {
+      // The user tapped down and then another recognizer won the arena. For example, in an app with both a
+      // TapSequenceGestureRecognizer and a HorizontalDragGestureRecognizer, when the user taps down and
+      // then drags horizontally, the onTapDown event is fired, and after that the HorizontalDragGestureRecognizer
+      // declares itself as the winner. Invoke onTapCancel to cancel the gesture.
+      _checkCancel();
+      if (_trackers.isEmpty) {
         _reset();
-      } else {
-        _checkCancel();
-        if (_trackers.isEmpty) {
-          _reset();
-        }
+      }
+      return;
+    }
+
+    if (tracker == _secondTap) {
+      // A double tap was registered and we were defeated on the gesture arena after that. Reset
+      // to clean up the tap trackers.
+      _reset();
+      return;
+    }
+
+    if (tracker == _firstTap) {
+      // A tap up was registered and we were defeated on the gesture arena after that. Reset
+      // to clean up the tap trackers.
+      _reset();
+      return;
+    }
+
+    if (_firstTap != null || _secondTap != null) {
+      // We have a single or double tap registered, but the tracker isn't related to any of them.
+      // It's not clear what this situation means.
+      _checkCancel();
+      if (_trackers.isEmpty) {
+        _reset();
       }
     }
   }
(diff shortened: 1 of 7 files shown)


super_editor/lib/src/infrastructure/multi_tap_gesture.dart
```diff
@@ -198,7 +198,10 @@ class TapSequenceGestureRecognizer extends GestureRecognizer {
     _trackers.remove(tracker.pointer);
     tracker.entry.resolve(GestureDisposition.rejected);
     _freezeTracker(tracker);
-    if (_firstTap != null || _secondTap != null) {
+
+    // Check if we need to cancel even when both _firstTap and _secondTap are null,
```
matthew-carroll, line 202: I don't really understand this comment. Can you try rephrasing it?

super_editor/lib/src/infrastructure/multi_tap_gesture.dart
```diff
@@ -372,7 +377,7 @@ class TapSequenceGestureRecognizer extends GestureRecognizer {
   }
 
   void _checkCancel() {
-    if (_firstTap == null && onTapCancel != null) {
+    if ((_firstTap == null || _firstTapDownDetails != null) && onTapCancel != null) {
```
matthew-carroll, line 380: Can you add an explanation comment to this line, too?

angelosilvestre: Updated.

super_editor/lib/src/infrastructure/multi_tap_gesture.dart
angelosilvestre: Updated.

super_editor/lib/src/infrastructure/multi_tap_gesture.dart
```diff
@@ -198,7 +198,16 @@ class TapSequenceGestureRecognizer extends GestureRecognizer {
     _trackers.remove(tracker.pointer);
     tracker.entry.resolve(GestureDisposition.rejected);
     _freezeTracker(tracker);
-    if (_firstTap != null || _secondTap != null) {
+
+    // Both _firstTap and _secondTap are only set when a pointer up event is received,
```
matthew-carroll, line 202: I think this is good info, but I'm wondering if we're combining unrelated situations in a single if-statement?

Most of this code was copied and adjusted from Flutter, so don't consider the existing code to be particularly readable. If we can improve the readability, we should.

I notice the first if-statement has 3 OR conditions, which sounds like 3 different things. Then within that if-statement we have another if/else, and the else contains another if. It's a lot of different conditions.

Can you try to refactor this little piece into unique conditions, where each condition explains what it represnts?

angelosilvestre: @matthew-carroll Updated. Can you take a look to see if it's better now? 

BTW, we still have some unusual code here. In `_reject`, there are situations that we call `_reset`. Then, on `_reset`, we can nullify `_firstTap` and `_secondTap` and recursively call `_reject` again a couple times.

matthew-carroll: I assume that's stuff copied from Flutter. I had no idea how this thing worked when I adapted it to our needs. Any time we figure out a little bit of it, we should add comments. Also, where possible, if we can simplify an area that we're working on, we should.

super_editor/lib/src/infrastructure/multi_tap_gesture.dart
angelosilvestre: It should be an `||`. If the user tapped down, but not released the tap, `_firstTap` is `null`.

matthew-carroll: That still sounds like an `&&` condition: "the user has tapped down AND the user has not released the tap" - isn't that the condition?

angelosilvestre: Currently, if the user:

- Taps down
- Release the tap
- Taps again
- Drags horizontally

We will still fire the cancel event. If we change it to `&&` it will no longer be the case. Is this expected?

matthew-carroll (approved):
LGTM

PR #2076, opened 2024-06-08
[SuperEditor][iOS] Don't show toolbar upon first tap (Resolves #2073)

### LINKED ISSUE #2073: [SuperEditor][iOS] - Popover toolbar appears when it shouldn't
**Package Version**
main

**To Reproduce**
1. Go to SuperEditor demo
2. Add a few empty paragraphs
3. Tap on an empty paragraph to place there the selection

**Actual behavior**
The selection is placed at the paragraph and the toolbar is shown.

**Expected behavior**
The selection is placed at the paragraph. Only a second tap should reveal the toolbar

**Platform**
Only reproducible on iOS.

**Flutter version**
Flutter master

**Screenshots**

https://github.com/superlistapp/super_editor/assets/31278849/d40ab717-971b-42e9-a334-911a8165d75a

**Additional context**
Similar issue to https://github.com/superlistapp/super_editor/issues/2005.

angelosilvestre:
[SuperEditor][iOS] Don't show toolbar upon first tap. Resolves #2073

On iOS, tapping to change the selection from one empty paragraph to another is causing the toolbar to appear.

The issue is that, when we check if the user tapped on an existing selection, we aren't checking for the node id. Therefore, whenever the user places the caret to another paragraph at the same offset as the previous selection, the toolbar is displayed:

https://github.com/superlistapp/super_editor/assets/7597082/ba7deab0-be10-478c-9462-b9e5ff26c5cc

This PR fixes the issue by adding the node id to the comparison.

### DIFF
--- a/super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
@@ -643,6 +643,7 @@ class _IosDocumentTouchInteractorState extends State<IosDocumentTouchInteractor>
     if (docPosition != null) {
       final didTapOnExistingSelection = selection != null &&
           selection.isCollapsed &&
+          selection.extent.nodeId == docPosition.nodeId &&
           selection.extent.nodePosition.isEquivalentTo(docPosition.nodePosition);
 
       if (didTapOnExistingSelection) {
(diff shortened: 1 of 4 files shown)


angelosilvestre:
@matthew-carroll A couple android golden tests started to fail due some small rendering change related to the scaleFactor. 

I increased the tolerance to make those tests pass, but we can re-generated then if needed.

matthew-carroll (approved):
LGTM

PR #2087, opened 2024-06-11
[SuperTextField] Re-generate failing caret test goldens (Resolves #2085)

### LINKED ISSUE #2085: Fix golden failures
We have 3 or 4 goldens currently failing. Try to find out why they're failing and let's get them passing again so PRs don't look like they broke new things.

angelosilvestre:
[SuperTextField] Re-generate failing caret test goldens. Resolves #2085

There are some failing tests related to caret rendering. These failures look like legitimate failures. However, comparing the master image with the test image, we can see that now the caret is better aligned with the toolbar (I draw a red line to demonstrate that).

Since these tests started to fail without any changes from our side, I assume that some caret related method in `RenderParagraph` is now returning a different value.

Master image:

SCR-20240611-rumb

Test image:
image

This PR updates the failing goldens.


super_editor/test_goldens/super_textfield/goldens/super_textfield_font_height.png
matthew-carroll, line 1: On this one it seems like maybe we lost the caret off to the left. Can you check, for example, where the caret appears on the left side of an image when the image takes full screen width?

KevinBrendel:
These look like the expected results of the changes I made in #1945 and #1982. Though for some reason, these particular goldens did not fail when making those pull requests.

super_editor/test_goldens/super_textfield/goldens/super_textfield_font_height.png
angelosilvestre: Yeah, the caret still appears:

image

We also have a golden test to ensure the caret appears.

matthew-carroll (approved):
LGTM

PR #2089, opened 2024-06-12
[golden_tester] Remove docker image after running the command (Resolves #2088)

### LINKED ISSUE #2088: [BUG] - Golden tester does not remove image after running
After using our golden test tool, the docker container is correctly removed, but the image isn't removed.

As a result, we end up with a bunch of dangling images.

Update the tool to remove the image after running.

angelosilvestre:
[golden_tester] Remove docker image after running the command. Resolves #2088

After running the golden tester commands, the container is removed but the image is kept. As a result, we end up with a bunch of dangling images.

This PR changes the tool to remove the image after running the commands. It also changes how we handle failures in the tests. Currently, we are throwing and exiting with an unhandled exception when tests fail. With these changes, we will gracefully exit the process after running the commands.

### DIFF
--- a/golden_runner/lib/src/commands.dart
+++ b/golden_runner/lib/src/commands.dart
@@ -111,7 +111,11 @@ class GoldenTestCommand extends Command {
         ...cmdArguments,
       ],
       description: 'Golden tests',
+      throwOnError: false,
     );
+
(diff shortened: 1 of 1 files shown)


matthew-carroll (approved):
LGTM - Great update. This will save TB in no time.

PR #2096, opened 2024-06-13
[SuperEditor][SuperTextField][iOS] Update selection handle appearance (Resolves #1924)

### LINKED ISSUE #1924: [SuperIOSTextField] Update text selection handles sizing and layout to match native iOS design spec
`SuperIOSTextField` text selection handles doesn't follow the native iOS design specs as documented [here](https://docs.google.com/document/d/1g7qHcknab9Av4kxw78145wyVrj5wF5_UEGyzt8Exhzk/edit). This includes the text selection handle bar and the circle that is shown above/below the handle.

Need to update following things to match the native behaviour: 
* iOS uses handle circle radius of 6px under all circumstances. `SuperIOSTextField` currently uses 4px. Update that. [read](https://docs.google.com/document/d/1g7qHcknab9Av4kxw78145wyVrj5wF5_UEGyzt8Exhzk/edit#heading=h.vtp4ztibjjz)
* On iOS, the drag handle rectangles are exactly as tall as the selection highlight rectangle for a given line, and the top and bottom of those handle rectangles match the selection rectangle. Update the handle bar layout to match the native behaviour.  [read](https://docs.google.com/document/d/1g7qHcknab9Av4kxw78145wyVrj5wF5_UEGyzt8Exhzk/edit#heading=h.cefu8wyhfn2s)
* On iOS, the drag handle circle sits completely above/below the very top/bottom edge of the handle. `SuperIOSTextField` circle overlaps with the selection area. Update it. [read](https://docs.google.com/document/d/1g7qHcknab9Av4kxw78145wyVrj5wF5_UEGyzt8Exhzk/edit#heading=h.cefu8wyhfn2s)

angelosilvestre:
[SuperEditor][SuperTextField][iOS] Update selection handle appearance. Resolves #1924

Our iOS drag handles have some differences compared to the native iOS selection handle:

1. The ball radius is much smaller than native iOS
2. The ball is displayed at the top of the selection box. On native iOS, the ball handle has some overlap with the selection box.

Also, our drag handle line isn't correctly aligned with the selection box. 

Native iOS:

image

SuperEditor:

https://github.com/superlistapp/super_editor/assets/7597082/4b971806-8bcb-4a5c-8142-637ad12bb56d

https://github.com/superlistapp/super_editor/assets/7597082/f62339d9-59d5-4991-8002-c5142bd99552

We use the caret offset and height provided by `RenderParagraph` to compute the drag handle position and height. This doesn't seem to match the selection boxes, also reported by `RenderParagraph`.

To solve that, I modified the way we compute the handle to also use selection boxes. We create an expanded selection at the upstream and downstream positions and compute the selection boxes for each selection edge. That way, the handles look aligned.

With the changes from this PR:

https://github.com/superlistapp/super_editor/assets/7597082/ec5b9b8d-f775-429e-8ad4-3da118c10004

https://github.com/superlistapp/super_editor/assets/7597082/46c19a10-b86f-4594-947c-eb039d7d65ea

There are places where we care about the handle ball diameter and places where we care about the ball radius. Maybe we should change it to always care about the radius?

BTW, the iOS toolbar shadow is looking weird, not sure if it's something that changed after impeller landed on iOS. Also, the `SuperTextField` iOS toolbar doesn't look like it's pointing to the correct position.


matthew-carroll:
> BTW, the iOS toolbar shadow is looking weird, not sure if it's something that changed after impeller landed on iOS. Also, the SuperTextField iOS toolbar doesn't look like it's pointing to the correct position.

Please file corresponding issues with repro's.

super_editor/lib/src/infrastructure/platforms/ios/ios_document_controls.dart
```diff
@@ -672,6 +672,50 @@ class IosControlsDocumentLayerState extends DocumentLayoutLayerState<IosHandlesD
     }
   }
 
+  /// Compute the `Rect` to display a drag handle for an expanded selection.
```
matthew-carroll, line 675: Please specify what the `Rect` means. Is it a zero width rectangle that represents the x offset and the height of the caret? Does it represent the area that the caret should paint (meaning it has a width)? Also what coordinate system is this reported in?

super_editor/lib/src/infrastructure/platforms/ios/ios_document_controls.dart
```diff
@@ -672,6 +672,50 @@ class IosControlsDocumentLayerState extends DocumentLayoutLayerState<IosHandlesD
...
+
+    if (extentNodePosition == null) {
+      // Couldn't find a valid position to the right. Look for a position to the left
+      // of the current position within the same node.
+      extentNodePosition = component.movePositionLeft(position.nodePosition);
+    }
+
+    if (extentNodePosition == null) {
+      // We couldn't expand the selection neither to the left of the right. Fallback
+      // to rect for the position, which matches the caret top and bottom.
```
matthew-carroll, line 699: Saying "matches the caret top and bottom" is a confusing reference because this method is probably being used to decide how to paint the caret, right? If my understanding is correct, then this comment should say something about Flutter's concept of a caret size and how that's different from everything else we're doing in this method.

super_editor/lib/src/super_textfield/ios/_editing_controls.dart
```diff
@@ -426,19 +423,22 @@ class _IOSEditingControlsState extends State<IOSEditingControls>
         ? TextAffinity.downstream
         : TextAffinity.upstream;
 
+    final selectionBoxes = _textLayout.getBoxesForSelection(widget.editingController.textController.selection);
+    final upstreamSelectionBox = selectionBoxes.first;
```
matthew-carroll, line 427: Is this list guaranteed to be non-empty?

super_editor/lib/src/super_textfield/ios/_editing_controls.dart
```diff
@@ -484,12 +484,23 @@ class _IOSEditingControlsState extends State<IOSEditingControls>
...
+    final ballRadius = defaultIosHandleBallDiameter / 2;
+
     return CompositedTransformFollower(
       key: handleKey,
       link: widget.textContentLayerLink,
       offset: followerOffset,
       child: Transform.translate(
-        offset: const Offset(-12, -5),
+        offset: Offset(
+          -12,
```
matthew-carroll, line 495: Let's do something about this magic number. First, is a static value going to have a problem with changing pixel densities? Second, let's at least comment what it means so it can be verified and updated in the future.

super_editor/lib/src/super_textfield/ios/_editing_controls.dart
```diff
@@ -484,12 +484,23 @@ class _IOSEditingControlsState extends State<IOSEditingControls>
...
+          isUpstreamHandle
+              // For the upstream handle, the ball is displayed above the text, partially
+              // overlapping the selected area. Decrease the ball diameter so the ball is positioned
```
matthew-carroll, line 498: > Decrease the ball diameter....

But you're not decreasing the ball diameter, are you?

super_text_layout/lib/src/text_selection_layer.dart
```diff
@@ -173,3 +178,17 @@ class TextSelectionPainter extends CustomPainter {
         selectionColor != oldDelegate.selectionColor;
   }
 }
+
+/// The number of pixels to decrement from the top of selection box to display
```
matthew-carroll, line 182: I'm not sure what this means. What's the "selection box"?

I'm also pretty uncomfortable with dangling global properties like this that have such a specific purpose.

super_editor/lib/src/super_textfield/ios/_editing_controls.dart
angelosilvestre: It doesn't seem to be impacted by pixel densities, but I'm not sure where this number came from.

This is how it looks if we use a zero x-offset: 

simulator_screenshot_F97AA426-C020-4B18-BBDB-7FAF238FE352

angelosilvestre: Even at the very first character it happens. The leader widget is the text itself, so I'm not sure why this is happening. Any ideas?

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
```diff
@@ -1887,4 +1887,4 @@ const defaultIosMagnifierSize = Size(133, 96);
 
 /// The diameter of the small circle that appears on the top and bottom of
 /// expanded iOS text handles.
-const defaultIosHandleBallDiameter = 8.0;
+const defaultIosHandleBallDiameter = 16.0;
```
matthew-carroll, line 1890: Can you clarify in the Dart Doc whether this is px or dpi?

super_editor/lib/src/infrastructure/platforms/ios/ios_document_controls.dart
```diff
@@ -672,6 +672,53 @@ class IosControlsDocumentLayerState extends DocumentLayoutLayerState<IosHandlesD
     }
   }
 
+  /// Computes a zero width `Rect` that represents the x and y offsets and the height
+  /// of the handle in content space.
```
matthew-carroll, line 676: Is this truly only for the "handle" or does it also apply to painting the caret? If so, you should probably say "upstream or downstream handle" for clarity.

super_editor/lib/src/infrastructure/platforms/ios/ios_document_controls.dart
```diff
@@ -672,6 +672,53 @@ class IosControlsDocumentLayerState extends DocumentLayoutLayerState<IosHandlesD
...
+  ///
+  /// The `Rect` returned by [DocumentLayout.getRectForPosition] doesn't match the
```
matthew-carroll, line 678: This part is confusing to read because the reader isn't expecting to hear about something in `DocumentLayout`. You should transition into the topic so that the reader has some context for what you're saying.

I was going to offer a recommendation, but I'm not even sure why we're referencing `getRectForPosition` here. How is that related to this method?

super_editor/lib/src/super_textfield/ios/_editing_controls.dart
matthew-carroll: Can you show a screenshot with it at the first character, and can you also add a border around the widget that's following so we can clearly see the alignment?

super_editor/lib/src/super_textfield/ios/_editing_controls.dart
```diff
@@ -484,12 +507,23 @@ class _IOSEditingControlsState extends State<IOSEditingControls>
...
       key: handleKey,
       link: widget.textContentLayerLink,
       offset: followerOffset,
       child: Transform.translate(
-        offset: const Offset(-12, -5),
+        offset: Offset(
+          -12,
+          isUpstreamHandle
+              // For the upstream handle, the ball is displayed above the text, partially
+              // overlapping the selected area. Decrease the ball diameter from the handle offset
```
matthew-carroll, line 521: I think this is still saying something you don't mean to say. "Decrease the ball diameter" means reduce the diameter of the ball. Do you mean "Move the ball..."?

super_text_layout/lib/src/text_selection_layer.dart
matthew-carroll: If "selection box" means "selection highlight box" then what does "to display the selection highlight" mean? I still don't understand what any of this Dart Doc is trying to say...

angelosilvestre: I updated the code trying to make it simpler. Could you take a look?

matthew-carroll (approved):
LGTM

matthew-carroll:
@angelosilvestre can you resolve the conflicts? Looks like mostly goldens.

angelosilvestre:
@matthew-carroll I resolved the conflicts and re-generated the goldens.

Issue #2098, opened 2024-06-17
[SuperTextField] - Add tapping WidgetTester extensions and implement missing tests


matthew-carroll:
When working on #2080 I wanted to duplicate the Super Editor tap behavior to SuperTextField. However, when writing the tests I noticed that we don't seem to have WidgetTester extensions to tap at character offsets, which makes the tests much more complicated.

Add those extensions and then implement the empty tests in `super_textfield_ios_selection_test.dart`.

PR #2100, opened 2024-06-17
[SuperEditor][SuperTextField][iOS] - When user taps, place caret at word boundary (Resolves #2080)

### LINKED ISSUE #2080: [SuperEditor][iOS] Cursor placement does not reflect native iOS behaviour
**Package Version**
stable branch

There is a difference in how Super Editor responds to user taps on text compared to native experiences. It appears to follow the Android behaviour of placing the cursor as close as possible to the tap location.

When tapping on a word, the majority of the time the cursor should snap to the end of it. If the user taps within a small distance from the start of the word (within 1 character), the cursor should snap to the beginning instead. 

Also, another small difference I just noticed is that the iOS cursor has a subtle animated pulse effect, whereas super editor follows Android design again by having no animation.

|Expected|Actual|
|--|--|
| <video src="https://github.com/superlistapp/super_editor/assets/10269289/28b12b4b-8b90-45b5-95d0-8861742ee0a9"></video> | <video src="https://github.com/superlistapp/super_editor/assets/10269289/b44047f6-1c1d-4395-be3c-988806dab99a"></video> |

Thank you for creating such a great package, I am really looking forward to seeing what I can do with it!

matthew-carroll:
[SuperEditor][SuperTextField][iOS] - When user taps, place caret at word boundary (Resolves #2080)

On iOS, when a user taps on text, the caret isn't placed where the user tapped - it's placed at a word boundary.

This PR updates `SuperEditor` and `SuperTextField` to use a similar heuristic for placing the caret. This PR also disables that behavior in tests by default, so that our tests can explicitly target desired caret positions.

There's some followup work in #2098

### DIFF
--- a/super_editor/lib/src/infrastructure/platforms/ios/selection_heuristics.dart
+++ b/super_editor/lib/src/infrastructure/platforms/ios/selection_heuristics.dart
@@ -1,0 +1,41 @@
+import 'package:super_editor/src/infrastructure/strings.dart';
+
+/// User interaction heuristics that simulate observed behavior on
+/// iOS devices.
+class IosHeuristics {
+  /// Adjusts a user's tap offset within a text field, or a paragraph, by
+  /// moving the caret to a word boundary, as observed on iOS.
+  static int adjustTapOffset(String text, int tapOffset) {
+    assert(tapOffset >= 0 && tapOffset < text.length);
+    if (tapOffset == 0) {
+      return 0;
+    }
+
+    final upstreamWordStart = text.moveOffsetUpstreamByWord(tapOffset) ?? 0;
+    final upstreamWordEnd = text.moveOffsetDownstreamByWord(upstreamWordStart) ?? text.length;
+
+    final downstreamWordEnd = text.moveOffsetDownstreamByWord(tapOffset) ?? text.length;
+    final downstreamWordStart = text.moveOffsetUpstreamByWord(downstreamWordEnd) ?? 0;
+
+    late final int adjustedSelectionOffset;
+    if (text[tapOffset] == " ") {
+      // User tapped between words. Pick the nearest word.
+      adjustedSelectionOffset =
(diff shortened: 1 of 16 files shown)


super_editor/lib/src/infrastructure/platforms/ios/selection_heuristics.dart
```diff
@@ -0,0 +1,41 @@
...
+      if (tapOffset <= upstreamWordEnd) {
+        // The tap position is within the upstream word.
+        adjustedSelectionOffset = tapOffset - upstreamWordStart <= 1 ? upstreamWordStart : upstreamWordEnd;
+      } else {
+        // The tap position is within the downstream word.
+        adjustedSelectionOffset = tapOffset - downstreamWordStart <= 1 ? downstreamWordStart : downstreamWordEnd;
+      }
+    }
+
+    return adjustedSelectionOffset;
```
angelosilvestre, line 37: Don't you think it would be more readable if we use early returns?

super_editor/test/super_editor/mobile/super_editor_ios_selection_test.dart
```diff
@@ -10,6 +10,95 @@ import '../supereditor_test_tools.dart';
 void main() {
   group("SuperEditor mobile selection >", () {
     group("iOS >", () {
+      group("on tap >", () {
+        testWidgetsOnIos("when beyond first character > places caret at end of word", (tester) async {
+          // Note: We pump the following text.
+          // "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod...",
+          await _pumpAppWithLongText(tester);
+
+          // Tap near the end of a word "consectet|ur"
```
angelosilvestre, line 19: Please add a dot at the end of this and the other comments in this file.

angelosilvestre (approved):
LGTM with a couple comments.

PR #2101, opened 2024-06-17
[SuperEditor] - Tasks are now indentable (Resolves #1895)

### LINKED ISSUE #1895: [SuperEditor] Task indentation implementation is missing
Unable to indent checkboxes using tab key, similar to ordered and unordered lists.

### Similar Issue
Pressing enter on an empty checkbox does not make it disappear,. This also works correctly in case of ordered and unordered lists.

matthew-carroll:
[SuperEditor] - Tasks are now indentable (Resolves #1895)

Includes rules that control allowable indents.

- The first task in a list can't be indented.
- A task that appears after another task can have at most the indent of its parent + 1

This PR hooks up Tab to increment the indent and Shift+Tab to decrement the indent.

Incrementing and decrementing indents are also available as `EditRequest`s for mobile integration.

An `EditReaction` is included, which ensures that after deleting a task, no remaining tasks have illegal indents.


matthew-carroll:
@angelosilvestre let me know if you have thoughts about how a task should calculate the distance of an indent, so that a task can co-exist with list items, without looking out of place.

super_editor/lib/src/default_editor/tasks.dart
```diff
@@ -258,6 +284,10 @@ class _TaskComponentState extends State<TaskComponent> with ProxyDocumentCompone
     return Row(
       crossAxisAlignment: CrossAxisAlignment.start,
       children: [
+        // TODO: create a space calculator like list items
+        SizedBox(
+          width: 24.0 * widget.viewModel.indent,
+        ),
```
angelosilvestre, line 290: If we use the same calculation that we use for unordered list items, doesn't it look correct?

super_editor/lib/src/default_editor/tasks.dart
```diff
@@ -366,6 +396,146 @@ ExecutionInstruction backspaceToConvertTaskToParagraph({
...
+  if (node.indent == 0) {
+    // Can't un-indent any further.
+    return ExecutionInstruction.continueExecution;
+  }
+
+  editContext.editor.execute([
+    UnIndentTaskRequest(node.id),
+  ]);
+
+  return ExecutionInstruction.haltExecution;
```
angelosilvestre, line 536: Should this run even if the caret is at the middle of the task? Or should it run only if it's at the beginning of it?

super_editor/test/super_editor/components/task_test.dart
```diff
@@ -91,536 +91,668 @@ void main() {
...
-        ],
-      );
+      testWidgetsOnDesktop("does not apply to child tasks", (tester) async {
+        final document = MutableDocument(
+          nodes: [
+            TaskNode(id: "1", text: AttributedText("parent"), isComplete: false),
+            TaskNode(id: "2", text: AttributedText("child one"), isComplete: false),
+            TaskNode(id: "3", text: AttributedText("child two"), isComplete: false),
+          ],
+        );
```
angelosilvestre, line 647: The test description says "does not apply to child tasks", but all tasks here have the same indentation level. Is this description correct?

super_editor/test/super_editor/components/task_test.dart
```diff
@@ -91,536 +91,668 @@ void main() {
...
+        expect(SuperEditorInspector.findTaskIndent("3"), 1);
+
+        // Ensure the legal tasks below the deleted task weren't impacted.
+        expect(SuperEditorInspector.findTaskIndent("4"), 1);
+        expect(SuperEditorInspector.findTaskIndent("5"), 2);
+        expect(SuperEditorInspector.findTaskIndent("6"), 0);
+      });
     });
   });
 }
```
angelosilvestre, line 752: How about a test to lock the behavior we want for the following scenario:

```
[ ] Level 1
    [ ] Level 2
        [ ] Level 3
```

If the user presses TAB on the Level 2 it shouldn't indent, right?

Also, it seems we are missing tests for the backspace shortcut.

super_editor/lib/src/default_editor/tasks.dart
matthew-carroll: Good point. No it shouldn't.

super_editor/lib/src/default_editor/tasks.dart
matthew-carroll: Do you think it makes sense to force text style knowledge into the task indent calculator? We can use the same one as list items, but that will require involving the text style.

super_editor/test/super_editor/components/task_test.dart
matthew-carroll: I guess I should change the verbiage. In this case child mean "following". I'll update it.

super_editor/test/super_editor/components/task_test.dart
matthew-carroll: Done

matthew-carroll:
I tried using the same indent calculator for tasks as list items, but it indented way too far. So I created a different one for list items vs tasks. The result is this.

Screenshot 2024-06-17 at 4 54 29 PM

super_editor/lib/src/default_editor/tasks.dart
```diff
@@ -185,6 +214,8 @@ class TaskComponentViewModel extends SingleColumnLayoutComponentViewModel with T
       super == other &&
           other is TaskComponentViewModel &&
           runtimeType == other.runtimeType &&
+          indent == other.indent &&
+          indentCalculator == other.indentCalculator &&
```
angelosilvestre, line 218: I think we shouldn't include the `indentCalculator` in equals and hashcode. 

Developers can pass anonymous methods to `indentCalculator`, this would cause any two`TaskComponentViewModel`s to be different.

I think we had the same problem with  `setComplete` in the past.

angelosilvestre (approved):
LGTM

PR #2110, opened 2024-06-19
Fix: Can now convert paragraph to task and begin with task completed (Resolves #2109)

### LINKED ISSUE #2109: [Bug][SuperEditor][SuperReader] - Tasks can't begin as completed
Due to a hard-coded value in the `ConvertParagraphToTaskCommand` we can't convert a paragraph to a task and immediately check it.

Replace the hard-coded value by passing the variable that's already made available in that command.

matthew-carroll:
Fix: Can now convert paragraph to task and begin with task completed (Resolves #2109)

The bug was simply that we weren't passing a variable forward - instead we hardcoded `false`.

I also cleaned up the corresponding test suite by using the pump scaffold method in all tests.

### DIFF
--- a/super_editor/test/super_editor/components/task_test.dart
+++ b/super_editor/test/super_editor/components/task_test.dart
@@ -19,24 +19,7 @@ void main() {
           TaskNode(id: "1", text: AttributedText("This is a task"), isComplete: false),
         ],
       );
-      final composer = MutableDocumentComposer();
-      final editor = createDefaultDocumentEditor(document: document, composer: composer);
-
-      await tester.pumpWidget(
-        MaterialApp(
-          home: Scaffold(
-            body: SuperEditor(
-              editor: editor,
-              document: document,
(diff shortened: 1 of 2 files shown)


super_editor/test/super_editor/components/task_test.dart
```diff
@@ -591,6 +411,42 @@ void main() {
...
+          nodes: [
+            ParagraphNode(id: "1", text: AttributedText("This is a task")),
+          ],
+        );
+        final editor = await _pumpScaffold(tester, document: document);
+
+        // Convert the paragraph to a task.
+        editor.execute([
+          const ConvertParagraphToTaskRequest(nodeId: "1", isComplete: true),
+        ]);
```
angelosilvestre, line 444: I think it should be "complete task". Both tests have the same description.

angelosilvestre (approved):
LGTM with a single comment.

PR #2122, opened 2024-06-22
[SuperEditor] Make selected text color strategy consider each colored span (Resolves #2093)

### LINKED ISSUE #2093: [BUG] - selectedTextColorStrategy's originalTextColor ignores ColorAttributions
**Package Version**

**User Info**
coneno GmbH, QuikFlow

**To Reproduce**
1. Select the colored text

**Minimal Reproduction Code**
<details>

  ```dart

void main() {
  runApp(const MainApp());
}

  const MainApp({super.key});

  @override
}

  static const _colorAttribution = ColorAttribution(Colors.red);
    ParagraphNode(
      id: Editor.createNodeId(),
      text: AttributedText(
          'Hello, world!',
          AttributedSpans(attributions: const [
            SpanMarker(attribution: _colorAttribution, offset: 0, markerType: SpanMarkerType.start),
            SpanMarker(attribution: _colorAttribution, offset: 4, markerType: SpanMarkerType.end),
          ])),
    ),
  ]);
  late final Editor _editor;

  @override
  void initState() {
    super.initState();
  }

  @override
    return MaterialApp(
      home: Scaffold(
        body: Center(
          child: SuperEditor(
            editor: _editor,
            document: _document,
            composer: _composer,
              selectedTextColorStrategy: ({required originalTextColor, required selectionHighlightColor}) =>
                  originalTextColor,
            ),
          ),
        ),
      ),
    );
  }
}

  ```
</details>

**Actual behavior**
The `selectedTextColorStrategy`, which just returns the `originalTextColor`, returns the text color as if there were no `ColorAttribution`s.

**Expected behavior**
The `originalTextColor` is provided with the ColorAttributions applied so that it reflects the actually visible color to the user and can be correctly modified to stay visible with regard to the `selectionHighlightColor`. Alternatively, the Attributions should be made available in this function as a parameter, so that it can itself implement the desired behavior.

**Platform**
macOS

**Flutter version**
Flutter 3.22.2 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 761747bfc5 (7 days ago) • 2024-06-05 22:15:13 +0200
Engine • revision edd8546116
Tools • Dart 3.4.3 • DevTools 2.34.3

**Screenshots**
Screenshot 2024-06-12 at 14 52 14

angelosilvestre:
[SuperEditor] Make selected text color strategy consider each colored span. Resolves #2093

`SingleColumnLayoutComponentViewModel` invokes the `SelectedTextColorStrategy` using only the default text color for the node as the `originalColor` argument, ignoring any `ColorAttribution`s. If we return the same original color, we loose the color attribution:

SCR-20240622-oxai

SCR-20240622-oxck

This PR changes `SingleColumnLayoutComponentViewModel` to consider each colored span and call `SelectedTextColorStrategy` for each of then:

SCR-20240622-oxlx

### DIFF
--- a/super_editor/lib/src/default_editor/layout_single_column/_styler_user_selection.dart
+++ b/super_editor/lib/src/default_editor/layout_single_column/_styler_user_selection.dart
@@ -132,25 +132,52 @@ class SingleColumnLayoutSelectionStyler extends SingleColumnLayoutStylePhase {
       editorStyleLog.finer('   - extent: ${textSelection?.extent}');
 
       if (viewModel is TextComponentViewModel) {
-        final componentTextColor = viewModel.textStyleBuilder({}).color;
+        AttributedText? textWithSelectionAttributions;
 
-        final textWithSelectionAttributions =
-            textSelection != null && _selectedTextColorStrategy != null && componentTextColor != null
-                ? (viewModel.text.copyText(0)
-                  ..addAttribution(
-                    ColorAttribution(_selectedTextColorStrategy!(
-                      originalTextColor: componentTextColor,
-                      selectionHighlightColor: _selectionStyles.selectionColor,
-                    )),
-                    SpanRange(textSelection.start, textSelection.end - 1),
-                    // The selected range might already have a color attribution. We want to override it
-                    // with the selected text color.
-                    overwriteConflictingSpans: true,
-                  ))
-                : viewModel.text;
+        if (textSelection != null && _selectedTextColorStrategy != null) {
+          final selectedRange = SpanRange(textSelection.start, textSelection.end - 1);
+
+          final componentTextColor = viewModel.textStyleBuilder({}).color;
+          if (componentTextColor != null) {
+            // Compute the selected text color for the default color of the node. If there is any
+            // text color attributions in the selected range, they will override this color.
+            textWithSelectionAttributions = viewModel.text.copyText(0)
+              ..addAttribution(
+                ColorAttribution(_selectedTextColorStrategy!(
+                  originalTextColor: componentTextColor,
+                  selectionHighlightColor: _selectionStyles.selectionColor,
+                )),
+                selectedRange,
+                // Override any existing color attributions.
+                overwriteConflictingSpans: true,
+              );
+          }
+
+          final coloredSpans = viewModel.text.getAttributionSpansInRange(
+            attributionFilter: (attr) => attr is ColorAttribution,
+            range: selectedRange,
+            // We should only change the selected portion of each span.
+            resizeSpansToFitInRange: true,
+          );
+          if (coloredSpans.isNotEmpty) {
+            // Compute and apply the selected text color for each span that has a color attribution.
+            textWithSelectionAttributions ??= viewModel.text.copyText(0);
+            for (final span in coloredSpans) {
+              textWithSelectionAttributions.addAttribution(
+                ColorAttribution(_selectedTextColorStrategy!(
+                  originalTextColor: (span.attribution as ColorAttribution).color,
+                  selectionHighlightColor: _selectionStyles.selectionColor,
+                )),
+                SpanRange(span.start, span.end),
+                // Override any existing color attributions.
+                overwriteConflictingSpans: true,
+              );
+            }
+          }
+        }
 
         viewModel
-          ..text = textWithSelectionAttributions
+          ..text = textWithSelectionAttributions ?? viewModel.text
           ..selection = textSelection
           ..selectionColor = _selectionStyles.selectionColor
           ..highlightWhenEmpty = highlightWhenEmpty;
--- a/super_editor/test/super_editor/supereditor_selection_test.dart
+++ b/super_editor/test/super_editor/supereditor_selection_test.dart
@@ -88,6 +88,79 @@ void main() {
         expect(richText.getSpanForPosition(const TextPosition(offset: 5))!.style!.color, Colors.green);
         expect(richText.getSpanForPosition(const TextPosition(offset: 16))!.style!.color, Colors.green);
       });
+
+      testWidgetsOnArbitraryDesktop("overrides each colored span", (tester) async {
+        final stylesheet = defaultStylesheet.copyWith(
+          selectedTextColorStrategy: ({required Color originalTextColor, required Color selectionHighlightColor}) {
+            if (originalTextColor == Colors.green) {
+              return Colors.red;
+            }
+
+            if (originalTextColor == Colors.yellow) {
+              return Colors.blue;
+            }
+
+            return Colors.white;
+          },
+        );
+
+        // Pump an editor with a paragraph with the following colors:
+        // Lorem ipsum dolor
+        // gggggg-----------
+        // ------yyyyyy-----
+        // ------------bbbbb (black, the default color)
+        await tester //
+            .createDocument()
+            .withCustomContent(
+              MutableDocument(
+                nodes: [
+                  ParagraphNode(
+                    id: '1',
+                    text: AttributedText(
+                      'Lorem ipsum dolor',
+                      AttributedSpans(
+                        attributions: [
+                          SpanMarker(
+                              attribution: const ColorAttribution(Colors.green),
+                              offset: 0,
+                              markerType: SpanMarkerType.start),
+                          SpanMarker(
+                              attribution: const ColorAttribution(Colors.green),
+                              offset: 5,
+                              markerType: SpanMarkerType.end),
+                          SpanMarker(
+                              attribution: const ColorAttribution(Colors.yellow),
+                              offset: 6,
+                              markerType: SpanMarkerType.start),
+                          SpanMarker(
+                              attribution: const ColorAttribution(Colors.yellow),
+                              offset: 11,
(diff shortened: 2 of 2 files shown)


super_editor/test/super_editor/supereditor_selection_test.dart
```diff
@@ -88,6 +88,79 @@ void main() {
         expect(richText.getSpanForPosition(const TextPosition(offset: 5))!.style!.color, Colors.green);
         expect(richText.getSpanForPosition(const TextPosition(offset: 16))!.style!.color, Colors.green);
       });
+
+      testWidgetsOnArbitraryDesktop("overrides each colored span", (tester) async {
```
matthew-carroll, line 92: When I read the name of this test, I don't know what "overrides each colored span" means. I don't know the UX impetus for this test, nor what the UI result is expected to be.

Moreover, when I look down at the bottom of the test to see what the expectations are looking for, the comment above them simply says that all the colors have changed. Whatever this test is verifying, it's currently too difficult to figure out what that is.

super_editor/lib/src/default_editor/layout_single_column/_styler_user_selection.dart
```diff
@@ -132,25 +132,52 @@ class SingleColumnLayoutSelectionStyler extends SingleColumnLayoutStylePhase {
...
-                  ))
-                : viewModel.text;
+        AttributedText? textWithSelectionAttributions;
+
+        if (textSelection != null && _selectedTextColorStrategy != null) {
+          final selectedRange = SpanRange(textSelection.start, textSelection.end - 1);
+
+          final componentTextColor = viewModel.textStyleBuilder({}).color;
+          if (componentTextColor != null) {
+            // Compute the selected text color for the default color of the node. If there is any
```
matthew-carroll, line 142: To help avoid ambiguity with the concept of "selection color", any place where you're currently saying "selected text color", you should probably say "color of the selected text".

Also, I think this comment is overall difficult to follow. I think the following is what you're doing:

```
// Copy the existing text, switching the color of the text from the regular text color to the
// color of selected text. For example, text that's normally black, might become white when
// selected with a dark selection box color.
//
// Any text with an explicit color attribution will remain unaffected, e.g., a red word will
// remain red.
```

super_editor/lib/src/default_editor/layout_single_column/_styler_user_selection.dart
```diff
@@ -132,25 +132,52 @@ class SingleColumnLayoutSelectionStyler extends SingleColumnLayoutStylePhase {
...
+            // text color attributions in the selected range, they will override this color.
+            textWithSelectionAttributions = viewModel.text.copyText(0)
+              ..addAttribution(
+                ColorAttribution(_selectedTextColorStrategy!(
```
matthew-carroll, line 146: I'm not sure the overall assumptions of this change are correct. We're assuming that any text with a color applied to it should retain that same color when selected.

That might be what some people want, but it's probably not true in general, for the same reason that we're letting people change the color of selected text in the first place. Colors that are high contrast when unselected might be low contrast when selected. This applies to spans of text with color attributions, too.

Shouldn't the selected text color strategy get an opportunity to switch any given color, as desired?

KevinBrendel: Indeed, as a user, I would like to know the original color (after the attributions were applied), but I want the color returned by selectedTextColorStrategy to be displayed without further modifications. I just need to know about the original color so that I can calculate the correct color to be displayed.

super_editor/lib/src/default_editor/layout_single_column/_styler_user_selection.dart
angelosilvestre: > // Any text with an explicit color attribution will remain unaffected, e.g., a red word will
            // remain red.

This isn't the case. For each span with a different color, we call  `_selectedTextColorStrategy` with that color as the `originalTextColor`. If the method returns a different color, the returned color will override the existing color.

super_editor/lib/src/default_editor/layout_single_column/_styler_user_selection.dart
angelosilvestre: > I'm not sure the overall assumptions of this change are correct. We're assuming that any text with a color applied to it should retain that same color when selected.

As explained in this [comment](https://github.com/superlistapp/super_editor/pull/2122#discussion_r1664891728), the selected color strategy has to opportunity to switch any given color. The difference now is that, if the selected range has multiple different colored spans, the selected color strategy can return a new color for each existing color.

super_editor/test/super_editor/supereditor_selection_test.dart
angelosilvestre: Would "allows overriding each color within the selected text" be more clear?

super_editor/lib/src/default_editor/layout_single_column/_styler_user_selection.dart
matthew-carroll: @angelosilvestre can you add a test for that so I can see it in action?

angelosilvestre: @matthew-carroll The test referenced in this comment https://github.com/superlistapp/super_editor/pull/2122/files#r1660478509 is an example

super_editor/test/super_editor/supereditor_selection_test.dart
matthew-carroll: how about "can choose new selected text color based on the original text color"?

super_editor/test/super_editor/supereditor_selection_test.dart
```diff
@@ -88,6 +88,79 @@ void main() {
...
+        final stylesheet = defaultStylesheet.copyWith(
+          selectedTextColorStrategy: ({required Color originalTextColor, required Color selectionHighlightColor}) {
```
matthew-carroll, line 94: If we consider that editors will be used in a light mode and a dark mode, it seems likely that users would benefit from a strategy that works well in dark mode, to go with our existing strategy for light mode. A dark mode strategy should probably use a heuristic that inverts the luminance of each existing text color. Can you see if you can put together a reasonable dark mode strategy that automatically selects new colors and then we could test that actual strategy here?

angelosilvestre: I tried a naive solution inverting the lightness of an hsl color with the following code, but it doesn't look great:

```dart
Color darkModeSelectedColorStrategy({
  required Color originalTextColor,
  required Color selectionHighlightColor,
}) {
  final hslColor = HSLColor.fromColor(originalTextColor);
  return hslColor.withLightness(1.0 - hslColor.lightness).toColor();    
}
```

Not selected:

SCR-20240709-tcza

With a strategy that returns the original color:

SCR-20240709-tdil

Color with an inverse lightness.

SCR-20240709-tdpn

What do you think? Do you have other ideas ? Maybe we should take the `selectionHighlightColor` also in consideration.

matthew-carroll: > Maybe we should take the selectionHighlightColor also in consideration.

Yeah I think we probably should. The primary consideration is to ensure that the selected text contrasts with the selection. Do you know of any other app references to look at for this, where selection colors are different from typical browser selections?

KevinBrendel: Word (at least on Mac) seems to change the text color slightly if the selection color is the exact same as the text color.

matthew-carroll:
@angelosilvestre do you need anything from me to continue this PR?

super_editor/test/super_editor/supereditor_selection_test.dart
angelosilvestre: I tried some apps:

Notion doesn't seem to change the font color when the text is selected:

image

image

Apple notes changes all text to white:

image

image

Google docs don't change the font color:

image

image

Word online also doesn't seem to change the font color:

image

image

angelosilvestre:
@matthew-carroll Updated the example app to use the same behavior as Apple Notes to color the text in dark mode.

Also, update a test name.

super_editor/example/lib/demos/example_editor/example_editor.dart
```diff
@@ -255,6 +255,16 @@ class _ExampleEditorState extends State<ExampleEditor> {
     _editorFocusNode.requestFocus();
   }
 
+  /// Makes text white, for use during dark mode styling.
+  ///
+  /// This is the same behavior observed in Apple Notes.
+  Color _darkModeSelectedColorStrategy({
```
matthew-carroll, line 261: Did we say that we wanted Apple Notes style highlighting as the default in Super Editor? If so, why is this in the example app? Shouldn't we have a public default object in Super Editor that does this in dark mode, and does the other thing in light mode?

angelosilvestre: Are you saying that we should include this in the default stylesheet? If so, we would need to include the `BuildContext` in the `SelectedTextColorStrategy` signature, otherwise we won't know if we are in light or dark mode.

PR #2125, opened 2024-06-24
[SuperTextField] New variations of AttributedTextEditingController.clear() (Resolves #2094)

### LINKED ISSUE #2094: [BUG] - AttributedTextEditingController.clear() not working as expected
**Package Version**
`SuperTextField` from `super_editor` on latest `main`

**User Info**
Superlist

**To Reproduce**
1. Type some text into a `SuperTextField`
2. Press a button which is hooked up to call `AttributedTextEditingController.clear()`
3. The text is not cleared, and no exception is printed to the console.
4. If you then continue to type text, an exception is printed to the console.

Note: I have found it possible to work around this by instead of calling `.clear()`:

```dart
    _textController.updateTextAndSelection(
      text: AttributedText(),
      selection: const TextSelection.collapsed(offset: 0),
    );
```

The above works fine visually:
* focus still present
* caret still blinking
* new typed text ends up in the text field

... however an exception is printed to the console.

**Minimal Reproduction Code**

<details>

  ```dart
import 'package:flutter/scheduler.dart';

void main() {
  runApp(const MyApp());
}

  const MyApp({super.key});

  @override
    return MaterialApp(
      title: 'Clear Text Field Example',
      theme: ThemeData(
        colorScheme: ColorScheme.fromSeed(seedColor: Colors.deepPurple),
        useMaterial3: true,
      ),
      home: const MyHomePage(title: 'Clear Text Field Example'),
    );
  }
}

class MyHomePage extends StatefulWidget {
  const MyHomePage({
    super.key,
    required this.title,
  });

  final String title;

  @override
  State<MyHomePage> createState() => _MyHomePageState();
}

class _MyHomePageState extends State<MyHomePage> {
  final _focusNode = FocusNode();
  final _textController = ImeAttributedTextEditingController();

  @override
  void initState() {
    super.initState();

    SchedulerBinding.instance.addPostFrameCallback((_) {
      // Why do I have to request focus in a post frame callback?
      //
      // I guess it's something about the focus node not yet being parented?
      // Other Flutter widgets don't seem to have this issue.
      //
      // Oddly enough if I don't use the post frame callback, the text field
      // takes up all available vertical space, and results in a layout
      // exception.
      _focusNode.requestFocus();
    });
  }

  @override
  void dispose() {
    _focusNode.dispose();
    _textController.dispose();
    super.dispose();
  }

  void _clearTextOption1() {
    // Text is not cleared.
    // No error printed to console.
    // Subsequently typed text does not show in text field, and an exception 
    // is printed to the console.
    _textController.clear();
  }

  void _clearTextOption2() {
    // This works perfectly visually:
    // * focus still present
    // * caret still blinking
    // * new typed text ends up in the text field
    //
    // ... but an exception is printed to the console
    _textController.updateTextAndSelection(
      text: AttributedText(),
      selection: const TextSelection.collapsed(offset: 0),
    );
  }

  @override
    final theme = Theme.of(context);
    final textStyle = theme.textTheme.bodyLarge!.copyWith(
      color: theme.colorScheme.primary,
    );
    final hintStyle = textStyle.copyWith(
      color: theme.colorScheme.secondary,
    );

    return Scaffold(
      appBar: AppBar(
        backgroundColor: Theme.of(context).colorScheme.inversePrimary,
        title: Text(widget.title),
      ),
      body: Align(
        alignment: Alignment.topCenter,
        child: Padding(
          padding: const EdgeInsets.all(32),
          child: Container(
            decoration: BoxDecoration(
              border: Border.all(
                width: 1,
                color: Theme.of(context).colorScheme.primary,
              ),
              borderRadius: BorderRadius.circular(8),
              color: Theme.of(context).colorScheme.surface,
            ),
            child: Row(
              mainAxisSize: MainAxisSize.max,
              children: [
                const SizedBox(width: 16),
                Expanded(
                  child: SuperTextField(
                    focusNode: _focusNode,
                    textController: _textController,
                    textStyleBuilder: (_) => textStyle,
                    hintBuilder: (_) => Text(
                      'Type something',
                      style: hintStyle,
                    ),
                  ),
                ),
                const SizedBox(width: 16),
                IconButton(
                  icon: const Icon(Icons.clear),
                  onPressed: _clearTextOption1,
                ),
              ],
            ),
          ),
        ),
      ),
    );
  }
}
  ```
</details>

**Actual behavior**
When calling `_clearTextOption1`:
- Text is not cleared.
- No error printed to console.
- Subsequently typed text does not show in text field, and an exception is printed to the console.

When calling `_clearTextOption2`:
- This works perfectly visually:
  * Focus is still present.
  * The caret still blinking.
  * Newly typed text ends up in the text field.
- An unexpected error is printed to the console.

**Expected behavior**
When calling `_clearTextOption1`:
- Text should be cleared.
- No error should be printed.
- Focus should still be present.
- The caret should still be blinking.
- Subsequently typed text should end up in the text field, without causing any errors.

When calling `_clearTextOption2`:
- Should have exact results as option 1, without any errors.
- Should not have to use because option 1 doesn't work.

**Platform**
Which device(s) and operating system(s) have experienced the problem?
All of above notes are about iOS.

Just checked macOS and it also suffers from the issues with `.clear()`, but the workaround does not seem to cause any errors to be printed to console.

Did not test on Android or web.

**Flutter version**
```
Flutter 3.22.0-22.0.pre.7 • channel main • https://github.com/flutter/flutter.git
Framework • revision 9f3be7dfd9 (6 weeks ago) • 2024-05-02 04:15:02 -0400
Engine • revision 58b031c096
Tools • Dart 3.5.0 (build 3.5.0-122.0.dev) • DevTools 2.35.0
```

**Screenshots**
Option 1:

https://github.com/superlistapp/super_editor/assets/5852503/b42100a1-cac1-4ce6-a14e-eceda85fba79

Option 2:

https://github.com/superlistapp/super_editor/assets/5852503/ecec124f-cfa9-4281-9240-6ee080ac8bfe

CillianMyles:
(Resolves #2094)

### DIFF
--- a/super_editor/test/super_textfield/attributed_text_editing_controller_test.dart
+++ b/super_editor/test/super_textfield/attributed_text_editing_controller_test.dart
@@ -835,8 +835,8 @@ void main() {
 
       test(
           'calling the clearTextAndSelection method clears the text, '
-          'composing attributions, composing region, and moves '
-          'the removes the text selection', () {
+          'composing attributions, composing region, and removes the '
+          'text selection', () {
         int listenerNotifyCount = 0;
         final controller = AttributedTextEditingController(
           text: AttributedText('my text'),
@@ -864,8 +864,8 @@ void main() {
 
       test(
           'calling the clear method clears the text, '
-          'composing attributions, composing region, and moves '
-          'the selection to the start', () {
+          'composing attributions, composing region, and removes the '
+          'text selection', () {
         int listenerNotifyCount = 0;
         final controller = AttributedTextEditingController(
           text: AttributedText('my text'),
@@ -879,6 +879,7 @@ void main() {
             listenerNotifyCount += 1;
           });
 
+        // ignore: deprecated_member_use_from_same_package
         controller.clear();
 
         expect(controller.text.text, isEmpty);
--- a/super_editor/test/super_textfield/attributed_text_editing_controller_test.dart
+++ b/super_editor/test/super_textfield/attributed_text_editing_controller_test.dart
@@ -804,39 +804,7 @@ void main() {
     });
 
     group("clearing text and selection", () {
-      test(
-          'calling the clearText method clears the text, '
-          'composing attributions, composing region, and moves '
-          'the selection to the start', () {
-        int listenerNotifyCount = 0;
-        final controller = AttributedTextEditingController(
-          text: AttributedText('my text'),
-          selection: TextSelection.collapsed(offset: 7),
-          composingRegion: TextRange(start: 3, end: 7),
-        )
-          ..composingAttributions = {
-            boldAttribution,
-          }
-          ..addListener(() {
-            listenerNotifyCount += 1;
-          });
-
-        controller.clearText();
-
-        expect(controller.text.text, isEmpty);
-        expect(
-          controller.selection,
-          const TextSelection.collapsed(offset: 0),
-        );
-        expect(controller.composingAttributions, isEmpty);
-        expect(controller.composingRegion, TextRange.empty);
-        expect(listenerNotifyCount, 1);
-      });
-
-      test(
-          'calling the clearTextAndSelection method clears the text, '
-          'composing attributions, composing region, and removes the '
-          'text selection', () {
+      test("can remove the text, selection, and composing region at the same time", () {
         int listenerNotifyCount = 0;
         final controller = AttributedTextEditingController(
           text: AttributedText('my text'),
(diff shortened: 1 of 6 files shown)


super_editor/test/super_textfield/attributed_text_editing_controller_test.dart
```diff
@@ -803,7 +803,192 @@ void main() {
       });
     });
 
+    group("clearText", () {
+      test('clears the text', () {
```
matthew-carroll, line 807: It looks like there are separate tests from simultaneous expected conditions. There probalby isn't any reason to separate them. It looks like you should probably verify all conditions in the same test. You'll just want to make sure the test name captures the goal.

CillianMyles: Combined all simultaneous expectations into single tests ✅

super_editor/test/super_textfield/attributed_text_editing_controller_test.dart
```diff
@@ -803,7 +803,192 @@ void main() {
       });
     });
 
+    group("clearText", () {
```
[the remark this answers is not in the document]
CillianMyles: Named the group and tests as best as I could with this direction. I ended up having to use `"calling the foo method ..."` terminology because there is no way that these events occur on their own, calling the method will be something devs will do in their own use cases. Let me know if this works!

super_editor/test/super_textfield/attributed_text_editing_controller_test.dart
```diff
@@ -803,18 +803,92 @@ void main() {
...
+        expect(
+          controller.selection,
+          const TextSelection.collapsed(offset: 0),
+        );
+        expect(controller.composingAttributions, isEmpty);
+        expect(controller.composingRegion, TextRange.empty);
+        expect(listenerNotifyCount, 1);
+      });
 
+      test(
```
matthew-carroll, line 836: I can't see the difference between this test name and the following test name. Can you check those? That's one of the costs of lengthy and similar test names.

I think the following two test names would cover your bases:

```
test("can remove the text, selection, and composing region at the same time", () {...});

test("can remove the text and composing region, and place the caret at the start, at the same time", () {...});
```

Also, let's keep the test names to a single line. Updating test names with multiple lines of quoted text is a nightmare.

CillianMyles: Thanks for the suggestions. I combined everything into 2 tests, since there are only 2 logical options/outcomes for an API consumer. It did leave me wondering where to place the test for the deprecated .clear() method. I put it inside the bottom of the 1st test block (alongside the test for the method it is replacing), and mentioned it can be removed when .clear() is removed. Let me know what you think

matthew-carroll (approved):
LGTM

matthew-carroll:
@angelosilvestre can you cherry pick this when you get a chance?

PR #2127, opened 2024-06-26
Ignore spurious metrics change notification

knopp:
On iOS with keyboard present there is occasional `didChangeMetrics` notification even if the metrics did not change. This is something that would be nice to fix in the engine, but the API itself does not make any guarantees about when the notification is called.

The problem with this is that the notification might come during scrolling, in which case it can interfere with scroll position when trying to position the cursor within the viewport.


matthew-carroll (approved):
LGTM

matthew-carroll:
@angelosilvestre can you cherry pick this to stable when you get a minute?

PR #2128, opened 2024-06-27
Clamp the scroll offset in ensureOffsetIsVisible

knopp:
The offset does seem to be out of range with `SliverPinnedHeader`, which causes quite a lot of jumpiness when typing. I'm not sure if there is better way to fix this, but clamping works.

### DIFF
--- a/super_editor/lib/src/infrastructure/platforms/mobile_documents.dart
+++ b/super_editor/lib/src/infrastructure/platforms/mobile_documents.dart
@@ -409,6 +409,11 @@ class DragHandleAutoScroller {
     final scrollPosition = _getScrollPosition();
     final currentScrollOffset = scrollPosition.pixels;
 
+    // The offset calculation below does not work correctly in custom scroll view with sliver header
+    // and causes overscroll so for now clamp the offset.
+    final max = scrollPosition.maxScrollExtent;
+    final min = scrollPosition.minScrollExtent;
+
     if (offsetInViewport.dy < _dragAutoScrollBoundary.leading) {
       // The offset is above the leading boundary. We need to scroll up
       editorGesturesLog.fine("The scrollable needs to scroll up to make offset visible.");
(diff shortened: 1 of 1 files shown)


matthew-carroll:
@knopp did you also try this change outside of a custom scrollview? Did you do enough regression analysis to ensure that this doesn't create meaningful new problems for existing uses?

knopp:
Not quite sure how exactly this could regress given that it only ensures that we don't jump outside of valid scroll positions? Can you maybe be more specific? It's a fairly trivial fix.

matthew-carroll:
@knopp I don't have any specific interactions in mind, but I've seen enough small, simple changes cause unexpected results. Given that scrolling is a critical requirement across all editor use-cases, I'd appreciate it if you could check the experiences that don't use a `CustomScrollView` and ensure that there aren't any obvious issues with scrolling, including auto-scrolling when editing, etc.

Ideally we would have a battery of tests for these things, but testing against scroll offsets is really tedious with a high maintenance cost.

BazinC:
I also fixed the same problem in a fork 2 weeks ago. It is reproductible on super_editor example, see repro steps in description: https://github.com/BazinC/super_editor/pull/9

matthew-carroll:
@BazinC - Did you try to the change without a surrounding `CustomScrollView`?

BazinC:
> @BazinC - Did you try to the change without a surrounding `CustomScrollView`?

Yes my change also works without a surrounding CustomScrollView. See video video in description https://github.com/BazinC/super_editor/pull/9, it's using an ExampleEditor. No CustomScrollView.

matthew-carroll:
Since @BazinC verified this behavior when not using a `CustomScrollView`, I'll go ahead and merge it.

matthew-carroll (approved):
LGTM

matthew-carroll:
@angelosilvestre can you cherry pick this when you get a minute?

PR #2131, opened 2024-06-30
[SuperTextLayout] Paint text selection with BorderRadius

domesticmouse:


domesticmouse:
@matthew-carroll I'm curious what tests you would like for this?

domesticmouse:
I think the goldens i generated on macOS aren't compatible with CI. As a googler, I can't use docker...

matthew-carroll:
@domesticmouse you can't use Docker on your personal machine? Or you can't use Docker on a G3 machine and that's all you've got?

super_text_layout/lib/src/text_selection_layer.dart
```diff
@@ -132,12 +133,14 @@ class TextSelectionPainter extends CustomPainter {
   TextSelectionPainter({
     required this.textLayout,
     required this.textSelection,
+    required this.borderRadius,
```
matthew-carroll, line 136: This is going to be a breaking change, as-is. Please switch this to a default value of zero so that nothing upstream breaks by adding this.

domesticmouse:
I don't currently have a personal dev machine. Can you update the goldens in this PR to make the CI happy?

super_text_layout/test_goldens/text_selection_layer_test.dart
```diff
@@ -77,4 +77,76 @@ void main() {
...
+        tester,
+        beneathBuilder: (context, textLayout) {
+          return TextLayoutEmptyHighlight(
+            textLayout: textLayout,
+            style: selectionStyle,
+          );
+        },
+      );
+
+      await screenMatchesGolden(tester, "TextSelectionLayer_small-highlight-when-empty-border-radius");
```
matthew-carroll, line 134: Is this golden image correct for this test? I don't see rounded corners on the empty highlight.

super_text_layout/lib/src/text_selection_layer.dart
domesticmouse: Done.

super_text_layout/test_goldens/text_selection_layer_test.dart
domesticmouse: The empty case was a separate codepath. Found and updated.

matthew-carroll:
@angelosilvestre can you try generating the goldens for this PR?

super_text_layout/lib/src/text_selection_layer.dart
```diff
@@ -156,11 +165,18 @@ class TextSelectionPainter extends CustomPainter {
...
+          bottomRight: borderRadius.bottomRight);
 
-      canvas.drawRect(
+      canvas.drawRRect(
         // Note: If the rect has no width then we've selected an empty line. Give
         //       that line a slight width for visibility.
-        rect.width > 0 ? rect : Rect.fromLTWH(rect.left, rect.top, 5, rect.height),
+        rect.width > 0
+            ? rrect
+            : RRect.fromRectAndRadius(Rect.fromLTWH(rect.left, rect.top, 5, rect.height), Radius.zero),
```
domesticmouse, line 179: @matthew-carroll I'm unsure what should be done here. Should we round these corners as well?

matthew-carroll: Can you provide some context? What's the code we're looking at, what does it do now, and what are you proposing that it should do?

domesticmouse: The code I'm looking at is the edge case for when the `rect` has zero width. The code currently just creates a rectangle of width 5. I'm wondering if this edge case should round the edge case created rectangle the same as the other rects. There currently isn't a golden test for this as changing it from not rounding the edge case rect to rounding it doesn't break the test suite.

angelosilvestre:
@domesticmouse I cannot push to this branch, so I created a PR to add the updated goldens. Could you please download the goldens and update this PR?

domesticmouse:
@angelosilvestre I believe I have pulled the content from https://github.com/superlistapp/super_editor/pull/2136 into this PR.

super_text_layout/lib/src/text_selection_layer.dart
matthew-carroll: > The code I'm looking at is the edge case for when the rect has zero width

Is this the same thing as the optional policy that we have to display a selection when empty? If so, I thought I saw that you already rounded the empty selection rectangle?

If this is something different, can you let me know where/when this 5px wide rectangle currently appears?

domesticmouse:
> @domesticmouse I cannot push to this branch, so I created a PR to add the updated goldens. Could you please download the goldens and update this PR?

Reading the code, this path is only triggered on empty lines, when this code expands the selection box from zero to five pixels wide.

domesticmouse:
Please land. I'd like to use this for my presentation in India in a couple of weeks.

matthew-carroll:
I haven't had a chance to look into that empty selection code path, but I'll merge as-is so you can use it.

matthew-carroll (approved):
LGTM

matthew-carroll:
@angelosilvestre can you cherry pick this, and also close your other PR if it's no longer needed?

Issue #1795, opened 2024-01-13
[SuperTextField][mobile] Multi-line textfield doesn't allow drag to scroll vertically


angelosilvestre:
**Package Version**
super_editor - main

**To Reproduce**
1. Open Example app with iOS or Android SuperTextField example
2. Run the short demo
3. Drag to scroll vertically

**Actual behavior**
The content doesn't scroll

**Expected behavior**
The content should scroll, like Flutter `TextField`

**Platform**
iOS and Android

**Flutter version**
3.17.0-19.0.pre.16 • channel master •

**Screenshots**

Flutter `TextField`:

[scroll_android.webm](https://github.com/superlistapp/super_editor/assets/7597082/4a144478-cd03-4955-b8b4-3dae110b2faf)

`SuperTextField`:

[supertextfield_scroll.webm](https://github.com/superlistapp/super_editor/assets/7597082/55ad2cc8-2a0d-43eb-b846-56194987c7fe)

angelosilvestre:
Cc @matthew-carroll

matthew-carroll:
I think this is now being worked on here: https://github.com/superlistapp/super_editor/pull/2139

PR #2139, opened 2024-07-03
[FEATURE] - Allow vertical scrolling of a multi-line SuperTextField when its text overflows (Resolves #2140 and #1795)

### LINKED ISSUE #2140: [FEATURE] - Allow vertical scrolling of a multi-line SuperTextField when its text overflows
**Your Info**
Using SuperEditor at Superlist

**User Behavior**
Multi-line SuperTextFields with text that overflow (e.g. 3 lines of text, when max lines is 2) are not vertically scrollable on mobile. I can confirm this is not an issue on desktop (at least I tested on macOS).

**Feature Value**
It would be great if we could vertically scroll up and down within the text field to access the overflowed text content.

**Additional context**
To work around this, you can initiate the magnifier and then vertically scroll, but this seems like it should be possible without this workaround.

For context, here is the desired behaviour visible in iMessage on iOS:

https://github.com/superlistapp/super_editor/assets/5852503/67bef542-c426-43e1-92ad-f15778620879

Here is a demo showing iOS scrolling is not possible, but macOS is:

https://github.com/superlistapp/super_editor/assets/5852503/561d19be-b230-4961-81d0-c19cc41fbb1b

CillianMyles:


super_editor/lib/src/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -423,7 +423,7 @@ class _TextScrollViewState extends State<TextScrollView>
       child: SingleChildScrollView(
         key: _textFieldViewportKey,
         controller: _scrollController,
-        physics: const NeverScrollableScrollPhysics(),
+        physics: isMultiline ? null : const NeverScrollableScrollPhysics(),
```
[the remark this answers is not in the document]
matthew-carroll: > Also as far as I could text this change only applies to mobile SuperTextFields.

Can you mention what that's based on? Without opening the source code I can't tell. Is there a mobile platform check somewhere in this code, or a related place?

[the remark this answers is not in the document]
CillianMyles: This code is within this widget called `TextScrollView`, which is only used in `SuperAndroidTextField` and `SuperIOSTextField`:

CleanShot 2024-07-09 at 21 37 51@2x

I said as far "as I could tell" because I'm not intimately familiar with the codebase, and did not want to say something incorrect or misleading.

super_editor/test/super_textfield/super_textfield_gesture_scrolling_test.dart
```diff
@@ -249,6 +249,80 @@ void main() {
...
+
+      // Ensure the text field has scrolled to the bottom.
+      expect(
+        SuperTextFieldInspector.findScrollOffset(),
+        greaterThan(0.0),
+      );
+
+      // Scroll upwards towards the top of the text field.
+      await tester.drag(find.byType(SuperTextField), const Offset(0, -1000));
+      await tester.pumpAndSettle();
```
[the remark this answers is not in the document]
CillianMyles: Admittedly I did not spend a lot of time trying to do so, wanted to make sure I was in the right ballpark first. Happy to fix these tests, or add more in other files as required.

matthew-carroll:
I checked usages of `TextScrollView` and you're correct that it's only currently used for Android and iOS `SuperTextField`s.

Did you check what happens with auto-scrolling when dragging selection handles, dragging the caret, etc?

Wrapped around the `TextScrollView` is a `AndroidTextFieldTouchInteractor` and a `IOSTextFieldTouchInteractor`, respectively. Both of those interactors are given access to the scroll controller for the `TextScrollView`.

I assume there was a reason that I disabled the standard scroll physics. I'd like to find what that was and then ensure we didn't break those requirements with this change. If this does break those requirements, then we might need to find another way to accomplish this.

matthew-carroll:
Also, let's move this out of draft mode, and can you also check the failing tests in CI?

CillianMyles:
@matthew-carroll - the tests I added for mobile and desktop are now passing. These tests check to see that multi-line SuperText fields can scroll up and down when the text overflows.

In terms of checking the existing functionality with selection handle, magnifiers, scrolling etc.: all existing tests pass, which I hope cover the rules in more detail than I can manually test. In any case I did some manual checks - see:

https://github.com/user-attachments/assets/a051b198-ef1c-4e51-9ecc-61c553491d20

I now have some golden failures. I tried to update the goldens locally but it ended up in way too many unrelated files changed. Is there something I'm missing with regard to updating the goldens? I am running latest Flutter main:
```
Flutter 3.24.0-1.0.pre.153 • channel main • https://github.com/flutter/flutter.git
Framework • revision 49d6d520cb (45 minutes ago) • 2024-07-17 11:18:10 -0400
Engine • revision 7e25796340
Tools • Dart 3.6.0 (build 3.6.0-48.0.dev) • DevTools 2.37.1
```

super_editor/test_goldens/super_textfield/super_textfield_toolbar_test.dart
```diff
@@ -8,15 +8,14 @@ import '../test_tools_goldens.dart';
...
+    testGoldensOniOS('displays toolbar pointing down for expanded selection', (tester) async {
       // Pumps a widget tree with a SuperTextField at the bottom of the screen.
       await _pumpSuperTextfieldToolbarTestApp(
         tester,
         child: Positioned(
           bottom: 50,
           child: _buildSuperTextField(
             text: 'Arrow pointing down',
-            configuration: SuperTextFieldPlatformConfiguration.iOS,
           ),
```
CillianMyles, line 19: @matthew-carroll not sure why these widget tests were setup this way, but since there is iOS in the golden file path name, I figured iOS was the intention, and adjusting this platform setup and regenerating the goldens was the only way I could get the golden tests to pass.

Also I was only able to get the goldens to pass by generating them on a Linux/Ubuntu machine, macOS seemed to completely change the font rendering.

CillianMyles:
@matthew-carroll all tests (including new tests & goldens) are now passing.

Not sure what to do about the Netlify/deploymant job failures. I don't think anything I did should affect them?

CillianMyles:
@matthew-carroll I think we should fix the goldens: merge either (https://github.com/superlistapp/super_editor/pull/2183) or (https://github.com/superlistapp/super_editor/pull/2184), and then I will rebase this PR. 

I see that option 1 works and 2 does not, and also the Netlify stuff passes in option 1 as well so probably will succeed here after I rebase this PR.

Please let me know how we should proceed.

CillianMyles:
@matthew-carroll this one has been rebased and tests are passing, and ready for another look!

matthew-carroll:
@CillianMyles do you know if those demo site failures are new to this PR vs broken in general? I don't think I saw those failing in other PRs...

super_editor/lib/src/super_textfield/infrastructure/text_scrollview.dart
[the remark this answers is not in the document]
matthew-carroll: Can you add a code comment that explains why this ternary is here, and can you also add a breadcrumb in that comment for where a developer should look to find scrolling behavior in the case of a single-line text field (for which we disable the physics)?

CillianMyles:
@matthew-carroll over the past few days/weeks working on a few PRs, I have noticed these fail sometimes. I can't remember in which PRs and it's a bit hard to track down older runs due to rebasing. 

To be honest thought I can't think why the code changes in this PR could have any affect on these demo site failures.

Not sure if it's any use, but now come to think of it, something I do remember when I created the option 1 and option 2 PRs at virtually the same time, the PR checks were actually running at the same time on both, and I noted that the demo site jobs failed on one PR and not the other. Not sure if it could be anything to do with concurrent jobs.

Aside from that, I don't have any ideas unfortunately.

super_editor/test/super_textfield/super_textfield_gesture_scrolling_test.dart
```diff
@@ -249,6 +249,102 @@ void main() {
...
+      // of text, which should overflow considering there are 3 lines of text.
+      await _pumpTestApp(
+        tester,
+        textController: controller,
+        minLines: 1,
+        maxLines: 2,
+        maxHeight: 40,
+      );
+
+      double getTextBottom() {
```
matthew-carroll, line 271: I'd prefer that you duplicate the actual tester lines instead of defining inline functions. Inline functions require the reader to become a human compiler to understand the code and I'd like to eliminate that in any place where the value of the function isn't immense.

super_editor/test/super_textfield/super_textfield_gesture_scrolling_test.dart
```diff
@@ -249,6 +249,102 @@ void main() {
...
+        return tester.getBottomRight(find.byType(SuperTextField)).dy;
+      }
+
+      double getViewportBottom() {
+        return tester.getBottomRight(find.byType(SuperText)).dy;
+      }
+
+      // Ensure the text field has not yet scrolled.
+      expect(getTextBottom(), lessThan(getViewportBottom()));
```
matthew-carroll, line 280: Is this how we verify scroll offsets in other tests? I don't remember, but I would assume that we'd access a scroll controller value. Perhaps we even have some test tools for this?

super_editor/test/super_textfield/super_textfield_gesture_scrolling_test.dart
```diff
@@ -249,6 +249,102 @@ void main() {
...
+
+      // Scroll down to reveal the last line of text.
+      await tester.drag(find.byType(SuperTextField), const Offset(0, -1000.0));
+      await tester.pumpAndSettle();
+
+      // Ensure the text field has scrolled to the bottom.
+      expect(getTextBottom(), equals(getViewportBottom()));
```
matthew-carroll, line 287: An exact scroll offset expectation has a decent chance of failing due to floating point issues. Can you look for one of our tests that uses a `moreOrLessEquals()` comparison? I think that's the name...

super_editor/test/super_textfield/super_textfield_gesture_scrolling_test.dart
```diff
@@ -249,6 +249,102 @@ void main() {
...
+
+      // Scroll back up to the top of the text field.
+      await tester.drag(find.byType(SuperTextField), const Offset(0, 1000.0));
+      await tester.pumpAndSettle();
+
+      // Ensure the text field has scrolled back to the top.
+      expect(getTextBottom(), lessThan(getViewportBottom()));
```
matthew-carroll, line 294: Is `lessThan` a sufficient check? One if we only scrolled up 1px....wouldn't this still pass?

super_editor/test/super_textfield/super_textfield_gesture_scrolling_test.dart
CillianMyles: Done ✅

super_editor/test/super_textfield/super_textfield_gesture_scrolling_test.dart
CillianMyles: Done ✅

super_editor/test/super_textfield/super_textfield_gesture_scrolling_test.dart
CillianMyles: You are correct, when changing to scroll only 1 pixel, the test still passes.

I switched approaches to always validate that either the tops or bottoms are aligned!

super_editor/test/super_textfield/super_textfield_gesture_scrolling_test.dart
CillianMyles: In different tests in this file, we use both this method, and the `SuperTextFieldInspector.findScrollOffset()` method:
https://github.com/superlistapp/super_editor/blob/main/super_editor/test/super_textfield/super_textfield_inspector.dart#L157-L191

I tried the `SuperTextFieldInspector.findScrollOffset()` method, but found it did not seem to have the correct value after scrolling back up (0), even though I visually validated the scroll by taking some golden images.

During some debugging, I found the values of the scroll controller, and the values of `TextScrollController.scrollOffset` to be off, even though you would expect them to be the same.

When I saw the following [comment](https://github.com/superlistapp/super_editor/blob/main/super_editor/lib/src/super_textfield/infrastructure/text_scrollview.dart#L523-L537), it made me think that it would be easier to compare the top and bottom values, since this method is used already in this file, and as I did not really know how to debug this any further.

CillianMyles:
@matthew-carroll I rebased and addressed all you points. Looking forward to your feedback!

super_editor/test/super_textfield/super_textfield_gesture_scrolling_test.dart
matthew-carroll: Can you file an issue about the inspector method returning the wrong results, and include repro steps?

super_editor/lib/src/super_textfield/infrastructure/text_scrollview.dart
```diff
@@ -423,6 +423,8 @@ class _TextScrollViewState extends State<TextScrollView>
       child: SingleChildScrollView(
         key: _textFieldViewportKey,
         controller: _scrollController,
+        // Passing null for scroll physics for multiline text fields will
+        // apply an appropriate physics for the underlying host platform.
```
matthew-carroll, line 427: The operative details we need to capture in this comment is why there's a difference, i.e., why we sometimes disable physics vs sometimes we want the underlying standard physics. Please see the request in my original comment.

CillianMyles: Updated ✅

super_editor/lib/src/super_textfield/infrastructure/text_scrollview.dart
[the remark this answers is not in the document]
CillianMyles: Comment added + updated! ✅

super_editor/test/super_textfield/super_textfield_gesture_scrolling_test.dart
CillianMyles: Yes I can do that. It would be easier after landing this PR, as I could modify the tests I've added to use this `SuperTextFieldInspector.findScrollOffset()` and show that it produces incorrect results. Would that work?

matthew-carroll (approved):
LGTM

matthew-carroll:

PR #2142, opened 2024-07-05
1962 slack mention clone minor changes

BazinC:
This PR includes minor changes we would like to get shipped with https://github.com/superlistapp/super_editor/pull/2038.

- Enable providing custom edit request handlers by constructor
- Enable providing the trigger tby constuctor (default to '@')
- Make the removeSlackComposingTokenAttribution method public
- Provide a way to set value on the SlackTagIndex
- Replace print() statements with logger.
- Remove 2 AssertionError thrown when running on Android. They lead to the editor stuck in an unusable state. Removed because the situation happen without encountering any edge case. Normal usage.

### DIFF
--- a/super_editor/lib/src/clones/slack/slack_tags.dart
+++ b/super_editor/lib/src/clones/slack/slack_tags.dart
@@ -31,12 +31,13 @@ class SlackTagPlugin extends SuperEditorPlugin {
   /// The key used to access the [SlackTagIndex] in an attached [Editor].
   static const slackTagIndexKey = "slackTagIndex";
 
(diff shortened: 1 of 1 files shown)


BazinC:
@matthew-carroll let me know if it is possible to merge this into your branch

matthew-carroll:
Looks like a bunch of tests are failing. I'll make the changes you listed directly on my branch.

In general, there's no reason to put up a PR to change a branch that someone is already working on. In this case, I would have preferred that you just tell me what you want me to adjust.

BazinC:
Understood

PR #2155, opened 2024-07-09
[SuperEditor][SuperReader][iOS] Fix horizontal drag scrolling the editor (Resolves #2081)

### LINKED ISSUE #2081: [SuperEditor] [iOS] Vertical scrolling occurring during horizontal drag gestures
**Package Version**
stable branch

In iOS apps, views containing a Super Editor can be scrolled vertically even when dragging horizontally. Once a horizontal drag gesture is initiated, only movement along the x-axis is consumed, and y-axis movements are ignored until a new drag gesture is started.

https://github.com/superlistapp/super_editor/assets/172330941/e45f555f-91a5-4d36-98bf-8bafbd40a163

angelosilvestre:
[SuperEditor][SuperReader][iOS] Fix horizontal drag scrolling the editor. Resolves #2081 

Dragging horizontally is causing the editor to scroll:

https://github.com/superlistapp/super_editor/assets/7597082/0f1e9077-4a44-4513-9ace-ef5c042a19fd

Additionally, if the user starts dragging horizontally, they can't drag vertically to scroll the document without releasing the finger:

https://github.com/superlistapp/super_editor/assets/7597082/b545cbbc-e553-4685-bc86-a7cc102fe663

Both issues happen because we use a combination of a `HorizontalDragRecognizer` and a `VerticalDragRecognizer`. We use this to work around the fact that, when used inside a `CustomScrollView`, the `CustomScrollView` vertical recognizer wins the gesture arena, preventing the user from dragging horizontally to drag the handles.

For the first issue, we start scrolling the document even if the drag event is a horizontal drag. For the second issue, the horizontal drag events always report a zero `dy` in the drag event details.

To solve this, we should use a modified version of `PanGestureRecognizer`, so we can win earlier in the arena. 

Unfortunately, Flutter's `DragGestureRecognizer` (the base class for `VerticalDragGestureRecognizer`, `HorizontalDragGestureRecognizer` and `PanGestureRecognizer`) relies on private abstract methods, making it impossible for us to create our own subclass. As a workaround, I copied `DragGestureRecognizer` and made the abstract methods public. I filed a ticket on the Flutter repo to that on the original recognizer: https://github.com/flutter/flutter/issues/151446

This PR introduces a custom pan gesture recognizer, I named it `EagerPanGestureRecognizer` for the lack of a better name.

It's main difference lives in the `hasSufficientGlobalDistanceToAccept` method. `PanGestureRecognizer` checks if the distance moved is bigger than the pan slop, while the `VerticalDragGestureRecognizer` checks if it's bigger the touch slop. Since the pan slop is twice as big as the touch slop, the `VerticalDragGestureRecognizer` self declares win earlier than the  `PanGestureRecognizer`.

With out pan recognizer using the touch slop instead of the pan slop, we can beat the `VerticalDragGestureRecognizer` in the arena:

https://github.com/superlistapp/super_editor/assets/7597082/986d46b9-ffb2-47d1-b63e-318a700683f2

https://github.com/superlistapp/super_editor/assets/7597082/ee304d3b-85cd-4c57-b689-9fd376dfe59d


super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
```diff
@@ -896,6 +898,9 @@ class _IosDocumentTouchInteractorState extends State<IosDocumentTouchInteractor>
   }
 
   void _onPanStart(DragStartDetails details) {
+    // Store the details so we can start a drag in the onPanUpdate.
+    _lastDragStartDetails = details;
```
matthew-carroll, line 902: If the root cause was winning the gesture arena, and you switched to a new gesture recognizer that wins earlier, can you explain why you needed so many additional changes to the panning callbacks? I.e., what problems are you solving here outside of winning the arena earlier?

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
```diff
@@ -1450,7 +1467,9 @@ enum DragMode {
   // around the selected word.
   longPress,
   // Dragging to scroll the document.
-  scroll
+  scroll,
+  // We still don't know if the user is dragging vertically or horizontally.
```
matthew-carroll, line 1471: What's the relevant difference between "unknown" and "scroll"? What would happen if we used "scroll" in the places where you're currently using "unknown"?

Also, in general, we should avoid referring to states as "unknown". They become catch-alls for any number of things and we'll never know which is which. If this state is actually something like "waitingForGestureArena" or something like that, then we should use the specific name so we can avoid ambiguity in the future.

super_editor/lib/src/infrastructure/eager_pan_gesture_recognizer.dart
```diff
@@ -0,0 +1,59 @@
+import 'package:flutter/foundation.dart';
+import 'package:flutter/gestures.dart' hide DragGestureRecognizer;
+
+import 'package:super_editor/src/infrastructure/flutter/monodrag.dart';
+
```
matthew-carroll, line 5: Would this file make more sense under `flutter`? Are we only putting direct copies under `flutter`, or does `flutter` also include extensions on existing Flutter capabilities?

super_editor/lib/src/infrastructure/flutter/monodrag.dart
```diff
@@ -0,0 +1,804 @@
...
+import 'package:flutter/foundation.dart';
+import 'package:flutter/gestures.dart';
+import 'package:flutter/scheduler.dart';
+
+/// A copy of Flutter's `DragGestureRecognizer` with public abstract methods.
+///
+/// Flutter's `DragGestureRecognizer` relies on private abstract methods to work. Since the
+/// built-in subclasses live on the same package, they can override these methods, but we don't.
+///
+/// This class can be removed if https://github.com/flutter/flutter/issues/151446 is resolved.
```
matthew-carroll, line 12: Can you link to the file where this was copied from? Please grab a link that refers to a specific hash so that future changes in Flutter won't make the link stop working.

super_editor/test/super_editor/mobile/super_editor_ios_selection_test.dart
```diff
@@ -428,6 +427,75 @@ void main() {
           await tester.pump();
         });
       });
+
+      group("horizontal drag", () {
+        testWidgetsOnIos("does not cause editor to scroll", (tester) async {
```
matthew-carroll, line 432: To be clear, when the horizontal drag was causing scroll behavior, was it truly a horizontal drag, or was it mostly horizontal with some amount of vertical movement included?

super_editor/test/super_editor/mobile/super_editor_ios_selection_test.dart
```diff
@@ -428,6 +427,75 @@ void main() {
...
+          );
+
+          // Drag horizontally.
+          for (int i = 1; i < 10; i += 1) {
+            await gesture.moveBy(const Offset(20, 0));
+            await tester.pump();
+          }
+
+          // Ensure that dragging doesn't cause the editor to scroll.
+          //expect(scrollController.offset, 0);
```
matthew-carroll, line 483: Should this be commented out?

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
angelosilvestre: The gesture arena is one issue. This other modifications are here to detect if the user is dragging vertically or horizontally.

In the pan start method we still don't know if the user is dragging vertically or horizontally (if it's possible, please correct me), so I'm deferring some decisions to the pan update callback.

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
angelosilvestre: Maybe "waitingForDisambiguiation"? In this state we are waiting to detect if the user is dragging vertically or horizontally.

If we just use scroll then we will still scroll the document when dragging horizontally.

super_editor/lib/src/infrastructure/eager_pan_gesture_recognizer.dart
angelosilvestre: We can move it there.

super_editor/lib/src/infrastructure/flutter/monodrag.dart
angelosilvestre: Done.

super_editor/test/super_editor/mobile/super_editor_ios_selection_test.dart
angelosilvestre: It was truly a horizontal drag.

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
matthew-carroll: Let's call this `_dragStartDetails` - the `last` is confusing for timing.

Also when you say "// Store the details so we can start a drag in the onPanUpdate.", this describe a mechanism, but not the purpose. The important details are "why".

It sounds like in this case the "why" is "// Store the gesture start details to disambiguate horizontal vs vertical dragging, later."

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
matthew-carroll: How about `waitingForScrollDirection`?

angelosilvestre:
@matthew-carroll I addressed your latest comments.

matthew-carroll (approved):
LGTM

PR #2158, opened 2024-07-10
Remove nodes list from Document to make Document immutable at API level (Resolves #2156)

### LINKED ISSUE #2156: [SuperEditor] - Remove mutability from Document API
The `Document` API includes access to a `List` of `DocumentNode`s, which means there's an implicit mutability in an interface that should be immutable. To avoid confusion, adjust the API to be explicitly immutable.

matthew-carroll:
Remove nodes list from Document to make Document immutable at API level (Resolves #2156)

This PR has a large change list, but it all boils down to removing the `nodes` property on `Document` in favor of accessing nodes with methods and properties, so that clients can't mutate a `Document`.

`Document` now implements `Iterable<DocumentNode>` so that it's still easy to iterate through the nodes in a document. `Iterable` also introduces a `first` and `last` property, which is honored. I also created `firstOrNull` and `lastOrNull` to access those properties in a possibly empty `Document`.

### DIFF
--- a/super_editor/lib/src/default_editor/multi_node_editing.dart
+++ b/super_editor/lib/src/default_editor/multi_node_editing.dart
@@ -22,7 +22,7 @@ class PasteStructuredContentEditorRequest implements EditRequest {
     required this.pastePosition,
   });
 
-  final List<DocumentNode> content;
+  final Document content;
   final DocumentPosition pastePosition;
 }
 
@@ -30,12 +30,12 @@ class PasteStructuredContentEditorRequest implements EditRequest {
 /// given paste position within the document.
 class PasteStructuredContentEditorCommand extends EditCommand {
   PasteStructuredContentEditorCommand({
-    required List<DocumentNode> content,
+    required Document content,
     required DocumentPosition pastePosition,
   })  : _content = content,
         _pastePosition = pastePosition;
 
-  final List<DocumentNode> _content;
+  final Document _content;
   final DocumentPosition _pastePosition;
 
   @override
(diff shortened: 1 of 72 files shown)


matthew-carroll:
@angelosilvestre I'm gonna merge this in so it doesn't go stale. Feel free to follow up with any issues that you find.

super_editor/lib/src/default_editor/multi_node_editing.dart
```diff
@@ -127,7 +127,7 @@ class PasteStructuredContentEditorCommand extends EditCommand {
   void _pasteMultipleNodes(
     CommandExecutor executor,
     MutableDocument document,
-    List<DocumentNode> pastedNodes,
+    Document pastedNodes,
```
angelosilvestre, line 130: Having a `pastedNodes` typed as a `Document` looks a bit strange IMO. 

What if we declare it as an `Iterable<DocumentNode>`. Since `Document` implements `Iterable<DocumentNode>` the compiler won't complain it we give it a `Document`.

angelosilvestre (commented):
LGTM with a single comment.

super_editor/lib/src/default_editor/multi_node_editing.dart
matthew-carroll: I was torn on this. I went back and forth. I think my question would be, what's the difference between a `Document` and a `List<DocumentNode>`? If you paste some content, are you not essentially pasting a document within a document? That answer could go either way, but I couldn't think of a strong reason not to call it a `Document`. Let me know if you can think of any.

PR #2181, opened 2024-07-22
[SuperEditor] Update CancelComposingActionTagCommand to use same logic as ActionTagComposingReaction._cancelComposingTag (Resolves #2177)

### LINKED ISSUE #2177: [BUG] - Exception when pressing space after cancelling action tag
**Package Version**

**User Info**
coneno, QuikFlow

**To Reproduce**
1. Go to the action tags demo in the example app
2. Write an action tag which does not match any of the options (so that "NO ACTIONS" is displayed)
3. Cancel using escape
4. Press space

**Actual behavior**
An exception is thrown.
Screenshot 2024-07-19 at 20 43 53

**Expected behavior**
A space is inserted and no exception is thrown.

**Platform**
macOS

**Flutter version**
Flutter 3.22.3 • channel stable • https://github.com/flutter/flutter.git
Framework • revision b0850beeb2 (3 days ago) • 2024-07-16 21:43:41 -0700
Engine • revision 235db911ba
Tools • Dart 3.4.4 • DevTools 2.34.3

KevinBrendel:
Fixes https://github.com/superlistapp/super_editor/issues/2177.

Effectively, when pressing space, `ActionTagComposingReaction._cancelComposingTag` removes the actionTagComposingAttribution and adds an actionTagCancelledAttribution to the first character (the trigger character).

CancelComposingActionTagCommand also removes the actionTagComposingAttribution, but adds the actionTagCancelledAttribution to the full length of the original actionTagComposingAttribution. ActionTagComposingReaction._healCancelledTags does not seem to properly deal with this, which leads to the exception outlined in the issue.


KevinBrendel:
Not sure why this goldens file is different. Visually, I can't tell a difference.

KevinBrendel:
Since the same Goldens are also failing in other PRs, it probably doesn't have anything to do with this PR. Can remove the update if you want to.

matthew-carroll:
@KevinBrendel I didn't see this PR because I wasn't tagged on it, so I already put up a PR that makes this change, along with some other adjustments. Please check that other PR and make sure it works for you. I'm going to close this PR in favor of that one.

KevinBrendel:
@matthew-carroll I didn't know you had already seen the issue so I assumed you would also see this PR because the issue shows that it is referenced by this PR.

matthew-carroll:
@KevinBrendel it's best to assume that if you don't "@" me in something, I probably won't see it. I've never been able to get GitHub to send me the right selection of update emails.

PR #2182, opened 2024-07-22
[SuperEditor] expose canUndo and canRedo

BazinC:
In order to build a toolbar with undo/redo buttons on mobile, it would be nice to expose the state of the history lists.
This way we can build button with enabled/disabled UI state.

### DIFF
--- a/super_editor/lib/src/core/editor.dart
+++ b/super_editor/lib/src/core/editor.dart
@@ -353,6 +353,10 @@ class Editor implements RequestDispatcher {
     _isReacting = false;
   }
 
+  bool canUndo() => _history.isNotEmpty;
+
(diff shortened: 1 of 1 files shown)


matthew-carroll:
History and future are both now publicly visible: https://github.com/superlistapp/super_editor/blob/main/super_editor/lib/src/core/editor.dart#L86

This is why I'd prefer that you start by filing an issue ticket instead of jumping straight to a PR. Your needs might correspond to other related needs, and we miss out on the chance to discover that when you go straight to a code change.

I'm going to close this PR because as far as I can tell you now have access to what you need. If you still can't achieve your goal, then please post your goal in an issue ticket where we can discuss what you're trying to do.

PR #2190, opened 2024-07-26
[SuperEditor] - Fix: Paste command remembers the nodes it inserted to play nice with undo/redo (Resolves #2173)

### LINKED ISSUE #2173: [BUG] - undo crash after pasting
**Package Version**

**To Reproduce**
1. Run web's example
2. Copy the last few nodes
3. Paste
4. Type in something
5. Cmd Z several times

**Platform**
Web, possibly also on Android.

**Flutter version**
3.22.2

**Screen recording**
https://github.com/user-attachments/assets/1c066631-ee90-4344-b361-437d51f14c1e

matthew-carroll:
[SuperEditor] - Fix: Paste command remembers the nodes it inserted to play nice with undo/redo (Resolves #2173)

Previously, the paste command was re-parsing the pasted content every time it ran. This resulted in random node IDs every time that it was undone and redone. Therefore, if the user typed more characters in the pasted content, calling "undo" would result in text insertions into nodes with IDs that no longer existed, because the paste command generated new node IDs when redone.

Now, the paste command only parses its content one time. When the paste command is re-run, it copies each of its `DocumentNode`s and inserts them in the document, resulting in the exact same node IDs and data every time.


angelosilvestre (approved):
LGTM

PR #2191, opened 2024-07-27
[SuperEditor] - Fix: Action tag exception when pressing ESC, then typing characters. (Resolves #2177)

### LINKED ISSUE #2177: [BUG] - Exception when pressing space after cancelling action tag

matthew-carroll:
[SuperEditor] - Fix: Action tag exception when pressing ESC, then typing characters. (Resolves #2177)

It's a little unclear, looking back at the Action Tag implementation, what the exact intention was in terms of attributing things when composing and attributing them after they're cancelled. However, it looks like we only apply the cancellation attribution to the trigger, e.g., "/". But, in one of the places where we implement cancellation, we were applying the cancellation to the whole token, which seems to have lead to an exception.

This PR adjusts the cancellation attribution to only apply to the first character.


angelosilvestre (approved):
LGTM

KevinBrendel:
@matthew-carroll Could this be merged, please? It is a pretty critical bug for us and we want to build a release candidate, soon.

PR #2193, opened 2024-07-27
[Spellcheck] Expose macOS spellcheck API's in dart (Resolves #2189)

### LINKED ISSUE #2189: [Spellcheck] - Expose Mac spellcheck APIs in Dart
Use method channels to expose Mac spellcheck APIs in Dart within the spellcheck plugin.

angelosilvestre:
[Spellcheck] Expose macOS spellcheck API's in dart (Resolves #2189)

This PR adds a method channel to call the macOS spellchecking API's. 

Pigeon was used to generate the binding code. There are some limitations, for example, lists must have nullable generic types. For this reason I had to define the list os suggestions as `List<String?>`.

I removed the default generated tests. I'm not sure which tests we should write for this, but I'm open to write any tests that would be useful.

The doc comments still need some improvements.

Here's a small demo:

https://github.com/user-attachments/assets/7d224871-b562-428d-9b8e-d22d7d7334f6


angelosilvestre:
@matthew-carroll Could you do a first pass review?

angelosilvestre:
I implemented a single method that call `checkSpelling` and `guesses` to return all of the suggestions in the text, but we can change that if needed.

super_editor_spellcheck/macos/Classes/SuperEditorSpellcheckPlugin.swift
```diff
@@ -1,19 +1,94 @@
...
+    SpellCheckApiSetup.setUp(binaryMessenger: registrar.messenger, api: instance)
   }
 
-  public func handle(_ call: FlutterMethodCall, result: @escaping FlutterResult) {
-    switch call.method {
-    case "getPlatformVersion":
-      result("macOS " + ProcessInfo.processInfo.operatingSystemVersionString)
-    default:
-      result(FlutterMethodNotImplemented)
+  /// Checks the given `text` for spelling errors with the given `language`.
```
matthew-carroll, line 13: Is `///` the actual syntax for multiline doc comments in Swift?

super_editor_spellcheck/macos/Classes/SuperEditorSpellcheckPlugin.swift
```diff
@@ -1,19 +1,94 @@
...
+  ///
+  /// Returns a list of `TextSuggestion`s, where each entry represents a
```
matthew-carroll, line 15: Can you document what happens when the same mis-spelled word exists multiple times? Do we get multiple `TextSuggestion`s for it? Or just one?

super_editor_spellcheck/macos/Classes/SuperEditorSpellcheckPlugin.swift
```diff
@@ -1,19 +1,94 @@
...
+  /// misspelled word, with the possible suggestions.
+  ///
+  /// Returns an empty list if no spelling errors are found or if the `language`
+  /// isn't supported by the spell checker.
+  func checkSpelling(language: String, text: String) throws -> [TextSuggestion] {
+    var result : [TextSuggestion] = [];
+
+    if (language.isEmpty || text.isEmpty) {
```
matthew-carroll, line 23: I'm wondering what we should do about errors. I suppose that `text` being empty is something that could happen in the regular course of things. But the `language` existence is more a contract within our own plugin, right? Is there any situation where `language` would be empty where a developer didn't make a mistake?

super_editor_spellcheck/macos/Classes/SuperEditorSpellcheckPlugin.swift
```diff
@@ -1,19 +1,94 @@
...
+      // We can't look for misspelled words without a language or text.
+      return result;
+    }
+
+    let spellChecker = NSSpellChecker.shared;
+
+    // Convert dart locale to macOS language code. For example, converts "pt-BR" to "pt_BR".
```
matthew-carroll, line 30: I'm wondering if we should do this conversion before we get here? My thought is that Dart code is easier to access, easier to read, and can be changed without recompiling platform code. Any reason not to hoist that up into Dart?

super_editor_spellcheck/macos/Classes/SuperEditorSpellcheckPlugin.swift
```diff
@@ -1,19 +1,94 @@
...
+    var languageCode = language.replacingOccurrences(of: "-", with: "_")
+    if (!spellChecker.availableLanguages.contains(languageCode)){
+      // The given language isn't supported by the spell checker. It might be the case that
+      // the user has a language configured with an incompatible region. For example,
+      // a user might have "en-BR" configured, which means that the language is English,
+      // but the region is Brazil. In this case, we should try to use only the language.
+      let firstPart = languageCode.components(separatedBy: "_").first;
```
matthew-carroll, line 37: Instead of calling this `firstPart`, let's give it a name that reflects what it is. Sounds like this is something like the `languageAbbreviation`, right?

super_editor_spellcheck/macos/Classes/SuperEditorSpellcheckPlugin.swift
```diff
@@ -1,19 +1,94 @@
...
+      if (firstPart == nil) {
+        // The language code isn't in the format language_REGION. Fizzle.
+        return result;
```
matthew-carroll, line 40: I'm wondering if we should differentiate between a failure to query for spelling errors, and actually finding zero spelling errors.

super_editor_spellcheck/macos/Classes/SuperEditorSpellcheckPlugin.swift
```diff
@@ -1,19 +1,94 @@
...
+      }
+
+      languageCode = firstPart!;
```
matthew-carroll, line 43: Looks like `firstPart` probably doesn't need to exist at all. You can just assign to `languageCode`, right?

super_editor_spellcheck/macos/Classes/SuperEditorSpellcheckPlugin.swift
```diff
@@ -1,19 +1,94 @@
...
+      let misspelledRange = spellChecker.checkSpelling(
+        of: text,
+        startingAt: currentOffset,
+        language: languageCode,
+        wrap: false,
+        inSpellDocumentWithTag: 0,
+        wordCount: nil
+      );
+
+      if (misspelledRange.location != NSNotFound) {
```
matthew-carroll, line 63: This conditional is tough to read because its a double negative "location is not equal to not found". Does Swift support a `continue` statement? If so, can we invert the condition and then continue to the next loop iteration?

matthew-carroll (commented):
I left some inline comments. I'm also wondering why we don't do a true method-for-method exposure of the spelling and grammar methods in the API?

https://developer.apple.com/documentation/appkit/nsspellchecker#1668670

Some of the methods appear stateful, and some are even visual, so I understand why we ignore those. But other than that, why not do a 1-to-1 exposure?

angelosilvestre:
> I'm also wondering why we don't do a true method-for-method exposure of the spelling and grammar methods in the API?

I agree we could expose each method separately. But I also think it's useful to keep this method to simplify the process of fetching suggestions for the whole text.

matthew-carroll:
> > I'm also wondering why we don't do a true method-for-method exposure of the spelling and grammar methods in the API?
> 
> I agree we could expose each method separately. But I also think it's useful to keep this method to simplify the process of fetching suggestions for the whole text.

Can that method become a Dart method instead? If we expose each relevant Mac OS method through the method channel, then whatever logic you wrote in the current plugin method could be done in Dart instead, right?

angelosilvestre:
> Can that method become a Dart method instead?

It can be. A downside would be that, if the user pastes a a big piece of text, we will be calling into the platform multiple times.

matthew-carroll:
> > Can that method become a Dart method instead?
> 
> It can be. A downside would be that, if the user pastes a a big piece of text, we will be calling into the platform multiple times.

Why is that?

super_editor_spellcheck/macos/Classes/SuperEditorSpellcheckPlugin.swift
angelosilvestre: Yes.

super_editor_spellcheck/macos/Classes/SuperEditorSpellcheckPlugin.swift
angelosilvestre: Updated to throw an error, as documented in https://github.com/flutter/packages/blob/main/packages/pigeon/example/README.md#swift

super_editor_spellcheck/macos/Classes/SuperEditorSpellcheckPlugin.swift
angelosilvestre: Updated.

super_editor_spellcheck/macos/Classes/SuperEditorSpellcheckPlugin.swift
angelosilvestre: Since we will expose almost all of the methods and we can query the available languages, should this be also moved to dart?

super_editor_spellcheck/macos/Classes/SuperEditorSpellcheckPlugin.swift
angelosilvestre: Updated.

angelosilvestre (commented):
I exposed more methods from `NSSpellChecker`.  There are still some to expose.

angelosilvestre:
BTW, I tried to keep the same docs that exist in the Apple's documentation.

super_editor_spellcheck/lib/src/spell_checker.dart
```diff
@@ -0,0 +1,278 @@
...
+import 'package:super_editor_spellcheck/src/messages.g.dart';
+
+/// Plugin to check spelling errors in text using the native macOS spell checker.
+class SuperEditorSpellCheckerPlugin {
+  final SpellCheckApi _spellCheckApi = SpellCheckApi();
+
+  /// Returns a unique tag to identified this spell checked object.
+  ///
+  /// Use this method to generate tags to avoid collisions with other objects that can be spell checked.
+  Future<int> uniqueSpellDocumentTag() async {
```
matthew-carroll, line 12: I see that we've replicated all the calls and this plugin is essentially operating as a proxy. I'm guessing we should take a different approach here.

First, does this plugin class do anything that the `SpellCheckApi` doesn't do? If not, why not just instantiate the `SpellCheckApi` wherever it's needed?

If this plugin does do things beyond just call to the `SpellCheckApi`, then would it make sense to expose the `SpellCheckApi` as a property of the plugin class, and then add whatever else you need beyond that within the plugin class?

super_editor_spellcheck/example/lib/main.dart
```diff
@@ -16,34 +18,98 @@ class MyApp extends StatefulWidget {
...
+      _superEditorSpellcheckPlugin.closeSpellDocumentWithTag(_documentTag!);
     }
+    super.dispose();
+  }
+
+  void _onTextChanged() {
+    _fetchSuggestions();
+  }
+
+  Future<void> _fetchSuggestions() async {
```
matthew-carroll, line 56: This method should probably wait for a slight delay, and also bail if another request comes in while it's running. Otherwise, you'll get results out of order.

matthew-carroll (commented):
I added comments. You also mentioned there are more methods that aren't ported over. What kind of behaviors do those methods cover?

super_editor_spellcheck/example/macos/Runner/AppDelegate.swift
```diff
@@ -6,4 +6,8 @@ class AppDelegate: FlutterAppDelegate {
   override func applicationShouldTerminateAfterLastWindowClosed(_ sender: NSApplication) -> Bool {
     return true
   }
+
+  override func applicationSupportsSecureRestorableState(_ app: NSApplication) -> Bool {
```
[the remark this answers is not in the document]
angelosilvestre: I think this was auto-generated, reverted it.

super_editor_spellcheck/lib/src/spell_checker.dart
angelosilvestre: It does change some things. Pigeon imposes us some limitations:
- We cannot import nothing. We have some methods that ideally would return a `TextRange`. Since we cannot import `dart:ui`, I had to define a custom `Range` class just to work with pigeon. Then, this class converts it to a `TextRange`.
- All generics (lists, maps, etc) must have nullable type parameters. For example, the `guesses` methods needs to return a list of nullable strings. In this class I re-map to non-nullable.

super_editor_spellcheck/pigeon/messages.dart
```diff
@@ -0,0 +1,197 @@
...
+  swiftOut: 'macos/Classes/messages.g.swift',
+))
+@HostApi()
+abstract class SpellCheckApi {
+  /// Checks the given [text] for spelling errors with the given [language].
+  ///
+  /// Returns a list of [TextSuggestion]s, where each spans represents a
+  /// misspelled word, with the possible suggestions.
+  ///
+  /// Returns an empty list if no spelling errors are found or if the [language]
```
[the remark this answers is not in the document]
angelosilvestre: Yeah, it just returns an empty list.

super_editor_spellcheck/pigeon/messages.dart
```diff
@@ -0,0 +1,197 @@
...
+  /// isn't supported by the spell checker.
+  List<TextSuggestion> fetchSuggestions({
+    required String text,
+    required String language,
+  });
+
+  /// Returns a unique tag to identified this spell checked object.
```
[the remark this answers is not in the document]
angelosilvestre: As far as I understood, this is used to enable multiple spell checking operations for the same app. For example, suppose we have two open documents, and we want to ignore a misspelled word only for one of them.

Most of the spell-checking methods can operate with this identifiers.

super_editor_spellcheck/pigeon/messages.dart
```diff
@@ -0,0 +1,197 @@
...
+  ///
+  /// Use this method to generate tags to avoid collisions with other objects that can be spell checked.
+  int uniqueSpellDocumentTag();
+
+  /// Notifies the receiver that the user has finished with the tagged document.
+  ///
+  /// The spell checker will release any resources associated with the document,
+  /// including but not necessarily limited to, ignored words.
+  void closeSpellDocument(int tag);
```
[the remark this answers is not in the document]
angelosilvestre: I explained that in my previous comment.

super_editor_spellcheck/pigeon/messages.dart
```diff
@@ -0,0 +1,197 @@
...
+
+  /// Starts the search for a misspelled word in [stringToCheck] starting at [startingOffset]
```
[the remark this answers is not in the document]
angelosilvestre: It isn't a long-lived action. I copied this doc from the Apple's documentation, but I agree we should change it. Updated.

super_editor_spellcheck/pigeon/messages.dart
```diff
@@ -0,0 +1,197 @@
...
+  /// within the string object.
+  ///
+  /// - [stringToCheck]: The string object containing the words to spellcheck.
+  /// - [startingOffset]: The offset within the string object at which to start the spellchecking.
+  /// - [language]: The language of the words in the string.
+  /// - [wrap]: `true` to indicate that spell checking should continue at the beginning of the string
```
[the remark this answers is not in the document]
angelosilvestre: Also from the Apple docs, it's indeed confusing. Basically, if `wrap` is true and we give a `startingOffset` greater than zero, if no spelling errors are found between `startingOffset` and the end of the string, it will start again from the beginning...

super_editor_spellcheck/pigeon/messages.dart
```diff
@@ -0,0 +1,197 @@
...
+  ///
+  /// Returns an array of strings containing possible replacement words.
+  List<String?>? guesses({
+    required Range range,
+    required String text,
+    String? language,
+    int inSpellDocumentWithTag = 0,
+  });
+
+  /// Initiates a grammatical analysis of a given string.
```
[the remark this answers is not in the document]
angelosilvestre: Updated to "Performs". Or maybe "Searches for grammatical errors in the given string" ?

super_editor_spellcheck/pigeon/messages.dart
```diff
@@ -0,0 +1,197 @@
...
+  /// Returns the list of complete words from the spell checker dictionary in the order
+  /// they should be presented to the user.
+  List<String>? completions({
+    required Range partialWordRange,
+    required String text,
+    String? language,
+    int inSpellDocumentWithTag = 0,
+  });
+
+  /// Returns the number of words in the specified string.
```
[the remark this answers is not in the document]
angelosilvestre: The Apple docs say "If language is nil, the current selection in the Spelling panel’s pop-up menu is used.", but I have no idea what the "Spelling panel’s pop-up" is.

super_editor_spellcheck/pigeon/messages.dart
```diff
@@ -0,0 +1,197 @@
...
+  void unlearnWord(String word);
+
+  /// Instructs the spell checker to ignore all future occurrences of [word] in the document
+  /// identified by [documentTag].
+  void ignoreWord({required String word, required int documentTag});
+
+  /// Returns the array of ignored words for a document identified by [documentTag].
+  List<String>? ignoredWords({required int documentTag});
+
+  /// Initializes the ignored-words document (a dictionary identified by [documentTag] with [words]),
```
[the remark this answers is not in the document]
angelosilvestre: Updated.

super_editor_spellcheck/pigeon/messages.dart
```diff
@@ -0,0 +1,197 @@
...
+  /// an array of words to ignore.
+  void setIgnoredWords({required List<String> words, required int documentTag});
+
+  /// Returns the dictionary used when replacing words.
```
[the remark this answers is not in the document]
angelosilvestre: Basically, this is the method that we would query to implement the text replacement, mentioned in this ticket: https://github.com/superlistapp/super_editor/issues/1890

This is a user-defined dictionary. I think it's safe to assume that it won't have thousands of words.

super_editor_spellcheck/pigeon/messages.dart
```diff
@@ -0,0 +1,197 @@
...
+///
+/// The [end] index is exclusive.
+class TextSuggestion {
+  TextSuggestion({
+    required this.start,
+    required this.end,
+    required this.suggestions,
+  });
+
+  final int start;
```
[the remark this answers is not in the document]
angelosilvestre: Pigeon crashes when we try to add a getter. But even if it didn't, we cannot import anything inside this file...

super_editor_spellcheck/pigeon/messages.dart
```diff
@@ -0,0 +1,197 @@
...
+/// the one at [end].
+///
+/// This is used because we can't use `TextRange` in pigeon.
+class Range {
+  Range({
+    required this.start,
+    required this.end,
+  });
+
+  final int start;
```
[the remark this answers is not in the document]
angelosilvestre: I haven't looked deeper to see why. But the generator explicitly says we cannot import anything useful:

```
Unsupported import 'dart:ui', only imports of 'package:pigeon/pigeon.dart' are supported.
```

super_editor_spellcheck/pigeon/messages.dart
```diff
@@ -0,0 +1,197 @@
...
+  /// The range of the first error found in the text or `null` if no errors were found.
+  final Range? firstError;
+
+  /// A list of details about the grammatical errors found in the text or `null`
+  /// if no errors were found.
+  final List<PlatformGrammaticalAnalysisDetail?>? details;
+}
+
+/// A detail about a grammatical error found in a text.
+class PlatformGrammaticalAnalysisDetail {
```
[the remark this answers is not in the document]
angelosilvestre: Related to the previous comment about the motivation for the `SuperEditorSpellCheckerMacPlugin` class.  Pigeon forces us to declare `details` as a list of nullable items. So I declared here with the "Platform" prefix and then added a implementation that return a list of non-nullable items named as `CheckGrammarResult`.

angelosilvestre:
> You also mentioned there are more methods that aren't ported over. What kind of behaviors do those methods cover?

The most relevant for us is these: 

- https://developer.apple.com/documentation/appkit/nsspellchecker/1535739-check, which performs multiple analysis in a single call, like orthography, grammar, locate dates and phone number, among other things.
- https://developer.apple.com/documentation/appkit/nsspellchecker/1531542-correction, which returns a single proposed correction for a misspelled word

We could also expose the available languages as a method.

Should we implement these?

angelosilvestre:
@matthew-carroll I pushed an alternative implementation, where `fetchSuggestions` is implemented in dart instead of swift.

super_editor_spellcheck/example/lib/main.dart
```diff
@@ -16,34 +18,111 @@ class MyApp extends StatefulWidget {
...
+      const Duration(milliseconds: 300),
+      _fetchSuggestions,
+    );
+  }
+
+  Future<void> _fetchSuggestions() async {
+    final textToSearch = _textController.text;
+
+    int? tag = _documentTag;
+    tag ??= await _superEditorSpellcheckPlugin.uniqueSpellDocumentTag();
```
matthew-carroll, line 65: After each of these `await` calls, you should check for `mounted`. Also, shouldn't we check to make sure that this request is still desired after each await call?

super_editor_spellcheck/example/lib/main.dart
```diff
@@ -16,34 +18,111 @@ class MyApp extends StatefulWidget {
...
+
+    final suggestions = await _superEditorSpellcheckPlugin.fetchSuggestions(
+      PlatformDispatcher.instance.locale,
```
matthew-carroll, line 68: I'm sure these calls run very quickly, but accessing a global mutable property across async boundaries, and assuming it never changes, is risky. We should probably query this before calling any of these methods and then pass the same variable into each one.

super_editor_spellcheck/lib/src/spell_checker.dart
```diff
@@ -0,0 +1,364 @@
+import 'dart:ui';
+
+import 'package:super_editor_spellcheck/src/messages.g.dart';
+
+/// Plugin to check spelling errors in text using the native macOS spell checker.
+class SuperEditorSpellCheckerMacPlugin {
+  final SpellCheckMac _spellCheckApi = SpellCheckMac();
+
+  /// A list containing all the available spell checking languages. The languages are ordered
```
matthew-carroll, line 9: Please make sure Dart Docs begin with only a single sentence.

super_editor_spellcheck/lib/src/spell_checker.dart
```diff
@@ -0,0 +1,364 @@
...
+  Future<List<String>> availableLanguages() async {
+    final languages = await _spellCheckApi.availableLanguages();
+
+    return languages //
+        .where((e) => e != null)
+        .cast<String>()
+        .toList();
+  }
+
+  /// Returns a unique tag to identified this spell checked object.
```
matthew-carroll, line 20: Probably a typo: "to identified" -> "to identify"

Also, when I read this, I don't know what "this spell checked object" refers to...

super_editor_spellcheck/lib/src/spell_checker.dart
```diff
@@ -0,0 +1,364 @@
...
+  ///
+  /// Use this method to generate tags to avoid collisions with other objects that can be spell checked.
+  Future<int> uniqueSpellDocumentTag() async {
+    return await _spellCheckApi.uniqueSpellDocumentTag();
+  }
+
+  /// Notifies the receiver that the user has finished with the tagged document.
```
matthew-carroll, line 27: Who's the "receiver"?

super_editor_spellcheck/lib/src/spell_checker.dart
```diff
@@ -0,0 +1,364 @@
...
+  ///
+  /// The spell checker will release any resources associated with the document,
+  /// including but not necessarily limited to, ignored words.
+  Future<void> closeSpellDocumentWithTag(int tag) async {
+    await _spellCheckApi.closeSpellDocument(tag);
+  }
+
+  /// Checks the given [text] for spelling errors with the given [locale].
+  ///
+  /// Returns a list of [TextSuggestion]s, where each spans represents a
```
matthew-carroll, line 37: "spans" -> "span"

super_editor_spellcheck/lib/src/spell_checker.dart
```diff
@@ -0,0 +1,364 @@
...
+  /// misspelled word, with the possible suggestions.
+  ///
+  /// Returns an empty list if no spelling errors are found or if the [locale]
```
matthew-carroll, line 40: Is returning an empty list for an unsupported locale the same behavior that Mac has?

super_editor_spellcheck/lib/src/spell_checker.dart
```diff
@@ -0,0 +1,364 @@
...
+  /// isn't supported by the spell checker.
+  ///
+  /// If the same misspelled word is found multiple times in the text, it will be
+  /// included in the list multiple times, since each range is different.
+  Future<List<TextSuggestion>> fetchSuggestions(
```
matthew-carroll, line 45: It looks like this method isn't really a Mac method - it's our own method that runs multiple Mac methods, including the repeated execution of `checkSpelling`. 

Can we rename this file to `spell_checker_mac.dart`, have the plugin only expose direct Mac behaviors, then have a `spell_checker.dart`, which offers our own chosen APIs and behavior?

When creating plugins, we want to retain the platform behavior, as it exists, and part of that goal also means not adding our own APIs, which have no corresponding platform implementation.

super_editor_spellcheck/lib/src/spell_checker.dart
```diff
@@ -0,0 +1,364 @@
...
+      }
+
+      // Place the offset after the current word to continue the search.
+      currentOffset += misspelledRange.end;
+    }
+
+    return result;
+  }
+
+  /// Starts the search for a misspelled word in [stringToCheck] starting at [startingOffset]
```
matthew-carroll, line 118: Let's reword this because it sounds like an ongoing process, and I don't think it is.

```
/// Searches for a misspelled word in [stringToCheck], starting at [startingOffset], and returns the [TextRange] surrounding the misspelled word.
///
/// If no misspelled word is found a [TextRange] is returned with bounds of `-1`, which can also be queried more conveniently with [TextRange.isValid].
///
/// To find all (or multiple) misspelled words in a given string, call this
method repeatedly, passing in different values for [startingOffset].
```

super_editor_spellcheck/lib/src/spell_checker.dart
```diff
@@ -0,0 +1,364 @@
...
+  /// Returns an array of possible substitutions for the specified string.
+  ///
+  /// - [range]: The range of the string to check.
+  /// - [text]: The string to guess.
+  /// - [inSpellDocumentWithTag]: An identifier unique within the application
+  ///   used to inform the spell checker which document that text is associated, potentially
+  ///   for many purposes, not necessarily just for ignored words. A value of 0 can be passed
+  ///   in for text not associated with a particular document.
+  ///
+  /// Returns an array of strings containing possible replacement words.
```
matthew-carroll, line 162: Replacements for wording in this Dart Doc:

"The range of the string to check" -> "The range, within the [text], for which possible substitutions should be generated."

"The string to guess" -> "The string containing the word/text for which substitutions should be generated."

"An identifier unique...." -> "The (optional) ID of the loaded document that contains the given [text], which is used to provide additional context to the substitution guesses. A value of '0' instructs the guessing system to consider the [text] in isolation, without connection to any given document."

Remove "Returns an array of string..."

super_editor_spellcheck/lib/src/spell_checker.dart
```diff
@@ -0,0 +1,364 @@
...
+      return <String>[];
+    }
+
+    return result //
+        .where((e) => e != null)
+        .cast<String>()
+        .toList();
+  }
+
+  /// Initiates a grammatical analysis of a given string.
```
matthew-carroll, line 186: Try to make changes to this Dart Doc that are similar to the ones I suggested for the `checkSpelling` method.

super_editor_spellcheck/lib/src/spell_checker.dart
```diff
@@ -0,0 +1,364 @@
...
+          [],
+    );
+  }
+
+  /// Provides a list of complete words that the user might be trying to type based on a
+  /// partial word in a given string.
+  ///
+  /// - [partialWordRange] - Range that identifies a partial word in string.
+  /// - [text] - String with the partial word from which to generate the result.
+  /// - [inSpellDocumentWithTag]: An identifier unique within the application
```
matthew-carroll, line 235: Try adjusting this description based on my recommendations for the `guesses` method.

super_editor_spellcheck/lib/src/spell_checker.dart
```diff
@@ -0,0 +1,364 @@
...
+        .toList();
+  }
+
+  /// Initializes the ignored-words document (a dictionary identified by [documentTag] with [words]),
+  /// an array of words to ignore.
+  Future<void> setIgnoredWords({required List<String> words, required int documentTag}) async {
+    await _spellCheckApi.setIgnoredWords(words: words, documentTag: documentTag);
+  }
+
+  /// Returns the dictionary used when replacing words.
```
matthew-carroll, line 313: Is there any more info we can provide about what this is, how big it is, how it might be used?

super_editor_spellcheck/lib/src/spell_checker.dart
```diff
@@ -0,0 +1,364 @@
...
+    if (locale == null) {
+      return null;
+    }
+
+    return locale.toLanguageTag().replaceAll("-", "_");
+  }
+}
+
+class CheckGrammarResult {
+  final TextRange? firstError;
```
matthew-carroll, line 336: Properties below constructors. Please update any such cases in the PR.

super_editor_spellcheck/lib/src/spell_checker.dart
```diff
@@ -0,0 +1,364 @@
...
+  final List<GrammaticalAnalysisDetail> details;
+
+  CheckGrammarResult({
+    this.firstError,
```
matthew-carroll, line 340: Is this actually the `firstError` in the text? Or is this just the `errorRange`, and the user might be iterating through the first, second, third, etc grammar errors in a string?

super_editor_spellcheck/lib/src/spell_checker.dart
```diff
@@ -0,0 +1,364 @@
...
+}
+
+class GrammaticalAnalysisDetail {
+  GrammaticalAnalysisDetail({
+    required this.range,
+    required this.userDescription,
+  });
+
+  final TextRange range;
+  final String userDescription;
```
matthew-carroll, line 352: Please describe what this is.

super_editor_spellcheck/macos/Classes/SuperEditorSpellcheckPlugin.swift
```diff
@@ -1,19 +1,215 @@
...
+  /// in the user’s preferred order as set in the system preferences.
+  func availableLanguages() throws -> [String?] {
+    return NSSpellChecker.shared.availableLanguages;
+  }
+
+  /// Returns a unique tag to identified this spell checked object.
+  ///
+  /// Use this method to generate tags to avoid collisions with other objects that can be spell checked.
+  func uniqueSpellDocumentTag() throws -> Int64 {
+    return Int64(NSSpellChecker.uniqueSpellDocumentTag());
```
matthew-carroll, line 23: How/where does this get associated with a document?

super_editor_spellcheck/macos/Classes/SuperEditorSpellcheckPlugin.swift
```diff
@@ -1,19 +1,215 @@
...
+      of: stringToCheck,
+      startingAt: Int(startingOffset),
+      language: language,
+      wrap: wrap,
+      inSpellDocumentWithTag: Int(inSpellDocumentWithTag),
+      details: &details
+    );
+    
+    if (grammarRange.location == NSNotFound || details == nil) {
+      return PlatformCheckGrammarResult(
```
matthew-carroll, line 120: Does Swift have named constructors? If so, can we name this something like `none`?

super_editor_spellcheck/pigeon/messages.dart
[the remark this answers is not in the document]
matthew-carroll: If the constraint is Pigeon, then that's put that in the name - because it doesn't sound like the constraint is actually the "Platform"

`PigeonGrammaticalAnalysisDetail`

super_editor_spellcheck/pigeon/messages.dart
```diff
@@ -0,0 +1,174 @@
...
+  Map<String, String> userReplacementsDictionary();
+}
+
+/// A range of characters in a string of text.
+///
+/// The text included in the range includes the character at [start], but not
+/// the one at [end].
+///
+/// This is used because we can't use `TextRange` in pigeon.
+class Range {
```
matthew-carroll, line 137: Should this be called `PigeonRange` to match the other places where special things were created for Pigeon? Or is there some other reason that this exists?

super_editor_spellcheck/lib/src/spell_checker.dart
```diff
@@ -0,0 +1,364 @@
...
+
+import 'package:super_editor_spellcheck/src/messages.g.dart';
+
+/// Plugin to check spelling errors in text using the native macOS spell checker.
+class SuperEditorSpellCheckerMacPlugin {
+  final SpellCheckMac _spellCheckApi = SpellCheckMac();
+
+  /// A list containing all the available spell checking languages. The languages are ordered
+  /// in the user’s preferred order as set in the system preferences.
+  Future<List<String>> availableLanguages() async {
```
matthew-carroll, line 11: For all the methods that have the same signature as the `SpellCheckMac`, let's de-duplicate them. Otherwise they'll drift apart.

Pick whichever of these classes you think should be the source of truth, write the Dart Docs for each method there, surround them with templates, and then just repeat the templates in the other location.

super_editor_spellcheck/lib/src/spell_checker.dart
angelosilvestre: Updated.

super_editor_spellcheck/lib/src/spell_checker.dart
angelosilvestre: It's the first error in the text that was found after the starting offset. If there are more errors, they will be in the details. I made it that way because it's how the mac API returns the results.

super_editor_spellcheck/lib/src/spell_checker.dart
angelosilvestre: Updated.

super_editor_spellcheck/macos/Classes/SuperEditorSpellcheckPlugin.swift
angelosilvestre: If I understood it correctly, it isn't associated with the text of a document. The main purpose is to associate ignored words to this tag. So when the user uses a different tag, these ignored words aren't considered.

super_editor_spellcheck/macos/Classes/SuperEditorSpellcheckPlugin.swift
angelosilvestre: Pigeon doesn't seem to support named constructors.

super_editor_spellcheck/pigeon/messages.dart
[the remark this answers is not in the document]
matthew-carroll: I didn't see an explanation of how/when/where to use this. Looking at this code related to tags, I have no idea what one is supposed to do with them...

super_editor_spellcheck/macos/Classes/SuperEditorSpellcheckPlugin.swift
matthew-carroll: Ok, we need to explain all this in the relevant docs. If Apple didn't explain it, then so be it, but we'll want Flutter developers to know how to use these things.

If there is no actual document behind a tag, then it sounds like a tag corresponds to an OS managed group of ignored words and whatever else you're able to mutate. You create two tags: A and B. You can then instruct Mac to learn the word "ljsglefs" for document A, but it will still be seen as misspelled in document B. And by "document" we just mean partitioned based on the tag. Is that right?

You may want to try it out if you're not sure. People will look to us to tell them how it works.

angelosilvestre: Updated.

matthew-carroll (approved):
LGTM - we'll handle further adjustments based on our experience using this plugin

PR #2201, opened 2024-07-29
[SuperEditor] Fix ActionTagComposingReaction looking for composing tag at selection base twice instead of at base and at extent.

KevinBrendel:
Without this fix, this could cause an exception, as the condition checks if `extent.nodePosition` is a `TextNodePosition` and then assumes that `base.nodePosition` is a `TextNodePosition`.

### DIFF
--- a/super_editor/test/super_editor/text_entry/tagging/action_tags_test.dart
+++ b/super_editor/test/super_editor/text_entry/tagging/action_tags_test.dart
@@ -668,8 +668,15 @@ void main() {
       await tester.pressRightArrow();
 
       // Select upstream towards the cancelled action tag
-      await tester.pressShiftLeftArrow();
-      await tester.pressShiftUpArrow();
+      await expectLater(
+        () async {
+          await tester.pressShiftLeftArrow();
+          await tester.pressShiftUpArrow();
+        },
+        returnsNormally,
+      );
+
+      // If we reach the end without exception, then ActionTagComposingReaction did not fail due to selection.base.nodePosition not being a TextNodePosition.
     });
   });
 }
--- a/super_editor/test/super_editor/text_entry/tagging/action_tags_test.dart
+++ b/super_editor/test/super_editor/text_entry/tagging/action_tags_test.dart
@@ -679,6 +679,7 @@ void main() {
       );
 
       // If we reach the end without exception, then ActionTagComposingReaction did not fail due to selection.base.nodePosition not being a TextNodePosition.
+      // See also: https://github.com/superlistapp/super_editor/pull/2201
     });
   });
 }
--- a/super_editor/test/super_editor/text_entry/tagging/action_tags_test.dart
+++ b/super_editor/test/super_editor/text_entry/tagging/action_tags_test.dart
@@ -653,7 +653,9 @@ void main() {
   });
 
   group("selections >", () {
-    testWidgetsOnAllPlatforms("upstream cancelled action tag from block node", (tester) async {
+    testWidgetsOnAllPlatforms(
+        "exception in ActionTagComposingReaction when selecting upstream from BlockNode while cancelled action tag exists",
+        (tester) async {
       await _pumpTestEditor(
         tester,
         paragraphThenHrDoc(),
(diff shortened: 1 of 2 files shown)


matthew-carroll:
Can you please express this bug as a failing test and then show that this change fixes the bug?

KevinBrendel:
Are there any existing tests related to action tags which I could use as a reference?

I currently don't have the setup to test it outside of our app, but essentially it can be triggered for example by:

1. Have one `ParagraphNode` with a cancelled ActionTag (careful not to trigger #2177)
2. Have a `HorizontalRuleNode`
3. Select upstream from the end of the `HorizontalRuleNode` into the `ParagraphNode`

It seems quite obvious to me that the code is not intended this way, as both the comment, as well as the condition indicate that extent is intended here. Base was already checked in the block before. (And it also fixes the exception, of course.)

https://github.com/user-attachments/assets/b6404371-c62a-4b97-b6ed-a69d1863a620

matthew-carroll:
> Are there any existing tests related to action tags which I could use as a reference?

Yeah, we try to do our best to comprehensively test all features. In general, all PRs that modify behavior (non docs) are expected to include corresponding tests.

Here are the current action tag tests:

https://github.com/superlistapp/super_editor/blob/main/super_editor/test/super_editor/text_entry/tagging/action_tags_test.dart

KevinBrendel:
@matthew-carroll Added a test which triggers the exception/demonstrates that the exception is no longer triggered with this fix.

super_editor/test/super_editor/text_entry/tagging/action_tags_test.dart
```diff
@@ -651,6 +651,27 @@ void main() {
       });
     });
   });
+
+  group("selections >", () {
+    testWidgetsOnAllPlatforms("upstream cancelled action tag from block node", (tester) async {
```
matthew-carroll, line 656: Can you come up with a better name for this test? As someone reading this who didn't work on the problem or the solution, I don't know what this string of words means...

super_editor/test/super_editor/text_entry/tagging/action_tags_test.dart
```diff
@@ -651,6 +651,27 @@ void main() {
...
+      await tester.placeCaretInParagraph("1", 0);
+      await tester.typeImeText("/header ");
+
+      // Place cursor at the end of the horizontal rule/block node
+      await tester.pressDownArrow();
+      await tester.pressRightArrow();
+
+      // Select upstream towards the cancelled action tag
+      await tester.pressShiftLeftArrow();
+      await tester.pressShiftUpArrow();
```
matthew-carroll, line 672: If possible, we'd like to formulate tests in a way that we can naturally declare expectations that are met in the test.

If there's no way to formulate this test with an expectation that makes sense, then at a minimum we need a comment at the bottom of this test explaining that reaching the comment without throwing an exception means the test passes. Please also briefly describe the bug that lead to this test, and mention the issue ticket number in the comment.

super_editor/test/super_editor/text_entry/tagging/action_tags_test.dart
KevinBrendel: Added a longer name to express it more explicitly.

super_editor/test/super_editor/text_entry/tagging/action_tags_test.dart
KevinBrendel: Added expectLater with returnsNormally. This does not appear to change the behavior of the test, but makes it more explicit. Also added comment and link.

KevinBrendel:
Not sure why the netlify deploys are failing now. Is it because this PR's base is now outdated vs the main branch?

matthew-carroll:
I tried to edit this PR directly so you wouldn't need to push another change, but it looks like I can't do that.

Please replace the test description with this:

```
"can find tag that surrounds the extent position when the selection is expanded"
```

Please replace the description at the bottom of the test with this:

```
// If we reach the end without exception, then ActionTagComposingReaction did not blow up due to the base or extent
// position, and type of content at those positions.
//
// Original bug: https://github.com/superlistapp/super_editor/pull/2201
```

KevinBrendel:
@matthew-carroll Done.

matthew-carroll (approved):
LGTM

matthew-carroll:

PR #2207, opened 2024-07-31
[Infrastructure] Fix widget tester tapping on block component (Resolves #2197)

### LINKED ISSUE #2197: [Infrastructure] - Fix widget tester tapping on block nodes downstream
I wrote a test where I tried to tap on the downstream edge of a fake image block node. However, it resulted in placing the caret on the upstream edge.

I've reproduced the exact test below.

Correct the widget tester extension so that it places the caret on the desired side of a given block node.

```dart
testWidgetsOnMac("undo when typing after an image", (tester) async {
        // A reported bug found that when inserting a paragraph after an image, typing some
        // text, and then undo'ing the text, the paragraph's text duplicates during the
        // undo operation: https://github.com/superlistapp/super_editor/issues/2164
        // TODO: The root cause of this problem was mutability of DocumentNode's. Delete this test after completing: https://github.com/superlistapp/super_editor/issues/2166
        final testContext = await tester
            .createDocument() //
            .withCustomContent(MutableDocument(
              nodes: [
                ImageNode(id: "1", imageUrl: "https://fakeimage.com/myimage.png"),
              ],
            ))
            .withComponentBuilders([
              const FakeImageComponentBuilder(size: Size(1000, 400)),
              ...defaultComponentBuilders,
            ])
            .autoFocus(true)
            .pump();

        // Place the caret on the trailing edge of the image.
        testContext.editor.execute([
          const ChangeSelectionRequest(
            DocumentSelection.collapsed(
              position: DocumentPosition(nodeId: "1", nodePosition: UpstreamDownstreamNodePosition.downstream()),
            ),
            SelectionChangeType.placeCaret,
            SelectionReason.userInteraction,
          ),
        ]);
        await tester.pump();

        // TODO: fix block node selection - we requested downstream and it selected upstream
        // await tester.tapAtDocumentPosition(
        //   const DocumentPosition(nodeId: "1", nodePosition: UpstreamDownstreamNodePosition.downstream()),
        // );

        // Press enter to insert a new paragraph.
        await tester.pressEnter();

        // Ensure we inserted a paragraph.
        expect(testContext.document.nodeCount, 2);
        expect(testContext.document.getNodeAt(0), isA<ImageNode>());
        expect(testContext.document.getNodeAt(1), isA<TextNode>());

        // Type some text.
        await tester.pressKey(LogicalKeyboardKey.keyA);

        // Wait long enough to avoid combining actions into a single transaction.
        await tester.pump(const Duration(seconds: 2));

        // Type more text.
        await tester.pressKey(LogicalKeyboardKey.keyB);

        // Ensure we inserted the text.
        expect((testContext.document.getNodeAt(1) as TextNode).text.text, "ab");

        // Undo the text insertion.
        // TODO: remove `tester` reference after updating flutter_test_robots
        await tester.pressCmdZ(tester);

        // Ensure that the paragraph removed the last entered character.
        expect((testContext.document.getNodeAt(1) as TextNode).text.text, "a");
      });
```

angelosilvestre:
[Infrastructure] Fix widget tester tapping on block component. Resolves #2197

Trying to tap on the downstream position of a block component is always resulting in an upstream selection.

The issue is that the computation to determine the offset to tap relies on `getRectForPosition`. For block components, this always return a rect of the entire component, as is documented in the method:

```dart
/// Returns a [Rect] that bounds this entire box component.
///
/// The behavior of this method is the same, regardless of whether the given
/// [nodePosition] is `upstream` or `downstream`.
@override
Rect getRectForPosition(NodePosition nodePosition) {
   ///
}
```

This PR changes the `tapAtDocumentPosition` to handle block components differently, by computing its edge position and returning a `Rect` of half of the component's width.

### DIFF
--- a/super_editor/lib/src/test/super_editor_test/supereditor_robot.dart
+++ b/super_editor/lib/src/test/super_editor_test/supereditor_robot.dart
@@ -164,7 +164,7 @@ extension SuperEditorRobot on WidgetTester {
   /// {@macro supereditor_finder}
   Future<void> tapAtDocumentPosition(DocumentPosition position, [Finder? superEditorFinder]) async {
     final documentLayout = _findDocumentLayout(superEditorFinder);
-    final positionRectInDoc = documentLayout.getRectForPosition(position)!;
+    final positionRectInDoc = _getRectForDocumentPosition(position, documentLayout, superEditorFinder);
     final globalTapOffset = documentLayout.getAncestorOffsetFromDocumentOffset(positionRectInDoc.center);
 
     await tapAt(globalTapOffset);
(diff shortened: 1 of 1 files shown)


super_editor/lib/src/test/super_editor_test/supereditor_robot.dart
```diff
@@ -436,17 +436,67 @@ extension SuperEditorRobot on WidgetTester {
...
+          // Subtract half of the width to make it start from the center.
+          : Rect.fromLTWH(
+              edge.left - componentBox.size.width / 2,
+              edge.top,
+              componentBox.size.width / 2,
+              componentBox.size.height,
+            );
+
+      // Translate the rect to global coordinates.
+      final documentLayoutElement = _findDocumentLayoutElement(superEditorFinder);
```
matthew-carroll, line 474: I'm thinking that the need to access the document element isn't great. Does the `DocumentLayout` interface not offer any way to accomplish this goal as-is?

matthew-carroll (commented):
I left a comment. Also, to ensure this works, can you adjust whichever test(s) I recently committed where I wanted to use this, but I couldn't?

super_editor/lib/src/test/super_editor_test/supereditor_robot.dart
angelosilvestre: I didn't find any method to allow that. The `Rect` returned by `Component.getEdgeForPosition` is in component space and I didn't find any other method to convert it to document space.

angelosilvestre:
@matthew-carroll I search on the current main but I didn't find any tests like the one posted in the issue ticket, where we manually place the selection using `UpstreamDownstreamNodePosition.downstream()`.

Should I include that test?

matthew-carroll (approved):
LGTM - We can leave this as-is for now so we can move forward, but this block position calculation should probably be a part of the real layout and component system, and that real system should also make it possible to get to global coordinates.

PR #2208, opened 2024-08-02
[Quill] - Moved delta doc matcher into source code and exposed publicly so other projects can use it.

matthew-carroll:
[Quill] - Moved delta doc matcher into source code and exposed publicly so other projects can use it.


angelosilvestre (commented):
LGTM

PR #2215, opened 2024-08-04
[SuperEditor] - Undo/redo can now be disabled (they are disabled by default) (Resolves #2214)

### LINKED ISSUE #2214: [SuperEditor] - Make it possible to disable undo/redo
Currently, undo/redo remembers the user's full edit history, which creates the opportunity for memory leaks.

Make it possible to disable undo/redo, and for now, disable it by default.

matthew-carroll:
[SuperEditor] - Undo/redo can now be disabled (they are disabled by default) (Resolves #2214)

The `Editor` now has a property called `isHistoryEnabled`. That property is respected inside the `Editor` - `undo()` and `redo()` bail if it's `false`. Also, the locations that would otherwise add to the history only do that when the property is `true`.

I made the property a constant on each `Editor` instance because attempting to change that value over time seems more likely to create bugs than to enable desired behavior.

I also updated the test configuration system to take a value for `isHistoryEnabled`.


angelosilvestre (approved):
LGTM

PR #2216, opened 2024-08-05
[SuperEditor][Android][iOS] - Add tooling to more easily create keyboard toolbar experiences (Resolves #2209)

### LINKED ISSUE #2209: [SuperEditor][Android][iOS] - Add tooling to more easily create keyboard toolbar experiences
Editing text on mobile typically requires a toolbar that sits on top of the keyboard. For more sophisticated editors, it's also common to switch between the software keyboard and a panel with buttons. The panel with buttons takes larger-scale actions on the document - things that don't always apply when editing, and things that might need more room than a little icon on the toolbar.

Let's evaluate what kind of tooling we could provide that makes it faster and easier for apps to setup mobile keyboard editing toolbars and panels.

@angelosilvestre let's begin with ~1hr of investigation into the value we can provide here, the way others are already doing this, how we can/should interact with the keyboard, etc. Then let's use the remaining time to work on the tools.

angelosilvestre:
[SuperEditor][Android][iOS] - Add tooling to more easily create keyboard toolbar experiences. Resolves #2209

This PR introduces the `KeyboardPanelScaffold` widget to make it easier to implement a panel behind keyboard widget. This is common for chat apps, where the user can switch between the regular software keyboard panel and an emoji panel.

This widget does not enforce any particular styling on the panel. I named it "keyboard panel" because the lack of a better name, maybe we could find a better name for it.

I modified the existing `PanelBehindKeyboardDemo` demo to use this new widget and also modified the demo to add content to the keyboard panel. The demo mentions it's only for Android, but I didn't find any issues on iOS.

One issue that I found is that our `SoftwareKeyboardController.close` closes the IME connection instead of just closing the keyboard. This might cause problems. We should probably be invoking `TextInput.hide` instead.

https://github.com/user-attachments/assets/0c0d2cd7-857f-43b1-879e-cf1c7d04499b

[android_panel.webm](https://github.com/user-attachments/assets/e239902c-7b76-4d42-ab69-6d4a17323d1b)


super_editor/lib/src/infrastructure/keyboard_panel_scaffold.dart
```diff
@@ -0,0 +1,250 @@
...
+/// and an emoji panel.
+///
+/// The widget returned by [keyboardPanelBuilder] is positioned at the bottom of the screen,
+/// with its height constrained to be equal to the software keyboard height.
+///
+/// The widget returned by [aboveKeyboardBuilder] is positioned above the keyboard panel, when
+/// visible, or above the software keyboard, when visible. If neither the keyboard panel nor
+/// the software keyboard are visible, the widget is positioned at the bottom of the screen.
+///
+/// The widget returned by [contentBuilder] is positioned at the top, using all the available
```
matthew-carroll, line 16: Please describe `contentBuilder` a bit more. It's a bit confusing to read "positioned at the top" right after reading that "[aboveKeyboardBuilder] is positioned above the keyboard panel"

super_editor/lib/src/infrastructure/keyboard_panel_scaffold.dart
```diff
@@ -0,0 +1,250 @@
...
+/// height.
+///
+/// Use the [controller] to show/hide the keyboard panel and software keyboard.
+///
+/// It is required that [Scaffold.resizeToAvoidBottomInset] is set to `false`.
```
matthew-carroll, line 21: Please elaborate on this requirement. Which `Scaffold`? And why is it required? What happens if it's `true`?

super_editor/lib/src/infrastructure/keyboard_panel_scaffold.dart
angelosilvestre: Updated explaining the it's positioned above the above-keyboard panel. Which kind of other information do you think I should add here?

super_editor/lib/src/infrastructure/keyboard_panel_scaffold.dart
```diff
@@ -0,0 +1,252 @@
...
+///
+/// The widget returned by [contentBuilder] is positioned above the above-keyboard panel,
+/// using all the remaining height.
+///
+/// Use the [controller] to show/hide the keyboard panel and software keyboard.
+///
+/// It is required that the enclosing [Scaffold] has `resizeToAvoidBottomInset` set to `false`,
+/// otherwise we can't get the software keyboard height to size the keyboard panel. If
+/// `resizeToAvoidBottomInset` is set to `true`, the panel won't be displayed.
+class KeyboardPanelScaffold extends StatefulWidget {
```
matthew-carroll, line 24: I thought I asked this before, but I don't see the comment, so maybe I didn't...

This approach looks like it requires that this widget be placed at a location that surrounds all visible area, right?

If so, I'm wondering how one handles situations where they're adding a text field or an editor into a single card within a list, or the primary content area on a tablet that sits next to a docked nav drawer, etc.

In other words, when the area of the code with the text field or editor doesn't naturally have access to the whole screen, what do we expect developers to do?

On the one hand, we could tell developers to go place this scaffold higher in the tree somewhere, but on the surface, that seems like an extra burden.

What would happen if this widget operated in the application overlay? Can we make it so that the control over keyboard toolbars and panels can happen from any given place in the UI, even when that place is just a single list item?

angelosilvestre: I pushed an update with an alternative implementation that uses an `OverlayPortal`. Can you take a look?

https://github.com/user-attachments/assets/dcd801cf-8f01-44d4-894c-1811927bb872

matthew-carroll: What's the purpose of the button you added in this video? And what's happening when the caret is placed but there's no toolbar above the keyboard until you later press the button?

angelosilvestre: This is just to show that the above-keyboard panel can be enabled/disabled.

matthew-carroll:
Some thoughts from trying to use this in a chat editor use-case:

* We should have an auto-display property such that, when true, the toolbar content is displayed whenever the editor has a selection.
* The content currently flows behind the toolbar, but this hides some of the content. This is especially a problem when the editor is a small, intrinsic height editor at the bottom of the screen. Essentially the whole editor disappears behind the toolbar.
  * We should have a widget property option to decide whether to expand the content behind the toolbar area or not.
  * This will be a little tricky to implement because the toolbar is in the overlay, and we need to add space below the content that's equal to the intrinsic height of the toolbar. Maybe a two frame sequence is needed - one frame that displayed the toolbar offscreen and measures it, and then on the second frame, once its measured, both the toolbar and the content are displayed (to prevent a flash where the editor is in the wrong place).
* Possible bug: In the app where I'm trying this out, sometimes I hot reload and the toolbar disappears. Not sure if that's a problem in the scaffold or this app.

angelosilvestre:
> The content currently flows behind the toolbar, but this hides some of the content. This is especially a problem when the editor is a small, intrinsic height editor at the bottom of the screen. Essentially the whole editor disappears behind the toolbar.

@matthew-carroll I added a widget to added a widget to address this issue. Updated the chat demo to use it.

[keyboard_toolbar.webm](https://github.com/user-attachments/assets/ec341838-ed2a-4734-8606-1b6c075ff170)

matthew-carroll:
@angelosilvestre I saw your commit saying that you added an option to auto show the toolbar above the keyboard, but I don't actually see any public options added in that commit.

How does one say whether they want the toolbar to be displayed automatically or not?

matthew-carroll:
It looks like a good number of tests are failing.

matthew-carroll:
It looks like a good number of tests are failing.

matthew-carroll:
I added a KeyboardScaffoldSafeArea in a location in the widget tree where I think it should make sense, but it didn't have any impact on the content location. The content still moves down behind the keyboard panel.

Is there anything I should know about how to use this?

matthew-carroll:
@angelosilvestre can you take a look at the modifications I've made? Before ClickUp can use this, we'll need to merge it in and cut a new release. So it would be great if we could finalize this pretty quickly. Let me know if you see anything concerning in here.

super_editor/example/lib/demos/experiments/demo_panel_behind_keyboard.dart
```diff
@@ -169,155 +161,225 @@ class _PanelBehindKeyboardDemoState extends State<PanelBehindKeyboardDemo> {
...
+    _isKeyboardPanelVisible = isKeyboardPanelVisible;
+
+    return SuperEditor(
+      focusNode: _focusNode,
+      editor: _editor,
+      softwareKeyboardController: _keyboardController,
+      selectionPolicies: const SuperEditorSelectionPolicies(
+        clearSelectionWhenEditorLosesFocus: false,
+        // Currently, closing the software keyboard causes the IME connection to close.
+        clearSelectionWhenImeConnectionCloses: false,
```
angelosilvestre, line 190: Is this still true after your latest changes?

angelosilvestre (commented):
@matthew-carroll  LGTM

super_editor/example/lib/demos/experiments/demo_panel_behind_keyboard.dart
matthew-carroll: Did you see any area where it changed? I didn't alter any existing tests related to that, and I don't recall altering any related code.

matthew-carroll: I'm going to merge this. Please let me know if you have any evidence that this behavior changed.

angelosilvestre: I thought we have talked about that, but I might be mistaken.

matthew-carroll:
@angelosilvestre can you do a final review of this? I think we've got it working in ClickUp. I'm sure it'll evolve, but let's get this version merged and released.

super_editor/test/super_editor/supereditor_test_tools.dart
```diff
@@ -465,24 +488,39 @@ class TestSuperEditorConfigurator {
...
+                    ),
                   ),
-                ),
-              )
-            : null,
-        body: superEditor,
+                )
+              : null,
+          body: superEditor,
+          resizeToAvoidBottomInset: false,
```
angelosilvestre, line 505: It would be a good idea to leave a comment here explaining why we are always setting `resizeToAvoidBottomInset` to `false`.

super_editor/test/infrastructure/keyboard_panel_scaffold_test.dart
```diff
@@ -0,0 +1,781 @@
...
+        // Ensure the toolbar is above the minimized keyboard area.
+        expect(
+          tester.getBottomLeft(find.byKey(_aboveKeyboardToolbarKey)).dy,
+          screenHeight - _minimizedIPadKeyboardHeight,
+        );
+      });
+    });
+
+    group('Android tablets >', () {
+      testWidgetsOnIPad('shows panel when keyboard is docked', (tester) async {
```
angelosilvestre, line 397: Should this be `testWidgetsOnAndroidTablet` ?

angelosilvestre (commented):
LGTM with a few comments.

PR #2227, opened 2024-08-08
[SuperEditor][SuperReader] Fix list item styling without stylesheet (Resolves #2224)

### LINKED ISSUE #2224: [SuperEditor][SuperReader] - Bug: List item view models erase their styling if it's not in the stylesheet
We have a bug in the interaction between the stylesheet and list items. I observed this issue specifically with unordered list items, but it probably applies to aspects of ordered list items, too.

Unordered list item view models have a dot color, dot size, and dot shape. These properties are set by the view model constructor, but they can be changed by the stylesheet.

The problem is that, currently, if the stylesheet doesn't include values for these properties, then these properties are reset during the styling phase.

Reproduction:
- Adjust the unordered component creation so that the component is created with a non-standard dot size, like `Size(16, 16)` (the standard is `Size(4, 4)`.
- Remove any such dot size specification from the editor's stylesheet.
- Run the editor and inspect the dot size.

Expected result:
The dot size is 16x16

Actual result:
The dot size is 4x4

My guess is that the fix is simply to use the existing view model values when the stylesheet doesn't provide a value. Check other view models to see if that's what they do - but I assume that it is.

angelosilvestre:
[SuperEditor][SuperReader] Fix list item styling without stylesheet. Resolves #2224

When the stylesheet doesn't include values for the the list items style properties, they are erased in the style phase. This means that an app can't provide the style directly to the view models.

The issue is that we are always re-assigning the default values if they aren't present in the stylesheet.

This PR fixes the issue by maintaining the styles already present in the view model when the stylesheet doesn't provide them.

I renamed the existing goldens, appending "_stylesheet" and added new test appending "_component_builder".


super_editor/lib/src/default_editor/list_items.dart
```diff
@@ -424,6 +424,20 @@ class ListItemDotStyle {
 
   @override
   int get hashCode => super.hashCode ^ color.hashCode ^ shape.hashCode ^ size.hashCode;
+
+  /// Returns a copy of this [ListItemDotStyle] with optional new values
+  /// for [color], [shape], and [size].
+  ListItemDotStyle copyWith({
```
matthew-carroll, line 430: Please put `copyWith()` methods above `==` and `hashCode`.

matthew-carroll (approved):
LGTM

PR #2238, opened 2024-08-13
[SuperEditor][SuperTextField][iOS] Allow animating caret blink (Resolves #2099)

### LINKED ISSUE #2099: [SuperEditor][SuperTextField][iOS] - Make it possible to animate the iOS caret
The following issue mentioned wanting an animated caret for iOS: https://github.com/superlistapp/super_editor/issues/2080

I believe we had that at one point but we switched to an on/off blinking caret to help Superlist with performance.

Bring back the traditional iOS caret fading animation, but make it configurable so that users can turn it off.

angelosilvestre:
[SuperEditor][SuperTextField][iOS] Allow animating caret blink. Resolves #2099 

Currently, our blinking caret has an on/off visibility, it doesn't animates the visibility changes. 

This PR adds a fade in/out transition when the caret switches from visible to invisible. 

Some problems that I found:

- Although we have an option to use a `Ticker` to drive the visibility change, we are setting the `_ticker` to `null` when `stopBlinking` is called. It means that once the user types or places the caret somewhere, which causes `stopBlinking` to be called to momentary stop blinking, we are switching to the timer mode:

https://github.com/superlistapp/super_editor/blob/4c7472b0e43460f27b7a2809ba704286dcec7b21/super_text_layout/lib/src/infrastructure/blink_controller.dart#L90-L96

https://github.com/superlistapp/super_editor/blob/4c7472b0e43460f27b7a2809ba704286dcec7b21/super_text_layout/lib/src/infrastructure/blink_controller.dart#L68-L88

- Fixing the previous issue creates an issue in some tests that check if the caret is blinking. Such tests were working before because we were always using the timer mode. When switching to ticker mode, there are some tests failing due to `pumpAndSettleTimeout`. The timeout happens because two methods used in those tests (`tester.ime.backspace` and `tester.doubleTapInParagraph`) internally call `pumpAndSettle`.

### Open questions:

- How should we expose the configuration to animate the caret blink? Via `SuperEditorIosControlsController` ?
- Should we add parameters to `tester.ime.backspace` and `tester.doubleTapInParagraph` to avoid calling `pumpAndSettle`?

### DIFF
--- a/super_text_layout/lib/src/infrastructure/blink_controller.dart
+++ b/super_text_layout/lib/src/infrastructure/blink_controller.dart
@@ -15,10 +15,18 @@ class BlinkController with ChangeNotifier {
   @visibleForTesting
   static bool indeterminateAnimationsEnabled = true;
 
+  /// Creates a [BlinkController] that uses a ticker to switch between visible
+  /// and invisible.
+  ///
+  /// If [animate] is `true`, the caret animates its [opacity] when switching
+  /// between visible and invisible. If [animate] is `false`, the caret
+  /// switches between fully opaque and fully transparent.
   BlinkController({
     required TickerProvider tickerProvider,
     Duration flashPeriod = const Duration(milliseconds: 500),
-  }) : _flashPeriod = flashPeriod {
+    bool animate = false,
+  })  : _flashPeriod = flashPeriod,
+        _animate = animate {
     _ticker = tickerProvider.createTicker(_onTick);
   }
 
@@ -63,7 +71,18 @@ class BlinkController with ChangeNotifier {
   }
 
   bool _isVisible = true;
-  double get opacity => _isVisible ? 1.0 : 0.0;
+
+  double _opacity = 1.0;
+  double get opacity => _opacity;
+
+  /// Whether or not the caret should animate its opacity.
+  bool _animate = false;
+
+  /// Whether or not the caret should maintain its opacity until the next blink.
+  ///
+  /// For example, when the user places the caret, we want it to be visible with
+  /// full opacity until the caret switches to invisible.
+  bool _remainOpaqueUntilNextBlink = false;
 
   void startBlinking() {
     if (!indeterminateAnimationsEnabled) {
@@ -73,6 +92,9 @@ class BlinkController with ChangeNotifier {
     }
 
     if (_ticker != null) {
+      // Don't animate the opacity until the flash period has elapsed.
+      _remainOpaqueUntilNextBlink = true;
+
       // We're using a Ticker to blink. Restart it.
       _ticker!
         ..stop()
@@ -88,12 +110,17 @@ class BlinkController with ChangeNotifier {
   }
 
   void stopBlinking() {
-    _isVisible = true; // If we're not blinking then we need to be visible
+    // If we're not blinking then we need to be visible with full opacity.
+    _isVisible = true;
+    _opacity = 1.0;
+
+    // Keep the caret opaque until the flash period has elapsed, when it
+    // starts to fade out.
+    _remainOpaqueUntilNextBlink = true;
 
     if (_ticker != null) {
       // We're using a Ticker to blink. Stop it.
       _ticker?.stop();
-      _ticker = null;
     } else {
       // We're using a Timer to blink. Stop it.
       _timer?.cancel();
@@ -118,6 +145,15 @@ class BlinkController with ChangeNotifier {
   }
 
   void _onTick(Duration elapsedTime) {
+    final fadeAnimationDuration = _flashPeriod ~/ 2;
+
+    if (isBlinking && _animate && !_remainOpaqueUntilNextBlink) {
+      final percentage =
+          ((elapsedTime - _lastBlinkTime).inMilliseconds / fadeAnimationDuration.inMilliseconds).clamp(0.0, 1.0);
+      _opacity = _isVisible ? percentage : 1 - percentage;
+      notifyListeners();
+    }
+
     if (elapsedTime - _lastBlinkTime >= _flashPeriod) {
       _blink();
       _lastBlinkTime = elapsedTime;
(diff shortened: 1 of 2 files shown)


angelosilvestre:
@matthew-carroll Can you take a look at the questions mentioned in the PR description?

super_text_layout/lib/src/infrastructure/blink_controller.dart
```diff
@@ -15,16 +15,27 @@ class BlinkController with ChangeNotifier {
   @visibleForTesting
   static bool indeterminateAnimationsEnabled = true;
 
+  /// Creates a [BlinkController] that uses a ticker to switch between visible
+  /// and invisible.
+  ///
+  /// If [animate] is `true`, the caret animates its [opacity] when switching
+  /// between visible and invisible. If [animate] is `false`, the caret
+  /// switches between fully opaque and fully transparent.
```
matthew-carroll, line 23: Can you also add a point here about performance? The reason we originally got rid of animated blinking was to reduce Flutter re-renders.

super_text_layout/lib/src/infrastructure/blink_controller.dart
```diff
@@ -15,16 +15,27 @@ class BlinkController with ChangeNotifier {
...
   BlinkController({
     required TickerProvider tickerProvider,
     Duration flashPeriod = const Duration(milliseconds: 500),
-  }) : _flashPeriod = flashPeriod {
+    bool animate = false,
+    Duration fadeDuration = const Duration(milliseconds: 200),
+  })  : _flashPeriod = flashPeriod,
+        _fadeDuration = fadeDuration,
```
matthew-carroll, line 30: How can the fade duration be independent of the flash period?

super_text_layout/lib/src/infrastructure/blink_controller.dart
```diff
@@ -62,8 +76,24 @@ class BlinkController with ChangeNotifier {
     notifyListeners();
   }
 
+  /// Whether or not the caret is currently visible.
+  ///
+  /// It's possible that [isVisible] is `false` while [opacity] is greater than `0.0`
```
matthew-carroll, line 81: Rather than sayign something is possible, please describe under which circumstances something happens or doesn't happen.

super_text_layout/lib/src/infrastructure/blink_controller.dart
```diff
@@ -62,8 +76,24 @@ class BlinkController with ChangeNotifier {
...
+  /// when the caret is fading out.
+  bool get isVisible => _isVisible;
   bool _isVisible = true;
-  double get opacity => _isVisible ? 1.0 : 0.0;
+
+  double get opacity => _opacity;
+  double _opacity = 1.0;
+
+  /// Whether or not the caret should animate its opacity.
```
matthew-carroll, line 89: Please also specify the alternative. It's possible a developer reads this and then asks "what does it mean to not animate the opacity?" - let's make the distinction between fading blinking vs on/off blinking.

super_text_layout/lib/src/infrastructure/blink_controller.dart
```diff
@@ -62,8 +76,24 @@ class BlinkController with ChangeNotifier {
...
+  bool _animate = false;
+
+  /// Whether or not the caret should maintain its opacity until the next blink.
```
matthew-carroll, line 92: This Dart Doc comment makes it sound like this property is a permanent configuration on this controller, but in fact it looks like a momentary flag that's switched on and off to capture the fact that we're at a specific moment in time. Please try to replace this description with one that describes what this property is tracking, including when we expect it to flip from true to false, or false to true. Also, the name of this should probably start with `_isSomething...`, if it represents a temporary condition.

super_text_layout/lib/src/infrastructure/blink_controller.dart
```diff
@@ -118,6 +156,12 @@ class BlinkController with ChangeNotifier {
   }
 
   void _onTick(Duration elapsedTime) {
+    if (isBlinking && _animate && !_remainOpaqueUntilNextBlink) {
+      final percentage = ((elapsedTime - _lastBlinkTime).inMilliseconds / _fadeDuration.inMilliseconds).clamp(0.0, 1.0);
+      _opacity = _isVisible ? percentage : 1 - percentage;
```
matthew-carroll, line 161: Please add a comment to clarify the meaning of `_isVisible` in this calculation, because you're actually using it as a proxy to decide whether we're "fading in" or "fading out", and that's not immediately obvious.

matthew-carroll:
> Should we add parameters to tester.ime.backspace and tester.doubleTapInParagraph to avoid calling pumpAndSettle?

My initial thought is, no. I think we should adjust the Super Editor side to play well with common test behaviors, rather than change the test behaviors to fit Super Editor.

> How should we expose the configuration to animate the caret blink? Via SuperEditorIosControlsController ?

I wanna make sure I understand this question - you exposed the property from the controller already, right?

super_text_layout/lib/src/infrastructure/blink_controller.dart
angelosilvestre: The caret isn't constantly fading in and out. It stays visible or invisible during the flash period, then quickly fades in/out.

Should we just use half of the flash period?

super_text_layout/lib/src/infrastructure/blink_controller.dart
angelosilvestre: Was the original performance issue really about re-renders? Or was it related to repaints? I'm wondering if using e `RepaintBoundary` in our overlayers would make a difference.

super_text_layout/lib/src/infrastructure/blink_controller.dart
angelosilvestre: Updated.

super_text_layout/lib/src/infrastructure/blink_controller.dart
angelosilvestre: Updates.

super_text_layout/lib/src/infrastructure/blink_controller.dart
angelosilvestre: I tried to improve it a little bit.

super_text_layout/lib/src/infrastructure/blink_controller.dart
angelosilvestre: Updated.

angelosilvestre:
> My initial thought is, no. I think we should adjust the Super Editor side to play well with common test behaviors, rather than change the test behaviors to fit Super Editor.

But how should we do that, if the ticker is constantly scheduling frames?

> I wanna make sure I understand this question - you exposed the property from the controller already, right?

I exposed from the controller, but it still isn't configurable in a way that apps can enable or disable it. The `BlinkController` is instantiated by us.

PR #2239, opened 2024-08-14
Ignore unused_element_parameter for unused parameters

srawlins:
In https://github.com/dart-lang/sdk/issues/49025, we are splitting the `unused_element` code, so that the warning reported is `unused_element_parameter`.

In this PR, I change each ignored case to ignore each warning name, to make the code forwards-compatible.


matthew-carroll (approved):
LGTM

matthew-carroll:

Issue #2254, opened 2024-08-18
[BUG] - [DemoApp] Toggling debug info crashes the editor


angelosilvestre:
**Package Version**
Latest main

**To Reproduce**
1. Run the example app
2. Click on the debug visuals toggle

**Actual behavior**
The editor crashes with the following exception:

```console
Restarted application in 333ms.

════════ Exception caught by widgets library ═══════════════════════════════════
The following assertion was thrown building AnimatedBuilder(listenable: ValueNotifier<TextInputConnection?>#99375(null), state: _AnimatedState#b3bd8):
A RenderViewport expected a child of type RenderSliver but received a child of type RenderStack.
RenderObjects expect specific types of children because they coordinate with their children during layout and paint. For example, a RenderSliver cannot be the child of a RenderBox because a RenderSliver does not understand the RenderBox layout protocol.

The RenderViewport that expected a RenderSliver child was created by: Viewport ← IgnorePointer-[GlobalKey#552d6] ← Semantics ← Listener ← _GestureSemantics ← RawGestureDetector-[LabeledGlobalKey<RawGestureDetectorState>#66afa] ← Listener ← _ScrollableScope ← _ScrollSemantics-[GlobalKey#7f3c6] ← NotificationListener<ScrollMetricsNotification> ← Scrollable ← CustomScrollView ← ⋯

The RenderStack that did not match the expected child type was created by: Stack ← AnimatedBuilder ← SuperEditorImeDebugVisuals ← SuperEditorImeInteractor ← Viewport ← IgnorePointer-[GlobalKey#552d6] ← Semantics ← Listener ← _GestureSemantics ← RawGestureDetector-[LabeledGlobalKey<RawGestureDetectorState>#66afa] ← Listener ← _ScrollableScope ← ⋯

The relevant error-causing widget was:
    AnimatedBuilder AnimatedBuilder:file:///Users/angelosilvestre/dev/super_editor/super_editor/lib/src/default_editor/debug_visualization.dart:108:12

When the exception was thrown, this was the stack:
#0      ContainerRenderObjectMixin.debugValidateChild.<anonymous closure> (package:flutter/src/rendering/object.dart:4239:9)
#1      ContainerRenderObjectMixin.debugValidateChild (package:flutter/src/rendering/object.dart:4266:6)
#2      MultiChildRenderObjectElement.insertRenderObjectChild (package:flutter/src/widgets/framework.dart:6973:25)
[...]

PR #2255, opened 2024-08-19
Fix toggling debug info crashing the editor

knopp:
Fixes https://github.com/superlistapp/super_editor/issues/2254

- Changes `Stack` in `SuperEditorImeDebugVisuals` to `SliverHybridStack` because the render object now lives in a sliver
- Implements `forgetChild` inside `ContentLayersElement`. This is necessary because the `content` of `ContentLayersElement` is being reparented using `GlobalKey` so it needs to be excluded from `visitChildren` when unmounting.

### DIFF
--- a/super_editor/lib/src/infrastructure/content_layers.dart
+++ b/super_editor/lib/src/infrastructure/content_layers.dart
@@ -1,3 +1,5 @@
+import 'dart:collection';
+
 import 'package:flutter/rendering.dart';
 import 'package:flutter/scheduler.dart';
 import 'package:flutter/widgets.dart';
@@ -108,6 +110,8 @@ class ContentLayersElement extends RenderObjectElement {
   List<Element> _underlays = <Element>[];
   Element? _content;
   List<Element> _overlays = <Element>[];
+
+  final Set<Element> _forgottenChildren = HashSet<Element>();
 
   @override
   ContentLayers get widget => super.widget as ContentLayers;
@@ -331,6 +335,7 @@ class ContentLayersElement extends RenderObjectElement {
     assert(!debugChildrenHaveDuplicateKeys(widget, [newContent]));
 
     _content = updateChild(_content, newContent, _contentSlot);
+    _forgottenChildren.clear();
   }
 
   @override
@@ -370,7 +375,6 @@ class ContentLayersElement extends RenderObjectElement {
 
   @override
   void removeRenderObjectChild(RenderObject child, Object? slot) {
-    assert(child is RenderBox);
     assert(child.parent == renderObject);
     assert(slot != null);
     assert(_isContentLayersSlot(slot!), "Invalid ContentLayers slot: $slot");
@@ -379,8 +383,14 @@ class ContentLayersElement extends RenderObjectElement {
   }
 
   @override
+  void forgetChild(Element child) {
+    _forgottenChildren.add(child);
+    super.forgetChild(child);
+  }
+
+  @override
   void visitChildren(ElementVisitor visitor) {
-    if (_content != null) {
+    if (_content != null && !_forgottenChildren.contains(_content)) {
       visitor(_content!);
     }
 
@@ -395,11 +405,15 @@ class ContentLayersElement extends RenderObjectElement {
     // ignore: invalid_use_of_protected_member
     if (!WidgetsBinding.instance.locked) {
       for (final Element child in _underlays) {
-        visitor(child);
+        if (!_forgottenChildren.contains(child)) {
+          visitor(child);
+        }
       }
 
       for (final Element child in _overlays) {
-        visitor(child);
+        if (!_forgottenChildren.contains(child)) {
+          visitor(child);
+        }
       }
     }
   }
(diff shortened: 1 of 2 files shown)


super_editor/lib/src/infrastructure/content_layers.dart
```diff
@@ -313,11 +317,11 @@ class ContentLayersElement extends RenderObjectElement {
   void _temporarilyForgetLayers() {
     contentLayersLog.finer("ContentLayersElement - temporarily forgetting layers");
     for (final underlay in _underlays) {
-      forgetChild(underlay);
+      super.forgetChild(underlay);
```
knopp, line 320: I don't exactly know what the purpose of calling `forgetChild` here outside of normal framework flow is. I changed it to `super.forgetChild` to preserve the original behavior but it doesn't feel right either way.

matthew-carroll: Please add a comment to both of the `super.forgetChild()` calls that explains why you're not calling the new implementation you added.

super_editor/lib/src/infrastructure/content_layers.dart
```diff
@@ -370,17 +375,22 @@ class ContentLayersElement extends RenderObjectElement {
...
+  @override
+  void forgetChild(Element child) {
+    _forgottenChildren.add(child);
+    super.forgetChild(child);
+  }
+
   @override
   void visitChildren(ElementVisitor visitor) {
-    if (_content != null) {
+    if (_content != null && !_forgottenChildren.contains(_content)) {
```
matthew-carroll, line 393: Was this if-statement a pre-existing problem? Is the framework that's forgetting the content? Because up above the only thing we forget are the layers, not the content, right?

super_editor/lib/src/infrastructure/content_layers.dart
```diff
@@ -109,6 +111,8 @@ class ContentLayersElement extends RenderObjectElement {
   Element? _content;
   List<Element> _overlays = <Element>[];
 
+  final Set<Element> _forgottenChildren = HashSet<Element>();
```
matthew-carroll, line 114: Please add a comment that explains why we're tracking forgotten children and the critical moments that maintaining developers should be aware of with this tracking set.

super_editor/lib/src/infrastructure/content_layers.dart
```diff
@@ -331,6 +335,7 @@ class ContentLayersElement extends RenderObjectElement {
     assert(!debugChildrenHaveDuplicateKeys(widget, [newContent]));
 
     _content = updateChild(_content, newContent, _contentSlot);
+    _forgottenChildren.clear();
```
matthew-carroll, line 338: Please add a comment as to why this is the right moment to clear the tracking set (I can never remember the order of operations in Flutter's build process)

super_editor/lib/src/infrastructure/content_layers.dart
knopp: `forgetChild` basically seems to mean "exclude child from visitChildren until next update()". Frameworks calls `forgetChild` for elements that are being reparented to avoid unmounting them when unmounting the parent element for example.

So I think it was a pre-existing problem, just nothing triggered it.

matthew-carroll:
@knopp are you still working on this, or do you think you responded to all my comments? Whenever you do, please re-request a review so that I'm notified.

knopp:
I added missing comments.

matthew-carroll:
@knopp I don't see any code comments in the change log

knopp:
> @knopp I don't see any code comments in the change log

That's because I forgot to actually push the changes :) Sorry about it. Should be there now.

super_editor/lib/src/infrastructure/content_layers.dart
```diff
@@ -313,11 +321,13 @@ class ContentLayersElement extends RenderObjectElement {
...
     contentLayersLog.finer("ContentLayersElement - temporarily forgetting layers");
     for (final underlay in _underlays) {
-      forgetChild(underlay);
+      // Calling super.forgetChild directly to avoid adding it to _forgottenChildren.
+      super.forgetChild(underlay);
     }
 
     for (final overlay in _overlays) {
-      forgetChild(overlay);
+      // Calling super.forgetChild directly to avoid adding it to _forgottenChildren.
```
matthew-carroll, line 329: I'm not sure what this and the other comment mean. You've described what the code does, but not why it does it. Why do we not want to add the overlays to `_forgottenChildren`? What would happen if we did? What is the `super` call doing with the `overlay`, instead?

super_editor/lib/src/infrastructure/content_layers.dart
matthew-carroll: It looks like we're still missing this requested comment.

super_editor/lib/src/infrastructure/content_layers.dart
matthew-carroll: Can you add a comment above the if-statement that mentions that the framework says its the responsibility of `visitChildren` to ignore all forgotten children (similar to your explanation in this thread)?

matthew-carroll:
@knopp can you give this PR another pass for comments?

knopp:
@matthew-carroll, let me know if this is any better.

matthew-carroll (approved):
LGTM - I directly committed a change to one of the comments to avoid another review round.

matthew-carroll:

PR #2256, opened 2024-08-19
Demonstrate issue where SuperTextFieldInspector.findScrollOffset() value is incorrect

CillianMyles:


CillianMyles:
cc/ @matthew-carroll

matthew-carroll:
I see that you added a number of expectations. I assume they all fail? What values are each of them reporting?

matthew-carroll:
@CillianMyles ping on the earlier review questions

super_editor/test/super_textfield/super_textfield_gesture_scrolling_test.dart
```diff
@@ -281,6 +282,10 @@ void main() {
       var textBottom = tester.getBottomRight(find.byType(SuperTextField)).dy;
       var viewportBottom = tester.getBottomRight(find.byType(SuperText)).dy;
       expect(textBottom, moreOrLessEquals(viewportBottom));
+      // Since the scrollable content is taller than the viewport, and since
+      // the bottom of the text field is aligned with the bottom of the
+      // viewport, the scroll offset should be greater than 0.
+      expect(SuperTextFieldInspector.findScrollOffset(), greaterThan(0.0));
```
CillianMyles, line 288: Part of the test:

```dart
testWidgetsOnMobile("multi-line is vertically scrollable when text spans more lines than maxLines", (tester) async {
```

And for both Android and iOS, the test fails because the value for `SuperTextFieldInspector.findScrollOffset()` is 0

super_editor/test/super_textfield/super_textfield_gesture_scrolling_test.dart
```diff
@@ -327,6 +334,10 @@ void main() {
       var textBottom = tester.getBottomRight(find.byType(SuperTextField)).dy;
       var viewportBottom = tester.getBottomRight(find.byType(SuperText)).dy;
       expect(textBottom, moreOrLessEquals(viewportBottom));
+      // Issue is not present on desktop, further leading me to think that the
+      // issue is somehow related to that comment about the scroll offsets
+      // being out of sync in the mobile scroll view.
+      expect(SuperTextFieldInspector.findScrollOffset(), greaterThan(0.0));
```
CillianMyles, line 340: Desktop equivalents pass (so the value is > 0)

CillianMyles: Printed the actual value here as 20, which makes sense considering we have a scrollable of height 40 with content of height 60

CillianMyles (commented):
Hi @matthew-carroll 

Apologies for the delay in getting back to you on this one!

I added some information around what happens in the tests + values you were looking for.

angelosilvestre:
@CillianMyles I ran into this issue while working on https://github.com/superlistapp/super_editor/issues/2316, but I'm not sure if it's the same issue you are facing.

It seems that sometimes the `TextScrollController` and the `ScrollController` can be out of sync. In my case, the  `ScrollController` was being restored from page storage. I changed  `SuperTextFieldInspector.findScrollOffset()` to read from the `ScrollController` instead of the `TextScrollController` and it fixed the test method. We still have the problem of the two controllers being out of sync though...

matthew-carroll:
@angelosilvestre I assigned this to you so that you can dig further into the root cause after we decide what to do about restoring the scroll offset.

Issue #2258, opened 2024-08-19
[BUG] - iOS cursor when long-pressing space key not working as expected when document is scrolled


miguelcmedeiros:
**Package Version**
super_editor main branch (commit https://github.com/superlistapp/super_editor/commit/5d57d893890a740e080dd357305839ce0b0582aa).

**User Info**
Superlist

**To Reproduce**
1. Go to SuperEditor demo
2. Scroll the document (e.g. middle of document )
3. Long-press space key to move caret

**Actual behavior**
The red cursor is not seen and it's unclear how the position of the grey caret changes.

**Expected behavior**
Red cursor and grey carets are not in sync. 

**Platform**
iOS

**Flutter version**
master

**Screenshots**

https://github.com/user-attachments/assets/0a7fa687-06e1-4449-8436-b8583d81d16e

miguelcmedeiros:
Cc @matthew-carroll @angelosilvestre

matthew-carroll:
FYI @knopp in case this ends up being a sliver issue

knopp:
Seems like a potential regression. Will take a look.

knopp:
This is a sliver issue. The floating point cursor is in viewport coordinates, which is not something that we can do in sliver world. We need to use document coordinates. 

I was wondering how I missed this so I reverted to before the sliver PR and noticed that it was also subtly broken before, where in sliver scenario it was actually in document coordinates (that's why I missed it), but parts of the code, i.e. autoscroll, assumed it was viewport coordinates and did not work correctly. I'll look into this.

PR #2260, opened 2024-08-19
Expose floating cursor geometry in document coordinates

knopp:
Fixes https://github.com/superlistapp/super_editor/issues/2258

The cursor geometry is currently exposed in viewport coordinates, which works correctly for autoscroll, but for correct positioning document coordinates are needed.


knopp:
@matthew-carroll, does this need a test? If so, what would be a good way to test this?

matthew-carroll (approved):
LGTM

matthew-carroll:

PR #2263, opened 2024-08-20
[SuperEditor][Android] Auto-scroll when selection changes (Resolves #2259)

### LINKED ISSUE #2259: [BUG] - On Android SuperEditor does not auto scroll while long-pressing space key and selection changes
**Package Version**
super_editor main branch (commit https://github.com/superlistapp/super_editor/commit/5d57d893890a740e080dd357305839ce0b0582aa).

**User Info**
Superlist

**To Reproduce**
1. Go to SuperEditor demo
2. Place the selection at the end of the visible content
3. Long-press space key and drag it to the right to move the selection 

**Actual behavior**
Selection changes but editor does not auto-scroll once the selection leaves the visible area of the editor.

**Expected behavior**
The editor auto-scrolls so that the selection is always visible.

**Platform**
Android

**Flutter version**
master

**Screenshots**

https://github.com/user-attachments/assets/dbfd08f6-3df8-4625-b5f1-c2221355875d

angelosilvestre:
[SuperEditor][Android] Auto-scroll when selection changes. Resolves #2259

When dragging the spacebar to change the selection, if the new selection happens to be outside of the viewport, we are not auto-scrolling the editor:

https://github.com/user-attachments/assets/dbfd08f6-3df8-4625-b5f1-c2221355875d

The issue is that we are only auto-scrolling when the document changes, and not when the selection changes, which is the case.

This PR fixes the issue by auto-scrolling when the selection changes and the new selection is collapsed. When the selection is expanded this isn't needed as the user is either dragging a handle, or we already auto-scrolled after the first tap if the user double taps to select a word.

[android_spacebar_selection.webm](https://github.com/user-attachments/assets/295a277b-9df9-4bba-adb8-9c51a593b6c2)

Unrelated: we should probably consider changing `testWidgetsOnAndroid` and the other test runner methods to automatically include the platform suffix.


matthew-carroll:
> Unrelated: we should probably consider changing testWidgetsOnAndroid and the other test runner methods to automatically include the platform suffix.

Please file an issue.

super_editor/test/super_editor/supereditor_scrolling_test.dart
```diff
@@ -292,6 +292,56 @@ void main() {
       );
     });
 
+    testWidgetsOnAndroid("auto-scrolls to caret position (on Android)", (tester) async {
```
matthew-carroll, line 295: This test name seems ambiguous. Are we talking about spacebar based dragging? Caret dragging? Changing the selection logically? Pressing arrow keys on a Bluetooth keyboard that move the caret? All of those behaviors might have different bugs and consequences, so we should be explicit about what we're locking down in tests.

matthew-carroll:
@angelosilvestre looks like we now have golden failures - probly not related to this, but we should get them fixed.

angelosilvestre:
> > Unrelated: we should probably consider changing testWidgetsOnAndroid and the other test runner methods to automatically include the platform suffix.
> 
> Please file an issue.
> 

Filed https://github.com/superlistapp/super_editor/issues/2274

angelosilvestre:
@matthew-carroll Added a tolerance for most of the goldens and re-generated one that had a bigger difference.

matthew-carroll (approved):
LGTM

PR #2264, opened 2024-08-20
[Spellcheck] Expose `correction` method

angelosilvestre:
[Spellcheck] Expose correction method.

This PR exposes the `correction` method from NSSpellChecker: https://developer.apple.com/documentation/appkit/nsspellchecker/1531542-correction


matthew-carroll (commented):
LGTM

PR #2272, opened 2024-08-26
[SuperEditor][Android] Honor handle builders (Resolves #1934)

### LINKED ISSUE #1934: Android handle builders don't seem to be used
**Package Version**
`main`, `stable`

**To Reproduce**
1. Provide a `SuperEditorAndroidControlsController` with a builder callback to `collapsedHandleBuilder` or `expandedHandlesBuilder`.
2. Build the editor
3. See them never get called

**Actual behavior**
It seems that out of all the `SuperEditorAndroidControlsController` builders:
- collapsedHandleBuilder
- expandedHandlesBuilder
- toolbarBuilder
- magnifierBuilder
Only the last two are ever used.

Looking at: https://github.com/superlistapp/super_editor/blob/661ecb84b4f79bd3f81c1bfd148b36ab7be6993b/super_editor/lib/src/default_editor/document_gestures_touch_android.dart#L112-L125
Seems to confirm this. These controller builders are never used.

angelosilvestre:
[SuperEditor][Android] Honor handle builders. Resolves #1934

`SuperEditorAndroidControlsController` exposes `collapsedHandleBuilder` and `expandedHandlesBuilder`, but neither of them are being used anywhere.

This PR changes the android touch interactor to honor the provided builders.

The drag handles involve various gesture callbacks to place/expand/collapse the selection, start/stop the blinking caret, show/hide the toolbar, among other things. It would be very difficult for developers to do that on their own.

Because of that, this PR introduces an object called `DocumentHandleGestureCallbacks`, which holds the gesture callbacks that should be attached to the drag handle. Developers would wrap the handle with a `GestureDetector` and set each property to its corresponding `GestureDetector` callback.

I needed to modify the expanded handles selection logic, because after these changes, when the user starts dragging  the expanded handle, the pan update event is trigger before the user moves the drag handle to another character, so we couldn't determine if the user is dragging upstream or downstream. This breaks our logic of selecting by word or character.

The original ticket also mentions that we use `magnifierBuilder` and `toolbarBuilder` in different ways. For `magnifierBuilder`, we expect that developers will include the `Follower` in the returned widget, while in `toolbarBuilder` we wrap the returned widget with a `Follower`. Letting the developers attach the `Follower` to the tree seems more logical to me, because it would be able to customize the offset and anchors. But changing that now seems like a big breaking change. Should we do that?

### DIFF
--- a/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
@@ -1681,14 +1681,43 @@ class SuperEditorAndroidControlsOverlayManagerState extends State<SuperEditorAnd
           );
         }
 
-        return AndroidDefaultFollowingCollapsedHandle(
-          focalPoint: _controlsController!.collapsedHandleFocalPoint,
-          shouldShow: shouldShow,
-          gestureCallbacks: gestureCallbacks,
-          child: AndroidSelectionHandle(
-            key: DocumentKeys.androidCaretHandle,
-            handleType: HandleType.collapsed,
-            color: _controlsController!.controlsColor ?? Theme.of(context).primaryColor,
+        // Note: If we pass this widget as the `child` property, it causes repeated starts and stops
+        // of the pan gesture. By building it here, pan events work as expected.
+        return Follower.withOffset(
+          link: _controlsController!.collapsedHandleFocalPoint,
+          leaderAnchor: Alignment.bottomCenter,
+          followerAnchor: Alignment.topCenter,
+          // Use the offset to account for the invisible expanded touch region around the handle.
+          offset: -Offset(0, AndroidSelectionHandle.defaultTouchRegionExpansion.top) *
+              MediaQuery.devicePixelRatioOf(context),
+          child: AnimatedOpacity(
+            // When the controller doesn't want the handle to be visible, hide it.
+            opacity: shouldShow ? 1.0 : 0.0,
+            duration: const Duration(milliseconds: 150),
+            child: IgnorePointer(
+              // Don't let the handle respond to touch events when the handle shouldn't
+              // be visible. This is needed because we don't remove the handle from the
+              // tree, we just make it invisible. In theory, invisible widgets aren't
+              // supposed to be hit-testable, but in tests I found that without this
+              // explicit IgnorePointer, gestures were still being captured by this handle.
+              ignoring: !shouldShow,
+              child: GestureDetector(
+                onTapDown: (_) {
+                  // Register tap down to win gesture arena ASAP.
+                },
+                onTap: gestureCallbacks.onTap,
+                onPanStart: gestureCallbacks.onPanStart,
+                onPanUpdate: gestureCallbacks.onPanUpdate,
+                onPanEnd: gestureCallbacks.onPanEnd,
+                onPanCancel: gestureCallbacks.onPanCancel,
+                dragStartBehavior: DragStartBehavior.down,
+                child: AndroidSelectionHandle(
+                  key: DocumentKeys.androidCaretHandle,
+                  handleType: HandleType.collapsed,
+                  color: _controlsController!.controlsColor ?? Theme.of(context).primaryColor,
+                ),
+              ),
+            ),
           ),
         );
       },
@@ -1899,66 +1928,6 @@ class SuperEditorAndroidControlsOverlayManagerState extends State<SuperEditorAnd
           ),
         );
       },
-    );
-  }
-}
-
-/// A default implementation of following behavior and entrance/exit animation for an
-/// Android collapsed selection handle.
-///
-/// The [child] follows the [focalPoint], being aligned with the bottom center of the
-/// [focalPoint]. The [child] is fades in and out of based on the [shouldShow].
-///
-/// Provide a [touchRegionExpansion] to increase the touch region around the handle.
-class AndroidDefaultFollowingCollapsedHandle extends StatelessWidget {
-  const AndroidDefaultFollowingCollapsedHandle({
-    super.key,
-    required this.focalPoint,
-    required this.gestureCallbacks,
-    required this.shouldShow,
-    this.touchRegionExpansion = AndroidSelectionHandle.defaultTouchRegionExpansion,
-    required this.child,
-  });
-
-  final LeaderLink focalPoint;
-  final DocumentHandleGestureCallbacks gestureCallbacks;
-  final EdgeInsets touchRegionExpansion;
-  final bool shouldShow;
-  final Widget child;
-
-  @override
-  Widget build(BuildContext context) {
-    return Follower.withOffset(
-      link: focalPoint,
-      leaderAnchor: Alignment.bottomCenter,
-      followerAnchor: Alignment.topCenter,
-      // Use the offset to account for the invisible expanded touch region around the handle.
-      offset: -Offset(0, touchRegionExpansion.top) * MediaQuery.devicePixelRatioOf(context),
-      child: AnimatedOpacity(
-        // When the controller doesn't want the handle to be visible, hide it.
-        opacity: shouldShow ? 1.0 : 0.0,
-        duration: const Duration(milliseconds: 150),
-        child: IgnorePointer(
-          // Don't let the handle respond to touch events when the handle shouldn't
-          // be visible. This is needed because we don't remove the handle from the
-          // tree, we just make it invisible. In theory, invisible widgets aren't
-          // supposed to be hit-testable, but in tests I found that without this
-          // explicit IgnorePointer, gestures were still being captured by this handle.
-          ignoring: !shouldShow,
-          child: GestureDetector(
-            onTapDown: (_) {
-              // Register tap down to win gesture arena ASAP.
-            },
-            onTap: gestureCallbacks.onTap,
-            onPanStart: gestureCallbacks.onPanStart,
-            onPanUpdate: gestureCallbacks.onPanUpdate,
-            onPanEnd: gestureCallbacks.onPanEnd,
-            onPanCancel: () => gestureCallbacks.onPanCancel,
-            dragStartBehavior: DragStartBehavior.down,
-            child: child,
-          ),
-        ),
-      ),
     );
   }
 }
--- a/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
@@ -1324,11 +1324,39 @@ class SuperEditorAndroidControlsOverlayManagerState extends State<SuperEditorAnd
   final _dragHandleSelectionGlobalFocalPoint = ValueNotifier<Offset?>(null);
   final _magnifierFocalPoint = ValueNotifier<Offset?>(null);
 
+  late final DocumentHandleGestureDelegate _collapsedHandleGestureDelegate;
+  late final DocumentHandleGestureDelegate _upstreamHandleGesturesDelegate;
+  late final DocumentHandleGestureDelegate _downstreamHandleGesturesDelegate;
+
   @override
   void initState() {
     super.initState();
     _overlayController.show();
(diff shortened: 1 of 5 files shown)


matthew-carroll:
@angelosilvestre is this PR both connecting the existing builders, and also modifying the tooling to build them? If so, can you split this into two PRs: one for fixing the existing bug, and another for whatever improvements you want to make?

angelosilvestre:
@matthew-carroll I think that at least we should keep `DocumentHandleGestureCallbacks` in this PR. Otherwise it will be very impractical for anyone to use a custom handle, since the drag handle selection depends on the drag handle gesture callbacks.

angelosilvestre:
@matthew-carroll Updated removing `AndroidDefaultFollowingCollapsedHandle` from this PR.

matthew-carroll:
@angelosilvestre can you update the PR description so that it only describes the changes in this PR after you split this PR into two?

angelosilvestre:
@matthew-carroll Updated the description.

matthew-carroll:
@angelosilvestre since you focused in this PR around making it simpler and easier to create custom handles, can you add a code sample to the PR description that shows a minimum setup for a custom drag handle? That will probably help evaluate those decisions.

WRT the `Follower` issue, let's file a new issue for that and handle it later. We can make the breaking change when it seems like a good time to do that. We've got a lot of stuff in flight right now.

matthew-carroll:
@angelosilvestre did you update the description? I see that my last comment was the last activity on this thread, so I wasn't sure if that happened, or not.

angelosilvestre:
> @angelosilvestre since you focused in this PR around making it simpler and easier to create custom handles, can you add a code sample to the PR description that shows a minimum setup for a custom drag handle? That will probably help evaluate those decisions.

In this PR I focused on just making the handle builder be honored. Making it simpler and easier would be moved to a separate PR, right?

matthew-carroll:
I would defer to the earlier conversation in the thread. I think that's all the conversation we've had about this. What I don't know is whether your current PR description describes what's actually in this PR, or whether the description is still about the original PR, which included stuff that you later removed.

angelosilvestre:
I updated the description.

matthew-carroll:
> I needed to modify the expanded handles selection logic, because after these changes, panning was triggering the pan update while the handle still focused the initially focused character, causing the selection strategy to be confused to determine if the user was dragging upstream or downstream.

I'm not sure what this means. Can you try to describe this in a clearer way?

angelosilvestre:
Updated it again. Does it look better now?

matthew-carroll:
Ok, now it reads:

> I needed to modify the expanded handles selection logic, because after these changes, when the user starts dragging the expanded handle, the pan update event is trigger before the user moves the drag handle to another character, so we couldn't determine if the user is dragging upstream or downstream. This breaks our logic of selecting by word or character.

Wouldn't it be accurate to say if the user moves left, or up its an upstream selection, and if the user moves right or down then it's a downstream selection? There should probably be a little slop before making that choice, but we shouldn't have to wait for the handle to actually move for us to know the direction of the drag, right?

angelosilvestre:
> Wouldn't it be accurate to say if the user moves left, or up its an upstream selection, and if the user moves right or down then it's a downstream selection? There should probably be a little slop before making that choice, but we shouldn't have to wait for the handle to actually move for us to know the direction of the drag, right?

The change is basically adding that left/right move handling. Do we need to change the whole behavior on this PR?

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
```diff
@@ -1708,6 +1708,24 @@ class SuperEditorAndroidControlsOverlayManagerState extends State<SuperEditorAnd
           return const SizedBox();
         }
 
+        final gestureCallbacks = DocumentHandleGestureCallbacks(
```
matthew-carroll, line 1711: For all of these callbacks, I'm wondering why we're constructing these on every build? Isn't it the same configuration for these objects on every frame?

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
```diff
@@ -1751,73 +1769,95 @@ class SuperEditorAndroidControlsOverlayManagerState extends State<SuperEditorAnd
...
-              onPanStart: (details) => _onHandlePanStart(details, HandleType.upstream),
-              onPanUpdate: _onHandlePanUpdate,
-              onPanEnd: (details) => _onHandlePanEnd(details, HandleType.upstream),
-              onPanCancel: () => _onHandlePanCancel(HandleType.upstream),
-              dragStartBehavior: DragStartBehavior.down,
-              child: AndroidSelectionHandle(
-                key: DocumentKeys.upstreamHandle,
-                handleType: HandleType.upstream,
-                color: _controlsController!.controlsColor ?? Theme.of(context).primaryColor,
+  Widget _buildExpandedHandles() {
```
matthew-carroll, line 1772: Do we need to change this from `List<Widget>` to `Widget`? I'm hesitant to add more layout widgets when not needed, because it increases the odds that at some point something isn't aligned correctly because of we've got a `Stack` in a `Stack` in a `Stack`, etc.

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
```diff
@@ -1751,73 +1769,95 @@ class SuperEditorAndroidControlsOverlayManagerState extends State<SuperEditorAnd
...
+          );
+        }
+
+        if (!shouldShow) {
+          return const SizedBox();
+        }
+
+        return Stack(
+          children: [
+            Follower.withOffset(
```
matthew-carroll, line 1815: When you mentioned in the issue ticket that it's difficult for users to configure their own handles, did you include the `Follower` in your description of it being difficult? I.e., did you want to wrap the custom handle with a `Follower` so the user doesn't have to? Or is it better like this to have the user provide the `Follower`?

super_editor/lib/src/infrastructure/platforms/mobile_documents.dart
```diff
@@ -67,12 +101,40 @@ typedef DocumentCollapsedHandleBuilder = Widget Function(BuildContext, Key handl
...
+});
+
+/// Callbacks for handling gestures on a document handle.
+///
+/// These callbacks are intended to make it easier for developers to customize
+/// the drag handles, without having to re-implement the gesture logic.
+///
+/// For example, wrap the handle in a `GestureDetector` and pass these callbacks
+/// to the corresponding gesture events.
+class DocumentHandleGestureCallbacks {
```
matthew-carroll, line 121: Let's call this a `DocumentHandleGestureDelegate` - please update all other places where the name is used.

super_editor/test/super_editor/mobile/super_editor_android_overlay_controls_test.dart
```diff
@@ -304,6 +305,91 @@ void main() {
       expect(SuperEditorInspector.isCaretVisible(), true);
     });
 
+    testWidgetsOnAndroid("allows customizing the collapsed handle", (tester) async {
```
matthew-carroll, line 308: Given that most of this PR is based on delegating gesture behavior, should this PR add a number of tests that use those gestures? Or is that fully covered by existing tests?

Also, why are we only testing for Android?

super_editor/test/super_editor/supereditor_test_tools.dart
```diff
@@ -309,6 +309,18 @@ class TestSuperEditorConfigurator {
     return this;
   }
 
+  /// Configures the [SuperEditor] to use the given [builder] as its android collapsed handle builder.
+  TestSuperEditorConfigurator withAndroidCollapsedHandleBuilder(DocumentCollapsedHandleBuilder? builder) {
```
matthew-carroll, line 313: Why only for Android?

angelosilvestre: The original ticket is only about Android. We don't have handle builders for iOS.

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
angelosilvestre: We can keep it as a list. Reverted.

super_editor/test/super_editor/mobile/super_editor_android_overlay_controls_test.dart
angelosilvestre: We currently have test ensuring that dragging the expanded handle changes the selection. Should we add a test for the collapsed handle? We have properties for handle customization on Android only.

matthew-carroll: Seems like tests for the collapsed handle would be as important as for expanded handles.

super_editor/lib/src/infrastructure/platforms/mobile_documents.dart
```diff
@@ -23,12 +23,24 @@ class DocumentKeys {
 /// The [handleKey] is used to find the handle in the widget tree for various purposes,
 /// e.g., within tests to verify the presence or absence of the handle.
 ///
+/// The [gestureDelegate] hold event handles that should be attached to the gesture recognizer
```
matthew-carroll, line 26: This paragraph is confusing - I'm not sure what exactly it's trying to say.

"The [gestureDelegate] hold event handles". First, it sounds like "hold" -> "holds". Beyond that, I don't know what it means to "hold event handles"...do you mean "handlers"? If so, the delegate doesn't "hold event handlers", a delegate "defines handlers for gesture events that occur on a given handle".

"For example, set [gestureDelegate.onTap] to the handle's `onTap` event."

I don't know what this is telling me to do. How do I set a method on the handle's `onTap`? Is this a statement about what I need to do as a reader, or a description of what is happening behind the scenes? I can't tell.

super_editor/lib/src/infrastructure/platforms/mobile_documents.dart
```diff
@@ -23,12 +23,24 @@ class DocumentKeys {
...
+/// attached to the handle. For example, set [gestureDelegate.onTap] to the handle's `onTap` event.
+///
+/// Use [shouldShow] to fade in/out the handle entrance/exit, for example, using an [AnimatedOpacity]
```
matthew-carroll, line 29: This paragraph is also unclear. When reading this, I can't really tell what you want me to do, or what Super Editor does. I generally understand that something is going on with fading, but there seems to be like 3 distinct details that are smushed into one statement and it's confusing.

super_editor/lib/src/infrastructure/platforms/mobile_documents.dart
```diff
@@ -67,12 +101,40 @@ typedef DocumentCollapsedHandleBuilder = Widget Function(BuildContext, Key handl
...
+  required DocumentHandleGestureDelegate downstreamGestureDelegate,
+  required bool shouldShow,
+});
+
+/// Delegate for handling gestures on a document handle.
+///
+/// These callbacks are intended to make it easier for developers to customize
+/// the drag handles, without having to re-implement the gesture logic.
+///
+/// To use it, for example, wrap the handle in a `GestureDetector` and pass
```
matthew-carroll, line 119: Can you replace this last sentence with a minimal code example showing this in use?

super_editor/lib/src/infrastructure/platforms/mobile_documents.dart
angelosilvestre: Updated.

super_editor/lib/src/infrastructure/platforms/mobile_documents.dart
angelosilvestre: I updated trying to make it a bit more clear.

super_editor/lib/src/infrastructure/platforms/mobile_documents.dart
angelosilvestre: Updated.

matthew-carroll (approved):
LGTM

PR #2284, opened 2024-08-28
Extend more types of attributions when typing (Resolves #2280)

### LINKED ISSUE #2280: [SuperEditor] - Add more styles to UpdateComposerTextStylesReaction for automatic extension
When a style like bold or italics are applied to text, and the user deletes some text or types more text, the `UpdateComposerTextStylesReaction` ensures that the those attributions remain active in the `DocumentComposer`'s preferences so that the next character that's typed also has that attribution.

Currently `UpdateComposerTextStylesReaction` has a list of default attributions that it automatically re-enables called `defaultExtendableStyles`. This list includes bold, italic, underline, and strikethrough.

Update this list to include inline code, foreground color, and background color.

To include foreground color and background color we need to adjust the API so that it's not purely based on value. While `boldAttribution` and `codeAttribution` are global constants, foreground color and background color have different instances based on the color. Therefore, instead of an equality check, we need to allow for a type check, and probably an arbitrary check, too.

Also, can we make it easier for clients to adjust this list? Currently, users need to customize the `Editor` creation process, copy all the existing `defaultEditorReactions` and then adjust the constructor for `UpdateComposerTextStylesReaction`.

matthew-carroll:
Extend more types of attributions when typing (Resolves #2280)

Auto-extend `codeAttribution`s. Also auto-extend any attribution of type `ColorAttribution` or `BackgroundColorAttribution`.

I wanted to find a way to configure these extensions without needing to use a custom configuration for `Editor`, but I wasn't able to think of anything compelling.

### DIFF
--- a/super_editor/lib/src/default_editor/composer/composer_reactions.dart
+++ b/super_editor/lib/src/default_editor/composer/composer_reactions.dart
@@ -205,6 +205,7 @@ final defaultExtendableStyles = Set.unmodifiable({
 });
 
 const defaultExtendableTypes = {
+  FontSizeAttribution,
   ColorAttribution,
   BackgroundColorAttribution,
 };
--- a/super_editor/lib/src/default_editor/composer/composer_reactions.dart
+++ b/super_editor/lib/src/default_editor/composer/composer_reactions.dart
@@ -17,7 +17,8 @@ import 'package:super_editor/src/default_editor/text.dart';
 /// be automatically applied to newly typed text. This reaction identifies these
 /// situations and activates the desired styles in the [DocumentComposer].
 ///
-/// Only the given `stylesToExtend` are automatically activated.
+/// Only the given [styleValuesToExtend], [styleTypesToExtend], [styleSelectorsToExtend]
+/// are automatically activated.
 ///
 /// Styles are activated when placing the caret at the beginning of a paragraph,
 /// and the first character has a style:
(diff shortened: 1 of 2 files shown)


super_editor/lib/src/default_editor/composer/composer_reactions.dart
```diff
@@ -160,9 +185,26 @@ class UpdateComposerTextStylesReaction extends EditReaction {
...
   italicsAttribution,
   underlineAttribution,
   strikethroughAttribution,
+  codeAttribution,
 });
+
+const defaultExtendableTypes = {
+  ColorAttribution,
+  BackgroundColorAttribution,
+};
```
angelosilvestre, line 210: What about also extending for `FontSizeAttribution`? At least in Google docs it seems to extend it.

super_editor/test/super_editor/supereditor_attributions_test.dart
```diff
@@ -12,6 +12,190 @@ import 'test_documents.dart';
...
+
+            final document = testContext.document;
+
+            // Place the caret at "Color|".
+            await tester.placeCaretInParagraph(document.first.id, 5);
+
+            // Type text that should expand the bold attribution.
+            await tester.typeImeText("s");
+
+            // Ensure the bold attribution was applied to the inserted text.
```
angelosilvestre, line 79: "bold" -> "color"

super_editor/test/super_editor/supereditor_attributions_test.dart
```diff
@@ -12,6 +12,190 @@ import 'test_documents.dart';
...
+            await tester.pressBackspace();
+            await tester.pressBackspace();
+            await tester.pressBackspace();
+
+            // Type at a character that should expand the bold attribution.
+            await tester.typeImeText("s");
+
+            // Ensure the bold attribution was applied to the inserted text.
+            final text = SuperEditorInspector.findTextInComponent(document.first.id);
+            expect(text.text, "Colors");
```
angelosilvestre, line 179: "bold" -> "color"

angelosilvestre (approved):
LGTM with a few minor comments.

PR #2287, opened 2024-08-28
[Quill] - Fix exceptions when parsing ambiguous custom block formats (Resolves #2286)

### LINKED ISSUE #2286: [Quill] - Bug: block deserialization formats blow up if assumptions not met
When deserializing block formats, we make certain assumptions that can blow up under certain reasonable circumstances.

Example: Consider a custom block format with an embedded list property:

```
{
  "list": {
    "list": "ordered"
  }
}
```

Compare the custom block format to the standard Delta list item format:

```
{
  "list": "ordered"
}
```

Both of these are allowed. Yet, if we send the custom delta through the default block formats, we get an exception because the list item delta parser expects the value of "list" to be a `String`.

matthew-carroll:
[Quill] - Fix exceptions when parsing ambiguous custom block formats (Resolves #2286)

When implementing a custom Delta format that happens to have the same key as a standard format, e.g., "list", we were blowing up with an exception because the standard format assumed the wrong value type for the given key.

### DIFF
--- a/super_editor_quill/test/parsing_test.dart
+++ b/super_editor_quill/test/parsing_test.dart
@@ -565,5 +565,114 @@ void main() {
         expect((document.getNodeAt(2)! as ParagraphNode).text.text, "");
       });
     });
+
+    group("custom parsers >", () {
+      test("works with ambiguous formats", () {
+        // Consider the standard Delta list item format:
+        // {
+        //   "list": "ordered"
+        // }
+        //
+        // We want to ensure that an ambiguous custom format is respected
+        // without issue, e.g.,
+        // {
+        //   "list": {
+        //     "list": "ordered"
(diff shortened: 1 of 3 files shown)


super_editor_quill/test/parsing_test.dart
```diff
@@ -565,5 +565,114 @@ void main() {
...
+                "align": "left",
+              },
+            },
+            "insert": "\n"
+          },
+          {"insert": "Paragraph six\n"},
+        ]);
+
+        // If we get here without an exception, the test passes. This means that
+        // the standard
```
angelosilvestre, line 633: It seems this sentence is incomplete.

angelosilvestre (commented):
LGTM with a minor comment.

PR #2290, opened 2024-08-30
[Quill] - Added parsing and serialization support for inline embeds (Resolves #2289)

### LINKED ISSUE #2289: [Quill] - Serialize and deserialize inline embeds
Quill has the concept of embeds. These can be block-level or inline. A block level embed might be an image, an inline embed might be a math formula, or a user mention.

Update the serialization and deserialization interfaces to support inline embeds.

matthew-carroll:
[Quill] - Added parsing and serialization support for inline embeds (Resolves #2289)

### DIFF
--- a/super_editor_quill/lib/src/parsing/inline_formats.dart
+++ b/super_editor_quill/lib/src/parsing/inline_formats.dart
@@ -183,3 +183,22 @@ class NamedInlineDeltaFormat implements InlineDeltaFormat {
 abstract interface class InlineDeltaFormat {
   Attribution? from(Operation operation);
 }
+
+abstract interface class InlineEmbedFormat {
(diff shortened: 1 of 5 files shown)


super_editor_quill/lib/src/parsing/inline_formats.dart
```diff
@@ -183,3 +183,22 @@ class NamedInlineDeltaFormat implements InlineDeltaFormat {
 abstract interface class InlineDeltaFormat {
   Attribution? from(Operation operation);
 }
+
+abstract interface class InlineEmbedFormat {
```
angelosilvestre, line 187: We should probably add a dart doc for this.

angelosilvestre (commented):
LGTM

PR #2295, opened 2024-09-02
Parse deltas using a custom editor (Resolves #2293)

### LINKED ISSUE #2293: [Quill] - Take custom Editors in parseQuillDeltaOps
The super_editor_quill package offers a couple global methods to parse deltas into a `MutableDocument`. Internally, these methods create a `MutableDocument`, `MutableDocumentComposer`, and an `Editor`.  The problem is that this `Editor` is configured with default request handlers, which prevents apps with custom request handlers from taking their desired actions.

Make it possible to pass a custom `Editor` into these methods.

matthew-carroll:
Parse deltas using a custom editor (Resolves #2293)


angelosilvestre (commented):
LGTM

PR #2320, opened 2024-09-16
[SuperEditor] - Don't clip iOS drag handles, add start of chat example app

matthew-carroll:
[SuperEditor] - Don't clip iOS drag handles, add start of chat example app

In a client app, the iOS handles were being cut off at the boundary of the editor. This PR disables clipping on the relevant `Stack`s. 

This PR also adds the start of a chat demo, which approximates the client's use-case.

Screenshot 2024-09-15 at 10 50 07 PM


angelosilvestre (commented):
LGTM

PR #1269, opened 2023-07-31
[SuperEditor][web] Fix text input after a deletion (Resolves #1224)

angelosilvestre:
[SuperEditor][web] Fix text input after a deletion. Resolves #1224

On web, if we quickly press backspace and type a character at the end of a paragraph we get the following exception and the text input crashes:

```
Couldn't map an IME position to a document position.
```

The issue is that we are handling the deletion twice on web. In IME mode, Flutter reports both the deletion delta and the backspace keypress.

This PR changes `defaultImeKeyboardActions` to include a different version of `deleteUpstreamContentWithBackspace`, which does nothing when running on web.

Issue #1224, opened 2023-07-06
Web text input works inconsistently


brian-superlist:
- Launch the Example Editor with Chrome. (`super_editor/example (branch stable) ❯ flutter run -d chrome`) 
- Type in the document for a bit
- Sometimes, the document updates in response to text input, then eventually stops responding to text input.
- Other times, the editor does not update in response to text input, then somehow starts updating in response to text input.
- May be related to spacebar or delete keys.

Flutter Doctor

```
Doctor summary (to see all details, run flutter doctor -v):
[!] Flutter (Channel stable, 3.10.5, on macOS 13.4.1 22F82 darwin-arm64, locale en-US)
    ! The flutter binary is not on your path. Consider adding /Users/brian/lab/flutter/bin to your path.
    ! The dart binary is not on your path. Consider adding /Users/brian/lab/flutter/bin to your path.
[✓] Android toolchain - develop for Android devices (Android SDK version 33.0.2)
[✓] Xcode - develop for iOS and macOS (Xcode 14.3.1)
[✓] Chrome - develop for the web
[!] Android Studio (not installed)
[✓] IntelliJ IDEA Community Edition (version 2023.1.3)
[✓] VS Code (version 1.79.2)
[✓] Connected device (3 available)
[✓] Network resources

! Doctor found issues in 2 categories.
```

miguelcmedeiros:
Cc @matthew-carroll

matthew-carroll:
@miguelcmedeiros what's the priority of this for Superlist? Is there a push towards web?

matthew-carroll:
> Sometimes, it works immediately then eventually stops working.

@brian-superlist can you please define "stops working"?

brian-superlist:
Heya @matthew-carroll. What I mean is that:

  - SuperEditor does not respond to text input
  - I can no longer enter text into the document.
  - If I type in letters, the SuperEditor Document does not update.

Here's a video of it happening. The editor works fine the first 15 seconds, then when I type into the editor, the document is not updated:

https://github.com/superlistapp/super_editor/assets/90610075/7090a191-f8ec-4de1-a55e-9690bf9436b2

brian-superlist:
With regards to priority, we are looking to launch our web app soon. We've got the app running against our production environment and are currently working through many bugs. I'd imagine we'll be ready to go on our side in 4 weeks time.

angelosilvestre:
An update:

[...]

PR #2327, opened 2024-09-20
[SuperEditor][Web] Fix selection change with CMD + RIGHT (Resolves #2304)

### LINKED ISSUE #2304: [BUG] [MacOS + Web] - CMD+Left/Right arrow does not move selection correctly
**Package Version**
super_editor, main branch, flutter master.

**User Info**
Superlist

**To Reproduce**
1. On a Mac
2. Update to latest version
3. Run the example app with Chrome
4. Place cursor on the beginning of the line "To get started with your own editing experience, take the following steps:"
5. Press CMD+Right arrow
6. **Expected:** Cursor is placed at the end of the line, **Actual:** Cursor is placed at the beginning of the word "steps"
7. Press CMD+Left arrow
8. **Expected:**, Cursor at beginning of line **Actual:** Cursor at beginning of line (all good!)
9. Press CMD+Left arrow again 
10. **Expected:** Cursor still at beginning of line, **Actual:** Cursor moved to previous line

**Minimal Reproduction Code**
None needed. Example app of super_editor repo.

**Platform**
MacOS + Web

**Flutter version**
Flutter 3.24.0-1.0.pre.565 • channel master • https://github.com/superlistapp/flutter.git

angelosilvestre:
[SuperEditor][Web] Fix selection change with CMD + RIGHT. Resolves #2304

Steps to reproduce: 

1. Run the example app with Chrome 
2. Place cursor on the beginning of the line "To get started with your own editing experience, take the following steps:"
3. Press CMD+Right arrow

Expected: Cursor is placed at the end of the line, Actual: Cursor is placed at the beginning of the word "steps"

This wrong offset is being reported by the selection change delta. On web, Flutter relies on html inputs to handle text input. When the user presses CMD + RIGHT, the browser moves the selection to where it thinks the end of the line is, and the Flutter code generates text deltas for us.

I haven't dug deep into this, but this seems our text layout and the browser text layout are different. For example, in the repro text, the browser thinks the line ends before "steps".

We started handling deltas for moving the caret with the arrow keys in https://github.com/superlistapp/super_editor/pull/1269. I think I failed to document why we need to let the IME generate non-text deltas for this. However, I tested the issues reported in the original ticket (https://github.com/superlistapp/super_editor/issues/1224) and I wasn't able to reproduce any of them.

This PR changes `doNothingWithLeftRightArrowKeysAtMiddleOfTextOnWeb` to avoid sending the key event back to the OS if CMD is pressed.

The ideal solution would be to investigate why our line breaks are different from the browser line breaks, but this will require a deeper investigation.

As a reference, in our "Regular Flutter TextField Demo", if we limit the width of the Flutter TextField to 400px, we can also see the caret moving to the wrong position.

### DIFF
--- a/super_editor/lib/src/default_editor/document_hardware_keyboard/document_keyboard_actions.dart
+++ b/super_editor/lib/src/default_editor/document_hardware_keyboard/document_keyboard_actions.dart
@@ -694,8 +694,7 @@ ExecutionInstruction moveToStartOrEndOfLineWithArrowKeysOnWeb({
   if ((CurrentPlatform.isApple && !HardwareKeyboard.instance.isMetaPressed) ||
       (const [TargetPlatform.windows, TargetPlatform.linux].contains(defaultTargetPlatform) &&
           !HardwareKeyboard.instance.isControlPressed)) {
-    // CMD or CTRL is not pressed. For character by character, word by word movement,
-    // we rely on the IME.
+    // CMD or CTRL is not pressed.
     return ExecutionInstruction.continueExecution;
   }
 
--- a/super_editor/lib/src/default_editor/document_hardware_keyboard/document_keyboard_actions.dart
+++ b/super_editor/lib/src/default_editor/document_hardware_keyboard/document_keyboard_actions.dart
@@ -635,6 +635,20 @@ ExecutionInstruction doNothingWithLeftRightArrowKeysAtMiddleOfTextOnWeb({
     return ExecutionInstruction.continueExecution;
   }
 
+  if ((CurrentPlatform.isApple && HardwareKeyboard.instance.isMetaPressed) ||
+      (const [TargetPlatform.windows, TargetPlatform.linux].contains(defaultTargetPlatform) &&
+          HardwareKeyboard.instance.isControlPressed)) {
+    // Pressing the arrow keys generates non-text deltas on web. However, when pressing CMD + RIGHT
+    // to move the selection to the end of the line, the selection change sometimes reports an offset which
+    // isn't at the end of the line. This seems to be an issue where our text layout is different
+    // from the browser's text layout. For example, the browser breaks the line at a different point
+    // than we do, so pressing CMD + RIGHT moves the selection to where the browser thinks the end of
+    // the line is.
+    //
+    // Don't block the event to let it be handled by our selection movement code.
+    return ExecutionInstruction.continueExecution;
+  }
+
   // On web, pressing left or right arrow keys generates non-text deltas.
   // We handle those deltas to change the selection. However, if the caret sits at the beginning
   // or end of a node, pressing these arrow keys doesn't generate any deltas.
--- a/super_editor/lib/src/default_editor/document_hardware_keyboard/document_keyboard_actions.dart
+++ b/super_editor/lib/src/default_editor/document_hardware_keyboard/document_keyboard_actions.dart
@@ -635,20 +635,6 @@ ExecutionInstruction doNothingWithLeftRightArrowKeysAtMiddleOfTextOnWeb({
     return ExecutionInstruction.continueExecution;
   }
 
-  if ((CurrentPlatform.isApple && HardwareKeyboard.instance.isMetaPressed) ||
-      (const [TargetPlatform.windows, TargetPlatform.linux].contains(defaultTargetPlatform) &&
-          HardwareKeyboard.instance.isControlPressed)) {
-    // Pressing the arrow keys generates non-text deltas on web. However, when pressing CMD + RIGHT
-    // to move the selection to the end of the line, the selection change sometimes reports an offset which
-    // isn't at the end of the line. This seems to be an issue where our text layout is different
-    // from the browser's text layout. For example, the browser breaks the line at a different point
-    // than we do, so pressing CMD + RIGHT moves the selection to where the browser thinks the end of
-    // the line is.
-    //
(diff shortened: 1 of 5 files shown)


angelosilvestre:
Playing around a bit with the css flutter generates, it seems that the html input does end the line before "steps" (red is the html input text):

image

matthew-carroll:
This probably shouldn't be too surprising that this happens given that Flutter uses a different text layout implementation than the browser. 

Can you file an issue with Flutter?

Also, did you test jumping by word to see if that's broken in a similar way?

super_editor/lib/src/default_editor/document_hardware_keyboard/document_keyboard_actions.dart
```diff
@@ -635,6 +635,20 @@ ExecutionInstruction doNothingWithLeftRightArrowKeysAtMiddleOfTextOnWeb({
     return ExecutionInstruction.continueExecution;
   }
 
+  if ((CurrentPlatform.isApple && HardwareKeyboard.instance.isMetaPressed) ||
```
matthew-carroll, line 638: Rather than have this blocking handler know about certain key combo exceptions, can you just move or replicate the relevant key combo above this one in the priority list, along with a comment about why?

angelosilvestre: Do you mean to have a `moveToStartOrEndOfLineWithArrowKeysOnWeb` ?

angelosilvestre:
> Can you file an issue with Flutter?

I'll try to make a repo.

>  Also, did you test jumping by word to see if that's broken in a similar way?

Jumping by word looks fine.

super_editor/lib/src/default_editor/document_hardware_keyboard/document_keyboard_actions.dart
matthew-carroll: I'm not sure about the specifics, because I don't know which handlers this one is currently blocking, but the name you mentioned seems to make sense. If we had such a handler higher up in priority, instead of blocking logic, would it still fix the issue?

angelosilvestre: Added a new handler for that.

super_editor/lib/src/default_editor/document_hardware_keyboard/document_keyboard_actions.dart
```diff
@@ -671,6 +671,63 @@ ExecutionInstruction doNothingWithLeftRightArrowKeysAtMiddleOfTextOnWeb({
...
+  }
+
+  const arrowKeys = [
+    LogicalKeyboardKey.arrowLeft,
+    LogicalKeyboardKey.arrowRight,
+  ];
+  if (!arrowKeys.contains(keyEvent.logicalKey)) {
+    return ExecutionInstruction.continueExecution;
+  }
+
```
matthew-carroll, line 693: Should we check for any other control keys, like Option? A quick check of cmd+option+arrow shows that something else happens in my browser. It doesn't move my cursor. Instead, it seems to switch browser tabs...

angelosilvestre: I tried in chrome and pressing cmd + option + arrow doesn't hit this handler. Maybe the browser is handling the shortcut before us and prevent the event?

angelosilvestre:
Updated the PR to prevent moving between nodes with CMD + LEFT/RIGHT arrow as discussed in the original ticket.

super_editor/test/super_editor/supereditor_input_keyboard_actions_test.dart
```diff
@@ -1616,6 +1616,58 @@ This is a paragraph
           ),
         );
       });
+
+      testWidgetsOnApple('with CMD + LEFT ARROW at the beginning of a paragraph', (tester) async {
```
matthew-carroll, line 1620: These tests verify that CMD + LEFT ARROW do nothing "at the beginning of the paragraph" and the right arrow does nothing "at the end of the paragraph". 

I'm confused by this.

CMD + LEFT/RIGHT ARROW should do nothing on Linux *ever*. Not just at the beginning and end of lines.

It's on a Mac that they should do nothing at the beginning/end of lines.

Also, isn't this ticket about web? If so, shouldn't the new tests simulate web, and reference web?

angelosilvestre: > CMD + LEFT/RIGHT ARROW should do nothing on Linux ever. Not just at the beginning and end of lines.

But we are not running on Linux, we are running on Apple Devices. I create a `testWidgetsOnMacDesktopAndWeb` that might be more appropriate here.

> Also, isn't this ticket about web? If so, shouldn't the new tests simulate web, and reference web?

Yes, but in [this comment](https://github.com/superlistapp/super_editor/issues/2304#issuecomment-2391048168) Brian agreed that we should keep this behavior consistent for web and desktop.

matthew-carroll: Aren't these tests declared inside of a "Linux" group?

angelosilvestre: I didn't notice that. Moved to "Mac and iOS >" group.

matthew-carroll (approved):
LGTM

PR #2328, opened 2024-09-21
[SuperEditor] Remove redundant code in UnknownComponentBuilder (Resolves #2309)

### LINKED ISSUE #2309: [BUG] - Redundant implementation in unknown_component.dart
(using stable)
I stumbled on a redundancy that looks like a mistake in lib/src/default_editor/unknown_component.dart

There's an `UnknownComponent` defined like this:
```
class UnknownComponent extends StatelessWidget {
  @override
    return const SizedBox(
      width: double.infinity,
      height: 54,
      child: Placeholder(),
    );
  }
}
```

but the `UnknownComponentBuilder` returns a nearly identical widget instead of the component itself:

```
@override
  Widget? createComponent(
      SingleColumnDocumentComponentContext componentContext, SingleColumnLayoutComponentViewModel componentViewModel) {
    editorLayoutLog.warning("Building component widget for unknown component: $componentViewModel");
    return SizedBox(
      key: componentContext.componentKey,
      width: double.infinity,
      height: 100,
      child: const Placeholder(),
    );
  }
```

angelosilvestre:
[SuperEditor] Remove redundant code in `UnknownComponentBuilder`. Resolves #2309

We have an `UnknownComponent`, but it isn't being used anywhere. Instead, we are using an nearly identical widget, only with the height different.

This PR changes `UnknownComponentBuilder` to use the `UnknownComponent` and changes `UnknownComponent` to be the same height as the widget that was being returned by the `UnknownComponentBuilder`.

### DIFF
--- a/super_editor/lib/src/default_editor/unknown_component.dart
+++ b/super_editor/lib/src/default_editor/unknown_component.dart
@@ -19,11 +19,8 @@ class UnknownComponentBuilder implements ComponentBuilder {
   Widget? createComponent(
       SingleColumnDocumentComponentContext componentContext, SingleColumnLayoutComponentViewModel componentViewModel) {
     editorLayoutLog.warning("Building component widget for unknown component: $componentViewModel");
-    return SizedBox(
+    return UnknownComponent(
       key: componentContext.componentKey,
-      width: double.infinity,
-      height: 100,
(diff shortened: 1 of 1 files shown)


matthew-carroll (approved):
LGTM

PR #2332, opened 2024-09-22
Cherry Pick: super_text_layout release, super_editor pubspec dep revision

matthew-carroll:
Cherry Pick: super_text_layout release, super_editor pubspec dep revision


matthew-carroll:
Test failure is due to non-existent screenshot files that currently exist on `main`. We need to copy those to `stable`.

PR #2338, opened 2024-09-24
Use a hack to fix Flutter text height mismatch (Resolves #2337)

### LINKED ISSUE #2337: [SuperEditor][SuperReader][SuperTextLayout] - Sometimes there's a mismatch of intrinsic text height when empty vs non-empty
For certain font sizes in Flutter, the reported height of an empty line of text is 1px off from that of a non-empty line of text.

This seems to be a Flutter bug. I filed: https://github.com/flutter/flutter/issues/155620

For now, we should hack around this by using our own measurement when the text is empty.

matthew-carroll:
Use a hack to fix Flutter text height mismatch (Resolves #2337)

For certain font sizes, Flutter reports a different height when the text is empty vs non-empty. I described that issue here: https://github.com/flutter/flutter/issues/155620

Because the effected font sizes are unpredictable, this PR introduces a hack that uses a TextPainter to measure a string with a single character when the text is empty. The TextPainter's height is used instead of the height measured by the underlying `RenderParagraph`.

I tried to reproduce the problem in a test so that we could write a test against this hack. However, the test always produced the same height whether the text was empty or not.


matthew-carroll:
Note: The analysis error is due to a deprecation on master that we haven't altered yet. I'm not doing that here because I want this PR to be able to cherry pick to stable.

angelosilvestre (approved):
LGTM

PR #2341, opened 2024-09-25
Use Timer instead of Ticker with mobile caret blink controller to avoid constantly pumping frames (Resolves #2340)

### LINKED ISSUE #2340: [SuperEditor][Mobile] - Bug: Constantly pumping frames
A client app found that we're constantly pumping frames on mobile, even when the editor doesn't have focus or selection.

The root cause is the use of a blinking controller with a `Ticker` instead of a `Timer`. We previously switched from `Ticker` to `Timer` on desktop to avoid this problem. Now we need to make this change on mobile.

matthew-carroll:
Use Timer instead of Ticker with mobile caret blink controller to avoid constantly pumping frames (Resolves #2340)

### DIFF
--- a/super_editor/lib/src/default_editor/super_editor.dart
+++ b/super_editor/lib/src/default_editor/super_editor.dart
@@ -1210,7 +1210,7 @@ const defaultComponentBuilders = <ComponentBuilder>[
 
 /// Default list of document overlays that are displayed on top of the document
 /// layout in a [SuperEditor].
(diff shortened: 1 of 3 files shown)


matthew-carroll:
I think the super_editor golden failure is related to recent work that hacked the caret size under certain circumstances.

angelosilvestre (approved):
LGTM

PR #2352, opened 2024-09-27
[SuperTextField] Fix golden failure (Resolves #2343)

### LINKED ISSUE #2343: [SuperTextField][SuperText] - SuperTextField golden failing after recent caret sizing fix
Recently we hacked a fix for the caret size due to a bug in Flutter when the caret sits at the end of text, after a space.

That hack generally seemed fine, but it apparently broke a golden test for SuperTextField. The failure isn't in the pixels, but rather in the layout process.

Update the caret hack so that it doesn't violate the layout assertion that was triggered in that test.

angelosilvestre:
[SuperTextField] Fix golden failure. Resolves #2343

The goldens are failing due to a layout crash with the following exception:

```console
The following assertion was thrown during performLayout():
RenderLeaderLayer does not meet its constraints.
Constraints:
  BoxConstraints(200.0<=w<=Infinity, 0.0<=h<=20.0)
Size:
  Size(200.0, 23.0)
```

This seems to be introduced by https://github.com/superlistapp/super_editor/pull/2338.

The crash is happening for an empty `SuperTextField` without any padding. The cause is that the height reported by the workaround code is bigger than the incoming constraints. This causes a crash during layout, because we are supposed to respect the incoming constraints.

This PR only fixes the crash by constraining the computed height, but we probably want to dig deeper to find out why the viewport is being smaller than it should be.

### DIFF
--- a/super_text_layout/lib/src/super_text.dart
+++ b/super_text_layout/lib/src/super_text.dart
@@ -429,7 +429,11 @@ class RenderLayoutAwareParagraph extends RenderParagraph {
           ..textDirection = textDirection
           ..textAlign = textAlign
           ..layout();
-        size = Size(size.width, _textPainter.height);
+
+        // Constrain the computed size because it can be bigger than the
+        // incoming constraints. Not respecting the incoming constraints
+        // is a Flutter violation and it causes a crash.
+        size = constraints.constrain(Size(size.width, _textPainter.height));
       }
     }
   }
--- a/super_text_layout/lib/src/super_text.dart
+++ b/super_text_layout/lib/src/super_text.dart
@@ -430,9 +430,11 @@ class RenderLayoutAwareParagraph extends RenderParagraph {
           ..textAlign = textAlign
           ..layout();
 
-        // Constrain the computed size because it can be bigger than the
-        // incoming constraints. Not respecting the incoming constraints
-        // is a Flutter violation and it causes a crash.
+        // We have no text, so we set the height of this render object to the line height
+        // of an arbitrary character of the given style. However, it's possible that our
+        // parent render object imposed constraints that are shorter than a single line
+        // of text. To avoid breaking Flutter's layout rules, we take the minimum height
+        // between the single line of text, and the incoming height constraints.
         size = constraints.constrain(Size(size.width, _textPainter.height));
       }
     }
--- a/super_editor/pubspec.yaml
+++ b/super_editor/pubspec.yaml
@@ -47,13 +47,13 @@ dependencies:
   flutter_test_robots: ^0.0.24
   clock: ^1.1.1
 
-#dependency_overrides:
+dependency_overrides:
   # Override to local mono-repo path so devs can test this repo
   # against changes that they're making to other mono-repo packages
-#  attributed_text:
-#    path: ../attributed_text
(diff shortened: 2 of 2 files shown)


matthew-carroll:
> This PR only fixes the crash by constraining the computed height, but we probably want to dig deeper to find out why the viewport is being smaller than it should be.

No need to hack around something. This is the time to dig into the root cause and resolve it.

angelosilvestre:
@matthew-carroll The root cause seems to be that this test uses a `lineHeight` of 20px and this isn't big enough for the computed height, even though the font size is set to 20px.

I think we could either:
- Modify the test to use a bigger `lineHeight`
- Modify the test to not specify a `lineHeight` and have the `TextScrollView` to use the same hack from https://github.com/superlistapp/super_editor/pull/2338 to size its viewport if the text is empty.

matthew-carroll:
> ...and this isn't big enough for the computed height...

@angelosilvestre what does that mean? Please try to be specific about cause and effect - when the explanation is ambiguous then we spend more time asking and providing clarifications.

angelosilvestre:
@matthew-carroll I mean that even though the font size is 20px the actual height of the text is bigger than that.

matthew-carroll:
I still don't know what that means. You're saying that text with a 20px font size isn't respecting the font size and it's rendering at a different size?

angelosilvestre:
It seems so, we are getting a 23.0px height for the "a" character.

super_text_layout/lib/src/super_text.dart
```diff
@@ -429,7 +429,11 @@ class RenderLayoutAwareParagraph extends RenderParagraph {
           ..textDirection = textDirection
           ..textAlign = textAlign
           ..layout();
-        size = Size(size.width, _textPainter.height);
+
+        // Constrain the computed size because it can be bigger than the
```
matthew-carroll, line 433: Let's change this comment to the following:

```
// We have no text, so we set the height of this render object to the line height
// of an arbitrary character of the given style. However, it's possible that our
// parent render object imposed constraints that are shorter than a single line
// of text. To avoid breaking Flutter's layout rules, we take the minimum height
// between the single line of text, and the incoming height constraints.
```

angelosilvestre: Updated.

matthew-carroll:
As mentioned in our Discord conversation, please check where the line height is being computed for the text scroller. Check if we should use this same height calculation there, and if we do, let's see if it fixes the test.

We can keep the size constraints in layout, but let's see if we can also solve the root problem.

matthew-carroll:
@angelosilvestre I'll go ahead and take this over because I've got some time right now.

matthew-carroll:
A note about the problem - it turns out that line height is passed into `SuperTextField`, rather than queried from the layout later. Therefore, changing how `SuperText` calculates and reports line height isn't sufficient to solve this problem.

I'll think about whether it's worth doing any further work on this, or if we should just merge the change that prevents the layout error.

matthew-carroll:
I think we'll go ahead and merge this because switching from a line height widget property to a layout calculated line height is an extensive change.

matthew-carroll (approved):
LGTM

Issue #1822, opened 2024-01-25
When parsing an empty sublist, `deserializeMarkdownToDocument` makes everything crash (probably because of an infinite loop)


manoletre:
**Package Version**
I am on the stable branch. I also just checked to be using the last commit (c69abd)

**To Reproduce**
1. Copy this string:

```
final test_string = """- hey there friend
    - """;
```

2. Run `deserializeMarkdownToDocument(test_string)`
3. Armageddon 💥

**Actual behavior**
The whole flutter app crashes. Probably due to an infinite loop

**Expected behavior**
The string should be parsed as this: (just without the dot on the sub list)

---

- hey there friend
     - .

---

**Platform**
I am on flutter web, on a MacOS. I am using Chrome.

**Flutter version**
```
Flutter 3.16.4 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 2e9cb0aa71 (6 weeks ago) • 2023-12-11 14:35:13 -0700
Engine • revision 54a7145303
Tools • Dart 3.2.3 • DevTools 2.28.4
```

**Additional context**
- #1711 and #1821 might be related

angelosilvestre:
@manoletre I've found the reason of the crash, but this input doesn't look like it generates a sublist. Most of the markdown readers I tried generate a single list item, with a "header 2" style. This also happens with the github comments:

Input:

```
- list item 1
    - 
```

Output:

- list item 1
    -

manoletre:
Yes you are right, most don't generate a sublist. I might have been wrong in expecting the sublist. I guess you can choose here what makes the most sense.

To me the main issue is the crash, how the string is actually parsed matters less. I stream markdown generated from OpenAI and sometimes it crashed. So as long as it does not crash anymore we are good 😁

PR #2364, opened 2024-10-04
upgrade markdown package reference

jezell:
This upgrades the markdown package to 7.2.1, also fixes an issue where encodeHtml is not passed to the markdown parser.

### DIFF
--- a/super_editor_markdown/lib/src/image_syntax.dart
+++ b/super_editor_markdown/lib/src/image_syntax.dart
@@ -24,7 +24,7 @@ class SuperEditorImageSyntax extends md.LinkSyntax {
         );
 
   @override
-  md.Node? close(
+  Iterable<md.Node>? close(
     md.InlineParser parser,
     covariant md.SimpleDelimiter opener,
     md.Delimiter? closer, {
@@ -38,7 +38,8 @@ class SuperEditorImageSyntax extends md.LinkSyntax {
     if (parser.pos + 1 >= parser.source.length) {
       // The `]` is at the end of the document, but this may still be a valid
       // shortcut reference link.
-      return _tryCreateReferenceLink(parser, text, getChildren: getChildren);
+      final link = _tryCreateReferenceLink(parser, text, getChildren: getChildren);
+      return link == null ? null : [ link ];
     }
 
     // Peek at the next character; don't advance, so as to avoid later stepping
@@ -51,7 +52,8 @@ class SuperEditorImageSyntax extends md.LinkSyntax {
       var leftParenIndex = parser.pos;
       var inlineLink = _parseInlineLink(parser);
       if (inlineLink != null) {
-        return _tryCreateInlineLink(parser, inlineLink, getChildren: getChildren);
+        final link = _tryCreateInlineLink(parser, inlineLink, getChildren: getChildren);
+        return link == null ? null : [ link ];
       }
       // At this point, we've matched `[...](`, but that `(` did not pan out to
       // be an inline link. We must now check if `[...]` is simply a shortcut
@@ -60,7 +62,8 @@ class SuperEditorImageSyntax extends md.LinkSyntax {
       // Reset the parser position.
       parser.pos = leftParenIndex;
       parser.advanceBy(-1);
-      return _tryCreateReferenceLink(parser, text, getChildren: getChildren);
+      final link = _tryCreateReferenceLink(parser, text, getChildren: getChildren);
+      return link == null ? null : [ link ];
     }
 
     if (char == AsciiTable.leftBracket) {
@@ -71,18 +74,21 @@ class SuperEditorImageSyntax extends md.LinkSyntax {
         // That opening `[` is not actually part of the link. Maybe a
         // *shortcut* reference link (followed by a `[`).
         parser.advanceBy(1);
-        return _tryCreateReferenceLink(parser, text, getChildren: getChildren);
+        final link = _tryCreateReferenceLink(parser, text, getChildren: getChildren);
+        return link == null ? null : [ link ];
       }
       var label = _parseReferenceLinkLabel(parser);
       if (label != null) {
-        return _tryCreateReferenceLink(parser, label, getChildren: getChildren);
+        final link = _tryCreateReferenceLink(parser, label, getChildren: getChildren);
+        return link == null ? null : [ link ];
       }
       return null;
     }
 
     // The link text (inside `[...]`) was not followed with a opening `(` nor
     // an opening `[`. Perhaps just a simple shortcut reference link (`[...]`).
-    return _tryCreateReferenceLink(parser, text, getChildren: getChildren);
+    final link = _tryCreateReferenceLink(parser, text, getChildren: getChildren);
+    return link == null ? null : [ link ];
   }
 
   /// Parses a size using the notation `=widthxheight`.
--- a/super_editor_markdown/lib/src/image_syntax.dart
+++ b/super_editor_markdown/lib/src/image_syntax.dart
@@ -38,8 +38,7 @@ class SuperEditorImageSyntax extends md.LinkSyntax {
     if (parser.pos + 1 >= parser.source.length) {
       // The `]` is at the end of the document, but this may still be a valid
       // shortcut reference link.
-      final link = _tryCreateReferenceLink(parser, text, getChildren: getChildren);
-      return link == null ? null : [ link ];
+      return _tryCreateReferenceLink(parser, text, getChildren: getChildren);
     }
 
     // Peek at the next character; don't advance, so as to avoid later stepping
@@ -52,8 +51,7 @@ class SuperEditorImageSyntax extends md.LinkSyntax {
       var leftParenIndex = parser.pos;
       var inlineLink = _parseInlineLink(parser);
       if (inlineLink != null) {
-        final link = _tryCreateInlineLink(parser, inlineLink, getChildren: getChildren);
-        return link == null ? null : [ link ];
+        return [ _tryCreateInlineLink(parser, inlineLink, getChildren: getChildren) ];
       }
       // At this point, we've matched `[...](`, but that `(` did not pan out to
       // be an inline link. We must now check if `[...]` is simply a shortcut
@@ -62,8 +60,7 @@ class SuperEditorImageSyntax extends md.LinkSyntax {
       // Reset the parser position.
       parser.pos = leftParenIndex;
       parser.advanceBy(-1);
-      final link = _tryCreateReferenceLink(parser, text, getChildren: getChildren);
-      return link == null ? null : [ link ];
+      return _tryCreateReferenceLink(parser, text, getChildren: getChildren);
     }
 
     if (char == AsciiTable.leftBracket) {
@@ -74,21 +71,18 @@ class SuperEditorImageSyntax extends md.LinkSyntax {
         // That opening `[` is not actually part of the link. Maybe a
         // *shortcut* reference link (followed by a `[`).
         parser.advanceBy(1);
-        final link = _tryCreateReferenceLink(parser, text, getChildren: getChildren);
-        return link == null ? null : [ link ];
+        return _tryCreateReferenceLink(parser, text, getChildren: getChildren);
       }
       var label = _parseReferenceLinkLabel(parser);
       if (label != null) {
-        final link = _tryCreateReferenceLink(parser, label, getChildren: getChildren);
-        return link == null ? null : [ link ];
+        return _tryCreateReferenceLink(parser, label, getChildren: getChildren);
       }
       return null;
     }
 
     // The link text (inside `[...]`) was not followed with a opening `(` nor
     // an opening `[`. Perhaps just a simple shortcut reference link (`[...]`).
-    final link = _tryCreateReferenceLink(parser, text, getChildren: getChildren);
-    return link == null ? null : [ link ];
+    return _tryCreateReferenceLink(parser, text, getChildren: getChildren);
   }
 
   /// Parses a size using the notation `=widthxheight`.
@@ -146,7 +140,7 @@ class SuperEditorImageSyntax extends md.LinkSyntax {
   /// Tries to create a reference link node.
   ///
   /// Returns the link if it was successfully created, `null` otherwise.
-  md.Node? _tryCreateReferenceLink(md.InlineParser parser, String label,
+  List<md.Node>? _tryCreateReferenceLink(md.InlineParser parser, String label,
       {required List<md.Node> Function() getChildren}) {
     return _resolveReferenceLink(label, parser.document.linkReferences, getChildren: getChildren);
   }
@@ -242,19 +236,21 @@ class SuperEditorImageSyntax extends md.LinkSyntax {
   /// Otherwise, returns `null`.
   ///
   /// [label] does not need to be normalized.
-  md.Node? _resolveReferenceLink(
+ List<md.Node>? _resolveReferenceLink(
     String label,
     Map<String, md.LinkReference> linkReferences, {
     required List<md.Node> Function() getChildren,
   }) {
     final linkReference = linkReferences[_normalizeLinkLabel(label)];
     if (linkReference != null) {
-      return createNode(
-        linkReference.destination,
-        linkReference.title,
-        //size: linkReference.size,
-        getChildren: getChildren,
-      );
+      return [
+        createNode(
+          linkReference.destination,
+          linkReference.title,
+          //size: linkReference.size,
+          getChildren: getChildren,
+        ) 
+      ];
     } else {
       // This link has no reference definition. But we allow users of the
       // library to specify a custom resolver function ([linkResolver]) that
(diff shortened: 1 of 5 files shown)


super_editor_markdown/lib/src/image_syntax.dart
```diff
@@ -38,7 +38,8 @@ class SuperEditorImageSyntax extends md.LinkSyntax {
     if (parser.pos + 1 >= parser.source.length) {
       // The `]` is at the end of the document, but this may still be a valid
       // shortcut reference link.
-      return _tryCreateReferenceLink(parser, text, getChildren: getChildren);
+      final link = _tryCreateReferenceLink(parser, text, getChildren: getChildren);
```
matthew-carroll, line 41: Looks like all uses of `_tryCreateReferenceLink()` gets converted to a list. Should we just change that method to return `Iterable<md.Node>?` and then continue to return it directly?

super_editor_markdown/lib/src/markdown_to_document_parsing.dart
```diff
@@ -23,9 +23,12 @@ MutableDocument deserializeMarkdownToDocument(
   List<ElementToNodeConverter> customElementToNodeConverters = const [],
   bool encodeHtml = false,
 }) {
-  final markdownLines = const LineSplitter().convert(markdown);
+  final markdownLines = const LineSplitter().convert(markdown).map((l) {
```
matthew-carroll, line 26: What happened to require this difference? I'm not sure what the existing code did, but it's not immediately obvious what the difference is in impact between the original code and this new code.

super_editor_markdown/lib/src/markdown_to_document_parsing.dart
```diff
@@ -519,22 +522,26 @@ abstract class ElementToNodeConverter {
...
     md.Delimiter opener,
     md.Delimiter closer, {
     required List<md.Node> Function() getChildren,
+    required String tag,
   }) {
-    return md.Element('u', getChildren());
+    final element = md.Element('u', getChildren());
+    return [ element ];
   }
+  static final _tags = [ md.DelimiterTag("u", 1) ];
```
matthew-carroll, line 544: Please declare all static members above all instance members (we probably need to add a lint for that at some point).

And can you also add a comment above `_tags` that describe what it is, and how it's used?

super_editor_markdown/lib/src/markdown_to_document_parsing.dart
```diff
@@ -548,7 +555,7 @@ class _ParagraphWithAlignmentSyntax extends _EmptyLinePreservingParagraphSyntax
 
   @override
   bool canParse(md.BlockParser parser) {
-    if (!_alignmentNotationPattern.hasMatch(parser.current)) {
+    if (!_alignmentNotationPattern.hasMatch(parser.current.content)) {
```
matthew-carroll, line 558: Was this a pre-existing error on our part? Should we have been accessing `parser.current.content` this whole time?

matthew-carroll:
FYI @angelosilvestre

matthew-carroll:
@jezell it looks like we've got a bunch of failing tests. I'm guessing most/all of them is because we need corresponding changes in other packages.

super_editor_markdown/lib/src/markdown_to_document_parsing.dart
jezell: Markdown lib changed from taking List<String> to taking List<Line>.

super_editor_markdown/lib/src/image_syntax.dart
jezell: Good catch. That would be better.

super_editor_markdown/lib/src/markdown_to_document_parsing.dart
jezell: This is required by md.DelimiterSyntax, in theory the constructor takes an optional tags member, but a bug in
the Markdown lib replaces that with a const [], and then tries to sort it. Will add a comment to that effect.

super_editor_markdown/lib/src/markdown_to_document_parsing.dart
jezell: Markdown lib changed from String to Line which has a content member, so this just redirects to where the string is now.

jezell:
Looking at the tests, I'm not sure about one of them, "unordered list followed by empty list item" seems to not actually convert in commonmark, so not sure if it's a bad test. Here too in this comment I've put the markdown and it doesn't end up as two list items, but the test wants it to. Maybe the new markdown parser is more correct here?

https://spec.commonmark.org/dingus/

- list item 1
    -

jezell:
The strikethrough test also looks bad, strikethrough is a double ~ not one, but the test has a single. Github markdown seems to support single, but that doesn't appear to be standard or documented (https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax):

https://www.markdownguide.org/extended-syntax/#:~:text=You%20can%20strikethrough%20words%20by,before%20and%20after%20the%20words.

Seems like maybe the dart markdown package got more strict. Definitely a change in behavior, but most guides seem to confirm it's supposed to be double not single tilde.

super_editor_markdown/test/super_editor_markdown_test.dart
```diff
@@ -1433,10 +1437,22 @@ with multiple lines
         expect(document.getNodeAt(25)!, isA<ParagraphNode>());
       });
 
-      test('paragraph with strikethrough', () {
+      test('paragraph with single ~ should not be strikethrough', () {
```
jezell, line 1440: This is a change in behavior, but it seems to be more correct. The underlying dart markdown parser seems to be more strict now, but it does seem to align better with the generally accepted pattern for strikethrough. 

Should the markdown parser accept the decision of the underlying markdown lib to change the behavior here?

super_editor_markdown/test/super_editor_markdown_test.dart
```diff
@@ -1129,17 +1129,21 @@ This is some code
         expect((document.first as ListItemNode).text.text, isEmpty);
       });
 
-      test('unordered list followd by empty list item', () {
-        const markdown = """- list item 1
-    - """;
+      test('unordered list followed by empty list item', () {
```
jezell, line 1132: There is a change in behavior here in this specific case introduced by the underlying parser changing, but was the intent of the test really to verify that the hanging - was deleted from the line in a weird case that markdown parsers don't seem to agree on? CommonMark spec seems to say that empty list items should be allowed, not stripped. This updates the test to verify that empty list items should be kept like the spec says, rather than test for something that seems to be up interpreted differently depending on the parser being used.

super_editor_markdown/test/super_editor_markdown_test.dart
```diff
@@ -1433,10 +1437,22 @@ with multiple lines
...
 
+        // Ensure text within the range isn't attributed.
+        expect(styledText.getAllAttributionsAt(0).contains(strikethroughAttribution), false);
+        expect(styledText.getAllAttributionsAt(6).contains(strikethroughAttribution), false);
+
+        // Ensure text outside the range isn't attributed.
+        expect(styledText.getAllAttributionsAt(7).contains(strikethroughAttribution), false);
+      });
+
+      test('paragraph with double strikethrough', () {
```
jezell, line 1452: Added a test for this case since it was missing

super_editor_markdown/test/super_editor_markdown_test.dart
matthew-carroll: It looks like this is where we introduced this test: https://github.com/superlistapp/super_editor/issues/1822

super_editor_markdown/test/super_editor_markdown_test.dart
matthew-carroll: I think the new name of this test is ambiguous. When I read it, I thought it meant a single standalone "~" without a closing "~". But based on the implementation, I take it that you're trying to describe the difference between "~" and "~~"?

If you're saying that the new package version doesn't parse "~something~" as strikethrough, then I think we need to adjust it to do so. Given we've had that behavior for a long time, I don't think we should break it for existing users. That said, users should have an opportunity to turn it off, if they choose.

Adding support for double "~~" should be fine because its a new feature, and if that really is the norm, then I don't foresee complaints.

super_editor_markdown/lib/src/markdown_to_document_parsing.dart
```diff
@@ -519,21 +522,28 @@ abstract class ElementToNodeConverter {
...
+/// A Markdown [DelimiterSyntax] that matches underline spans of text, which are represented in
 /// Markdown with surrounding `¬` tags, e.g., "this is ¬underline¬ text".
 ///
-/// This [TagSyntax] produces `Element`s with a `u` tag.
-class UnderlineSyntax extends md.TagSyntax {
-  UnderlineSyntax() : super('¬', requiresDelimiterRun: true, allowIntraWord: true);
+/// This [DelimiterSyntax] produces `Element`s with a `u` tag.
+class UnderlineSyntax extends md.DelimiterSyntax {
+
+  /// This is required by md.DelimiterSyntax, in theory the constructor takes an optional tags member, but a bug in
```
matthew-carroll, line 531: This comment still doesn't tell us what this thing is. What does "u" mean? What's a `DelimiterTag`? Why is it required? What happens if it's not passed to the super constructor? You somewhat answered that last part, but didn't mention the consequences.

Assume the next person who's responsible for this has very little understanding of the inside of the markdown package.

super_editor_markdown/lib/src/markdown_to_document_parsing.dart
matthew-carroll: Can you add an explicit type parameter to your map function so that the difference is unambiguous? Part of the confusion here is that reading the code doesn't tell us what type "l" is.

matthew-carroll:
Looks like we have 40 failing tests in super_editor - most/all of them look like auto-conversion of text, which is based on Markdown parsing.

super_editor_markdown/lib/src/markdown_to_document_parsing.dart
jezell: Was trying to explain that it's not actually supposed to be required from the class definition:

https://pub.dev/documentation/markdown/latest/markdown/DelimiterSyntax-class.html

The constructor takes a nullable. However, the problem is there is a bug in the underlying dart library if you don't pass it. Due to these two lines, one sets it to const [] if not passed, then the next tries to sort. So we have to pass something or it blows up. That's the only reason it's there.

https://github.com/dart-lang/markdown/blob/d53feae0760a4f0aae5ffdfb12d8e6acccf14b40/lib/src/inline_syntaxes/delimiter_syntax.dart#L67

https://github.com/dart-lang/markdown/blob/d53feae0760a4f0aae5ffdfb12d8e6acccf14b40/lib/src/inline_syntaxes/delimiter_syntax.dart#L319

matthew-carroll: I understand that point about the bug in the package - but those questions I raised above are still real questions that a reader will have. Can you please address those in the code comment?

jezell:
> Looks like we have 40 failing tests in super_editor - most/all of them look like auto-conversion of text, which is based on Markdown parsing.

@matthew-carroll looks like what was happening is the new darter markdown parser drops of extra newlines if there are hanging newlines at the end of the file after a list. There aren't any tests for this in the markdown package, but the tests in the super_editor package relies on it for setting the caret in an empty node after the lists. I'm adding a test for this case and some extra logic in the parser. It's a little dirty, but seems like the desire here is to insulate people from the markdown lib changing. At least the test will catch if they change the behavior back.

super_editor_markdown/test/super_editor_markdown_test.dart
jezell: @matthew-carroll added single line strikethrough parsing node, a bit hacky to get it to work since underlying markdown parser doesn't like having two sort of similar nodes, but does work as long as the order in the list is correct. Added comments to explain.

super_editor_markdown/lib/src/markdown_to_document_parsing.dart
jezell: Added this explanation to the code, hope it clarifies the reasoning this is required better.

super_editor_markdown/test/super_editor_markdown_test.dart
```diff
@@ -1129,17 +1129,21 @@ This is some code
         expect((document.first as ListItemNode).text.text, isEmpty);
       });
 
-      test('unordered list followd by empty list item', () {
-        const markdown = """- list item 1
-    - """;
+      test('unordered list followed by empty list item', () {
+        const markdown = """
```
matthew-carroll, line 1133: Does this leading newline need to be here? If a Markdown blob begins with a list item, it won't have a leading newline.

super_editor_markdown/test/super_editor_markdown_test.dart
```diff
@@ -1433,10 +1437,22 @@ with multiple lines
         expect(document.getNodeAt(25)!, isA<ParagraphNode>());
       });
 
-      test('paragraph with strikethrough', () {
+      test('paragraph with single strikethrough', () {
         final doc = deserializeMarkdownToDocument('~This is~ a paragraph.');
         final styledText = (doc.getNodeAt(0)! as ParagraphNode).text;
 
+        // Ensure text within the range isn't attributed.
```
matthew-carroll, line 1444: Shouldn't this test be reverted, given that you added the single-character parser?

super_editor_markdown/test/super_editor_markdown_test.dart
```diff
@@ -1541,6 +1557,23 @@ Paragraph3""";
         expect((doc.getNodeAt(2)! as ParagraphNode).text.text, 'Paragraph3');
       });
 
+       test('empty paragraph after paragraphs', () {
```
matthew-carroll, line 1560: Can you phrase the test name in terms of what you're verifying? I'm actually not sure at first glance which detail you're locking down here. Is it that 2 trailing newlines are reduced to 1? Is it that 2+ trailing newlines are reduced to 1?

matthew-carroll:
@angelosilvestre can you also take a look at this PR and see if you find anything that I've missed?

super_editor_markdown/test/super_editor_markdown_test.dart
jezell: Good catch. Test was reverted but comment wasn't.

super_editor_markdown/test/super_editor_markdown_test.dart
jezell: Doesn't need to be there, will remove the leading.

super_editor_markdown/test/super_editor_markdown_test.dart
jezell: updating to 'every 2 newlines after a list are a paragraph'. The logic is a bit weird but it seems to be what the old dart markdown parser was doing with new lines after lists. The super_editor tests that deal with inserting newlines after lists fail with null reference exceptions due to missing nodes if this doesn't happen. I first tried adding code to preserve all hanging newlines regardless of what came between, but other tests in the markdown package want a no newlines after other types of content. Seems like the old parser wasn't really consistent with hanging newlines and that leaked into the tests. So this retains old behavior, but also codifies old inconsistencies.

super_editor_markdown/lib/src/markdown_to_document_parsing.dart
```diff
@@ -58,6 +61,14 @@ MutableDocument deserializeMarkdownToDocument(
     );
   }
 
+  // Add 1 hanging line for every 2 blank lines at the end, need this to preserve behavior pre markdown 7.2.1
+  final hangingEmptyLines = markdownLines.reversed.takeWhile((md.Line l) => l.isBlankLine);
+  if(hangingEmptyLines.isNotEmpty && documentNodes.lastOrNull is ListItemNode) {
+    for(var i = 0; i < hangingEmptyLines.length >> 1; i++) {
```
angelosilvestre, line 67: Won't this be a bit more readable if we use `hangingEmptyLines.length ~/ 2` instead of `hangingEmptyLines.length >> 1`?

angelosilvestre (commented):
LGTM with a single comment.

super_editor_markdown/lib/src/markdown_to_document_parsing.dart
matthew-carroll: I didn't catch that for loop, but yeah, @jezell let's not use bit shifting unless we absolutely have to. If you wanna divide by two, let's use the standard operators for that.

matthew-carroll: @jezell can you push an update with this change?

jezell: pushed

matthew-carroll:
@angelosilvestre can you cherry pick when you get a chance?

PR #2368, opened 2024-10-06
[SuperEditor] Add ability to set a block node as non-deletable (Resolves #2334)

### LINKED ISSUE #2334: [FEATURE] - Allow to configure block node to not be selectable and/or deletable
**Your Info**
Superlist

**User Behavior**
The request is to have a block node that once placed in the document, it can't be selected or deleted by means of keyboard interactions (e.g. pressing backspace/delete).

Perhaps both aspected can be configured separately, i.e. the node should not be selectable but deletable, and vice-versa.

angelosilvestre:
[SuperEditor] Add ability to set a block node as undeletable. Resolves #2334

The issue ticket also mentions making block nodes unselectable. This is already possible by setting `isVisuallySelectable` `true` in  `BoxComponent`. There is a demo in `UnselectableHrDemo`.

This PR adds the ability to mark a node as undeletable. The goal at this moment is for us to decide if this is the approach that we want and also confirm the expected behavior for some cases. After that, I'll add tests to the PR.

Marking a node as undeletable can be achieved by setting a flag in the node's metadata:

```dart
HorizontalRuleNode(
  id: Editor.createNodeId(),
  metadata: {
    NodeMetadata.deletable: false,
  },
),
```

https://github.com/user-attachments/assets/56bbeb27-7275-47dc-98ab-089af5fea128

https://github.com/user-attachments/assets/06dfc0d0-f5d5-40e0-9860-b84bec16ea4a

### DIFF
--- a/super_editor/lib/src/core/document.dart
+++ b/super_editor/lib/src/core/document.dart
@@ -416,6 +416,8 @@ abstract class DocumentNode implements ChangeNotifier {
 
   DocumentNode copy();
 
+  bool get isDeletable => _metadata[NodeMetadata.isDeletable] != false;
+
   @override
   bool operator ==(Object other) =>
       identical(this, other) ||
--- a/super_editor/lib/src/core/document.dart
+++ b/super_editor/lib/src/core/document.dart
@@ -309,6 +309,10 @@ class DocumentPosition {
 
 /// A single content node within a [Document].
 abstract class DocumentNode implements ChangeNotifier {
+  DocumentNode({
+    Map<String, dynamic>? metadata,
+  }) : _metadata = metadata ?? {};
+
   /// ID that is unique within a [Document].
   String get id;
 
@@ -373,7 +377,7 @@ abstract class DocumentNode implements ChangeNotifier {
   /// Returns all metadata attached to this [DocumentNode].
   Map<String, dynamic> get metadata => _metadata;
 
-  final Map<String, dynamic> _metadata = {};
+  final Map<String, dynamic> _metadata;
 
   /// Sets all metadata for this [DocumentNode], removing all
   /// existing values.
@@ -465,8 +469,11 @@ class NodeMetadata {
 
   /// Whether or not the node is deletable.
   ///
-  /// A deletable node can be removed from the document, for example,
-  /// when pressing backspace when the node is selected.
+  /// A non-deletable node cannot be removed from the document by user
+  /// interaction. For exammple, selecting a non-deletable node and pressing
+  /// backspace has no effect.
+  ///
+  /// Apps can still remove non-deletable nodes by issuing a `DeleteNodeRequest`.
   ///
   /// If the node doesn't have this metadata, it is assumed to be deletable.
   static const String isDeletable = 'isDeletable';
--- a/super_editor/lib/src/core/document.dart
+++ b/super_editor/lib/src/core/document.dart
@@ -465,6 +465,9 @@ class NodeMetadata {
 
   /// Whether or not the node is deletable.
   ///
+  /// A deletable node can be removed from the document, for example,
+  /// when pressing backspace when the node is selected.
+  ///
   /// If the node doesn't have this metadata, it is assumed to be deletable.
-  static const String deletable = 'deletable';
+  static const String isDeletable = 'isDeletable';
 }
--- a/super_editor/lib/src/core/document.dart
+++ b/super_editor/lib/src/core/document.dart
@@ -457,3 +457,14 @@ abstract class NodePosition {
   /// even though [==] returns `false`.
   bool isEquivalentTo(NodePosition other);
 }
+
+/// Keys to access metadata on a [DocumentNode].
+class NodeMetadata {
+  /// Applies an [Attribution] to the node.
+  static const String blockType = 'blockType';
+
+  /// Whether or not the node is deletable.
+  ///
+  /// If the node doesn't have this metadata, it is assumed to be deletable.
+  static const String deletable = 'deletable';
+}
--- a/super_editor/lib/src/default_editor/common_editor_operations.dart
+++ b/super_editor/lib/src/default_editor/common_editor_operations.dart
@@ -1251,9 +1251,18 @@ class CommonEditorOperations {
   ///
   /// This can be used, for example, to effectively delete an image by replacing
   /// it with an empty paragraph.
-  void replaceBlockNodeWithEmptyParagraphAndCollapsedSelection(String nodeId) {
+  ///
+  /// If [abortIfUndeletable] is `true`, the operation is be aborted if the
+  /// node is not deletable.
+  void replaceBlockNodeWithEmptyParagraphAndCollapsedSelection(
+    String nodeId, {
+    bool abortIfUndeletable = true,
+  }) {
     editor.execute([
-      ReplaceNodeWithEmptyParagraphWithCaretRequest(nodeId: nodeId),
+      ReplaceNodeWithEmptyParagraphWithCaretRequest(
+        nodeId: nodeId,
+        abortIfUndeletable: abortIfUndeletable,
+      ),
     ]);
   }
 
@@ -1284,11 +1293,9 @@ class CommonEditorOperations {
 
     // Delete the selected content.
     editor.execute([
-      DeleteContentRequest(documentRange: composer.selection!),
-      ChangeSelectionRequest(
-        DocumentSelection.collapsed(position: newSelectionPosition),
-        SelectionChangeType.deleteContent,
-        SelectionReason.userInteraction,
+      DeleteContentRequest(
+        documentRange: composer.selection!,
+        newSelection: DocumentSelection.collapsed(position: newSelectionPosition),
       ),
     ]);
   }
--- a/super_editor/lib/src/default_editor/common_editor_operations.dart
+++ b/super_editor/lib/src/default_editor/common_editor_operations.dart
@@ -865,7 +865,7 @@ class CommonEditorOperations {
 
     if (!composer.selection!.isCollapsed) {
       // A span of content is selected. Delete the selection.
-      _deleteExpandedSelection();
+      _deleteExpandedSelection(TextAffinity.downstream);
       return true;
     }
 
@@ -1056,7 +1056,7 @@ class CommonEditorOperations {
 
     if (!composer.selection!.isCollapsed) {
       // A span of content is selected. Delete the selection.
-      _deleteExpandedSelection();
+      _deleteExpandedSelection(TextAffinity.upstream);
       return true;
     }
 
@@ -1332,7 +1332,7 @@ class CommonEditorOperations {
   ///
   /// Returns [true] if content was deleted, or [false] if no content was
   /// selected.
-  bool deleteSelection() {
+  bool deleteSelection(TextAffinity affinity) {
     if (composer.selection == null) {
       return false;
     }
@@ -1343,14 +1343,14 @@ class CommonEditorOperations {
 
     // The document selection includes a span of content. It may or may not
     // cross nodes. Either way, delete the selected content.
-    _deleteExpandedSelection();
+    _deleteExpandedSelection(affinity);
     return true;
   }
 
-  void _deleteExpandedSelection() {
+  void _deleteExpandedSelection(TextAffinity affinity) {
     // Delete the selected content.
     editor.execute([
-      const DeleteSelectionRequest(),
+      DeleteSelectionRequest(affinity: affinity),
     ]);
   }
 
@@ -1602,7 +1602,7 @@ class CommonEditorOperations {
       // Without this, the new text doesn't preserve the attributions of the replaced text.
       final composerAttributions = {...composer.preferences.currentAttributions};
 
-      _deleteExpandedSelection();
+      _deleteExpandedSelection(TextAffinity.downstream);
 
       // Restore the previous attributions.
       composer.preferences
@@ -1659,7 +1659,7 @@ class CommonEditorOperations {
     }
 
     if (!composer.selection!.isCollapsed) {
-      _deleteExpandedSelection();
+      _deleteExpandedSelection(TextAffinity.downstream);
     }
 
     final extentNodePosition = composer.selection!.extent.nodePosition;
@@ -1745,7 +1745,7 @@ class CommonEditorOperations {
       // The selection is not collapsed. Delete the selected content first,
       // then continue the process.
       editorOpsLog.finer("Deleting selection before inserting block-level newline");
-      _deleteExpandedSelection();
+      _deleteExpandedSelection(TextAffinity.downstream);
     }
 
     final newNodeId = Editor.createNodeId();
@@ -2183,7 +2183,7 @@ class CommonEditorOperations {
     //       need to be carried out in response to user input.
     _saveToClipboard(textToCut);
 
-    deleteSelection();
+    deleteSelection(TextAffinity.downstream);
   }
 
   Future<void> _saveToClipboard(String text) {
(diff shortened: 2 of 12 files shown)


angelosilvestre:
@matthew-carroll Could you take a look at this?

matthew-carroll:
> What should we do when there are multiple selected nodes, one of the in undeletable, and the user tries to delete them?

We should delete all the deletable content and place the caret at the upstream edge of the selection.

> What should we do when there are multiple selected nodes, one of the in undeletable, and the user types any key trying to replace the selected content?

We should delete the deletable content and insert the character at the upstream edge of the selection.

super_editor/lib/src/core/document.dart
```diff
@@ -457,3 +457,14 @@ abstract class NodePosition {
   /// even though [==] returns `false`.
   bool isEquivalentTo(NodePosition other);
 }
+
+/// Keys to access metadata on a [DocumentNode].
+class NodeMetadata {
+  /// Applies an [Attribution] to the node.
+  static const String blockType = 'blockType';
```
matthew-carroll, line 464: `blockType` is specific to `TextNode`s isn't it? If so, `document.dart` probably isn't the right place to define this.

Also, the idea behind the metadata is that any node type might add its own metadata. So we should have an approach to defining keys that doesn't require or imply centralization in a single package class.

super_editor/lib/src/core/document.dart
```diff
@@ -457,3 +457,14 @@ abstract class NodePosition {
...
+
+  /// Whether or not the node is deletable.
+  ///
+  /// If the node doesn't have this metadata, it is assumed to be deletable.
+  static const String deletable = 'deletable';
```
matthew-carroll, line 469: Please call this "isDeletable" to make clear that this is a boolean condition.

super_editor/lib/src/core/document.dart
```diff
@@ -457,3 +457,14 @@ abstract class NodePosition {
   /// even though [==] returns `false`.
   bool isEquivalentTo(NodePosition other);
 }
+
+/// Keys to access metadata on a [DocumentNode].
+class NodeMetadata {
+  /// Applies an [Attribution] to the node.
+  static const String blockType = 'blockType';
+
+  /// Whether or not the node is deletable.
```
matthew-carroll, line 466: Please expand on what it means to be "deletable" or not, or point the reader to a place they can find out.

super_editor/lib/src/default_editor/common_editor_operations.dart
```diff
@@ -1284,11 +1293,9 @@ class CommonEditorOperations {
 
     // Delete the selected content.
     editor.execute([
-      DeleteContentRequest(documentRange: composer.selection!),
-      ChangeSelectionRequest(
-        DocumentSelection.collapsed(position: newSelectionPosition),
-        SelectionChangeType.deleteContent,
-        SelectionReason.userInteraction,
+      DeleteContentRequest(
```
matthew-carroll, line 1296: I don't think we should change `DeleteContentRequest` to include selection changes. I think we should define a `DeleteSelectionRequest`, which implicitly would include selection knowledge and adjustment. There's nothing wrong with deleting content without any selection, and `DeleteContentRequeset` should make that possible.

super_editor/lib/src/default_editor/horizontal_rule.dart
```diff
@@ -12,7 +12,9 @@ import 'layout_single_column/layout_single_column.dart';
 class HorizontalRuleNode extends BlockNode with ChangeNotifier {
   HorizontalRuleNode({
     required this.id,
+    Map<String, dynamic>? metadata,
   }) {
+    super.metadata = metadata ?? {};
```
matthew-carroll, line 17: Is there a reason that you didn't declare `super.metadata` in the constructor parameters instead of initializing after construction?

angelosilvestre (commented):
@matthew-carroll Can you take a look again at this?

miguelcmedeiros:
@angelosilvestre if a component (e.g. horizontal rule) is not visually selectable and we have the following content in the editor:
```
First paragraph
<hr>
|Second paragraph
```
If the caret is at the beginning of the second paragraph and the user presses backspace, then the horizontal rule is deleted. 

In this case I would expect that the node is not deleted and the caret moves to the end of the first paragraph.

angelosilvestre:
> If the caret is at the beginning of the second paragraph and the user presses backspace, then the horizontal rule is >deleted.
> 
> In this case I would expect that the node is not deleted and the caret moves to the end of the first paragraph.

@miguelcmedeiros To make sure I understood. Are you saying you tried this PR and it didn't work, or is it just about the selection change when the node is not deletable?

miguelcmedeiros:
@angelosilvestre 
Yep, I tried this PR and in the case above, the horizontal rule node was deleted.

miguelcmedeiros:
> Just confirming, Did you set the NodeMetadata.isDeletable to false in the document node?

@angelosilvestre yes, I did.

angelosilvestre:
@miguelcmedeiros  Ok, thanks for trying this PR and reporting this. I'll take a look as soon as I can.

angelosilvestre:
@miguelcmedeiros I was able to reproduce it and I fixed this issue.

miguelcmedeiros:
@angelosilvestre Now it doesn't delete the horizontal rule but instead of just moving the selection, it would be great that the non-deletable nodes are simply ignored.

In this example the horizontal rule is non-deletable and can be selectable:

https://github.com/user-attachments/assets/b4c46b76-e069-4620-b1f3-77d11c5f04d3

When backspace is pressed, it moves the cursor to the first selectable node upstream, but the content ("editor") remains below the horizontal rule. 

My bad for previously not detailing out properly the requirements, but I'm gradually realizing some of these interactions as I'm testing out this PR.

matthew-carroll:
> When backspace is pressed, it moves the cursor to the first selectable node upstream, but the content ("editor") remains below the horizontal rule.

@miguelcmedeiros I'm not sure what this means.

Can you detail each interaction that you expect? E.g.:
- caret downstream and backspace is pressed
- caret upstream is delete is pressed
- selecting text above and below the node and pressing backspace (and delete)
- select the non-deletable node and dragging down to the node below and pressing backspace (and delete)
- select the non-deletable node and dragging up to the node above and pressing backspace (and delete)
- any other interaction you can think of

miguelcmedeiros:
> * caret downstream and backspace is pressed
> * caret upstream is delete is pressed

In both cases, ignore the next (above the downstream position or below the upstream position) non-deletable and non-selectable node. In the following case:
```
text 1
<non-deletable/non-selectable node>
text 2
```

If the care is at the end of the first paragraph and the DELETE is pressed, then I would expect:
```
text 1text2
<non-deletable/non-selectable node>
```

If the caret in at the beginning of the second paragraph and BACKSPACE is pressed, then I would expect:
```
text 1text2
<non-deletable/non-selectable node>
```

> * selecting text above and below the node and pressing backspace (and delete)

Similar to the scenarios above. If the following selection (`|>` and `<|`denote the selection) is made:
```
text |>1
<non-deletable/non-selectable node>
text <|2
```
If DELETE or BACKSPACE is pressed, then the result is:
```
text 2
<non-deletable/non-selectable node>
```

> * select the non-deletable node and dragging down to the node below and pressing backspace (and delete)

In the following scenario (assuming this is what was meant):
```
<non-deletable node>|>
<|text
```

Pressing BACKSPACE would do nothing. Having a selectable but non-deletable node can lead to this outcome I guess.
Pressing DELETE would move the caret to the beginning of the node below.

> * select the non-deletable node and dragging up to the node above and pressing backspace (and delete)

In the following scenario (assuming this is what was meant):
```
text|>
<|<non-deletable node>
```

Pressing DELETE would do nothing.
Pressing BACKSPACE would move the caret to the beginning of the node below.

> any other interaction you can think of

At the moment, no.

angelosilvestre:
@miguelcmedeiros Would you mind testing this again? I believe I handled all of these cases. I just have one case that I'm not sure is correct:

> 
> In the following scenario (assuming this is what was meant):
> 
> ```
> text|>
> <|<non-deletable node>
>```
> Pressing DELETE would do nothing.
> Pressing BACKSPACE would move the caret to the beginning of the node below.

Shouldn't backspace move the caret to the end of the node above here?

@matthew-carroll I had to add the affinity to the `DeleteSelectionRequest`, otherwise we won't be able to able to know if the user pressed BACKSPACE or DELETE. If this isn't desirable we could try to find another solution.

miguelcmedeiros:
> Shouldn't backspace move the caret to the end of the node above here?

@angelosilvestre you are right 👍 

Tested it out and works really well. Thanks!

super_editor/test/super_editor/supereditor_undeletable_content_test.dart
```diff
@@ -0,0 +1,1757 @@
...
+import 'package:flutter_test_runners/flutter_test_runners.dart';
+import 'package:super_editor/super_editor.dart';
+import 'package:super_editor/super_editor_test.dart';
+
+import '../test_tools.dart';
+import 'supereditor_test_tools.dart';
+
+void main() {
+  group('SuperEditor > undeletable content > prevents deletion > ', () {
+    // The desktop tests run for all desktop platforms because on mac the backspace
```
matthew-carroll, line 14: I'm confused by this comment. The comment seems to explain why we *shouldn't* run them  all the same, but then the comment says that we do run them all the same way...

super_editor/test/super_editor/supereditor_undeletable_content_test.dart
```diff
@@ -0,0 +1,1757 @@
...
+          )
+        ]);
+        await tester.pump();
+
+        // Press backspace a few times trying to delete the node.
+        await tester.pressBackspace();
+        await tester.pressBackspace();
+        await tester.pressBackspace();
+
+        // Ensure that the horizontal rule was not deleted.
```
matthew-carroll, line 118: Please be sure to verify the expected selection. It's not obvious what that selection would be given that a non-deletable node is fully selected.

super_editor/lib/src/core/document.dart
```diff
@@ -309,6 +309,10 @@ class DocumentPosition {
 
 /// A single content node within a [Document].
 abstract class DocumentNode implements ChangeNotifier {
+  DocumentNode({
```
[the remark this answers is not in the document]
angelosilvestre: To make it const we should declare it like this:

```dart
const DocumentNode({
    Map<String, dynamic>? metadata,
  }) : _metadata = metadata ?? const {};
```

Since we need to use `const {}`, doesn't it mean that any `DocumentNode` created without a `metadata` parameter will have its metadata point to the same object?

super_editor/lib/src/default_editor/common_editor_operations.dart
```diff
@@ -959,7 +968,11 @@ class CommonEditorOperations {
       return false;
     }
 
-    final nodeAfter = document.getNodeAfter(node);
+    DocumentNode? nodeAfter = document.getNodeAfter(node);
+    while (nodeAfter is BlockNode && !nodeAfter.isDeletable) {
```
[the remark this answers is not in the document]
angelosilvestre: Then we will do nothing and return after the `nodeAfter == null` check.

super_editor/lib/src/default_editor/common_editor_operations.dart
```diff
@@ -1167,13 +1188,63 @@ class CommonEditorOperations {
     return true;
   }
 
+  bool moveSelectionToEndOfFirstSelectableUpstreamNode() {
```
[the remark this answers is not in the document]
angelosilvestre: It's probably best to keep it private. But I added a dart doc.

super_editor/test/super_editor/supereditor_undeletable_content_test.dart
angelosilvestre: What I meant is that we should use `testWidgetsOnDesktop` instead of `testWidgetsOnArbitraryDesktop` to test both the key handlers and the mac selectors.

super_editor/lib/src/core/document.dart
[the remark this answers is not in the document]
matthew-carroll: I don't think that's what it would mean, but we can leave these non-const for now and then clean them up when making all nodes immutable.

super_editor/lib/src/default_editor/common_editor_operations.dart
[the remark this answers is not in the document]
matthew-carroll: But what does that mean in practice? What's the effect of that decision?

super_editor/test/super_editor/supereditor_undeletable_content_test.dart
matthew-carroll: It looks like you skipped past the context then. Your comment should essentially read like "[Instead of only testing a single arbitrary desktop,] the desktop tests run for all desktop platforms because..." But you didn't include the beginning context to understand what the rest of the comment means.

super_editor/lib/src/default_editor/multi_node_editing.dart
```diff
@@ -1042,6 +1076,139 @@ class DeleteContentCommand extends EditCommand {
   }
 }
 
+/// Deletes the selected content within the document.
+///
+/// Any non-deletable nodes within the selection are ignored.
+///
+/// The [affinity] defines the direction to where the user is trying to
```
matthew-carroll, line 1083: This comment doesn't describe why affinity is relevant. What's the difference between passing upstream or downstream? Why does it matter?

super_editor/lib/src/default_editor/multi_node_editing.dart
```diff
@@ -1042,6 +1076,139 @@ class DeleteContentCommand extends EditCommand {
   }
 }
 
+/// Deletes the selected content within the document.
+///
+/// Any non-deletable nodes within the selection are ignored.
```
matthew-carroll, line 1081: Instead of saying "ignored" please be specific. Selected, non-deletable nodes are retained without removal or modification.

super_editor/lib/src/default_editor/paragraph.dart
```diff
@@ -476,12 +477,29 @@ class CombineParagraphsCommand extends EditCommand {
...
 
-    final nodeAbove = document.getNodeBefore(secondNode);
+    DocumentNode? nodeAbove = document.getNodeBefore(secondNode);
     if (nodeAbove == null) {
       editorDocLog.info('At top of document. Cannot merge with node above.');
       return;
     }
-    if (nodeAbove.id != firstNodeId) {
+
+    // Search for the node above the second node that matches the first node.
```
matthew-carroll, line 486: It's not clear what "matches" means. Do you mean that you're searching upstream for a node that's mergeable with `secondNode`?

super_editor/lib/src/default_editor/paragraph.dart
```diff
@@ -758,6 +780,10 @@ class DeleteUpstreamAtBeginningOfParagraphCommand extends EditCommand {
     }
   }
 
+  /// Merges the selected [TextNode] with the upstream [TextNode].
+  ///
+  /// If there are non-deletable [BlockNode]s between the two [TextNode]s,
+  /// the [BlockNode]s are ignored.
```
matthew-carroll, line 786: Here again, instead of saying "ignored", please specify the technical result to avoid ambiguity.

super_editor/lib/src/default_editor/paragraph.dart
angelosilvestre: Updated.

super_editor/lib/src/default_editor/multi_node_editing.dart
angelosilvestre: Updated.

super_editor/test/super_editor/supereditor_undeletable_content_test.dart
angelosilvestre: Updated.

super_editor/lib/src/default_editor/common_editor_operations.dart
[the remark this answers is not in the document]
angelosilvestre: It means that nothing will change in the document.

PR #2378, opened 2024-10-16
[SuperEditor][mobile] Implement spellchecking support (Resolves #2353)

### LINKED ISSUE #2353: [FEATURE] - Add spellcheck support for mobile platforms
We are working on spellcheck support for mac, which includes decorate misspelled words and show a popover with suggestions.

For reference, here's some references of spellchecking in other apps:

Google Docs (Android)

https://github.com/user-attachments/assets/5382a23e-e425-4646-b452-7cc72b40ddce

Microsoft Word (Android)

https://github.com/user-attachments/assets/d73d1803-2eda-41fa-93cc-f52aaf6dc585

Google Docs (iOS)

https://github.com/user-attachments/assets/382fd930-4585-4ece-b33b-2cb1d35a65d5

Notion (iOS)

https://github.com/user-attachments/assets/9d155f22-9364-495a-bbc0-6f1c8f8118f2

Obsidian (iOS)

https://github.com/user-attachments/assets/4e65ce52-3b3d-40eb-9a87-b3be6ea43a27

Apple Notes (iOS)

https://github.com/user-attachments/assets/35c601dc-0b0c-4607-ab18-d81779ce7565

angelosilvestre:
[SuperEditor][mobile] Implement spellchecking support. Resolves #2353

Our current spellchecker plugin only works on macOS. This PR adds support for spellchecking on Android and iOS on the spellchecker plugin.

Currently, the desktop implementation is showing the suggestions toolbar whenever the currently selection sits within a misspelled word with available suggestions. However, this wouldn't work for mobile because, on mobile platforms, the suggestions toolbar is displayed when tapping on a misspelled word, but not when dragging the handle over a misspelled word. This creates an implementation problem, because a gesture (tap up) should trigger the suggestions popover and we don't have an extension point for plugins to do that.

To solve this issue, this PR adds the ability for plugins to install a delegate for tap handlers. The plugin now has the opportunity to handle the tap the way it wants, preventing our gesture system for handling them.

The tap delegate shows/hides the suggestions popover by interacting with a `SpellCheckerPopoverController`, a class that exposes the spellchecker popover functionality, which uses a `SpellCheckerPopoverDelegate` (usually a document layer) to effectively display the popover toolbar.

This PR also add an option to temporarily disable selection handles. This is needed because, when tapping on a misspelled word, we want to select the whole word, but prevent the selection handles from being displayed.

On mobile platforms, the selection color is different when the suggestions popover is visible. Currently, we don't have an option to override the selection color. On our style pipeline, the selection styler is always last, so even if we use a custom style phase, the selection styler always resets the selection color.

To solve that, this PR exposes an `appendedStylePhases` property on `SuperEditor`, so we can add style phases that are positioned after the selection styler, adding the ability to override the selection color.

This PR adds default implementations for iOS-style and Android-style suggestions popovers.

https://github.com/user-attachments/assets/b5931600-cc61-4118-ab00-11a9036b94af

[android_spellcheck.webm](https://github.com/user-attachments/assets/3638e20e-bf09-4d7c-a2fe-60726fcd138b)


matthew-carroll:
@angelosilvestre - Is this the PR you wanted me to review? If so, there should still be a description explaining what you did and how you did it - that's what directs the PR review. Also, whenever you want a review, please tag me in the PR.

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
```diff
@@ -427,6 +429,8 @@ class AndroidDocumentTouchInteractor extends StatefulWidget {
   /// a link when the user taps on text with a link attribution.
   final ContentTapDelegate? contentTapHandler;
 
+  final SpellCheckerPopoverController? spellCheckerPopoverController;
```
matthew-carroll, line 432: I just pulled up this PR, but the first thing that I'm wondering is why the spell check controller wasn't added to the area with the handles, caret, magnifier, and editing toolbar? That area already has one type of toolbar - why wouldn't this controller/behavior belong there, too?

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
```diff
@@ -1009,6 +1014,7 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
       } else {
         // The user tapped somewhere else in the document. Hide the toolbar.
         _controlsController!.hideToolbar();
+        widget.spellCheckerPopoverController?.show(widget.selection.value!);
```
matthew-carroll, line 1017: Why are we blindly telling the spell checker popover to show itself? I don't see any conditional check here for a mis-spelled word...

super_editor/lib/src/default_editor/spelling_and_grammar/spell_checker_popover_controller.dart
```diff
@@ -0,0 +1,35 @@
+import 'package:super_editor/src/core/document_selection.dart';
+
+/// Shows/hides a popover with spelling suggestions.
+class SpellCheckerPopoverController {
+  SpellCheckerPopoverDelegate? _delegate;
+
+  /// Attaches this controller to a delegate that knows how to
+  /// show a popover with spelling suggestions.
+  void attach(SpellCheckerPopoverDelegate delegate) {
```
matthew-carroll, line 9: Did we use an attach/detach approach for the existing mobile editing toolbar, too?

super_editor/lib/src/default_editor/spelling_and_grammar/spell_checker_popover_controller.dart
```diff
@@ -0,0 +1,35 @@
+import 'package:super_editor/src/core/document_selection.dart';
+
+/// Shows/hides a popover with spelling suggestions.
```
matthew-carroll, line 3: There should be a lot more information in this Dart Doc. Imagine reading this as someone who needs to use this - there's no guiding information about where/when/how.

super_editor/lib/src/default_editor/spelling_and_grammar/spell_checker_popover_controller.dart
```diff
@@ -0,0 +1,35 @@
...
+    _delegate?.showSuggestionsForWordAt(wordRange);
+  }
+
+  /// Hides the spelling suggestions popover.
+  void hide() {
+    _delegate?.hideSuggestionsPopover();
+  }
+}
+
+abstract class SpellCheckerPopoverDelegate {
```
matthew-carroll, line 29: Missing Dart Doc. This should have a lot of info about who is supposed to implement this.

super_editor/lib/src/default_editor/spelling_and_grammar/spell_checker_popover_controller.dart
```diff
@@ -0,0 +1,35 @@
...
+  /// Shows spelling suggestions for the word at [wordRange].
```
matthew-carroll, line 30: What if the document changes? What if the user's selection changes? Please fully describe expected behaviors.

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -127,6 +128,7 @@ class SuperEditor extends StatefulWidget {
     this.selectorHandlers,
     this.gestureMode,
     this.contentTapDelegateFactory = superEditorLaunchLinkTapHandlerFactory,
+    this.spellCheckerPopoverController,
```
matthew-carroll, line 131: Why is this being passed into `SuperEditor`? We don't pass anything for the mobile editing toolbars, magnifier, handles, etc, do we?

matthew-carroll (commented):
I did a partial review. Before reviewing the rest, I think we need to align on why this spell check controller was injected in the places that it was. When we talked about this online, I mentioned treating this spell check toolbar the same way we treat the mobile editing toolbars, but as far as I can tell, that's not what this PR is currently doing.

angelosilvestre:
> @angelosilvestre - Is this the PR you wanted me to review? If so, there should still be a description explaining what you did and how you did it - that's what directs the PR review. Also, whenever you want a review, please tag me in the PR.

This PR is meant to review the controller usage only. I should have added that to the description, sorry.

angelosilvestre:
> I did a partial review. Before reviewing the rest, I think we need to align on why this spell check controller was injected in the places that it was. When we talked about this online, I mentioned treating this spell check toolbar the same way we treat the mobile editing toolbars, but as far as I can tell, that's not what this PR is currently doing.

@matthew-carroll  I moved the controller to `SuperEditorIosControlsController` and `SuperEditorAndroidControlsController`. Can you take a look to see if that's how we should proceed?

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
```diff
@@ -312,6 +314,16 @@ class SuperEditorAndroidControlsController {
   /// If [toolbarBuilder] is `null`, a default Android toolbar is displayed.
   final DocumentFloatingToolbarBuilder? toolbarBuilder;
 
+  final SpellCheckerPopoverController? spellCheckerPopoverController;
```
matthew-carroll, line 317: The existing editing toolbar just takes a builder. Is there a reason that this one needs a controller?

super_editor/lib/src/default_editor/spelling_and_grammar/spell_checker_popover_controller.dart
angelosilvestre: Added some docs, can you take a look to see what else should be detailed?

super_editor/lib/src/default_editor/spelling_and_grammar/spell_checker_popover_controller.dart
angelosilvestre: Added some docs, can you take a look to see what else should be detailed?

super_editor/lib/src/default_editor/spelling_and_grammar/spelling_and_grammar_styler.dart
```diff
@@ -34,6 +35,15 @@ class SpellingAndGrammarStyler extends SingleColumnLayoutStylePhase {
     markDirty();
   }
 
+  /// Whether or not we should to override the default selection color with [selectionHighlightColor].
+  ///
+  /// On mobile platforms, when the suggestions popover is opened, the selected text uses a different
```
matthew-carroll, line 40: This line is a bit confusing. Does this line mean that this property will be `true` on mobile?

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -847,7 +861,13 @@ class SuperEditorState extends State<SuperEditor> {
           getDocumentLayout: () => editContext.documentLayout,
           selectionChanges: editContext.composer.selectionChanges,
           selectionNotifier: editContext.composer.selectionNotifier,
-          contentTapHandler: _contentTapDelegate,
+          contentTapHandlers: [
+            if (_contentTapDelegate != null) //
```
matthew-carroll, line 865: The term "delegate" probably needs to be replaced with "handler" because typically there's only one delegate. There might be many handlers.

Given that SuperEditor can take a list of handlers, is there a reason that plugins only get one?

super_editor/lib/src/infrastructure/platforms/android/android_document_controls.dart
```diff
@@ -311,6 +314,12 @@ class AndroidControlsDocumentLayerState
     });
   }
 
+  void _onSelectionHandlesAllowedChange() {
+    setState(() {
+      //
```
matthew-carroll, line 319: Please specify what's being rebuilt and why

super_editor/lib/src/infrastructure/platforms/android/android_document_controls.dart
```diff
@@ -319,6 +328,11 @@ class AndroidControlsDocumentLayerState
       return null;
     }
 
+    if (_controlsController!.areSelectionHandlesAllowed.value == false) {
```
matthew-carroll, line 331: Is this value nullable? It seems strange that `areSelectionHandlesAllowed` would ever be `null`...

super_editor_spellcheck/example/lib/main_super_editor_spellcheck.dart
```diff
@@ -98,19 +119,26 @@ class _SuperEditorSpellcheckScreenState extends State<_SuperEditorSpellcheckScre
...
-          ],
+      body: SuperEditorAndroidControlsScope(
+        controller: _androidControlsController,
+        child: SuperEditorIosControlsScope(
+          controller: _iosControlsController,
+          child: SuperEditor(
+            autofocus: true,
+            editor: _editor,
+            appendedStylePhases: [
+              _spellingAndGrammarPlugin.styler,
```
matthew-carroll, line 130: Shouldn't this be added automatically as part of the plugin?

angelosilvestre:
@matthew-carroll This should wait on https://github.com/superlistapp/super_editor/pull/2397, because we will have some conflicts.

super_editor/lib/src/default_editor/document_gestures_mouse.dart
```diff
@@ -118,7 +119,11 @@ class _DocumentMouseInteractorState extends State<DocumentMouseInteractor> with
     widget.autoScroller
       ..addListener(_updateDragSelection)
       ..addListener(_updateMouseCursorAtLatestOffset);
-    widget.contentTapHandler?.addListener(_updateMouseCursorAtLatestOffset);
+    if (widget.contentTapHandlers != null) {
+      for (final handler in widget.contentTapHandlers!) {
+        handler.addListener(_updateMouseCursorAtLatestOffset);
```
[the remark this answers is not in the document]
angelosilvestre: It's used by the `SuperEditorLaunchLinkTapHandler` to notify that it wants to re-evaluate the mouse cursor when switching between the interaction mode and regular mode.

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
```diff
@@ -784,6 +807,23 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
...
 
+    editorGesturesLog.info("Tap down on document");
+    final docOffset = _getDocumentOffsetFromGlobalOffset(details.globalPosition);
+    editorGesturesLog.fine(" - document offset: $docOffset");
+    final docPosition = _docLayout.getDocumentPositionNearestToOffset(docOffset);
+    editorGesturesLog.fine(" - tapped document position: $docPosition");
+
+    if (widget.contentTapHandlers != null && docPosition != null) {
+      for (final handler in widget.contentTapHandlers!) {
+        final result = handler.onTap(docPosition);
```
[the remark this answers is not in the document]
angelosilvestre: Updated after the scrolling rework and the https://github.com/superlistapp/super_editor/pull/2397. I guess it should be correct now.

super_editor/lib/src/default_editor/spelling_and_grammar/spelling_and_grammar_styler.dart
angelosilvestre: This is a transitive state, while the suggestions popover is open this will be `true`, otherwise it will be `false`.

super_editor/lib/src/default_editor/super_editor.dart
angelosilvestre: > The term "delegate" probably needs to be replaced with "handler" because typically there's only one delegate. There might be many handlers.

Updated on a previous PR.

super_editor/lib/src/infrastructure/platforms/android/android_document_controls.dart
angelosilvestre: Updated.

super_editor/lib/src/infrastructure/platforms/android/android_document_controls.dart
angelosilvestre: Updated.

angelosilvestre:
@matthew-carroll I also added the Android and iOS platforms to the spellcheck demo app.

super_editor_spellcheck/lib/src/super_editor/spelling_error_suggestion_overlay.dart
```diff
@@ -413,14 +542,41 @@ class DesktopSpellingSuggestionToolbar extends StatefulWidget {
...
+    widget.editor.document.removeListener(_onDocumentChange);
+    super.dispose();
+  }
+
   void _applySpellingFix(String replacement) {
     widget.editor.fixMisspelledWord(widget.selectedWordRange!, replacement);
+    widget.closeToolbar();
+  }
+
+  void _onDocumentChange(DocumentChangeLog changeLog) {
```
matthew-carroll, line 578: Notice that `_onDocumentChange()` is referenced from all the lifecycle methods, but `_applySpellingFix()` isn't. Please switch the order of these two methods.

super_editor_spellcheck/lib/src/super_editor/spelling_error_suggestion_overlay.dart
```diff
@@ -513,3 +669,273 @@ class _DesktopSpellingSuggestionToolbarState extends State<DesktopSpellingSugges
     }
   }
 }
+
+/// A spelling suggestion toolbar, designed for the Android platform,
+/// which displays a vertical list alternative spellings for a given mis-spelled
```
matthew-carroll, line 674: I don't know what "a vertical list alternative spellings" is.

super_editor_spellcheck/lib/src/super_editor/spelling_error_suggestion_overlay.dart
```diff
@@ -513,3 +669,273 @@ class _DesktopSpellingSuggestionToolbarState extends State<DesktopSpellingSugges
...
+    required this.selectedWordBounds,
+    required this.closeToolbar,
+  });
+
+  final FocusNode editorFocusNode;
+  final Object? tapRegionId;
+  final Editor editor;
+  final DocumentRange selectedWordRange;
+  final List<String> suggestions;
+  final Rect selectedWordBounds;
```
matthew-carroll, line 696: The ordering of this property seems to be inconsistent throughout the code.

super_editor_spellcheck/lib/src/super_editor/spelling_error_suggestion_overlay.dart
```diff
@@ -513,3 +669,273 @@ class _DesktopSpellingSuggestionToolbarState extends State<DesktopSpellingSugges
...
+        child: Text(
+          title,
+          style: const TextStyle(fontSize: 12),
+        ),
+      ),
+    );
+  }
+}
+
+const iOSToolbarDarkBackgroundColor = Color(0xFF333333);
```
matthew-carroll, line 934: Should these be public or private?

super_editor_spellcheck/lib/src/super_editor/spelling_error_suggestion_overlay.dart
```diff
@@ -371,22 +452,69 @@ typedef SpellingErrorSuggestionToolbarBuilder = Widget Function(
   required DocumentRange selectedWordRange,
   required List<String> suggestions,
   required VoidCallback onCancelPressed,
+  required VoidCallback closeToolbar,
+  required Rect selectedWordBounds,
```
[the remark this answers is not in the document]
angelosilvestre: Updated.

super_editor_spellcheck/lib/src/super_editor/spelling_error_suggestion_overlay.dart
```diff
@@ -371,22 +452,69 @@ typedef SpellingErrorSuggestionToolbarBuilder = Widget Function(
...
+        selectedWordBounds: selectedWordBounds,
+        closeToolbar: closeToolbar,
+      );
+    case TargetPlatform.android:
+      return AndroidSpellingSuggestionToolbar(
+        editorFocusNode: editorFocusNode,
+        editor: editor,
+        selectedWordRange: selectedWordRange,
+        suggestions: suggestions,
+        selectedWordBounds: selectedWordBounds,
```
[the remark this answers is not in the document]
angelosilvestre: I removed from the Android implementation. We need it for iOS to configure the focal point of the cupertino toolbar.

super_editor_spellcheck/lib/src/super_editor/spelling_error_suggestion_overlay.dart
angelosilvestre: Updated.

super_editor_spellcheck/lib/src/super_editor/spelling_and_grammar_plugin.dart
```diff
@@ -1,9 +1,12 @@
...
 
 import 'package:flutter/foundation.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
 import 'package:follow_the_leader/follow_the_leader.dart';
 import 'package:super_editor/super_editor.dart';
 import 'package:super_editor_spellcheck/src/platform/spell_checker.dart';
+import 'package:super_editor_spellcheck/src/super_editor/spell_checker_popover_controller.dart';
 import 'package:super_editor_spellcheck/src/super_editor/spelling_error_suggestion_overlay.dart';
 import 'package:super_editor_spellcheck/src/super_editor/spelling_error_suggestions.dart';
```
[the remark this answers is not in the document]
angelosilvestre: Added a dart doc to the constructor to explain the parameters.

super_editor_spellcheck/lib/src/super_editor/spell_checker_popover_controller.dart
```diff
@@ -0,0 +1,118 @@
...
+    _delegate = delegate;
+  }
+
+  /// Detaches this controller from the delegate.
+  ///
+  /// This controller can't show/hide the popover while detached from a delegate.
+  ///
+  /// A [SpellCheckerPopoverDelegate] must call this method after
+  /// being unmounted from the widget tree.
+  void detach() {
```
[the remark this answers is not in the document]
matthew-carroll: That doesn't really address my question. delegate1 doesn't know that delegate2 attached itself, so delegate1 never knew to call `detach()`.

super_editor_spellcheck/lib/src/super_editor/spelling_and_grammar_plugin.dart
```diff
@@ -328,4 +415,281 @@ class SpellingAndGrammarReaction implements EditReaction {
...
+      return TapHandlingInstruction.continueHandling;
+    }
+
+    final tapPosition = details.documentLayout.getDocumentPositionNearestToOffset(details.layoutOffset);
+    if (tapPosition == null) {
+      return TapHandlingInstruction.continueHandling;
+    }
+
+    final spelling = popoverController.findSuggestionsForWordAt(DocumentSelection.collapsed(position: tapPosition));
+    if (spelling == null || spelling.suggestions.isEmpty) {
```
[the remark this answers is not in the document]
matthew-carroll: How about we add a `bool` to `SpellingErrorSuggestion` for `isWordMisspelled`?

super_editor_spellcheck/lib/src/super_editor/spelling_error_suggestion_overlay.dart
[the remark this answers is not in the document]
matthew-carroll: Right, that's demonstrating my point. So iOS happens to need some layout info, the others don't. What happens when one of them needs yet some other layout info? Will we just add random properties endlessly to the function signature?

Similar to another place where we had to deal with layout querying, would it make sense to pass the `DocumentLayout` into this builder? Or, perhaps, would it make sense to setup a `Leader`/`Follower` relationship somewhere? How do we solve similar things elsewhere?

super_editor_spellcheck/lib/src/super_editor/spell_checker_popover_controller.dart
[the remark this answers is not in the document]
angelosilvestre: Added callbacks for when the delegate is attached/detached.

super_editor_spellcheck/lib/src/super_editor/spelling_and_grammar_plugin.dart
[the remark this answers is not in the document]
angelosilvestre: But doesn't `SpellingErrorSuggestion` always mean that the word is misspelled?

super_editor_spellcheck/lib/src/super_editor/spelling_error_suggestion_overlay.dart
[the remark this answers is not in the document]
angelosilvestre: Updated.

matthew-carroll:
On our call we discovered that behavior in iOS 17 is different than iOS 16.

On iOS 17:
 * When the user taps a mis-spelled word, the mis-spelled word becomes fully selected. The selection color is the normal selection color. No handles are shown.
 * Pressing and dragging from either end of the word begins to drag that end of the selection. The side that's being dragged shows a non-blinking caret (instead of a handle). The other side of the selection shows a handle.
 * When the user stops dragging, the caret turns into a handle.

We looked at Android but found inconsistent behaviors between Keep, Gmail, SMS, etc.

Gmail and Calendar do the following:
 * Places the caret where tapped
 * Hides the caret
 * Highlights the misspelled word in red
 * Doesn't show any handles
 * Shows a popover list of spelling suggestions - this list is vertical
 * It seems like the suggestions are sitting in an invisible fullscreen modal. Tapping anywhere else causes the suggestions to close, and the previous caret position is restored. Even tapping on buttons while the suggestions are open doesn't result in the buttons being activated.

We should ensure that these behaviors work by default with this plugin.

angelosilvestre:
@matthew-carroll I updated the behavior for iOS:

https://github.com/user-attachments/assets/bfbca2d9-db68-4e9b-b721-91b6cb5e25f5

angelosilvestre:
@matthew-carroll I updated the Android behavior:

[android_suggestions.webm](https://github.com/user-attachments/assets/d699e5f9-3e49-4865-a916-8b9b0f98e9f9)

super_editor_spellcheck/example/lib/main_super_editor_spellcheck.dart
```diff
@@ -65,12 +67,23 @@ class _SuperEditorSpellcheckScreen extends StatefulWidget {
...
 
   @override
   void initState() {
     super.initState();
 
+    _iosControlsController = SuperEditorIosControlsController();
+    _androidControlsController = SuperEditorAndroidControlsController();
+
+    _spellingAndGrammarPlugin = SpellingAndGrammarPlugin(
+      iosControlsController: _iosControlsController,
```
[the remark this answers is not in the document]
angelosilvestre: They are not required anymore. We only have assertions in the constructor to ensure that, when running on Android or iOS, the controller is provided.

matthew-carroll (approved):
LGTM

PR #2391, opened 2024-10-31
[SuperTextField][iOS] Fix selection highlight (Resolves #2346)

### LINKED ISSUE #2346: [SuperTextField][iOS] - Selection highlight doesn't appear
When using a `SuperIOSTextField` (not a SuperTextField), the selection highlight doesn't appear.

Oddly, when using a `SuperTextField` on iOS, the selection highlight does appear.

Screenshot 2024-09-25 at 4 14 02 PM

```dart
return DecoratedBox(
      decoration: BoxDecoration(
        border: Border.all(color: Colors.grey),
        borderRadius: BorderRadius.circular(4),
        // color: const Color(0xFF222222),
      ),
      child: SuperIOSTextField(
        padding: const EdgeInsets.all(12),
        caretStyle: CaretStyle(color: Colors.red),
        selectionColor: defaultSelectionColor,
        handlesColor: Colors.red,
          return defaultTextFieldStyleBuilder(attributions).copyWith(
            color: Colors.white,
            fontSize: 18,
          );
        },
        hintBuilder: (_) {
          return Text(
            "Enter text and open toolbar",
            style: TextStyle(
              color: Colors.grey,
              fontSize: 18,
            ),
          );
        },
        popoverToolbarBuilder: iOSSystemPopoverTextFieldToolbarWithFallback,
      ),
    );
```

angelosilvestre:
[SuperTextField][iOS] Fix selection highlight. Resolves #2346

When using a `SuperIOSTextField` (not a `SuperTextField`), the selection highlight doesn't appear. This only happens if we don't provide a `textController` in `SuperIOSTextField` constructor.

The issue is that, in the places we check for an existing selection to add the highlights, we are always checking for the  `textController` provided by the widget constructor instead of using the internal controller (which uses the constructor's `textController` if provided).

This PR fixes the issue by making `SuperIOSTextField` always use the internal `textController`.

I'm not sure the text is good, since we are looking at too much implementation details. Maybe we should just use a golden test instead.

### DIFF
--- a/super_editor/lib/src/super_textfield/ios/ios_textfield.dart
+++ b/super_editor/lib/src/super_textfield/ios/ios_textfield.dart
@@ -629,17 +629,17 @@ class SuperIOSTextFieldState extends State<SuperIOSTextField>
         return Stack(
           clipBehavior: Clip.none,
           children: [
-            if (widget.textController?.selection.isValid == true)
+            if (_textEditingController.selection.isValid == true)
               // Selection highlight beneath the text.
               TextLayoutSelectionHighlight(
                 textLayout: textLayout,
                 style: SelectionHighlightStyle(
                   color: widget.selectionColor,
                 ),
-                selection: widget.textController?.selection,
+                selection: _textEditingController.selection,
               ),
             // Underline beneath the composing region.
-            if (widget.textController?.composingRegion.isValid == true && widget.showComposingUnderline)
+            if (_textEditingController.composingRegion.isValid == true && widget.showComposingUnderline)
               TextUnderlineLayer(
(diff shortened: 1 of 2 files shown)


matthew-carroll:
> I'm not sure the text is good, since we are looking at too much implementation details. Maybe we should just use a golden test instead.

Sure we can use a golden test. Did we not already have a golden test that included a selection for an iOS text field?

angelosilvestre:
@matthew-carroll We have for `SuperTextField`, but not for `SuperIOSTextField`.

matthew-carroll (approved):
LGTM

PR #2396, opened 2024-11-01
WIP: [SuperEditor] Add the ability to add an empty paragraph when tapping a non-selectable node at the end of the document (Resolves #2395)

### LINKED ISSUE #2395: [SuperEditor] - Add empty paragraph at the end of document when tapping on the empty area below it
**Your Info**
Superlist

**User Behavior**
When a user taps on the empty area below the document content, it will place the selection at the nearest position, ignoring any node that is configured as being visually non-selectable.

Instead, we would like to add an empty paragraph at the end. If the last node is already empty paragraph, then just place the selection there.

Since one option is not necessarily better than the other, just different, maybe `SuperEditor` can offer a way for the client code to implement a custom behaviour or offer a configuration field when the client code pick one or the other option.

**Feature Value**
The behaviour as described in this ticket is usual in tools like Notion, and is for users used to such tools the expected behaviour.

angelosilvestre:
[SuperEditor] Add the ability to add an empty paragraph when tapping a non-selectable node at the end of the document .Resolves #2395

When tapping at a non-selectable node, the editor moves the selection to the nearest selectable node.

This PR adds the ability to add a new empty paragraph when the user taps at a non-selectable node at the end of the document.

This PR isn't complete. The goal is to decide if that's how we want to implement it. We talked about using a reaction for that,  but this special handling of non-selectable nodes happens before issuing a selection change request. Also, checking if the node is selectable requires access to the document layout to get the component for the node. We don't have access to that in reactions.

This PR introduces a `SelectNearestSelectableNodeRequest`, that takes a `targetNodeId` (the id of the non-selectable node that the user tapped), and the nearest selection.

In the command, we can choose to add a new node if the node with `targetNodeId` is the last node of the document.

Open question: How should we make this configurable? Create two commands, one that only changes the selection and other that adds an empty paragraph if the non-selectable node is the last in the document? Create a constructor parameter in `SelectNearestSelectableNodeCommand` to configure this? Another option?

### DIFF
--- a/super_editor/lib/src/core/document_composer.dart
+++ b/super_editor/lib/src/core/document_composer.dart
@@ -4,6 +4,9 @@ import 'dart:ui';
 import 'package:attributed_text/attributed_text.dart';
 import 'package:flutter/foundation.dart';
 import 'package:super_editor/src/core/document.dart';
+import 'package:super_editor/src/default_editor/multi_node_editing.dart';
+import 'package:super_editor/src/default_editor/paragraph.dart';
+import 'package:super_editor/src/default_editor/text.dart';
 import 'package:super_editor/src/infrastructure/_logging.dart';
 import 'package:super_editor/src/infrastructure/pausable_value_notifier.dart';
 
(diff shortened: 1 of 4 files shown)


miguelcmedeiros:
@angelosilvestre for our use case, when a non-selectable node is tapped it shouldn't create an empty paragraph beneath or move to the nearest available selection. The current behavior of doing nothing in this situation is just fine.

The request in #2395 is about what happens when the empty area in the editor below the content is tapped. In that ticket I mentioned what happens when the last node is a non-selectable node since it's the reason for the ticket in the first place, but if the last node is selectable, the request would be to still add an empty paragraph at the end. Unless there's already one, and in this case, just move the selection there.

angelosilvestre:
@miguelcmedeiros I guess I read the ticket too  quickly. I'll update the PR as soon as possible.

angelosilvestre:
I'm closing this PR in favor of https://github.com/superlistapp/super_editor/pull/2397, because the implementation is completely different.

PR #2397, opened 2024-11-05
[SuperEditor] Add tap handler to add an empty paragraph when tapping below the end of document (Resolves #2395)

### LINKED ISSUE #2395: [SuperEditor] - Add empty paragraph at the end of document when tapping on the empty area below it

angelosilvestre:
[SuperEditor] Add tap handler to add an empty paragraph when tapping below the end of document. Resolves #2395 

Currently, tapping below the end of the document places the caret at the end of the last selectable component.

This PR modifies the `ContentTapDelegate` to include more detailed information about the tap, including whether or not the tap happened below the last node of above the first one. 

This PR also introduces a `SuperEditorAddEmptyParagraphTapHandler`, so apps can choose to add an empty paragraph to the end of the document instead.

This is a breaking change, but resolving it is pretty simple.

Instead of using booleans to describe if the gesture happened above the first node/after the last node, we could use an enum, because it won't make sense to have both `isGestureAboveStartOfDocument` and `isGestureBelowEndOfDocument` set to true.

### DIFF
--- a/super_editor/lib/src/default_editor/super_editor.dart
+++ b/super_editor/lib/src/default_editor/super_editor.dart
@@ -20,6 +20,7 @@ import 'package:super_editor/src/default_editor/document_gestures_touch_ios.dart
 import 'package:super_editor/src/default_editor/document_scrollable.dart';
 import 'package:super_editor/src/default_editor/layout_single_column/_styler_composing_region.dart';
 import 'package:super_editor/src/default_editor/list_items.dart';
+import 'package:super_editor/src/default_editor/multi_node_editing.dart';
 import 'package:super_editor/src/default_editor/tasks.dart';
 import 'package:super_editor/src/infrastructure/_logging.dart';
 import 'package:super_editor/src/infrastructure/content_layers.dart';
@@ -1682,7 +1683,8 @@ class SuperEditorLaunchLinkTapHandler extends ContentTapDelegate {
   }
 
   @override
-  TapHandlingInstruction onTap(DocumentPosition tapPosition) {
+  TapHandlingInstruction onTap(DocumentTapDetails details) {
+    final tapPosition = details.position;
     if (!composer.isInInteractionMode.value) {
       // The editor isn't in "interaction mode". We don't want to allow
       // users to open links by tapping on them.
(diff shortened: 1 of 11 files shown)


matthew-carroll:
> This is a breaking change, but resolving it is pretty simple.

When creating a breaking change with a known solution, please demonstrate that change in the PR so that readers can come here and copy the fix.

super_editor/lib/src/infrastructure/document_gestures_interaction_overrides.dart
```diff
@@ -12,19 +12,40 @@ abstract class ContentTapDelegate with ChangeNotifier {
...
+  });
+
+  /// The position in the document where the gesture occurred.
+  ///
+  /// When there is no content at the gesture location, this position
+  /// holds the nearest position in the document.
+  final DocumentPosition position;
+
+  /// Whether the gesture occurred above the first node in the document.
+  final bool isGestureAboveStartOfDocument;
```
matthew-carroll, line 43: This is definitely far too specific to this one particular goal. There could be 100 different queries that apps might want. We'd end up with 100 properties.

Can you please propose an alternate approach? Can we pass a reference to the document layout and let the implementer check whatever they want?

angelosilvestre: I pushed a alternative approach, placing the checks directly on the handler.

matthew-carroll:
I'm still curious about the concept of passing the document layout. Can you think of any reason why we wouldn't want to pass the layout with the event?

We're currently reporting the "tap position" in terms of the tapped document node. But this is already a bit of overreach on our end. What if the handler wants to find the "nearest position"? Or, if we're currently looking up the nearest position, what if the handler only wants to consider positions that are actually touched?

So I'm wondering, what if we pass the document layout and the global offset? Can you think of any downsides from that? Any hurdles that might get in our way?

angelosilvestre:
> Can you think of any downsides from that? Any hurdles that might get in our way?

I don't have anything against passing the document layout.

Are you saying we should remove `tapPosition` and let implementers obtain this information directly from the `DocumentLayout`?

matthew-carroll:
> Are you saying we should remove tapPosition and let implementers obtain this information directly from the DocumentLayout?

Yeah, I think we can leave it up to the client to decide what info they wanna pull out of the layout.

angelosilvestre:
@matthew-carroll Updated.

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -1723,3 +1729,82 @@ class SuperEditorLaunchLinkTapHandler extends ContentTapDelegate {
     return null;
   }
 }
+
```
matthew-carroll, line 1732: Putting this comment here, but it's related to the delegate, itself.

Should we switch to a chain of responsibility so that multiple delegates can be provided? If we don't, and a developer wants multiple situations to be considered, then that developer will have to implement his own chain of responsibility.

Is there any reason that it's important for us to have only a single delegate?

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -1723,3 +1729,82 @@ class SuperEditorLaunchLinkTapHandler extends ContentTapDelegate {
...
+SuperEditorAddEmptyParagraphTapHandler superEditorAddEmptyParagraphTapHandlerFactory(SuperEditorContext editContext) =>
```
matthew-carroll, line 1733: This probably doesn't belong in super_editor.dart - we should create a more appropriate source file for such things, or add it to a related existing file.

super_editor/lib/src/default_editor/super_editor.dart
angelosilvestre: Updated to multiple delegates. We would already need multiple delegates for the spellcheck PR.

super_editor/lib/src/default_editor/super_editor.dart

matthew-carroll:
@miguelcmedeiros would you like to sign off on this PR before we merge it? I think it generally looks good to me, but would be good for you to verify that it works for you.

miguelcmedeiros (approved):
LGTM

PR #2406, opened 2024-11-13
[super_editor_markdown] Fix compilation error

angelosilvestre:
[super_editor_markdown] Fix compilation error

Now `CommonEditorOperations.getDocumentPositionAfterExpandedDeletion` returns a nullable type. This caused a compilation error in `pasteMarkdown`.

This PR fixes that issue.

### DIFF
--- a/super_editor_markdown/lib/src/super_editor_paste_markdown.dart
+++ b/super_editor_markdown/lib/src/super_editor_paste_markdown.dart
@@ -40,7 +40,7 @@ Future<void> pasteMarkdown({
   required Document document,
   required DocumentComposer composer,
 }) async {
-  DocumentPosition pastePosition = composer.selection!.extent;
+  DocumentPosition? pastePosition = composer.selection!.extent;
 
   // Delete all currently selected content.
   if (!composer.selection!.isCollapsed) {
@@ -48,6 +48,10 @@ Future<void> pasteMarkdown({
       document: document,
       selection: composer.selection!,
     );
+
+    if (pastePosition == null) {
+      return;
+    }
 
     // Delete the selected content.
     editor.execute([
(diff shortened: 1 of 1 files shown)


super_editor_markdown/lib/src/super_editor_paste_markdown.dart
```diff
@@ -49,6 +49,10 @@ Future<void> pasteMarkdown({
       selection: composer.selection!,
     );
 
+    if (pastePosition == null) {
```
matthew-carroll, line 52: Shouldn't this guard clause appear directly below the `pastePosition` assignment? In what situation can we paste without a selection?

angelosilvestre: The first assignment is assumed to be always non-null. I added the nullability annotation because `CommonEditorOperations.getDocumentPositionAfterExpandedDeletion` can return `null`.

angelosilvestre: > Aside from the return type, in what situation would deletion return a null selection?

It returns `null` if there are only non-deletable nodes in the selection.

angelosilvestre: Only if all nodes are non-deletable.

angelosilvestre: Updated.

matthew-carroll (approved):
LGTM

matthew-carroll:
@angelosilvestre something went wrong with the cherry pick here. Can you do it manually?

angelosilvestre:
@matthew-carroll The changes were already included in https://github.com/superlistapp/super_editor/pull/2422

PR #2407, opened 2024-11-15
Update index.jinja

suragch:
Fix text to copy when the button is pressed. It should copy the shown text. An even better solution would be to make a single source of truth for both the shown text and the copied text.

This PR also fixes a broken link to the SuperEditor Quickstart guides.


doc/website/source/index.jinja
```diff
@@ -213,8 +213,8 @@ layout:
 
     <script>
       function copyActivationToClipboard() {
-        navigator.clipboard.writeText("dart pub global activate static_shock_cli");
+        navigator.clipboard.writeText("super_editor: ^0.2.7");
```
matthew-carroll, line 216: This should be injectable. Can you locate the code that shows the value in the header? The value should be coming from the GitHub plugin, and injected through a Jinja variable. You should be able to use that same injection here for "^0.2.7".

suragch: OK, the version number should be injected now as per my latest commit.

matthew-carroll (approved):
LGTM - Thanks!

matthew-carroll:
@suragch - Whenever we merge something into `main` we also need to cherry pick that into `stable`. Can you put up a PR pointing at `stable` that includes a cherry pick of this commit from `main`?

suragch:
@matthew-carroll The `stable` branch doesn't have `doc/website/source/index.jinja` at all, so I'm getting a conflict with the cherry pick. What is the difference between `main` and `stable` here?

matthew-carroll:
Ok, I guess I never cherry picked the original docs. In that case, don't worry about cherry picking in this case.

PR #2421, opened 2024-11-26
[SuperEditor] - Fix: When navigating from screen with keyboard open to screen with no IME connection, KeyboardScaffoldSafeArea pushes the content up above the keyboard even though its closed. (Resolves #2419)

### LINKED ISSUE #2419: [SuperEditor] - Keyboard panel scaffold thinks keyboard is up when it's down on new screen
Reproduction steps:
 - On Screen 1 with a `KeyboardScaffoldSafeArea`, and a `KeyboardPanelScaffold`, open the keyboard
 - With the keyboard open, navigate to Screen 2, which contain a `KeyboardScaffoldSafeArea` but NOT a `KeyboardPanelScaffold` - the keyboard should close when navigation happens.

Expected results:
Screen 2 takes all space on the screen.

Actual results:
Screen 2 shows an empty area where the keyboard used to be, because the `KeyboardScaffoldSafeArea` thinks the keyboard is up, and there's no `KeyboardPanelScaffold` to correct it.

matthew-carroll:
[SuperEditor] - Fix: When navigating from screen with keyboard open to screen with no IME connection, KeyboardScaffoldSafeArea pushes the content up above the keyboard even though its closed. (Resolves #2419)

A situation was found in ClickUp where `KeyboardScaffoldSafeArea` wasn't doing what was expected.

The app was on Screen 1 with an editor open, and keyboard up. This screen has a `KeyboardScaffoldSafeArea` and a `KeyboardPanelScaffold`.

Then the app navigates to Screen 2.

Screen 2 contains a `KeyboardScaffoldSafeArea` but doesn't contain a `KeyboardPanelScaffold`. 

Screen 2 continues to show an empty area for the keyboard, even though the keyboard has closed.

The reason this happened is because when Screen 2 is first pumped and initialized, the keyboard is still up. Then, as Screen 2 animates in through the navigator, the keyboard closes.

The bug is that when there's not a descendant `KeyboardPanelScaffold`, the `KeyboardScaffoldSafeArea` never updates its bottom insets. It initialized itself with the insets for the keyboard, but never reduced them as the keyboard closed.

This PR adjusts the `KeyboardScaffoldSafeArea` so that if it initializes with insets from a `MediaQuery`, it continues to honor the `MediaQuery` changes. Then, if a descendant ever explicitly sets the insets on the `KeyboardScaffoldSafeArea`, it stops paying attention to the `MediaQuery`.

### DIFF
--- a/super_editor/lib/src/infrastructure/keyboard_panel_scaffold.dart
+++ b/super_editor/lib/src/infrastructure/keyboard_panel_scaffold.dart
@@ -743,7 +743,7 @@ class _KeyboardScaffoldSafeAreaState extends State<KeyboardScaffoldSafeArea>
         bottomPadding: _ancestorSafeArea?.geometry.bottomPadding ?? MediaQuery.paddingOf(context).bottom,
       );
 
-      // We track whether our safe is from MediaQuery (instead of an another KeyboardSafeAreaGeometry).
+      // We track whether our safe area is from MediaQuery (instead of an another KeyboardSafeAreaGeometry).
       // We do this in case the MediaQuery value changes when we don't have any descendant
       // KeyboardPanelScaffold.
       //
--- a/super_editor/lib/src/infrastructure/keyboard_panel_scaffold.dart
+++ b/super_editor/lib/src/infrastructure/keyboard_panel_scaffold.dart
@@ -707,6 +707,7 @@ class _KeyboardScaffoldSafeAreaState extends State<KeyboardScaffoldSafeArea>
   KeyboardSafeAreaGeometry? _keyboardSafeAreaData;
 
   KeyboardScaffoldSafeAreaMutator? _ancestorSafeArea;
+  bool _isSafeAreaFromMediaQuery = false;
 
   @override
   void didChangeDependencies() {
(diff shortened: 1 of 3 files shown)


super_editor/lib/src/infrastructure/keyboard_panel_scaffold.dart
```diff
@@ -723,21 +724,54 @@ class _KeyboardScaffoldSafeAreaState extends State<KeyboardScaffoldSafeArea>
...
-    );
+
+    if (_keyboardSafeAreaData == null) {
+      // This is the first call to didChangeDependencies. Initialize our safe area.
+      _keyboardSafeAreaData = KeyboardSafeAreaGeometry(
+        bottomInsets: _ancestorSafeArea?.geometry.bottomInsets ?? MediaQuery.viewInsetsOf(context).bottom,
+        bottomPadding: _ancestorSafeArea?.geometry.bottomPadding ?? MediaQuery.paddingOf(context).bottom,
+      );
+
+      // We track whether our safe is from MediaQuery (instead of an another KeyboardSafeAreaGeometry).
```
angelosilvestre, line 746: "We track whether our safe is" > "We track whether our safe area is" ?

angelosilvestre (approved):
LGTM with a single comment.

PR #2428, opened 2024-11-27
[SuperEditor][SuperTextField][iOS] Fix focal point for native toolbar (Resolves #2345)

### LINKED ISSUE #2345: [SuperEditor][SuperReader][SuperTextField] - Reports wrong focal point to floating toolbar when double tapping to select word
When double tapping to select a word on iOS, the focal point is reported as the initial caret offset, rather than the center of the selection rectangle.

I found this problem in Super Editor, but Super Reader and Super Text Field should be checked, too.

I found this problem on iOS, but Android should be checked, too.

https://github.com/user-attachments/assets/e87b8b7e-b326-4fa3-8e2f-10be3fc39ae7

angelosilvestre:
[SuperEditor][SuperTextField][iOS] Fix focal point for native toolbar. Resolves #2345

When using the native iOS toolbar, the toolbar has a wrong focal point:

https://github.com/user-attachments/assets/e87b8b7e-b326-4fa3-8e2f-10be3fc39ae7

We compute the toolbar anchor using the `LeaderLink`'s offset and size. The issue is that we are trying to get the leader size before it is computed, because the toolbar builder is being called before the `Leader` runs `performLayout`.

I added some print statements to show what's happening:

```console
flutter: after RenderLeader performLayout(), leader size: Size(0.0, 25.2)
flutter: building toolbar, leader size: Size(0.0, 25.2)
flutter: after RenderLeader performLayout(), leader size: Size(85.3, 21.5)
```

The first `performLayout` runs after the first tap and it's the result is the caret size. This means we are using the bounds computed for the caret instead of the bounds of the expanded selection.

This PR changes the native iOS toolbar builder to query the selection bounds from the `DocumentLayout` instead of relying on the `LeaderLink`. Maybe we want to change `iOSSystemPopoverEditorToolbarWithFallbackBuilder` to use named parameters instead of positional parameters.

https://github.com/user-attachments/assets/41ef3a83-2b07-4104-b391-55d1ded55929

### DIFF
--- a/super_editor/example/lib/demos/in_the_lab/feature_ios_native_context_menu.dart
+++ b/super_editor/example/lib/demos/in_the_lab/feature_ios_native_context_menu.dart
@@ -133,7 +133,6 @@ class _NativeIosContextMenuFeatureDemoState extends State<NativeIosContextMenuFe
       focalPoint,
       _commonEditorOperations,
       SuperEditorIosControlsScope.rootOf(context),
-      _documentLayoutKey.currentState as DocumentLayout,
     );
   }
 
--- a/super_editor/example/lib/demos/in_the_lab/feature_ios_native_context_menu.dart
+++ b/super_editor/example/lib/demos/in_the_lab/feature_ios_native_context_menu.dart
@@ -133,7 +133,6 @@ class _NativeIosContextMenuFeatureDemoState extends State<NativeIosContextMenuFe
       focalPoint,
       _commonEditorOperations,
       SuperEditorIosControlsScope.rootOf(context),
-      _editor.composer.selection!,
       _documentLayoutKey.currentState as DocumentLayout,
     );
   }
--- a/super_editor/example/lib/demos/in_the_lab/feature_ios_native_context_menu.dart
+++ b/super_editor/example/lib/demos/in_the_lab/feature_ios_native_context_menu.dart
@@ -123,12 +123,18 @@ class _NativeIosContextMenuFeatureDemoState extends State<NativeIosContextMenuFe
     Key mobileToolbarKey,
     LeaderLink focalPoint,
   ) {
+    if (_editor.composer.selection == null) {
+      return const SizedBox();
+    }
+
     return iOSSystemPopoverEditorToolbarWithFallbackBuilder(
       context,
       mobileToolbarKey,
       focalPoint,
       _commonEditorOperations,
       SuperEditorIosControlsScope.rootOf(context),
+      _editor.composer.selection!,
+      _documentLayoutKey.currentState as DocumentLayout,
     );
   }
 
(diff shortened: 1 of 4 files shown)


matthew-carroll:
> This PR changes the native iOS toolbar builder to query the selection bounds from the DocumentLayout instead of relying on the LeaderLink.

This doesn't seem like a good idea. Correct me if I'm wrong, but it sounds like you're essentially saying "we have a problem with content-sized leader and followers so I'm switching to a completely different approach".

Shouldn't we solve the actual problem with the leader size reporting instead of kicking that can down the road to the next time something like this happens? 

What happens when the leader size is needed for something that isn't easily queryable from a `DocumentLayout`? Also, since when does the `DocumentLayout` know how we render selection boxes? It looks like you've essentially implied knowledge of implementation details around selection box placement sizing within the toolbar builder. One of the reasons that we used a leader in the first place was so that visual decorations could blindly follow a location chosen by some other part of the system.

angelosilvestre:
> One of the reasons that we used a leader in the first place was so that visual decorations could blindly follow a location chosen by some other part of the system.

@matthew-carroll This works when using a `Follower` to position the desired widget. To query the offset and size we need to be aware of the timing of things, that size will be available only after layout and position will be available only after paint.

Another solution to this is to add an option to create an `IOSSystemContextMenu` passing a `LeaderLink` instead of a `Rect`. When passing a `LeaderLink` we would show the system menu at the end of the frame. I tried this solution and it seems to work.

matthew-carroll:
> To query the offset and size we need to be aware of the timing of things

Timing is already relevant. You can't `build()` a `Follower` before its `Leader`. This appears to be yet another timing consideration on top of that.

> that size will be available only after layout and position will be available only after paint.

Aren't we already dealing with similar ordering issues in `follow_the_leader`? For example, how are we handling screen and widget boundary positioning? In fact, even alignment depends on size. How are we positioning `Follower`s on the right side of a `Leader` if we aren't dealing with ordering?

angelosilvestre:
@matthew-carroll The follower alignment and boundary handling happen during paint, were we already know the size and position. The problem is that we are querying this information during build. That's why I suggested the alternative approach to query this information at the end of the frame.

matthew-carroll:
@angelosilvestre I'm asking whether or not we can make an adjustment to utilize the same approach/timing as those other behaviors. You said those other behaviors work because we do them in paint. So I'm asking if we can similarly utilize the timing of paint to achieve this other goal.

Generally speaking, we never want to rely on "end of frame" because that always necessitates a "followup frame". There's no such thing as visual positioning at the end of a frame, there's only end-of-frame positioning followed by an extra frame. At best we're one frame behind. At worst, weird invalidation stuff happens and that next frame never happens, resulting in an erroneous paint. So I want to exhaust all other options before accepting that.

angelosilvestre:
@matthew-carroll We could turn `IOSSystemContextMenu` into a `RenderObjectWidget` that receives a `LeaderLink`. Then, on `paint` we would show the native menu.

matthew-carroll (approved):
LGTM

Issue #2429, opened 2024-11-28
Update web documentation for super_editor 0.3.0


suragch:
There is a large difference between version 0.2.7 and 0.3.0 and at least a portion of the current docs don't even work with version 0.2.7.

It would be helpful to have documentation that is up-to-date and easy to follow for first time Super Editor users.

PR #2430, opened 2024-11-28
[SuperEditor][Web][Guides]: Updates guides for version 0.3.0 (#2429)

suragch:
This PR aims to update all of the SuperEditor web guides to work with `super_editor` 0.3.0 with an eye for first time users.

Since I myself am a first time user, my goal is to write the docs that I would have wanted to have. They should not only help the reader to write code that runs but also explain any confusing aspects along the way.

I'm initially submitting this PR as an update to a single guide (SuperEditor Quickstart) for early feedback and will then ping the reviewer at various intervals afterwards (after every guide to start with). The PR doesn't need to be merged until all of the guides are finished.

### DIFF
--- a/doc/website/source/super-editor/guides/quickstart.md
+++ b/doc/website/source/super-editor/guides/quickstart.md
@@ -2,68 +2,83 @@
 title: Super Editor Quickstart
 contentRenderers: ["jinja", "markdown"]
 ---
+
 # Super Editor Quickstart
-Super Editor comes with sane defaults to help you get started with an editor experience, quickly. These defaults include support for images, list items, blockquotes, and horizontal rules, as well as selection gestures, and various keyboard shortcuts.
+
+Super Editor comes with sane defaults to help you get quickly started with an editor experience. These defaults include support for images, list items, blockquotes, and horizontal rules, as well as selection gestures and various keyboard shortcuts.
 
 Drop in the default editor and start editing.
 
 ## Add <code>super_editor</code> to your project
+
 To use <code>super_editor</code>, add a dependency in your <code>pubspec.yaml</code>.
 
 ```yaml
 dependencies:
-  super_editor: {{ pub.super_editor.version }}
+  super_editor: ^0.3.0
 ```
 
 ## Display an editor
-A visual editor first requires a logical editor. A logical editor holds an underlying document, which the user edits, and a composer to manage the user's selection.
 
-Initialize the logical editor.
+Super Editor is both the visual editor that users see and interact with, as well as the logical editor that handles those interactions behind the scenes. 
+
+Start by initializing the logical editor and its required components:
 
 ```dart
-class MyApp extends StatefulWidget {
-    State<MyApp> createState() => _MyApp();
+import 'package:flutter/widgets.dart';
+import 'package:super_editor/super_editor.dart';
+
+class MyEditorPage extends StatefulWidget {
+  const MyEditorPage({super.key});
+
+  @override
+  State<MyEditorPage> createState() => _MyEditorPageState();
 }
 
-class _MyApp extends State<MyApp> {
-    late final Editor _editor;
-    late final MutableDocument _document;
-    late final MutableDocumentComposer _composer;
-    
-    void initState() {
-        super.initState();
-    
-        _document = MutableDocument.empty();
-    
-        _composer = MutableDocumentComposer();
-    
-        _editor = Editor();
-    }
-    
-    void dispose() {
-        _editor.dispose();
-        _composer.dispose();
-        _document.dispose();
-    
-        super.dispose();
-    }
+class _MyEditorPageState extends State<MyEditorPage> {
+  late MutableDocument _document;
+  late MutableDocumentComposer _composer;
+  late Editor _editor;
+
+  @override
+  void initState() {
+    super.initState();
+    _document = MutableDocument.empty();
+    _composer = MutableDocumentComposer();
+    _editor = createDefaultDocumentEditor(
+      document: _document,
+      composer: _composer,
+    );
+  }
+
+  @override
+  void dispose() {
+    _composer.dispose();
+    _document.dispose();
+    super.dispose();
+  }
+
+  // More to come...
 }
 ```
+
+Here are a few points to note:
+
+- The logical editor holds an underlying document and a composer to manage the user's selection.
+- The document, <code>MutableDocument</code>, contains a list of nodes for content like text, images, and so on, which users can edit. In this case, you're starting with an empty list.
+- <code>createDefaultDocumentEditor</code> is a convenience method from <code>super_editor</code> to give you some of those sane defaults mentioned earlier.
     
-With the logical pieces ready, you can now display a visual editor. Build a <code>SuperEditor</code> widget and return it from your <code>build()</code> method.
+With the logical pieces ready, you can now display a visual editor. Add a <code>build()</code> method that returns a <code>SuperEditor</code> widget with its logical editor:
     
 ```dart
-class _MyApp extends State<MyApp> {
-    // ...
-  
-    Widget build(BuildContext context) {
-      return SuperEditor(
-          editor: _editor,
-      );
-    }
+@override
+Widget build(BuildContext context) {
+  return SuperEditor(
+    editor: _editor,
+  );
 }
 ```
 
 That's all it takes to get started with your very own editor. Run your app, tap in the editor, and start typing!
 
-The next step is configuration. Check out the other guides for more help.
+Check out the other guides for more help.


suragch:
@matthew-carroll 

My changes to the quickstart guide will be a good test to see how much you'll allow me to change the current wording. The former wording wasn't bad, I just expand it. I'm also following these principles:

- As much as possible, I'll include entire code blocks that are copy-pastable. This will help users to get started quickly and avoid ambiguity.
- I'll include optional pieces like `@override`. Leaving them off may save a few lines, but the result is compiler warnings that makes the tutorial experience not quite as nice.
- I'll tell the reader what to do before the code block and add explanation after the code block to illuminate any confusing parts.

Is it OK to use backticks (\`) rather that `<code></code>`?

doc/website/source/super-editor/guides/quickstart.md
```diff
@@ -2,68 +2,83 @@
 title: Super Editor Quickstart
 contentRenderers: ["jinja", "markdown"]
 ---
+
 # Super Editor Quickstart
-Super Editor comes with sane defaults to help you get started with an editor experience, quickly. These defaults include support for images, list items, blockquotes, and horizontal rules, as well as selection gestures, and various keyboard shortcuts.
+
```
matthew-carroll, line 7: Let's stay consistent with spacing beneath headers. Previously we had no blank lines here, and down below there's still a header with no blank line. So let's remove these two added blank lines.

doc/website/source/super-editor/guides/quickstart.md
```diff
@@ -2,68 +2,83 @@
...
+Super Editor comes with sane defaults to help you get quickly started with an editor experience. These defaults include support for images, list items, blockquotes, and horizontal rules, as well as selection gestures and various keyboard shortcuts.
```
matthew-carroll, line 8: "you get quickly started" -> "you get started quickly"

doc/website/source/super-editor/guides/quickstart.md
```diff
@@ -2,68 +2,83 @@
...
 Drop in the default editor and start editing.
 
 ## Add <code>super_editor</code> to your project
+
 To use <code>super_editor</code>, add a dependency in your <code>pubspec.yaml</code>.
 
 ```yaml
 dependencies:
-  super_editor: {{ pub.super_editor.version }}
+  super_editor: ^0.3.0
```
matthew-carroll, line 18: Since we're specifying a non-pub version, it's probably a good idea to use a global variable in case this is referenced in multiple places.

You can something like `super_editor_version` in a top-level `_data.yaml` with a value of "^0.3.0".

doc/website/source/super-editor/guides/quickstart.md
```diff
@@ -2,68 +2,83 @@
...
+  late MutableDocument _document;
+  late MutableDocumentComposer _composer;
+  late Editor _editor;
+
+  @override
+  void initState() {
+    super.initState();
+    _document = MutableDocument.empty();
+    _composer = MutableDocumentComposer();
+    _editor = createDefaultDocumentEditor(
```
matthew-carroll, line 48: I think we can now avoid storing the document and composer:

```dart
_editor = _createDefaultDocumentEditor(
  document: MutableDocument.empty(),
  composer: MutableDocumentComposer(),
);
```

Sometime soon we should probably adjust `createDefaultDocumentEditor()` to make those parameters optional and instantiate sane defaults if they're not provided.

doc/website/source/super-editor/guides/quickstart.md
```diff
@@ -2,68 +2,83 @@
...
 ```dart
-class _MyApp extends State<MyApp> {
-    // ...
-  
-    Widget build(BuildContext context) {
-      return SuperEditor(
-          editor: _editor,
-      );
-    }
+@override
```
matthew-carroll, line 74: Given the expansion of explanation above, there's probably no reason to separate this code sample from the previous one.

Might as well show everything and then explain it.

doc/website/source/super-editor/guides/quickstart.md
```diff
@@ -2,68 +2,83 @@
...
+    _composer.dispose();
+    _document.dispose();
+    super.dispose();
+  }
+
+  // More to come...
 }
 ```
+
+Here are a few points to note:
```
matthew-carroll, line 65: In general, I think it's better to use qualifiers that add substance, rather than qualifies that essentially tell the reader that there's going to be more to read.

I also prefer that lists not be used to essentially segment paragraphs. We can just use paragraphs.

Consider the following alternative to what's written.

> Multiple objects work together to edit documents. A `Document` provides a consistent structure for content within a document. A `DocumentComposer` holds the user's current selection, along with any styles that should be applied to newly typed text. An `Editor` alters the `Document`.
>
> The `Editor` fulfills a number of responsibilities, each of which is configurable. Rather than force every user to fully configure an `Editor`, `super_editor` provides a global factory called `createDefaultDocumentEditor`, which configures an `Editor` with sane defaults. To adjust those defaults, consider copying the implementation of `createDefaultDocumentEditor` and then altering the implementation to meet your needs.
>
> The `SuperEditor` widget creates a user interface for visualizing the `Document`, changing the selection in the `DocumentComposer`, and submitting change requests to the `Editor`. The `SuperEditor` widget is the part that most people think of when they think of "document editing". The `SuperEditor` widget includes many configurable properties, all of which focus on user interactions, e.g., selection and focus policies, gesture interceptors, scroll control, mobile selection handles, and more.
>
> Continue your Super Editor journey with more beginner guides:
> // Here's where you can link to more info about Document, DocumentComposer, Editor, SuperEditor, and other useful next step guides.

matthew-carroll (commented):
Left comments on the initial adjustments to `quickstart.md`.

To your question about `<code>` tags. Yes you can use backticks instead. In fact, you should only use HTML tags in Markdown when you have no other choice.

doc/website/source/super-editor/guides/quickstart.md
suragch: I'd like to argue for hard-coding the version number right in the text of the guide. That makes it obvious what version the code in the guide was written for. It's too easy to update a variable and forget that that doesn't automatically migrate the code. I have no idea what version of Super Editor the previous Quickstart used, but it certainly wasn't 0.2.7 or 0.3.0. 

Seeing an outdated pub version in the guides will:

1. remind us that we need to update the version and perform a full read-though and manual code check,
2. prompt users to open a GitHub issue reminding us that the guides out of date, and
3. provide a pointer to a version of super_editor that does work in case a user can't get the most recent version to work.

Auto-updating the version number, on the other hand, can very likely lead to what I experienced: code that doesn't compile and a tedious process of finding the cause. Most casual users would just give on Super Editor.

doc/website/source/super-editor/guides/quickstart.md
suragch: That would be useful. I'll add a TODO in the guide for now until `createDefaultDocumentEditor()` is updated.

doc/website/source/super-editor/guides/quickstart.md
matthew-carroll: > I'd like to argue for hard-coding the version number right in the text of the guide. That makes it obvious what version the code in the guide was written for. It's too easy to update a variable and forget that that doesn't automatically migrate the code.

This doesn't account for the maintenance burden of keeping these things updated. It's possible to make an error in both directions. Whether you hard code it, or use a variable, there are ways to get it wrong. But which situation is more common, and which situation is more likely to create errors? I think that almost certainly scattering an unknown number of version numbers across the guides is far more likely to produce errors on a more regular basis.

I'll address each of your specific points:

> remind us that we need to update the version and perform a full read-though and manual code check

It won't remind you of anything if you forget that there's a version number sitting in some random doc. And after you forget about it for a few versions, there won't even be a great way to find it. If you try to do a full code search for something like `0.3.0-dev.9`, you'll also need to serach `0.3.0-dev.8` and so on, for every possible version number that you may have forgotten. 

> prompt users to open a GitHub issue reminding us that the guides out of date

Almost no one ever files an issue based on documentation. We shouldn't expect or rely on any such thing happening.

However, what we should do, eventually, is create buttons in the docs that take people to a pre-configured issue ticket to do things like that.

> provide a pointer to a version of super_editor that does work in case a user can't get the most recent version to work.

The right answer for situations like this is to have API docs that are partitioned by API version. We'll need to add infrastructure to do that, but that's the way to go.

suragch: To be clear, when I talk about "a full read-though and manual code check", I mean reading and updating the code in all of the guides, not just a single document or doing a search-and-replace type of update. This is the type of update I've performed numerous times on books and articles.

I agree that it would be useful to have docs partitioned by API version, similar to how Docusaurus does it.

Anyway, I'll go with your guidance and add the version to a variable name in `_data.yaml`.

matthew-carroll: > when I talk about "a full read-though and manual code check", I mean reading and updating the code in all of the guides, not just a single document or doing a search-and-replace type of update

We should try to leverage process and infrastructure to avoid such a requirement. I know that I won't have time to do that, and most others won't either. You might be OK doing that now, but I'm sure you won't want that responsibility either - especially when we've got 100 guides.

doc/website/source/super-editor/guides/quickstart.md
```diff
@@ -2,68 +2,81 @@
...
-  super_editor: {{ pub.super_editor.version }}
+  super_editor: {{ super_editor_version }}
 ```
 
 ## Display an editor
-A visual editor first requires a logical editor. A logical editor holds an underlying document, which the user edits, and a composer to manage the user's selection.
+Super Editor is both the visual editor that users see and interact with, as well as the logical editor that handles those interactions behind the scenes. 
 
-Initialize the logical editor.
+Start by initializing the logical editor and its required components:
```
matthew-carroll, line 22: We should avoid using the term "components" in a generic way because we actually have things we refer to as "components", which are block-level widgets within a document layout.

doc/website/source/super-editor/guides/quickstart.md
matthew-carroll: The update I was originally mentioning here was to get rid of `_document` and `_composer`. I think those are no longer needed as instance members. You can instantiate them directly in the call to `createDefaultDocumentEditor()`.

doc/website/source/super-editor/guides/quickstart.md
suragch: I understand that updating the guides is a lot of work, but when the underlying package goes through breaking changes, how else do we update the guides? Changing a variable name won't fix a legacy example that needs a migration and a new explanation.

doc/website/source/super-editor/guides/quickstart.md
suragch: Ah, got it!

doc/website/source/super-editor/guides/quickstart.md
matthew-carroll: Part of the answer to that question is to invent the answers. My point here is that rather than assume that you just have to re-validate dozens of guides every couple weeks, try to look for ways to eliminate that need. If a problem is intractable, how can we make it tractable?

One such area of focus, which I've discussed with Angelo in the past, is to use code samples that come from actual source code, which is actually compiled in CI. If the code samples no longer compile, then they'll trigger a CI failure. We'll be directed to exactly the problematic code, avoiding 98% of the reading and review time.

suragch: Those are good ideas.

suragch:
@matthew-carroll 
I've updated the [Document from Markdown](https://supereditor.dev/super-editor/guides/document-from-markdown/) guide. Do we need to provide special instructions for importing the `super_editor_markdown` package? It doesn't work as directed in the guides. A dependency override is necessary. Is that always going to be required?

matthew-carroll:
> A dependency override is necessary.

@suragch - Is this actually true, after our recent conversation on Discord?

doc/website/source/super-editor/guides/document-from-markdown.md
```diff
@@ -17,13 +17,27 @@ dependencies:
...
+const markdown = '''
+# Header
+This is a _Super_ Editor!
+''';
+
 final document = deserializeMarkdownToDocument(markdown);
 ```
 
-The de-serialized document is a `MutableDocument`. Check other guides to find out how to use it.
+The de-serialized document is a `MutableDocument`. Add it to your `Editor` similarly to how you did it in the Quickstart guide:
```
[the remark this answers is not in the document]
suragch: Since people presumably just finished reading the Quickstart guide, my thinking is that we don't need to re-explain what a `MutableDocument` is. I thought having a quick refresher on where to use the `document` object would be helpful, though.

I take your point about the "add" word, and including a link is a good idea.

But if you would prefer to remove the code block or add the extended explanation, I can do that.

doc/website/source/super-editor/guides/document-from-markdown.md
```diff
@@ -17,13 +17,27 @@ dependencies:
...
+
+```dart
+_editor = createDefaultDocumentEditor(
+  document: document,
+  composer: MutableDocumentComposer(),
+);
+```
+
+Run that, and you'll see the document rendered on the screen.
```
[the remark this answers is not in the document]
suragch: I'm not sure if you noticed that I had added the block:

```
const markdown = '''
# Header
This is a _Super_ Editor!
''';
```

I did that so that this guide would work more like a tutorial and running the code would produce a visible result. The code in some of the other guides don't run as written. For example, `document.deleteNodeAt(2)` in the Assemble a Document guide was a code fragment meant to be used as an example and not meant to be run as a part of a tutorial (since `2` would have been out of range). "Run that" is a signal to the user that given code is tutorial style rather than a random snippet, and they should be able to try it out.

Also, the current Quickstart guide that is live now says "Run your app, tap in the editor, and start typing!" I was following this tutorial style pattern. 

Is your intention for most of the guides that they should just be code snippets rather than tutorials?

doc/website/source/super-editor/guides/document-from-markdown.md
```diff
@@ -17,13 +17,27 @@ dependencies:
...
 
 ## Serialize Markdown
-Serialize a `Document` to a Markdown `String`.
+You can also go the other direction. Here's how you would serialize a `Document` to a Markdown `String`:
```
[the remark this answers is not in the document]
suragch: I think I misunderstood the original intent in updating the guides from a new user's perspective. I was updating them as if I were working through the guides as a new user reading them in the order they appear in the menu and reading each guide from top to bottom. I see now that that was not your intention.

suragch:
> Is this actually true, after our recent conversation on Discord?

Our Discord discussion resolved the issue.

suragch:
@matthew-carroll In the [Add a Popover Toolbar](https://supereditor.dev/super-editor/guides/add-a-popover-toolbar/) guide, there is the following line:

```
_toolbarAligner = CupertinoPopoverToolbarAligner(_viewportKey);
```

`CupertinoPopoverToolbarAligner` doesn't compile for me and I can't find it or a good substitute in `follow_the_leader`. What is the current way of doing this?

doc/website/source/super-editor/guides/document-from-markdown.md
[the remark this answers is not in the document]
suragch: I've reverted this change back to the original.

doc/website/source/super-editor/guides/document-from-markdown.md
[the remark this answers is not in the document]
suragch: I've removed this change.

doc/website/source/super-editor/guides/document-from-markdown.md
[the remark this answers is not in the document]
suragch: I've reverted this change back to the original.

suragch:
@matthew-carroll 

I've reverted several of my previous changes. Partly I had misunderstood the scope and purpose of the edits I was making. The other part was I let my pride get the better of me and tried to argue my side when that wasn't what you wanted for this site. I forgot that I'm here to be a learner. I'm sorry about that.

I've left in a few TODO questions. If you have any answers for those, I'll add them in. Otherwise, we can ignore them.

I wasn't able to get the Popover guides working. (See my question above about that.)

The [Add New Type of Content](https://supereditor.dev/super-editor/guides/add-new-type-of-content) and [Add a Mobile Keyboard Toolbar](https://supereditor.dev/super-editor/guides/add-a-mobile-keyboard-toolbar) guides are essentially empty. Should we remove those from the menu for now?

I'll limit this PR to the Super Editor guides. If I update the Super Reader or Super Text Field guides, I create a new PR for them later.

PR #2435, opened 2024-11-30
[SuperEditor][SuperReader] - Reduce uses of node index queries in Document (Resolves #2434)

### LINKED ISSUE #2434: [SuperEditor][SuperReader] - Minimize uses of node indices
When we change a `Document` from a list of nodes to a tree of nodes (#2433) the concept of a "node index" will change in a significant way. There will no longer be a global node index. Instead, a node will have an index within some parent node. The root node will be called the "document".

To prepare us for this change, convert every unnecessary use of node indices to some other node access mechanism, such that we can use the same access mechanism with a tree. For example, I expect that we'll retain `getNodeById()` at a document level, which means those uses won't be impacted by the change to a tree.

matthew-carroll:
[SuperEditor][SuperReader] - Reduce uses of node index queries in Document (Resolves #2434)

This PR changes as many calls as possible from index-based node access to either getting a node by ID, or getting a node before/after another node. This is in preparation for changes to a tree document, for which those behaviors still make sense, but accessing at an index doesn't.

### DIFF
--- a/super_editor/lib/src/core/editor.dart
+++ b/super_editor/lib/src/core/editor.dart
@@ -1291,7 +1291,7 @@ class MutableDocument with Iterable<DocumentNode> implements Document, Editable 
     }
 
     for (int i = 0; i < _nodes.length; ++i) {
-      if (!_nodes[i].hasEquivalentContent(other.getNodeAt(i)!)) {
+      if (!_nodes[i].hasEquivalentContent(other.getNodeById(_nodes[i].id)!)) {
         return false;
       }
     }
--- a/super_editor/lib/src/core/editor.dart
+++ b/super_editor/lib/src/core/editor.dart
@@ -1278,7 +1278,7 @@ class MutableDocument with Iterable<DocumentNode> implements Document, Editable 
     }
   }
 
-  /// Returns [true] if the content of the [other] [Document] is equivalent
+  /// Returns `true` if the content of the [other] [Document] is equivalent
   /// to the content of this [Document].
   ///
   /// Content equivalency compares types of content nodes, and the content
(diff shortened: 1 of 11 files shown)


super_editor/lib/src/core/editor.dart
```diff
@@ -1291,7 +1291,7 @@ class MutableDocument with Iterable<DocumentNode> implements Document, Editable
     }
 
     for (int i = 0; i < _nodes.length; ++i) {
-      if (!_nodes[i].hasEquivalentContent(other.getNodeAt(i)!)) {
+      if (!_nodes[i].hasEquivalentContent(other.getNodeById(_nodes[i].id)!)) {
         return false;
       }
```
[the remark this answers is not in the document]
matthew-carroll: I re-implemented the method. Let me know if there are remaining inconsistencies.

super_editor/lib/src/default_editor/multi_node_editing.dart
```diff
@@ -923,28 +923,38 @@ class DeleteContentCommand extends EditCommand {
...
+        ));
+        document.deleteNode(nodeToDelete);
       }
-      changes.add(DocumentEdit(
-        NodeRemovedEvent(removedNode.id, removedNode),
-      ));
-      document.deleteNodeAt(i);
+
+      // Move to the next node.
+      nodeToDelete = document.getNodeAfter(nodeToDelete);
```
angelosilvestre, line 957: Shouldn't we query `document.getNodeAfter` before deleting the node? Otherwise I guess `getNodeAfter` would return `null`, since `nodeToDelete` won't be present in the document anymore.

matthew-carroll: Yeah. Updated.

angelosilvestre (approved):
LGTM

PR #2436, opened 2024-12-02
[SuperEditor] Make getDocumentPositionAfterExpandedDeletion return null for collapsed selections (Resolves #2431)

### LINKED ISSUE #2431: [BUG] - getDocumentPositionAfterExpandedDeletion with collapsed selection on block node returns TextPosition
Can be easily reproducible. The position should be `UpstreamDownstreamNodePosition` but is `TextPosition` instead.

```dart
test("getDocumentPositionAfterExpandedDeletion on block node", () {
  final node = HorizontalRuleNode(
    id: "1",
  );
  final document = MutableDocument(nodes: [
    node,
  ]);
  final selection = DocumentSelection.collapsed(
    position: DocumentPosition(
      nodeId: node.id,
      nodePosition: node.endPosition,
    ),
  );
  final position = CommonEditorOperations.getDocumentPositionAfterExpandedDeletion(
    document: document,
    selection: selection,
  );
  expect(position is UpstreamDownstreamNodePosition, isTrue);
});
```

angelosilvestre:
[SuperEditor] Make getDocumentPositionAfterExpandedDeletion return null for collapsed selections. Resolves #2431

The `CommonEditorOperations.getDocumentPositionAfterExpandedDeletion` returns a `TextNodePosition` even when the selection is collapsed. In this situation, we expect nothing to be deleted and no selection change.

Currently, if all nodes in the selection are non-deletable, this method returns `null`. To keep consistency, this PR changes this method to also return `null` when the selection is collapsed.f

### DIFF
--- a/super_editor/lib/src/default_editor/common_editor_operations.dart
+++ b/super_editor/lib/src/default_editor/common_editor_operations.dart
@@ -1396,6 +1396,13 @@ class CommonEditorOperations {
   }) {
     // Figure out where the caret should appear after the
     // deletion.
+
+    if (selection.isCollapsed) {
+      // There is no expanded deletion when the selection is collapsed. Therefore,
+      // no selection change is expected.
+      return null;
+    }
+
     // TODO: This calculation depends upon the first
     //       selected node still existing after the deletion. This
     //       is a fragile expectation and should be revisited.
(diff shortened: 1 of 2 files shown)


matthew-carroll (approved):
LGTM

PR #2445, opened 2024-12-08
Adds placeholders to AttributedText (Resolves #2440)

### LINKED ISSUE #2440: [AttributedText] - Add single-character placeholders
We want to display small inline images in `SuperEditor`, `SuperReader`, and `SuperTextField`. This begins with a representation of such things in `AttributedText`. `AttributedText` doesn't know anything about rendering, so we don't want to insert any concept of `Widget`s, layout, etc. Instead, we probably want some concept of a "placeholder" which, in effect, is just an extra character whose content isn't chosen until rendering happen.

For this ticket, add the concept of a placeholder to `AttributedText`.

 * Length of `AttributedText` should include a +1 for each placeholder
 * I think accessing plain `text` should produce a string that doesn't have any extra characters for each placeholder. If needed, maybe we make it configurable to inject a desired character in place of each placeholder, but we can wait until that's a requirement.
   * One issue here is any code that has a text offset, and then accesses the character in `text` at that offset, might get the wrong character because the offset includes placeholders, but `text` doesn't include those placeholders, so there's an offset misalignment.
 * Placeholders need to be able to contain arbitrary data so that clients can store whatever is needed to render the eventual widgets.

matthew-carroll:
Adds placeholders to AttributedText (Resolves #2440)

We want to display inline widgets in `SuperEditor`. This requires a way to represent the existence of such widgets within the text model, which is `AttributedText`.

This PR adds placeholders to `AttributedText`. A placeholder is an `Object` that takes up the space of a single character. This spacing allows for text positions before and after the placeholder, but not within the placeholder.

Each placeholder is represented by some kind of `Object`. We currently don't place any restriction on what that is. Mapping from placeholders to widgets is handled by the visual system.

### DIFF
--- a/attributed_text/lib/src/attributed_text.dart
+++ b/attributed_text/lib/src/attributed_text.dart
@@ -1,3 +1,5 @@
+import 'package:collection/collection.dart';
+
 import 'attributed_spans.dart';
 import 'attribution.dart';
 import 'logging.dart';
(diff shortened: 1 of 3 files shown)


attributed_text/lib/src/attributed_text.dart
```diff
@@ -17,25 +19,133 @@ final _log = attributionsLog;
 // TODO: there is a mixture of mutable and immutable behavior in this class.
 //       Pick one or the other, or offer 2 classes: mutable and immutable (#113)
 class AttributedText {
+  static const placeholderCharacter = '\uFFFC';
```
angelosilvestre, line 22: Please add a dart doc for this.

angelosilvestre (approved):
LGTM

PR #2448, opened 2024-12-09
[SuperText] - Add support for inline widgets (Resolves #2441)

### LINKED ISSUE #2441: [SuperText] - Render inline widgets
We want to display small inline images in `SuperEditor`, `SuperReader`, and `SuperTextField`. This begins with a representation of such things in `AttributedText`: https://github.com/superlistapp/super_editor/issues/2440

Once placeholders are added to `AttributedText`, we need to render widgets in place of those placeholders.

For this ticket, adjust `SuperText` to render small images and gifs based on `AttributedText` placeholders.

Ensure that selection rendering works correctly when inline images are rendered.

Consider edge cases such as a paragraph that only contains a single inline image, and a paragraph that contains only a series of inline images.

matthew-carroll:
[SuperText] - Add support for inline widgets (Resolves #2441)

Changes:
 * Expand all selection boxes to "max" so that the selection rectangle for inline widgets isn't reduced to the height of the inline widget.
 * Position the caret forcing a downstream position to reduce buggyness from Flutter. Flutter is really broken with caret positioning near inline widgets. But "upstream" is more broken than "downstream".
 * Added a couple likely useful inline image placeholders.


matthew-carroll:
I'm gonna merge this in so I can put up a PR for `SuperEditor` integration. Please leave comments on anything in here that seems wrong.

super_text_layout/test_goldens/inline_widgets_test.dart
```diff
@@ -0,0 +1,339 @@
...
+import 'package:super_text_layout/super_text_layout.dart';
+
+import 'test_tools_goldens.dart';
+
+void main() {
+  group("SuperText inline widgets >", () {
+    testGoldensOnAndroid("vertical alignments", (tester) async {
+      await tester.pumpWidget(
+        _buildScaffold(
+          // ignore: prefer_const_constructors
```
angelosilvestre, line 13: Why do we need to ignore `prefer_const_constructors` here?

angelosilvestre (commented):
LGTM

super_text_layout/test_goldens/inline_widgets_test.dart
matthew-carroll: I think this might be a holdover from a copy/paste

Issue #2424, opened 2024-11-27
[SuperEditor] - Attributions not placed according to markdown standard when serializing document


edhom:
The `AttributedTextMarkdownSerializer` places markdown formatting symbols according to the offset of attribution markers. This results in invalid markdown, as the markdown standard doesn't allow leading/trailing whitespaces.

I'm unsure whether the correct fix is in the `AttributedTextMarkdownSerializer`, or instead to update the attribution markers already before to exclude whitespace. 

**To Reproduce**

https://github.com/user-attachments/assets/07d98389-d587-4c75-97f6-b50df300b861

**Actual behavior**
Super** editor**

**Expected behavior**
Super **editor**

edhom:
> I'm unsure whether the correct fix is in the AttributedTextMarkdownSerializer, or instead to update the attribution markers already while editing to exclude whitespace.
@matthew-carroll If you could guide me on this question, I would try to come up with a PR:)

matthew-carroll:
Hi @edhom - can you please link to the part of the Markdown spec that says this?

WRT a PR/work - the best place to start is to write a bundle of failing tests that demonstrate all the edge cases you can think of for this rule. We already have many tests written in the super_editor_markdown package. You can check if there's a place within the existing files where these tests would fit. If for some reason these tests are sufficiently different in purpose from all existing tests, then you can create a new file.

If you put up a draft PR and tag me and @angelosilvestre with those tests then we can better understand the implications of this change, and help figure out where the change should be implemented.

edhom:
@matthew-carroll Thanks a lot, I followed up here: https://github.com/superlistapp/super_editor/pull/2452

PR #2452, opened 2024-12-10
[SuperEditorMarkdown] Attributions not placed according to markdown standard when serializing document (#2424)

edhom:
Hi @matthew-carroll, hi @angelosilvestre :) Here's a PR with failing tests following up on your comment [here](https://github.com/superlistapp/super_editor/issues/2424#issuecomment-2524134384). I also added some background below to motivate why I think this is an issue. Thanks again for your support here and I hope I can contribute myself after some guidance!

### Background
The `AttributedTextMarkdownSerializer` places markdown formatting symbols according to the offset of attribution markers. This results in invalid markdown, as the markdown standard doesn't allow leading/trailing whitespaces. The closest official source I find to back my issue of incorrect markdown annoations is the following:

> To bold/italicize text, add one/two asterisks before and after a word or phrase. ([markdownguide.org](https://www.markdownguide.org/basic-syntax/#bold))

Also from the original spec from [John Gruber](https://daringfireball.net/projects/markdown/syntax#em), it's obvious that the annotation should be around a word/phrase and not white space.

GitHub  shows ** bold text** and not **bold text** just as `flutter_markdown`.

### DIFF
--- a/super_editor_markdown/test/attributed_text_markdown_test.dart
+++ b/super_editor_markdown/test/attributed_text_markdown_test.dart
@@ -119,5 +119,24 @@ void main() {
         "This is ***bold*** text.",
       );
     });
+
+    test("First and last character in one attribution span is white space", () {
+      expect(
+        AttributedText(
+          //       b  b
+          //      i    i
+          "This is bold text.",
+          AttributedSpans(
+            attributions: [
+              const SpanMarker(attribution: boldAttribution, offset: 8, markerType: SpanMarkerType.start),
+              const SpanMarker(attribution: boldAttribution, offset: 11, markerType: SpanMarkerType.end), 
+              const SpanMarker(attribution: italicsAttribution, offset: 7, markerType: SpanMarkerType.start),
+              const SpanMarker(attribution: italicsAttribution, offset: 12, markerType: SpanMarkerType.end),
+            ],
+          ),
+        ).toMarkdown(),
+        "This is ***bold*** text.",
+      );
+    });
   });
 }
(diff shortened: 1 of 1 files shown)


matthew-carroll:
> Also from the original spec from [John Gruber](https://daringfireball.net/projects/markdown/syntax#em), it's obvious that the annotation should be around a word/phrase and not white space.

Why is this "obvious"? What specifically are you referring to in the original spec? Neither of your links seem to mention or address the question of spaces at all.

edhom:
@matthew-carroll Sorry I wanted to refer to the examples, they consistently show the delimiters placed directly around the text without spaces. And you're right it's not obvious at all that it shouldn't be allowed to have whitespaces. But that's what most markdown renderers implemented.

matthew-carroll:
Can you show a few examples of editors that behave the way you're describing here? Namely, editors that aren't using Flutter - editors that have a longer track record and user base.

edhom:
Sure!

| Editor | Preview |
|--------|--------|
| VSCode, built-in markdown preview | Screenshot 2024-12-12 at 07 50 28 |
| Typora | Screenshot 2024-12-12 at 07 57 22 |
| StackEdit | Screenshot 2024-12-12 at 08 01 00 |
| Dillinger | Screenshot 2024-12-12 at 08 06 05 |
| GitHub | ** bold text **, **bold text** |

edhom:
@matthew-carroll Do you think the serializer should account for this? If not wanted, I will post-edit myself the result after serialization...

PR #2455, opened 2024-12-13
[SuperTextField] Add ability to override tap gestures (Resolves #2447)

### LINKED ISSUE #2447: [FEATURE] - [SuperTextField] Allow `LinkAttributions` to be clickable
**Your Info**
Superlist

**User Behavior**
Currently the `SuperTextField` (for desktop) does not allow for customization of the `MouseCursor` when a text field is hovered.
Moreover the `SuperTextField` does not allow for custom left-click/tap actions based on what the user clicked on in the text field.

Ideally, the cursor and left-click action should be overridable based on what Attribution a user is hovering while the `SuperTextField` doesn't have focus.
This would allow a 'click to open' on a `LinkAttribution` within a `SuperTextField`.

angelosilvestre:
[SuperTextField] Add ability to override tap gestures. Resolves #2447

This PR allows apps to register custom handlers, to enable features like changing the mouse cursor when hovering content with specific attributions, like a `LinkAttribution` and performing custom logic when tapping on it.

The implementation is based on the `SuperEditor`'s `ContentTapDelegate`.

I noticed we also have a `RightClickListener`. We could probably deprecate this property and make everything available through the new `SuperTextFieldTapHandler`.

This PR includes a `SuperTextFieldLaunchLinkTapHandler` that changes the mouse cursor when hovering a link, and opens the URL when tapping on it.

On `SuperEditor`, we have the concept of "interaction mode", which is enabled when the users presses CMD. The mouse cursor and tap handling on the link handler only runs if the editor is in interaction mode. This is not yet implemented by this PR, because the original ticket didn't mention it. If it's needed, we can add that.

### DIFF
--- a/super_editor/lib/src/super_textfield/infrastructure/text_field_gestures_interaction_overrides.dart
+++ b/super_editor/lib/src/super_textfield/infrastructure/text_field_gestures_interaction_overrides.dart
@@ -11,14 +11,33 @@ import 'package:super_text_layout/super_text_layout.dart';
 abstract class SuperTextFieldTapHandler with ChangeNotifier {
   MouseCursor? mouseCursorForContentHover(SuperTextFieldGestureDetails details) => null;
 
-  TapHandlingInstruction onTap(SuperTextFieldGestureDetails details) => TapHandlingInstruction.continueHandling;
+  TapHandlingInstruction onTapDown(SuperTextFieldGestureDetails details) => TapHandlingInstruction.continueHandling;
 
-  TapHandlingInstruction onDoubleTap(SuperTextFieldGestureDetails details) => TapHandlingInstruction.continueHandling;
+  TapHandlingInstruction onTapUp(SuperTextFieldGestureDetails details) => TapHandlingInstruction.continueHandling;
 
-  TapHandlingInstruction onTripleTap(SuperTextFieldGestureDetails details) => TapHandlingInstruction.continueHandling;
+  TapHandlingInstruction onTapCancel() => TapHandlingInstruction.continueHandling;
 
-  TapHandlingInstruction onSecondaryTap(SuperTextFieldGestureDetails details) =>
+  TapHandlingInstruction onDoubleTapDown(SuperTextFieldGestureDetails details) =>
       TapHandlingInstruction.continueHandling;
+
+  TapHandlingInstruction onDoubleTapUp(SuperTextFieldGestureDetails details) => TapHandlingInstruction.continueHandling;
+
+  TapHandlingInstruction onDoubleTapCancel() => TapHandlingInstruction.continueHandling;
+
+  TapHandlingInstruction onTripleTapDown(SuperTextFieldGestureDetails details) =>
+      TapHandlingInstruction.continueHandling;
+
+  TapHandlingInstruction onTripleTapUp(SuperTextFieldGestureDetails details) => TapHandlingInstruction.continueHandling;
+
+  TapHandlingInstruction onTripleTapCancel() => TapHandlingInstruction.continueHandling;
+
+  TapHandlingInstruction onSecondaryTapDown(SuperTextFieldGestureDetails details) =>
+      TapHandlingInstruction.continueHandling;
+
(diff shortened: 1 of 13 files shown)


super_editor/lib/src/super_textfield/infrastructure/text_field_gestures_interaction_overrides.dart
```diff
@@ -0,0 +1,51 @@
+import 'package:flutter/widgets.dart';
+import 'package:super_editor/src/infrastructure/document_gestures_interaction_overrides.dart';
+import 'package:super_editor/super_text_field.dart';
+import 'package:super_text_layout/super_text_layout.dart';
+
+/// Delegate for mouse status and clicking on special types of content,
```
matthew-carroll, line 6: Please replace with:

```
/// Tap handler that can (optionally) respond to single, double, and triple taps, as well as dictate the cursor
/// appearance on desktop.
```

I also want to add a section about listening, but after reviewing the code that registers as a listener, I'm not actually sure how the change notifications are supposed to work?

super_editor/lib/src/super_textfield/infrastructure/text_field_gestures_interaction_overrides.dart
```diff
@@ -0,0 +1,51 @@
...
+
+  TapHandlingInstruction onTap(TextFieldGestureDetails details) => TapHandlingInstruction.continueHandling;
+
+  TapHandlingInstruction onDoubleTap(TextFieldGestureDetails details) => TapHandlingInstruction.continueHandling;
+
+  TapHandlingInstruction onTripleTap(TextFieldGestureDetails details) => TapHandlingInstruction.continueHandling;
+}
+
+/// Information about a gesture that happened within a [SuperTextField].
+class TextFieldGestureDetails {
```
matthew-carroll, line 22: Should we call this `SuperTextFieldGestureDetails` to match other names?

super_editor/test/super_textfield/super_textfield_gestures_interaction_overrides_test.dart
```diff
@@ -0,0 +1,315 @@
...
+          expect(
+            SuperTextFieldInspector.findSelection(),
+            const TextSelection.collapsed(offset: -1),
+          );
+        });
+      });
+
+      group('multiple handlers >', () {
+        testWidgetsOnAllPlatforms('run seach handler until the gesture is handled', (tester) async {
+          final noopHandler = _NoopTextFieldTapHandler();
```
matthew-carroll, line 41: Please check capitalization: "noopHandler" -> "noOpHandler"

matthew-carroll:
> I noticed we also have a RightClickListener. We could probably deprecate this property and make everything available through the new SuperTextFieldTapHandler.

Ok, let's go ahead and deprecate.

BTW, while we're here on this ticket, can we also offer right-click support? 

https://github.com/superlistapp/super_editor/issues/278

https://github.com/superlistapp/super_editor/pull/692

angelosilvestre:
@matthew-carroll Since https://github.com/superlistapp/super_editor/issues/278 mentions tap down vs tap up events, should we provide down/up variations for the all tap handlers (single tap, double tap, triple tap) ?

super_editor/lib/src/super_textfield/infrastructure/text_field_gestures_interaction_overrides.dart
angelosilvestre: > I also want to add a section about listening, but after reviewing the code that registers as a listener, I'm not actually sure how the change notifications are supposed to work?

On `SuperEditor`, we have the concept on interaction mode in the document composer, when the user presses CMD the interaction mode is enabled. The `SuperEditorLaunchLinkTapHandler` listens for interaction mode changes and notifies its own listeners when a change happens.

On `SuperTextField`, we don't yet have an interaction mode, nor a composer. It's not clear to me where this should be handled.

matthew-carroll (approved):
LGTM

PR #2459, opened 2024-12-17
[SuperEditor][SuperTextField][iOS] Make caret word snapping less aggressive (Resolves #2152)

### LINKED ISSUE #2152: [SuperEditor][SuperTextField][iOS] - caret word snapping is too aggressive
Based on iOS behavior, when the user taps to place the caret somewhere new, the caret should jump to the edge of the word.

However, our current implementation is too aggressive. We snap even when the user is tapping on the existing caret location, which should instead open the edit toolbar.

https://github.com/superlistapp/super_editor/assets/7259036/3eea0e53-eeb1-41af-90d7-5e4f1f86b209

angelosilvestre:
[SuperEditor][SuperTextField][iOS] Make caret word snapping less aggressive. Resolves #2152

On iOS, the caret snaps to the edges of a word. However, this shouldn't happen when the user taps at the caret position.

This PR changes `SuperEditor` and `SuperTextField` to avoid snapping the caret when tapping at the caret.

Some required adjustments:
- The check for a tap over the caret was using the snapped position instead of the tap position. Modified to use the exact tap  position.
- The check for a tap over the caret was expanding the caret 24px in all directions. This causes an issue, because tapping on a line below (or up) can be interpreted as a tap on the caret. Modified to expand only horizontally.

We still have a problem in `SuperEditor` regarding showing the toolbar when tapping at the caret. We have this code:

```dart
if (didTapOnExistingSelection && _isKeyboardOpen) {
  // Toggle the toolbar display when the user taps on the collapsed caret,
  // or on top of an existing selection.
  //
  // But we only do this when the keyboard is already open. This is because
  // we don't want to show the toolbar when the user taps simply to open
  // the keyboard. That would feel unintentional, like a bug.
  _controlsController!.toggleToolbar();
} else {
  // The user tapped somewhere else in the document. Hide the toolbar.
  _controlsController!.hideToolbar();
}
```

However, `_isKeyboardOpen` uses the bottom view insets. This doesn't work if there are any `Scaffold`s in the tree without `resizeToAvoidBottomInset` set to `false`. Maybe we can check for an ancestor `KeyboardScaffoldSafeArea` and if we don't find one, we don't do the `_isKeyboardOpen` check.

### DIFF
--- a/super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
@@ -247,6 +247,7 @@ class IosDocumentTouchInteractor extends StatefulWidget {
     required this.selection,
     this.openKeyboardWhenTappingExistingSelection = true,
     required this.openSoftwareKeyboard,
+    required this.isImeConnected,
     required this.scrollController,
     required this.dragHandleAutoScroller,
     required this.fillViewport,
@@ -268,6 +269,10 @@ class IosDocumentTouchInteractor extends StatefulWidget {
 
   /// A callback that should open the software keyboard when invoked.
   final VoidCallback openSoftwareKeyboard;
+
+  /// A [ValueListenable] that should notify this widget when the IME connects
+  /// and disconnects.
+  final ValueListenable<bool> isImeConnected;
 
   /// Optional list of handlers that respond to taps on content, e.g., opening
   /// a link when the user taps on text with a link attribution.
@@ -637,7 +642,7 @@ class _IosDocumentTouchInteractorState extends State<IosDocumentTouchInteractor>
           selection.extent.nodeId == docPosition.nodeId &&
           selection.extent.nodePosition.isEquivalentTo(docPosition.nodePosition);
 
-      if (didTapOnExistingSelection && _isKeyboardOpen) {
+      if (didTapOnExistingSelection && widget.isImeConnected.value) {
         // Toggle the toolbar display when the user taps on the collapsed caret,
         // or on top of an existing selection.
         //
@@ -650,31 +655,14 @@ class _IosDocumentTouchInteractorState extends State<IosDocumentTouchInteractor>
         _controlsController!.hideToolbar();
       }
 
-      final tappedComponent = _docLayout.getComponentByNodeId(adjustedSelectionPosition.nodeId)!;
-      if (!tappedComponent.isVisualSelectionSupported()) {
-        // The user tapped a non-selectable component.
-        // Place the document selection at the nearest selectable node
-        // to the tapped component.
-        moveSelectionToNearestSelectableNode(
-          editor: widget.editor,
-          document: widget.document,
-          documentLayoutResolver: widget.getDocumentLayout,
-          currentSelection: widget.selection.value,
-          startingNode: widget.document.getNodeById(adjustedSelectionPosition.nodeId)!,
-        );
-        return;
-      } else {
-        // Place the document selection at the location where the
-        // user tapped.
-        _selectPosition(adjustedSelectionPosition);
-      }
-
-      if (didTapOnExistingSelection && widget.openKeyboardWhenTappingExistingSelection) {
-        // The user tapped on the existing selection. Show the software keyboard.
-        //
-        // If the user didn't tap on an existing selection, the software keyboard will
-        // already be visible.
-        widget.openSoftwareKeyboard();
+      if (didTapOnExistingSelection) {
+        if (widget.openKeyboardWhenTappingExistingSelection) {
+          // The user tapped on the existing selection. Show the software keyboard.
+          //
+          // If the user didn't tap on an existing selection, the software keyboard will
+          // already be visible.
+          widget.openSoftwareKeyboard();
+        }
       } else {
         final tappedComponent = _docLayout.getComponentByNodeId(adjustedSelectionPosition.nodeId)!;
         if (!tappedComponent.isVisualSelectionSupported()) {
@@ -703,16 +691,6 @@ class _IosDocumentTouchInteractorState extends State<IosDocumentTouchInteractor>
     }
 
     widget.focusNode.requestFocus();
-  }
-
-  /// Returns `true` if we *think* the software keyboard is currently open, or
-  /// `false` otherwise.
-  ///
-  /// We say "think" because Flutter doesn't report this info to us. Instead, we
-  /// inspect the bottom insets on the window, and we assume any insets greater than
-  /// zero means a keyboard is visible.
-  bool get _isKeyboardOpen {
-    return MediaQuery.viewInsetsOf(context).bottom > 0;
   }
 
   DocumentPosition _moveTapPositionToWordBoundary(DocumentPosition docPosition) {
--- a/super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
@@ -130,6 +130,7 @@ class SuperEditorIosControlsController {
     _shouldShowToolbar.dispose();
   }
 
+  /// {@template ios_use_selection_heuristics}
   /// Whether to adjust the user's selection similar to the way iOS does.
   ///
   /// For example: iOS doesn't let users tap directly on a text offset. Instead,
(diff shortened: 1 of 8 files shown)


angelosilvestre:
@matthew-carroll 

There are three failing tests that I'm not sure what to do:

- SuperEditor robot taps to place caret at a non-linebreak offset with different affinities (on iOS)
- SuperEditor text affinity upstream and downstream positions render differently at a line break (on iOS)
- SuperEditor text affinity upstream and downstream positions render the same if not at a line break (on iOS)

All of these tests place the caret at a specific position, then try to place again in the same position, but with different affinity. So, the taps are being considered to be over the caret, and we are not changing the selection.

On Android, it seems we change the selection even when tapping at the caret.

Should these tests attempt to clear the selection before trying to set the new selection?

angelosilvestre:
@matthew-carroll I modified also "SuperEditor text affinity upstream and downstream positions render the same if not at a line break (on iOS)" because it also uses the same document position with different affinities.

If this isn't ideal, then we probably should modify the checks to instead of looking at the document position look at the current caret rect.

super_editor/lib/src/super_textfield/ios/user_interaction.dart
```diff
@@ -52,6 +52,9 @@ class IOSTextFieldTouchInteractor extends StatefulWidget {
     required this.child,
   }) : super(key: key);
 
+  @visibleForTesting
+  static bool useIosSelectionHeuristics = true;
```
matthew-carroll, line 56: Please move this to the top of the class, and please document it, including why/when it should be changed.

super_editor/test/flutter_test_config.dart
```diff
@@ -1,12 +1,16 @@
...
 
+import 'package:super_editor/src/super_textfield/ios/ios_textfield.dart';
 import 'package:super_editor/src/test/test_globals.dart';
 import 'package:super_text_layout/super_text_layout.dart';
 
 Future<void> testExecutable(FutureOr<void> Function() testMain) async {
   // Disable indeterminate animations
   BlinkController.indeterminateAnimationsEnabled = false;
 
+  // Disable iOS selection heuristics.
```
matthew-carroll, line 11: Please mention what those heuristics are. The reader of this probably has no idea.

super_editor/test/super_editor/mobile/super_editor_ios_overlay_controls_test.dart
```diff
@@ -214,6 +217,63 @@ void main() {
...
+        SuperEditorInspector.findDocumentSelection(),
+        selectionEquivalentTo(const DocumentSelection.collapsed(
+          position: DocumentPosition(
+            nodeId: "1",
+            nodePosition: TextNodePosition(offset: 39),
+          ),
+        )),
+      );
+
+      // Press and drag the caret to "con|sectetur".
```
matthew-carroll, line 237: Please check all of these comments. For example, we say we're dragging the caret, but we show it in the same place. Down below we say we moved the caret...to the same place.

super_editor/test/super_editor/mobile/super_editor_ios_overlay_controls_test.dart
```diff
@@ -214,6 +217,63 @@ void main() {
...
+        SuperEditorInspector.findDocumentSelection(),
+        selectionEquivalentTo(const DocumentSelection.collapsed(
+          position: DocumentPosition(
+            nodeId: "1",
+            nodePosition: TextNodePosition(offset: 32),
+          ),
+        )),
+      );
+
+      // Tap at the caret.
```
matthew-carroll, line 259: Let's say "Tap on the caret" instead of "at".

super_editor/test/super_editor/mobile/super_editor_ios_overlay_controls_test.dart
```diff
@@ -214,6 +217,63 @@ void main() {
       expect(SuperEditorInspector.isCaretVisible(), true);
     });
 
+    testWidgetsOnIos("keeps current selection when tapping on caret", (tester) async {
```
matthew-carroll, line 220: Between the name of this test, and the dragging happening in the test, it's not clear to me what we're really trying to demonstrate, or why....

super_editor/test/super_editor/mobile/super_editor_ios_overlay_controls_test.dart
```diff
@@ -309,11 +369,15 @@ void main() {
...
+Future<void> _pumpSingleParagraphApp(
+  WidgetTester tester, {
+  bool useIosSelectionHeuristics = false,
+}) async {
   await tester
       .createDocument()
       // Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor...
       .withSingleParagraph()
       .simulateSoftwareKeyboardInsets(true)
+      .useIosSelectionHeuristics(useIosSelectionHeuristics)
```
matthew-carroll, line 381: If this is configured on our editor configurator, then why do we also have it defined statically?

super_editor/test/super_textfield/ios/super_textfield_ios_selection_test.dart
```diff
@@ -53,6 +54,56 @@ void main() {
...
+        ),
+      );
+
+      // Ensure there's no selection to begin with.
+      expect(
+        SuperTextFieldInspector.findSelection(),
+        const TextSelection.collapsed(offset: -1),
+      );
+
+      // Tap at "ips|um" to place the caret at the end of the word.
```
matthew-carroll, line 75: Please mention how it ends up at the end of the word. When reading this comment in isolated, it looks like the first half of the comment contradicts the second half of the comment.

super_editor/test/super_textfield/ios/super_textfield_ios_selection_test.dart
```diff
@@ -53,6 +54,56 @@ void main() {
       // Ensure that the text field toolbar disappeared.
       expect(find.byType(IOSTextEditingFloatingToolbar), findsNothing);
     });
+
+    testWidgetsOnIos("keeps current selection when tapping on caret", (tester) async {
```
matthew-carroll, line 58: Similar to the other test - is this really just about tapping on a caret? Because dragging doesn't seem necessary for that specific check. Seems like we're missing some additional important details about the goal.

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
```diff
@@ -908,7 +908,12 @@ class _IosDocumentTouchInteractorState extends State<IosDocumentTouchInteractor>
     }
 
     final extentRect = _docLayout.getRectForPosition(collapsedPosition)!;
-    final caretRect = Rect.fromLTWH(extentRect.left - 1, extentRect.center.dy, 1, 1).inflate(24);
+    final caretRect = Rect.fromLTRB(
```
matthew-carroll, line 911: Let's call this `caretHitArea`

super_editor/test/flutter_test_config.dart
angelosilvestre: Updated.

super_editor/test/super_editor/mobile/super_editor_ios_overlay_controls_test.dart
angelosilvestre: I'm dragging the caret because that's the only way to simulate an user interaction to place the caret at a specific position that's not a word boundary.

Without dragging we need to manually issue a selection change request.

super_editor/test/super_editor/mobile/super_editor_ios_overlay_controls_test.dart
angelosilvestre: We are not showing it in the same place, we are dragging from offset 39 up to 32.

```dart
// con|sectetur
//     <------|
```

super_editor/test/super_editor/mobile/super_editor_ios_overlay_controls_test.dart
angelosilvestre: It defaults to `false` in the configurator. Here it allows tests to enable the heuristics.

super_editor/test/super_textfield/ios/super_textfield_ios_selection_test.dart
angelosilvestre: Dragging is the only user interaction that can place the caret at the middle of a word when the selection heuristics are enabled. Should I mention this in the comment?

super_editor/test/super_textfield/ios/super_textfield_ios_selection_test.dart
angelosilvestre: Updated.

angelosilvestre:
@matthew-carroll I pushed a change that I forgot to push before.

matthew-carroll (approved):
LGTM

PR #2460, opened 2024-12-19
[SuperEditor] Add `ClearDocumentRequest` (Resolves #2360)

### LINKED ISSUE #2360: [SuperEditor] - Create a `clearDocumentRequest`
Sometimes an app wants to clear all the content in a document. This is especially common when `SuperEditor` is used to send chat messages. The document needs to be cleared when each message is sent.

Currently, this process involves sending a few requests:
- insert new empty paragraph at start of document
- move caret to new empty paragraph
- clear composing region
- delete all existing nodes

To help apps avoid bugs, we should make this process easier by creating a singular `clearDocumentRequest` which does all of these things on the app's behalf.

When writing tests, try to include verification that the IME mapping doesn't mess up as a result of this request. For example, the document should serialize to the IME without issue, the selection should map from the document to the IME, and the composing region should map from the document to the IME.

angelosilvestre:
[SuperEditor] Add `ClearDocumentRequest`. Resolves #2360

Clearing a document involves:
- Removing all the nodes
- Adding a new empty paragraph
- Placing the caret at the newly created paragraph
- Clearing the composing region

This PR adds a `ClearDocumentRequest`, which packs all of those actions in a single request.

### DIFF
--- a/super_editor/lib/src/core/editor.dart
+++ b/super_editor/lib/src/core/editor.dart
@@ -1246,6 +1246,12 @@ class MutableDocument with Iterable<DocumentNode> implements Document, Editable 
     return isRemoved;
   }
 
+  /// Deletes all nodes from the [Document].
+  void clear() {
+    _nodes.clear();
+    _refreshNodeIdCaches();
+  }
+
   /// Moves a [DocumentNode] matching the given [nodeId] from its current index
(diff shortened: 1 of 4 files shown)


matthew-carroll (approved):
LGTM

PR #2476, opened 2024-12-23
[SuperEditor] - Make it possible to open a keyboard panel before the keyboard opens

matthew-carroll:
[SuperEditor] - Make it possible to open a keyboard panel before the keyboard opens


matthew-carroll:
@angelosilvestre I'm gonna merge this so I can cut a release and use this over in ClickUp. Feel free to follow up with any issues.

angelosilvestre (commented):
LGTM

PR #2480, opened 2024-12-27
[SuperEditor] Fix crash with selection across nodes when using the tags plugin (Resolves #2479)

### LINKED ISSUE #2479: [BUG] - Crash when selecting across paragraphs with action tag plugin
**Package Version**
Latest main

**To Reproduce**
Select a paragraph node completely (all text) and then continue the selection to another paragraph node above it. Or simply select multiple nodes partially.

**Actual behavior**
The editor crashes with the following exception: 

```
Exception has occurred.
RangeError (RangeError (end): Invalid value: Not in inclusive range 10..56: 68)
```

**Expected behavior**
The editor shouldn't crash.

**Additional context**
Introduced in https://github.com/superlistapp/super_editor/pull/2387

angelosilvestre:
[SuperEditor] Fix crash with selection across nodes when using the tags plugin. Resolves #2479

To reproduce the issue:

Select a paragraph node completely (all text) and then continue the selection to another paragraph node above it. Or simply select multiple nodes partially.

Exception:

```
Exception has occurred.
RangeError (RangeError (end): Invalid value: Not in inclusive range 10..56: 68)
```

The issue is that we are trying to use the caret position to extract the text. However, when multiple paragraphs are selected, we are always trying to extract the text using the end position of the selection, which will be invalid when one paragraph is larger then another.

We discussed about modifying the tag find logic for the action tags to only care about collapsed selections, because an action tag only makes sense while the user is typing, and when the user is typing, the selection is always collapsed.

There are some tests that ensure we extract tags with expanded selections, so I had to modify those tests to ensure the opposite.

Also, I suggest we change the composing reaction to only start composing if the user is actually typing. Currently, if a text contains a substring that is matched as a tag, for example "/hello", just by placing the caret somewhere in the middle of "hello" the reaction already starts a tag composition. We have a ticket for that: https://github.com/superlistapp/super_editor/issues/2392.

### DIFF
--- a/super_editor/lib/src/default_editor/text_tokenizing/action_tags.dart
+++ b/super_editor/lib/src/default_editor/text_tokenizing/action_tags.dart
@@ -123,11 +123,19 @@ class SubmitComposingActionTagCommand extends EditCommand {
   void execute(EditContext context, CommandExecutor executor) {
     final document = context.document;
     final composer = context.find<MutableDocumentComposer>(Editor.composerKey);
-    if (composer.selection == null) {
+    final selection = composer.selection;
+
+    if (selection == null) {
       return;
     }
 
-    final extent = composer.selection!.extent;
+    if (!selection.isCollapsed) {
+      // Action tags are composed while the user is typing. Since the
+      // selection is expanded, the user is not typing.
+      return;
+    }
+
+    final extent = selection.extent;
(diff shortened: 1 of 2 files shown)


angelosilvestre:
@alterhuman Would you mind trying this PR?

matthew-carroll (approved):
LGTM

PR #2490, opened 2025-01-04
Fix: Make KeyboardScaffoldSafeArea work when not at bottom of screen (Resolves #2489)

### LINKED ISSUE #2489: [BUG] - Keyboard safe area doesn't handle widgets that aren't on bottom of screen
Keyboard safe area doesn't handle widgets that aren't on bottom of screen.

The keyboard safe area has been used to create mobile chat experiences. To date, these experiences always placed the chat editor at the bottom of the screen. The keyboard safe area pushes the editor up to sit above the keyboard and the toolbar. 

It turns out we never tested this behavior in situations where the chat editor sits above other content. For example, consider an app with persistent bottom navigation tabs. In such an app, the editor sits about 70 pixels above the bottom of the screen. This extra 70 pixels is then retained when moving the editor up. The result is a 70px gap between the editor and the keyboard. This gap matches whatever content appears below the editor when the keyboard is collapsed.

matthew-carroll:
Fix: Make KeyboardScaffoldSafeArea work when not at bottom of screen (Resolves #2489)

Apparently when I recently created the new implementation of the keyboard scaffold I never tested with widgets below the chat editor, such as bottom nav tabs. When there's a widget below the editor, we were pushing the editor up too far, equal to the height of whatever was below it before pushing it up.

There's one known shortcoming even with the PR. When the content below the editor changes size, the editor position is one frame behind, leading to some jitter. This is likely to happen because the widget at the bottom of the screen is probably wrapped in a `SafeArea` to deal with bottom insets.

Here's an example of the jitter:

https://github.com/user-attachments/assets/c19d4e96-d27d-424b-ba0f-87e9fa61daf5

The same demo on an SE, which doesn't the bottom safe area space, is stable:

https://github.com/user-attachments/assets/ce7a5343-bc6a-40bb-af98-40e6d6704e2e

### DIFF
--- a/super_editor/example/lib/demos/mobile_chat/demo_mobile_chat.dart
+++ b/super_editor/example/lib/demos/mobile_chat/demo_mobile_chat.dart
@@ -101,7 +101,10 @@ class _MobileChatDemoState extends State<MobileChatDemo> {
 
   @override
   Widget build(BuildContext context) {
-    return KeyboardScaffoldSafeArea(
+    return KeyboardScaffoldSafeAreaScope(
+      // ^ Share keyboard inset info throughout all subtrees. The insets will
+      //   be reported by the subtree with the editor. Those insets might then
+      //   be used by the subtree with page content, etc.
       child: DefaultTabController(
         length: 2,
         child: Scaffold(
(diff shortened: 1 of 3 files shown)


angelosilvestre (approved):
LGTM

PR #2496, opened 2025-01-08
[DemoApp] Fix editor not re-gaining focus when the app is brought to foreground (Resolves #2279)

### LINKED ISSUE #2279: [BUG] - SuperEditor clears selection when application window loses focus
**Package Version**
super_editor main branch (commit https://github.com/superlistapp/super_editor/commit/7eae101b6f1b12956aa1bb348139f50689b61244).

**User Info**
Superlist

**To Reproduce**
1. Go to SuperEditor demo on macOS (or web)
2. Click on select a word
3. Move focus to another application (e.g. chrome)
4. Move focus back to demo app

**Actual behavior**
Selection is cleared.

**Expected behavior**
Selection should be the same as before switching to another application.

**Platform**
macOS and web.

**Flutter version**
master

**Screenshots**

https://github.com/user-attachments/assets/900f7d87-0398-4ba7-aa44-bda84de84d46

angelosilvestre:
[DemoApp] Fix editor not re-gaining focus when the app is brought to foreground. Resolves #2279

There are some places where we request focus to the editor's focus node when the toolbar is closed. These code paths can be called in response to the app going to background. Doing this prevents the editor from re-gaining focus when the app is restored.

This PR changes these places to only request focus to the editor if the primary focus is not at the root  `FocusScopeNode` (which might signify that the app is going to background).

### DIFF
--- a/super_editor/example/lib/demos/example_editor/example_editor.dart
+++ b/super_editor/example/lib/demos/example_editor/example_editor.dart
@@ -170,7 +170,16 @@ class _ExampleEditorState extends State<ExampleEditor> {
     // I tried explicitly unfocus()'ing the URL textfield
     // in the toolbar but it didn't return focus to the
     // editor. I'm not sure why.
-    _editorFocusNode.requestFocus();
+    //
+    // Only do that if the primary focus is not at the root focus scope because
+    // this might signify that the app is going to the background. Removing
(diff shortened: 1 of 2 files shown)


matthew-carroll (approved):
LGTM

PR #2503, opened 2025-01-15
[Super Editor] - Fix KeyboardScaffoldSafeArea so that when there's only a single scope, that scope is used instead of MediaQuery (Resolves #2502)

### LINKED ISSUE #2502: [Super Editor] - Keyboard safe area doesn't have correct safe area in some situations
There are some situations where `KeyboardScaffoldSafeArea` reverts back to `MediaQuery` when it should use a descendant `KeyboardPanelScaffold` insets.

I don't have a repro yet, but the root cause is that the `KeyboardScaffoldSafeArea` checks for an ancestor scope, and if it doesn't find an ancestor scope, it defers to `MediaQuery`. However, the `KeyboardScaffoldSafeArea` might, itself, be the highest level scope. In that case, it should use its own geometry. But it doesn't. I just defers to `MediaQuery`.

matthew-carroll:
[Super Editor] - Fix KeyboardScaffoldSafeArea so that when there's only a single scope, that scope is used instead of MediaQuery (Resolves #2502)

The `KeyboardScaffoldSafeArea` and `KeyboardScaffoldSafeAreaScope` talk to each other in the widget tree so that they respect their ancestors. However, I focused so much on that intercommunication that I screwed up the case of a single safe area scope. This PR fixes that issue by applying the safe area insets from the singular safe area scope, when there's no ancestor.

### DIFF
--- a/super_editor/clones/quill/pubspec.lock
+++ b/super_editor/clones/quill/pubspec.lock
@@ -39,7 +39,7 @@ packages:
       path: "../../../attributed_text"
       relative: true
     source: path
-    version: "0.4.0"
+    version: "0.4.1"
   boolean_selector:
     dependency: transitive
     description:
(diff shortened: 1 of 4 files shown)


matthew-carroll:
@angelosilvestre I'm gonna merge and cut a new release to fix this bug in ClickUp. Feel free to comment after.

angelosilvestre (commented):
LGTM

PR #2511, opened 2025-01-15
[Quill] - Make consecutive node merges configurable (Resolves #2510)

### LINKED ISSUE #2510: [Quill] - Make consecutive block merges configurable
We currently have hard-coded logic in `parser.dart`, which looks for back-to-back code blocks, and then combines them into one.

This hard-coded logic doesn't work for ClickUp because they have their own code block attribution, so the `super_editor_quill` code isn't picking those up.

Moreover, there could be any variety of blocks that an app might want to combine when back-to-back, e.g., blockquote, banner.

Make this merging behavior configurable when parsing.

matthew-carroll:
[Quill] - Make consecutive node merges configurable (Resolves #2510)

We previously hard-coded the merging of consecutive code blocks into a single block, which seemed to be the standard Quill editor behavior. The Quill editor didn't seem to merge any other blocks.

However, an app might want to merge any number of block types, including blockquotes. An app might also want to merge custom block types, which this package can't possibly know about. Therefore, this PR makes the merge decision configurable.

### DIFF
--- a/super_editor_quill/lib/src/parsing/parser.dart
+++ b/super_editor_quill/lib/src/parsing/parser.dart
@@ -243,29 +243,10 @@ extension OperationParser on Operation {
           break;
         }
 
+        // Beginning with the last non-empty node, move backwards, collecting all
+        // nodes that should be merged into one.
         final nodeBeforeTrailingNewline = document.getNodeBefore(document.last)!;
         final blockTypeToMerge = nodeBeforeTrailingNewline.getMetadataValue(NodeMetadata.blockType);
-        var shouldMerge = false;
-        for (final rule in blockMergeRules) {
-          final ruleShouldMerge = rule.shouldMerge(blockTypeToMerge);
-          if (ruleShouldMerge == true) {
-            // The rule says we definitely want to merge.
-            shouldMerge = true;
-            break;
-          }
-          if (ruleShouldMerge == false) {
-            // The rule says we definitely don't want to merge.
-            shouldMerge = false;
-            break;
-          }
-        }
-        if (!shouldMerge) {
-          // Our merge rules don't want us to merge this node.
-          break;
-        }
-
-        // The most recently inserted node supports merging. Check if we should merge with
-        // the previous node.
         var blocksToMerge = <ParagraphNode>[];
         for (int i = document.nodeCount - 2; i >= 0; i -= 1) {
           final node = document.getNodeAt(i)!;
@@ -273,9 +254,22 @@ extension OperationParser on Operation {
             break;
           }
 
-          if (node.getMetadataValue("blockType") != blockTypeToMerge) {
-            // This node doesn't have the same block type as the one(s) we want
-            // to merge. Break out the of loop.
+          var shouldMerge = false;
+          for (final rule in blockMergeRules) {
+            final ruleShouldMerge = rule.shouldMerge(blockTypeToMerge, node.getMetadataValue(NodeMetadata.blockType));
+            if (ruleShouldMerge == true) {
+              // The rule says we definitely want to merge.
+              shouldMerge = true;
+              break;
+            }
+            if (ruleShouldMerge == false) {
+              // The rule says we definitely don't want to merge.
+              shouldMerge = false;
+              break;
+            }
+          }
+          if (!shouldMerge) {
+            // Our merge rules don't want us to merge this node.
             break;
           }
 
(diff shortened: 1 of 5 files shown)


matthew-carroll:
@angelosilvestre I'm gonna merge in to cut a release. Feel free to review after.

super_editor_quill/lib/src/parsing/parser.dart
```diff
@@ -197,44 +236,65 @@ extension OperationParser on Operation {
...
         for (int i = document.nodeCount - 2; i >= 0; i -= 1) {
           final node = document.getNodeAt(i)!;
           if (node is! ParagraphNode) {
             break;
           }
-          if (node.getMetadataValue("blockType") != codeAttribution) {
+
+          var shouldMerge = false;
+          for (final rule in blockMergeRules) {
+            final ruleShouldMerge = rule.shouldMerge(blockTypeToMerge, node.getMetadataValue(NodeMetadata.blockType));
```
angelosilvestre, line 259: Just a small detail, we could cache `node.getMetadataValue(NodeMetadata.blockType)` in a variable outside of this for loop.

super_editor_quill/lib/src/parsing/parser.dart
```diff
@@ -510,3 +570,39 @@ enum DeltaOperationType {
...
+
+/// A rule that decides whether a given [DocumentNode] should be merged into
+/// the node before it, when creating a [Document] from Quill Deltas.
+///
+/// This is useful, for example, to place multiple lines of code within a
+/// single code block.
+abstract interface class DeltaBlockMergeRule {
+  /// Returns `true` if two consecutive blocks with the given types should merge,
+  /// `false` if they shouldn't, or `null` if this rule has no opinion about the merge.
+  bool? shouldMerge(Attribution block1, Attribution block2);
```
angelosilvestre, line 588: Using a nullable `bool` looks a bit weird to me, because when reading the code that uses this, it isn't immediately obvious. I would suggest to use an enum instead.

super_editor_quill/test/parsing/multiline_parsing_test.dart
```diff
@@ -0,0 +1,184 @@
...
+            },
+          );
+
+          expect(
+            (document.getNodeAt(0)! as ParagraphNode).text.toPlainText(),
+            "This is a blockquote\nThis is line two",
+          );
+          expect((document.getNodeAt(0)! as ParagraphNode).getMetadataValue("blockType"), blockquoteAttribution);
+          expect((document.getNodeAt(1)! as ParagraphNode).text.toPlainText(), "");
+          expect(document.nodeCount, 2);
```
angelosilvestre, line 65: Shouldn't the node count expectation be placed before the other ones? If, for some reason, there is only one node in the document the test will fail in the line above.

super_editor_quill/test/parsing/multiline_parsing_test.dart
```diff
@@ -0,0 +1,184 @@
...
+            "",
+          );
+          expect(
+            (document.getNodeAt(2)! as ParagraphNode).text.toPlainText(),
+            "This is line two",
+          );
+          expect((document.getNodeAt(0)! as ParagraphNode).getMetadataValue("blockType"), blockquoteAttribution);
+          expect((document.getNodeAt(1)! as ParagraphNode).text.toPlainText(), "");
+          expect((document.getNodeAt(2)! as ParagraphNode).getMetadataValue("blockType"), blockquoteAttribution);
+          expect(document.nodeCount, 4);
```
angelosilvestre, line 106: Same here.

super_editor_quill/test/parsing/multiline_parsing_test.dart
```diff
@@ -0,0 +1,184 @@
...
+            },
+          );
+
+          expect(
+            (document.getNodeAt(0)! as ParagraphNode).text.toPlainText(),
+            "This is a code block\nThis is line two\nThis is line three",
+          );
+          expect((document.getNodeAt(0)! as ParagraphNode).getMetadataValue("blockType"), codeAttribution);
+          expect((document.getNodeAt(1)! as ParagraphNode).text.toPlainText(), "");
+          expect(document.nodeCount, 2);
```
angelosilvestre, line 142: And here.

super_editor_quill/test/parsing/multiline_parsing_test.dart
```diff
@@ -0,0 +1,184 @@
...
+            "",
+          );
+          expect(
+            (document.getNodeAt(2)! as ParagraphNode).text.toPlainText(),
+            "This is line two",
+          );
+          expect((document.getNodeAt(0)! as ParagraphNode).getMetadataValue("blockType"), codeAttribution);
+          expect((document.getNodeAt(1)! as ParagraphNode).text.toPlainText(), "");
+          expect((document.getNodeAt(2)! as ParagraphNode).getMetadataValue("blockType"), codeAttribution);
+          expect(document.nodeCount, 4);
```
angelosilvestre, line 179: And here.

PR #2514, opened 2025-01-17
Export src/infrastructure/platforms/platform.dart

miguelcmedeiros:
Exports `src/infrastructure/platforms/platform.dart` that it's possible to access `debugIsWebOverride` in tests.


matthew-carroll:
If you need this for tests, and if that's the only reasonable public use-case, can you export from the test lib, instead of the main lib?

miguelcmedeiros:
> If you need this for tests, and if that's the only reasonable public use-case, can you export from the test lib, instead of the main lib?

@matthew-carroll ✅

matthew-carroll (approved):
LGTM

matthew-carroll:
@angelosilvestre can you cherry pick this whenever you get a chance?

PR #2516, opened 2025-01-18
[Attributed Text] - Fix: A few methods still (erroneously) using plain text length without accounting for placeholders (Resolves #2515)

### LINKED ISSUE #2515: [Attributed Text] - Fix copyAndAppend() and other places not accounting for placeholders
In a client app I discovered that when calling `AttributedText.copyAndAppend(moreText)` it can throw an exception when a placeholder is in the originating `AttributedText`. This is because `copyAndAppend()` (and maybe other methods) are still using `_text.length`. That should never be used anymore. We must always use the `length` of the `AttributedText` so that we account for placeholders.

Make that change, add relevant tests, cut a new release.

matthew-carroll:
[Attributed Text] - Fix: A few methods still (erroneously) using plain text length without accounting for placeholders (Resolves #2515)

Apparently I missed some uses of `_text.length` in `AttributedText` when I implemented placeholders. Except for special situations of internal accounting, we should never use `_text.length` anymore because it misses the placeholders.

I added a failing test for each effected method.

### DIFF
--- a/attributed_text/lib/src/attributed_text.dart
+++ b/attributed_text/lib/src/attributed_text.dart
@@ -270,14 +270,14 @@ class AttributedText {
   /// Returns all spans in this [AttributedText] for the given [attributions].
   Set<AttributionSpan> getAttributionSpans(Set<Attribution> attributions) => getAttributionSpansInRange(
         attributionFilter: (a) => attributions.contains(a),
-        range: SpanRange(0, _text.length),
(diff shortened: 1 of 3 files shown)


matthew-carroll:
@angelosilvestre I'm gonna merge so I can cut a release. Feel free to review after.

angelosilvestre (commented):
LGTM

PR #2518, opened 2025-01-18
[SuperEditor][SuperTextField] Improve RTL support (Resolves #2472)

### LINKED ISSUE #2472: Cursor Does Not Stay on the Left for RTL Input in SuperTextField
I am using the SuperTextField widget from the Super Editor package, and I encountered an issue when handling right-to-left (RTL) input. While the text direction dynamically adjusts to RTL based on the input using Directionality widget, the cursor does not stay on the left side of the text field when in RTL mode. Instead, it behaves as if it’s in left-to-right (LTR) mode, which is inconsistent with standard RTL text editing behavior.

This behavior causes confusion for users who expect the cursor to align properly for RTL languages such as Arabic or Hebrew.

Steps to Reproduce:
1- Use the SuperTextField widget with a text controller that updates Directionality dynamically based on input.
2- Enter any RTL text (e.g., Arabic or Hebrew).
3- Observe the text aligns correctly to the right, but the cursor stays on the right instead of shifting to the left, as expected for RTL input.
Expected Behavior:
  When in RTL mode:
  -The text should align to the right.
  -The cursor should remain at the leftmost position of the text field, consistent with RTL text editing behavior.
Actual Behavior:
  When in RTL mode:
  -The text aligns to the right (as expected).
  -However, the cursor stays on the right, behaving as if the input is in LTR mode.

angelosilvestre:
[SuperEditor][SuperTextField] Improve RTL support. Resolves #2472

We have some issues in RTL support:

- When typing in a SuperEditor with a RTL language, the caret always stays at the right side.
- When typing in a SuperTextField with a RTL language, the caret always stays at the right side and the text is left-aligned.

For `SuperEditor`, we have code to infer the text direction from the text itself, but we weren't passing the text direction down the  `TextComponent`, and we were also not passing it to `LayoutAwareRichText`. As the result, the caret position is computed as if we are in LTR.

For `SuperTextField`, we weren't inferring the text direction from the text.

This PR makes the following changes:

- Allows passing the text direction to `LayoutAwareRichText`.
- Makes the `SuperTextField`'s `textAlign` nullable, to choose the alignment based on the text direction, if the app doesn't enforce a text alignment.
- Applies a directionality to list items and tasks, to place the checkbox and list indicators in the right side when in RTL mode.


super_editor/lib/src/super_textfield/android/android_textfield.dart
```diff
@@ -175,6 +179,21 @@ class SuperAndroidTextFieldState extends State<SuperAndroidTextField>
 
   late ImeAttributedTextEditingController _textEditingController;
 
+  /// The text direction of the first character in the text.
+  ///
+  /// Used to align and position the caret depending on whether
+  /// the text is RTL or LTR.
+  TextDirection? _contentTextDirection;
```
matthew-carroll, line 186: What if the text includes segments in both directions? Are we assuming that a text field only contains a single text direction?

super_editor/test_goldens/editor/goldens/super-editor-rtl-caret-ordered-list-item-iOS.png
matthew-carroll, line 1: Is there a reason that these goldens use enough text to wrap lines? It seems to me that the multiple lines only gets in the way of understanding what these tests want to show - especially this one where we have black squares on both sides of the image, so it's not obvious what's being rendered. Just looks like a bunch of boxes.

Also, these file names should probably include some reference to the part we care about. So this file name makes it clear that we've got an ordered list item with RTL on iOS, but where do we expect the caret to sit? It's probably a good idea to mention that in the file name.

BTW, if this is an "ordered list item", why don't we have a numeral on it?

super_editor/test_goldens/editor/supereditor_rtl_test.dart
```diff
@@ -0,0 +1,126 @@
...
+import 'package:super_editor/src/test/super_editor_test/supereditor_robot.dart';
+import 'package:super_editor/super_editor.dart';
+
+import '../../test/super_editor/supereditor_test_tools.dart';
+import '../test_tools_goldens.dart';
+
+void main() {
+  group('SuperEditor > RTL mode >', () {
+    testGoldensOnAllPlatforms(
+      'inserts text and paints caret on the left side of paragraph',
```
matthew-carroll, line 17: Instead of just saying "left" and "right", we probably need to include a reference to both the side and the relative position. Maybe the words "upstream" and "downstream". Whatever makes the most sense. But since the point of these tests is to deal with cases where "left" and "right" mean different things in terms of content, we should make that clear.

super_editor/test_goldens/editor/goldens/super-editor-rtl-caret-ordered-list-item-iOS.png
angelosilvestre: Updated to use a single line and updated the test names.

> BTW, if this is an "ordered list item", why don't we have a numeral on it?

I'm not sure why we have the numeral on Android and Linux, but not on the other platforms.

super_editor/test_goldens/editor/supereditor_rtl_test.dart
angelosilvestre: Updated.

super_editor/lib/src/super_textfield/android/android_textfield.dart
angelosilvestre: I guess the caret probably won't work correctly if have multiple directions in the same text field. I'll need to check if `RenderParagraph` returns the correct position.

super_editor/test_goldens/editor/goldens/super-editor-rtl-caret-ordered-list-item-iOS.png
matthew-carroll: Seems like a bug. Please file an issue if there's not something obvious to fix in this PR.

matthew-carroll (approved):
LGTM - Does this PR include any breaking changes? If so, can you call them out in the PR description and provide a migration guide?

angelosilvestre:
@matthew-carroll It doesn't include breaking changes.

PR #2519, opened 2025-01-18
[SuperEditor][web] Fix double tap closing the IME connection on Safari/Firefox (Resolves #2513)

### LINKED ISSUE #2513: [SuperEditor] [Firefox/Safari] - Backspace to delete expanded selection breaks the editor
**To Reproduce**
1. Go to SuperEditor demo
2. Select a word
3. Press backspace

**Actual behavior**
Content is not deleted and afterwards it's not possible to interact with the editor anymore. Seems to be in a broken state.

**Expected behavior**
Word is deleted.

**Platform**
Worked fine on Chrome and Brave browsers. Safari and Firefox seems to be broken.

angelosilvestre:
[SuperEditor][web] Fix double tap closing the IME connection on Safari/Firefox. Resolves #2513

On Safari/Firefox, double tapping to select a word and then pressing backspace, delete, or any letter does nothing and stops the editor from responding to key presses.

I noticed this only happens when the selection is expanded by double tapping. When expanding by dragging, or pressing SHIFT + ARROW key it works correctly. What's happening is that we are closing the IME connection and then immediately reopening it, and it doesn't seem to work correctly on Safari/Firefox. This is the sequence of events:

- The user double taps
- We clear the existing selection
- Because the default value of `SuperEditorImePolicies.closeKeyboardOnSelectionLost` is `true`, we close the IME connection
- We place the new selection
- We try to open the IME selection again (which doesn't seem to work)

This PR modifies the double tap handler to only clear the existing selection if we weren't able to place a new selection. I tried with both Safari and Firefox and this seems to fix the issue.

### DIFF
--- a/super_editor/lib/src/default_editor/document_gestures_mouse.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_mouse.dart
@@ -364,10 +364,10 @@ class _DocumentMouseInteractorState extends State<DocumentMouseInteractor> with 
     }
 
     _selectionType = SelectionType.word;
-    _clearSelection();
+    bool didSelectContent = false;
 
     if (docPosition != null) {
-      bool didSelectContent = _selectWordAt(
+      didSelectContent = _selectWordAt(
         docPosition: docPosition,
         docLayout: _docLayout,
       );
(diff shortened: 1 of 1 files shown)


matthew-carroll (approved):
LGTM

PR #2520, opened 2025-01-19
[SuperTextField] Add support for inline widgets (Resolves #2507)

### LINKED ISSUE #2507: [SuperTextField] - Bring inline widgets to SuperTextField
We've completed work in `AttributedText` and `SuperEditor` to render inline widgets. Bring that capability to `SuperTextField`.

For reference, here's the meta ticket, which links to the other tickets: https://github.com/superlistapp/super_editor/issues/2439

angelosilvestre:
[SuperTextField] Add support for inline widgets. Resolves #2507

We already implemented support for inline widgets in `SuperEditor`. This PR adds support for inline widgets in `SuperTextField`.

We have a `TextSpan buildTextSpan(AttributionStyleBuilder styleBuilder)` method inside `AttributedTextEditingController`.  This method doesn't seem to belong here, and it isn't being used in the repo. I marked this method as deprecated.


super_editor/test/super_textfield/super_textfield_attributions_test.dart
```diff
@@ -31,10 +31,12 @@ void main() {
         );
 
         // Ensure the text is colored orange.
-        expect(
-          SuperTextFieldInspector.findRichText().style!.color,
-          Colors.orange,
-        );
+        for (int i = 0; i <= 9; i++) {
```
matthew-carroll, line 34: Is there something we can do to make this simpler? Doing this once or twice isn't terrible, but in theory one might want to verify details like text color for any number of cases.

In this particular case, it looks like there's a span per index? That doesn't seem correct. If we have a series of characters with identical styles, they should still be available within a single span, right?

super_editor/test/super_textfield/super_textfield_inline_widgets_test.dart
```diff
@@ -0,0 +1,412 @@
...
+      expect(
+        find.byPlaceholderName('1'),
+        findsOneWidget,
+      );
+
+      // Ensure the inline widget was rendered at the end of the textfield.
+      final inlineWidgetRect = tester.getRect(find.byPlaceholderName('1'));
+      expect(
+        inlineWidgetRect.left,
+        _getOffsetAtPosition(tester, const TextPosition(offset: 5)).dx,
```
matthew-carroll, line 87: Is this offset the left side or the right side of the character rectangle?

super_editor/test/super_textfield/super_textfield_inline_widgets_test.dart
```diff
@@ -0,0 +1,412 @@
+import 'package:flutter/gestures.dart';
```
matthew-carroll, line 1: Should we also have a few golden tests to check caret position, caret height, as well as selection rectangles when selecting over and near inline widgets? These details would be relevant both for single-line text blocks and multi-line text blocks within a text field.

super_editor/test/super_textfield/super_textfield_attributions_test.dart
angelosilvestre: > In this particular case, it looks like there's a span per index? That doesn't seem correct. If we have a series of characters with identical styles, they should still be available within a single span, right?

We do have a single span, the goal of this change is to make sure that every character has the same color. If the span at offset zero has the right color, but, somehow, we have other spans with other colors, this test would pass.

We could modify this to ensure that the `InlineSpan` has only one child span.

super_editor/test/super_textfield/super_textfield_inline_widgets_test.dart
angelosilvestre: It's the left side.

super_editor/test/super_textfield/super_textfield_inline_widgets_test.dart
angelosilvestre: Added golden tests.

super_editor/test/super_textfield/super_textfield_attributions_test.dart
matthew-carroll: Don't we usually check for exact matches of span markers, which then ensures that only two color markers exist: start and end. And also ensures no other markers exist at all?

super_editor/test/super_textfield/super_textfield_inline_widgets_test.dart
matthew-carroll: Then we should probably switch that to the right side, right? If the widget overlaps the final character then that would be a bug.

angelosilvestre: In fact, offset 5 IS the inline widget. Maybe we should check if it has a bigger offset than the character at offset 4. I update this.

super_editor/test/super_textfield/super_textfield_attributions_test.dart
angelosilvestre: But span markers are at the attributed text level. This is at the Flutter InlineSpan level.

matthew-carroll: Ok. We can leave it as-is.

matthew-carroll (approved):

PR #2524, opened 2025-01-21
[Spellcheck] - Upgrade super_editor dependency from 0.3.0-dev.3 to ^0.3.0-dev.15.

matthew-carroll:
[Spellcheck] - Upgrade `super_editor` dependency from 0.3.0-dev.3 to ^0.3.0-dev.15 to get access to new APIs that otherwise create a compilation error.


matthew-carroll:
FYI @angelosilvestre

PR #2527, opened 2025-01-21
[Spellcheck] - Check for spelling errors on initial document content (Resolves #2526)

### LINKED ISSUE #2526: [BUG] - [Spellcheck]: Not checking for spelling errors on initial document content.
After the recent completion of spellcheck on mobile, I noticed that if a document is created with spelling errors, those errors aren't identified.

Spell checking should always be applied, so if a document has pre-existing content, we need to check spelling there when the plugin initializes.

matthew-carroll:
[Spellcheck] - Check for spelling errors on initial document content (Resolves #2526)

Editors can begin with non-empty Documents. In that case, we need to find spelling errors in the initial content. This PR does that.


matthew-carroll:
@angelosilvestre I'm gonna merge this when CI goes green so I can use it with ClickUp. Feel free to review.

angelosilvestre (commented):
LGTM

PR #2532, opened 2025-01-23
[AttributedText] - Fix incorrect character lookup by index (Resolves #2531)

### LINKED ISSUE #2531: [BUG] - [AttributedText] - Character lookup logic is wrong
`AttributedText` supports looking up characters and placeholders by index. The implementation is wrong for character lookup. It pulls characters from the plain text, not the text with placeholders, which can lead to the wrong characters being returned, and exceptions trying to access indices beyond the end of the text string.

matthew-carroll:
[AttributedText] - Fix incorrect character lookup by index (Resolves #2531)

We need to lookup text characters while accounting for placeholders. Previously were were looking up characters in the plain text string.

### DIFF
--- a/attributed_text/lib/src/attributed_text.dart
+++ b/attributed_text/lib/src/attributed_text.dart
@@ -119,13 +119,13 @@ class AttributedText {
   late final String _textWithPlaceholders;
 
   /// Returns the character or placeholder at offset zero.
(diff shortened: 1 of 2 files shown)


angelosilvestre (approved):
LGTM

PR #2536, opened 2025-01-24
[Super Editor] - Prevent NaN and negative bottom insets in KeyboardScaffoldSafeArea (Resolves #2535)

### LINKED ISSUE #2535: [BUG] - [Super Editor]: Keyboard scaffold safe area using NaN for bottom insets
In a client app it was discovered that under certain conditions, when `KeyboardScaffoldSafeArea` looks up its global position, its getting back `NaN`. When attempting to use that value in padding, things blow up.

It's unclear why exactly we're getting `NaN`. The only unusual detail I can see when debugging this in the client app is that the render box for safe area says it "needs paint". This is probably the culprit because `localToGlobal()` is based on paint metrics.

However, I can't figure out why the render box is marked as needing paint. I checked the following properties:
 * hasSize: true
 * context is mounted: true
 * render box is attached: true
 * paint bounds: Rect.fromLTRB(0.0, 0.0, 430.0, 782.0)

Given that this condition seems to be out of our control, we should check for `NaN` and fizzle if that's what we get. In debug mode, we should also log a warning.

matthew-carroll:
[Super Editor] - Prevent NaN and negative bottom insets in KeyboardScaffoldSafeArea (Resolves #2535)

See the issue ticket for an explanation of the problem.

### DIFF
--- a/super_editor/lib/src/infrastructure/keyboard_panel_scaffold.dart
+++ b/super_editor/lib/src/infrastructure/keyboard_panel_scaffold.dart
@@ -1077,49 +1077,7 @@ class _KeyboardScaffoldSafeAreaState extends State<KeyboardScaffoldSafeArea> {
           return widget.child;
         }
 
-        // There's no ancestor KeyboardScaffoldSafeArea, but there might be an ancestor
-        // KeyboardScaffoldSafeAreaScope, whose insets we should use.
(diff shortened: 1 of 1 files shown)


matthew-carroll:
@angelosilvestre I'm gonna merge this to resolve a production bug in ClickUp.

angelosilvestre (commented):
LGTM

PR #2540, opened 2025-01-25
[Infrastructure] Add test platform suffixed by default (Resolves #2274)

### LINKED ISSUE #2274: [Infrastructure] - Include platform suffix in tests by default
Currently, the platform suffixes that we add in tests ("on Android", "on iOS", "on Windows", etc) are added by the test methods that run the same tests for multiple platforms, like `testWidgetsOnAllPlatforms` and `testWidgetsOnMobile`.

When we want to write a test that runs only on a specific platform, for example, using `testWidgetsOnAndroid`, we need to add the platform suffixes ourselves. This is easy to forget, so we might end up overwriting tests. For example:

```dart
testWidgetsOnAndroid("test something", (tester) async{});

testWidgetsOniOS("test something", (tester) async{});
```

Move the platform suffixes to the test methods specific to a platform.

angelosilvestre:
[Infrastructure] Add test platform suffixed by default (Resolves #2274)

In some of the tests that are specific to a single platform, we are adding the platform suffixes manually. For example: `testWidgetsOnAndroid("test something", (tester) async{});`

We already made the change to include the platform suffix to the test methods from the `flutter_test_runners` package a while ago (https://github.com/Flutter-Bounty-Hunters/testing/pull/1), but we did not release a new version since, so we should release a new version.

This PR removes the manually added suffixes and also modify some test runners declared directly in this repo.

### DIFF
--- a/super_editor/test/super_editor/mobile/super_editor_ios_swipe_to_pop_test.dart
+++ b/super_editor/test/super_editor/mobile/super_editor_ios_swipe_to_pop_test.dart
@@ -9,8 +9,7 @@ import '../supereditor_test_tools.dart';
 
 void main() {
   group('SuperEditor', () {
-    testWidgetsOnIos('keeps current selection and does not show mobile controls when swipping to pop (on iOS)',
-        (tester) async {
+    testWidgetsOnIos('keeps current selection and does not show mobile controls when swipping to pop', (tester) async {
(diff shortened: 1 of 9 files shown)


matthew-carroll (approved):

matthew-carroll:
Can you file a ticket to release flutter_test_robots and bring it into super_editor?

PR #2545, opened 2025-01-25
[SuperEditor][Web] Fix backspace to un-indent (Resolves #2456)

### LINKED ISSUE #2456: [Super Editor Web BUG] -Inconsistent Tab Deletion Behavior on First Line and Between Nodes
When trying to delete a tab in certain scenarios, the behaviour is inconsistent:
1. If the tab is at the beginning of the sentence and is on the first line, it cannot be deleted.
2. When deleting a tab between nodes, the node itself is deleted, and its content is appended to the previous node instead of just removing the tab.

**Package Version**
super_editor, stable branch

**User Info**
Reflection.app

**To Reproduce**
1. First-Line Tab Issue
- Start with a sentence on the first line with a tab at the beginning.
- Try to delete the tab with backspace.

2. Between-Nodes Tab Issue
- Place the cursor at the start of a node after a tab.
- Press backspace or delete to remove the tab.

**Actual behaviour**
- The tab at the beginning of the first line cannot be deleted.
- Deleting a tab between nodes removes the node and appends its content to the previous node.

**Expected behaviour**
- The tab at the beginning of the first line should be deleted.
- Deleting a tab between nodes should only remove the tab without merging nodes.

**Platform**
Web (Chrome)

**Flutter version**
Flutter 3.24.4 

**Videos**
https://github.com/user-attachments/assets/58bd4ff7-a8c5-4408-b10e-cd30a194f9ed

angelosilvestre:
[SuperEditor][Web] Fix backspace to un-indent (Resolves #2456)

On web, pressing BACKSPACE at the beginning of an indented paragraph or list item does not unindent it.

The issue is that, on web, we let the browser handle the deletions, so we are sending the BACKSPACE back to the IME. This was preventing the shortcuts that use backspace from executing.

This PR fixes this issue by placing `doNothingWithBackspaceOnWeb` after the shortcuts that use backspace in the keyboard handler list.

There is an existing test, but it was not running for web, I modified so it also runs on web. Also, we should probably make `SuperEditor` tests use IME as the input source by default, since I suppose the majority of apps will use IME mode.


matthew-carroll:
> Also, we should probably make SuperEditor tests use IME as the input source by default, since I suppose the majority of apps will use IME mode.

Please file an issue if this wasn't done in this PR

super_editor/lib/src/default_editor/super_editor.dart
```diff
@@ -1436,18 +1436,19 @@ final defaultImeKeyboardActions = <DocumentKeyboardAction>[
   selectAllWhenCmdAIsPressed,
   cmdBToToggleBold,
   cmdIToToggleItalics,
+  backspaceToUnIndentListItem,
+  backspaceToUnIndentParagraph,
+  backspaceToUnIndentTask,
+  backspaceToConvertTaskToParagraph,
+  backspaceToClearParagraphBlockType,
+  // Let the browser IME handle the deletion.
```
matthew-carroll, line 1444: I don't think I'm going to understand this comment within a few days.

Is this comment referring to the backspace items above? Or the "do nothing" below?

Based on what you've described in this PR, it seems like the comment clarification belongs above the backspace items, and explains why those items need to appear before the `doNothingWithBackspaceOnWeb`.

super_editor/test/super_editor/components/list_items_test.dart
```diff
@@ -328,6 +329,24 @@ void main() {
         expect(doc.last.asListItem.indent, 0);
       });
 
+      testWidgetsOnDesktopAndWeb('unindents with BACKSPACE at the beginning of the text', (tester) async {
```
matthew-carroll, line 332: "beginning of the text" is too generic. Sounds like you might be referring to the whole document.

I think this should be "unindents with BACKSPACE with caret at beginning of list item"

super_editor/lib/src/default_editor/super_editor.dart

matthew-carroll (approved):

PR #2552, opened 2025-01-26
[SuperEditor][Android] Fix toolbar being hidden when releasing a long press (Resolves #2481)

### LINKED ISSUE #2481: [BUG] - long press without selection opens the toolbar AND closes it again immidiately
**Package Version**
super_editor, GitHub, stable

**User Info**
individual

**To Reproduce**
1. Run the example app
2. Long press anywhere (with the intent of inserting someting, so without an expanded selection)
4. See the toolbar to insert someting from the clipboard appear and immidiately close again

**Minimal Reproduction Code**
Simply use the example app, no change needed.

**Actual behavior**
The Toolbar opens and closes, making it impossible to tap one of the buttons

**Expected behavior**
The long-press should open the toolbar and only close it when taping somewhere else again. That allows actually pressing a button on the toolbar.

**Platform**
Android, did not investigate others, may be relevant there too.

**Flutter version**
Flutter 3.27.1, Dart 3.6.0

**Screenshots**

See the video:

https://github.com/user-attachments/assets/3f9abe14-f4ef-44e6-8022-edefc43fdccb

**Additional context**

At first glance, it looks like the following like is incorrect and should be removed to fix this:

https://github.com/superlistapp/super_editor/blob/583d978b53896462be3e9f3a1bf3dd4607e55a53/super_editor/lib/src/default_editor/document_gestures_touch_android.dart#L731

with more context visible, the following line seems to be incorrect, which is also what i did in the video:

```dart
  void _onTapUp(TapUpDetails details) {
    // Stop waiting for a long-press to start.
    _tapDownLongPressTimer?.cancel();

    // Cancel any on-going long-press.
    if (_isLongPressInProgress) {
      _longPressStrategy = null;
      _magnifierGlobalOffset.value = null;
      _showAndHideEditingControlsAfterTapSelection(didTapOnExistingSelection: false); // ← * SEEMS INCORRECT HERE *
      return;
    }
```

angelosilvestre:
[SuperEditor][Android] Fix toolbar being hidden when releasing a long press (Resolves #2481).

On Android, long pressing and releasing the gesture on an empty paragraph, or on a whitespace, is closing the toolbar:

[before.webm](https://github.com/user-attachments/assets/b7e869cd-d778-4a5b-9384-863986cbdd1b)

This is caused because we are calling the same method that updates the overlay controls when the user just performs a single tap. That method hides the toolbar when the selection is collapsed.

This PR changes the tap up handling to call `_onLongPressEnd` if the user was performing a long press, and modify this method to show the toolbar. We also need to hide the toolbar when the document changes, so it hides after the user starts typing. This will be a breaking change if an app performs some document change when interacting the toolbar and expects it to remain open. However, we already do that on iOS, so it might not be a problem.

Also, this PR makes a change, that we already made in iOS, to avoid checking if the keyboard is open by looking for bottom inserts, because that is unreliable. Instead, we are now checking for an IME connection. This doesn't guarantee that the keyboard is visible,  but we don't have an ideal alternative yet.

[after.webm](https://github.com/user-attachments/assets/9675352f-529e-4a87-a9b9-435d9384251b)

On iOS, this seems to be already working as expected, so this PR adds a test for it.

### DIFF
--- a/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
@@ -430,6 +430,7 @@ class AndroidDocumentTouchInteractor extends StatefulWidget {
     required this.selection,
     this.openKeyboardWhenTappingExistingSelection = true,
     required this.openSoftwareKeyboard,
+    required this.isImeConnected,
     required this.scrollController,
     required this.fillViewport,
     this.contentTapHandlers,
@@ -451,6 +452,10 @@ class AndroidDocumentTouchInteractor extends StatefulWidget {
 
   /// A callback that should open the software keyboard when invoked.
   final VoidCallback openSoftwareKeyboard;
+
+  /// A [ValueListenable] that should notify this widget when the IME connects
+  /// and disconnects.
+  final ValueListenable<bool> isImeConnected;
 
   /// Optional list of handlers that respond to taps on content, e.g., opening
   /// a link when the user taps on text with a link attribution.
@@ -671,6 +676,9 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
   }
 
   void _onDocumentChange(_) {
+    // The user might start typing when the toolbar is visible. Hide it.
+    _controlsController!.hideToolbar();
+
     onNextFrame((_) {
       _ensureSelectionExtentIsVisible();
     });
@@ -751,7 +759,7 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
     if (_isLongPressInProgress) {
       _longPressStrategy = null;
       _magnifierGlobalOffset.value = null;
-      _showAndHideEditingControlsAfterTapSelection(didTapOnExistingSelection: false);
+      _onLongPressEnd();
       return;
     }
 
@@ -985,7 +993,7 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
         ..hideMagnifier()
         ..blinkCaret();
 
-      if (didTapOnExistingSelection && _isKeyboardOpen) {
+      if (didTapOnExistingSelection && widget.isImeConnected.value) {
         // Toggle the toolbar display when the user taps on the collapsed caret,
         // or on top of an existing selection.
         //
@@ -998,16 +1006,6 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
         _controlsController!.hideToolbar();
       }
     }
-  }
-
-  /// Returns `true` if we *think* the software keyboard is currently open, or
-  /// `false` otherwise.
-  ///
-  /// We say "think" because Flutter doesn't report this info to us. Instead, we
-  /// inspect the bottom insets on the window, and we assume any insets greater than
-  /// zero means a keyboard is visible.
-  bool get _isKeyboardOpen {
-    return MediaQuery.viewInsetsOf(context).bottom > 0;
   }
 
   void _onPanStart(DragStartDetails details) {
@@ -1168,7 +1166,7 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
   }
 
   void _onLongPressEnd() {
-    _longPressStrategy!.onLongPressEnd();
+    _longPressStrategy?.onLongPressEnd();
 
     // Cancel any on-going long-press.
     _longPressStrategy = null;
@@ -1178,10 +1176,12 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
 
     _controlsController!.hideMagnifier();
     if (!widget.selection.value!.isCollapsed) {
-      _controlsController!
-        ..showExpandedHandles()
-        ..showToolbar();
+      _controlsController!.showExpandedHandles();
     }
+
+    // Show the toolbar even the selection is collapsed, because the user might
+    // be long-pressing on an empty paragraph or on a whitespace.
+    _controlsController!.showToolbar();
   }
 
   void _onCaretDragEnd() {
--- a/super_editor/test/super_editor/mobile/super_editor_android_overlay_controls_test.dart
+++ b/super_editor/test/super_editor/mobile/super_editor_android_overlay_controls_test.dart
@@ -168,6 +168,36 @@ void main() {
       expect(SuperEditorInspector.isMobileMagnifierVisible(), isFalse);
     });
 
+    testWidgetsOnAndroid("shows toolbar when long pressing on an empty paragraph and hides it after typing",
+        (tester) async {
+      await tester //
+          .createDocument()
+          .withSingleEmptyParagraph()
+          .pump();
+
+      // Ensure the toolbar is not visible.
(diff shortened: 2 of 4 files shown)


super_editor/lib/src/default_editor/document_gestures_touch_android.dart
```diff
@@ -985,7 +993,7 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
         ..hideMagnifier()
         ..blinkCaret();
 
-      if (didTapOnExistingSelection && _isKeyboardOpen) {
+      if (didTapOnExistingSelection && widget.isImeConnected.value) {
```
matthew-carroll, line 996: This was "is keyboard open" now it's "is IME connected". These aren't the same thing. Which one do we actually want here?

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
```diff
@@ -1168,7 +1166,7 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
   }
 
   void _onLongPressEnd() {
-    _longPressStrategy!.onLongPressEnd();
+    _longPressStrategy?.onLongPressEnd();
```
matthew-carroll, line 1169: Why change to `?`? How do we get here without a long press strategy?

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
```diff
@@ -1178,10 +1176,12 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
 
     _controlsController!.hideMagnifier();
     if (!widget.selection.value!.isCollapsed) {
-      _controlsController!
-        ..showExpandedHandles()
-        ..showToolbar();
+      _controlsController!.showExpandedHandles();
     }
+
+    // Show the toolbar even the selection is collapsed, because the user might
```
matthew-carroll, line 1182: I'm confused about the timing of the user interactions that we're solving with this PR. I thought the PR described the toolbar disappearing after it was already visible. But this code seems to be making the toolbar appear. What's the actual bug here?

super_editor/test/super_editor/mobile/super_editor_android_overlay_controls_test.dart
```diff
@@ -168,6 +168,36 @@ void main() {
...
+        (tester) async {
+      await tester //
+          .createDocument()
+          .withSingleEmptyParagraph()
+          .pump();
+
+      // Ensure the toolbar is not visible.
+      expect(SuperEditorInspector.isMobileToolbarVisible(), isFalse);
+
+      // Long press to show the toolbar.
```
matthew-carroll, line 181: Shouldn't long press open the magnifier?

super_editor/test/super_editor/mobile/super_editor_ios_overlay_controls_test.dart
```diff
@@ -132,6 +132,36 @@ void main() {
...
+
+      // Long press, this shouldn't show the toolbar.
+      final gesture = await tester.longPressDownInParagraph('1', 0);
+
+      // Ensure the toolbar is not visible yet.
+      expect(SuperEditorInspector.isMobileToolbarVisible(), isFalse);
+
+      // Release the finger.
+      await gesture.up();
+      await tester.pumpAndSettle();
```
matthew-carroll, line 153: Why do we need to settle?

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
angelosilvestre: We actually want to know if the keyboard is open, but the method this is using isn't reliable. It will always report `false` if there is an ancestor `Scaffold` that resizes to avoid the bottom insets.

We already did this change on iOS.

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
angelosilvestre: I was calling `_onLongPressEnd` at the wrong time. Fixed it.

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
angelosilvestre: It seems we don't need this code. Just need the `_onLongPressEnd` to be called.

super_editor/test/super_editor/mobile/super_editor_android_overlay_controls_test.dart
angelosilvestre: On Android, it shouldn't. The magnifier is displayed when dragging a handle.

super_editor/test/super_editor/mobile/super_editor_ios_overlay_controls_test.dart
angelosilvestre: It seems `pump` is enough.

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
matthew-carroll: Can you use super_keyboard to get this info?

angelosilvestre: Updated. O also modified to do the same for iOS.

angelosilvestre: One caveat, I had to use `TestSuperKeyboard` to simulate the keyboard being reported as open. This is something that we might want to do by default for all mobile tests. Otherwise we might hide some bug, because the keyboard will be always reported as closed.

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
```diff
@@ -671,6 +672,9 @@ class _AndroidDocumentTouchInteractorState extends State<AndroidDocumentTouchInt
   }
 
   void _onDocumentChange(_) {
+    // The user might start typing when the toolbar is visible. Hide it.
+    _controlsController!.hideToolbar();
```
matthew-carroll, line 676: Can't this be triggered by things like other users typing elsewhere, or theoretically any change that might not be connected to the user doing anything?

super_editor/test/super_editor/mobile/super_editor_android_overlay_controls_test.dart
```diff
@@ -168,6 +169,41 @@ void main() {
...
+      final gesture = await tester.longPressDownInParagraph('1', 0);
+
+      // Ensure the toolbar is visible.
+      expect(SuperEditorInspector.isMobileToolbarVisible(), isTrue);
+
+      // Release the finger.
+      await gesture.up();
+      await tester.pumpAndSettle();
```
matthew-carroll, line 195: Please comment why we need the settle

angelosilvestre: It seems a `pump` is enough.

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
angelosilvestre: I think it's possible that other users typing will trigger this. At the moment, the document change events don't have a property to determine if the change is caused by an user interaction or by the content being changed by another source.

PR #2556, opened 2025-01-28
[Super Editor] - Fix: applying styles to text with inline placeholders deletes the placeholders (Resolves #2555)

### LINKED ISSUE #2555: [BUG][Super Editor] - Inline widgets disappear when applying text styles
Repro:
1. Create some text with an inline widget
2. Select text and select across the inline widget
3. Apply a style, like bold or italics

Expect result:
Style is applied to selected text

Actual result:
Style is applied to selected text, but the inline widget disappears.

matthew-carroll:
[Super Editor] - Fix: applying styles to text with inline placeholders deletes the placeholders (Resolves #2555)

The problem was that the code to toggle the styles was only copying the "plain text" version of the `AttributedText`, which then ignored any inline placeholders. The result was a loss of all placeholders when toggling styles.


matthew-carroll:
@angelosilvestre I'm gonna merge this to cut a release.

angelosilvestre (commented):
LGTM

PR #2562, opened 2025-02-02
[SuperEditor] Remove the usage of ScrollbarWithCustomPhysics (Resolves #2561)

### LINKED ISSUE #2561: [FEATURE] - Remove `ScrollbarWithCustomPhysics` usage from `SuperEditor`
In https://github.com/superlistapp/super_editor/pull/1649, we introduced `ScrollbarWithCustomPhysics`, to allow the use of a custom `ScrollPhysics` because, at the time, we handled the scrolling ourselves.

We don't need to use `ScrollbarWithCustomPhysics` in `SuperEditor` anymore, because we no longer handle scrolling manually.

Remove its usage from `SuperEditor`. However, we still need it in `SuperTextField` yet.

angelosilvestre:
[SuperEditor] Remove the usage of ScrollbarWithCustomPhysics (Resolves #2561)

In https://github.com/superlistapp/super_editor/pull/1649, we introduced `ScrollbarWithCustomPhysics`, to allow the use of a custom `ScrollPhysics` because, at the time, we handled the scrolling ourselves.

We don't need to use `ScrollbarWithCustomPhysics` in `SuperEditor` anymore, because we no longer handle scrolling manually. Thus, we don't need custom physics.

This PR removes the usage of `ScrollbarWithCustomPhysics` in `SuperEditor`. However, we still handle the scrolling in `SuperTextField`, so `ScrollbarWithCustomPhysics` is still used there.

This PR also use a new extension method from a `flutter_test_robots` PR to make it easier writing tests that interact with the scrollbar.

Before landing this PR, we need to release a new version of `flutter_test_robots` and remove the dependency override.

### DIFF
--- a/super_editor/lib/src/default_editor/document_scrollable.dart
+++ b/super_editor/lib/src/default_editor/document_scrollable.dart
@@ -7,7 +7,6 @@ import 'package:super_editor/src/infrastructure/_logging.dart';
 import 'package:super_editor/src/infrastructure/documents/document_scroller.dart';
 import 'package:super_editor/src/infrastructure/flutter/build_context.dart';
 import 'package:super_editor/src/infrastructure/flutter/flutter_scheduler.dart';
-import 'package:super_editor/src/infrastructure/flutter/material_scrollbar.dart';
 import 'package:super_editor/src/infrastructure/scrolling_diagnostics/_scrolling_minimap.dart';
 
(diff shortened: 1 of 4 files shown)


matthew-carroll (commented):
LGTM - Let's get test robots updated and released first, then we can finalize this PR.

PR #2567, opened 2025-02-07
[SuperEditorSpellCheck] Add API to ignore text during spellcheck (Resolves #2564)

### LINKED ISSUE #2564: [FEATURE][Spellcheck] - Add APIs to ignore text for spellcheck
Users want to ignore certain pieces of text when running spellcheck, e.g., links, user mentions, task mentions.

Add APIs to spellcheck that make it easy to ignore desired text
* By text span
* By text pattern, e.g., "@user"
* By attribution value, e.g., `== boldAttribution`
* By attribution filter, e.g., `(attribution) => some condition`

angelosilvestre:
[SuperEditorSpellCheck] Add API to ignore text during spellcheck (Resolves #2564)

This PR adds an optional `ignoreRules` to the plugin constructor, so apps can choose to ignore certain texts, like links, tags, etc.

Each rule is a function that returns a list of `TextRange`s that should be ignored. All ranges from all rules are ignored.

Since on mobile we pass the whole text at once to the spellchecker, we need a workaround to do this. This PR replaces all the ignored ranges with whitespaces, so the spellchecker doesn't report anything on those ranges.

This PR implements built-in rules for ignore by pattern, attribution and attribution filters.

Without the ignored ranges:

image

With ignored ranges:

image

### DIFF
--- a/super_editor_spellcheck/lib/src/super_editor/spelling_and_grammar_plugin.dart
+++ b/super_editor_spellcheck/lib/src/super_editor/spelling_and_grammar_plugin.dart
@@ -285,13 +285,12 @@ class SpellingAndGrammarReaction implements EditReaction {
   /// Checks every [TextNode] in the given document for spelling and grammar
   /// errors and stores them for visual styling.
   void analyzeWholeDocument(EditContext editorContext) {
-    final document = editorContext.document;
     for (final node in editorContext.document) {
       if (node is! TextNode) {
         continue;
       }
 
-      _findSpellingAndGrammarErrors(node, document);
+      _findSpellingAndGrammarErrors(node);
     }
   }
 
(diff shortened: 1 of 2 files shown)


angelosilvestre:
@matthew-carroll Do you have any ideas on what is the ideal approach to add tests for this? I think we could make the plugin take a `SpellCheckService`(Flutter's interface for spellchecking) as an optional parameter and implement a `SpellCheckService` that returns spelling errors for everything that isn't a whitespace.

Also, which other built-in rules should we add?

super_editor_spellcheck/lib/src/super_editor/spelling_and_grammar_plugin.dart
```diff
@@ -514,6 +530,83 @@ class SpellingAndGrammarReaction implements EditReaction {
     // see suggestions and select them.
     _suggestions.putSuggestions(textNode.id, spellingSuggestions);
   }
+
+  /// Filters out ranges that should be ignored from spellchecking.
+  ///
+  /// This method replaces the ignored ranges with whitespaces so that the spellchecker
+  /// doesn't see them.
+  String _filterIgnoredRanges(TextNode node, Document document) {
```
matthew-carroll, line 538: Can we remove the `document` from this? This seems to spread the `document` around to a bunch of callsites, and in theory the broader `document` shouldn't be necessary to evaluate rules for patterns, attributions, or explicit text spans.

super_editor_spellcheck/lib/src/super_editor/spelling_and_grammar_plugin.dart
```diff
@@ -785,3 +878,38 @@ class _SpellCheckerContentTapDelegate extends ContentTapDelegate {
 
   Editor? editor;
 }
+
+/// A function that determines ranges to be ignored from spellchecking.
+typedef SpellingIgnoreRule = List<TextRange> Function(TextNode node, Document document);
+
+/// A collection of built-in rules for ignoring spans of text from spellchecking.
+class SpellingIgnoreRules {
```
matthew-carroll, line 886: We should add a rule for explicit text spans, just in case an app already knows the range of text to ignore.

super_editor_spellcheck/lib/src/super_editor/spelling_and_grammar_plugin.dart
angelosilvestre: Removed.

super_editor_spellcheck/lib/src/super_editor/spelling_and_grammar_plugin.dart
angelosilvestre: Like we talked before, the knowledge about ignored ranges lives only in the `SpellingAndGrammarReaction`. So, a rule that takes a `TextRange` would apply that range for all nodes in the document. A rule to ignore a portion of the document would need to take a `DocumentRange` and a `Document`, to generate different `TextRanges` for each node, as the rules return the ignored range for a single `TextNode`.

Regarding adding option directly on our native spellchecker plugin, I'm not sure it's a good idea, because the plugin basically just exposes the macOS API's, so I think we shouldn't be adding arbitrary parameters.

matthew-carroll (approved):
LGTM - We should add some kind of tests for this. Maybe send a fake spell checker and verify that the text that's sent to the checker has blanked out the various ignore span. But I'll merge this as-is so that ClickUp can use it.

angelosilvestre:
@matthew-carroll Added a PR with tests: https://github.com/superlistapp/super_editor/pull/2574

PR #2568, opened 2025-02-07
[SuperEditor] Fix immutability violation in spellchecking styler (Resolves #2560)

### LINKED ISSUE #2560: [Spellcheck][Bug] - Immutability violation
We have some kind of mutability issue in the spellcheck styler.

For example, a client app had a crash within `spelling_and_grammar.dart` within `_applyErrors()` where we try to do the following:

```dart
viewModel.grammarErrors
  ..clear()
  ..addAll([
    //....
  ]);
```

Either we want this list to be immutable, in which case we should replace it instead of alter it.

Or, we want this list to be mutable, in which case we need to lookup where we're assigning an unmodifiable list to this variable.

angelosilvestre:
[SuperEditor] Fix immutability violation in spellchecking styler (Resolves #2560)

A crash was reported within `spelling_and_grammar.dart` within `_applyErrors()` where we try to do the following:

```dart
viewModel.grammarErrors
  ..clear()
  ..addAll([
    //....
  ]);
```

I didn't find any places where we return an unmodifiable list from the the `grammarErrors`, but I think it's safer to replace  the list instead of modifying it, because apps can add custom view models, that might return an unmodifiable list for `grammarErrors`.

This PR changes the `SpellingAndGrammarStyler` to override the viewmodel's `spellingErrors` and `grammarErrors`.

### DIFF
--- a/super_editor/lib/src/default_editor/spelling_and_grammar/spelling_and_grammar_styler.dart
+++ b/super_editor/lib/src/default_editor/spelling_and_grammar/spelling_and_grammar_styler.dart
@@ -120,11 +120,7 @@ class SpellingAndGrammarStyler extends SingleColumnLayoutStylePhase {
       // Apply it.
       viewModel.spellingErrorUnderlineStyle = _spellingErrorUnderlineStyle!;
     }
-    viewModel.spellingErrors
-      ..clear()
-      ..addAll([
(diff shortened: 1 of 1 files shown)


matthew-carroll (approved):
LGTM

PR #2574, opened 2025-02-11
[Spellcheck] Add tests for ignore rules (Resolves #2573)

### LINKED ISSUE #2573: [SpellCheck] Add tests for ignore rules
In https://github.com/superlistapp/super_editor/pull/2567 we added API's for ignoring certain pieces of text from spellchecking.

Add tests to lock that behavior.

angelosilvestre:
[Spellcheck] Add tests for ignore rules (Resolves #2573)

This is a follow up PR for https://github.com/superlistapp/super_editor/pull/2567.

We added support for "ignore rules", to allow apps to ignore from spellcheck certain pieces of text, like mentions, tags, texts with specific attributions, etc.

This PR modifies the `SpellingAndGrammarPlugin` to take an optional `SpellCheckService` parameter. This allows us to provide a fake spellchecker in tests, to verify which portions of texts were queried.

### DIFF
--- a/super_editor_spellcheck/lib/src/super_editor/spelling_and_grammar_plugin.dart
+++ b/super_editor_spellcheck/lib/src/super_editor/spelling_and_grammar_plugin.dart
@@ -37,6 +37,9 @@ class SpellingAndGrammarPlugin extends SuperEditorPlugin {
   /// - [ignoreRules]: a list of rules that determine ranges that should be ignored from spellchecking.
   ///   It can be used,  for example, to ignore links or text with specific attributions. See [SpellingIgnoreRules]
   ///   for a list of built-in rules.
+  /// - [spellCheckService]: a custom spell check service to use for spell checking. If this is provided,
+  ///   the plugin will use this service instead of the default spell check service. This is intended to
+  ///   be used in tests.
   SpellingAndGrammarPlugin({
     bool isSpellingCheckEnabled = true,
     UnderlineStyle spellingErrorUnderlineStyle = defaultSpellingErrorUnderlineStyle,
@@ -47,6 +50,7 @@ class SpellingAndGrammarPlugin extends SuperEditorPlugin {
     SuperEditorAndroidControlsController? androidControlsController,
     SuperEditorIosControlsController? iosControlsController,
     List<SpellingIgnoreRule> ignoreRules = const [],
+    SpellCheckService? spellCheckService,
   })  : _isSpellCheckEnabled = isSpellingCheckEnabled,
         _isGrammarCheckEnabled = isGrammarCheckEnabled {
     assert(defaultTargetPlatform != TargetPlatform.android || androidControlsController != null,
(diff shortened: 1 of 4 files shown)


super_editor_spellcheck/lib/src/super_editor/spelling_and_grammar_plugin.dart
```diff
@@ -88,6 +94,8 @@ class SpellingAndGrammarPlugin extends SuperEditorPlugin {
     };
   }
 
+  late final SpellCheckService? _spellCheckService;
```
matthew-carroll, line 97: Please document this. What is it? Where does it likely come from? What should a user know about how/when to provide this property?

super_editor_spellcheck/lib/src/super_editor/spelling_and_grammar_plugin.dart
```diff
@@ -254,14 +262,16 @@ extension SpellingAndGrammarEditorExtensions on Editor {
...
-  SpellingAndGrammarReaction(this._suggestions, this._styler, this._ignoreRules);
+  SpellingAndGrammarReaction(this._suggestions, this._styler, this._ignoreRules, this._customSpellCheckService);
 
   final SpellingErrorSuggestions _suggestions;
 
   final SpellingAndGrammarStyler _styler;
 
   final List<SpellingIgnoreRule> _ignoreRules;
 
+  final SpellCheckService? _customSpellCheckService;
```
matthew-carroll, line 273: Why is this called "custom"? Why not always defer to this, and by default the implementation is whatever we already have?

angelosilvestre: > Why not always defer to this, and by default the implementation is whatever we already have?

On macOS, we don't implement a `SpellCheckService`, we manually call our mac spellcheck plugin. We can't make it implement this interface, because `SpellCheckService` has no knowledge about grammar checking, which we do support on mac.

super_editor_spellcheck/lib/src/super_editor/spelling_and_grammar_plugin.dart
angelosilvestre: Updated.

super_editor_spellcheck/lib/src/super_editor/spelling_and_grammar_plugin.dart
matthew-carroll: Sounds like there are probably multiple angles of analysis and rework that needs to be done then. If grammar and spelling aren't support on all platforms, then that would seem to indicate that our public API surface should separate spelling and grammar.

Once separated, I would hope that the same interface could be applied to Mac. But if not, then we need to identify what those differences are, and then rework the way we publicly structure the concept of a spellcheck service, because clearly there's no such thing as a singular spellcheck service.

If we have something called `SpellCheckService` - it needs to work everywhere.

angelosilvestre: I added a new `GrammarCheckService`, following the same convention as `SpellCheckService`, applied both interfaces to mac and modified the plugin to accept both a `SpellCheckService` and a `GrammarCheckService`.

Can you take a look if that is what you were thinking?

matthew-carroll (approved):

matthew-carroll:
@angelosilvestre can you cherry pick this manually? GitHub won't do it because the workflow changed.

PR #2582, opened 2025-02-20
[SuperEditor][iOS] Fix selecting all document when using the native toolbar (Resolves #2579)

### LINKED ISSUE #2579: [SuperEditor][Bug] - iOS native toolbar "select all" only selects the paragraph with the caret
[SuperEditor][Bug] - iOS native toolbar "select all" only selects the paragraph with the caret

https://github.com/user-attachments/assets/02e1344a-9038-4cf4-8ce2-6a1cbcc07fe8

angelosilvestre:
[SuperEditor][iOS] Fix selecting all document when using the native toolbar (Resolves #2579).

When using the iOS native toolbar, "select all" only selects the paragraph with the caret:

https://github.com/user-attachments/assets/02e1344a-9038-4cf4-8ce2-6a1cbcc07fe8

This happens because we only send to the IME the text of the currently selected nodes. So the IME "thinks" that this is the entire text.

Ideally, Flutter should report an action for us to handle that, but I'm not sure if this is possible.

To workaround this, whenever we receive a non-text delta selecting an entire node, on iOS, we select then entire document instead:

https://github.com/user-attachments/assets/230786d5-b756-4aa2-8cf8-7444c4fbe5ba

### DIFF
--- a/super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
+++ b/super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
@@ -261,46 +261,58 @@ class TextDeltasDocumentEditor {
       // where the user can swipe over the spacebar to change the selection. This also happens
       // when the app uses the native iOS text selection toolbar and the user presses "Select all".
 
-      final extentNode = document.getNodeById(docSelection.extent.nodeId)!;
-      final isWholeNodeSelected = docSelection.start.nodeId == docSelection.end.nodeId &&
-          docSelection.start.nodePosition == extentNode.beginningPosition &&
-          docSelection.end.nodePosition == extentNode.endPosition;
-      if (defaultTargetPlatform == TargetPlatform.iOS && isWholeNodeSelected) {
-        // On iOS, when the app uses the native text selection toolbar and the user presses "Select all",
-        // Flutter sends us a non-text delta with the selection change. However, since we only send to the IME the text
-        // of the currently selected nodes, the delta reports the node being selected, not the entire
-        // document. To workaroud this, whenever iOS reports a selection change that selects an entire node,
-        // we select the entire document instead.
-        editor.execute([
-          ChangeSelectionRequest(
-            DocumentSelection(
-              base: DocumentPosition(
-                nodeId: document.first.id,
-                nodePosition: document.first.beginningPosition,
-              ),
-              extent: DocumentPosition(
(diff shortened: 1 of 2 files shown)


super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
```diff
@@ -258,15 +258,45 @@ class TextDeltasDocumentEditor {
...
-          docSelection,
-          docSelection.isCollapsed ? SelectionChangeType.placeCaret : SelectionChangeType.expandSelection,
-          SelectionReason.userInteraction,
-        ),
-        ChangeComposingRegionRequest(docComposingRegion),
-      ]);
+      // where the user can swipe over the spacebar to change the selection. This also happens
+      // when the app uses the native iOS text selection toolbar and the user presses "Select all".
+
+      final extentNode = document.getNodeById(docSelection.extent.nodeId)!;
```
matthew-carroll, line 264: Can we extract this somehow to a private function below this? There's a lot going on here that's specific to this iOS workaround.

matthew-carroll (approved):
LGTM. I'd like a readability adjustment that I mentioned in a comment. Also, let's continue digging into the root cause to figure out if it's at least possible for Flutter to tell us when the "select all" signal comes through.

super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
angelosilvestre: Adjusted. Can you take a look?

super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
```diff
@@ -258,7 +258,11 @@ class TextDeltasDocumentEditor {
     if (docSelection != null) {
       // We got a selection from the platform.
       // This could happen in some software keyboards, like GBoard,
-      // where the user can swipe over the spacebar to change the selection.
+      // where the user can swipe over the spacebar to change the selection. This also happens
+      // when the app uses the native iOS text selection toolbar and the user presses "Select all".
+
+      docSelection = _ajustWholeDocumentSelectionOnIos(docSelection);
```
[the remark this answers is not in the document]
angelosilvestre: Updated.

matthew-carroll (approved):

PR #2589, opened 2025-02-23
[SuperEditor] Add more specific SelectionChangeTypes (Resolves #2367)

### LINKED ISSUE #2367: [SuperEditor] - moving caret with up and down arrows generates a SelectionChangeEvent of type collapsedSelection
current super_editor, main branch

I'm writing a reaction which has correct the caret position if the user has moved the caret into a region of the document which is supposed to be currently hidden. For this, I search the changeList for a `SelectionChangeEvent` of type `SelectionChangeType.pushCaret`. This works for simple upstream/downstream movement of the caret by pressing left or right arrows. However, when the user presses up or down buttons, the current implementation in `CommonEditorOperations.moveCaretUp` and `moveCaretDown` will instead produce a `SelectionChangeType.collapseSelection`, *even when the selection was already collapsed before`. Maybe I misunderstood the change types, but to me this naming is confusing.

I would expect `moveCaretUp` and `moveCaretDown` to fire an event of `collapseSelection` when there actually was an expanded selection before, else fire an event of `pushCaret`. If `pushCaret` for some reason was meant for one-step only movements, I'd rather expect another type of `moveCaret`.

I can work around this by reacting to `collapseSelection` as well; I still think this is confusing.

angelosilvestre:
[SuperEditor] Add more specific SelectionChangeTypes (Resolves #2367)

Currently, when moving the caret using up or down arrow, the editor generates a selection change event with the type `SelectionChangeType.collapseSelection`.  This makes it difficult for reactions to determine if the selection moved up or down.

Also, our `SelectionChangeType`s for pushing the caret or extent don't carry the direction of the selection change.

This PR removes `SelectionChangeType.pushCaret` and `SelectionChangeType.pushExtent` in favor of more specific types:

- `SelectionChangeType.pushDownstream`
- `SelectionChangeType.pushCaretUpstream`
- `SelectionChangeType.pushCaretDown`
- `SelectionChangeType.pushCaretUp`
- `SelectionChangeType.pushExtentDownstream`
- `SelectionChangeType.pushExtentUpstream`
- `SelectionChangeType.pushExtentDown`
- `SelectionChangeType.pushExtentUp`

This is a breaking change, but I think it's much more safer to generate a compilation error instead of keeping `SelectionChangeType.pushCaretUp` and possibly silently failing at runtime.

The upgrade should be straightforward: 

Before

```dart
if (selectionChangeType == SelectionChangeType.pushCaret) {

}

switch (selectionChangeType) {
  case SelectionChangeType.pushCaret:         
    break;
}
```

After:

```dart
if (const [
  SelectionChangeType.pushCaretDownstream,
  SelectionChangeType.pushCaretUpstream,
  SelectionChangeType.pushCaretDown,
  SelectionChangeType.pushCaretUp
].contains(selectionChangeType)) {

}

switch (selectionChangeType) {
  case SelectionChangeType.pushCaretDownstream:
  case SelectionChangeType.pushCaretUpstream:
  case SelectionChangeType.pushCaretDown:
  case SelectionChangeType.pushCaretUp:
    break;
}

// The same for pushExtent.
```

On Android, we are always generating a selection change type of `SelectionChangeType.pushCaret`. This PR changes it to `SelectionChangeType.placeCaret`, but it might be better to let that decision to the touch interactor instead.

On web, things are a bit trickier. We handle selection changes by IME deltas, so we don't have information about the intent of the selection change. Should we compare the new selection, with the selection that is expected when the user presses up/down/left/right arrow to determine what is the correct selection change type?

### DIFF
--- a/super_editor/lib/src/core/document_composer.dart
+++ b/super_editor/lib/src/core/document_composer.dart
@@ -255,8 +255,8 @@ class PushCaretRequest extends ChangeSelectionRequest {
   PushCaretRequest(
     DocumentPosition newPosition,
     this.direction,
-  ) : super(DocumentSelection.collapsed(position: newPosition), SelectionChangeType.pushCaret,
-            SelectionReason.userInteraction);
+    SelectionChangeType changeType,
+  ) : super(DocumentSelection.collapsed(position: newPosition), changeType, SelectionReason.userInteraction);
 
   final TextAffinity direction;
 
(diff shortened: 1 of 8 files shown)


angelosilvestre:
@JostSchenck Could you please try this PR to see if it works for you?

JostSchenck:
@angelosilvestre Thanks a lot. Trying it out now, but will respond to the issue for the context there.

JostSchenck:
@angelosilvestre This PR solves my problem, thank you very much.

super_editor/lib/src/core/document_composer.dart
```diff
@@ -483,12 +483,33 @@ enum SelectionChangeType {
   /// dragging with the mouse.
   placeExtent,
 
-  /// Place the caret based on a desire to move the previous caret position upstream or downstream.
-  pushCaret,
+  /// Place the caret based on a desire to move the previous caret position downstream.
```
matthew-carroll, line 486: I think I/we made these descriptions too verbose to begin with. There are so many words that it actually confuses the meaning.

Let's go with: "Move the caret one unit downstream."

Please attempt to adjust the other descriptions to have similar concision for the sake of clarity.

angelosilvestre: Updated.

JostSchenck:
@angelosilvestre  @matthew-carroll Is there a chance this will be merged in the near future?

PR #2603, opened 2025-03-09
[SuperEditor][Android][Web] Fix ENTER not correctly splitting list items and tasks (Resolves #2600)

### LINKED ISSUE #2600: [SuperEditor][Web/Android] - Pressing return in list items is broken
**To Reproduce**
Depending on where the selection is in list item when return key is pressed, it will give different results, but all them are incorrect. 

Pre-condition: build the demo app for web and open an Android device (used Chrome)

1. Place the selection in a list item (ordered or unordered)
2. Press return

**Actual behavior**
Depends on where the selection is:
* end of text: 1 new empty paragraph is inserted below with the selection
* elsewhere: 2 new list item are inserted 

**Expected behavior**
Depends on where the selection is:
* end of text: 1 new empty list item is inserted below with the selection
* elsewhere: 1 new list item are inserted

**Platform**
Web on an Android device

**Flutter version**
Flutter `master`

**Screenshots**

https://github.com/user-attachments/assets/2c9137a4-3c04-4f50-a31e-b1bd770ad5c2

angelosilvestre:
[SuperEditor][Android][Web] Fix ENTER not correctly splitting list items and tasks (Resolves #2600)

On Android, pressing ENTER on a list item or task does not correctly splits the list item:

https://github.com/user-attachments/assets/2c9137a4-3c04-4f50-a31e-b1bd770ad5c2

On Android native, pressing ENTER generates only a new line insertion delta. Because of that, when receive a new line insertion, we call `onPerformAction` to notify that the `TextInputAction.newline` happened.

The issue is that, on Android Web, pressing ENTER generates both an `onPerformAction` call and a new line insertion delta. So we end up with `onPerformAction` being called twice. This causes a new list item to be inserted and being immediately converted to a paragraph.

This PR modifies the IME code to avoid calling `onPerformAction` for a new line insertion on Android Web, because the `onPerformAction` will already be called automatically.

We have tests for Android native, where we are only reporting the new line insertion deltas (as expected for Android native). I duplicated those tests to report the new line insertion and the `onPerformAction` call. 

The `pressEnterWithIme` method from `flutter_test_robots` always report both signals:

```dart
Future<void> pressEnterWithIme({
    Finder? finder,
    GetDeltaTextInputClient? getter,
}) async {
  if (!testTextInput.hasAnyClients) {
    // There isn't any IME connections.
    return;
  }

  await ime.typeText('\n', finder: finder, getter: getter);
  await pump();
  await testTextInput.receiveAction(TextInputAction.newline);
  await pump();
}
```

We should call the `receiveAction` on Android only if it's running on Web. The issue is that `flutter_test_robots` doesn't have `WebPlatformOverride` as we have in `SuperEditor`, so we can't simulate a web test here. We should find a way to make it easy to use `pressEnterWithIme` in both native and web tests without needing to modify each test that uses it.

### DIFF
--- a/super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
+++ b/super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
@@ -124,10 +124,10 @@ class TextDeltasDocumentEditor {
     editorImeLog.fine('Old text: "${delta.oldText}"');
 
     if (delta.textInserted == "\n") {
-      // On iOS, newlines are reported here and also to performAction().
-      // On Android, newlines are only reported here. So, on Android,
+      // On iOS and Android Web, newlines are reported here and also to performAction().
+      // On Android native, newlines are only reported here. So, on Android native,
       // we forward the newline action to performAction.
-      if (defaultTargetPlatform == TargetPlatform.android) {
+      if (defaultTargetPlatform == TargetPlatform.android && !CurrentPlatform.isWeb) {
         editorImeLog.fine("Received a newline insertion on Android. Forwarding to newline input action.");
         onPerformAction(TextInputAction.newline);
       } else {
@@ -197,10 +197,10 @@ class TextDeltasDocumentEditor {
     editorImeLog.fine('Old text: "${delta.oldText}"');
 
     if (delta.replacementText == "\n") {
-      // On iOS, newlines are reported here and also to performAction().
-      // On Android, newlines are only reported here. So, on Android,
+      // On iOS and Android Web, newlines are reported here and also to performAction().
+      // On Android native, newlines are only reported here. So, on Android native,
       // we forward the newline action to performAction.
-      if (defaultTargetPlatform == TargetPlatform.android) {
+      if (defaultTargetPlatform == TargetPlatform.android && !CurrentPlatform.isWeb) {
         editorImeLog.fine("Received a newline replacement on Android. Forwarding to newline input action.");
         onPerformAction(TextInputAction.newline);
       } else {
--- a/super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
+++ b/super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
@@ -124,7 +124,7 @@ class TextDeltasDocumentEditor {
     editorImeLog.fine('Old text: "${delta.oldText}"');
 
     if (delta.textInserted == "\n") {
-      // On iOS and Android Web, newlines are reported here and also to performAction().
+      // On iOS native and Android Web, newlines are reported here and also to performAction().
       // On Android native, newlines are only reported here. So, on Android native,
       // we forward the newline action to performAction.
       if (defaultTargetPlatform == TargetPlatform.android && !CurrentPlatform.isWeb) {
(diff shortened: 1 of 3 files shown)


angelosilvestre:
We have a weird golden failure that doesn't seem to be related to the changes from this PR.

supereditor_android_magnifier_screen_edges_masterImage

supereditor_android_magnifier_screen_edges_testImage

matthew-carroll:
@angelosilvestre that golden failure looks real. Can you please file a corresponding issue?

super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
```diff
@@ -124,10 +124,10 @@ class TextDeltasDocumentEditor {
     editorImeLog.fine('Old text: "${delta.oldText}"');
 
     if (delta.textInserted == "\n") {
-      // On iOS, newlines are reported here and also to performAction().
-      // On Android, newlines are only reported here. So, on Android,
+      // On iOS and Android Web, newlines are reported here and also to performAction().
```
matthew-carroll, line 127: When I first read this, I thought you meant iOS web and Android web. To make this crystal clear, please update the verbiage so that we always mention "native" or "web" after every reference to iOS or Android, within this conditional and related conditionals.

super_editor/test/super_editor/components/list_items_test.dart
```diff
@@ -443,6 +443,54 @@ void main() {
...
+        await tester.ime.sendDeltas(const [
+          TextEditingDeltaNonTextUpdate(
+            oldText: '. Item 12',
+            selection: TextSelection.collapsed(offset: 9),
+            composing: TextRange.collapsed(9),
+          ),
+        ], getter: imeClientGetter);
+
+        // On Android Web, pressing ENTER generates both a "\n" insertion and a newline input action.
+        await tester.pressEnterWithIme(getter: imeClientGetter);
```
matthew-carroll, line 469: Can you file two issue?
1. test runners: add web mocking (or whatever we call it)
2. test robots: add an optional parameter to `pressEnterWithIme()` called `simulateQuirks` which we can use to do different things based on platform.

super_editor/test/super_editor/components/list_items_test.dart
```diff
@@ -443,6 +443,54 @@ void main() {
         );
       });
 
+      testWidgetsOnWebAndroid("inserts new item upon new line insertion at end of existing item", (tester) async {
```
matthew-carroll, line 446: Can you move both of these tests after their corresponding generalized version, instead of appearing before them? I.e., I'd like to read the non-android-web version of inserting a list item before I read the android-web version of the same test.

angelosilvestre: It's already like this. The web tests were added after the native tests.

super_editor/test/super_editor/components/list_items_test.dart
angelosilvestre: I'm not sure I understood what changes are necessary for the test runners package. Currently, our web mocking lives inside `SuperEditor`.

matthew-carroll: Yeah, I think my point was that we should try to move web mocking into flutter test runners so that it's available to all other packages.

matthew-carroll (approved):

Issue #2608, opened 2025-03-12
[FEATURE] - inline widgets in List Item Component


kasyrm:
**Your Info**
Creating posts, Knips

**User Behavior**
Inline widgets should work in List Item Components

Inline widgets builders are passed to UnorderedListItemComponentViewModel and OrderedListItemComponentViewModel but they are omitted in copy() function which results in inline widgets not being displayed in list items. 

I noticed it should be enough to change the copy() function and pass builders to components.  I applied the changes [here](https://github.com/superlistapp/super_editor/pull/2609).

matthew-carroll:
Completed via PR

PR #2609, opened 2025-03-12
[FEATURE] - inline widgets in List Item Component #2608

kasyrm:
[FEATURE] - inline widgets in List Item Component (https://github.com/superlistapp/super_editor/issues/2608)

Inline widget builders passed from componentViewModel to List Item Components. Inline widget builders preserved in UnorderedListItemComponentViewModel and OrderedListItemComponentViewModel copy methods.


matthew-carroll:
@angelosilvestre can you review this? This review is for Knips.

angelosilvestre (approved):
LGTM

matthew-carroll (approved):
We may need to re-work the general approach to inline widgets based on what's reported in #2611 but for now this seems reasonable.

matthew-carroll:
@angelosilvestre can you cherry pick this when you get a minute?

PR #2616, opened 2025-03-22
[SuperEditor] - Fix selection highlight alignment for RTL languages

mohamadlounnas:
before:

Screenshot 2025-03-22 at 6 50 00 PM

after the fix:

Screenshot 2025-03-22 at 6 49 47 PM

### DIFF
--- a/super_editor_quill/lib/src/serializing/serializers.dart
+++ b/super_editor_quill/lib/src/serializing/serializers.dart
@@ -232,8 +232,8 @@ class TextBlockDeltaSerializer implements DeltaSerializer {
     // We didn't have a natural trailing newline. Insert a newline as per the
     // Delta spec.
     final newlineDelta = Operation.insert("\n", blockFormats.isNotEmpty ? blockFormats : null);
-    final previousDelta = deltas.operations[deltas.operations.length - 1];
-    if (newlineDelta.canMergeWith(previousDelta)) {
+    final previousDelta = deltas.operations.isNotEmpty ? deltas.operations[deltas.operations.length - 1] : null;
+    if (previousDelta != null && newlineDelta.canMergeWith(previousDelta)) {
       deltas.operations[deltas.operations.length - 1] = newlineDelta.mergeWith(previousDelta);
     } else {
       deltas.operations.add(newlineDelta);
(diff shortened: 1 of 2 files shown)


matthew-carroll:
@mohamadlounnas I see that you added a `Positioned` widget, but I have no idea why that fixed anything.

Can you please describe the root of the problem, and why a `Positioned` widget solves it?

mohamadlounnas:
> @mohamadlounnas I see that you added a `Positioned` widget, but I have no idea why that fixed anything.
> 
> Can you please describe the root of the problem, and why a `Positioned` widget solves it?

Hey @matthew-carroll,

The issue occurs when selecting RTL text ,the highlight is flipped because RTL text itself is mirrored, causing the highlight to start from -100%:
Screenshot 2025-03-22 at 6 50 00 PM

You can try it yourself. I fixed it by forcing the highlight layer to stay within the positive area of the stack, as shown in the image:
Screenshot 2025-03-22 at 6 49 47 PM

matthew-carroll:
@angelosilvestre can you check on the root cause for this? I'm skeptical that the correct solution is to add a `Positioned` widget, but I'm still not sure what the root cause of this really is.

mohamadlounnas:
> @angelosilvestre can you check on the root cause for this? I'm skeptical that the correct solution is to add a `Positioned` widget, but I'm still not sure what the root cause of this really is.

The main issue is that `CustomPainter` starts drawing from left to right only, and currently has no inherent size (0,0). In RTL mode, this causes it to stick to the top right and begin drawing from there towards the right.

To fix this, we can use `Positioned` (as i did) to enforce constraints on its children or wrap the `CustomPainter` with a `SizedBox` set to `width: double.infinity` to ensure it receives constraints

Screenshot 2025-04-02 at 5 37 08 AM


Screenshot 2025-04-02 at 5 17 03 AM

mohamadlounnas:
Since the issue originates from `super_text_layout` (and it may also be used in other projects), I think it's better to fix it within that package.  

For now, wrapping it with `SizedBox.shrink` seems to resolve the problem.
new pull request: https://github.com/superlistapp/super_editor/pull/2638

PR #2618, opened 2025-03-26
[SuperEditor] Fixes the overlays obscuring editor view in devtools painting mode (Resolves #2610)

### LINKED ISSUE #2610: [Super Editor][BUG] - Renders blank when Flutter dev tools shows layout bounds
Here's a screenshot from the chat example. I had the dev tools opened and had layout bounds turned on. This caused Super Editor to become a single color. Either the content is no longer rendering, or an overlay in the document is being painted opaque and blocking the content below it. Either way, we should adjust whatever we need to so that we can effectively use dev tools with Super Editor.

Image

rutvik110:
[SuperEditor] Fixes the overlays obscuring editor view in devtools painting mode (Resolves #2610)

The greyish overlays obscuring the editor view in painting mode seems to have been result of using the `SizedBox` within the document overlay builders in `defaultSuperEditorDocumentOverlayBuilders` when they don't want to build anything. 

For ex. within `SuperEditorIosToolbarFocalPointDocumentLayerBuilder`,
```dart
    if (defaultTargetPlatform != TargetPlatform.iOS || SuperEditorIosControlsScope.maybeNearestOf(context) == null) {
      // There's no controls scope. This probably means SuperEditor is configured with
      // a non-iOS gesture mode. Build nothing.
      return const ContentLayerProxyWidget(child: SizedBox());
    }
```

This is due to the render object `RenderConstrainedBox` used by `SizedBox` paints the widget with a greyish color in debug mode resulting in the overlays we see when painting mode is on. This can be confirmed by testing this in a regular app with just `SizedBox` widgets stacked on top of one another and going in painting mode.
```dart
  @override
  void debugPaintSize(PaintingContext context, Offset offset) {
    super.debugPaintSize(context, offset);
    assert(() {
      final Paint paint;
      if (child == null || child!.size.isEmpty) {
        paint = Paint()..color = const Color(0x90909090);
        context.canvas.drawRect(offset & size, paint);
      }
      return true;
    }());
  }
```

**Solution:** 

Initially tried using `SizedBox.shrink` but that seems to have no effect. Finally ended up creating a custom widget `BuildNothingBox` that builds and paints nothing and is used in place of `SizedBox` within overlay builders. This solves the issue of the greyish overlay in painting mode. 

After: 

Screenshot 2025-03-26 at 11 22 43 PM

Before:

Screenshot 2025-03-26 at 11 25 36 PM


rutvik110:
> > seems to have been result of using the SizedBox within the document overlay builders in defaultSuperEditorDocumentOverlayBuilders when they don't want to build anything
> 
> What about when things are built in the overlay? Do we end up with gray boxes again?

We don't. Things look normal in the case as far as I can tell.

rutvik110:
> > seems to have been result of using the SizedBox within the document overlay builders in defaultSuperEditorDocumentOverlayBuilders when they don't want to build anything
> 
> What about when things are built in the overlay? Do we end up with gray boxes again?

In the case of the default document overlay builders, we don't.

super_editor/lib/src/default_editor/document_caret_overlay.dart
```diff
@@ -7,6 +7,9 @@ import 'package:super_editor/src/infrastructure/documents/document_layers.dart';
 import 'package:super_editor/src/infrastructure/platforms/mobile_documents.dart';
 import 'package:super_text_layout/super_text_layout.dart';
 
+import '../infrastructure/flutter/build_nothing_box.dart';
```
matthew-carroll, line 10: Please always use package based paths - these relative paths always break when things get refactored.

super_editor/lib/src/infrastructure/flutter/build_nothing_box.dart
```diff
@@ -0,0 +1,14 @@
+import 'package:flutter/material.dart';
+import 'package:flutter/rendering.dart';
+
+/// A widget that builds and paints nothing.
+class BuildNothingBox extends LeafRenderObjectWidget {
```
matthew-carroll, line 5: Let's call this an `EmptyBox` for brevity. Also, this widget isn't really about "building" - what you've done is bypass a paint phase action, which is the lowest level action that takes place in a render object.

Please change the Dart Doc to the following, which accurately documents the purpose this widget:
```
/// A widget that takes up all available space, but unlike `SizedBox`, this widget
/// doesn't paint anything in debug mode.
///
/// Typically, when a Flutter app wants to take up space with a widget without
/// painting anything, a `SizedBox` used. However, when using Flutter debug
/// tools to see layout boundaries, every `SizedBox` paints itself with a gray
/// color. This is especially a problem when a `SizedBox` is displayed in an
/// overlay. To solve that problem, `EmptyBox` takes up space where a
/// `SizedBox` may have been used, but paints nothing in debug mode, allowing
/// users to see the content beneath the overlay.
```

super_editor/lib/src/infrastructure/flutter/build_nothing_box.dart
```diff
@@ -0,0 +1,14 @@
+import 'package:flutter/material.dart';
```
matthew-carroll, line 1: I doubt that you need to import `material` for this. Can you just import `widgets` or `foundation` or something lower level?

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
```diff
@@ -34,6 +35,7 @@ import 'package:super_editor/src/infrastructure/touch_controls.dart';
 
 import '../infrastructure/document_gestures.dart';
 import '../infrastructure/document_gestures_interaction_overrides.dart';
+import '../infrastructure/flutter/build_nothing_box.dart';
```
matthew-carroll, line 38: Same note here about imports.

super_editor/lib/src/default_editor/document_gestures_touch_android.dart
```diff
@@ -34,6 +34,8 @@ import 'package:super_editor/src/infrastructure/touch_controls.dart';
 
 import '../infrastructure/document_gestures.dart';
 import '../infrastructure/document_gestures_interaction_overrides.dart';
+import '../infrastructure/flutter/build_nothing_box.dart';
```
matthew-carroll, line 37: Same note here about imports

super_editor/lib/src/default_editor/document_gestures_touch_ios.dart
matthew-carroll, line 38: Same note here about imports

super_editor/lib/src/default_editor/document_caret_overlay.dart
rutvik110: Updated.

super_editor/lib/src/infrastructure/flutter/build_nothing_box.dart
rutvik110: Updated.

rutvik110:
@matthew-carroll Some of the goldens seems be failing on the ci. After updating the goldens locally, they're passing locally but for some reason still failing on ci. Not sure what's going on there. 

Running on Flutter master 3.31.0-1.0.pre.348

rutvik110:
@matthew-carroll Reverted the goldens updates. Good for merge now.

matthew-carroll (approved):

PR #2623, opened 2025-03-28
[Super Keyboard][Android] - Limit keyboard tracking to between onResume and onPause (Resolves #2622)

### LINKED ISSUE #2622: [Super Keyboard][BUG] - Handle Android activity pause and resume
Instead of reporting keyboard changes from `onCreate` to `onDestroy`, only report from `onResume` to `onPause`. This is because after we're paused, the IME changes reported to the `Activity` refer to the other `Activity` when the user is switching between apps.

matthew-carroll:
[Super Keyboard][Android] - Limit keyboard tracking to between onResume and onPause (Resolves #2622)

After `onPause` I found that our app seemed to be getting IME size reports from the other app that the user is switching to. Obviously, we don't want to report those to our Flutter app. Also, I think Flutter might stop processing channel messages once in the paused state.

Also, I found a place in the code where it looks like I previously forgot to update the keyboard state property to "open" or "closed", so I added those updates.

This PR also updates the demo app by adding a toggle switch to activate or deactivate the user's ability to close the keyboard by tapping outside the text field. This control is important because verifying Android "swipe to go back" behavior requires that we be able to swipe back without immediately removing focus from the text field. We want to test the actual Android back behavior, not focus change behavior.


matthew-carroll:
@angelosilvestre I'm gonna merge this so I can cut a release

angelosilvestre (commented):
LGTM

PR #2625, opened 2025-03-28
[SuperEditor] Fix Structured Content Paste Edge Case (Resolves #2626)

### LINKED ISSUE #2626: [BUG] - PasteStructuredContentEditorCommand drops initial paragraph when pasting mergeable content into blank paragraph
**Package Version**
Commit 68d0ee33ade4bf7c60f

**Minimal Reproduction Code**

<details>

  ```dart
  final document = MutableDocument.empty("1");
      final composer = MutableDocumentComposer();
      final editor = Editor(
        editables: {
          Editor.documentKey: document,
          Editor.composerKey: composer,
        },
        requestHandlers: [
          (editor, request) => request is PasteStructuredContentEditorRequest
              ? PasteStructuredContentEditorCommand(
                  content: request.content,
                  pastePosition: request.pastePosition,
                )
              : null,
          ...defaultRequestHandlers,
        ],
        reactionPipeline: List.from(defaultEditorReactions),
      );

      final pasteContent = MutableDocument(nodes: [
        ParagraphNode(id: Editor.createNodeId(), text: AttributedText("Misc Text")),
        ParagraphNode(id: Editor.createNodeId(), text: AttributedText("Other Stuff")),
      ]);

      editor.execute([
        PasteStructuredContentEditorRequest(
          content: pasteContent,
          pastePosition: DocumentPosition(
            nodeId: document.first.id,
            nodePosition: const TextNodePosition(offset: 0),
          ),
        )
      ]);
  ```
</details>

**Actual behavior**
Single paragraph with "Other Stuff" is added to the document.

**Expected behavior**
Two paragraphs with "Misc Text" and "Other Stuff"

**Platform**
Web (but presumably affects all platforms)

**Flutter version**
Please list your Flutter and Dart version info. You can find this by running `flutter --version` at the command line.

**Screenshots**

**Additional context**

AlabasterAxe:
Previously, when pasting mergable multi-paragraph content at the beginning of an empty document, the first paragraph would get dropped. This was because the code was merging the content of the initial paragraph with the blank initial ParagraphNode. The code would then check if the _original_ node was empty to determine if it should be removed but at that point it had been replaced in the document with the content of the initial pasted paragraph.

This pr just refetches the selection node by the supplied node id and uses that to determine if the initial paragraph needs to be removed.

#2626 

## Tests

  - [x] existing paste tests pass
  - [x] added new test for this case

### DIFF
--- a/super_editor/lib/src/default_editor/multi_node_editing.dart
+++ b/super_editor/lib/src/default_editor/multi_node_editing.dart
@@ -164,7 +164,11 @@ class PasteStructuredContentEditorCommand extends EditCommand {
       // We've pasted the first new node. Remove it from the nodes to insert.
       nodesToInsert.removeAt(0);
     }
-    if (currentNodeWithSelection.text.length == 0) {
+
+    // Get the initial node with selection again, as it may have been modified.
+    final latestCurrentNodeWithSelection = document.getNodeById(currentNodeWithSelection.id);
+
+    if (latestCurrentNodeWithSelection is TextNode && latestCurrentNodeWithSelection.text.length == 0) {
       // The node with the selection is an empty text node. After we use that node's
       // position to insert other nodes, we want to delete that first node, as if the
       // pasted content replaced it.
--- a/super_editor/test/super_editor/supereditor_copy_and_paste_test.dart
+++ b/super_editor/test/super_editor/supereditor_copy_and_paste_test.dart
@@ -187,5 +187,46 @@ This is the third paragraph''');
         "This is the third paragraph",
       );
     });
+
+    test("can paste at the beginning of an empty document (with merging text)", () async {
+      final document = MutableDocument.empty("1");
+      final composer = MutableDocumentComposer();
+      final editor = Editor(
+        editables: {
+          Editor.documentKey: document,
+          Editor.composerKey: composer,
+        },
+        requestHandlers: [
+          (editor, request) => request is PasteStructuredContentEditorRequest
+              ? PasteStructuredContentEditorCommand(
+                  content: request.content,
+                  pastePosition: request.pastePosition,
+                )
+              : null,
+          ...defaultRequestHandlers,
+        ],
+        reactionPipeline: List.from(defaultEditorReactions),
+      );
+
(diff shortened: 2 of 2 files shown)


matthew-carroll:
@AlabasterAxe can you clarify what the user-facing issue is? I couldn't figure it out based on your description.

Are you saying that an extra paragraph exists after paste? Or an expected paragraph disappears? Or an exception is thrown?

AlabasterAxe:
Yeah, the first paragraph in the pasted content would disappear.

AlabasterAxe:
I think the golden test failures are unrelated and the unit test failures above happened because the test was not in the same package as the fix so I moved the test to the super_editor package but let me know if there is a more preferable place for it.

matthew-carroll:
@AlabasterAxe to be crystal clear, when you say...

> Yeah, the first paragraph in the pasted content would disappear.

You're saying that if I paste the following:

```
Hello - paragraph one
World - paragraph two
```

The resulting document will be:

```
World - paragraph two
```

AlabasterAxe:
@matthew-carroll Is there anything else that I need to do for this PR?

Just to explain a bit more about why this happens, it seems like this is an artifact of the node merging logic. If the `blockType` of the nodes match, first paragraph of the pasted content is inserted into the node where the user's cursor is. Otherwise, the first node of the content is inserted into the document after the initial paragraph.

This has an unintended side effect where the user had their cursor in an empty paragraph initially. If the content is not merged with the node then there is a left-over initial paragraph which would show up as a blank line where the user pasted with their content showing up below. To avoid that case, the code checks the initial paragraph where the user pasted to see if it is empty, in which case that node is removed.

The problem arises when the first paragraph of the pasted content is mergable with the empty paragraph _and_ the node that has the cursor is empty. The content is merged into the initial paragraph by replacing the node in the document. The code then uses the now-stale node object to determine whether the initial node should be removed.

It sees that the original node _was_ empty originally and removes it, deleting the content of the initial paragraph along with it. 

Let me know if there's more tests you think I should add or if you have any feedback on the code!

PR #2634, opened 2025-04-01
[Quill] - Serialize quill inline embeds for inline placeholders (Resolves #2590)

### LINKED ISSUE #2590: [Quill] - Add hooks to create inline embeds for AttributedText placeholders
In a given document, inline placeholders within an `AttributedText` *might* need to be encoded as inline embeds within the Quill Deltas.

For example, an app might support custom emojis and/or inline images/gifs. These need to be encoded as inline embeds.

Create APIs so that apps can replace inline placeholders (when desired) with inline embeds.

Also, make sure that something reasonable happens when serializing placeholders without any desired inline embed serialization.

matthew-carroll:
[Quill] - Serialize quill inline embeds for inline placeholders (Resolves #2590)


matthew-carroll:
@angelosilvestre this can use a review at your earliest convenience

angelosilvestre (approved):
LGTM

PR #2638, opened 2025-04-02
enforce constraints to fix highlight start position

mohamadlounnas:
Screenshot 2025-04-02 at 5 37 08 AM

### DIFF
--- a/super_text_layout/lib/src/text_selection_layer.dart
+++ b/super_text_layout/lib/src/text_selection_layer.dart
@@ -22,12 +22,14 @@ class TextLayoutSelectionHighlight extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    return CustomPaint(
-      painter: TextSelectionPainter(
-        textLayout: textLayout,
-        selectionColor: style.color,
-        borderRadius: style.borderRadius,
-        textSelection: selection,
+    return SizedBox.shrink(
+      child: CustomPaint(
+        painter: TextSelectionPainter(
+          textLayout: textLayout,
+          selectionColor: style.color,
+          borderRadius: style.borderRadius,
+          textSelection: selection,
+        ),
       ),
     );
   }


matthew-carroll:
@mohamadlounnas  I see that this PR forcibly expands a widget tree to infinite width. With this change, which widget determines the final width of the custom paint?

@angelosilvestre do you have any thoughts on this change?

mohamadlounnas:
> @mohamadlounnas I see that this PR forcibly expands a widget tree to infinite width. With this change, which widget determines the final width of the custom paint?
> 
> @angelosilvestre do you have any thoughts on this change?

hey @matthew-carroll 

The `CustomPainter` starts drawing from left to right only, even when it's positioned on the right (because of how RTL layouts work).  
The issue is simply that the canvas **has no size**, so it starts drawing from position 0—which, in an RTL layout, is already on the far right!

So the solution is simply to give the canvas a defined size, so it's not forced to overflow its parent.

There are several ways to fix this. One solution is to use `Positioned`  to enforce constraints on its children. Alternatively, you can wrap the `CustomPainter` with a `SizedBox` set to `width: double.infinity` to ensure it receives proper constraints.

angelosilvestre:
@mohamadlounnas Did you find out why the `CustomPaint` is starting at the top right? I think it should always match the size of the text, i.e., starting at the top left.

mohamadlounnas:
> @mohamadlounnas Did you find out why the `CustomPaint` is starting at the top right? I think it should always match the size of the text, i.e., starting at the top left.

This is expected behavior because in RTL, content starts from the right side.
However, the canvas itself always draws from left to right — flipping it would incorrectly mirror the content.
Similar to a textbox: in RTL, it aligns to the right, but internally still draws left to right; the layout constraints prevent it from overflowing.
In this case, since the canvas size is 0, the layout can't properly align it to the right. As a result, it starts from the left (canvas (0,0)) and draws outward, which causes it to overflow beyond the parent.

look at the image I attached in the first comment

ismail-go:
I reproduced the considered bug and it still occurs. @mohamadlounnas has solved the bug partially but not completely. If you do not have a better solution for the bug please merge it with main @matthew-carroll and @angelosilvestre. My issue was the similar with this. #2472 .

And i have another request. The text direction is detected from input text in the ParagraphComponentBuilder. But if we do not have any character in the document it cannot starts from right for rtl languages. It starts from left and after the first input character it aligns to right. But we want that the caret should start from right and input characters should be the right side of the caret. As a solution for this issue you need to get a optional parameter as a textDirection for the ParagraphComponentBuilder. And after that we can set the direction of document dynamically. You can get Flutter TextFormField widget as a reference for the text direction @matthew-carroll @angelosilvestre

PR #2658, opened 2025-05-07
[Docs] - Website: New landing page, added guides.

matthew-carroll:
[Docs] - Website: New landing page, added guides.


matthew-carroll:
I'm gonna merge this to verify that the new landing page works when deployed. If it doesn't, I'll follow up with a fix. If it does, I'm going to start iterating on design changes for the guide pages.

PR #2666, opened 2025-05-10
[Clones] - Add CI build for each clone (Resolves #2664)

### LINKED ISSUE #2664: [Clones] - Add CI build for each clone
We should ensure that each clone project is able to compile on every PR so that we don't accidentally break them with API changes.

For this ticket, add a new CI build for each clone project. Each clone project should have its own CI job.

angelosilvestre:


super_clones/quill/pubspec.yaml
```diff
@@ -1,11 +1,11 @@
 name: feather
 description: "A Flutter clone of Quill"
-publish_to: 'none' # Remove this line if you wish to publish to pub.dev
+publish_to: "none" # Remove this line if you wish to publish to pub.dev
 
 version: 1.0.0+1
 
 environment:
-  sdk: '>=3.0.0 <4.0.0'
+  sdk: ">=3.0.0 <4.0.0"
```
matthew-carroll, line 8: Should we limit the upper bound on all of these to 3.6 so that we don't risk the auto-formatter kicking in?

matthew-carroll (approved):

matthew-carroll:
I approved, but just saw all tests are failing now. Might need to adjust the Flutter version or something.

angelosilvestre:
@matthew-carroll I modified how the version constraint is defined. It was reporting a version conflict.

matthew-carroll:
Now it's `^3.6.0` - will that allow resolution as `3.7.0`? Because I think `3.7.0` is where the auto formatter kicks in...

angelosilvestre:
Yeah, I reviewed and this will still allow dart 3.7.0

With this constraint:

```
sdk: ">=3.4.1 <3.7.0"
```
We are getting the following error:

```console
The current Dart SDK version is 3.9.0-100.2.beta.

Because bear requires SDK version >=3.4.1 <3.7.0, version solving failed.
```

Does that mean we cannot force it to use a previous version?

matthew-carroll:
Hm. I'm not sure what's happening there. What if you set it to exactly `3.6.0`?

matthew-carroll:
When you use `^3.6.0` does the autoformatter change everything?

angelosilvestre:
> When you use ^3.6.0 does the autoformatter change everything?

It doesn't look like is doing that.

matthew-carroll:
Ok then we can go with `^3.6.0`

matthew-carroll (approved):

PR #2667, opened 2025-05-10
[Docs] - Build website on every PR (Resolves #2661)

### LINKED ISSUE #2661: [Docs] - Build website on every PR
In the interest of never breaking the doc website, we should build it on every PR.

This doesn't mean deploy it on every PR - we should continue to do that manually. But we should ensure that the Static Shock build process completes without issue.

This check should only run on PRs against "main".

angelosilvestre:
[Docs] - Build website on every PR (Resolves #2661)

This PR adds a workflow to build the docs website for each PR that targets the main branch.


matthew-carroll:
Can we add this to the existing PR workflow?

angelosilvestre:
> Can we add this to the existing PR workflow?

I think the "on" clause can only be defined at workflow level, so we can't have a job running manually and other jobs running on PR's.

matthew-carroll:
Our existing validation and test workflow runs on PRs, and this new one runs on PRs. Where's the mismatch?

angelosilvestre:
I thought you were mentioning the build and release workflow. Moved the job to the PR validation workflow.

matthew-carroll (approved):

PR #2677, opened 2025-05-19
[SuperEditor][SuperReader] - Custom underline style configuration (Resolves #2675)

### LINKED ISSUE #2675: [SuperEditor] - Generalize underlines
We've introduced 3 specific types of underlines in Super Editor: composing region, spelling errors, and grammar errors.

Users should be able to contribute their own underline types and styles with minimal intervention.

matthew-carroll:
[SuperEditor][SuperReader] - Custom underline style configuration (Resolves #2675)

Text can now be attributed for a custom underline, and that custom underline is painted by `SuperEditor` and `SuperReader`.

The first step is to attribute some text with a `CustomUnderlineAttribution`. This attribution includes a `type` which is a `String` that can be anything. The `type` is interpreted by the style system, so as long as the `type` maps to an underline style, it will be honored. An unknown type will paint a straight black underline.

Each `TextComponentViewModel` has the stylesheet applied to it in `applyStyles()`. Each text view model now looks for a collection of underline styles: `customUnderlineStyles = styles[Styles.customUnderlineStyles];`. Each `TextComponentViewModel` holds on to those styles.

A new style phase called `CustomUnderlineStyler` inspects each `TextComponentViewModel` and looks for `CustomUnderlineAttribution`s in the text. If it finds any, then the `CustomUnderlineStyler` adds underlines to the `TextComponentViewModel` for each attribution.

After each view model runs `applyStyles()` and after the `CustomUnderlineStyler` parses the underlines out of the text, each `TextComponentViewModel` has enough information to create a list of underline styles for the text in the node. The method for this already existed before this PR - it's called `createUnderlines()`. That method now also creates these new custom underlines in addition to the existing composing region, spelling, and grammar underlines. The `createUnderlines()` method is called when the component is configured with the view model, e.g., when `ParagraphComponent` is created from the view model.

This PR also adds an `offset` property to the existing underline styles so that the underlines can be pushed up/down from their natural position.

## Unrelated Changes
As I tried to add a new property to all text component view models, it became clear that it's tedious and error prone to maintain our approach to mixing in `TextComponentViewModel` and implementing `copy()`, `==`, and `hashCode`. This requires replicating many lines per implementation and it's easy to miss one.

I refactored these relationships to significantly reduce tedious boilerplate.

Previously a call to `copy()` would create a new view model by passing every relevant property into its constructor. This required every text component to repeat all of these properties in the call to their constructors. The new approach in this PR is to leverage mutability, and let the mixin mutate all the properties that it wants. In other words, classes like `ParagraphComponentViewModel` and `ListItemComponentViewModel` now delegate to `TextComponentViewModel` to copy all the properties from `TextComponentViewModel`. As a result, we can implement new properties on `TextComponentViewModel` without altering any class that mixes in `TextComponentViewModel`.

A similar change was made to help with implementing `==` and `hashCode` via delegation to `TextComponentViewModel`.

## Demo
Screenshot 2025-05-18 at 10 52 29 PM

### DIFF
--- a/doc/website/source/super-editor/guides/_data.yaml
+++ b/doc/website/source/super-editor/guides/_data.yaml
@@ -82,6 +82,8 @@ navigation:
           url: super-editor/guides/styling/dark-mode-and-light-mode
         - title: Style a Document
           url: super-editor/guides/styling/style-a-document
+        - title: Text Underlines
+          url: super-editor/guides/styling/text-underlines
 
     - title: Editing UI
       items:
--- a/doc/website/source/super-editor/guides/styling/text-underlines.md
+++ b/doc/website/source/super-editor/guides/styling/text-underlines.md
@@ -1,0 +1,167 @@
+---
+title: Text Underlines
+---
+Underlines in Flutter text don't support any styles. They're always the same
+thickness, the same distance from the text, the same color as the text, and
+have the same square end-caps. It should be possible to control these styles,
+but Flutter doesn't expose the lower level text layout controls.
+
+Editors require custom underline painting for styles and design languages that
+don't exactly match the standard text underline. Super Editor supports custom 
+painting of underlines by manually positioning the painted lines beneath the
+relevant spans of text.
+
+## Special Underlines
+Super Editor treats some underlines as special. These include:
+
+ * The user's composing region.
+ * Spelling errors.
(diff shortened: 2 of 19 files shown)


angelosilvestre (approved):
LGTM

PR #2680, opened 2025-05-20
[SuperEditor] - Fix: Don't hide toolbar when user presses 'select all' on toolbar (Resolves #2679)

### LINKED ISSUE #2679: [SuperEditor] - Popover toolbar doesn't appear after "select all"
We expect the popover toolbar to appear when the user initially expands a selection. This works as intended when double tapping to select a word - the toolbar appears. However, when the user presses "select all" in the toolbar, the toolbar then disappears, and there's no way to get it back. This prevents the user from copying the selected content.

matthew-carroll:
[SuperEditor] - Fix: Don't hide toolbar when user presses 'select all' on toolbar (Resolves #2679)

Previously, for some reason, we were hiding the toolbar when the user presses "Select All" on the toolbar. And once the toolbar disappears, there's no way to get it back (tapping on the selection just places the caret).

This PR keeps the toolbar open after the user taps "Select All" so that the user can then copy or paste using the toolbar.

## Related Changes
This feature change broke a couple tests because in those tests, the toolbar became visible, and then blocked a tap position that was previously working. It's actually possible that these tests weren't testing what we thought up to this point, but now that they've failed by accidentally pressing a button on the toolbar, I verified with golden checks that those tests are not testing what we think.

### DIFF
--- a/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
+++ b/super_editor/lib/src/default_editor/document_gestures_touch_android.dart
@@ -316,13 +316,22 @@ class SuperEditorAndroidControlsController {
   final _shouldShowToolbar = ValueNotifier<bool>(false);
 
   /// Shows the toolbar by setting [shouldShowToolbar] to `true`.
-  void showToolbar() => _shouldShowToolbar.value = true;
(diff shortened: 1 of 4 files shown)


angelosilvestre (approved):
LGTM

PR #2687, opened 2025-05-26
[SuperKeyboard] - Report all changes together, also add bottom padding (Resolves #2686)

### LINKED ISSUE #2686: [SuperKeyboard] - Report all keyboard changes together
Currently we report each keyboard change on its own, e.g., keyboard state vs keyboard height.

This opens the door to bugs related to order of operation.

Bundle the keyboard info together and dispatch it all at the same time.

matthew-carroll:
[SuperKeyboard] - Report all changes together, also add bottom insets (Resolves #2686)

This PR reports changes to the keyboard state and keyboard height at the same time in a new geometry data structure. This was done to prevent issues that might come from order of operations where the keyboard state change is reported and THEN the keyboard height change is reported, or vis-a-versa. 

This PR also adds reporting of bottom padding on Android as a way to work around a buggy lifecycle on Android.

The buggy lifecycle on Android was found to be caused by incorrect handling of permissions in a client app. However, that lifecycle issue also demonstrated that when it happens, Flutter loses synchronization with the bottom insets and bottom padding, reporting it incorrectly for the remainder of the app run. By reporting the bottom padding ourselves, we ensure that we use the correct value.


matthew-carroll:
FYI @angelosilvestre feel free to review whenever. I'm gonna merge so I can cut a release.

PR #2690, opened 2025-05-27
[SuperEditor][Chat] - Add option for KeyboardPanelScaffold to bypass MediaQuery (Resolves #2689)

matthew-carroll:
[SuperEditor][Chat] - Add option for KeyboardPanelScaffold to bypass MediaQuery (Resolves #2689)

Apps can now explicitly request that `KeyboardPanelScaffold` pull the keyboard height from `SuperKeyboard` instead of `MediaQuery`. See the original issue for an explanation why.


angelosilvestre (commented):
LGTM

PR #2715, opened 2025-06-16
[super_editor_markdown] Fix deserialization of a list item followed by a task (Resolves #2714)

### LINKED ISSUE #2714: [SuperEditorMarkdown] - Incorrect document when deserializing a markdown with a unordered list using "-" followed by a task
**Minimal Reproduction Code**
The following unit test captures the issue:

```dart
      test('unordered list item with a dash followed by a task', () {
        const markdown = '''
- list item node 
- [ ] task node
''';

        final document = deserializeMarkdownToDocument(markdown);

        expect(document.nodeCount, 2);
        expect(document.getNodeAt(0)!, isA<ListItemNode>());
        expect(document.getNodeAt(1)!, isA<TaskNode>());
      });
```

**Actual behavior**
Creates a document with 2 unordered list item nodes.

**Expected behavior**
Creates a document with an unordered list item and a task

angelosilvestre:
[super_editor_markdown] Fix deserialization of a list item followed by a task (Resolves #2714)

Consider the following markdown, where we have a list item followed by a task:

```
- list item node 
- [ ] task node
```

This markdown produces a document with two list item nodes. 

The issue is that the list item parser consumes all contiguous items in a single call to `parse`. Since the list item syntax matches the task pattern, it considers the task as a list item.

I noticed that the markdown version that we use already supports tasks with the `UnorderedListWithCheckboxSyntax` (which just signals the base list item parser to handle tasks). With this syntax, we can correctly parse task items following list items.

I chose this approach because the other option is to copy the entire list item parser to modify it to allow us to break the list parsing when we find the task syntax. 

We already have some tests that parse tasks, which continue to pass after this change.

### DIFF
--- a/super_editor_markdown/lib/src/markdown_to_document_parsing.dart
+++ b/super_editor_markdown/lib/src/markdown_to_document_parsing.dart
@@ -36,7 +36,7 @@ MutableDocument deserializeMarkdownToDocument(
         const _ParagraphWithAlignmentSyntax(),
       ],
       const _EmptyLinePreservingParagraphSyntax(),
-      const _TaskSyntax(),
+      const md.UnorderedListWithCheckboxSyntax(),
     ],
   );
   final blockParser = md.BlockParser(markdownLines, markdownDoc);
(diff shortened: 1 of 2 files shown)


miguelcmedeiros:
@matthew-carroll @angelosilvestre 
Tested the fix and it now works well 🚀

PR #2723, opened 2025-06-20
[Spellcheck][Bug] - Fix incorrect downstream spelling underline bounds after text insertion or deletion (Resolves #2722)

### LINKED ISSUE #2722: [Spellcheck] - Underlines are in the wrong place after replacing a word
Repro:
 - Write a sentence with multiple misspelled words that have suggestions.
 - Replace the first misspelled word with a suggested correction that's a different length.

Expected result: The spelling error underlines immediately adjust so that the spelling error underlines continue to sit beneath the misspelled words.

Actual result: Initially, the spelling error underlines stay at the same indices for a couple seconds and then adjust to the new offsets of the originally misspelled words.

matthew-carroll:
[Spellcheck][Bug] - Fix incorrect downstream spelling underline bounds after text insertion or deletion (Resolves #2722)

Problem was that when text was inserted or deleted, the indices of all following text would change, but we weren't changing the text offsets of any errors that existed downstream.

This PR looks through all errors after the insertion/deletion point and either moves them back or moves them forward by the change length.

I used golden tests to verify this because we don't yet have a convenient way to query underline text offsets.


matthew-carroll:
FYI @angelosilvestre, I used our golden runner image and was able to get spellcheck goldens to actually pass in this PR in CI, so moving forward we shouldn't expect to see those fail in PRs any more.

PR #2736, opened 2025-07-07
[SuperKeyboard] - Render simulated keyboard when using KeyboardHeightSimulator (Resolves #2735)

### LINKED ISSUE #2735: [SuperKeyboard] - Show simulated software keyboard within KeyboardHeightSimulator
[SuperKeyboard] - Show simulated software keyboard within KeyboardHeightSimulator

Currently we have a placeholder widget where the simulated keyboard should appear. Replace that with something that looks somewhat like a software keyboard, so that users can write golden tests that better reflect the situation.

matthew-carroll:
[SuperKeyboard] - Render simulated keyboard when using KeyboardHeightSimulator (Resolves #2735)


matthew-carroll:
FYI @angelosilvestre

PR #2748, opened 2025-07-22
[Super Editor] - Delete selected content when user enters text with IME (Resolves #2747)

### LINKED ISSUE #2747: [SuperEditor] - When selection includes a block node, inserting a character through IME doesn't delete the content (all platforms)
Repro:
 * Insert an image
 * Below that, insert some text
 * Select the image and some of the text
 * Press a key on the software keyboard

Expected result: The selection is deleted and the character is inserted.

Actual result: Nothing happens.

Note: Pressing backspace on the software keyboard correctly deletes the selection.

Update: Root cause is that `InsertPlainTextAtCaretCommand` checks that the `start` node is a text node and bails if it isn't.

matthew-carroll:
[Super Editor] - Delete selected content when user enters text with IME (Resolves #2747)

Root cause was that `InsertPlainTextAtCaretCommand` would immediately bail out if the `start` node for text insertion wasn't a text node. This was being triggered in the case that the user drags up from text over an image and then tries to insert a character.

I change the command to delete whatever the selection is. After deleting the selection, if we're still not in a text node, we insert a text node.

I tried writing a test for this but no matter what I tried I wasn't able to get the IME simulation to work. Something within the test structure always broke. So I gave up. I tested this manually in the Mac desktop app.

### DIFF
--- a/super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
+++ b/super_editor/lib/src/default_editor/document_ime/document_delta_editing.dart
@@ -389,6 +389,7 @@ class TextDeltasDocumentEditor {
   }
 
   void replace(TextRange replacedRange, String replacementText) {
+    editorImeLog.fine("Replacing content in IME range: $replacedRange, with new text: '$replacementText'");
     final replacementSelection = _serializedDoc.imeToDocumentSelection(TextSelection(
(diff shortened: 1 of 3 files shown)


matthew-carroll:
Verified this change works in ClickUp. Merging.

PR #2766, opened 2025-08-06
[SuperEditor] Add block node for tables. (Resolves #2761)

### LINKED ISSUE #2761: [Markdown][Super Editor][Super Reader] - Add block node for tables
After #2758 we should another small step towards table support. This ticket is essentially for a "read only" version of a table node, which really amounts to the table node being a block node, which behaves like an image, or a horizontal rule.

We can call this node `TableBlockNode` and it should extend `BlockNode`. This node should support a finite number of rows and columns. 

Open question: We know that one of the places we're going to pull data from is Markdown. We know Markdown explicitly has the concept of column headers. Should we include the explicit concept of headers in this node? Or should we leave that up to the app that uses it?

In this PR, in addition to defining the node, update the existing Markdown parser to use this node.

angelosilvestre:


super_editor/lib/src/default_editor/table.dart
```diff
@@ -0,0 +1,115 @@
...
+class TableBlockNode extends BlockNode {
+  TableBlockNode({
+    required this.id,
+    required List<List<TextNode>> rows,
+    super.metadata,
+  }) : _rows = List.from(rows.map((row) => List<TextNode>.from(row))) {
+    initAddToMetadata({"blockType": const NamedAttribution("tableBlock")});
+  }
+
+  List<List<TextNode>> get rows => UnmodifiableListView(_rows);
```
matthew-carroll, line 19: This is really more of a `getCellsByRow`. Is this access really needed when you've got `getRow()`, `getColumn()`, and `getCell()`?

super_editor/lib/src/default_editor/table.dart
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  List<TextNode> getColumn(int index) {
+    if (_rows.isEmpty || index < 0 || index >= _rows[0].length) {
+      throw RangeError.range(index, 0, _rows[0].length - 1, 'index');
+    }
+    return UnmodifiableListView(_rows.map((row) => row[index]));
+  }
+
+  TextNode getCell(int rowIndex, int cellIndex) {
```
matthew-carroll, line 36: Since these parameters are both `int`s, less make them named. Also, for consistency, `cellIndex` should become `columnIndex`. A table is keyed by a row and a column.

super_editor/lib/src/default_editor/table.dart
```diff
@@ -0,0 +1,115 @@
...
+    }
+    final row = _rows[rowIndex];
+    if (cellIndex < 0 || cellIndex >= row.length) {
+      throw RangeError.range(cellIndex, 0, row.length - 1, 'cellIndex');
+    }
+    return row[cellIndex];
+  }
+
+  @override
+  final String id;
```
matthew-carroll, line 48: Please move this up to be the first property.

super_editor/lib/src/default_editor/table.dart
```diff
@@ -0,0 +1,115 @@
...
+    return buffer.toString();
+  }
+
+  @override
+  bool operator ==(Object other) =>
+      identical(this, other) ||
+      other is TableBlockNode &&
+          runtimeType == other.runtimeType &&
+          id == other.id &&
+          const DeepCollectionEquality().equals(_rows, other._rows) &&
```
matthew-carroll, line 110: As a matter of performance, it's usually a good idea to do a deep collection equality last, so that if any other simple data doesn't match, you can escape before doing the deep collection equality.

super_editor/lib/src/infrastructure/serialization/html/html_table.dart
```diff
@@ -0,0 +1,82 @@
...
+  if (node is! TableBlockNode) {
+    return null;
+  }
+  if (selection != null) {
+    if (selection is! UpstreamDownstreamNodeSelection) {
+      // We don't know how to handle this selection type.
+      return null;
+    }
+    if (selection.isCollapsed) {
+      // This selection doesn't include the table - it's a collapsed selection
```
matthew-carroll, line 23: Can you double check the other serializers and make sure we return `null` there, too? I'm wondering if I set this up to return `null` for an empty result, or whether I went with an empty string. Because `null` might mean "we don't handle this".

super_editor/lib/src/infrastructure/serialization/html/html_table.dart
```diff
@@ -0,0 +1,82 @@
...
+  return node.toHtml(document, inlineSerializers);
+}
+
+extension TableBlockNodeToHtml on TableBlockNode {
+  String toHtml(Document document, InlineHtmlSerializerChain inlineSerializers) {
+    final htmlBuffer = StringBuffer();
+    htmlBuffer.write('<table>');
+    if (rows.isNotEmpty) {
+      final headerRow = rows.first;
+      htmlBuffer.write('<thead>');
```
matthew-carroll, line 38: This is a place where HTML and Markdown differ. There's no requirement in HTML that there be any headers at all. Also, while I knew about `<th>` I don't think I knew about `<thead>`, which poses a problem.

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/thead

`<th>` can apply to any cell in an HTML table, which is what I mentioned previously about how any cell might be a header. Therefore, the use of `<th>` probably needs to come from a per-cell metadata.

Apparently groups of rows can be surrouned with `<thead>`, `<tbody>`, and `<tfoot>`.

A `<thead>` can only appear once, and must come before any `<tbody>`s. A `<tfoot>` can only appear once, and must come after any `<tbody>`s.

`<tbody>` can appear any number of times. It looks like HTML has some kind of default visual style for dividing `<tbody>`s rows from each other.

`<thead>`, `<tbody>`, and `<tfoot>`, can all have multiple rows within them.

I think we probably can't fully honor these structures without designing a table specifically for that purpose. However, I do think we can support `<thead>` and `<th>`.

Create a `tableHeader` attribution, which is applied to every cell that wants it.

When generating HTML, any cell that has that attribution should be a `<th>` instead of a `<td>`. 

Also, for `<thead>`, start scanning each row and collect all of the initial rows with 100% header cells. For each row at the beginning of the table that has 100% header cells, put those rows within a `<thead>`.

If we want to better honor the full HTML structure, we should come back later and create an `HtmlTableNode` that explicitly supported this structure.

super_editor/lib/src/infrastructure/serialization/html/html_table.dart
```diff
@@ -0,0 +1,82 @@
...
+      if (rows.length > 1) {
+        htmlBuffer.write('<tbody>');
+        for (int i = 1; i < rows.length; i++) {
+          final row = rows[i];
+          htmlBuffer.write('<tr>');
+          for (final cell in row) {
+            final cellContent = cell.text.toHtml(
+              serializers: inlineSerializers,
+            );
+            htmlBuffer.write('<td${_getTextAlignAttribute(cell)}>$cellContent</td>');
```
matthew-carroll, line 58: Please see the docs on `<td>` - I don't think alignment is supposed to be included as an attribute anymore.

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/td#deprecated_attributes

super_editor_markdown/lib/src/document_to_markdown_serializer.dart
```diff
@@ -27,21 +30,57 @@ String serializeDocumentToMarkdown(
...
+        extent: node.endPosition,
+      );
+    } else if (selectedRange != null && node.id == selectedRange.end.nodeId) {
+      // The selection starts at the beginning of this node and ends somewhere within this node.
+      nodeSelection = node.computeSelection(
+        base: node.beginningPosition,
+        extent: selectedRange.end.nodePosition,
+      );
+    } else {
+      nodeSelection = null;
```
matthew-carroll, line 73: Should we `continue` here?

super_editor_markdown/lib/src/document_to_markdown_serializer.dart
```diff
@@ -27,21 +30,57 @@ String serializeDocumentToMarkdown(
...
     }
 
-    // Serialize the current node to markdown.
-    final node = doc.getNodeAt(i)!;
     for (final serializer in nodeSerializers) {
-      final serialization = serializer.serialize(doc, node);
+      final serialization = serializer.serialize(doc, node, selection: nodeSelection);
       if (serialization != null) {
+        if (i > 0) {
```
matthew-carroll, line 79: Can we avoid this little bit of complexity by always writing a newline after each serialization?

super_editor_markdown/lib/src/document_to_markdown_serializer.dart
```diff
@@ -53,7 +92,11 @@ String serializeDocumentToMarkdown(
 
 /// Serializes a given [DocumentNode] to a Markdown `String`.
 abstract class DocumentNodeMarkdownSerializer {
-  String? serialize(Document document, DocumentNode node);
+  String? serialize(
```
matthew-carroll, line 95: Please add a Dart Doc for this method, and explain the `selection`.

super_editor_markdown/lib/src/document_to_markdown_serializer.dart
```diff
@@ -13,6 +15,7 @@ import 'super_editor_syntax.dart';
 /// provide [customNodeSerializers] to serialize those custom nodes.
 String serializeDocumentToMarkdown(
   Document doc, {
+  DocumentSelection? selection,
```
matthew-carroll, line 18: Please update the Dart Doc

super_editor_markdown/lib/src/document_to_markdown_serializer.dart
```diff
@@ -217,8 +334,25 @@ class TaskNodeSerializer extends NodeTypedDocumentNodeMarkdownSerializer<TaskNod
...
+    final textSelection = selection as TextNodeSelection?;
+    if (textSelection != null && textSelection.isCollapsed) {
+      // Selection is collapsed. Nothing is selected for copy.
+      return '';
+    }
+    final textToConvert = textSelection != null //
+        ? node.text.copyText(textSelection.start, textSelection.end)
+        : node.text;
+
+    return '- [${node.isComplete ? 'x' : ' '}] ${textToConvert.toPlainText()}';
```
matthew-carroll, line 355: Should this be `toMarkdown()`? And please ensure we add tests to catch situations like this for each type of node.

super_editor/lib/src/default_editor/table.dart
```diff
@@ -0,0 +1,115 @@
...
+import 'package:super_editor/src/default_editor/text.dart';
+
+@immutable
+class TableBlockNode extends BlockNode {
+  TableBlockNode({
+    required this.id,
+    required List<List<TextNode>> rows,
+    super.metadata,
+  }) : _rows = List.from(rows.map((row) => List<TextNode>.from(row))) {
+    initAddToMetadata({"blockType": const NamedAttribution("tableBlock")});
```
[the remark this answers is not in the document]
angelosilvestre: Maybe it's better to define a `tableBlockAttribution`?

super_editor/lib/src/infrastructure/serialization/html/html_table.dart
angelosilvestre: What this method actually does is to return a style attribute. Renamed it.

super_editor_markdown/lib/src/document_to_markdown_serializer.dart
angelosilvestre: Then we would end up with an extra newline after the last node.

super_editor_markdown/lib/src/document_to_markdown_serializer.dart
angelosilvestre: Since we do deserialize inline markdown in tests, we should also serialize it.

I had to modify a test that serializes a multiline task, because when serializing a multiline text node we should add a soft line break (two spaces at the end of the line).

Other types of nodes already have tests for inline styles.

super_editor/lib/src/default_editor/table.dart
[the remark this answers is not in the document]
matthew-carroll: More important question first - do we set a block type for images and horizontal rules?

super_editor/lib/src/infrastructure/serialization/html/html_table.dart
matthew-carroll: Did you look into whether it's more correct/appropriate to apply that style/attribute to the <td> vs applying it to a <p> that surrounds the content in the cell?

super_editor/lib/src/default_editor/table.dart
```diff
@@ -0,0 +1,130 @@
...
+
+/// A [DocumentNode] that represents a read-only block table.
+///
+/// Being a block node means that the table is either fully selected or not selected at all,
+/// i.e., there is no selection of individual cells.
+@immutable
+class TableBlockNode extends BlockNode {
+  /// Creates a [TableBlockNode] with the given [cells].
+  ///
+  /// [cells] is a list of rows, and each row is a list of [TextNode]s
```
matthew-carroll, line 17: Rather than describe lists of lists, a user can much more quickly understand what you mean if you say something like "The [cells] grid is indexed as `cells[row][column]`." Says the same thing, but with half the words, and focuses on how to interact with `cells` instead of its technical structure.

super_editor/lib/src/default_editor/table.dart
```diff
@@ -0,0 +1,130 @@
...
+    required List<List<TextNode>> cells,
+    super.metadata,
+  }) : _cells = List.from(cells.map((row) => List<TextNode>.from(row))) {
+    initAddToMetadata({NodeMetadata.blockType: tableBlockAttribution});
+  }
+
+  @override
+  final String id;
+
+  int get rowCount => _cells.length;
```
matthew-carroll, line 30: Please add a `columnCount`. Also, I think I would move `_cells` up to be under `id` - otherwise we have this kind of strange division between `rowCount` and `columnCount` vs `getRow()`, `getColumn()`, etc.

super_editor/lib/src/default_editor/table.dart
```diff
@@ -0,0 +1,130 @@
...
+          id == other.id &&
+          metadata == other.metadata &&
+          const DeepCollectionEquality().equals(_cells, other._cells);
+
+  @override
+  int get hashCode => id.hashCode ^ _cells.hashCode ^ metadata.hashCode;
+}
+
+const tableBlockAttribution = NamedAttribution("tableBlock");
+const tableHeaderAttribution = NamedAttribution("tableHeader");
```
matthew-carroll, line 130: Do other similar attributions end with `Attribution`? E.g., paragraph, header, code?

super_editor/lib/src/infrastructure/serialization/html/html_table.dart
```diff
@@ -0,0 +1,114 @@
+import 'package:flutter/material.dart';
+import 'package:super_editor/src/core/document.dart';
+import 'package:super_editor/src/default_editor/selection_upstream_downstream.dart';
+import 'package:super_editor/src/default_editor/table.dart';
+import 'package:super_editor/src/default_editor/text.dart';
+import 'package:super_editor/src/infrastructure/serialization/html/html_inline_text_styles.dart';
+
+String? defaultTableToHtmlSerializer(
```
matthew-carroll, line 8: Dart Doc - especially given that we only serialize a subset of table internal tags.

super_editor/test/infrastructure/serialization/html/document_to_html_test.dart
```diff
@@ -278,6 +278,428 @@ void main() {
...
+            '<td><strong>Value 1.1</strong></td>',
+            '<td><i>Value 1.2</i></td>',
+            '</tr>',
+            '</tbody>',
+            '</table>',
+          ].join(),
+        );
+      });
+
+      test("with multiple header rows", () {
```
matthew-carroll, line 510: Let's move this up beneath the first header row test

super_editor/test/infrastructure/serialization/html/document_to_html_test.dart
```diff
@@ -278,6 +278,428 @@ void main() {
...
+            '<td>Value 1.1</td>',
+            '<td>Value 1.2</td>',
+            '</tr>',
+            '</tbody>',
+            '</table>',
+          ].join(),
+        );
+      });
+
+      test("with header cell inside data row", () {
```
matthew-carroll, line 580: Let's move this up above the alignment test, so we can gather all the variations of header row, body row, header cell in one area.

super_editor/test/infrastructure/serialization/html/document_to_html_test.dart
```diff
@@ -278,6 +278,428 @@ void main() {
...
+            '<td>Value 2.1</td>',
+            '<th>Value 2.2</th>',
+            '</tr>',
+            '</tbody>',
+            '</table>',
+          ].join(),
+        );
+      });
+
+      test("rows after the first data row cannot be added to <thead>", () {
```
matthew-carroll, line 632: This test should be renamed. It's stated like a mutation. The reality is that the user can't do anything to a table after it's created. Please find a phrasing that refers to how parsing works, which is what's being tested.

Also, please move this up to be closer to the most closely related test.

super_editor/test/infrastructure/serialization/html/document_to_html_test.dart
```diff
@@ -278,6 +278,428 @@ void main() {
...
+            '<th>Column 2</th>',
+            '</tr>',
+            '</thead>',
+            '<tbody>',
+            '<tr>',
+            '<td>Value 1</td>',
+            '<td>Value 2</td>',
+            '</tr>',
+            '<tr>',
+            '<th>Value 3</th>',
```
matthew-carroll, line 693: I see that this test creates a full header row beneath a body row. That's a good case to test. But I think we're still missing one test case, which is a table whose first row is all headers, but its second row is only partially headers. That's probably an important case to lock down to make sure we really are inspecting each entire row at the beginning.

super_editor_markdown/lib/src/document_to_markdown_serializer.dart
```diff
@@ -160,30 +269,47 @@ class ParagraphNodeSerializer extends NodeTypedDocumentNodeMarkdownSerializer<Pa
...
+        ? node.text.copyText(textSelection.start, textSelection.end)
+        : node.text;
+
     final buffer = StringBuffer();
 
     final Attribution? blockType = node.getMetadataValue('blockType');
 
     if (blockType == header1Attribution) {
-      buffer.write('# ${node.text.toMarkdown()}');
+      buffer.write('# ${textToConvert.toMarkdown()}');
```
matthew-carroll, line 295: Seems like we're calling `toMarkdown()` is a whole bunch of places. Should we just call that one time up above?

super_editor/test/infrastructure/serialization/html/document_to_html_test.dart
angelosilvestre: Modified it a little bit. Please take a look if it's better.

super_editor/test/infrastructure/serialization/html/document_to_html_test.dart
angelosilvestre: Added a test for that.

super_editor/lib/src/infrastructure/serialization/html/html_table.dart
angelosilvestre: Added.

super_editor/lib/src/default_editor/table.dart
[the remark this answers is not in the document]
angelosilvestre: Yeah, we set it for those nodes.

super_editor/lib/src/infrastructure/serialization/html/html_table.dart
angelosilvestre: It seems to be appropriate to apply directly to the `td`.

https://www.w3schools.com/css/css_table_align.asp

super_editor_clipboard/lib/src/document_copy.dart
```diff
@@ -20,9 +21,10 @@ extension RichTextCopy on Document {
       inlineSerializers: SuperEditorClipboardConfig.inlineHtmlSerializers,
     )));
 
-    // Serialize a backup copy in plain text so that this clipboard content
-    // can be pasted into plain-text apps, too.
-    item.add(Formats.plainText(toPlainText(selection: selection)));
+    // Serialize to Markdown as a plain text representation of rich text.
+    item.add(Formats.plainText(
```
matthew-carroll, line 25: Let's actually make 2 of these methods so we can still retain plain text for users who want it:

```
copyAsRichTextWithPlainTextFallback() {}

copyAsRichTextWithMarkdownFallback() {}
```

super_editor_markdown/lib/src/document_to_markdown_serializer.dart
```diff
@@ -65,16 +117,24 @@ abstract class NodeTypedDocumentNodeMarkdownSerializer<NodeType> implements Docu
   const NodeTypedDocumentNodeMarkdownSerializer();
 
   @override
-  String? serialize(Document document, DocumentNode node) {
+  String? serialize(
```
matthew-carroll, line 120: Would it make sense to add a generic type to this class for `SelectionType` and do the selection gating here? We can also ask `selection.isCollapsed` here, too. Then I think you can remove those additions from each individual serializer.

matthew-carroll: Eh, maybe in the future. Probably not worth trying to figure that out right now. I'd have to really think things through to ensure that we can guarantee that every possible internal node selection has a concept of a collapsed selection.

PR #2784, opened 2025-08-27
[SuperEditor] - Fix: Use inline widget builders in hint component (Resolves #2783)

### LINKED ISSUE #2783: [SuperEditor] - Fix: Pass inline widget builders into `HintTextComponent`
It looks like `HintTextComponent` is never provided with inline widget builders, and therefore breaks any use-case that wants to build inline widgets in a paragraph that includes a hint.

Pipe through the inline widget builders from the view model.

matthew-carroll:
[SuperEditor] - Fix: Use inline widget builders in hint component (Resolves #2783)

Also, delete the hint component builder that we defined within the test directory because we publish a hint component builder now.


angelosilvestre (commented):
LGTM

PR #2787, opened 2025-08-27
[AttributedText] - Make constructor agnostic to placeholder order in placeholder map (Resolves #2786)

### LINKED ISSUE #2786: [AttributedText][Bug] - Wrong placeholder map order breaks constructor
It looks like our `AttributedText` constructor is order-dependent on the entries in the placeholder map. This is undocumented and unintentional. Make the constructor order agnostic.

matthew-carroll:
[AttributedText] - Make constructor agnostic to placeholder order in placeholder map (Resolves #2786)

Before this PR, if someone passes in an inline placeholder map into `AttributedText`, and the entries in that map aren't in the same order as the placeholder text positions, the constructor will blow up. This PR sorts them in the constructor to avoid that issue.

### DIFF
--- a/attributed_text/lib/src/attributed_text.dart
+++ b/attributed_text/lib/src/attributed_text.dart
@@ -70,7 +70,10 @@ class AttributedText {
       final buffer = StringBuffer();
       int start = 0;
       int insertedPlaceholders = 0;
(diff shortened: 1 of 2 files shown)


angelosilvestre (commented):
LGTM

PR #2799, opened 2025-09-25
[SuperEditor][SuperReader][SuperTextField] Update Android toolbar. (Resolves #2791)

### LINKED ISSUE #2791: [SuperEditor][SuperReader][SuperTextField] - Update Android popover toolbar to match OS
It looks like Android may have (somewhat recently) changed or unified their popover toolbar UI. We should update the Super Editor Android default toolbar to match.

The new toolbar is initially pill shaped, with non bold text. There's no visual division between buttons, but when pressed, the button highlight area is rectangular.

Image

The toolbar often has a "3 dots" expansion button. When that button is pressed, the horizontal pill-shaped toolbar animates to a rectangle with a vertical list of the items that weren't originally visible on the toolbar.

Image

We want to make sure we match the following:
 * Shape (including both the pill shape and the height)
 * Button text styles
 * Button press highlight area
 * Color scheme (which is probably tied to something in the app's theme and/or brightness)
 * Animation from a horizontal list to a vertical overflow menu (also, pay attention to when/why the overflow menu appears above the selection vs below the selection, and check what happens in situations with limited space)

angelosilvestre:
[SuperEditor][SuperReader][SuperTextField] Update Android toolbar. (Resolves #2791)

### DIFF
--- a/super_editor/lib/src/infrastructure/platforms/android/toolbar.dart
+++ b/super_editor/lib/src/infrastructure/platforms/android/toolbar.dart
@@ -1,9 +1,14 @@
-import 'package:flutter/material.dart';
-import 'package:super_editor/src/infrastructure/platforms/android/colors.dart';
+import 'dart:math' as math;
 
-class AndroidTextEditingFloatingToolbar extends StatelessWidget {
+import 'package:flutter/foundation.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter/rendering.dart';
+import 'package:follow_the_leader/follow_the_leader.dart';
+
+class AndroidTextEditingFloatingToolbar extends StatefulWidget {
   const AndroidTextEditingFloatingToolbar({
     Key? key,
+    this.focalPoint,
     this.floatingToolbarKey,
     this.onCutPressed,
     this.onCopyPressed,
@@ -12,6 +17,7 @@ class AndroidTextEditingFloatingToolbar extends StatelessWidget {
   }) : super(key: key);
 
   final Key? floatingToolbarKey;
+  final LeaderLink? focalPoint;
 
   final VoidCallback? onCutPressed;
   final VoidCallback? onCopyPressed;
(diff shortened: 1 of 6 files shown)


super_editor/test/super_editor/mobile/super_editor_android_overlay_controls_test.dart
```diff
@@ -724,5 +724,8 @@ Future<TestDocumentContext> _pumpSingleParagraphApp(WidgetTester tester) async {
       .createDocument()
       // Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor...
       .withSingleParagraph()
+      // It seems that when we have the toolbar visible, we are tapping the toolbar
```
matthew-carroll, line 727: Why is this the solution? Are we tapping the toolbar instead of the handle because the toolbar is pushed over the handle? If so, that's a real possible problem, right? What happens if a user wants to grab a ahndle but it's covered by the toolbar?

super_editor/lib/src/infrastructure/platforms/android/toolbar.dart
```diff
@@ -12,73 +17,835 @@ class AndroidTextEditingFloatingToolbar extends StatelessWidget {
...
-        child: Text(
-          title,
-          style: const TextStyle(fontSize: 14),
+const double _kToolbarHeight = 54.0;
+
+Widget _defaultToolbarBuilder(BuildContext context, Widget child) {
+  return _TextSelectionToolbarContainer(child: child);
+}
+
+/// A toolbar containing the given children. If they overflow the width
```
matthew-carroll, line 138: Let's move this to its own file in our flutter directory where we put other Flutter stuff. Also, is this the same name that Flutter used for it? Seems like a very generic and ambiguous name. Is there any reason not to call this something like `AndroidPopoverToolbar`? Or `AndroidTextFormattingToolbar`?

matthew-carroll:
Also, another question for this - what about theming? Do we get dark mode and light mode? Also, does Android impose any particular colors that we want to match?

angelosilvestre:
> Also, another question for this - what about theming? Do we get dark mode and light mode? Also, does Android impose any particular colors that we want to match?

Yes, we get light and dark mode. I updated the colors to match the colors from the API level 35 emulator.

PR #2919, opened 2026-02-16
[Super Editor Clipboard] - Fix native paste due to mis-use of super_clipboard API

matthew-carroll:
[Super Editor Clipboard] - Fix native paste due to mis-use of super_clipboard API

I'm not sure if there's a bug in `super_clipboard` or I was mis-using the API or both. Regardless, this PR adjusts our custom paste method to call `super_clipboard` in a slightly different way, which should fix a problem of paste simply not doing anything in one of our client apps.

### DIFF
--- a/super_editor_clipboard/lib/src/super_editor_paste.dart
+++ b/super_editor_clipboard/lib/src/super_editor_paste.dart
@@ -136,12 +136,15 @@ Future<void> pasteIntoEditorFromNativeClipboard(
   Editor editor, {
   CustomPasteDataInserter? customInserter,
 }) async {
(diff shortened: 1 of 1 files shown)


matthew-carroll:
Tests are failing because of a Flutter breaking change: https://github.com/Flutter-Bounty-Hunters/super_editor/pull/2917 - so I'm ignoring them.

For some reason I can't seem to resolve this PR hash as a dependency in the client app to verify the fix. I'm gonna have to merge to `main` and try to verify from there. As a result, there may be followup corrections to this.

PR #2940, opened 2026-02-25
[super_editor] super_message | contentTapDelegate not passed to SuperMessageAndroidTouchInteractor

piotrmoskala:
Problem: 
- Tap gestures are not being registered on Android only.
Solution: 
- Pass _contentTapDelegate to SuperMessageAndroidTouchInteractor (which is being correctly passed to SuperMessageIosTouchInteractor)

### DIFF
--- a/super_editor/lib/src/chat/super_message.dart
+++ b/super_editor/lib/src/chat/super_message.dart
@@ -373,6 +373,7 @@ class _SuperMessageState extends State<SuperMessage> {
                 focusNode: _focusNode,
                 editor: widget.editor,
                 getDocumentLayout: () => _messageContext.documentLayout,
+                contentTapHandlers: _contentTapDelegate != null ? [_contentTapDelegate!] : null,
                 showDebugPaint: widget.debugPaint.gestures,
                 child: SuperMessageAndroidControlsOverlayManager(
                   editor: widget.editor,


matthew-carroll:
@piotrmoskala I took over this PR to get it finished ASAP. Here are some notes and the new PR:

**Tests:** Unless you're dealing with OS or Flutter framework behaviors that can't be reasonably simulated, we expect that every PR include tests.

`SuperEditor` already had tests for this behavior, so it would have been trivial to add tests to this PR. You just needed to copy those tests to a new file in the right directory and then make some adjustments.

**Consistent behaviors:** Once I copied over the `SuperEditor` tests for `SuperMessage`, I found that they were failing. This is because `SuperEditor` includes a link clicking handler by default, but `SuperMessage` doesn't. `SuperMessage` should behave similar to `SuperEditor` unless there's a specific reason not to. So this was, in fact, a bug, which was made clear by the tests. In the new PR, I updated `SuperMessage` to include that link click handler by default.

**Consistent APIs:** After the tests showed me that we needed to add the link handler by default, I noticed a mismatch in API between `SuperMessage` and `SuperEditor`. `SuperEditor` takes a priority list of handlers, but `SuperMessage` only takes one handler. I don't remember why I did that. Maybe it was just an oversight. So I changed `SuperMessage` to match `SuperEditor` and take a priority list of handlers. I updated all `SuperMessage` implementation details accordingly, and then the tests passed.

**No shortcuts:** The original version of this PR added what appeared to be a missing argument to one of the interactors in `SuperMessage`. While the argument was definitely missing, that was only the beginning of the story. When you added it, you had to put it in a list to get it to compile. This should have been a red flag and investigated. Why does `SuperMessage` take a single handler when the interactor takes a list? Why does the Android interactor take a list, but the iOS interactor does not? It's critical that these questions be asked, and answered, when contributing to `super_editor`. I would argue that such things must be done when contributing to any package that wants to survive long term. Any PR you open where you don't do that, will require that I do it, so please make sure you go all the way down any rabbit hole you come across. You can ask me any questions you have while doing so.

Updated PR:
https://github.com/Flutter-Bounty-Hunters/super_editor/pull/2973

PR #3010, opened 2026-05-07
[AttributedText] Improve performance of getAttributionSpansInRange

knopp:
`AttributedSpans.getAttributionSpansInRange` frequently appears in a hot codepath and current implementation seems to have cubic complexity. This PR replaces it with simpler implementation that only traverses the attributes once.


knopp:
The updated test originally seemed to rely on ordering within the set which doesn't seem like behavior that we should depend on.

knopp:
`AttributedSpans.getMatchingAttributionsWithin` is also showing in profiler trace but it's much less prevalent. It might benefit from similar treatment.

matthew-carroll (approved):

PR #3019, opened 2026-05-26
[SuperEditor] - Update ime integration to work with custom selection types, and add AttachmentListNode as proof of concept

matthew-carroll:
[SuperEditor] - Update ime integration to work with custom selection types, and add AttachmentListNode as proof of concept


matthew-carroll:
I finished this work against a different stable branch, merged it, and then cherry picked to stable and main. Closing this.

PR #3044, opened 2026-06-11
Cherry Pick (to main): [SuperEditor][Markdown] - Parse images with captions immediately above them (Resolves #3041) (#3042)

### LINKED ISSUE #3041: [SuperEditor][Markdown] - Parse text immediately preceding an image as an image caption
In a client app, we have Markdown where image captions are placed on the line directly above an image. In our current parser, this seems to be lost, or otherwise messes up the image parsing.

We should update our syntax parser to handle a line of text above image syntax.

Examples that we should support:

    A caption:
    ![a]

and

    A caption:
    ![a]
    B caption:
    ![b]

matthew-carroll:
Cherry Pick (to main): [SuperEditor][Markdown] - Parse images with captions immediately above them (Resolves #3041) (#3042)

### DIFF
--- a/super_editor/lib/src/default_editor/tables/table_markdown.dart
+++ b/super_editor/lib/src/default_editor/tables/table_markdown.dart
@@ -49,7 +49,7 @@ class MarkdownTableComponentBuilder implements ComponentBuilder {
                 nodeId: cell.id,
                 createdAt: cell.metadata[NodeMetadata.createdAt],
                 text: cell.text,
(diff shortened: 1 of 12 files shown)


matthew-carroll:
It looks like this work, which we're cherry picking, legitimately created an issue with RTL paragraph alignment. It looks like we really do need to allow `null` `TextNode` `"textAlign"`. I'm going to merge this and follow up with a reversal on that.

PR #3045, opened 2026-06-11
[SuperEditor][Markdown] - Reverse part of the previous PR where we made every TextNode 'left' aligned by default, which broke RTL

matthew-carroll:
[SuperEditor][Markdown] - Reverse part of the previous PR where we made every TextNode 'left' aligned by default, which broke RTL


matthew-carroll:
The `super_editor_clipboard` tests were failing. I fixed them when compiling against the local `super_editor`, but it looks like we need to release a new version of `super_editor` to upgrade the dep in `super_editor_clipboard` to get the tests passing again.

Issue #2650, opened 2025-04-29
[BUG] - Cannot applyed markdown styling because of the space character in the end of the text


ismail-go:
Hi, @matthew-carroll  I have a problem in the serialization of the markdown styling.

If the markdown styled text has a space in the begining or end of the text it does not apply markdown styling to the text correctly.

For example;
**bold text** -> applyed bold styling correctlty but,
**bold text ** -> it seems as it is in the markdown displaying. Cannot applying the markdown styling.

as it is in bold styling, it is also works wrong in the italic, strikethrough etc..

We need to handle this because of the using keyboard suggested words. it puts space character to the end of the word.

How can we solve the problem? @matthew-carroll

matthew-carroll:
What version of super_editor and super_editor_markdown are you using? Does this happen with an English keyboard? Can you show example code that directly converts a MutableDocument to markdown which demonstrates this problem?

ismail-go:
Flutter (Channel stable, 3.29.3, on macOS 15.0.1 24A348 darwin-arm64, locale en-TR) [1,438ms]
• Flutter version 3.29.3 on channel stable at /Users/ismail/Flutter/Sdk/flutter
• Upstream repository https://github.com/flutter/flutter.git
• Framework revision ea121f8859 (3 weeks ago), 2025-04-11 19:10:07 +0000
• Engine revision cf56914b32
• Dart version 3.7.2
• DevTools version 2.42.3

super_editor_markdown:
git:
url: https://github.com/superlistapp/super_editor.git
path: super_editor_markdown

used last version of the super editor

You can reproduce the bug event like that. 
For example, Firstly select bold style from keyboard editor toolbar and try to text and put the space character end of the bold styled sentence. When you save the changes and rebuild the markdown editor the markdown styling broke down.

ismail-go:
Yes, of course! @ibrahimdevs @matthew-carroll 

https://github.com/user-attachments/assets/8149dabb-9a1f-4a03-a2f0-8d4da275e3cc

import 'dart:io';

import 'package:flutter/foundation.dart';
import 'package:flutter/material.dart';
import 'package:path_provider/path_provider.dart';
import 'package:super_editor/super_editor.dart';
import 'package:super_editor_markdown/super_editor_markdown.dart';

void main() {
  runApp(MyApp());
}

class MyApp extends StatelessWidget {
  @override
  Widget build(BuildContext context) {
    return MaterialApp(
[...]

PR #3079, opened 2026-07-05
[Super Editor] - Markdown: Serialize inline styles to valid CommonMark (Related to #2424, #2650)

donaldheppner:
## What this does

Rewrites `AttributedTextMarkdownSerializer` so that its Markdown output re-parses to the same styled text. Today, several common inputs serialize to Markdown that no CommonMark parser (including this package's own parser) reads back with the same styling, so styles silently vanish or change on the next load.

- **Whitespace at span edges** — a bold span that includes a trailing space serializes as `**bold ** text`, which is not valid CommonMark emphasis (the closing delimiter can't be preceded by whitespace, per [CommonMark spec §6.2, left-/right-flanking delimiter runs](https://spec.commonmark.org/0.31.2/#left-flanking-delimiter-run)). The serializer now moves edge whitespace outside the markers: `**bold** text`. This is easy to hit in practice: mobile keyboard word suggestions insert a trailing space, which ends up inside the styled span (#2650). Related to #2424 and the earlier attempt in #2452 — the renderer comparison table in #2452 (GitHub, flutter_markdown, VSCode, Typora, StackEdit, Dillinger all refuse space-adjacent delimiters) covers the "is this really invalid?" question that stalled that PR.
- **Backslash escapes** — parsing `3\*4` now records a `markdownEscape` attribution on the `*`, and serialization re-emits the backslash. Previously the escape was destroyed on the first round trip and the bare `*` could become an emphasis delimiter on the next parse.
- **Overlapping (non-nested) spans** — serialized by closing and re-opening styles so markers nest properly, and the output re-parses to the same attributions. Styles opening at the same offset open longest-span-first. Adjacent same-style spans merge instead of emitting `**a****b**`.
- **Safety net** — the serializer re-parses its own output and compares attributions; if the styles don't survive, it backslash-escapes markdown-significant characters and uses that form instead (protects plain text like `3*4 and 5*6` from gaining emphasis on the next parse).
- **Block-trigger escaping (opt-in)** — plain paragraphs serialize with line-leading block syntax escaped (`\# not a heading`), so a paragraph that merely starts with `#`, `- `, `1. `, `>`, a fence, or a thematic break doesn't change block type on the next parse. Node serializers that supply their own block syntax (headers, list items, table cells) leave this off.
- **Abutting spans coalesce before trimming** — editing can leave one bold run stored as two spans meeting at a space (toggle bold off and back on over part of a run); trimming each span separately would silently un-style that internal space. Runs are coalesced first, so only the true edges of a styled run are trimmed.

Four existing test expectations encoded the invalid serializations (`text** bold**`, `**a****b**`, `**Fixed **` — that last one, in `ime_ios_exceptional_cases_test.dart`, is exactly the #2650 keyboard-suggestion scenario) and are updated to the valid forms, which re-parse to the same styled text.

## What this doesn't change

Strikethrough still serializes as single `~` and underline as `¬` in this PR; moving those to GFM `~~` / `<u>` is split into the follow-up PR so this one stays reviewable.

## Tests

New round-trip cases in `attributed_text_markdown_test.dart` covering each bullet above. `flutter test` passes for the package.

## Context

Extracted from [MemNote](https://memnote.randomfact.com)'s super_editor fork, where these fixes shipped after we traced note corruption to serialize/parse round trips (every save/load cycle compounded the damage).

*This is PR 1 of a 6-PR series upstreaming our Markdown codec fixes; each is independently revertable but they apply in order. I'll rebase the rest as predecessors merge.*

### DIFF
--- a/super_editor/lib/src/infrastructure/serialization/markdown/document_to_markdown_serializer.dart
+++ b/super_editor/lib/src/infrastructure/serialization/markdown/document_to_markdown_serializer.dart
@@ -2,6 +2,7 @@ import 'dart:ui';
 
 import 'package:attributed_text/attributed_text.dart';
 import 'package:flutter/foundation.dart';
+import 'package:markdown/markdown.dart' as md;
 import 'package:super_editor/src/core/document.dart';
 import 'package:super_editor/src/core/document_selection.dart';
 import 'package:super_editor/src/default_editor/attributions.dart';
@@ -13,6 +14,7 @@ import 'package:super_editor/src/default_editor/selection_upstream_downstream.da
 import 'package:super_editor/src/default_editor/tables/table_block.dart';
 import 'package:super_editor/src/default_editor/tasks.dart';
 import 'package:super_editor/src/default_editor/text.dart';
+import 'package:super_editor/src/infrastructure/serialization/markdown/markdown_inline_parser.dart';
 import 'package:super_editor/src/infrastructure/serialization/markdown/super_editor_syntax.dart';
 
 /// Serializes the given [doc] to Markdown text.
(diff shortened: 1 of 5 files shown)


donaldheppner:

PR #3080, opened 2026-07-05
[Super Editor] - Markdown: Serialize strikethrough as GFM "~~" and underline as "<u>", parsing legacy forms

donaldheppner:
## What this does

Serializes strikethrough and underline to standard Markdown forms, while still parsing the legacy proprietary forms:

- **Strikethrough** serializes as GFM `~~text~~` instead of single `~text~`. Single-tilde is not CommonMark/GFM — GitHub, VSCode, and most renderers don't strike it through. The parser keeps accepting the legacy single-tilde form (`SingleStrikethroughSyntax` is unchanged), and legacy content heals to `~~` on the next serialization.
- **Underline** serializes as `<u>text</u>` instead of the proprietary `¬text¬` marker. Inline HTML tags are the only widely supported way to express underline in Markdown; `¬` renders as a literal not-sign everywhere outside Super Editor. A new `HtmlUnderlineSyntax` parses `<u>...</u>` (including nested styles like `<u>**bold**</u>`), and the legacy `¬` form is still parsed, healing to `<u>` on the next serialization.

Both changes are one-time, idempotent normalizations of stored content: old documents load unchanged and re-save in the standard form.

## Behavior change

Consumers that post-process serialized Markdown and expect `~`/`¬` will see `~~`/`<u>` instead. Both forms parse back identically.

## Tests

Round-trip and healing cases in `attributed_text_markdown_test.dart` and `super_editor_markdown_test.dart`.

*PR 2 of the Markdown codec series — stacked on #3079; review the last commit. Extracted from MemNote's super_editor fork.*

### DIFF
--- a/super_editor/lib/src/infrastructure/serialization/markdown/document_to_markdown_serializer.dart
+++ b/super_editor/lib/src/infrastructure/serialization/markdown/document_to_markdown_serializer.dart
@@ -2,6 +2,7 @@ import 'dart:ui';
 
 import 'package:attributed_text/attributed_text.dart';
 import 'package:flutter/foundation.dart';
+import 'package:markdown/markdown.dart' as md;
 import 'package:super_editor/src/core/document.dart';
 import 'package:super_editor/src/core/document_selection.dart';
 import 'package:super_editor/src/default_editor/attributions.dart';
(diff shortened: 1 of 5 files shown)


donaldheppner:

PR #3081, opened 2026-07-05
[Super Editor] - Markdown: Preserve code fence language and serialize fence content literally (Resolves #3006)

### LINKED ISSUE #3006: [BUG] - [Markdown] - Fenced code blocks accumulate trailing whitespace on every parse → serialize round-trip
**Package Version**
super_editor, GitHub `main` branch (verified at commit `291a6930`, one commit ahead of tag `super_editor_v0.3.0-dev.51`). Also reproduces in published `super_editor_markdown 0.2.0` shipped alongside `super_editor 0.3.0-dev.40` — the markdown code was consolidated into `super_editor` itself in #2842, and the buggy path is byte-identical between the two locations.

**User Info**
individual

**To Reproduce**

1. Create a Dart/Flutter project that depends on `super_editor` from `main` (or on published `super_editor_markdown 0.2.0`).
2. Run the minimal reproduction code below.
3. Compare the printed output to the input.

**Minimal Reproduction Code**

<details>

  ````dart

  void main() {
    const input = '```\nline1\nline2\nline3\n```';
    final output =
        serializeDocumentToMarkdown(deserializeMarkdownToDocument(input));
    print(output);
    print('---');
    print('input == output? ${output == input}');
  }
  ````
</details>

**Actual behavior**

The output is **not** byte-identical to the input. Every line of the fenced code block has gained two trailing spaces, and a spurious blank line has appeared before the closing fence:

````
```
line1··
line2··
line3··

```
````

*(where `··` denotes two literal trailing space characters)*

A second round-trip stacks another `  ` on top of every line, so the damage compounds indefinitely. Over ~50 open/save cycles, individual code-block lines accumulated ~100 trailing spaces each, producing visible churn in version-control diffs and a "dirty file on open" experience.

**Expected behavior**

The output should be byte-identical to the input.

**Platform**

Platform-independent. Reproduced via pure Dart in `flutter_test`. Originally surfaced on Windows 11.

**Flutter version**

Flutter 3.41.6 stable, Dart 3.11.4. The bug is pure Dart logic and not Flutter-version-dependent.

**Screenshots**

N/A

**Additional context**

We analyzed the serializer and parser and confirmed this is a serializer bug, not a parser bug. Between `deserializeMarkdownToDocument` and `serializeDocumentToMarkdown`, the parsed `ParagraphNode`'s text is stored cleanly as `"line1\nline2\nline3\n"` — no `  ` has been inserted anywhere.

The damage happens in `ParagraphNodeSerializer.doSerialization` in `super_editor/lib/src/infrastructure/serialization/markdown/document_to_markdown_serializer.dart`. The `codeAttribution` branch routes the code node's text through `AttributedTextMarkdownSerializer.serialize` via `node.text.toMarkdown()`:

```dart
} else if (blockType == codeAttribution) {
  buffer //
    ..writeln('```') //
    ..writeln(inlineMarkdown) //
    ..write('```');
}
```

That attribution-aware path is fine for paragraphs but wrong here, because `AttributedTextMarkdownSerializer._writeTextToBuffer` unconditionally inserts the markdown hard-break marker (`  \n`) before every embedded `\n`:

```dart
void _writeTextToBuffer(String text) {
  final lines = text.split('\n');
  for (int i = 0; i < lines.length; i++) {
    if (i > 0) {
      // Adds two spaces before line breaks.
      _buffer.write('  ');
      _buffer.write('\n');
    }
    _buffer.write(lines[i]);
  }
}
```

That `  ` marker is the correct convention for paragraph text — a line ending in `  \n` is a CommonMark hard line break — but it is incorrect inside a fenced code block, where per CommonMark §4.5 the contents are literal and hard-break syntax does not apply. So on every round-trip, each embedded `\n` gains another `  ` in front of it, and the damage compounds indefinitely.

We also noticed a second, smaller defect in the same branch: the parser preserves the trailing `\n` between the last code line and the closing fence as part of the node's text, and `writeln(inlineMarkdown)` above then doubles it, producing a spurious blank line before the closing fence on every round-trip.

**Proposed Solution**

Bypass `AttributedTextMarkdownSerializer` entirely for code blocks. Code content has no valid inline attributions, so the attribution-aware path has nothing to contribute and its hard-break convention is actively harmful. Use `toPlainText()` and strip the single trailing newline before handing the body to `writeln`:

```dart
} else if (blockType == codeAttribution) {
  // Use the raw plain text for code blocks. Markdown hard-break markers
  // ("  \n") and inline style markers are not valid inside fenced code
  // blocks per CommonMark §4.5, so routing code content through
  // AttributedTextMarkdownSerializer would inflate trailing whitespace
  // on every round-trip.
  final codeText = (textSelection != null //
          ? node.text.copyText(textSelection.start, textSelection.end)
          : node.text)
      .toPlainText();
  // Strip a single trailing newline so writeln('```') below does not
  // produce a spurious blank line before the closing fence on round-trip
  // (the parser preserves the newline between the last code line and
  // the closing fence as part of the node's text).
  final codeBody = codeText.endsWith('\n') //
      ? codeText.substring(0, codeText.length - 1)
      : codeText;
  buffer //
    ..writeln('```') //
    ..writeln(codeBody) //
    ..write('```');
}
```

This replacement handles both defects: the hard-break inflation (via `toPlainText()` bypassing `AttributedTextMarkdownSerializer`) and the spurious blank line (via the `endsWith('\n')` strip).

**Test coverage**

We verified the fix against the following regression tests, all passing post-fix:

1. `fenced code block survives parse->serialize unchanged` — byte-identical single round-trip.
2. `two round-trips do not compound trailing whitespace inside a fenced code block` — guards specifically against the compounding behavior.
3. `non-code paragraph still emits hard-break marker` — regression guard ensuring the fix does not touch normal paragraph serialization (the `  \n` convention is correct there).

donaldheppner:
## What this does

Fixes fenced code blocks corrupting on every serialize/parse round trip (Resolves #3006):

- **Fence content is serialized literally** via `toPlainText()` instead of routing through the inline Markdown serializer. The inline path appended two-space hard-break markers to every line — one more `"  "` per line per round trip, compounding exactly as described in #3006 — and backslash-escaped code content.
- **The parser's trailing newline is stripped** when serializing, so the fence doesn't grow a spurious blank line before the closing ` ``` ` on every round trip (the second symptom in #3006). This matches the fix proposed in the issue body.
- **The fence info string survives**: parsing ` ```dart ` stores `dart` in the code node's `codeLanguage` metadata (from the `language-` class the block parser produces), and serialization re-emits it. Previously the language was silently dropped on the first round trip.

## Tests

New serialization and deserialization cases in `super_editor_markdown_test.dart`: language preservation both directions, literal (unescaped) fence content, and `codeLanguage` metadata absent for bare fences.

*PR 3 of the Markdown codec series — stacked on #3079 and #3080; review the last commit. Extracted from MemNote's super_editor fork.*

### DIFF
--- a/super_editor/lib/src/infrastructure/serialization/markdown/document_to_markdown_serializer.dart
+++ b/super_editor/lib/src/infrastructure/serialization/markdown/document_to_markdown_serializer.dart
@@ -2,6 +2,7 @@ import 'dart:ui';
 
 import 'package:attributed_text/attributed_text.dart';
 import 'package:flutter/foundation.dart';
+import 'package:markdown/markdown.dart' as md;
 import 'package:super_editor/src/core/document.dart';
 import 'package:super_editor/src/core/document_selection.dart';
 import 'package:super_editor/src/default_editor/attributions.dart';
@@ -13,6 +14,7 @@ import 'package:super_editor/src/default_editor/selection_upstream_downstream.da
 import 'package:super_editor/src/default_editor/tables/table_block.dart';
 import 'package:super_editor/src/default_editor/tasks.dart';
 import 'package:super_editor/src/default_editor/text.dart';
+import 'package:super_editor/src/infrastructure/serialization/markdown/markdown_inline_parser.dart';
 import 'package:super_editor/src/infrastructure/serialization/markdown/super_editor_syntax.dart';
 
 /// Serializes the given [doc] to Markdown text.
@@ -298,10 +300,10 @@ class ParagraphNodeSerializer extends NodeTypedDocumentNodeMarkdownSerializer<Pa
 
     final Attribution? blockType = node.getMetadataValue('blockType');
 
-    final inlineMarkdown = (textSelection != null //
-            ? node.text.copyText(textSelection.start, textSelection.end)
-            : node.text)
-        .toMarkdown();
+    final textToConvert = textSelection != null //
+        ? node.text.copyText(textSelection.start, textSelection.end)
+        : node.text;
+    final inlineMarkdown = textToConvert.toMarkdown();
 
     if (blockType == header1Attribution) {
       buffer.write('# $inlineMarkdown');
@@ -332,7 +334,10 @@ class ParagraphNodeSerializer extends NodeTypedDocumentNodeMarkdownSerializer<Pa
           buffer.writeln(alignmentToken);
         }
       }
-      buffer.write(inlineMarkdown);
+      // A plain paragraph has no block-level syntax of its own, so a line starting
+      // with "#", "- ", "1. ", etc., would change block type on the next parse.
+      // Serialize with those trigger characters escaped.
+      buffer.write(textToConvert.toMarkdown(escapeLineStartTriggers: true));
     }
 
     // We're not at the end of the document yet. Add a blank line after the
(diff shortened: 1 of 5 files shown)


donaldheppner:

Issue #2924, opened 2026-02-19
[BUG] - Empty line between unordered and task list items breaks parsing/formatting


timcreatedit:
**Package Version**
`super_editor: ^0.3.0-dev.48`

**User Info**
ClickUp

**To Reproduce**
When an empty line is included between some unordered list items and task list items, the parsing and formatting break down.

Task list items loose their completion state, but also some unwanted padding is added when rendering.

**Minimal Reproduction Code**
To reproduce visually

<details>
  <summary>Minimal, Runnable Code Sample</summary>

  ```dart
  import 'package:flutter/material.dart';
import 'package:super_editor/super_editor.dart';

void main() {
  runApp(const MainApp());
}

class MainApp extends StatefulWidget {
  const MainApp({super.key});

  @override
  State<MainApp> createState() => _MainAppState();
}

class _MainAppState extends State<MainApp> {
  bool emptyLine = false;

  late Editor editor = createDefaultDocumentEditor(
    document: deserializeMarkdownToDocument(markdown),
  );

  String get markdown =>
      '''
- Unordered item 1
- Unordered item 2${emptyLine ? '\n' : ''}
- [ ] Task item 1
- [x] Task item 2
''';

  void _updateDocument() {
    editor.dispose();
    editor = createDefaultDocumentEditor(
      document: deserializeMarkdownToDocument(markdown),
    );
  }

  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      home: Scaffold(
        body: SingleChildScrollView(
          child: Column(
            spacing: 24,
            children: [
              SwitchListTile(
                title: Text('Include empty line between list items'),
                value: emptyLine,
                onChanged: (value) {
                  setState(() {
                    emptyLine = value;
                    _updateDocument();
                  });
                },
              ),
              SuperMessage(
                editor: editor,
                componentBuilders: [...readOnlyDefaultComponentBuilders],
              ),
            ],
          ),
        ),
      ),
    );
  }
}

  ```
</details>

To reproduce parsing failure

<details>
  <summary>Minimal, Runnable Code Sample</summary>

  ```dart
import 'package:flutter_test/flutter_test.dart';
import 'package:super_editor/super_editor.dart';

void main() {
  group('document', () {
    setUp(() {});

    testWidgets('deserialize', (tester) async {
      final document = deserializeMarkdownToDocument('''
- Unordered item 1
- Unordered item 2
- [ ] Task item 1
- [x] Task item 2
[...]

PR #3082, opened 2026-07-05
[Super Editor] - Markdown: Serialize list items with real ordinals and round-trippable nesting indentation

donaldheppner:
## What this does

Makes list serialization round-trippable. Today, every serialize/parse cycle mutates lists:

- **Ordered items all serialize as `1.`** — real ordinals are lost. They now serialize with sequential ordinals (`1. 2. 3.`) computed from their position among consecutive same-indent ordered siblings.
- **Every item gets a 2-space base indent** (`  * item`), so top-level lists arrive indented and keep drifting. Top-level items now serialize with zero leading spaces.
- **Nested items use a fixed 2-space step**, which is not enough to nest under an ordered parent (`1. ` needs a 3-column indent) — the nesting silently flattens on the next parse. Nested items are now indented to their parent marker's content column (2 under `- `, 3 under `1. `, 4 under `10. `), computed from the ancestor run.
- **Unordered items serialize with a canonical `- ` marker** instead of `*` (a one-time, idempotent normalization; `*` input still parses).
- **A list item followed by a different-type list item no longer gets an extra blank line**, which re-parsed as two separate lists and reset ordered numbering (the serialization-side sibling of the parse problems discussed in #2924).

## Tests

Updated expectations plus new cases in `super_editor_markdown_test.dart`: a nested-ordered-list round trip asserting indent levels survive, and mixed-type nesting (`1.` / nested `-` / `2.`). Paste-test fixtures canonicalized to the new output.

*PR 4 of the Markdown codec series — stacked on #3079–#3081; review the last commit. Extracted from MemNote's super_editor fork.*

### DIFF
--- a/super_editor/lib/src/infrastructure/serialization/markdown/document_to_markdown_serializer.dart
+++ b/super_editor/lib/src/infrastructure/serialization/markdown/document_to_markdown_serializer.dart
@@ -2,6 +2,7 @@ import 'dart:ui';
 
 import 'package:attributed_text/attributed_text.dart';
 import 'package:flutter/foundation.dart';
+import 'package:markdown/markdown.dart' as md;
 import 'package:super_editor/src/core/document.dart';
 import 'package:super_editor/src/core/document_selection.dart';
 import 'package:super_editor/src/default_editor/attributions.dart';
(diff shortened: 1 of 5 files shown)


donaldheppner:

